PR #73, opened 2025-09-02
Bump Go to 1.25

dgageot:


dgageot:
Needs an updated version of the `golangci-lint-action` action

PR #105, opened 2025-09-05
:brain: :wrench: fix correct mem memory param in mem.yaml example

marcellodesales:
# 🐛 Current Bug

* The current version doesn't parse the yaml file, so no linting output.

# ✅ Fixed Demo

The correct version is specified in the docs

# 💻 Screenshots


rumpl (approved):
Thanks!

PR #116, opened 2025-09-05
Show the current version on the lower left corner in the status bar

rumpl:
Screenshot 2025-09-05 at 17 29 59


krissetto:
"lower right corner" i assume :)

rumpl:
> "lower right corner" i assume :)

oops

PR #117, opened 2025-09-05
Support for azure openai via the standard openai provider

### LINKED ISSUE #112: [FEAT] - Support Azure OpenAI in addition to OpenAI
**What you'd like to see**  

I would like to use Azure OpenAI instead of pure OpenAI

**Why you'd like to see it**  

That would allow me to use `cagent` in more contexts as Azure OpenAI has some of the bells and whistles expected in a corporate context

**Workarounds?**  

no

**Additional context**  

If I try to call my Azure OpenAI endpoint using the existing `openai` provider, I would do something like this

```
models:
  cloud-gpt-4o:
    provider: openai
    model: gpt-4o
    base_url: https://openai-testtf.openai.azure.com/openai/deployments/gpt-4o
```

But this causes an error 404

Image

The reason most likely is that Azure OpenAI expects a version query parameter. E.g. a `POST` to `https://openai-testtf.openai.azure.com/openai/deployments/gpt-4o/chat/completions?api-version=2025-01-01-preview` works, whereas a `POST` to `https://openai-testtf.openai.azure.com/openai/deployments/gpt-4o/chat/completions` also causes a 404.

The OpenAI client library used in cagent (if I understand the code correctly) seems to also support Azure OpenAI, but with different methods, see e.g. https://github.com/sashabaranov/go-openai/blob/8e5611cc5efdc2533b80e5667b69741c3fad875c/README.md?plain=1#L617-L618

krissetto:
Proposal for supporting AzureOpenAI using the standard OpenAI provider

closes #112


pkg/model/provider/openai/client.go
```diff
@@ -102,6 +140,14 @@ func (c *Client) newGatewayClient(ctx context.Context) *openai.Client {
 	return openai.NewClientWithConfig(cfg)
 }
 
+// isAzureBaseURL returns true if the base URL points to Azure OpenAI
+func isAzureBaseURL(u string) bool {
```
krissetto: Alternatively to this, we could maybe use another `provider_opts` key such as `azure: true` 

```yaml
models:
  azure:
    provider: openai
    ...
    provider_opts:
        azure: true
        ...
```

tfenster: I could imagine that at some point Azure OpenAI would support custom domains. Then it would be useful to have the option key. Or maybe react on the azure_api_version? If that is set, it has to be Azure OpenAI and vice versa

tfenster: Testing a bit more, I now indeed got an endpoint that ends with `.cognitiveservices.azure.com`. That doesn't work with the currently proposed implementatoin

krissetto: Good idea, I changed the code to react on the existence of any `azure_` prefixed provider option.

If you give it a try by building off of this branch let me know if it all works as expected for you (the lower latency AzureOpenAI tends to have really is quite nice 😄)

tfenster: Works perfectly fine for me on Linux and Windows 👍😊 Thanks for the extremely fast turnaround!

stanislavHamara:
Implemented by https://github.com/docker/cagent/pull/153

PR #126, opened 2025-09-09
Support for OpenRouter via a new provider

xhanin:
Summary
- Add OpenRouter provider and integrate with agent/CLI.

Details
- Add OpenRouter provider:
  - pkg/model/provider/openrouter/adapter.go
  - pkg/model/provider/openrouter/client.go
  - pkg/model/provider/openrouter/client_test.go
- Register OpenRouter in provider registry (pkg/model/provider/provider.go)
- Update agent creation (internal/creator/agent.go)
- Update CLI `new` command (cmd/root/new.go)
- Update documentation (README.md)


xhanin:
I see that the approach for Azure was to adjust the openai client, and not a new provider. The approach here introduces too much duplication, it "works for me", but this is not enouch to share it, sorry for the noise.

PR #179, opened 2025-09-12
dockerfile: use cross-comp target to avoid emulation

crazy-max:
relates to https://github.com/docker/cagent/actions/runs/17650827920/job/50161382814

`build-agent` does not use cross compilation so build is done through emulation when building in GitHub Actions for `linux/arm64` architecture. Changed to use `builder` one instead.

Also update workflow to run `build` job on pull request without push to make sure there is no regression with this target.

Before it took ~19m to build:

image

Now ~2m: https://github.com/docker/cagent/actions/runs/17671685388/job/50224866246?pr=179

image

Will open follow-ups to consolidate things in the Dockerfile.

cc @jeanlaurent

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,8 +98,8 @@ jobs:
           TELEMETRY_HEADER: ${{ secrets.TELEMETRY_HEADER }}
 
   build-image:
-    if: startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
-    needs: [build]
+    needs:
+      - build
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -124,8 +124,9 @@ jobs:
         with:
           images: docker/cagent
           tags: |
-            type=raw,value=latest,enable={{is_default_branch}}
-            type=ref,event=tag
+            type=semver,pattern={{version}}
+            type=edge
+            type=ref,event=pr
```
crazy-max, line 129: Versioning strategy is slightly different but should be aligned with what we should expect when releasing a product.

* On release semver git tag like `v1.2.0`, will set `latest` and `v1.2.0` tags
* On pre-release semver git tag like `v1.2.0-rc1`, will set `v1.2.0-rc1` tag
* Merged on default branch, will set `edge` tag

PR #183, opened 2025-09-12
enhanced dockerfile

crazy-max:
See individual commits

As follow-up will add a bake definition.


jeanlaurent:
Hey @crazy-max let's try to close PR as fast as we can, can you split it between enhancements and zig support ?

crazy-max:
> Hey @crazy-max let's try to close PR as fast as we can, can you split it between enhancements and zig support ?

@jeanlaurent We discussed offline with @rumpl about it and we could get rid of CGO if we switch to https://pkg.go.dev/modernc.org/sqlite?utm_source=godoc. Then zig or osxcross would not be needed.

PR #193, opened 2025-09-15
Build image only on manual dispatch or tag

jeanlaurent:


.github/workflows/ci.yml
```diff
@@ -103,6 +103,7 @@ jobs:
           TELEMETRY_HEADER: ${{ secrets.TELEMETRY_HEADER }}
 
   build-image:
+    if: startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
```
crazy-max, line 106: Why? It was useful to run it on PR so we make sure there is no regression, this job only takes ~1m to run.

jeanlaurent: cc @rumpl

rumpl: Just talked with crazymax, the `build-image` was done at the very end, after the `build` step. We'll bring the build-image back but it will run in parallel with the build. Also crazymax has that other PR that will make the build slightly faster

PR #198, opened 2025-09-15
Handling Oauth flows for MCP servers that follow MCP Oauth

trungutt:
Handling Oauth flows for MCP servers that follow MCP Oauth


pkg/server/server.go
```diff
@@ -122,6 +122,10 @@ func New(sessionStore session.Store, runConfig config.RuntimeConfig, teams map[s
 
 	group.GET("/desktop/token", s.getDesktopToken)
 
+	// Resume to start an OAuth flow
+	group.POST("/:id/resumeStartOauth", s.resumeStartOauth)
+	group.POST("/resumeCodeReceivedOauth", s.resumeCodeReceivedOauth)
```
rumpl, line 127: These should be 

```suggestion
	group.POST("/sessions/:id/resumeStartOauth", s.resumeStartOauth)
	group.POST("/sessions/:id/resumeCodeReceivedOauth", s.resumeCodeReceivedOauth)
```

Why doesn't the second one have a session id?

dgageot (dismissed):

pkg/tools/mcp/stdio.go
```diff
@@ -322,3 +322,66 @@ func (c *stdioMCPClient) ReadResource(ctx context.Context, request mcp.ReadResou
 
 	return mcp.ParseReadResourceResult(response)
 }
+
+// Ping checks if the server is alive
```
rumpl: What are these? Are they related in any way with Oauth?

pkg/agent/agent.go
```diff
@@ -88,9 +104,28 @@ func (a *Agent) Model() provider.Provider {
...
+				if serverInfoProvider, ok := toolSet.(ServerInfoProvider); ok {
+					serverURL, serverType := serverInfoProvider.GetServerInfo()
+					return nil, &OAuthAuthorizationRequiredError{
+						Err:        err,
+						ServerURL:  serverURL,
+						ServerType: serverType,
+					}
+				}
+			}
+		}
```
rumpl: Shouldn't this go into the runtime rather than here?

pkg/tools/mcp/stdio.go
trungutt: I revert the first commit, I'll think about that later on

pkg/agent/agent.go
trungutt: Done

pkg/server/server.go
trungutt: We don't need the id from the second one. Let me explain:

- When building Oauth url to open in the browser, we build a `state` param (traditionally used for security purpose), but we also encode the session id in it
- When Oauth makes deeplink back, it gives a `code` and the exact `state`. The external client to cagent just needs to send back that `state` to the endpoint which is in charge of decode it, and then get the right session runtime. This way the encode/decode functions are in cagent side

rumpl: Gotcha!

PR #206, opened 2025-09-16
Fix summarization provider config

### LINKED ISSUE #138: [DMR provider] creating chat completion: error, status code: 500
**Describe the bug**

After the end of the tool completion I got this error:
```
creating chat completion: error, status code: 500, status: 500 Internal Server Error, message: Assistant message must contain either 'content' or 'tool_calls'!
```
> - tested on arm darwin and on arm linux
> - tested with 2 models: `ai/qwen2.5:latest` and `hf.co/menlo/jan-nano-gguf:q4_k_m`

**Version affected**

Release `1.0.6`, but `cagent version` displays:
```
cagent version dev
Build time: unknown
Commit: unknown
```

**How To Reproduce**

https://youtu.be/vmZv6AN1TZs

**OS and Terminal type**

- Darwin tests: Sequoia 15.6 - iTerm2
- Linux test: from a container using the terminal of VSCode

**Config yaml file**

```yaml
version: 1
agents:
  root:
    model: jan-nano
    description: Bob
    instruction: |
      You are a helpful AI assistant named "Bob".
      You are an expert in programming, software development, and technology.
      You are also skilled in using tools to assist with tasks.
      You always answer as helpfully as possible, while being safe.
      When you don't know the answer to a question, you admit that you don't know.
      You never make up answers or provide false information.
      You always format your answers in markdown, including any code snippets.
      You always use bullet points or numbered lists when providing multiple items.
      You always ask clarifying questions if a question is ambiguous.
    toolsets:
      - type: mcp
        remote:
          url: http://localhost:9011/mcp
          transport_type: streamable
   
models:
  jan-nano:
    provider: dmr
    model: hf.co/menlo/jan-nano-gguf:q4_k_m
    temperature: 0.7
    top_p: 0.9
    top_k: 20
    min_p: 0.0
```

krissetto:
This change configures the summarization model to have a higher max token count then the original model configuration.

If a model hits the configured max_tokens limit while streaming its response to the user, when we then try to summarize the session right after we might hit the exact same limit and get no summary (especially with non-thinking models, since when they return we'll have a thread that's already at the max_tokens limit).

The issue is less noticeable with thinking models (e.g. gpt-5 family) because the API doesn't return to us all the thinking tokens that the model has consumed, so subsequent responses will have some wiggle room in the configured max_tokens

Closes #138

### DIFF
--- a/pkg/model/provider/anthropic/client.go
+++ b/pkg/model/provider/anthropic/client.go
@@ -189,6 +189,15 @@ func (c *Client) CreateChatCompletion(
 	return response.Content[0].Text, nil
 }
 
+// Config returns a copy of the model configuration
+func (c *Client) Config() *latest.ModelConfig {
+	if c == nil || c.config == nil {
+		return nil
+	}
+	configCopy := *c.config
+	return &configCopy
+}
+
 func convertMessages(messages []chat.Message) []anthropic.MessageParam {
 	var anthropicMessages []anthropic.MessageParam
 
--- a/pkg/model/provider/dmr/client.go
+++ b/pkg/model/provider/dmr/client.go
@@ -391,6 +391,15 @@ func (c *Client) CreateChatCompletion(
 	return response.Choices[0].Message.Content, nil
 }
 
+// Config returns a copy of the model configuration
+func (c *Client) Config() *latest.ModelConfig {
+	if c == nil || c.config == nil {
+		return nil
+	}
+	configCopy := *c.config
+	return &configCopy
+}
+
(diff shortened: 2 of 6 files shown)


pkg/model/provider/anthropic/client.go
```diff
@@ -189,6 +189,15 @@ func (c *Client) CreateChatCompletion(
 	return response.Content[0].Text, nil
 }
 
+// Config returns a copy of the model configuration
+func (c *Client) Config() *latest.ModelConfig {
```
krissetto, line 193: I'll consolidate these into some base struct in a separate change

PR #221, opened 2025-09-16
Automatic Retry for Chat Completion Failures

a-dubs:
## 🔄 **Feature: Automatic Retry for Chat Completion Failures**

Adds robust retry functionality to automatically recover from transient chat completion and gateway errors, improving reliability for long-running agent sessions.

### **What's New**
- **Automatic Recovery**: Retry failed chat completions up to 3 times
- **Smart Backoff**: Exponential delay between retry attempts (1s, 2s, 3s)
- **Comprehensive Coverage**: Retries both stream creation and processing errors
- **Detailed Logging**: Structured logging with retry count tracking
- **Per-Request Reset**: Retry counter resets on successful operations

### **Usage**
```bash
# Enable retry for improved reliability
cagent run config.yaml --retry

# Perfect for automation and CI/CD
cagent exec config.yaml --retry --yolo

# Combine with other resilience features
cagent run config.yaml --retry --debug
```

### **Error Scenarios Covered**
- **Network Issues**: Temporary connectivity problems
- **Rate Limiting**: Provider API rate limit responses  
- **Gateway Errors**: Intermittent gateway failures
- **Model Loading**: Model initialization delays

### **Technical Implementation**
- **Retry Logic**: Integrated into runtime event loop
- **Exponential Backoff**: `time.Sleep(time.Duration(retryCount) * time.Second)`
- **Context Awareness**: Respects cancellation and timeouts
- **Telemetry Integration**: Error tracking for monitoring
- **Configuration**: Runtime option with CLI flag support

### **Logging Output**
```
INFO  Retrying chat completion stream creation agent=coder retry_count=1 max_retries=3
INFO  Retrying stream handling agent=coder retry_count=2 max_retries=3
ERROR Failed to create chat completion stream agent=coder retry_count=3 (final attempt)
```

### **Files Changed**
- `pkg/runtime/runtime.go`: Core retry logic and backoff implementation
- `cmd/root/{exec,run}.go`: CLI flag integration
- `pkg/config/runtime.go`: Configuration structure updates

---

## How to test changes:

create an agent config like:

```yaml
# test-retry-failure.yaml
version: "2"

agents:
  root:
    model: openai/gpt-4o-mini
    description: Test agent for retry functionality with failure simulation
    instruction: |
      You are a test agent. Use the empty_output tool to trigger a failure, then respond normally.
    toolsets:
      - type: script
        shell:
          empty_output:
            cmd: "echo -n"
            description: A tool that returns empty output to simulate failures
            args:
              message:
                type: string
                description: A message parameter (ignored)
            required: []
```

And then run the following command with and without the `--retry` flag and you will see it work.
```bash
cagent run test-retry-failure.yaml --retry --yolo --tui=false "Please use the empty_output tool with message 'test' and then tell me what happened"
```

### DIFF
--- a/cmd/root/new.go
+++ b/cmd/root/new.go
@@ -121,6 +121,12 @@ func NewNewCmd() *cobra.Command {
 						llmIsTyping = false
 					}
 					printToolCallResponse(e.ToolCall, e.Response)
+				case *runtime.RetryAttemptEvent:
+					if llmIsTyping {
+						fmt.Println()
+						llmIsTyping = false
+					}
+					printRetryAttempt(e)
 				case *runtime.ErrorEvent:
 					if llmIsTyping {
 						fmt.Println()
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -454,6 +454,12 @@ func runWithoutTUI(ctx context.Context, agentFilename string, rt runtime.Runtime
 				if e.ToolCall.ID == lastConfirmedToolCallID {
 					lastConfirmedToolCallID = ""
 				}
+			case *runtime.RetryAttemptEvent:
+				if llmIsTyping {
+					fmt.Println()
+					llmIsTyping = false
+				}
+				printRetryAttempt(e)
 			case *runtime.ErrorEvent:
 				if llmIsTyping {
 					fmt.Println()
--- a/cmd/root/run_text_utils.go
+++ b/cmd/root/run_text_utils.go
@@ -10,6 +10,7 @@ import (
 	"github.com/fatih/color"
 	"golang.org/x/term"
 
+	"github.com/docker/cagent/pkg/runtime"
 	"github.com/docker/cagent/pkg/tools"
 )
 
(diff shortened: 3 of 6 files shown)


krissetto:
Nice. 😎 

I'm wondering how we should display retries to the user in the TUI because i think retries should be shown and not hidden behind the scenes, but WDYT?

a-dubs:
@krissetto that is a good call. let me try this out. definitely don't want to obfuscate that from the user. Also what do you think about making retries a default feature? For this, I kept it feature flagged just to demonstrate the difference but personally I would never _not_ want to retry. It's super frustrating doing a one-shot cagent run and then having it fail towards the end when its so close to completing its task just due to a random timeout or gateway issue and would love to not have to append --retry to every invocation of cagent. But i guess i could just bake it into the shebang at the top of my configs if i really want.

dgageot:
I'm going to close this Pull Request for now. Let's revisit that in the future

PR #222, opened 2025-09-16
don't use /internal/ from /pkg/

dgageot:


rumpl:
But we don’t have internal any more

dgageot:
@rumpl I want to move back a few things that shouldn't be in `pkg`. This is to make sure we do it right

rumpl (approved):

Issue #227, opened 2025-09-17
Ability to define max iterations on an agent


krissetto:
**What you'd like to see**

Something along the lines of a `--max-iterations` parameter for `cagent new`, and a `max_iterations` field on an agent's yaml definition

Examples: 

`cagent new --model dmr/ai/gemma3n:2B-F16 --max-iterations 50`

---

```
...
agents:
  model: dmr/ai/qwen3:14B-Q6_K
  max_iterations: 10
  description: ...
  instruction: ...
  ...
```

**Why you'd like to see it**

Some models (especially thinking about smaller local models) can sometimes get confused during the agent loop and iterate forever either repeating themselves or performing random actions that lead nowhere.

PR #228, opened 2025-09-17
max iteration support for cagent new

krissetto:
Allow defining `--max-iterations` parameter on the `cagent new` command.

The rest of the cagent commands (run, api, ..) maintain the current behavior by calling `rt.RunStream` with `maxIterations` set to 0 (unlimited)

When using a model via dmr, the default max_iterations is set to 20.  
For all other providers, there is no set default (so iterations are unlimited).

When the max iteration count is reached, the user will be prompted if they want to continue for 10 more iterations or stop the execution of the agent.

**Screenshots**

Examples using `--max-iterations 2`:

Screenshot 2025-09-17 at 13 02 14

---

Screenshot 2025-09-17 at 13 01 45

---

references #227

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -40,7 +40,7 @@ type Runtime interface {
 	// CurrentAgent returns the currently active agent
 	CurrentAgent() *agent.Agent
 	// RunStream starts the agent's interaction loop and returns a channel of events
-	RunStream(ctx context.Context, sess *session.Session) <-chan Event
+	RunStream(ctx context.Context, sess *session.Session, maxIterations int) <-chan Event
 	// Run starts the agent's interaction loop and returns the final messages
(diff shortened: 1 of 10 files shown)


pkg/runtime/runtime.go
```diff
@@ -168,8 +168,9 @@ func (r *runtime) finalizeEventChannel(ctx context.Context, sess *session.Sessio
 	}
 }
 
-// Run starts the agent's interaction loop
-func (r *runtime) RunStream(ctx context.Context, sess *session.Session) <-chan Event {
+// RunStream starts the agent's interaction loop with an optional iteration limit
+// If maxIterations is 0, there is no limit
+func (r *runtime) RunStream(ctx context.Context, sess *session.Session, maxIterations int) <-chan Event {
```
rumpl: `maxIterations` should be a propery of the session

PR #229, opened 2025-09-17
Fix: Handle empty tool messages

mqasimsarfraz:
Include a default output if tool call returns an empty response!

Fixes: https://github.com/docker/cagent/issues/220

### Before

image

### After

Screenshot from 2025-09-17 16-23-00

### DIFF
--- a/pkg/tools/mcp/client.go
+++ b/pkg/tools/mcp/client.go
@@ -204,6 +204,11 @@ func processMCPContent(toolResult *mcp.CallToolResult) *tools.ToolCallResult {
 		}
 	}
 
+	// Handle an empty response. This can happen if the MCP tool does not return any content.
+	if finalContent == "" {
+		finalContent = "no output"
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/model/provider/openai/client.go
```diff
@@ -148,6 +148,11 @@ func convertMessages(messages []chat.Message) []openai.ChatCompletionMessage {
 			continue
 		}
 
+		// Handle an empty tool message. This can happen if the tool produced no output.
```
rumpl: This could be done once for all the providers if we handle it in the mcp client: https://github.com/docker/cagent/blob/main/pkg/tools/mcp/client.go#L199-L210

mqasimsarfraz: Thanks, that makes more sense. Tested it and works fine! 

image

rumpl (approved):
<3

rumpl:
Thanks @mqasimsarfraz! 🎉

PR #265, opened 2025-09-22
feat/examples: Add K8s debugger

mqasimsarfraz:
This PR adds an example of using K8s debugger with cagent. 

## Testing done:

[Created a pod](https://gist.githubusercontent.com/mqasimsarfraz/d72127ab5a7e92f74fdf1c1c49be2bd7/raw/236c03952268aff363d867b6c0823f4509b30d5f/gistfile1.txt) and performed the following steps:

Screenshot from 2025-09-22 12-08-55
Screenshot from 2025-09-22 12-09-20
Screenshot from 2025-09-22 12-09-42

**Note**: I wanted to use `mcp/kubernetes` directly but it seems it needs some fixes in https://github.com/docker/mcp-registry/blob/main/servers/kubernetes/server.yaml for handle kubeconfig so I using `docker run ...` but using the image for MCP hub.

### DIFF
--- a/examples/k8s_debugger.yaml
+++ b/examples/k8s_debugger.yaml
@@ -1,0 +1,44 @@
+#!/usr/bin/env cagent run
+version: "2"
+
+agents:
+  root:
+    model: openai/gpt-5
+    description: An AI Assistant for Debugging Kubernetes Clusters
+    instruction: |
+      You are an AI assistant for debugging Kubernetes clusters. Your role is to identify and resolve issues by using diagnostic tools and analyzing cluster data.
+
+      Instructions:
+        1. Ask if the user has observed specific issues or symptoms in the cluster.
+        2. Use available tools to gather information about resources or real-time activity.
+        3. Analyze the collected data to detect potential problems.
+        4. Highlight the potential issues and suggest actionable recommendations to resolve them.
+      
+      Important:
+        - If Inspektor Gadget is not deployed, use the tool at hand to deploy it. Ask user to restart the agent session after deployment.
+        - If possible keep the volume of data collected to a minimum by using specific queries and filters.
+        - Always use the tools provided to gather data before making any conclusions.
+
+    toolsets:
+      - type: mcp
+        ref: docker:inspektor-gadget
+        ## See full if of tools: https://hub.docker.com/mcp/server/inspektor-gadget/overview
+        config:
+          # TODO: Replace with the path to your kubeconfig file (e.g /home/user/.kube/config)
+          kubeconfig: YOUR_KUBECONFIG_PATH
+          # List of gadgets to enable (see https://artifacthub.io/packages/search?kind=22&verified_publisher=true&official=true&cncf=true&sort=relevance&page=1 for a full list)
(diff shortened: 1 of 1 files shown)


slimslenderslacks:
Let's fix that server.yaml entry for kubernetes. It does not have a way to mount the user's ~/.kube/config

slimslenderslacks:
I've added this issue: https://github.com/docker/mcp-registry/issues/275

mqasimsarfraz:
> I've added this issue: [docker/mcp-registry#275](https://github.com/docker/mcp-registry/issues/275)

Perfect. Let me propose a fix for it!

mqasimsarfraz:
I proposed the fix: https://github.com/docker/mcp-registry/pull/276

examples/k8s_debugger.yaml
```diff
@@ -0,0 +1,106 @@
...
+    
+            1. DNS Activity:
+            • The pod made multiple DNS queries to resolve the domain  example.com .
+            • Initially, the DNS request seemed to query an external nameserver at  1.2.3.4 , which did not return any answers.
+            • Subsequently, queries were made to the internal Kubernetes  kube-dns  service ( 10.0.0.10 ) for  example.com .
+            2. Other Networking Details:
+              • The DNS query latencies were low, indicating no major delays in network responses.
+          </ASSISTANT>
+        </CHAT>
+      </CHATS>
```
krissetto: i think we shouldn't have this whole chat example in the prompt of the agent

examples/k8s_debugger.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+models:
+  azure_model:
+    provider: azure
+    ## TODO: Replace with your Azure OpenAI deployment name
+    model: gpt-4o
+    # TODO: Replace with your Azure OpenAI base URL
+    base_url: https://YOUR-PROJECT-RESOURCE.cognitiveservices.azure.com
+    provider_opts:
+      api_version: 2025-01-01-preview
```
krissetto: for the sake of simplicity i think we should have this example use a provider with less need for configuration, like `openai`/`anthropic`

examples/k8s_debugger.yaml
mqasimsarfraz: Done. I agree it makes things too specific!

examples/k8s_debugger.yaml
mqasimsarfraz: Done. Probably we can have a separate example for a azure provider? I feel like it will be helpful since it requires more work! :)

mqasimsarfraz:
> Note: I wanted to use mcp/kubernetes directly but it seems it needs some fixes in https://github.com/docker/mcp-registry/blob/main/servers/kubernetes/server.yaml for handle kubeconfig so I using docker run ... but using the image for MCP hub.

Now that https://github.com/docker/mcp-registry/pull/278 is merged. I updated the PR to use `mcp/kubernetes` directly!

stasyu2009-ux (approved):
Everything is correct but the code version is incorrect so it causes an error, fix it then everything will be correct, thank you for the good work

PR #275, opened 2025-09-23
Don't use colours in non-tui mode

dgageot:


cmd/root/new.go
```diff
@@ -54,19 +54,19 @@ func NewNewCmd() *cobra.Command {
 					switch {
 					case os.Getenv("ANTHROPIC_API_KEY") != "":
 						modelProvider = "anthropic"
-						fmt.Printf("%s\n\n", gray("ANTHROPIC_API_KEY found, using Anthropic"))
+						fmt.Printf("%s\n\n", white("ANTHROPIC_API_KEY found, using Anthropic"))
```
krissetto, line 57: haven't tested this yet, but wouldn't this end up being white on white (completely illegible) for all those not using a dark mode terminal?

dgageot: yeah, that's true. Honestly, outside of the TUI, we should refrain from using colours. Right now it's grey on grey for me.

PR #297, opened 2025-09-25
Compute messages once and start handling errors

dgageot:


pkg/runtime/runtime.go
```diff
@@ -160,15 +160,20 @@ func (r *runtime) RunStream(ctx context.Context, sess *session.Session) <-chan E
...
 		}
 
+		a := r.team.Agent(r.currentAgent)
+		model := a.Model()
+		modelID := model.ID()
+
+		messages, err := sess.GetMessages(a)
+		if err != nil {
+			events <- Error(fmt.Sprintf("getting messages: %v", err))
+			return
```
rumpl, line 170: Not sending back the user message will break the TUI in this case, the TUI only shows the user message when it receives the UserMessage event

dgageot: Doesn't seem to break for me. It sure shows an error message but that's what I wanted.
Not sure what else it should do

rumpl: Show the user message too?

dgageot:
Closing in favour of https://github.com/docker/cagent/pull/304

PR #337, opened 2025-09-27
Misc simplifications

dgageot:


pkg/tui/components/tool/tool.go
```diff
@@ -18,13 +18,10 @@ import (
 // toolModel implements Model
 type toolModel struct {
 	message *types.Message
-
```
rumpl, line 21: Kinda liked these

rumpl (changes requested):
I would like to keep the history. Am counting on adding it to the tui

pkg/tui/components/tool/tool.go
dgageot: ...

rumpl: !!!

PR #341, opened 2025-09-29
Add session and agent info in the stream start/stop events

rumpl:


krissetto:
I think it would make sense to emit specific "agent delegation events", which would make things easier to use than having to parse all the tool events in search of the special one. WDYT?

rumpl:
> I think it would make sense to emit specific "agent delegation events", which would make things easier to use than having to parse all the tool events in search of the special one. WDYT?

Sounds overkill, a general thing could be that the sub-session start event would send the sessionId and also the parent session ID, but none of that is needed so we can come back to this once there's a need

PR #342, opened 2025-09-29
Implement Code Mode for calling tools

dgageot:
See https://blog.cloudflare.com/code-mode/

Example:
```console
cagent exec --yolo --code-mode ./examples/gopher.yaml "Read all my _test.go files and count the total number of lines"
```


pkg/codemode/exec.go
```diff
@@ -0,0 +1,69 @@
...
+	v, err := vm.RunString(script)
+	if err != nil {
+		return fmt.Sprintf("Error running script: %s", err), nil
+	}
+
+	// Some script are fire and forget and don't return anything.
+	// In that case we return "done." to please the LLM which can't deal with empty responses.
+	result := v.Export()
+	if result == nil {
+		return "done.", nil
```
krissetto: What if a tool outputs nothing erroneously (or because misused by the llm)? 

I think returning `done.`  _could_ confuse the llm in those scenarios, so IMHO we should use something like `<no output>` instead (which could help local/less performant models better understand whats going on)

WDYT?

cmd/root/eval.go
```diff
@@ -19,6 +19,7 @@ func NewEvalCmd() *cobra.Command {
 	}
 
 	cmd.PersistentFlags().StringSliceVar(&runConfig.EnvFiles, "env-from-file", nil, "Set environment variables from file")
+	cmd.PersistentFlags().BoolVar(&runConfig.GlobalCodeMode, "code-mode", false, "Provide a single tool to call other tools via Javascript")
```
krissetto: nit: maybe `--code-mode-tools` so its more explicit? `--code-mode` could be useful for others things as well

PR #343, opened 2025-09-29
Add the tool definition to the partial tool call and tool call events

### LINKED ISSUE #296: Issues with TUI slowness after tool calls
After some tools calls, the TUI becomes sluggish, even the loader in the sidebar spins at ~1 frame per 10 seconds...

rumpl:
Adds the tool definition to the two PartialToolCall and ToolCall events

This can help clients get the display name for example.

Fixes #296


rumpl:
I added some TODOS, parts of the code are not great but fixing them needs more cleanup work in the runtime

krissetto (approved):

PR #355, opened 2025-09-29
feat: Implement stack-based multiple dialogs management in TUI

rumpl:
This PR implements a stack-based dialog management system that supports multiple overlapping dialogs in the TUI.


rumpl:
> Noice. The tests mentioned in the pr description aren't there though, is that intentional?

Fixed

PR #365, opened 2025-09-30
Also lint tests

dgageot:


pkg/memory/database/sqlite/sqlite_test.go
```diff
@@ -176,13 +169,13 @@ func TestDatabaseOperationsWithCanceledContext(t *testing.T) {
 	}
 
 	err := db.AddMemory(ctx, memory)
-	assert.Error(t, err, "AddMemory should fail with canceled context")
+	require.Error(t, err, "AddMemory should fail with canceled context")
```
rumpl, line 172: assert is fine here no? I _do_ want things to continue, that way we can see all the failures at once

rumpl (approved):

PR #377, opened 2025-10-01
Add support for commands in the CLI/TUI/API

### LINKED ISSUE #128: Create agent dedicated commands
You can assign a given prompt to a set of agents.

Example:

With my calendar-management.yaml I can create a command that is named `15min` which will look for 15 min slot in today

I could run this with. `cagent run calendar-management.yaml 15min` and the agent will start right away with the commandm without any initial prompting.

krissetto:
Proposal for adding support for commands in the CLI/TUI/API

Example yaml file

```yaml
version: "2"

agents:
  root:
    model: openai/gpt-5-nano
    instruction: you are a helpful computer assistant
    commands:
      df: "check how much free space i have on my disk"
      ls: "list the files in the current directory"
    toolsets:
      - type: filesystem
      - type: shell

  another:
    model: openai/gpt-5-nano
    instruction: you are a helpful computer assistant
    commands:
      - fun: "tell me a joke about computers"
      - info: "tell me some information about computers"
```

Example CLI usage

```bash
# run a command
$ cagent run test.yaml -c df
... opens tui and sends the chosen command as the user message

$ cagent run test.yaml --command ls
... opens tui and sends the chosen command as the user message


# list commands
$ cagent run test.yaml -c

For any feedback, please visit: https://docker.qualtrics.com/jfe/form/SV_cNsCIg92nQemlfw

Available commands for agent 'root':
 - df: check how much free space i have on my disk
 - ls: list the files in the current directory
```

closes #128

### DIFF
--- a/cagent-schema.json
+++ b/cagent-schema.json
@@ -8,8 +8,8 @@
     "version": {
       "type": "string",
       "description": "Configuration version",
-      "enum": ["2"],
-      "examples": ["2"]
+      "enum": ["1", "2", "v1", "v2"],
+      "examples": ["1", "2", "v1", "v2"]
     },
     "agents": {
       "type": "object",
@@ -93,6 +93,22 @@
           "items": {
             "type": "string"
           }
+        },
+        "commands": {
+          "description": "Named prompts for quick-start commands used with --command/-c",
+          "oneOf": [
+            {
+              "type": "object",
+              "additionalProperties": { "type": "string" }
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "object",
+                "additionalProperties": { "type": "string" }
+              }
+            }
+          ]
         }
       },
       "additionalProperties": false
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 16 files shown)


rumpl:
I _think_ I would prefer if we could run a command with `cagent run agent.yaml df` instead of `-c df`

And maybe have `cagent commands agent.yaml` ?

Things got merged and this has conflicts now, sorry

krissetto:
conflicts resolved. i'm fine without the -c, but do we still want to support messages? right now `cagent run agent.yaml hello` works and that conflicts with what you're saying.

i can add the `commands` command if we want that

krissetto:
-c without a command currently does the listing like i assume you want `cagent commands agent.yaml` to do, just in case a user misuses the flag

rumpl:
`cagent run agent.yaml --command` that lists commands definitely looks weird to me.

What about:

```console
$ cagent command ls agent.yaml
```

```console
$ cagent command run agent.yaml <command>
```

I'm only asking mind you :)

krissetto:
Oh listing with --command/-c wasn't meant as **the** way to list available commands, but more as a ux thing for users who mess up so they don't have to type another command to see what commands are available. 

I'm ok with either approach really, having a dedicated cagent command for commands (lol) feels a bit overkill to me right now for what we're trying to do, which is just have a default first message from the agent file sent automatically, but I can make those changes if that's what we agree upon. Would love to hear from the others as well but if I don't then I'll just make these changes and move on with other things :)

krissetto:
Unless there are hard feelings around this, I'd honestly prefer if we merge this as is for now (meaning without a new top level CLI command) so this doesn't remain stuck here, and then iterate if we need to.

After using it a bit we can decide if we feel the need to make a top level `cagent command` command, especially if we feel the need for more features than currently implemented. To me it feels like a variation of the `run` command that's too small to warrant it's own top level `command` from the get-go

rumpl (approved):

rumpl:
Let's go

PR #423, opened 2025-10-06
Start to use recent fs APIs to restrict which directory we have access to

dgageot:


pkg/config/config.go
```diff
@@ -197,3 +143,58 @@ func validateConfig(cfg *latest.Config) error {
 func boolPtr(b bool) *bool {
 	return &b
 }
+
+func ValidatePathInDirectory(path, allowedDir string) (string, error) {
```
rumpl, line 147: This one doesn't look used any more? Or is it used somewhere else?

dgageot: Yeah, it's used in many other places in the codebase...

.vscode/mcp.json
krissetto, line 1: not needed? we could add .vscode to the gitignore

pkg/teamloader/teamloader.go
```diff
@@ -118,7 +118,7 @@ func Load(ctx context.Context, path string, runtimeConfig config.RuntimeConfig)
 	)
 
 	// Load the agent's configuration
-	cfg, err := config.LoadConfigSecure(fileName, parentDir)
+	cfg, err := config.LoadConfigSecureDeprecated(fileName, parentDir)
```
krissetto, line 121: are these uses of LoadConfigSecureDeprecated also meant to be eventually updated?

krissetto (approved):

pkg/teamloader/teamloader.go
dgageot: yes

.vscode/mcp.json
dgageot: Good catch! I opened another PR

PR #427, opened 2025-10-06
Delete session properly

trungutt:
When deleting session which is running, things need to get cancelled


rumpl:
I think we want the cancel maps to be guarded by a mutex or be a sync.Map

trungutt:
@rumpl Done

PR #434, opened 2025-10-07
Support options that produce errors

dgageot:
The plan is to open a root fs when setting the agentsDir. This can fail.


rumpl:
100% of the time when I create options I say to myself "We don't need errors here", 100% of the time at some point we do... :D

PR #436, opened 2025-10-07
Add API for reading/editing agent YAML config

trungutt:
Add API for reading/editing agent YAML config

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -85,6 +85,10 @@ func New(sessionStore session.Store, runConfig config.RuntimeConfig, teams map[s
 	group.GET("/agents", s.getAgents)
 	// Get an agent by id
 	group.GET("/agents/:id", s.getAgentConfig)
+	// Get an agent's raw YAML configuration by id
+	group.GET("/agents/:id/yaml", s.getAgentConfigYAML)
+	// Edit an agent's raw YAML configuration by id
+	group.PUT("/agents/:id/yaml", s.editAgentConfigYAML)
 	// Edit an agent configuration by id
 	group.PUT("/agents/config", s.editAgentConfig)
 	// Create a new agent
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -163,6 +167,86 @@ func (s *Server) getAgentConfig(c echo.Context) error {
...
+	body, err := io.ReadAll(c.Request().Body)
+	if err != nil {
+		slog.Error("Failed to read request body", "error", err)
+		return c.JSON(http.StatusBadRequest, map[string]string{"error": "failed to read request body"})
+	}
+	yamlContent := string(body)
+
+	// Validate the YAML content by attempting to parse it
+	var tmpConfig latest.Config
+	if err := yaml.UnmarshalWithOptions([]byte(yamlContent), &tmpConfig, yaml.Strict()); err != nil {
```
krissetto: in `config.LoadConfig` we are passing the `yaml.ReferenceDirs(dir))` option when unmarshalling the yaml, which would in theory allow referencing other yaml files in the same directory. Also, we do further validation by parsing the config and potentially migrating old configs. 

Should we be doing most of the same validations here so users don't save broken configs (but technically valid yaml), or was the intention to allow that and break later on `teamloader.Load`?

pkg/server/server.go
```diff
@@ -163,6 +167,86 @@ func (s *Server) getAgentConfig(c echo.Context) error {
...
+	if oldTeam, exists := s.teams[agentKey]; exists {
+		// Stop old team's toolsets before replacing
+		if err := oldTeam.StopToolSets(); err != nil {
+			slog.Error("Failed to stop old team toolsets", "agentKey", agentKey, "error", err)
+		}
+	}
+	s.teams[agentKey] = t
+
+	slog.Info("Agent YAML configuration updated successfully", "path", agentPath)
+	return c.String(http.StatusOK, yamlContent)
```
krissetto, line 249: nit: why return the full yamlContent again?

trungutt: This is common practice right

pkg/server/server.go
trungutt: Done

pkg/server/server.go
```diff
@@ -155,6 +167,88 @@ func (s *Server) getAgentConfig(c echo.Context) error {
...
+	if err := yaml.UnmarshalWithOptions([]byte(yamlContent), &tmpConfig, yaml.Strict(), yaml.ReferenceDirs(agentDir)); err != nil {
+		slog.Error("Invalid YAML content", "error", err)
+		return c.JSON(http.StatusBadRequest, map[string]string{"error": "invalid YAML content: " + err.Error()})
+	}
+
+	// Reload the agent to update the in-memory configuration
+	t, err := teamloader.Load(c.Request().Context(), agentPath, s.runConfig)
+	if err != nil {
+		slog.Error("Failed to reload agent after YAML edit", "path", agentPath, "error", err)
+		return c.JSON(http.StatusInternalServerError, map[string]string{"error": "failed to reload agent configuration"})
```
krissetto: do we want to surface the exact error to the user of the api so they can fix the file without checking the logs?

trungutt: Done

Issue #70, opened 2025-09-01
[FEAT] - Thinking budgets


krissetto:
**What you'd like to see**  
 
`cagent` should support configuring thinking budgets that can be used for thinking models like gemini-2.5-pro, gpt-5, etc.

**Why you'd like to see it**  

It'd help with tailoring the behavior of those models, where sometimes a user might prefer a low amount of thinking is better because of lower latency, and sometimes a user might want the model to think as much as possible to solve a harder problem

**Workarounds?**  

Not really, just picking different models based on their default behavior

**Additional context**  

e.g.
```yaml
models:
  low-thinker:
    provider: openai
    model: gpt-5
    thinking_budget: "low"/"medium"/"high"
    ...
  ...

# or we could even support a token based budget, depending on model support
models:
  average-thinker:
    provider: google
    model: gemini-2.5-pro
    thinking_budget: 500 # tokens!
    ...
  non-thinker:
    provider: google
    model: gemini-2.5-pro
    thinking_budget: 0 # tokens!
  ...
```

PR #440, opened 2025-10-08
OpenAI thinking budget support

krissetto:
Adds support for defining a thinking budget for OpenAI reasoning models

`thinking_budget` can be defined on a model definition, or on an agent definition (which will override the model definition)

Support for other providers to come in separate PRs

**Example usage**:

```yaml
version: "2"

models:
  gpt-5-mini:
    provider: openai
    model: gpt-5-mini
    thinking_budget: minimal

agents:
  root:
    model: gpt-5-mini
    instruction: you are a helpful assistant
    # uses thinking_budget defined on model definition

  bob:
    model: gpt-5-mini
    instruction: you are a helpful assistant that thinks a lot
    thinking_budget: high # overrides thinking_budget from model definition
```

References #70


rumpl:
I don't think I like the thinking_budget being on the agent, especially since thinking budgets are different between providers IIUC

krissetto:
> I don't think I like the thinking_budget being on the agent, especially since thinking budgets are different between providers IIUC

i had the same thought but ya never know :) can remove if we don't like it

krissetto:
For experimenting, it can be nice to have the option to override the value on an agent without duplicating model definitions (say i want to keep other model settings the same, such as token limit, temperature etc), but i do agree and think it's more of a model specific property so no hard feelings at all if we want to remove that part for now

krissetto:
@rumpl Rebased and removed agent overriding for now, we can see if people end up asking for this

PR #442, opened 2025-10-08
Fix cagent build

dgageot:


pkg/oci/Dockerfile.template
```diff
@@ -24,7 +24,9 @@ VOLUME /data
 ENV TELEMETRY_ENABLED=true
 ENV CAGENT_HIDE_FEEDBACK_LINK=0
 ENV CAGENT_HIDE_TELEMETRY_BANNER=0
-CMD ["api", "--listen", "unix:///data/agent.sock", "--session-db", "/data/session.db", "/agent.yaml"]
-RUN cat <<EOF >/agent.yaml
+USER root
+RUN cat <<EOF >/agent.yaml && chmod 777 /agent.yaml
```
krissetto, line 28: chmod before the EOF?

dgageot: why?

Issue #27, opened 2025-07-17
Workflows


rumpl:
People seem to want to have workflows

jeanlaurent:
# Cagent Workflow Patterns

## Overview

This document explains the three core workflow execution patterns in `docker/cagent`:

1. **Sequential Step Execution**
2. **Conditional Branching & Loops**
3. **Parallel Step Execution**

---

## 1. Sequential Step Execution

**Description**: Agents execute one after another in a linear chain. Each agent's output becomes available as input context for the next agent.

**Example**:
```yaml
workflow:
  - type: agent
    name: generator
  - type: agent
    name: translator
  - type: agent
    name: publisher
```

**Flow Diagram**:
```
┌───────────┐
│ generator │
└─────┬─────┘
      │
      ▼
┌───────────┐
│translator │
└─────┬─────┘
      │
      ▼
┌───────────┐
│ publisher │
└───────────┘
```

**Behavior**: 
- `generator` runs first and completes
- `translator` receives generator's output and processes it
- `publisher` receives translator's output and finalizes

---

## 2. Conditional Branching & Loops

**Description**: Workflow can branch based on agent outputs using conditions. When a condition routes back to an earlier step, it creates a loop pattern.

**Example**:
```yaml
workflow:
  - type: agent
    name: generator
  - type: agent
    name: translator
  - type: condition
    name: qa_check
    condition: "{{ $qa.output.is_approved }}"
    true:
      - type: agent
        name: publisher
    false:
      - type: agent
        name: translator
```

**Flow Diagram**:
```
┌───────────┐
│ generator │
└─────┬─────┘
      │
      ▼
┌───────────┐
│translator │
└─────┬─────┘
      │
      ▼
┌───────────┐
│ qa_check  │
└─────┬─────┘
      │
      ├─── true ───► ┌───────────┐
      │              │ publisher │
      │              └───────────┘
      │
      └─── false ──► ┌───────────┐
                     │translator │ (loops back)
                     └───────────┘
```

**Behavior**:
- The `qa` agent evaluates the translation quality
- If `is_approved == true`: workflow proceeds to `publisher`
- If `is_approved == false`: workflow routes back to `translator` (creating a retry loop)

---

## 3. Parallel Step Execution

**Description**: Multiple agents execute concurrently. The workflow waits for all parallel steps to complete before moving to the next sequential step.

**Example**:
```yaml
workflow:
  - type: parallel
    steps:
      - generator
      - generator
  - type: agent
    name: translator
[...]

Issue #430, opened 2025-10-07
Add JSON Schema Support for Structured AI Outputs


oleander:
**What you'd like to see**

**Proposal:** Add JSON Schema support for structured AI outputs. This would allow agents to define output schemas that get validated at the API level for supported models (OpenAI, Grok), ensuring consistent and reliable structured responses from AI models.

Key components:

- New `schemas` section in YAML configuration files with native YAML syntax
- Schema references using `$ref` for composition and reuse
- Model-specific schema configuration in agent definitions
- Support for both inline schemas and external schema files
- Configuration validation that fails when schemas are specified for unsupported models

Example configuration:

```yaml
schemas:
  event:
    type: object
    required: [title]
    properties:
      title: 
        type: string
        description: "Event title"
  response:
    type: object
    required: [events]
    properties:
      events: 
        type: array
        items: 
          $ref: "#/schemas/event"

agents:
  root:
    model: openai
    description: "Event extractor"
    instruction: "Extract events from text"
    schema:
      output: event
```

**Why you'd like to see it**

Structured outputs would help me build more reliable AI agents that can consistently return data in expected formats. Currently, I have to implement post-processing validation and retry logic when AI responses don't match expected structures, which adds complexity and reduces reliability. This feature would eliminate the need for manual validation and retry mechanisms, making agent responses more predictable and reducing development overhead.

The feature would be particularly valuable for:

- Data extraction agents that need consistent JSON output
- API integration scenarios requiring specific response formats
- Multi-agent workflows where downstream agents depend on structured data from upstream agents

**Workarounds?**

Currently using manual JSON parsing and validation in post-processing steps, along with retry logic when responses don't match expected formats. This involves:

- Custom validation functions for each expected output structure
- Retry mechanisms with different prompts when validation fails
- Error handling for malformed JSON responses
- Manual schema enforcement in agent instructions

**Additional context**

- The approach is inspired by `jsonSchema` in [GitHub's gh-models project](https://github.com/github/gh-models/blob/f38cd97dc0417a116879b5a6dedec0c8753cad4b/examples/json_schema_prompt.yml#L4)
- OpenAI ([structured output](https://platform.openai.com/docs/guides/structured-outputs) using `response_format`)
- Grok [\(structured outputs\)](https://docs.x.ai/docs/guides/structured-outputs)

[...]

PR #472, opened 2025-10-10
Explore Sequential and Parallel Workflow execution

jeanlaurent:
Partly implements #27 
with Sequential  and Parallel execution.

If this is the right direction, will implement  loops

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -58,6 +58,7 @@ linters:
       disabled-checks:
         - dupImport
         - hugeParam
+        - rangeValCopy
         - unnamedResult
         - appendAssign
       enabled-tags:
--- a/cagent-schema.json
+++ b/cagent-schema.json
@@ -8,8 +8,18 @@
     "version": {
       "type": "string",
       "description": "Configuration version",
-      "enum": ["1", "2", "v1", "v2"],
-      "examples": ["1", "2", "v1", "v2"]
+      "enum": [
+        "1",
+        "2",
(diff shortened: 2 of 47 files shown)


krissetto:
should we maybe draft this?

btw, i see schema being mentioned in the yamls, is this also partially implementing support for #430?

jeanlaurent:
Yeah I'm removing the schema this does too many things.
What do you mean draft?

krissetto:
>  What do you mean draft?

Draft PR, so we know when it should be reviewed 😉

krissetto:
I think it'd be cool if we supported defining multiple named workflows in the yaml, and also allowed agents to trigger a workflow themselves. It would allow for more complex hybrid scenarios where agents or subagents could trigger workflows at their discretion at any point during a conversation.

Amy thoughts around this?

jeanlaurent:
> think it'd be cool if we supported defining multiple named workflows in the yaml, and also allowed agents to trigger a workflow themselves.
You should add this to #27 @krissetto. I think I'm trying to do a step by step here of an imcomplete yet functioning part of the workflow.

jeanlaurent:
@rumpl let me know what we should do with this.
I think the 1st commit would already do something
the second on // execution I'm afraid will trigger a ton of threadsafety issue.

dgageot:
I'm going to close this one for now. Many pieces of cagent will not like things to run concurrently for now.

PR #474, opened 2025-10-10
Implement structured outputs

### LINKED ISSUE #28: Structured output
We should be maybe able to define the structured output we want from the LLM, relates to docker/new-cagent#27

rumpl:
Only for openai and gemini for now, will return an error if the agent is using anthropic

Fixes #28

### DIFF
--- a/cagent-schema.json
+++ b/cagent-schema.json
@@ -8,8 +8,18 @@
     "version": {
       "type": "string",
       "description": "Configuration version",
-      "enum": ["1", "2", "v1", "v2"],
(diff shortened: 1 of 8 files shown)


rumpl:
Just realised it’s missing the types for v0 and v1, will fix

krissetto:
Noice! I think we should add the same to DMR provider since it uses the same chat_completions api, even if it may not work with every single local model out there

rumpl:
Demo command added 👍

PR #475, opened 2025-10-10
Remove the "CreateChatCompletion" function

rumpl:
It's not really used, we only use the streaming one


krissetto:
i wonder if non streaming outputs could be valuable in workflow/structured output scenarios that don't involve a chat UI..

ok with removing it for now though

PR #480, opened 2025-10-13
JSON Schema rework

dgageot:
Make sure we support any JSON schema on any provider.

+ Remove the Go type we used to describe a schema. Use what the MCP servers send us
+ Still make sure that what we send to the providers (especially DMR) contains the required attributes (type and non empty properties)
+ Add more tests
+ Use better schemas, generated from go types, for our builtin tools


examples/golibrary/tool/main.go
```diff
@@ -46,24 +54,14 @@ func main() {
...
-				"a": map[string]any{
-					"type": "number",
-				},
-				"b": map[string]any{
-					"type": "number",
-				},
-			},
-		},
-		Handler: addNumbers,
+		Parameters:  tools.MustSchemaFor[AddNumbersArgs](),
```
rumpl, line 63: How do I define which args are required?

PR #481, opened 2025-10-13
Better support remote MCP servers references by their name in the MCP Catalog

dgageot:
For example:

```yaml
toolsets:
      - type: mcp
        ref: docker:apify
```

### DIFF
--- a/pkg/tools/mcp/remote.go
+++ b/pkg/tools/mcp/remote.go
@@ -92,7 +92,7 @@ func (c *remoteMCPClient) Initialize(ctx context.Context, request *mcp.Initializ
 			Endpoint:   c.url,
 			HTTPClient: httpClient,
 		}
-	case "streamable":
+	case "streamable", "streamable-http":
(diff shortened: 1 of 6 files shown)


pkg/tools/mcp/remote.go
```diff
@@ -92,13 +92,11 @@ func (c *remoteMCPClient) Initialize(ctx context.Context, request *mcp.Initializ
 			Endpoint:   c.url,
 			HTTPClient: httpClient,
 		}
-	case "streamable":
+	default:
```
rumpl: Why though?

dgageot: Everybody has a different name for `streamable`. Sometimes it's `http`, sometimes `stream`, sometimes `streaming` or `streaming_http`. 
The idea is to embrace that fact and recognise everything that's not sse as `streamable`

dgageot: Let me change that and we can discuss some more in the future.

PR #486, opened 2025-10-13
Fix OAuth flow breaking MCP session initialization

trungutt:
Fixed https://github.com/docker/cagent/issues/485.

When remote MCP servers require OAuth authentication, the OAuth flow might be triggered during the MCP session initialization handshake. This caused the session state to become corrupted because:

1. `client.Connect()` starts the MCP protocol handshake
2. Initial HTTP request returns 401 Unauthorized
3. `oauthTransport` intercepts and runs the full OAuth flow
4. After OAuth succeeds, the retry finds the MCP session in a broken state
5. Subsequent requests fail with "400 Bad Request: broken session"

This fix adds a pre-flight authentication check that triggers OAuth before establishing the MCP session, ensuring:

- OAuth completes at the HTTP transport layer first
- MCP session initialization happens with authentication already in place
- No interruption to the MCP protocol state machine

The pre-flight check uses a lightweight HEAD request to the endpoint, allowing the `oauthTransport` to handle any 401 responses and complete the OAuth flow before any MCP protocol messages are exchanged.

### DIFF
--- a/pkg/tools/mcp/remote.go
+++ b/pkg/tools/mcp/remote.go
@@ -84,6 +84,14 @@ func (c *remoteMCPClient) Initialize(ctx context.Context, request *mcp.Initializ
 	// Create HTTP client with OAuth support
 	httpClient := c.createHTTPClient()
 
+	// Pre-flight request to trigger OAuth flow if needed, before establishing MCP session
+	// This prevents OAuth from interrupting the MCP session initialization
+	slog.Debug("Performing pre-flight authentication check", "url", c.url)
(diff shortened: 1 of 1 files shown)


rumpl:
This shouldn’t be needed really, not sure how it fixes the case of two mcp servers. But sure, merge, I’ll have to check tho. 

No preflight or additional http calls should be needed

trungutt:
@rumpl The issue is confusing as I discovered it with two MCP Servers (Notion and Prisma). Turned out that Prisma was the main cause: its way of working is different than other MCP Servers, I cannot think of any other way of an upstream "ping" to make sure that `.Connect` call works

Issue #485, opened 2025-10-13
OAuth flow failure Prisma MCP Server


trungutt:
**Describe the bug**

The OAuth flow fails with Prisma

1. Create an agent with

```
#!/usr/bin/env cagent run
version: '2'

agents:
  root:
    model: anthropic/claude-3-5-sonnet-latest
    description: Prisma Expert
    instruction: You are an AI assistant with expertise in the Prisma project's documentation.
    toolsets:
      - type: mcp
        remote:
          url: https://mcp.prisma.io/mcp
          transport_type: streamable
```

2. Start running this agent with random question. The OAuth flow seems to pass through, but then subsequent requests fail with 400

rumpl:
The prisma MCP server works in two modes: sse and streamable 

https://github.com/prisma/mcp/blob/798c371db42bdb8caa7572d5656c1e3e5b204dd7/server.json#L13-L38

Our current code without the fix from #490 works in SSE mode but gets a 400 in streamable mode.

In streamable mode, once we get the token and retry we get this response from Prisma:

```html
<html>
  <head>
    <title>400 Bad Request</title>
  </head>
  <body>
    <center><h1>400 Bad Request</h1></center>
    <hr />
    <center>cloudflare</center>
  </body>
</html>
```

This is the request: 

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "initialize",
  "params": {
    "capabilities": { "roots": { "listChanged": true }, "elicitation": {} },
    "clientInfo": { "name": "cagent", "version": "1.0.0" },
    "protocolVersion": "2025-06-18"
  }
}

```

PR #490, opened 2025-10-13
Fix OAuth flow breaking MCP session initialization

trungutt:
Fixed https://github.com/docker/cagent/issues/485.

When remote MCP servers require OAuth authentication AND the OAuth flow is triggered during the MCP session initialization handshake. This caused the session state to become corrupted because:

1. client.Connect() starts the MCP protocol handshake
2. Initial HTTP request returns 401 Unauthorized
3. oauthTransport intercepts and runs the full OAuth flow
4. After OAuth succeeds, the retry finds the MCP session in a broken state
5. Subsequent requests fail with "400 Bad Request: broken session"

This fix adds retry logic that detects "broken session" errors during initialization and automatically retries once after OAuth completes, ensuring:

- OAuth completes at the HTTP transport layer first
- MCP session initialization happens with authentication already in place
- No permanent session corruption from the OAuth interruption

The retry is limited to OAuth-related "broken session" errors to avoid masking other legitimate connection failures.

Fixes the issue where OAuth-protected MCP servers (like mcp.prisma.io) would fail to initialize despite successful user authentication.

**Note:** An attempt to fix the same issue was https://github.com/docker/cagent/pull/486, but that fix broken other MCP server like atlassian.

### DIFF
--- a/pkg/tools/mcp/remote.go
+++ b/pkg/tools/mcp/remote.go
@@ -6,7 +6,9 @@ import (
 	"iter"
 	"log/slog"
 	"net/http"
+	"strings"
 	"sync"
+	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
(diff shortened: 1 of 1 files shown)


rumpl:
#485 should be updated. Or we can close it and open a new issue only for prisma

trungutt:
@rumpl https://github.com/docker/cagent/issues/485 has been changed

krissetto (approved):

PR #502, opened 2025-10-15
Add shortcut management for agent configurations

jeanlaurent:
Add alias system for agent configurations

  Adds support for creating aliases for agent paths, making it easier to run
  frequently-used agents.

  Commands

  cagent alias [<name> <agent-path>] - Manage aliases
  - cagent alias - List all aliases
  - cagent alias <name> <path> - Create an alias
  - cagent alias remove <name> - Remove an alias

  cagent run <name> - Automatically resolves aliases
  - Checks for alias first, then falls back to file path or catalog reference
  - No change to existing behavior for files or catalog references

  Usage

  # Create an alias
  $ cagent alias code agentcatalog/notion-expert

  # Run using the alias
  $ cagent run code

  # List aliases
  $ cagent alias

  # Remove an alias
  $ cagent alias remove code

  Implementation

  - Aliases stored in ~/.config/cagent/shortcuts.yaml
  - Resolution order in run command: alias → local file → catalog reference

### DIFF
--- a/cmd/root/alias.go
+++ b/cmd/root/alias.go
@@ -45,7 +45,7 @@ func newAliasAddCmd() *cobra.Command {
 		Use:   "add <alias-name> <agent-path>",
 		Short: "Add a new alias",
 		Args:  cobra.ExactArgs(2),
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(_ *cobra.Command, args []string) error {
 			telemetry.TrackCommand("alias", append([]string{"add"}, args...))
 			return createAlias(args[0], args[1])
 		},
@@ -59,7 +59,7 @@ func newAliasListCmd() *cobra.Command {
 		Aliases: []string{"ls"},
 		Short:   "List all registered aliases",
 		Args:    cobra.NoArgs,
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(*cobra.Command, []string) error {
 			telemetry.TrackCommand("alias", []string{"list"})
 			return listAliases()
 		},
@@ -73,7 +73,7 @@ func newAliasRemoveCmd() *cobra.Command {
 		Aliases: []string{"rm"},
 		Short:   "Remove a registered alias",
 		Args:    cobra.ExactArgs(1),
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(_ *cobra.Command, args []string) error {
 			telemetry.TrackCommand("alias", append([]string{"remove"}, args...))
 			return removeAlias(args[0])
 		},
@@ -89,14 +89,13 @@ func createAlias(name, agentPath string) error {
 	}
 
 	// Expand tilde in path if it's a local file path
-	if strings.HasPrefix(agentPath, "~/") {
-		if abs, err := expandTilde(agentPath); err == nil {
-			agentPath = abs
-		}
+	absAgentPath, err := expandTilde(agentPath)
+	if err != nil {
+		return err
 	}
 
 	// Store the alias
-	s.Set(name, agentPath)
+	s.Set(name, absAgentPath)
 
 	// Save to file
 	if err := s.Save(); err != nil {
@@ -105,7 +104,7 @@ func createAlias(name, agentPath string) error {
 
 	fmt.Printf("Alias '%s' created successfully\n", name)
 	fmt.Printf("  Alias: %s\n", name)
-	fmt.Printf("  Agent: %s\n", agentPath)
+	fmt.Printf("  Agent: %s\n", absAgentPath)
 	fmt.Printf("\nYou can now run: cagent run %s\n", name)
 
 	return nil
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -127,7 +127,7 @@ models:
 
 #### Testing Best Practices
 
-This project uses `github.com/stretchr/testify` for assertions. 
+This project uses `github.com/stretchr/testify` for assertions.
 
 In Go tests, always prefer `require` and `assert` from the `testify` package over manual error handling.
 
--- a/pkg/aliases/aliases.go
+++ b/pkg/aliases/aliases.go
@@ -11,37 +11,30 @@ import (
 )
 
 // Aliases represents the aliases configuration
-type Aliases struct {
-	Aliases map[string]string `yaml:"aliases"`
-}
+type Aliases map[string]string
 
-// GetAliasesFilePath returns the path to the aliases file
-func GetAliasesFilePath() string {
+func aliasesFilePath() string {
 	return filepath.Join(paths.GetConfigDir(), "aliases.yaml")
 }
 
 // Load loads aliases from the configuration file
 func Load() (*Aliases, error) {
-	return LoadFrom(GetAliasesFilePath())
+	return loadFrom(aliasesFilePath())
 }
 
-// LoadFrom loads aliases from a specific file path
-func LoadFrom(path string) (*Aliases, error) {
+// loadFrom loads aliases from a specific file path
+func loadFrom(path string) (*Aliases, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return &Aliases{Aliases: make(map[string]string)}, nil
+			return &Aliases{}, nil
 		}
 		return nil, fmt.Errorf("failed to read aliases file: %w", err)
 	}
 
-	var s Aliases
+	s := Aliases{}
 	if err := yaml.Unmarshal(data, &s); err != nil {
 		return nil, fmt.Errorf("failed to parse aliases file: %w", err)
-	}
-
-	if s.Aliases == nil {
-		s.Aliases = make(map[string]string)
 	}
 
 	return &s, nil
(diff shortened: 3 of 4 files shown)


krissetto:
I didn't really dig into the implementation here, but I'm not really sure this is very useful.. The `register` command seems to register aliases for agent filenames (e.g. `code` instead of `./some-coding-agent.yaml`), the `please` command seems equivalent to the `run` command.

I think this could be done equivalently with a single command (e.g. `register` or `alias` etc.), and we could integrate the check for aliases directly into the `run` command similar to how we can run `namespace/artifactName` and .yaml files already.

The main benefit i see if only shortening the agent name, is there something else i'm missing?

jeanlaurent:
If I boils down your point 
1. So rename register to alias --> ok 
2. Remove please de the logic of running aliases behind run

All of this make perfect sense from of pure flag logic pov. 

I wanted to have this conversation to put commands/aliases at the top level of the cli. Like cagent code. 

I think the ultimate idea/goal from a user standpoint would be to have a few hand picked agents that you use on a daily basis with 'cagent code' or 'cagent check-gh-issue'. Now i wasn't sure really of the complexity of this implementation especially regarding the potential name clash with existing cli
Commands. 

I did the suggested change, but let's have a conversation on top level cli commands before.

jeanlaurent (commented):
## 📋 Review Summary

This PR introduces a valuable shortcut management system for cagent configurations through aliases. The implementation appears well-thought-out with clear usage patterns and a sensible resolution order for aliases.

## 🔍 General Feedback

- The feature adds good user experience improvements by allowing shortcuts to frequently used agents
- The resolution order (alias → local file → catalog reference) is logical and maintains backward compatibility
- The command structure and usage examples are clear and follow CLI best practices
- Using ~/.config/cagent/shortcuts.yaml follows standard XDG configuration conventions

To provide more detailed feedback on the implementation details and code changes, I would need access to the actual code changes in the PR.

cmd/root/alias.go
```diff
@@ -0,0 +1,200 @@
...
+				telemetry.TrackCommand("alias", args)
+				return createAlias(args[0], args[1])
+			}
+
+			// Single argument is invalid in this context
+			return fmt.Errorf("invalid arguments: use 'alias list' or 'alias <name> <agent-path>'")
+		},
+	}
+
+	cmd.AddCommand(newAliasListCmd())
```
dgageot, line 36: It would be more standard to have a `cagent alias add` command.

Right now, if I type `cagent alias`, I expect to see the sub commands but that's not the case.

cmd/root/alias.go
```diff
@@ -0,0 +1,200 @@
...
+			telemetry.TrackCommand("alias", []string{"list"})
+			return listAliases()
+		},
+	}
+}
+
+// newAliasRemoveCmd creates the remove subcommand
+func newAliasRemoveCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "remove <alias-name>",
```
dgageot: Can you add `rm` as an alias to this command?

cmd/root/alias.go
```diff
@@ -0,0 +1,200 @@
...
+	cmd.AddCommand(newAliasListCmd())
+	cmd.AddCommand(newAliasRemoveCmd())
+
+	return cmd
+}
+
+// newAliasListCmd creates the list subcommand
+func newAliasListCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "list",
```
dgageot: Can you add `ls` as an alias to this command?

cmd/root/alias.go
```diff
@@ -0,0 +1,200 @@
...
+			telemetry.TrackCommand("alias", append([]string{"remove"}, args...))
+			return removeAlias(args[0])
+		},
+	}
+}
+
+// createAlias creates a new alias
+func createAlias(name, agentPath string) error {
+	// Load existing aliases
+	s, err := shortcuts.Load()
```
dgageot: should le package be called `aliases` instead of `shortcuts`?

cmd/root/alias.go
```diff
@@ -0,0 +1,200 @@
...
+	fmt.Printf("  Alias: %s\n", name)
+	fmt.Printf("  Agent: %s\n", agentPath)
+	fmt.Printf("\nYou can now run: cagent run %s\n", name)
+
+	return nil
+}
+
+// removeAlias removes an alias
+func removeAlias(name string) error {
+	s, err := shortcuts.Load()
```
dgageot: It would be nice to have a `--force` added in a future PR

cmd/root/alias.go
```diff
@@ -0,0 +1,200 @@
...
+	return nil
+}
+
+// expandTilde expands the tilde in a path to the user's home directory
+func expandTilde(path string) (string, error) {
+	if !strings.HasPrefix(path, "~/") {
+		return path, nil
+	}
+
+	homeDir, err := filepath.Abs(filepath.Join(filepath.Dir(path), ".."))
```
dgageot: How is that a home dir?

cmd/root/run.go
```diff
@@ -94,6 +95,15 @@ func doRunCommand(ctx context.Context, args []string, exec bool) error {
 	slog.Debug("Starting agent", "agent", agentName, "debug_mode", debugMode)
 
 	agentFilename := args[0]
+
+	// Try to resolve as an alias first
+	if aliases, aliasErr := shortcuts.Load(); aliasErr == nil {
```
dgageot: Can you use `err` instead?

pkg/shortcuts/shortcuts.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// LoadFrom loads shortcuts from a specific file path
+func LoadFrom(path string) (*Shortcuts, error) {
+	// If file doesn't exist, return empty shortcuts
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return &Shortcuts{Shortcuts: make(map[string]string)}, nil
+	}
+
+	data, err := os.ReadFile(path)
```
dgageot: You don't need to stat if you read

pkg/shortcuts/shortcuts.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, fmt.Errorf("failed to read shortcuts file: %w", err)
+	}
+
+	var s Shortcuts
+	if err := yaml.Unmarshal(data, &s); err != nil {
+		return nil, fmt.Errorf("failed to parse shortcuts file: %w", err)
+	}
+
+	if s.Shortcuts == nil {
+		s.Shortcuts = make(map[string]string)
```
dgageot: We can probably move that to line 40

pkg/shortcuts/shortcuts_test.go
```diff
@@ -0,0 +1,136 @@
+package shortcuts
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestShortcuts(t *testing.T) {
+	// Create a temporary directory for testing
```
dgageot: We don't need that comment

pkg/shortcuts/shortcuts_test.go
```diff
@@ -0,0 +1,136 @@
...
+	tmpDir := t.TempDir()
+	shortcutsFile := filepath.Join(tmpDir, "shortcuts.yaml")
+
+	t.Run("Load empty shortcuts", func(t *testing.T) {
+		s, err := LoadFrom(shortcutsFile)
+		if err != nil {
```
dgageot: You should use require.NoError(t, err)

pkg/shortcuts/shortcuts_test.go
```diff
@@ -0,0 +1,136 @@
...
+			t.Fatalf("Failed to load shortcuts: %v", err)
+		}
+		if len(s.Shortcuts) != 0 {
```
dgageot: You should use assert.Len(...)

cmd/root/alias.go
jeanlaurent: done

cmd/root/alias.go
jeanlaurent: done

cmd/root/alias.go
jeanlaurent: done

cmd/root/alias.go
jeanlaurent: done

pkg/shortcuts/shortcuts_test.go
jeanlaurent: done

pkg/shortcuts/shortcuts_test.go
jeanlaurent: done

pkg/shortcuts/shortcuts_test.go
jeanlaurent: done

cmd/root/alias.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/docker/cagent/pkg/telemetry"
+)
+
+// NewAliasCmd creates a new alias command for managing aliases
+func NewAliasCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "alias",
+		Short: "Manage aliases for agents",
+		Long:  `Create and manage aliases for agent configurations or catalog references.`,
+		Example: `  # Create an alias for a catalog agent
```
dgageot, line 22: We could start by making this command Hidden until it fully works

pkg/shortcuts/shortcuts.go
jeanlaurent: done

cmd/root/run.go
jeanlaurent: done

cmd/root/alias.go
jeanlaurent: Ooops, big error here. Fixed

PR #519, opened 2025-10-16
Slash commands

### LINKED ISSUE #517: Add `/` commands to the TUI
Make it possible to call a command directly from the user input:

- `/copy`
- `/compact`
- `/reset`

Image

jeanlaurent:
The only way to copy from the clipboard is through CMD-P then select copy.
This adds `/copy`, `/reset and `/compact` commands in all modes (including --tui=false).

Fixes #517

### DIFF
--- a/pkg/tui/page/chat/chat.go
+++ b/pkg/tui/page/chat/chat.go
@@ -182,6 +182,10 @@ func (p *chatPage) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return p, cmd
 
 	case editor.SendMsg:
+		// Check for slash commands first
+		if handled, cmd := p.handleSlashCommand(msg.Content); handled {
+			return p, cmd
+		}
 		cmd := p.processMessage(msg.Content)
 		return p, cmd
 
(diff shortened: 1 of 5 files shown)


krissetto:
Is this supposed to resolve #517? I'm not sure if /copy and /clipboard have the same meaning or if I'm missing some other detail

jeanlaurent:
Yes /clipboard might not be the best named function

pkg/tui/page/chat/chat.go
```diff
@@ -430,6 +434,31 @@ func (p *chatPage) switchFocus() {
 	}
 }
 
+// handleSlashCommand checks if the content is a slash command and handles it
+// Returns (true, cmd) if command was handled, (false, nil) otherwise
+func (p *chatPage) handleSlashCommand(content string) (bool, tea.Cmd) {
```
rumpl, line 439: These commands already exist in the tui.go 

https://github.com/docker/cagent/blob/605590918ce98a5cf6053afacc0e965286def382/pkg/tui/tui.go#L293-L333

We should reuse them

dgageot:
@jeanlaurent @rumpl What's the status on this one?

dgageot:
@jeanlaurent I'm going to close this one. There are other PRs that try to address that

PR #520, opened 2025-10-16
Add a category to every builtin tool

dgageot:


dgageot:
@krissetto I'm trying to leverage them on the UI side

Issue #232, opened 2025-09-17
🐛 unexpected 400 Bad Request


nunocoracao:
**Describe the bug**

Getting the following error while running an agent in TUI

```
 │ error receiving from stream: POST "https://api.anthropic.com/v1/messages": 400 Bad Request (Request-ID: req_011CTEDQrFjrizY9CJfxJwAg)
 {"type":"error","error":{"type":"invalid_request_error","message":"messages.66: `tool_use` ids were found without `tool_result` blocks immediately after:
 toolu_01Hn3XhD7dhiFWRwn3VSh3RY. Each `tool_use` block must have a corresponding `tool_result` block in the next
 message."},"request_id":"req_011CTEDQrFjrizY9CJfxJwAg"}
```

**Version affected**

```
cagent version dev
Build time: 2025-09-17T14:27:16Z
Commit: 7c227c1f424a2d813cf17edf0e0a19040212d2b7
```

**How To Reproduce**

Not sure, it stopped happening once I started a new session

**OS and Terminal type**

Please include what OS and terminal you are using

**Additional context**

Any other info you consider useful can be included here

krissetto:
is this still reproducible?
if so, can you include a repro

jeanlaurent:
Yes 

```
time=2025-10-16T03:06:25.483Z level=ERROR msg="Error handling stream" agent=root error="error receiving from stream: POST \"[https://api.anthropic.com/v1/messages\](https://api.anthropic.com/v1/messages/)": 400 Bad Request (Request-ID: req_011CUA8zAg43gFz2pbeXWPr1) {\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"messages.17: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_018K3Z91JokEQMQtDZh8ToKM. Each `tool_use` block must have a corresponding `tool_result` block in the next message.\"},\"request_id\":\"req_011CUA8zAg43gFz2pbeXWPr1\"}"
I see I need to create a new pending review first:
[...]

PR #521, opened 2025-10-16
Add msg sequencing guards for anthropic

krissetto:
Adds logs for scenarios where msg sequencing is wrong, and adds a basic repair functionality for these issues to not block agent loops if this happens so the agent can try to fix its own problems (adds missing tool responses with `(tool execution failed)` so the agent can continue its work knowing something went wrong)

references #232


krissetto:
I have a feeling most of the underlying errors will be solved with #522, this is mostly for extra safety and logging purposes

PR #526, opened 2025-10-16
Handle long-running shell commands

trungutt:
This PR attempts to suggestion a solution to https://github.com/docker/cagent/issues/505 and https://github.com/docker/cagent/issues/34

Previously, shell commands would block indefinitely waiting for completion, causing the events channel to never close and making the system appear hung when running long-running commands like "npm run dev" or background services.

This commit implements automatic timeout detection with streaming output:

- Commands completing within 30 seconds return synchronously (backward compatible)
- Commands exceeding 30 seconds automatically switch to background streaming mode
- Background goroutine streams incremental output every 2 seconds via events
- Tool handler returns immediately after timeout, allowing conversation to continue
- Adds an optional `Background` boolean parameter to the shell tool, allowing LLMs to explicitly request
  background execution for long-running commands

### DIFF
--- a/pkg/tools/builtin/shell.go
+++ b/pkg/tools/builtin/shell.go
@@ -52,7 +52,10 @@ func (h *shellHandler) RunShell(ctx context.Context, toolCall tools.ToolCall) (*
 		return nil, fmt.Errorf("invalid arguments: %w", err)
 	}
 
-	cmd := exec.CommandContext(ctx, h.shell, append(h.shellArgsPrefix, params.Cmd)...)
+	// Use context.Background() instead of ctx to prevent command interruption when
+	// the incoming context is cancelled (e.g., when an API request ends).
+	// Commands should only be stopped explicitly via Stop() or process cleanup.
+	cmd := exec.CommandContext(context.Background(), h.shell, append(h.shellArgsPrefix, params.Cmd)...)
 	cmd.Env = h.env
 	if params.Cwd != "" {
 		cmd.Dir = params.Cwd


pkg/tools/builtin/shell.go
```diff
@@ -23,11 +25,19 @@ type ShellTool struct {
...
+}
+
+type backgroundCommand struct {
+	cmd    *exec.Cmd
+	outBuf *bytes.Buffer
+	errBuf *bytes.Buffer
+	done   chan error
 }
 
 type RunShellArgs struct {
```
rumpl, line 43: We should add a `Background bool` to the shell tool, LLMs know about `npm start` and such so they could decided on their own that the tool should be run in the background

pkg/tools/builtin/shell_test.go
```diff
@@ -84,7 +87,7 @@ func TestShellTool_HandlerEcho(t *testing.T) {
 	// Get handler from tool
 	tls, err := tool.Tools(t.Context())
 	require.NoError(t, err)
-	require.Len(t, tls, 1)
+	require.Len(t, tls, 2) // shell and get_logs
```
rumpl: All these comments can be removed

pkg/tui/page/chat/chat.go
```diff
@@ -196,9 +196,6 @@ func (p *chatPage) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	case *runtime.MCPInitFinishedEvent:
 		spinnerCmd := p.sidebar.SetMCPInitializing(false)
 		return p, spinnerCmd
-	case *runtime.ShellOutputEvent:
```
rumpl, line 199: Should we remove the ShellOutputEvent completely?

trungutt: Oups, I'd restore this in the PR. Will make another one after this one

pkg/tools/builtin/shell.go
trungutt: Done

pkg/tools/builtin/shell_test.go
```diff
@@ -244,3 +252,154 @@ func TestShellTool_ParametersAreObjects(t *testing.T) {
...
+	var pid int
+	_, err = fmt.Sscanf(result.Output, "Command is running in background (PID: %d)", &pid)
+	require.NoError(t, err)
+	assert.Positive(t, pid)
+
+	// Wait a bit for command to produce output
+	// (in real usage, the LLM would call get_logs when it wants to check)
+	// Using a small sleep to let the echo complete
+	// Note: sleep 1 in the command ensures it stays alive long enough
+	// time.Sleep(100 * time.Millisecond)
```
rumpl: Is this needed or can we remove the whole comment block?

pkg/tools/builtin/shell.go
```diff
@@ -91,27 +111,139 @@ func (h *shellHandler) RunShell(ctx context.Context, toolCall tools.ToolCall) (*
...
-	output := outBuf.String() + errBuf.String()
+	// Read stderr
+	go func() {
+		defer wg.Done()
+		_, _ = io.Copy(&errBuf, stderrPipe)
+	}()
 
-	if err != nil {
+	// Wait for command completion with timeout
+	const quickCommandTimeout = 30 * time.Second
```
rumpl: Let's put this const at the top of the file, also 30 seconds is maybe too long? Let's try with 10 maybe?

pkg/tools/builtin/shell.go
rumpl: Since the command is called with the context here it will be interrupted once the context is done. In the api for example the context is the request context so the command will be interrupted once the request is done, we don't want that.

rumpl (changes requested):

pkg/tools/builtin/shell.go
trungutt: Done

pkg/tools/builtin/shell.go
```diff
@@ -41,7 +52,10 @@ func (h *shellHandler) RunShell(ctx context.Context, toolCall tools.ToolCall) (*
 		return nil, fmt.Errorf("invalid arguments: %w", err)
 	}
 
-	cmd := exec.CommandContext(ctx, h.shell, append(h.shellArgsPrefix, params.Cmd)...)
+	// Use context.Background() instead of ctx to prevent command interruption when
+	// the incoming context is cancelled (e.g., when an API request ends).
+	// Commands should only be stopped explicitly via Stop() or process cleanup.
+	cmd := exec.CommandContext(context.Background(), h.shell, append(h.shellArgsPrefix, params.Cmd)...)
```
rumpl, line 58: I'm really not a fan of this actually, if I, the user, see that a command is taking too long I want to be able to stop the stream, which would stop the command and continue talking to the agent. With the code as is, even if I stop the stream the command will continue running in the background...

With the `background` parameter things will be better no?

PR #540, opened 2025-10-17
Attempt to fix SQLite database locking issues with concurrent access

trungutt:
Fixes https://github.com/docker/cagent/issues/544.

This PR configures SQLite connections with proper concurrency handling to prevent "database is locked" errors when multiple operations access the database simultaneously.

Details:
- Add busy_timeout (5s) to wait for locks instead of failing immediately
- Enable WAL mode for better concurrent read/write performance
- Configure connection pool to serialize writes (SQLite best practice)

Fixes session store and memory database SQLite implementations.


krissetto (approved):
there are more sqlite optimizations we could do, but lets get this in for now

PR #541, opened 2025-10-17
Use our own markdown rendering

dgageot:
It's faster, has fewer bugs and is easier to fix and customize


rumpl:
Yay!

rumpl:
When this is merged we should create issues to make this better, mainly:

- make the colors nicer
- there is no syntax colors for code blocks
- table rendering is not great

dgageot:
@rumpl I'll try to fix some of those before merging

dgageot:
I'll close this for now. The main issue with block quotes is fixed.
However, we still need a better markdown rendering

PR #546, opened 2025-10-17
Use appropriate amount of context for title generation

### LINKED ISSUE #545: Title generation breaks on long responses
Session title generation can break if an agent replies with a log of text, iterates a lot using tools etc.

Example:
- A user asks an agent for something complex as their first message (e.g. "write an spa app in react that does a, b, c, etc.");
- The agent iterates a lot and/or generates a lot of tokens because of tool usage etc;
- The title generation goes over the context size because we currently use all the messages in the thread (system prompt and tool responses included) to generate a title

krissetto:
Changes session title generation to only consider the first user message

Closes #545

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -972,29 +972,26 @@ func (r *runtime) generateSessionTitle(ctx context.Context, sess *session.Sessio
 	var conversationHistory strings.Builder
 	messages := sess.GetAllMessages()
 	for i := range messages {
-		role := "Unknown"
-		switch messages[i].Message.Role {
-		case "user":
-			role = "User"
-		case "assistant":
-			role = "Assistant"
-		case "system":
-			role = "System"
+		if messages[i].Message.Role == chat.MessageRoleUser {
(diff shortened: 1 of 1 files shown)


rumpl:
We should make it even simpler and just use the first user message when generating a title

krissetto:
> We should make it even simpler and just use the first user message when generating a title

I'm not convinced we should.. I did that in my initial implementation, but it gives quite bad results if the user message isn't specific enough to begin with.

As a basic example think of more open ended messages like "tell me a story". With messages like that the title can really only end up being something like "A story".  
If the assistant replies with a story about a goose at least the title could end up being something like "adventures of bob the silly goose".

If we really don't want that and going from one extreme (all messages) to the other (only first user msg) is preferred, I'll make those changes, but given the amount of code involved i don't think its a huge amount of logic to keep (most of the changes in this PR are from #543)

rumpl:
> but it gives quite bad results if the user message isn't specific enough to begin with.

Which is fine

krissetto:
> Which is fine

Strong disagree here, but that's fine. I changed this to only consider the first user message as-is so we can keep things moving.

PR #547, opened 2025-10-17
Fix oauth with streamable mcp servers

### LINKED ISSUE #485: OAuth flow failure Prisma MCP Server

rumpl:
req.Clone makes a shallow copy of the body so the first call would
consume the body completely, making the next call after oauth flow an
empty request, this would give us a 400 Bad Request.

Keep the body so that we can re-inject it into the original request
before re-sending with the token.

Fixes #485

### DIFF
--- a/pkg/tools/mcp/oauth.go
+++ b/pkg/tools/mcp/oauth.go
@@ -336,6 +336,16 @@ type oauthTransport struct {
 }
 
 func (t *oauthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	var bodyBytes []byte
+	if req.Body != nil && req.Body != http.NoBody {
+		var err error
+		bodyBytes, err = io.ReadAll(req.Body)
+		if err != nil {
+			return nil, err
+		}
+		req.Body = io.NopCloser(strings.NewReader(string(bodyBytes)))
+	}
+
 	reqClone := req.Clone(req.Context())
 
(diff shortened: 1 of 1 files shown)


pkg/tools/mcp/oauth.go
```diff
@@ -336,6 +336,16 @@ type oauthTransport struct {
 }
 
 func (t *oauthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	var bodyBytes []byte
+	if req.Body != nil && req.Body != http.NoBody {
+		var err error
+		bodyBytes, err = io.ReadAll(req.Body)
```
findleyr, line 342: Hi, @jba referred me here as I'm fixing this in the MCP client implementation.

I think you need to close the body after reading it, as RoundTripper says that RoundTrip must always close the body.

PR #552, opened 2025-10-19
Fix/tui token count fix

### LINKED ISSUE #504: Token tracking when using subagents
At the moment the token tracking code in the TUI only seems to consider the currently active agent/session. 

This means that the values displayed in the TUI change when an agent delegates a task to a sub-agent or a sub-agent finishes its task and returns control to it's parent, not giving the user a precise indication of the tokens actually used when running an agent team

It'd be nice to have an overall view of total tokens consumed by the full agent team, and also a way to see the consumption and cost per individual session (which would mean per sub-agent as well).

riturajFi:
#504


riturajFi:
@krissetto kindly review this solution

riturajFi:
Hi @krissetto , i found out that there were some double counting issues - i am fxing them now

krissetto:
Hey @riturajFi! :wave: is this ready for testing? If not we should put the PR back into draft for the meantime

riturajFi:
Hi @krissetto , the PR is ready for testing - i am just adding some comments, removing some unnecessary stuff and cleaning it up

riturajFi:
Hi @krissetto , would love to see your reviews on this. Kindly suggest any changes and i shall do them asap. Would love seeing it merged.

krissetto:
Hey @riturajFi! I'm not at my machine right now, but I'll give it a look and try it out as soon as I can over the weekend

riturajFi:
Thanks a lot @krissetto . Please assign me one more issue in tve meanwhile. I shall work on it! It's very insignful to work on first principle projects like these.

krissetto:
sure! wanna take a look at #665? shouldn't require many changes

pkg/runtime/usage_tracker.go
krissetto, line 1: I think we can simplify things and don't really need this. Each session already contains its own usage data, so we could use less code to implement this total count by traversing the session tree (going from the root agent's session down to its sub-agent sessions) instead of having a separate tracking mechanism

pkg/tui/components/sidebar/sidebar.go
```diff
@@ -91,6 +95,34 @@ func formatTokenCount(count int) string {
 	return fmt.Sprintf("%d", count)
 }
 
+// formatCost renders small currency values with enough precision for tiny spends.
+func formatCost(cost float64) string {
```
krissetto: nit: I don't think we need to display sub 1cent spending, shorter numbers can help with formatting in the limited sidebar space too

pkg/runtime/usage_tracker.go
krissetto: Actually, even simpler and less code to implement would probably be to append the values from the `emitUsageEvent` events in the TUI code, and not traverse sessions at all. What do you think?

riturajFi:
Thanks @krissetto . I am on it

riturajFi:
Hi @krissetto , opened a new PR with a completely new apprach and minimal changes so clsing this one

PR #561, opened 2025-10-20
Pagination for session messages

stanislavHamara:
A session can have 10000s of messages potentially. Pagination makes the consumption of messages easier for GUI, with a potential to implement "Load more" or infinite scroll approaches for chats.

### DIFF
--- a/pkg/api/pagination.go
+++ b/pkg/api/pagination.go
@@ -38,7 +38,7 @@ func PaginateMessages(messages []session.Message, params PaginationParams) ([]se
 	}
 
 	startIdx := 0
-	endIdx := totalCount
+	var endIdx int
 
 	if params.Before != "" {
 		endIdx = beforeIndex


pkg/api/pagination.go
```diff
@@ -0,0 +1,112 @@
...
+		metadata.HasMore = startIdx > 0
+	} else {
+		metadata.HasMore = endIdx < totalCount
+	}
+
+	if len(paginatedMessages) > 0 {
+		lastMsg := paginatedMessages[len(paginatedMessages)-1]
+		lastIdx := endIdx - 1
+		nextCursor := MessageCursor{
+			Timestamp: lastMsg.Message.CreatedAt,
```
rumpl: The timestamp is only returned but not checked anywhere, do we really need a timestamp?

We can either:
- use the timestamp somewhere to check that the request is valid
- remove the timestamp

pkg/api/types.go
```diff
@@ -126,14 +126,24 @@ type SessionsResponse struct {
...
+	OutputTokens  int                 `json:"output_tokens"`
+	WorkingDir    string              `json:"working_dir,omitempty"`
+	Pagination    *PaginationMetadata `json:"pagination,omitempty"`
+}
+
+// PaginationMetadata contains pagination information
+type PaginationMetadata struct {
+	TotalMessages int    `json:"total_messages"`        // Total number of messages in session
+	Limit         int    `json:"limit"`                 // Number of messages in this response
+	HasMore       bool   `json:"has_more"`              // Whether more messages exist
```
rumpl: There are two cursors, next and prev, but only one "has_more", how do we know which way it has more messages? Is this needed really? An empty cursor could mean that there are no more messages in that direction.

pkg/api/pagination.go
stanislavHamara: The timestamp is used for creating the cursor hash in the Encode function. Is that what you are asking?

pkg/api/types.go
stanislavHamara: Yeah you are right, tbh the next cursor here is actually not needed at all as we will always start with the most recent messages and we only need a single direction cursor to load more older messages.

As far as having `hasMore` I just saw that as a standard in paginated responses. I am happy to go either way, empty cursor will work too.

pkg/api/pagination.go
rumpl: Sure, but you are not using it anywhere else, like when you get a request with a cursor you are not checking it, it's completely unused it seems

stanislavHamara: Ah fair, understand now, index is sufficient for the cursor.

dgageot (approved):

PR #563, opened 2025-10-20
Continue simplifying tools

dgageot:


pkg/tools/builtin/filesystem.go
```diff
@@ -337,20 +329,7 @@ func (t *FilesystemTool) Tools(context.Context) ([]tools.Tool, error) {
 				Title: "Write File",
 			},
 		},
-	}
-
-	if len(t.allowedTools) == 0 {
```
rumpl, line 342: Ok for now but it's going to bite us one day, the system prompt for a toolset should know what tools it has so that it can tweak itself.

dgageot: One day, yes

rumpl (approved):

PR #564, opened 2025-10-20
Fix shell process cleanup

### LINKED ISSUE #562: Fails to build on Windows platform
relates to https://github.com/docker/packaging/actions/runs/18655026306/job/53181921407?pr=258#step:7:1673

```
0.286 + go build -trimpath -ldflags '-w -X '\''github.com/docker/cagent/pkg/version.Version=v0.0.0-20251020133613-26330cc'\'' -X '\''github.com/docker/cagent/pkg/version.Commit=26330cc0e1579ab91fd24a3459986256680d8330'\''' -o /build/cagent/cagent.exe .
0.286 /usr/local/src/cagent /build
28.48 # github.com/docker/cagent/pkg/tools/builtin
28.48 pkg/tools/builtin/shell.go:59:4: unknown field Setpgid in struct literal of type syscall.SysProcAttr
28.48 pkg/tools/builtin/shell.go:260:16: undefined: syscall.Kill
```

we should also look to have this target built in GHA to check for regressions.

rumpl:
Instead of using `exec.CommandContext` and keeping the process handlers in the tool, select either the tool completion or the context cancellation, in the case a cancelled context, terminate the whole process group.

Draft for now, need to do it for windows too.

Fixes #562 (once I'm done)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,28 @@
+# Auto detect text files and normalize to LF
+* text=auto eol=lf
+
+# Go source files
+*.go text eol=lf
+
+# Shell scripts
+*.sh text eol=lf
+
+# Windows specific files
+*.bat text eol=crlf
+*.cmd text eol=crlf
+*.ps1 text eol=crlf
+
+# Binary files
+*.exe binary
+*.dll binary
+*.so binary
+*.dylib binary
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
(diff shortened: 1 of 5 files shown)


trungutt:
But it still doesn't handle cleanly long-running process yet

rumpl:
If the LLM starts a `npm dev` that never stops then you can cancel the stream and the thing will end. Here I am only handling clean stop of shell tools.

Background tools will come later if needed

PR #567, opened 2025-10-21
Add bake definition

crazy-max:
First containerized most of the tasks to avoid drifting with external tooling like golangci-lint, go version and so on.

Adds bake definition so we can avoid arbitrary targets from Taskfile and be consistent and repro between inner loop and ci environment. Here is a breakdown of the changes:

* The `local` stage in the Dockerfile has been removed since we now create non opinionated binary name.
* A `releaser` stage has been added to create proper release artifacts if we want to upload them as GitHub Releases in the future.
* Git info has been containerized to make everything portable.
* The new bake definition includes the necessary targets for inner loop and ci environment:
  * Defaults to building `binaries` if no target is specified
  * `validate` target group runs `lint`, `test` and `licenses`
  * `image` and `image-cross` build the cagent image
  * `release` target  does a cross-build of the binaries and output release artifacts to `./bin/release`
  * `binaries` and `binaries-cross` build binaries and output to `./bin/build`
  * `test` run go test with coverage and output coverage result to `./bin/coverage` that could then be uploaded to codecov if we want to, like we do for other projects https://app.codecov.io/gh/docker
  * `licenses` runs go-licenses in a containerized env without need to install the binary.
  * `lint` runs golangci-lint also in a containerized env without need to install the binary.
* GitHub Actions now uses bake to run validation, building binaries and the image. No need for extra tools to be installed on the runner. Just Docker.
  * I took the opportunity to do cross compilation for binaries related to https://github.com/docker/cagent/issues/562 so we can make sure there is no regression. Also uploads binaries to GitHub Artifacts. We could potentially create a GitHub Release when a tag is pushed and upload the binaries there but I let that for a follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,11 +22,11 @@ jobs:
       includes: ${{ steps.generate.outputs.matrix }}
     steps:
       - name: Checkout
-        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+        uses: actions/checkout@v5
       -
         name: Generate matrix
         id: generate
-        uses: docker/bake-action/subaction/matrix@3acf805d94d93a86cce4ca44798a76464a75b88c # v6.9.0
+        uses: docker/bake-action/subaction/matrix@v6
         with:
           target: validate
 
@@ -40,10 +40,10 @@ jobs:
         include: ${{ fromJson(needs.prepare.outputs.includes) }}
     steps:
       - name: Checkout
-        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+        uses: actions/checkout@v5
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
+        uses: docker/setup-buildx-action@v3
 
       - name: Validate
         uses: docker/bake-action@v6
@@ -56,11 +56,11 @@ jobs:
       includes: ${{ steps.generate.outputs.matrix }}
     steps:
       - name: Checkout
-        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+        uses: actions/checkout@v5
 
       - name: Generate matrix
         id: generate
-        uses: docker/bake-action/subaction/matrix@3acf805d94d93a86cce4ca44798a76464a75b88c # v6.9.0
+        uses: docker/bake-action/subaction/matrix@v6
         with:
           target: release
           fields: platforms
@@ -83,12 +83,12 @@ jobs:
           echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
 
       - name: Checkout
-        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
+        uses: docker/setup-buildx-action@v3
 
       - name: Build
         uses: docker/bake-action@v6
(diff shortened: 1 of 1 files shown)


pkg/memory/database/sqlite/sqlite_test.go
```diff
@@ -39,7 +39,7 @@ func TestNewMemoryDatabase(t *testing.T) {
 	assert.NotNil(t, db, "Database should be created successfully")
 
 	// Test with invalid path
-	_, err := NewMemoryDatabase("/:invalid:path")
```
crazy-max, line 42: `/:invalid:path` is technically valid:

```
#18 6.407 --- FAIL: TestNewMemoryDatabase (0.05s)
#18 6.407     sqlite_test.go:43:
#18 6.407               Error Trace:    /src/pkg/memory/database/sqlite/sqlite_test.go:43
#18 6.407               Error:          An error is expected but got nil.
#18 6.407               Test:           TestNewMemoryDatabase
#18 6.407               Messages:       Should fail with invalid database path
#18 6.407 FAIL
```

.dockerignore
```diff
@@ -1,6 +1 @@
-*
```
rumpl, line 1: @dgageot prefers the "Ignore everything first and then add things that are needed one by one"

Taskfile.yml
```diff
@@ -1,77 +1,31 @@
...
-    sources:
-      - "{{.GO_SOURCES}}"
-      - ".golangci.yml"
+    cmds:
+      - ln -sf {{.USER_WORKING_DIR}}/{{.DESTDIR}}/cagent {{.HOME}}/bin/cagent
 
   test:
     desc: Run tests
-    cmd: go test ./...
-    sources:
```
rumpl, line 56: Why did you remove the sources? Is this even needed now that we have the bake target, or maybe this should just call bake?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/docker/cagent
 
-go 1.25.3
```
rumpl, line 3: Why?

rumpl: I'm sure there is a completely valid reason, I would just like to know why please :)

.dockerignore
dgageot: I do :-)

go.mod
crazy-max: The `go` directive should theoretically declare the **minimum required Go version** for the module. The use of a patch version in `go.mod` (e.g., `go 1.21.0`) was introduced in Go 1.21:

> Go 1.21 introduces a small change to the numbering of releases. In the past, we used Go 1.N to refer to both the overall Go language version and release family as well as the first release in that family.  
> — [Go 1.21 Release Notes](https://tip.golang.org/doc/go1.21)

In other words, `go 1.21` now refers to the *release family*, while `go 1.21.0` refers to a specific version.

As of Go 1.21 and higher:

> * The `go` line declares a required minimum version of Go to use with this module.  
> * …  
> — [Module Reference](https://go.dev/ref/mod#go-mod-file-go)

It’s best practice to specify `.0` for the patch version — for example, `go 1.21.0`.  
Patch releases are, by design, **non-breaking**: they contain only backward-compatible fixes and improvements. Using `.0` signals that the module is compatible with the entire patch series (`1.21.x`) without requiring users to upgrade to a specific patch version.  

This approach ensures that we only compel users to upgrade their Go toolchain when the module genuinely depends on a feature introduced in a newer Go release. In general, keeping the patch version at `.0` helps maintain compatibility and reduces friction for downstream users integrating the module.

Taskfile.yml
crazy-max: I just kept this target so you can still run tests natively on your host. Specially important on darwin/windows that we can't run in containers.

go.mod
crazy-max: But now looking at https://github.com/docker/cagent/commit/5dda943bb5818f6b8ed0144e1377acfd49c6a3c0

```
go: github.com/alpkeskin/gotoon@v0.1.0: module github.com/alpkeskin/gotoon@v0.1.0 requires go >= 1.25.3
```

We need to set minimum to 1.25.3 anyway.

.dockerignore
crazy-max: Np, updated

crazy-max: Hum this breaks the versioning: https://github.com/docker/cagent/actions/runs/19065808007/job/54456250623?pr=567#step:5:402

```
#23 0.115 + xx-go build -trimpath -ldflags '-s -w -X github.com/docker/cagent/pkg/version.Version=v1.9.7-68-g4acaa88.m -X github.com/docker/cagent/pkg/version.Commit=4acaa88ba46dfdb9108419234d1a385935dafa80.m' -o /binaries/cagent .
```

Because the working tree is not clean within the container now.

dgageot:
@rumpl @crazy-max what do we do with this?

crazy-max:
> @rumpl @crazy-max what do we do with this?

This is not portable and fully containerized with current dockerignore strategy https://github.com/docker/cagent/pull/567#discussion_r2489884128

dgageot:
@crazy-max I'm going to close this one. Sorry. I know it's hard work but we are not using it

PR #568, opened 2025-10-21
Implement ACP

rumpl:
This implements the ACP protocol in cagent, you can now configure your ACP client with `cagent acp agent.yaml` and then you can use your agent in your IDE.

Start of a session
Screenshot 2025-10-23 at 11 00 27

Tool confirmation
Screenshot 2025-10-23 at 11 00 45

Final thing, with file edit tracking in Zed.
Screenshot 2025-10-23 at 11 00 58

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -57,6 +57,9 @@ $ cagent exec config.yaml --yolo          # Run the agent once and auto-accept a
 $ cagent api config.yaml
 $ cagent api config.yaml --listen :8080
 
+# ACP Server (Agent Client Protocol via stdio)
+$ cagent acp config.yaml                 # Start ACP server on stdio
+
 # Other commands
 $ cagent new                          # Initialize new project
 $ cagent new --model openai/gpt-5-mini --max-tokens 32000  # Override max tokens during generation
--- a/pkg/acp/agent.go
+++ b/pkg/acp/agent.go
@@ -1,0 +1,422 @@
+package acp
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"sync"
+
+	"github.com/coder/acp-go-sdk"
+	"github.com/google/uuid"
+
+	"github.com/docker/cagent/pkg/config"
+	"github.com/docker/cagent/pkg/runtime"
+	"github.com/docker/cagent/pkg/session"
+	"github.com/docker/cagent/pkg/team"
+	"github.com/docker/cagent/pkg/teamloader"
+	"github.com/docker/cagent/pkg/tools"
+)
+
+// Agent implements the ACP Agent interface for cagent
+type Agent struct {
+	conn          *acp.AgentSideConnection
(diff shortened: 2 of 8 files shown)


docs/USAGE.md
```diff
@@ -57,6 +57,9 @@ $ cagent exec config.yaml --yolo          # Run the agent once and auto-accept a
 $ cagent api config.yaml
 $ cagent api config.yaml --listen :8080
 
+# ACP Server (Agent Client Protocol via stdio)
```
krissetto, line 60: maybe here we could add a brief mention of what its for, and a link to https://agentclientprotocol.com/overview/introduction?

pkg/acp/agent.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	// Cancel any previous turn
+	a.mu.Lock()
+	if acpSess.cancel != nil {
+		prev := acpSess.cancel
+		a.mu.Unlock()
+		prev()
+		// Wait a bit for cancellation to complete
+		time.Sleep(100 * time.Millisecond)
```
krissetto: is the cancellation async here that we need to time.sleep?

rumpl:
Note: I am not handling embedded resources still, will do in a followup

pkg/acp/agent.go
```diff
@@ -0,0 +1,425 @@
...
+	for _, content := range params.Prompt {
+		if content.Text != nil {
+			userContent += content.Text.Text
+		}
+		if content.ResourceLink != nil {
+			slog.Debug("resource link", "link", content.ResourceLink)
+		}
+		if content.Resource != nil {
+			slog.Debug("embedded context", "context", content.Resource)
+			slog.Debug(content.Resource.Resource.TextResourceContents.Text)
```
krissetto, line 181: shall we add at least the debug logs for the other types of content as well so we dont forget to implement them later? (image, audio, blob resources)

https://agentclientprotocol.com/protocol/content#param-resource

krissetto (approved):
Just for context since i haven't gone deep in ACP, is filesystem the only toolset that needs/wiil need an ACP specific wrapper?

Also a quick mention in the README is probably warranted, to show users how they can use the project

pkg/acp/agent.go
rumpl: not needed, will remove

gianpaj:
Nice one! Could you provide an example `agent.yaml` please?

rumpl:
In the screenshot I was using the `golang_developer.yaml` that's in the root of this repo, I added this to my Zed config:

```
  "agent_servers": {
    "cagent": {
      "command": "cagent",
      "args": ["acp", "./golang_developer.yaml"]
    }
  }
```

You can also try the `examples/gopher.yaml` one, it's nice too.

You can also `cagent new` and ask it to generate a developer agent for you. Or you can craft your own that suits your needs.

marcellodesales:
@rumpl This definitely needs to be a Docker Event Session!

PR #570, opened 2025-10-21
Anthropic - set `max_tokens` based on available context

### LINKED ISSUE #569: Anthropic - input tokens + `max_tokens` > context length
When using the Anthropic provider and reaching the upper limits of the available context length, we can end up in scenarios where the input tokens + the max_tokens value is > than the max supported context length for a given model.

To fix this, we need to keep in consideration the length of the input and set  an appropriate `max_tokens` value before sending requests to the Anthropic API

Example error
```json
400 Bad Request

{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "input length and `max_tokens` exceed context limit: 178959 + 64000 > 200000, decrease input length or `max_tokens` and try again"
  },
  "request_id":"req_011CUB2XUHWkFnQzxcKBbjpF"
}
```

krissetto:
Calculate `max_tokens` value for Anthropic requests based on the available context **using the official token counting API**.

The PR fixes errors such as:

```json
400 Bad Request

{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "input length and `max_tokens` exceed context limit: 178959 + 64000 > 200000, decrease input length or `max_tokens` and try again"
  },
  "request_id":"req_011CUB2XUHWkFnQzxcKBbjpF"
}
```

Closes #569

### DIFF
--- a/pkg/model/provider/anthropic/beta_client.go
+++ b/pkg/model/provider/anthropic/beta_client.go
@@ -38,6 +38,16 @@ func (c *Client) createBetaStream(
 		}
 	}
 
+	sys := extractBetaSystemBlocks(messages)
+
+	if used, err := countAnthropicTokensBeta(ctx, client, anthropic.Model(c.ModelConfig.Model), converted, sys, allTools); err == nil {
+		configuredMaxTokens := maxTokens
+		maxTokens = clampMaxTokens(anthropicContextLimit(c.ModelConfig.Model), used, maxTokens)
+		if maxTokens < configuredMaxTokens {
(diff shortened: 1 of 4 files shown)


krissetto:
Right now I implemented this calculation using the official Anthropic count tokens API.

Another option would be to do some approximated calculation with some external library, and use that estimate instead of an additional API call.

rumpl:
Rather than doing double calls all the time, can we parse the error and do it only when needed?

krissetto:
> Rather than doing double calls all the time, can we parse the error and do it only when needed?

Sure, I'll take a look at that before merging

dgageot:
@krissetto I'll merge it first. Please look at the improvement in a second PR

PR #579, opened 2025-10-22
Nicer error when DD is not running

### LINKED ISSUE #571: better error messaging when docker not running
**What you'd like to see**

Repro steps:
- Turn off docker
- Run any agent with a docker-based mcp server (e.g., github.yaml)

Expected result:
a useful error message

Actual result:
Image

dgageot:
Fixes #571

### DIFF
--- a/examples/search.yaml
+++ b/examples/search.yaml
@@ -1,0 +1,11 @@
+#!/usr/bin/env cagent run
+version: "2"
+
+agents:
+  root:
+    model: anthropic/claude-sonnet-4-5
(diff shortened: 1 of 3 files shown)


rumpl (approved):

krissetto (approved):

dvdksn (commented):
I was trying to use cagent with mcp gateway standalone, on docker-ce, but I kept getting this error. So I guess what we really should check is if the gateway is running (?)

PR #582, opened 2025-10-22
Support attaching images via API

stanislavHamara:
### 1. API Types (`pkg/api/types.go`)
- Added `MultiContent []chat.MessagePart` field to `Message` struct
- Allows API clients to send messages with mixed content (text + images)

### 2. Server (`pkg/server/server.go`)
- **Critical Fix**: Updated `runAgent` handler to properly handle `MultiContent`
- When a message has `MultiContent`, create a session message with the full multi-content structure
- Falls back to plain text for messages without `MultiContent`
- Added import for `chat` package to access message types

### 3. Runtime (`pkg/runtime/runtime.go`)
- **Enhancement**: Updated `UserMessage` event emission to extract text from `MultiContent`
- When `Content` is empty but `MultiContent` exists, extract text parts for display
- This ensures UI/CLI displays meaningful text even for multi-content messages
- Note: This is cosmetic - the actual message processing was already working

## API Usage Example

```bash
# Encode image to base64
IMAGE_DATA=$(base64 -i passport.jpg)

# Send request with image attachment
curl -X POST http://localhost:8080/api/sessions/$SESSION_ID/agent/image_analyzer \
  -H "Content-Type: application/json" \
  -d '[{
    "role": "user",
    "content": "",
    "multi_content": [
      {"type": "text", "text": "Extract text from this image"},
      {"type": "image_url", "image_url": {"url": "data:image/jpeg;base64,'$IMAGE_DATA'"}}
    ]
  }]'
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -23,7 +23,6 @@ import (
 	"github.com/labstack/echo/v4/middleware"
 
 	"github.com/docker/cagent/pkg/api"
-	"github.com/docker/cagent/pkg/chat"
 	"github.com/docker/cagent/pkg/config"
 	latest "github.com/docker/cagent/pkg/config/v2"
 	"github.com/docker/cagent/pkg/content"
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -1089,14 +1090,29 @@ func (s *Server) runAgent(c echo.Context) error {
...
 		if err := json.NewDecoder(c.Request().Body).Decode(&messages); err != nil {
 			return echo.NewHTTPError(http.StatusBadRequest, "invalid request body")
 		}
 
 		for _, msg := range messages {
-			sess.AddMessage(session.UserMessage(agentFilename, msg.Content))
+			// Check if the message has multi-content (e.g., text + images)
+			if len(msg.MultiContent) > 0 {
+				// Create message with multi-content support
+				sess.AddMessage(&session.Message{
```
krissetto: nit: i know we're not currently super consistent with this across the codebase, but maybe we should adjust the `session.UserMessage` func to accept multipart content. WDYT?

pkg/runtime/runtime.go
```diff
@@ -238,7 +238,22 @@ func (r *runtime) RunStream(ctx context.Context, sess *session.Session) <-chan E
 
 		messages := sess.GetMessages(a)
 		if sess.SendUserMessage {
-			events <- UserMessage(messages[len(messages)-1].Content)
+			lastMsg := messages[len(messages)-1]
```
rumpl: Let's only send content for now, this event is kinda used for the TUI to know that the agent started, but now that we have the MCP servers loading events this might now be useful any more .

pkg/server/server.go
stanislavHamara: Have a look to see if that's what you had in mind or if it is too javascripty :D

PR #590, opened 2025-10-23
winget: automatically publish releases

chelnak:
Adds a new GitHub action that automates the publishing of new winget manifests.

The workflow is triggered when a release is published and requires a repository secret to be set that has public repo access.

Info on requirements for the token can be found [here](https://github.com/microsoft/winget-pkgs/pull/306270).

I made the assumption that your tags will always be your release version (that seems to be the case so far).

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,0 +1,24 @@
+name: winget
+
+on:
+  release:
+    types: [published]
+
+env:
+  WINGET_CREATE_GITHUB_TOKEN: ${{ secrets.WINGET_TOKEN }}
+
+jobs:
+  publish:
+    runs-on: windows-latest
+    steps:
+      - name: publish
+        run: |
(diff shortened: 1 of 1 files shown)


dgageot:
@chelnak I'm not sure I understand what needs to be configured from the link you shared

chelnak:
> @chelnak I'm not sure I understand what needs to be configured from the link you shared

Oh I must have copied the wrong link.

I'll find the correct one!

chelnak:
Here's the correct doc: 

https://github.com/microsoft/winget-create?tab=readme-ov-file#using-windows-package-manager-manifest-creator-in-a-cicd-pipeline

You'll need a GitHub token that has public repo access because it will create a PR to the `winget-pkgs` repository.

The workflow in this PR expects a secret named `WINGET_TOKEN`.

IIRC, `wingetcreate` works by forking the packages repository then creates the upstream PR.

PR #600, opened 2025-10-24
Fix stream cancellation in the TUI

### LINKED ISSUE #597: Cancelling a stream breaks the TUI
Canceling a response mid-stream leaves the TUI in a broken state where the "Working" and loading spinners remain active instead of also being stopped, and no further messages can be sent by the user

Screenshot of a canceled stream

Image

krissetto:
Fixes stream cancellation logic in the TUI by stopping and removing all spinners, allowing the user to send followup messages, and adds a message indicating to the user that the stream has been canceled

---

Screenshot

Screenshot From 2025-10-24 13-49-30

---

Closes #597

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -20,6 +20,11 @@ import (
 	"github.com/docker/cagent/pkg/tui/types"
 )
 
+// StreamCancelledMsg notifies components that the stream has been cancelled
+type StreamCancelledMsg struct {
+	ShowMessage bool // Whether to show a cancellation message after cleanup
+}
+
 // Model represents a chat message list component
 type Model interface {
 	layout.Model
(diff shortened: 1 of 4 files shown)


pkg/tui/components/messages/messages.go
```diff
@@ -31,12 +31,15 @@ type Model interface {
...
 	AddAssistantMessage() tea.Cmd
 	AddSeparatorMessage() tea.Cmd
+	AddCancelledMessage() tea.Cmd
 	AddOrUpdateToolCall(agentName string, toolCall tools.ToolCall, toolDef tools.Tool, status types.ToolStatus) tea.Cmd
 	AddToolResult(msg *runtime.ToolCallResponseEvent, status types.ToolStatus) tea.Cmd
 	AppendToLastMessage(agentName string, messageType types.MessageType, content string) tea.Cmd
 	ScrollToBottom() tea.Cmd
 	AddShellOutputMessage(content string) tea.Cmd
 	AddSystemMessage(content string) tea.Cmd
+	RemoveSpinner()
```
rumpl: Instead of making this public it would be nicer maybe to have a command that is sent down to all the components that tell them that the stream is cancelled, WDYT? Just asking

krissetto: Agree, I changed this

rumpl (approved):

PR #602, opened 2025-10-24
Auto-pull DMR models

dgageot:


pkg/model/provider/dmr/client.go
```diff
@@ -430,6 +436,56 @@ func parseDMRProviderOpts(cfg *latest.ModelConfig) (contextSize int, runtimeFlag
...
+		return fmt.Errorf("failed to pull model %s: %w", model, err)
+	}
+
+	slog.Info("Model pulled successfully", "model", model)
+	fmt.Printf("Model %s pulled successfully.\n", model)
+	return nil
+}
+
+func modelExists(ctx context.Context, model string) bool {
+	cmd := exec.CommandContext(ctx, "docker", "model", "inspect", model)
```
rumpl, line 486: There is no api for this?

dgageot: I didn't want to go that route... But I might have because it also needs to detect if the platform supports DMR at all.

dgageot:
@krissetto do you know what's the cleanest way to detect that DMR is not supported and therefore not trigger the auto-pull?

PR #607, opened 2025-10-24
feat: add environment variable placeholder expansion for agent commands (#604)

### LINKED ISSUE #604: commands support $ placeholders
Commands defined in the agent.yaml should support $placeholders that are replaced with env var values.

smazmi:
## Description

Implements support for `$placeholder` syntax in agent commands defined in `agent.yaml` files. When a command is invoked using the `-c/--command` flag, any `$VARIABLE_NAME` or `${VARIABLE_NAME}` placeholders are automatically replaced with their corresponding environment variable values.

Fixes #604

## Changes

- **Modified `cmd/root/run.go`**:
  - Added import for `pkg/environment` package
  - Added placeholder expansion logic after command lookup (lines 246-251)
  - Uses `environment.NewDefaultProvider()` to access env vars from multiple sources (OS env, secrets, keychain, etc.)
  - Returns clear error message if a required environment variable is not set

## Implementation Details

The expansion leverages existing infrastructure:
- `environment.Expand()` - Handles both `$VAR` and `${VAR}` syntax
- `environment.NewDefaultProvider()` - Provides access to multiple environment sources

Maintains **full backward compatibility** - commands without placeholders work exactly as before.

## Example Usage

**Before:**
```bash
$ cagent run agent.yaml -c greet
# Agent receives literal: "Say hello to $USER"
```

**After:**
```bash
$ export USER=alice
$ cagent run agent.yaml -c greet
# Agent receives expanded: "Say hello to alice"
```

**Agent configuration example:**
```yaml
agents:
  root:
    model: anthropic/claude-sonnet-4-0
    commands:
      greet: "Say hello to $USER and ask how their day is going"
      analyze: "Analyze the project named $PROJECT_NAME"
      path: "List contents of ${WORKING_DIR}"
```

## Testing
Tested with Docker Model Runner (DMR) using `ai/gpt-oss:latest`
### Manual Testing
Tested with multiple scenarios:
- [x] Single placeholder expansion (`$USER`)
- [x] Multiple placeholders in one command (`$PROJECT_NAME`, `$OWNER`, `$LOCATION`)
- [x] Both `$VAR` and `${VAR}` syntax
- [x] Commands without placeholders (backward compatibility)
- [x] Error handling when environment variable is missing

## Checklist

- [x] Code follows project style and patterns
- [x] Builds successfully (`task build`)
- [x] No lint errors (`task lint`)
- [x] Manually tested multiple scenarios
- [x] Backward compatibility maintained
- [x] Clear error messages for missing variables
- [x] Uses existing `environment` package infrastructure

Signed-off-by: Sadique Azmi [sadiquemobaraka5@gmail.com](mailto:sadiquemobaraka5@gmail.com)

### DIFF
--- a/pkg/teamloader/teamloader.go
+++ b/pkg/teamloader/teamloader.go
@@ -115,14 +115,12 @@ func createThinkTool(ctx context.Context, toolset latest.Toolset, parentDir stri
 // Undefined variables are allowed and will expand to empty strings, enabling users to
 // only define the environment variables needed for the commands they actually use.
 func expandCommandPlaceholders(ctx context.Context, commands map[string]string, envProvider environment.Provider) map[string]string {
-	if len(commands) == 0 {
-		return commands
-	}
+	expanded := map[string]string{}
 
-	expanded := make(map[string]string, len(commands))
 	for name, value := range commands {
 		expanded[name] = environment.ExpandLenient(ctx, value, envProvider)
 	}
+
 	return expanded
 }
 
@@ -350,9 +348,6 @@ func Load(ctx context.Context, p string, runtimeConfig config.RuntimeConfig, opt
 	agentsByName := make(map[string]*agent.Agent)
 
 	for name, agentConfig := range cfg.Agents {
-		// Expand environment variable placeholders in commands
-		expandedCommands := expandCommandPlaceholders(ctx, agentConfig.Commands, env)
-
 		opts := []agent.Opt{
 			agent.WithName(name),
 			agent.WithDescription(agentConfig.Description),
@@ -361,7 +356,7 @@ func Load(ctx context.Context, p string, runtimeConfig config.RuntimeConfig, opt
 			agent.WithAddPromptFiles(agentConfig.AddPromptFiles),
 			agent.WithMaxIterations(agentConfig.MaxIterations),
 			agent.WithNumHistoryItems(agentConfig.NumHistoryItems),
-			agent.WithCommands(expandedCommands),
+			agent.WithCommands(expandCommandPlaceholders(ctx, agentConfig.Commands, env)),
 		}
 
 		models, err := getModelsForAgent(ctx, cfg, &agentConfig, env, runtimeConfig)
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -114,41 +114,27 @@ agents:
     commands: # Either mapping or list of singleton maps
       df: "check how much free space i have on my disk"
       ls: "list the files in the current directory"
+      greet: "Say hello to $USER and ask how their day is going"
+      analyze: "Analyze the project named $PROJECT_NAME in the $ENVIRONMENT environment"
 ```
 
-### Running with named commands
-
-- Example YAML forms supported:
-
-```yaml
-commands:
-  df: "check how much free space i have on my disk"
-  ls: "list the files in the current directory"
-  greet: "Say hello to $USER and ask how their day is going"
-  analyze: "Analyze the project named $PROJECT_NAME in the $ENVIRONMENT environment"
-```
-
-```yaml
-commands:
-  - df: "check how much free space i have on my disk"
-  - ls: "list the files in the current directory"
-  - greet: "Say hello to $USER and ask how their day is going"
-```
-
-Commands support environment variable placeholders using `$VARIABLE_NAME` or `${VARIABLE_NAME}` syntax.
-
-Run:
+### Running named commands
 
 ```bash
 cagent run ./agent.yaml /df
 cagent run ./agent.yaml /ls
+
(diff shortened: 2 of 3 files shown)


smazmi:
I've tested this thoroughly with different placeholder scenarios and it all works well! I have some example YAML files I created during development - would you like me to include those in this PR at like `examples/` or keep it minimal? Happy either way! 😊

rumpl:
Thank you!

Adding examples would be great yes. And if you have the time, a couple of lines of documentation in docs/usage.md would be amazing.

cmd/root/run.go
```diff
@@ -243,7 +244,13 @@ func doRunCommand(ctx context.Context, args []string, exec bool) error {
 			return fmt.Errorf("agent '%s' has no commands", agentName)
 		}
 		if msg, ok := cmds[trimmed]; ok {
-			commandFirstMessage = &msg
+			// Expand $placeholders with environment variable values
+			env := environment.NewDefaultProvider()
```
rumpl: Commenting from my phone, I _think_ this code should be in the teamloader, that way all the other commands (exec, api…) can have the same behavior

smazmi:
> Thank you!
> 
> Adding examples would be great yes. And if you have the time, a couple of lines of documentation in docs/usage.md would be amazing.

Absolutely! I’ll get it done

cmd/root/run.go
smazmi: Ah! That makes sense, I'll take a look.

smazmi: Quick question about the current behavior right now, undefined variables cause an error. This means if a user runs -c <some_command> without defining a variable used in that command, it fails. However, if I move the expansion logic to the teamloader, it attempts to expand all variables during agent loading and fails if any variable is undefined.

I’m considering introducing a more lenient expand function that still allows the agent to load even if some variables are undefined in that case, those variables would just be replaced with empty strings instead of causing an error. Does that sound right?

rumpl: I think it's fine to not fail but replace with empty, after all `echo $SOME_VAR_THAT_DOESNT_EXIST` doesn't fail but just returns an empty string.

In the future we might add the same interpolation logic as [compose has](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/) but we can do that in a followup :)

smazmi: > after all `echo $SOME_VAR_THAT_DOESNT_EXIST` doesn't fail but just returns an empty string.

Yeah, that’s what I was thinking too, makes perfect sense. I’ll get it done then, glad I asked! Thanks :)

pkg/teamloader/teamloader_test.go
```diff
@@ -106,3 +107,75 @@ func TestOverrideModel(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create a test environment provider
+			env := &testEnvProvider{vars: tt.envVars}
+
+			// Expand the commands
+			result := expandCommandPlaceholders(t.Context(), tt.commands, env)
+
+			if tt.expectError {
```
dgageot: Please use `require.NoError` here

pkg/teamloader/teamloader.go
```diff
@@ -352,6 +367,9 @@ func Load(ctx context.Context, p string, runtimeConfig config.RuntimeConfig, opt
 	agentsByName := make(map[string]*agent.Agent)
 
 	for name, agentConfig := range cfg.Agents {
+		// Expand environment variable placeholders in commands
```
dgageot: Could the expansion happen just when the command is about to run?
If that's too complicated, that's ok. I just want things to be as lazy as possible.

smazmi: Hey @dgageot! I appreciate the feedback about lazy evaluation.

Quick question: if we do lazy expansion, doesn't that essentially mean we're back to my original implementation - expanding in `run.go` when the command is used?

The only difference would be calling a teamloader helper function instead of `environment.Expand()` directly. @rumpl suggested keeping expansion logic in teamloader for architectural consistency, but for lazy expansion, the actual call would happen in `run.go`.

So is this the right direction - keep the expansion utility in teamloader but call it lazily from run.go when the command is actually used?

dgageot: I'll sync with @rumpl and will come back to you!

smazmi: Sure

rumpl: Small update, we didn't have the time to sync today sorry, we'll get to it as soon as possible

smazmi: Sure, No problem

smazmi:
Thanks!

PR #612, opened 2025-10-25
feat: add agent commands support in TUI

### LINKED ISSUE #603: commands should also be /commands
An agent.yaml can contain commands. Those commands should be available as /commands in the TUI and also be shown in the command palette.

Deepam02:
Fixes #603

Agent commands can be invoked by typing /command in chat 
Commands now appear in the command palette (Ctrl+P) under "Agent Commands" category

### DIFF
--- a/pkg/tui/page/chat/chat.go
+++ b/pkg/tui/page/chat/chat.go
@@ -96,7 +96,7 @@ func New(a *app.App) Page {
 		title:        a.Title(),
 		sidebar:      sidebar.New(),
 		messages:     messages.New(a),
-		editor:       editor.New(),
+		editor:       editor.New(a.ResolveCommand),
 		focusedPanel: PanelEditor,
 		app:          a,
 		keyMap:       defaultKeyMap(),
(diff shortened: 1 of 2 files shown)


pkg/tui/tui.go
```diff
@@ -293,7 +293,7 @@ func toFullscreenView(content string) tea.View {
 
 // buildCommandCategories builds the list of command categories for the command palette
 func (a *appModel) buildCommandCategories() []dialog.CommandCategory {
-	return []dialog.CommandCategory{
+	categories := []dialog.CommandCategory{
 		{
```
Deepam02, line 297: Should I break this method into smaller parts or keep it as is since the logic is straightforward?

rumpl: Should be fine for now. Am on my phone now, hard to review, will take a closer look tomorrow

Deepam02:
@rumpl have you had a chance to review this pr?

rumpl:
@Deepam02 when the prompt for the command is too long the lines wrap, could we trim the command prompt so that it takes only one line?
Screenshot 2025-10-28 at 21 24 12

pkg/tui/page/chat/chat.go
```diff
@@ -179,7 +180,9 @@ func (p *chatPage) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return p, cmd
 
 	case editor.SendMsg:
-		cmd := p.processMessage(msg.Content)
+		// Resolve agent commands (if any)
+		content := p.app.ResolveCommand(msg.Content)
```
rumpl: Could the editor do the resolve? It kinda feels like it would be its job and not here, so basically the editor sends a message, and we don't care if the user typed it or if it was extracted from a command

Deepam02: Done! I moved the command resolution into the editor component. Now the editor resolves commands before sending the message.

PR #627, opened 2025-10-29
Add the ability to toon-ify the tool results

rumpl:
Useful for mcp servers that return big json outputs.

More info here https://github.com/johannschopplich/toon

### DIFF
--- a/pkg/tui/components/tool/tool.go
+++ b/pkg/tui/components/tool/tool.go
@@ -3,7 +3,6 @@ package tool
 import (
 	"encoding/json"
 	"fmt"
-	"log/slog"
 	"strings"
 
 	"github.com/charmbracelet/bubbles/v2/spinner"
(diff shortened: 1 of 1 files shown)


examples/github-toon.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+agents:
+  root:
+    model: anthropic/claude-sonnet-4-0
+    description: GitHub Agent - Help with GitHub using MCP tools
+    instruction: You are a helpful assistant that can help with all things GitHub
+    toolsets:
+      - type: mcp
+        ref: docker:github-official
+        toon: ".*" # toonify all the tools
```
krissetto, line 12: any specific reason for giving the choice of tools to toonify here? i was initially thinking it'd be simpler to do all or nothing, so i'm curious to know if you encountered any weird behavior etc along the way

krissetto:
a wee bit of docs would be nice :)

dgageot:
@rumpl What if we enable this for every mcp server?

krissetto:
> What if we enable this for every mcp server?

this is an option I'd like as well, potentially even globally for all agents in a team or as a configurable default (especially if we had a per-user option config file in .cagent). Setting this toon option under every toolset on every agent can be quite tedious

examples/github-toon.yaml
rumpl: Not all tools will return a json

rumpl:
Sure, I can add on an agent level, something like “toon: true”

But I do want to keep the option on each toolset

dgageot:
Let's merge it and iterate from there

PR #628, opened 2025-10-29
Cache markdown rendering

krissetto:
This drastically reduces the amount of times we re-render markdown if the terminal width or msg content hasn't changed, lowering the cpu load on the UI thread. Particularly noticeable on long threads where the conversation becomes much easier to scroll with these changes

### DIFF
--- a/pkg/tui/components/message/message.go
+++ b/pkg/tui/components/message/message.go
@@ -28,6 +28,10 @@ type messageModel struct {
 	height  int
 	focused bool
 	spinner spinner.Model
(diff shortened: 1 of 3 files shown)


dgageot:
@krissetto I think this has been removed by @rumpl. I let the two of you fight! :-)

rumpl:
Will take a look, messages component already has a cache so not sure what this does

krissetto:
closing for now since we cache thinking msgs in messages.go now

PR #633, opened 2025-10-29
Add xai api

mfranzon:
Add compatibility with xAI provider in order to use Grok models.

### DIFF
--- a/examples/grok.yaml
+++ b/examples/grok.yaml
@@ -2,7 +2,7 @@ version: "2"
 
 agents:
   root:
-    model: grok/grok-3-mini
+    model: xai/grok-3-mini
     description: A helpful AI assistant powered by Grok
     instruction: |
       You are Grok, a helpful and maximally truthful AI built by xAI.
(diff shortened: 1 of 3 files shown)


rumpl:
Woop woop, could you add it to the list of providers in the readme?

mfranzon:
> Woop woop, could you add it to the list of providers in the readme?

Absolutely!

dgageot (approved):

examples/grok.yaml
```diff
@@ -0,0 +1,17 @@
...
+    model: grok/grok-3-mini
+    description: A helpful AI assistant powered by Grok
+    instruction: |
+      You are Grok, a helpful and maximally truthful AI built by xAI.
+      Be helpful, accurate, and concise in your responses.
+      You are not based on any other companies and their models.
+
+models:
+  grok-model:
+    provider: grok
```
rumpl: Right, so we use models.dev for a bunch of things and it would seem that the provider name should be `xai`?

mfranzon: you are right. I changed It on the `provider.go` file. Thank you!

rumpl:
And it's in, thanks!

PR #635, opened 2025-10-29
Auto create directories in write_file

### LINKED ISSUE #621: `write_file` builtin should create the directory structure
If the LLM asks us to write to a file in a subdirectory that doesn't exist, we should create it and write the file, instead of returning an error like we do now

Deepam02:
closes #621

The `write_file` builtin tool now automatically creates parent directory structures when writing files to nested paths that don't exist.


rumpl (approved):
LGTM

PR #643, opened 2025-10-30
Use javascript to expand /commands

dgageot:


krissetto:
curiosity, why javascript templates for this implementation instead of go templates? they are nicer? :)

dgageot:
@krissetto nicer and more widespread

PR #654, opened 2025-10-30
Simpler code for history

dgageot:


pkg/tui/components/editor/editor.go
```diff
@@ -87,29 +74,20 @@ func (e *editor) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 			}
 			return e, nil
 		case "ctrl+c":
 			return e, tea.Quit
 		case "up":
-			// Consume the key when we replace the buffer with an older command.
-			if e.navigateHistory(NAVIGATEPREVIOUS) {
-				return e, nil
-			}
+			e.textarea.SetValue(e.hist.Previous())
```
rumpl, line 120: The current implementation has "moveHistoryCursorToLatest" to make sure we always start from the end of the history, this is not the case any more right? We need a way to rewind the history internal state

dgageot: I'll take a look

PR #677, opened 2025-11-01
Fix text selection styling

### LINKED ISSUE #665: Text selection styling
Mostly a nit, but the styling for the text when it is selected with a mouse changes the color of the underlying text on the lines that are selected. 

The text before the selection, on the same line, has one color. the text after the selection, also on the same line as the line were the selection ends, has a different color compared to the text before the selection

**Screenshot**

Image

Deepam02:
Fixes #665

Replaced string manipulation logic to use `ansi.Cut()` which preserves ANSI escape codes (colors, formatting) in unselected text segments while applying selection styling only to the selected portion.

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -950,25 +950,32 @@ func (m *model) applySelectionHighlight(lines []string, viewportStartLine int) [
 }
 
 func (m *model) highlightLine(line string, startCol, endCol int) string {
+	// Get plain text for boundary checks
 	plainLine := ansi.Strip(line)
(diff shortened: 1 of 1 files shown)


Deepam02:
Screenshot of the fix 
image

rumpl:
Would it be possible to also keep the styling on the selected text? Not sure how good that will look though. 

I’m curious to see how code selection looks like too

Deepam02:
I experimented with keeping the original styling visible when text is selected, but it didn’t work out cleanly. The problem is that when you layer a background color over existing ANSI color codes, it either wipes out the foreground colors or fails to apply properly.

PR #678, opened 2025-11-01
Nebius provider

shivaylamba:
Add compatibility with Nebius provider in order to use various OSS models.

### DIFF
--- a/examples/nebius.yaml
+++ b/examples/nebius.yaml
@@ -2,7 +2,7 @@ version: "2"
 
 agents:
   root:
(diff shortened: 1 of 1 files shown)


examples/nebius.yaml
```diff
@@ -0,0 +1,17 @@
+version: "2"
+
+agents:
+  root:
+    model: moonshotai/Kimi-K2-Instruct
```
dgageot:
```suggestion
    model: grok-model
```

PR #680, opened 2025-11-01
Use the TUI for agent creation

rumpl:


rumpl:
Rebased, ready for review. There's a ton we can do still but it's ok for now

dgageot (approved):

PR #693, opened 2025-11-03
Support lauching the api command with an oci artifact

### LINKED ISSUE #689: `cagent api` should work with an oci artifact name
It would auto-pull the agent's yaml like `cagent run` does.

Ideally, it would also refresh the yaml regularly when running for a long time.

krissetto:
Enables running the api server for agent teams in an oci artifact (e.g. pushed with `cagent push agent.yaml namespace/repo`)

Example to run from an oci artifact and auto-pull new versions every 10 mins:

```sh
cagent api namespace/repo --pull-interval 10
```

Also adds --pull-interval to allow for auto-pulling and reloading of an agent team from an oci reference

Closes #689

### DIFF
--- a/cmd/root/api.go
+++ b/cmd/root/api.go
@@ -6,10 +6,14 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"time"
 
+	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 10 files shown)


krissetto:
~2/3 of the changes are tests so don't worry too much about reviewing :wink:

dgageot (approved):

PR #699, opened 2025-11-03
Add a cagent mcp command to expose root and subagents as mcp tools in an mcp server

jeanlaurent:
## Add MCP command to expose agents as MCP tools

  Adds `cagent mcp` command that starts an MCP server exposing agents as tools via stdio.

  ### What it does
  - Each agent in the config becomes an MCP tool
  - Tool description = agent description
  - Works with single and multi-agent configurations
  - Sub-agents can be called directly

  ### Example

  ```bash
  # Start MCP server
  cagent mcp ./examples/dev-team.yaml
```
  This exposes 3 tools: root, designer, awesome_engineer

  MCP clients can call them:
```
  {
    "method": "tools/call",
    "params": {
      "name": "designer",
      "arguments": {
        "message": "Design a login page"
      }
    }
  }
  ```


cmd/root/mcp.go
```diff
@@ -0,0 +1,184 @@
...
+		sess.AddMessage(session.UserMessage(agentFilename, input.Message))
+
+		// Run the agent and collect the response
+		_, err = rt.Run(ctx, sess)
+		if err != nil {
+			slog.Error("Agent execution failed", "agent", agentName, "error", err)
+			return nil, ToolOutput{}, fmt.Errorf("agent execution failed: %w", err)
+		}
+
+		// Extract the assistant's response from session items
```
rumpl: The session object has a `GetLastAssistantMessageContent` method that does this

cmd/root/mcp.go
```diff
@@ -0,0 +1,184 @@
...
+	defer func() {
+		if err := t.StopToolSets(ctx); err != nil {
+			slog.Error("Failed to stop tool sets", "error", err)
+		}
+	}()
+
+	// Create MCP server
+	server := mcp.NewServer(&mcp.Implementation{
+		Name:    "cagent",
+		Version: "1.0.0",
```
rumpl:
```suggestion
		Version: version.Version,
```

cmd/root/mcp.go
```diff
@@ -0,0 +1,184 @@
...
+		toolDef := &mcp.Tool{
+			Name:        agentName,
+			Description: description,
+			InputSchema: map[string]any{
+				"type": "object",
+				"properties": map[string]any{
+					"message": map[string]any{
+						"type":        "string",
+						"description": "The message to send to the agent",
+					},
```
rumpl, line 104: This is fine for now but we might play around with the same tool as our `transfer_task` tool that has a `task` and an `expected_output` that helps guide the agent to do its job

cmd/root/mcp.go
jeanlaurent: good catch !

cmd/root/mcp.go
```diff
@@ -0,0 +1,184 @@
...
+		slog.Debug("MCP tool called", "agent", agentName, "message", input.Message)
+
+		// Get the agent to check its configuration
+		agent, err := t.Agent(agentName)
+		if err != nil {
+			return nil, ToolOutput{}, fmt.Errorf("failed to get agent: %w", err)
+		}
+
+		// Create session first with agent's max iterations (same as run command)
+		sess := session.New(session.WithMaxIterations(agent.MaxIterations()))
```
rumpl: Add a session title here so that the runtime doesn't try and generate one, it's not really needed for an MCP call. It's not a great API, I have it in the back of my mind to make this more explicit. 

```suggestion
		sess := session.New(session.WithTitle("MCP tool call"), session.WithMaxIterations(agent.MaxIterations()))
```

cmd/root/mcp.go
```diff
@@ -0,0 +1,184 @@
...
+		rt, err := runtime.New(t,
+			runtime.WithCurrentAgent(agentName),
+			runtime.WithRootSessionID(sess.ID),
+		)
+		if err != nil {
+			return nil, ToolOutput{}, fmt.Errorf("failed to create runtime: %w", err)
+		}
+
+		// Add the user message to the session
+		sess.AddMessage(session.UserMessage(agentFilename, input.Message))
```
rumpl: nit, you could add this when you create the session: `session.WithUserMessage(agentFilename, input.Message)`

rumpl (approved):
LGTM after minor change requests, let's get this one in so I can play with it!

cmd/root/mcp.go
jeanlaurent: ack, on it.

cmd/root/mcp.go
jeanlaurent: Will skip this for now.

PR #706, opened 2025-11-03
Add timeout to shell tool

### LINKED ISSUE #694: shell toolset should aways set a timeout
When calling a tool to run a shell command, we should always set a short time, like 30s.
We tend must add an optional parameter to the tool call so that the agent can raise the timeout.

This should reduce a lot the cases where the agent runs a never ending command.

Deepam02:
fixes  #694

### DIFF
--- a/pkg/tools/builtin/shell.go
+++ b/pkg/tools/builtin/shell.go
@@ -92,6 +92,7 @@ func (h *shellHandler) RunShell(ctx context.Context, toolCall tools.ToolCall) (*
 		if cmd.Process != nil {
 			_ = kill(cmd.Process, pg)
 		}
+		output := outBuf.String()
 		// Check if parent context was cancelled or if it was a timeout
 		if ctx.Err() != nil {
 			// Parent context was cancelled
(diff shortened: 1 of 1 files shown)


pkg/tools/builtin/shell.go
```diff
@@ -75,12 +88,20 @@ func (h *shellHandler) RunShell(ctx context.Context, toolCall tools.ToolCall) (*
...
+		if ctx.Err() != nil {
+			// Parent context was cancelled
+			return &tools.ToolCallResult{
+				Output: "Command cancelled",
+			}, nil
+		}
+		// Timeout occurred
 		return &tools.ToolCallResult{
-			Output: "Command cancelled",
+			Output: fmt.Sprintf("Command timed out after %v", effectiveTimeout),
```
rumpl: It would be great if we could also return the output of the shell command execution here, it might help the llm figure out what to do next.

Deepam02: Sure, I have added that.

rumpl (approved):

PR #722, opened 2025-11-04
The sidebar becomes a title bar when window too small

rumpl:
Screenshot 2025-11-04 at 17 48 07
Screenshot 2025-11-04 at 17 48 12


jeanlaurent:
thank you !

PR #724, opened 2025-11-04
Add tests for tool call formatting

dgageot:


rumpl:
FYI we have `gotest.tools/v3` in our dependencies now so you might want to maybe use golden?

dgageot:
I'll take a look.
I also need to add more tests but the code is currently random on maps.
I found out by adding assertions on e2e tests.

PR #725, opened 2025-11-04
Feature/welcome message

### LINKED ISSUE #718: Agents should have an optional Welcome Message
When running an agent, it would be nice that it shows a welcome message that givers basic instructions to the user.

We could start by:
+ Adding an optional static message to each agent in an agent.yaml
+ Showing this static message at the top of the TUI when `cagent run` starts.

In the future, the static text could contain placeholders interpolated by goja. e.g to show the name of the model it's going to use.

We could also imagine that the welcome message is totally auto-generated by the LLM from the information about the agent. Let's do that later and focus on the simpler case first.

Deepam02:
fixes #718

### DIFF
--- a/examples/welcome_message.yaml
+++ b/examples/welcome_message.yaml
@@ -1,0 +1,23 @@
+#!/usr/bin/env cagent run
+version: "2"
+
+agents:
+  root:
+    model: openai/gpt-4o-mini
+    description: A helpful AI assistant with a welcome message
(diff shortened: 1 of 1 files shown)


dgageot:
Hi @Deepam02, sorry, the code is going to be a bit more complicated because of recent changes in the codebase
 I'll probably start from your PR and add the changes that are needed.

PR #736, opened 2025-11-05
Don't just fmt.Println

dgageot:
+ Sometimes it makes senses to use a Printer, when the output is important for the end user.
+ Sometimes, it's better to use slog, especially when the code runs in server mode (cagent api)
+ Sometimes just drop the output


jeanlaurent:
Shouldn't we add this to the linter?

PR #748, opened 2025-11-07
api to tool call toolset

rumpl:


dgageot (approved):

dgageot:
Maybe we could add an output schema in the future. Very verbose but useful!

PR #760, opened 2025-11-11
Add a2a support

dgageot:
This has been tested with https://github.com/a2aproject/a2a-inspector


cmd/root/adk.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type adkFlags struct {
+	agentName  string
+	workingDir string
+	port       int
+	runConfig  config.RuntimeConfig
+}
+
+func newADKCmd() *cobra.Command {
```
rumpl: Should be called a2a, like the spec, not like the dev kit

pkg/a2a/testdata/basic.yaml
```diff
@@ -0,0 +1,9 @@
+version: "2"
```
rumpl, line 1: No need for this any more

dgageot: It's better to have those in test files

PR #795, opened 2025-11-14
File autocomplete ignore vcs

### LINKED ISSUE #786: File autocomplete ignore vcs
This PR https://github.com/docker/cagent/pull/723 added the possibility to ignore VCS, it would be really neat if the filesystem autocomplete also ignored things that are .gitignore'd

Deepam02:
closes #786

### DIFF
--- a/pkg/tools/builtin/filesystem.go
+++ b/pkg/tools/builtin/filesystem.go
@@ -13,9 +13,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing/format/gitignore"
-
 	"github.com/docker/cagent/pkg/fsx"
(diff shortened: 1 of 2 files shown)


pkg/tools/builtin/filesystem.go
rumpl, line 455: Do we still need this function here?

Deepam02: Yes, I think it is in the correct place. this function handles FilesystemTool specific logic like the ignoreVCS flag and the cached matcher setup, while the actual VCS matching is delegated to `fsx.VCSMatcher.ShouldIgnore()`
So keeping it here makes sense for me.

rumpl (approved):

pkg/tools/builtin/filesystem.go
rumpl: Got it, was confused, sorry

PR #822, opened 2025-11-17
Add background job support to shell

### LINKED ISSUE #811: Enhance shell toolset with background jobs
The shell toolset can run foreground commands. 

It should also be able to:
+ run a command in the background
+ list all background commands
+ view a background command's output
+ stop a background command

Important: background commands should all stop with the agent.

The typical use-case is for an agent to start a web a server and run other tools that connect to it.

Deepam02:
closes #811 
Adds background job management to the shell toolset.

**New Tools:**
- `run_shell_background` - Start long-running commands
- `list_background_jobs` - View all jobs and their status
- `view_background_job` - Check job output
- `stop_background_job` - Terminate running jobs

### DIFF
--- a/pkg/tools/builtin/shell.go
+++ b/pkg/tools/builtin/shell.go
@@ -320,11 +320,11 @@ func (h *shellHandler) ListBackgroundJobs(_ context.Context, _ tools.ToolCall) (
 		status := job.status.Load()
 		statusStr := statusToString(status)
 
-		runtime := time.Since(job.startTime).Round(time.Second)
+		elapsed := time.Since(job.startTime).Round(time.Second)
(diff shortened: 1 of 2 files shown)


pkg/tools/builtin/shell.go
```diff
@@ -9,12 +9,21 @@ import (
...
 	"time"
 
+	"github.com/docker/cagent/pkg/concurrent"
 	"github.com/docker/cagent/pkg/tools"
 )
 
-const ToolNameShell = "shell"
+const (
+	ToolNameShell              = "shell"
+	ToolNameRunShellBackground = "run_shell_background"
```
dgageot, line 22: Should it be `run_background_job`?

dgageot:
Let's merge it as is and iterate

PR #840, opened 2025-11-18
Fix/improve todo tool output formatting

### LINKED ISSUE #836: Don't print todo call results
1. Listing todos:
 Todos are explicitly listed with generic names.

```
Created 6 todos: [todo_1], [todo_2], [todo_3], [todo_4], [todo_5], [todo_6]
```

We should either skip that output, or since they are just listed before, we should not list their name at all.

2. completing todos
The output of completing a TODO is also refering to the key of the todo instead of it's name.
I like to know *which* TODO has been completed, here I kind of guess it's the 3rd one.

```
✓ Update TODO
✓ todo_3 → completed
Updated todo [todo_3] to status: [completed]
```

I think rather than outputing all the above we should write
```
✓ Update TODO
✓ The super task I was doing → completed
```

benodiwal:
Closes #836

- [x] Removed ID enumeration from createTodos output
- [x] Changed updateTodo to display descriptions instead of raw IDs
- [x] Converts todo_1 → Task1, UUIDs/other formats → Task in case of no description found  

Screenshot 2025-11-18 at 9 51 07 AM
Screenshot 2025-11-18 at 12 32 52 PM
Screenshot 2025-11-18 at 12 33 27 PM

### DIFF
--- a/pkg/tui/components/tool/todotool/integration_test.go
+++ b/pkg/tui/components/tool/todotool/integration_test.go
@@ -190,7 +190,7 @@ Some random text
 			}
 
 			// Verify output contains some basic formatting
-			assert.True(t, len(output) > 0, "Output should contain some content")
+			assert.NotEmpty(t, output, "Output should contain some content")
 		})
 	}
 }
--- a/pkg/tui/components/tool/todotool/parser_test.go
+++ b/pkg/tui/components/tool/todotool/parser_test.go
@@ -241,7 +241,7 @@ This is not a todo line
 		t.Run(tt.name, func(t *testing.T) {
 			result, err := parser.ParseTodoList(tt.input)
 			require.NoError(t, err)
-			assert.Equal(t, len(tt.expected), len(result))
+			assert.Len(t, result, len(tt.expected))
 
 			for i, expected := range tt.expected {
 				if i < len(result) {
@@ -285,7 +285,7 @@ func TestParseTodoListWithFallback(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			todos, err := parser.ParseTodoListWithFallback(tt.input)
 			require.NoError(t, err)
-			assert.Equal(t, tt.expectedCount, len(todos))
+			assert.Len(t, todos, tt.expectedCount)
 
 			// Check that expected descriptions are present
 			descriptions := make([]string, len(todos))
@@ -313,7 +313,7 @@ func TestRenderParsedTodo(t *testing.T) {
 	assert.Contains(t, result, "Test todo")
 
 	// Should contain some styled content (icon + description)
-	assert.True(t, len(result) > len("Test todo"))
+	assert.Greater(t, len(result), len("Test todo"))
 }
 
 func TestParsedTodo_ToTodoType(t *testing.T) {


benodiwal:
@rumpl  @jeanlaurent, there is a case when task description is not rendering i have written a fallback for that, can u check if its fine or we should take some other approach for this case, it is visible like this for that case

Screenshot 2025-11-18 at 12 51 09 PM

jeanlaurent:
First, Thanks for working on this @benodiwal
What prevents a task description from not rendering?

pkg/tools/builtin/todo.go
```diff
@@ -116,16 +116,8 @@ func (h *todoHandler) createTodos(_ context.Context, toolCall tools.ToolCall) (*
...
-	for i, id := range ids {
-		if i > 0 {
-			output += ", "
-		}
-		output += fmt.Sprintf("[%s]", id)
-	}
-
 	return &tools.ToolCallResult{
-		Output: output,
+		Output: fmt.Sprintf("Created %d todos", len(params.Descriptions)),
```
rumpl: We can't remove the current output from the `create_todos` tool, the LLM needs the IDs of the created todos so that it can update them at a later stage. If we don't show the IDs it will have to make an additional call to `list_todos` to get their IDs which just wastes tokens.

For this issue we need to only change the way we render the todo calls.

pkg/tui/components/tool/todotool/component.go
```diff
@@ -158,17 +159,32 @@ func (c *Component) renderList() string {
...
-					icon, style := renderTodoIcon("pending")
-					styledLines = append(styledLines, style.Render(icon)+" "+style.Render(strings.TrimSuffix(strings.TrimSpace(line[2:]), " (Status: pending)")))
-				case strings.Contains(line, "(Status: in-progress)"):
-					icon, style := renderTodoIcon("in-progress")
-					styledLines = append(styledLines, style.Render(icon)+" "+style.Render(strings.TrimSuffix(strings.TrimSpace(line[2:]), " (Status: in-progress)")))
-				case strings.Contains(line, "(Status: completed)"):
-					icon, style := renderTodoIcon("completed")
-					styledLines = append(styledLines, style.Render(icon)+" "+style.Render(strings.TrimSuffix(strings.TrimSpace(line[2:]), " (Status: completed)")))
-				default:
+				// Extract todo content, removing the ID portion
```
rumpl: This is very brittle, we should find a way to test it and make sure the TUI doesn't break if we change the output of the todo tool

pkg/tools/builtin/todo.go
benodiwal: Yeeah, ig that is the issue task description was not rendering which i mentioned above for the initial todo updates.

cc: @jeanlaurent

benodiwal:
@rumpl I have implemented a simple parser to make the logic less fragile with some unit tests for the parser and integration tests for the tui as well, can u check if there are some changes required or not

benodiwal:
hey @rumpl, gentle ping to review :)

dgageot:
@rumpl do we still need this?

rumpl:
We don't, sorry @benodiwal I worked on all the tool calls recently and made changes that are good enough for now, closing this

PR #843, opened 2025-11-18
RAG support 

### LINKED ISSUE #38: RAG
Support wiring a RAG to an agent

krissetto:
### RAG

I had some free time this past weekend so i started looking into the options we have around providing RAG support for cagent. This is one of our longest standing open issues #38, and it'd be awesome to support it in a comprehensive way.

The following PR is a proposal for how we could support giving agents access to knowledge bases via a variety of RAG approaches. The goal was to allow users to go from agent to agent + rag in a matter of minutes with minimal configuration required

### Screencast

https://github.com/user-attachments/assets/87cae5ea-6c14-41d1-99db-08518b2f6d2f

### Example config

Chunked embeddings RAG configuration (example files also included in the PR changes):

```yaml
agents:
  root:
    model: anthropic/haiku
    description: assistant that can search a knowledge base
    instruction: |
      You are a helpful assistant. You have access to a knowledge base you
      can use when it makes sense to do so, based on the user's question.
      If you receive sources from the knowledge base, always include them as
      a markdown list of links to local files at the very end of your response.
    rag: 
      - blork_knowledge_base

rag:
  blork_knowledge_base:
    description: documents about the blorks
    docs:
      - ./rag/docs
      - ./rag/blork_field_guide.txt
    strategies:
      - type: chunked-embeddings
        model: openai/text-embedding-3-small
        database: ./rag/chunked_embeddings.db
        threshold: 0.5
        vector_dimensions: 1536
        chunking:
          size: 1000
          overlap: 100
          respect_word_boundaries: true
```

The RAG space is quite open ended are there are lots of cool things we could support. I started with support for a handful of things as I was playing around, but this can be extended quite a lot more too.   

This PR currently includes:
- Background indexing of text-based documents;
- Automatic re-indexing of documents that have changed, been added, etc
- Support for chunked embeddings and bm25 retrieval strategies
- Support for using hybrid retrieval, which means gathering documents with more than one retrieval strategy and then "fusing" the results together
- When an agent is given access to one or more rag sources, it gets a new tool injected that it can use to query them

The changes consist mostly of new code that's isolated into a new `rag` package, examples and example fictitious docs to test the feature with. I tried to keep runtime changes minimal, only what was necessary to plug in rag support

The retrieval strategy concept is implemented in a way that the community can quite easily extend.

This PR is not perfect, but it is in a working state and could be merged as-is if we like the idea and want to iterate on it but, given the large amount of changes, I'm totally open to revising the approach.  

Let me know any thoughts, perplexities and if there are differing perspectives around how we should go about supporting RAG in the project, I'm looking forward to it :v:  

---

Closes #38


pkg/app/app.go
```diff
@@ -22,14 +22,25 @@ type App struct {
...
 		firstMessage:     firstMessage,
 		events:           make(chan tea.Msg, 128),
 		throttleDuration: 50 * time.Millisecond, // Throttle rapid events
 	}
+
+	// If the runtime supports background RAG initialization, start it
+	// and forward events to the TUI. Remote runtimes typically handle RAG server-side
+	// and won't implement this optional interface.
+	if ragRuntime, ok := rt.(runtime.RAGInitializer); ok {
+		go ragRuntime.StartBackgroundRAGInit(context.Background(), func(event runtime.Event) {
```
dgageot: This non-interruptible context is not great

pkg/team/team.go
```diff
@@ -74,6 +86,50 @@ func (t *Team) StopToolSets(ctx context.Context) error {
...
 }
+
+// RAGManagers returns the RAG managers for this team
+func (t *Team) RAGManagers() map[string]*rag.Manager {
+	return t.ragManagers
+}
+
+// InitializeRAG initializes all RAG managers in the background
+func (t *Team) InitializeRAG(ctx context.Context) {
+	if len(t.ragManagers) == 0 {
```
dgageot: not really needed

pkg/team/team.go
```diff
@@ -74,6 +86,50 @@ func (t *Team) StopToolSets(ctx context.Context) error {
...
+			} else {
+				slog.Info("RAG manager initialized successfully", "rag", mgr.Name())
+			}
+		}()
+	}
+}
+
+// StartRAGFileWatchers starts file watchers for all RAG managers
+func (t *Team) StartRAGFileWatchers(ctx context.Context) {
+	if len(t.ragManagers) == 0 {
```
dgageot: same

pkg/teamloader/teamloader.go
```diff
@@ -280,3 +303,26 @@ func getToolsForAgent(ctx context.Context, a *latest.AgentConfig, parentDir stri
...
+		return nil
+	}
+
+	// Filter managers for this agent
+	agentManagers := make(map[string]*rag.Manager)
+	for _, ragName := range agentConfig.RAG {
+		if mgr, exists := allManagers[ragName]; exists {
+			agentManagers[ragName] = mgr
+		} else {
+			slog.Warn("RAG source not found for agent", "rag", ragName)
```
dgageot: isn't that an error?

pkg/app/app.go
krissetto: good catch, missed this one, will fix

docs/USAGE.md
```diff
@@ -622,6 +622,239 @@ them to delegate tasks to other agents:
...
+
+Look for logs tagged with:
+- `[RAG Manager]` - Overall operations
+- `[Chunked-Embeddings Strategy]` - Chunked-embeddings retrieval
+- `[BM25 Strategy]` - BM25 retrieval
+- `[RRF Fusion]` / `[Weighted Fusion]` / `[Max Fusion]` - Result fusion
+
+### RAG Examples
+
+See `examples/` directory:
```
krissetto: note to self: update these

PR #868, opened 2025-11-19
Support CTRL-ENTER in the input window

jeanlaurent:
This add supporting CTRL-ENTER to add new line in the input window of the TUI.


dgageot:
@jeanlaurent Let's have @rockygeekz give it a go!

Issue #871, opened 2025-11-19
Improve implementation of shift-enter


dgageot:
Support for shift-enter needs to be improved. 

See https://github.com/docker/cagent/pull/875#issuecomment-3556706345

rockygeekz:
Hi @dgageot ! Can I take this issue and work on it? Thanks!

dgageot:
@rockygeekz Sure! I'll assign you.

Be careful, not every terminal support shift-enter...
I think that [crush](https://github.com/charmbracelet/crush) uses shift-enter when possible and ctrl-j otherwise  (e.g. in Apple's  terminal on macOS)

rockygeekz:
Thank you! I’ll make sure to check the terminal compatibility for Shift-Enter and Ctrl-J. I’ll get started on this now!

PR #875, opened 2025-11-19
fix: add proper Shift+Enter and Ctrl+J newline handling (#871)

rockygeekz:
Hey!
This PR adds the right behavior for Enter, Shift+Enter, and Ctrl+J as discussed in issue #871.

Enter now submits the message

Shift+Enter adds a newline (when the terminal actually sends that key)

Ctrl+J always adds a newline as a fallback on terminals that don’t support Shift+Enter

I tested this in a couple of environments:

Ubuntu terminal → Shift+Enter works and inserts a newline

WSL/Windows Terminal → Shift+Enter isn’t detected, but Ctrl+J works as expected

This follows the guidance you mentioned:
“use shift-enter when possible, and ctrl-j otherwise.”

If you want me to adjust anything or add more tests, I’m happy to update it!

### DIFF
--- a/pkg/tui/components/editor/editor.go
+++ b/pkg/tui/components/editor/editor.go
@@ -238,16 +238,55 @@ func (e *editor) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 		return e, nil
 	case tea.KeyPressMsg:
 		switch msg.String() {
-		case "enter":
+		// Handle send/newline keys:
+		// - Enter: submit current input (if textarea inserted a newline, submit previous buffer).
+		// - Shift+Enter: insert newline when supported by terminal.
+		// - Ctrl+J: fallback to insert '\n' when Shift+Enter isn’t reported.
+
+		case "enter", "shift+enter", "ctrl+j":
 			if !e.textarea.Focused() {
 				return e, nil
 			}
+
+			prev := e.textarea.Value()
+			e.textarea, _ = e.textarea.Update(msg)
+
 			value := e.textarea.Value()
+
+			// If the textarea inserted a newline due to plain Enter, submit the previous value.
+			if value != prev && msg.String() == "enter" {
+				if prev != "" && !e.working {
+					e.textarea.SetValue(prev)
+					e.textarea.MoveToEnd()
+					e.textarea.Reset()
+					e.refreshSuggestion()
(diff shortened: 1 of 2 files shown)


dgageot:
@rockygeekz could you tell the user which key stroke to use?

Screenshot 2025-11-20 at 08 30 54

rockygeekz:
@dgageot Behavior works as expected, and the UI now looks like this (see the image).
If you need any changes or adjustments, please feel free to let me know.

Screenshot 2025-11-20 134803

rumpl:
Just got time to look at this, we should do this differently, bubbletea has some things that can help us for this, it's weird and hidden but here's how it looks.

We should, in the tui.go (maybe?) listen to the event `tea.KeyboardEnhancementsMsg` and send it down to the editor, the editor can then check for the `msg.Flags`, if `mgs.Flags` is anything but `0` then we can set `shift+enter`, if not, we should set `ctrl+j` as the keys to add a new line.

We should set the keyboard shortcuts dynamically so that we always have only one shortcut for newlines at any time.

If you are on a mac you can test both cases easily: the default terminal doesn't support keyboard enhancements, while iterm does. Not sure how you would test this on a Windows machine...

dgageot:
@rumpl I'm going to merge this one and improve the implementation in another PR

dgageot (approved):

rockygeekz:
Hi @dgageot Thanks for merging! I wasn’t able to test the mac/iTerm side, but hope the contribution was helpful.

benodiwal:
> Just got time to look at this, we should do this differently, bubbletea has some things that can help us for this, it's weird and hidden but here's how it looks.
> 
> We should, in the tui.go (maybe?) listen to the event `tea.KeyboardEnhancementsMsg` and send it down to the editor, the editor can then check for the `msg.Flags`, if `mgs.Flags` is anything but `0` then we can set `shift+enter`, if not, we should set `ctrl+j` as the keys to add a new line.
> 
> We should set the keyboard shortcuts dynamically so that we always have only one shortcut for newlines at any time.
> 
> If you are on a mac you can test both cases easily: the default terminal doesn't support keyboard enhancements, while iterm does. Not sure how you would test this on a Windows machine...

Hey @rumpl, can u give a read to my thoughts here #930 once. Thanks !!

PR #878, opened 2025-11-20
Handle more complex gateway URLs

dgageot:
When a gateway URL contains query params, they need special handling otherwise clients will drop them.

### DIFF
--- a/pkg/model/provider/gemini/client.go
+++ b/pkg/model/provider/gemini/client.go
@@ -5,8 +5,10 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"log/slog"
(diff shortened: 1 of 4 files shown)


pkg/model/provider/gemini/client.go
```diff
@@ -79,16 +81,24 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 				return nil, errors.New("failed to get Docker Desktop token for Gateway")
 			}
 
+			// Pass a cleaned-up base URL to the OpenAI client
```
rumpl: Openai?

rumpl (dismissed):

rumpl (approved):

PR #900, opened 2025-11-21
runConfig is a pointer (*config.RuntimeConfig), this code was making all concurrent sessions share and modify the same runtime config

derekmisler:
Resolves runConfig is a pointer (*config.RuntimeConfig), this code was making all concurrent sessions share and modify the same runtime config


dgageot:
Closing in favour of https://github.com/docker/cagent/pull/902

PR #901, opened 2025-11-21
Support piping

dgageot:
Support  piping of cagent's output:

```
cagent exec ./examples/pirate.yaml "hello" | mate
cagent exec --yolo ./examples/shell.yaml "Is there a README.md?" | mate
echo "Is there a README.md?" | cagent exec --yolo ./examples/shell.yaml - | mate
```

### DIFF
--- a/pkg/cli/text.go
+++ b/pkg/cli/text.go
@@ -9,7 +9,8 @@ import (
 	"strings"
 
 	"github.com/fatih/color"
-	"github.com/wk8/go-ordered-map/v2"
+	"github.com/mattn/go-isatty"
+	orderedmap "github.com/wk8/go-ordered-map/v2"
 	"golang.org/x/term"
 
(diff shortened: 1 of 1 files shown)


pkg/cli/text.go
```diff
@@ -76,6 +77,10 @@ func (p *Printer) PrintToolCallWithConfirmation(ctx context.Context, toolCall to
 	p.PrintToolCall(toolCall)
 	p.Printf("\n%s", bold("Can I run this tool? ([y]es/[a]ll/[n]o): "))
 
+	if !isatty.IsTerminal(os.Stdout.Fd()) {
```
krissetto, line 80: does this need to consider the possible presence of the --yolo flag?

dgageot: No, in yolo mode, we don't enter this function.

PR #907, opened 2025-11-22
Tool rendering niceties

rumpl:
A minimal view for shell tool calls

Screenshot 2025-11-22 at 19 44 02

Later commits changed the way we render tools, they are now much more compact.

Screenshot 2025-11-22 at 23 53 08

Screenshot 2025-11-22 at 23 53 51

Screenshot 2025-11-22 at 23 54 31


rumpl:
Also shows a spinner while we are waiting for the arguments of a tool call to be streamed

Screenshot 2025-11-23 at 00 19 27

PR #930, opened 2025-11-25
feat: dynamic newline shortcuts based on terminal capabilities

### LINKED ISSUE #871: Improve implementation of shift-enter

benodiwal:
Closes #871

Tested on iTerm, Ghostty, and default Mac Terminal to verify the behavior across different terminal capabilities.

Hey @rumpl, I have some thoughts regarding the implementation after working on this issue. I've implemented keyboard enhancement detection as suggested, listening to `tea.KeyboardEnhancementsMsg`, checking `msg.Flags`, and dynamically configuring shortcuts. 

However, during implementation I discovered a terminal protocol limitation: even with keyboard enhancements enabled `(Flags != 0)`, terminals like `iTerm` send `shift+enter` as `ctrl+j` without preserving the `shift modifier` (debug shows: Key="ctrl+j", Mod=4 (ModCtrl), ShiftMod=false), making them indistinguishable at the protocol level. Given this reality, the implementation shows both `"shift+enter / ctrl+j"` in the status bar for modern terminals (since both work due to the protocol mapping) and only "ctrl+j" for legacy terminals. 

This approach matches what Claude Code, Gemini, and other modern terminal applications do, they acknowledge both shortcuts work in enhanced terminals rather than artificially restricting to one when the terminal protocol
 doesn't allow true separation.

### DIFF
--- a/pkg/tui/page/chat/chat.go
+++ b/pkg/tui/page/chat/chat.go
@@ -559,7 +559,6 @@ func (p *chatPage) cancelStream(showCancelMessage bool) tea.Cmd {
 	)
 }
 
-
 // processMessage processes a message with the runtime
 func (p *chatPage) processMessage(content string) tea.Cmd {
 	if p.msgCancel != nil {


dgageot:
cc @rumpl

rumpl (approved):

rumpl:
Tested, works, thanks <3

PR #934, opened 2025-11-25
Add mouse wheel scrolling support to TUI input editor

### LINKED ISSUE #931: Scrollable input area in TUI
The input area in the TUI is not scrollable and has a fixed height. 

If prompts become long, using the arrow keys to move the cursor up and down in the text is the only way to view.

It'd be nice to be able to scroll the input area with a mouse :)

Deepam02:
closes #931

### DIFF
--- a/pkg/tui/components/editor/editor.go
+++ b/pkg/tui/components/editor/editor.go
@@ -214,6 +214,30 @@ func (e *editor) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 	case tea.WindowSizeMsg:
 		e.textarea.SetWidth(msg.Width - 2)
 		return e, nil
+
+	// Handle mouse events
+	case tea.MouseWheelMsg:
+		// Forward mouse wheel as cursor movements to textarea for scrolling
+		// This bypasses history navigation and allows viewport scrolling
+		switch msg.Button.String() {
+		case "wheelup":
+			// Move cursor up (scrolls viewport if needed)
+			e.textarea.CursorUp()
+		case "wheeldown":
+			// Move cursor down (scrolls viewport if needed)
+			e.textarea.CursorDown()
+		}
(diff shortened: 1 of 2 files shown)


dgageot:
cc @krissetto

pkg/tui/components/editor/editor.go
```diff
@@ -214,6 +214,30 @@ func (e *editor) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
...
+	case tea.MouseWheelMsg:
+		// Forward mouse wheel as cursor movements to textarea for scrolling
+		// This bypasses history navigation and allows viewport scrolling
+		switch msg.Button.String() {
+		case "wheelup":
+			// Move cursor up (scrolls viewport if needed)
+			e.textarea.CursorUp()
+		case "wheeldown":
+			// Move cursor down (scrolls viewport if needed)
+			e.textarea.CursorDown()
```
krissetto, line 228: This does work, but I think it'd be a better UX to scroll the viewport directly rather than moving the cursor itself up. 

If we make the input area taller in height or responsive to terminal size, and say for example it becomes 10 lines tall,we'd need to scroll 11 lines up to scroll up by 12, then one more per line as we continue to go up. Same thing would happen when scrolling down. 

We could merge this as is and prepare a followup PR for what i mentioned, or just fix things directly here. Let me know what you prefer @Deepam02 :)

Deepam02: I tried that first, but the textarea component doesn't let us control the viewport directly

Deepam02: The  issue is that textarea with `SetHeight(3`) is limiting the scrollable content

krissetto: Then for now let's merge this as-is, and iterate :)

PR #966, opened 2025-11-27
More cleanup

dgageot:


pkg/server/server.go
```diff
@@ -98,7 +98,7 @@ func (s *Server) Serve(ctx context.Context, ln net.Listener) error {
 // API handlers
 
 func (s *Server) getAgents(c echo.Context) error {
-	agents := []api.Agent{}
+	var agents []api.Agent
```
rumpl: There was a comment but it was removed, do not change this one, idiomatic or not, I want this endpoint to always return a slice and never nil.

dgageot: Add a test man!

dgageot: I will :-)

rumpl (approved):

pkg/server/server.go
rumpl: The comment did great for _a long time_, I'm pretty sure you are capable of removing the test too :D

PR #967, opened 2025-11-27
Code aware chunking in RAG strategies

krissetto:
The goal of this change is to allow RAG strategy chunking configurations to define they are `code_aware`.

When `code_aware: true`, the rag system will use treesitter bindings to parse the AST of the code for the chunking, in order to not spit up logical blocks of code like functions.

Only Go code is supported in this initial implementation, but we can add support for more languages fairly easily in followup PRs.

Example:
```yaml
...
rag:
  codebase:
    docs: [./src]
    strategies:
      - type: chunked-embeddings
        embedding_model: openai/text-embedding-3-small
        database: ./code.db
        chunking:
          size: 2000
          **code_aware: true**    # <- Enable AST-based chunking
    results:
      limit: 5
```

The treesitter AST parsing will also be used in follow up rag strategies to gather the semantic meaning of the analyzed code before generating embeddings :smirk:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,9 @@ ARG ALPINE_VERSION="3.22"
 # xx is a helper for cross-compilation
 FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.7.0 AS xx
 
+# osxcross contains the MacOSX cross toolchain for xx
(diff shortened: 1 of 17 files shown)


krissetto:
needs some extra work to prepare for using CGO in CI

dgageot (approved):

PR #982, opened 2025-11-28
RAG: Allow `respect_vcs` config

krissetto:
Allow setting `respect_vcs` in RAG sources.
Can be overridden in strategy configurations as necessary. 
Defaults to true

Also moved doc gathering code to fsx package


docs/USAGE.md
```diff
@@ -974,6 +974,7 @@ models:
 |-------|------|-------------|
 | `docs` | []string | Document paths/directories (shared across strategies) |
 | `description` | string | Human-readable description |
+| `respect_gitignore` | boolean | Whether to respect .gitignore files (default: `true`) |
```
rumpl: Can we name this something else? I know, git is king but maybe something more general like VSC or something

krissetto: `respect_vsc`? other suggestions?

pkg/fsx/collect.go
```diff
@@ -0,0 +1,162 @@
...
+	"strings"
+
+	"github.com/bmatcuk/doublestar/v4"
+)
+
+// CollectFiles recursively collects all files from given paths.
+// Supports glob patterns (via doublestar), directories, and individual files.
+// Skips paths that don't exist instead of returning an error.
+// Optional shouldIgnore filter can exclude files/directories (return true to skip).
+func CollectFiles(paths []string, shouldIgnore func(path string) bool) ([]string, error) {
```
rumpl, line 16: Why didn't you use `fsx.DirectoryTree`?

krissetto: I wanted to consolidate in separate PR, the dir walking logic is a bit different between the two and i didn't want to break other things in the meantime

krissetto: I extracted and re-used some parts, but left the CollectFiles func for now as it has a few extra features

rumpl: Okay, thanks

rumpl (approved):

PR #1000, opened 2025-12-01
Allow to provide a eval file name

jeanlaurent:
This allows the `/eval` command to accept an optional file name.

`/eval` works as usual

`/eval foobar` will save the file as `foobar.json`

### DIFF
--- a/pkg/evaluation/save.go
+++ b/pkg/evaluation/save.go
@@ -5,22 +5,38 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/cagent/pkg/session"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/evaluation/save.go
```diff
@@ -5,22 +5,38 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/cagent/pkg/session"
 )
 
-func Save(sess *session.Session) (string, error) {
+func Save(sess *session.Session, filename ...string) (string, error) {
```
rumpl, line 13: Why is this `...string`?

dgageot:
Let's close it for now. I want it too but it should be implemented differently.

PR #1003, opened 2025-12-01
add hot-reload yaml configuration changes

### LINKED ISSUE #282: Feature Request: Hot-reload YAML configuration changes
Support for automatically reloading YAML configuration files when they change on disk, without requiring a restart of Cagent.

**Implementation:**
- Watch agent YAML files for changes using fsnotify
- When a file changes, validate the new configuration
- Apply the changes to running agents without losing state
- Fall back to previous config if validation fails

## **Why you'd like to see it**

**Problem:** Changing any agent configuration requires a full restart of Cagent, which:
- Loses all agent context and ongoing conversations  
- Interrupts service availability
- Makes iterative prompt engineering painful (edit → stop → start → test → repeat)

**Solution benefits:**
- Instant prompt/instruction updates while testing
- Zero-downtime config updates in production
- Faster development cycle when tuning agents

## **Workarounds?**

Currently using:
- Manual restart after each YAML edit (loses state)
- Running multiple Cagent instances and switching between them (complex)

Neither preserves agent state or provides seamless updates.

## **Additional context**

This is a common pattern in cloud-native tools:
- Kubernetes automatically updates mounted ConfigMaps
- Prometheus reloads config on SIGHUP
- Most modern servers support graceful config reload

**Simple MVP scope:**
- Only watch files specified at startup
- Only reload agent instructions/prompts (not model changes)
- Log changes for debugging
- Keep it simple - no fancy rollback mechanisms initially

Similar to how Docker Compose watches for file changes in development mode.

Deepam02:
Fixes #282

Implements automatic configuration reloading when agent YAML files change on disk.
### What's included:
- File watching with `fsnotify` and 500ms debounce for multiple rapid writes
- Validates new config before applying (falls back on error)
- Reloads `instruction`, `description`, and `commands` properties
- TUI notifications for reload events (started/success/failure)

### What's NOT included:
- reloading `Model` or `toolset` changes

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -81,9 +81,10 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 	}
 
 	var (
-		rt   runtime.Runtime
-		sess *session.Session
-		err  error
+		rt          runtime.Runtime
+		sess        *session.Session
+		err         error
+		agentSource config.Source
 	)
 	if f.remoteAddress != "" {
 		rt, sess, err = f.createRemoteRuntimeAndSession(ctx, agentFileName)
@@ -91,7 +92,7 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 			return err
 		}
 	} else {
-		agentSource, err := config.Resolve(agentFileName)
+		agentSource, err = config.Resolve(agentFileName)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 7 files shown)


Deepam02:
Are there any other agent fields which should be included in this PR?

Model and toolset reloading looks more complex, since it needs proper lifecycle handling and cleanup. Should that be done as a separate PR? I can take that up if it makes sense.

jeanlaurent:
That's cool ! This allow agents self updating themselves

dgageot:
@Deepam02 I'm going to close this one for now. We still want the feature but it probably requires more preparation work

PR #1004, opened 2025-12-01
Better file attachment support with `@somefile`

krissetto:
Makes @ references for files actually send the file contents along with the message to the LLM, so the agent doesn't have to round-trip to then get the files contents (_if it has access to a tool with which to do that_). 

Files references via @somefile get attached to a structured <attachments> section at the end of the user message, while the @somefile references remain in the message where the user typed them so the LLM has a better idea of the context and where the file contents are referenced by the user

Full attachment contents are not shown in the TUI to keep things easier to read

Also fixes a visual bug where historical completions were showing alongside @ file completions

#### Example

What the user sees:

image

What actually gets sent as the contents of the user message:

~~~
hey whats in here? @minimal.yaml

<attachments>
@minimal.yaml:
```
agents:
    root:
        model: openai/gpt-5-mini
        description: Writes a nice story about a subject
        instruction: You are helpful assistant
```
</attachments>
~~~

### DIFF
--- a/pkg/tui/page/chat/chat.go
+++ b/pkg/tui/page/chat/chat.go
@@ -259,8 +259,17 @@ func (p *chatPage) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 		return p, cmd
 
 	case editor.SendMsg:
-		cmd := p.processMessage(msg.Content)
-		return p, cmd
+		// Add user message to UI immediately using the display content (with @filename placeholders)
+		displayCmd := p.messages.AddUserMessage(msg.DisplayContent)
+		// Persist display content to history (not expanded file contents)
+		if p.history != nil {
+			if err := p.history.Add(msg.DisplayContent); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to persist command history: %v\n", err)
+			}
+		}
+		// Process the full content (with expanded files) for the agent
+		processCmd := p.processMessage(msg.Content)
+		return p, tea.Batch(displayCmd, processCmd)
 
 	case messages.StreamCancelledMsg:
 		model, cmd := p.messages.Update(msg)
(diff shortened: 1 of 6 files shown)


pkg/tui/page/chat/chat.go
```diff
@@ -259,8 +259,17 @@ func (p *chatPage) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 		return p, cmd
 
 	case editor.SendMsg:
-		cmd := p.processMessage(msg.Content)
-		return p, cmd
+		// Add user message to UI immediately using the display content (with @filename placeholders)
+		displayCmd := p.messages.AddUserMessage(msg.DisplayContent)
```
rumpl, line 263: Is this change needed for this feature? I guess it's fine to have optimistic rendering, my idea of the `UserMessageEvent` was for times when you connect the tui to a remote server

pkg/tui/page/chat/chat.go
```diff
@@ -294,8 +303,9 @@ func (p *chatPage) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 		p.sidebar = model.(sidebar.Model)
 		return p, cmd
 	case *runtime.UserMessageEvent:
-		cmd := p.messages.AddUserMessage(msg.Message)
-		return p, cmd
+		// User message is already added in the SendMsg handler with display content,
```
rumpl, line 306: We can remove this comment IMO

pkg/tui/page/chat/chat.go
krissetto: yeah good Q, honestly i didn't think too much about it.. maybe we should consolidate and have only one way to do things

rumpl: What do you mean only one way to do things? We do have one way to do things: react to the events sent from the runtime :D

PR #1005, opened 2025-12-02
Attachment improvements - Pasted text attachments

krissetto:
Includes a simple banner on top of the text area to show what files have been attached to the message. 
If a user clicks on an attachment in the banner, they can see a preview of the file.

If the user pastes more than 10 lines or 500 chars, create a temp file and attach it instead
This pasted content will then behave the same as normal attachments

Since this is already more than big enough, a subsequent PR will bring more improvements like showing attachments on the messages too

Screenshots:
Screenshot From 2025-12-09 15-35-38
Screenshot From 2025-12-09 15-35-43

### DIFF
--- a/pkg/tui/components/editor/editor.go
+++ b/pkg/tui/components/editor/editor.go
@@ -1,19 +1,24 @@
 package editor
 
 import (
+	"fmt"
 	"log/slog"
 	"os"
+	"path/filepath"
 	"regexp"
-	"slices"
 	"strings"
 
+	"charm.land/bubbles/v2/key"
 	"charm.land/bubbles/v2/textarea"
 	tea "charm.land/bubbletea/v2"
 	"charm.land/lipgloss/v2"
+	"github.com/atotto/clipboard"
+	"github.com/docker/go-units"
 	"github.com/mattn/go-runewidth"
 
 	"github.com/docker/cagent/pkg/app"
 	"github.com/docker/cagent/pkg/history"
+	"github.com/docker/cagent/pkg/paths"
 	"github.com/docker/cagent/pkg/tui/components/completion"
 	"github.com/docker/cagent/pkg/tui/components/editor/completions"
 	"github.com/docker/cagent/pkg/tui/core"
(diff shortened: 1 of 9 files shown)


krissetto:
moving to draft because i have to reconcile these changes with the changes made in https://github.com/docker/cagent/pull/1012 that i was unaware of

pkg/tui/components/editor/editor.go
```diff
@@ -514,61 +657,185 @@ func (e *editor) SetValue(content string) {
...
+	return attachment{
+		path:        file.Name(),
+		placeholder: "@" + displayName,
+		label:       fmt.Sprintf("%s (%s)", displayName, formatBytes(len(content))),
+		sizeBytes:   len(content),
+		isTemp:      true,
+	}, nil
+}
+
+func formatBytes(n int) string {
```
rumpl: we can use go-units

rumpl:
I pasted and clicked on the thing and the dialog seems completely broken. Do we really need this dialog and the mouse clicking?

Screenshot 2025-12-10 at 08 58 46

krissetto:
@rumpl Well, i do think the TUI should become more interactive, and i do think being able to click on the attachment to see the contents is quite nice, especially when multiple could be present. We could have a similar feature via keyboard but that feels quite clunky to use and is less intuitive than just point and click in my opinion.

I noticed this issue you faced was due to how some terminals handle horizontal scrolling and the related control codes, so i made some changes to enable word wrapping with no horizontal scrolling so it renders properly on most terminals (at least what i tested). 

I just tested the latest changes on:
- Terminal, Ghostty, VSCode's terminal and iTerm2 on MacOS;
- Ghostty and Ptyxis on Ubuntu;
- Windows Terminal + Powershell and Command Prompt on Windows

Could you give it another test and see if it works as expected?
Then if we decide we don't want the feature i can close the PR

rumpl:
> Then if we decide we don't want the feature i can close the PR

I do want the feature, but it's a lot of code and I can't really review it, the thing I'm having second thoughts is the dialog, that's all, I just pasted the text, I know what's in there.

krissetto:
I see your point.. Personally I've erroneously pasted the wrong things way more than once in the past, but maybe that's mostly a me problem? :thinking: Also if you paste multiple snippets the prompt you send ends up with things like "paste-1" and "paste-2", which may not be super clear especially if writing something that's not super short. 

My idea was to eventually also show these attachments on the messages themselves and then reuse the dialog logic to display them, so a user can see what they have attached in past messages as well.

There's no rush for merging this anyway. I can see to maybe split it up a bit, like removing the dialog for now? Some code can eventually be consolidated as well, for example the edit tool rendering dialog is similar, but there are some subtle differences and i didn't want to touch too many things at once

rumpl:
I just got scared by one 1k+ lines commit I think, I was actually driving this branch all morning, pasting away and it's a nice feature, I didn't click one though :)

rumpl (approved):

PR #1015, opened 2025-12-03
Fail fast if unix socket is already in use by another process

### LINKED ISSUE #1014: Only one server
It's possible to start two `cagent api` servers that listen on the same socket. The second `cagent api` should fail in the case of one already listening on that socket.

krissetto:
Make sure `cagent api` fails fast if the desired unix socket is in use by another process

image

Closes #1014

### DIFF
--- a/pkg/server/listen.go
+++ b/pkg/server/listen.go
@@ -2,10 +2,12 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 )
 
 func Listen(ctx context.Context, addr string) (net.Listener, error) {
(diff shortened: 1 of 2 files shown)


pkg/server/listen.go
```diff
@@ -21,8 +23,20 @@ func Listen(ctx context.Context, addr string) (net.Listener, error) {
 }
 
 func listenUnix(ctx context.Context, path string) (net.Listener, error) {
-	if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
-		return nil, err
+	// Check if socket file exists
+	if _, err := os.Stat(path); err == nil {
+		// Socket file exists - check if another process is using it
+		conn, err := net.DialTimeout("unix", path, 100*time.Millisecond)
```
rumpl, line 29: We should only check for the socket existence, not try and connect to it. This means we also need to make sure we cleanup the socket when `cagent api` exits

krissetto: fair, adjusting now

krissetto:
@rumpl feel free to undraft or close this once you know what you want to achieve here

dgageot:
Let's close this one for now. That's the standard behaviour of unix domain sockets

PR #1028, opened 2025-12-05
Faster tests

dgageot:
Used `cagent run ./examples/gopher.yaml` to optimize our slowest tests


rumpl (approved):
LGTM, we should take a closer look at some of those, maybe we can use `synctest` in some cases

PR #1036, opened 2025-12-05
Nicer demo

dgageot:


rumpl:
Are you using github in light mode?

pkg/teamloader/teamloader_test.go
```diff
@@ -1,173 +0,0 @@
-package teamloader
```
rumpl, line 1: What?

PR #1045, opened 2025-12-06
Add AWS Bedrock provider support

mfenderov:
Add AWS Bedrock provider using the Converse API. Supports all Bedrock models (Claude, Titan, Llama, Mistral) with AWS credential chain and Bedrock API key authentication.

```yaml
models:
  claude:
    provider: bedrock
    model: global.anthropic.claude-sonnet-4-5-20250929-v1:0
    provider_opts:
      region: us-east-1
```

This is an initial implementation to keep PR size manageable. Future additions: prompt caching, IAM policy examples, troubleshooting docs.

P.S.
There's an existing PR #542 that also adds Bedrock support, but there's been no movement in more than a month.

### DIFF
--- a/pkg/teamloader/teamloader_test.go
+++ b/pkg/teamloader/teamloader_test.go
@@ -14,6 +14,12 @@ import (
 	"github.com/docker/cagent/pkg/config/latest"
 )
 
+// skipExamples contains example files that require cloud-specific configurations
+// (e.g., AWS profiles, GCP credentials) that can't be mocked with dummy env vars.
+var skipExamples = map[string]string{
+	"pr-reviewer-bedrock.yaml": "requires AWS profile configuration",
+}
+
 func collectExamples(t *testing.T) []string {
 	t.Helper()
 
@@ -23,6 +29,10 @@ func collectExamples(t *testing.T) []string {
 			return err
 		}
 		if !d.IsDir() && filepath.Ext(path) == ".yaml" {
+			if reason, skip := skipExamples[filepath.Base(path)]; skip {
+				t.Logf("Skipping %s: %s", path, reason)
+				return nil
+			}
 			files = append(files, path)
 		}
 		return nil


inercia:
Seems related to https://github.com/docker/cagent/pull/542

krissetto:
Finally got around to testing things with bedrock setup properly, sorry for the delay. 

This provider seems to work, but i noticed that cost/usage information is not being handled properly, the TUI shows both tokens and cost at zero. The models.dev data we use for pricing should include plenty of bedrock models, so we can use that and consider it good enough. With usage tracked, I think it'd be enough to merge this for now

pkg/config/examples_test.go
```diff
@@ -57,7 +57,7 @@ func TestParseExamples(t *testing.T) {
 				require.NotEmpty(t, model.Provider)
 				require.NotEmpty(t, model.Model)
 				// Skip providers that don't have entries in models.dev
-				if model.Provider == "dmr" {
+				if model.Provider == "dmr" || model.Provider == "bedrock" {
```
krissetto: models.dev does have bedrock models under the `amazon-bedrock` provider id. maybe we can use that instead of just `bedrock` everywhere for simplicity

mfenderov: Already on it 😁

pkg/config/examples_test.go
```diff
@@ -57,7 +57,7 @@ func TestParseExamples(t *testing.T) {
 				require.NotEmpty(t, model.Provider)
 				require.NotEmpty(t, model.Model)
 				// Skip providers that don't have entries in models.dev
-				if model.Provider == "dmr" {
+				if model.Provider == "dmr" || model.Provider == "amazon-bedrock" {
```
krissetto: no need to skip bedrock here or else we won't get costs :)

krissetto: nit of course since its a test

pkg/model/provider/bedrock/client.go
```diff
@@ -0,0 +1,241 @@
...
+	var globalOptions options.ModelOptions
+	for _, opt := range opts {
+		opt(&globalOptions)
+	}
+
+	// Check for Bedrock API key (simpler auth method)
+	bearerToken, _ := env.Get(ctx, "AWS_BEARER_TOKEN_BEDROCK")
+	if bearerToken == "" {
+		bearerToken = getProviderOpt[string](cfg.ProviderOpts, "api_key")
+	}
```
krissetto: Last thing: we shouldn't allow the user to put their token directly in the agent.yaml. 

The goal is to keep the agent teams themselves shareable, and publishable on Hub, so they should never contain sensitive info/secrets etc.

What we do instead with other providers is use `token_key` as a way to allow the user to say "my token is in an env var with this name", so they can customize that way.

krissetto: e.g.

```yaml
models:
  thinku:
    provider: anthropic
    model: claude-haiku-4-5-20251001
    max_tokens: 64000
    token_key: CUSTOM_API_KEY_VAR # <--- this
    thinking_budget: 16384
    provider_opts:
      interleaved_thinking: true
...
```

krissetto:
After that last issue regarding the `api_key`, the rest looks good and I'm ready to merge :)

rumpl (approved):

PR #1046, opened 2025-12-07
Support agent paths with slashes in API routing

mcavage:
Replace two separate agent routes with a single wildcard route that can handle agent paths containing slashes. This allows agents to be organized in nested directories or namespaces.

The new routing logic:
- Uses a wildcard parameter to capture the full agent path
- Handles paths with or without leading slashes
- Supports optional agent_name as the final path segment
- Maintains backward compatibility with existing paths

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -70,8 +70,8 @@ func New(sessionStore session.Store, runConfig *config.RuntimeConfig, agentSourc
 	// Delete a session
 	group.DELETE("/sessions/:id", s.deleteSession)
 	// Run an agent loop
-	group.POST("/sessions/:id/agent/:agent", s.runAgent)
-	group.POST("/sessions/:id/agent/:agent/:agent_name", s.runAgent)
+	// Use wildcard to capture agent path which may contain slashes
+	group.POST("/sessions/:id/agent/*", s.runAgent)
 	group.POST("/sessions/:id/elicitation", s.elicitation)
 
 	// Health check endpoint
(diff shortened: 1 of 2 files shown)


BobDickinson:
FWIW, I'm building a TypeScript client to talk to the cagent API and I ran into this exact issue (agent id from API contained full path, and when using that exact ID in subsequent call paths the calls failed because the router wasn't processing the path segments properly).  I applied this PR to a feature branch to test and it solved my problem.

BobDickinson:
## Fix Errors

The PR has linter errors (which would also fail build if not caught there) because it's based on code from before the `sessionManager` refactoring (commit `9f152767`). Here are the required fixes:

### `pkg/server/server.go`

1. **Add `strings` import** (line ~11):
```go
import (
	// ... existing imports ...
	"strings"  // ADD THIS
	"time"
)
```

2. **Replace `s.agentSources` with `s.sm.sources`** (3 occurrences):
   - Line ~225: `s.agentSources[path]` → `s.sm.sources[path]`
   - Line ~230: `s.agentSources[withSlash]` → `s.sm.sources[withSlash]`
   - Line ~235: `s.agentSources[withoutSlash]` → `s.sm.sources[withoutSlash]`

### `pkg/server/server_test.go`

3. **Update `New()` call** (line ~101):
```go
// Before:
srv, err := New(store, &runConfig, sources)

// After:
ctx := t.Context()  // ctx is already declared earlier in the function
srv, err := New(ctx, store, &runConfig, 0, sources)
```

These changes align the PR with the current codebase structure that uses `sessionManager` instead of direct `agentSources` field and will allow it to pass the CI checks.

I had to apply these fixes to my feature branch to test the PR.

dgageot:
@rumpl do we still need this?

rumpl:
We don't this was fixed in v1.18.1, the file source now indexes by file name

PR #1047, opened 2025-12-07
api: implement GET /api/agents/:id to fix remote-mode welcome message

### LINKED ISSUE #973: Welcome message doesn't show in remote mode

26tanishabanik:
This PR adds the missing GET /api/agents/:id, since the remote client already calls GetAgent and needs the exact config. Without it, remote mode can’t read welcome_message, hence the welcome never shows.

Fixes #973

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -36,10 +36,8 @@ func New(ctx context.Context, sessionStore session.Store, runConfig *config.Runt
 
 	group := e.Group("/api")
 
-	// List all available agents
-	group.GET("/agents", s.getAgents)
-	// Get a specific agent configuration by ID
-	group.GET("/agents/:id", s.getAgent)
+	// List all available agents or get a specific agent by ID
+	group.GET("/agents/*", s.getAgents)
 
 	// List all sessions
 	group.GET("/sessions", s.getSessions)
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -122,6 +125,34 @@ func (s *Server) getAgents(c echo.Context) error {
 	return c.JSON(http.StatusOK, agents)
 }
 
+func (s *Server) getAgent(c echo.Context) error {
+	id := c.Param("id")
+
+	src, ok := s.sm.sources[id]
+	if !ok {
```
rumpl: Let's remove a 404 if the id isn't found, we do have the `/agents` so the ID should be known

26tanishabanik: Sure

26tanishabanik: I ran the cagent with below command:
```bash
> ./bin/cagent api examples/welcome_message.yaml
Listening on [::]:8080
{"time":"2025-12-08T14:42:03.689901+05:30","id":"","remote_ip":"::1","host":"localhost:8080","method":"POST","uri":"/api/sessions","user_agent":"Go-http-client/1.1","status":200,"error":"","latency":841333,"latency_human":"841.333µs","bytes_in":228,"bytes_out":230}
Fetching agent with ID: examples/welcome_message.yaml
Agent found? false
{"time":"2025-12-08T14:42:03.691819+05:30","id":"","remote_ip":"::1","host":"localhost:8080","method":"GET","uri":"/api/agents/examples/welcome_message.yaml","user_agent":"Go-http-client/1.1","status":200,"error":"","latency":439000,"latency_human":"439µs","bytes_in":0,"bytes_out":736}
Fetching agent with ID: examples/welcome_message.yaml
Agent found? false
{"time":"2025-12-08T14:42:03.694068+05:30","id":"","remote_ip":"::1","host":"localhost:8080","method":"GET","uri":"/api/agents/examples/welcome_message.yaml","user_agent":"Go-http-client/1.1","status":200,"error":"","latency":216833,"latency_human":"216.833µs","bytes_in":0,"bytes_out":736}
```

And then sent a request like this:

```bash
./bin/cagent run --remote http://localhost:8080 examples/welcome_message.yaml
```

Without the first `!ok` block, the welcome message doesn't come, do you want me to just remove the second `!ok` block? @rumpl

rumpl: We would actually need something like this: https://github.com/docker/cagent/pull/1046

Basically it would seem that our urls are not defined as they should

26tanishabanik: updated the code

dgageot:
Let's close it. This has been fixed with https://github.com/docker/cagent/pull/1143

PR #1052, opened 2025-12-09
Choose which agent/agents to expose with MCP

dgageot:


rumpl:
You say `agent/agents` but I don't see where in the code one can define multiple agents

PR #1056, opened 2025-12-09
Add the session store to the runtime

rumpl:
The runtime will save intermediate messages it gets during a loop

### DIFF
--- a/e2e/cagent_exec_test.go
+++ b/e2e/cagent_exec_test.go
@@ -104,7 +104,7 @@ func cagentExec(t *testing.T, moreArgs ...string) string {
 
 	// Start a recording AI proxy to record and replay traffic.
 	svr, _ := startRecordingAIProxy(t)
-	args = append(args, "--models-gateway", svr.URL)
+	args = append(args, "--models-gateway", svr.URL, "--session-db", "/tmp/session.db")
 
 	// Run cagent exec
 	var stdout bytes.Buffer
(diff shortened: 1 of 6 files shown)


e2e/cagent_exec_test.go
```diff
@@ -104,7 +104,7 @@ func cagentExec(t *testing.T, moreArgs ...string) string {
 
 	// Start a recording AI proxy to record and replay traffic.
 	svr, _ := startRecordingAIProxy(t)
-	args = append(args, "--models-gateway", svr.URL)
+	args = append(args, "--models-gateway", svr.URL, "--session-db", "/tmp/session.db")
```
dgageot, line 107: What happens when tests run concurrently?

pkg/session/store.go
```diff
@@ -34,6 +34,67 @@ type Store interface {
 	UpdateSession(ctx context.Context, session *Session) error
 }
 
+type InMemorySessionStore struct {
+	sessions map[string]*Session
```
dgageot: Can this store be shared between multiple sessions? Should it be synchronised?

rumpl: Updated

e2e/cagent_exec_test.go
rumpl: The store should be able to handle concurrent requests so this is fine.

dgageot (approved):

PR #1057, opened 2025-12-09
Implement ctrl-G to open prompt in editor

dgageot:
Fix https://github.com/docker/cagent/issues/1040

### DIFF
--- a/pkg/tui/page/chat/editor.go
+++ b/pkg/tui/page/chat/editor.go
@@ -1,0 +1,92 @@
+package chat
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+
+	tea "charm.land/bubbletea/v2"
+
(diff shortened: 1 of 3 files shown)


pkg/tui/page/chat/editor.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		updatedContent, readErr := os.ReadFile(tmpPath)
+		os.Remove(tmpPath)
+
+		if readErr != nil {
+			return msgtypes.ExternalEditorResultMsg{Err: fmt.Errorf("failed to read edited file: %w", readErr)}
+		}
+
+		return msgtypes.ExternalEditorResultMsg{Content: string(updatedContent)}
```
rumpl: We wouldn't need the `handleExternalEditorResult` if we return `editor.SendMsg` here directly

PR #1073, opened 2025-12-12
Support Vertex API

dgageot:


pkg/config/testdata/env/google_model.yaml
```diff
@@ -9,3 +9,6 @@ models:
   gemini:
     provider: google
     model: gemini-2-0
+    provider_opts:
+      project: my-gcp-project
```
krissetto: huh is this a vertex-only concept?

pkg/model/provider/gemini/client.go
```diff
@@ -48,15 +48,40 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
...
+		var httpClient *http.Client
+		apiKey := ""
+		backend := genai.BackendVertexAI
+
+		project := providerOption(cfg, "project")
+		location := providerOption(cfg, "location")
+		if project == "" && location == "" {
+			apiKey := defaultsTo(env.Get(ctx, "GOOGLE_API_KEY"), env.Get(ctx, "GEMINI_API_KEY"))
+			if apiKey == "" {
+				return nil, errors.New("either GOOGLE_API_KEY or GEMINI_API_KEY environment variable is required")
```
krissetto: is this msg correct if in gather.go we set GOOGLE_API_KEY in the requiredEnv map?

krissetto (approved):

pkg/config/testdata/env/google_model.yaml
dgageot: yes

PR #1110, opened 2025-12-15
Add a "Meta" property to the tool results

rumpl:
This makes it possible for the tools to send additional data that is not meant to be sent to the LLM. We use this in the case of the todo toolset to always send the current state of all of the todos. With this meta we no longer have to parse the arguments on the TUI side.

### DIFF
--- a/pkg/tools/builtin/filesystem.go
+++ b/pkg/tools/builtin/filesystem.go
@@ -124,6 +124,17 @@ type ReadMultipleFilesArgs struct {
 	JSON  bool     `json:"json,omitempty" jsonschema:"Whether to return the result as JSON"`
 }
 
+type ReadMultipleFilesEntry struct {
+	Path      string `json:"path"`
+	Content   string `json:"content"`
+	LineCount int    `json:"lineCount"`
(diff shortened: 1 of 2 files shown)


pkg/runtime/event.go
```diff
@@ -82,18 +82,20 @@ func ToolCallConfirmation(toolCall tools.ToolCall, toolDefinition tools.Tool, ag
...
 type ToolCallResponseEvent struct {
-	Type           string         `json:"type"`
-	ToolCall       tools.ToolCall `json:"tool_call"`
-	ToolDefinition tools.Tool     `json:"tool_definition"`
-	Response       string         `json:"response"`
+	Type           string                `json:"type"`
+	ToolCall       tools.ToolCall        `json:"tool_call"`
+	ToolDefinition tools.Tool            `json:"tool_definition"`
+	Response       string                `json:"response"`
+	Result         *tools.ToolCallResult `json:"result,omitempty"`
```
rumpl, line 89: This one is weird... For now, at some point in the future we will be able to remove the `Response`, once no other parts depend on it

dgageot (approved):

Issue #1075, opened 2025-12-13
Handoffs example won't return control to root agent


BobDickinson:
**Describe the bug**

I have done several runs with the example agent: /examples/handoff.yaml and have been unable to coerce it to return control to the root agent.  When the web search subagent is directed to return control it fails to do so, both ignoring the message, and repeating the original search that launched the web search agent (even if that isn't the most recent search by that agent in the session).  The not returning control is an issue (I get it, agents are unpredictable), but the ignoring the prompt and repeating the first search feels like a bug.

**Version affected**

cagent version v1.9.29
Commit: d404b37bfea5e82855eed5a748bff3f791009c34

**How To Reproduce**

Run `cagent run handoff.yaml`

1. Give a prompt that requires a search
2. After search, note that search agent is still running
3. Try another search if you want
4. Prompt it to return control to the root agent

It ignores that prompt (usually), and repeats the initial search (see output below).
I've run this 5 times (with variations in the prompts), and it failed as above on all but one occasion, where it actually didn't repeat the initial search and told me it returned control (it did not).

In no case did I ever successfully coerce the sub-agent to return control to the root agent.

**Expectation**

I expected it to return control to the root agent after the search.  When reviewing the actual prompt the agent provided to the sub-agent, I could see how that might not happen, but I would at least expect it to happen when directly prompted (or very worst case, to treat the prompt as something to search).  But the fact that it ignores the prompt and repeats the first search (in most cases) makes me think this is a bug and not just an intransigent agent (or it's a bug in addition to an intransigent agent).

**Screenshots**

┃
┃  what is the current google doodle
┃

┃
┃ ✓ Handoff Conversation
┃ agent:
┃ web_search
┃
┃
┃ -> output:
┃ The agent root handed off the conversation to you, look at the history of the conversation and continue where
┃ it
┃  left off. Once you are done with your task or if the user asks you, handoff the conversation back to root.
┃

┃
┃ ✓ search
┃ query:
┃ current Google Doodle
┃
┃
┃ -> output (truncated):
┃ Found 10 search results:
┃
┃ 1. Google Doodles - Google's Search Logo Changes for Every Occasion
┃    URL: https://doodles.google/
┃    Summary: Discover the latest and pastDoodlesfrom around the world, created byGoogleto celebrate local
┃ heroes,
┃  events, culture, and places. Explore interactive and videoDoodles, find your birthdayDoodle, and learn fun
┃ fact
┃ s aboutDoodlehistory and artists.
┃
[...]

PR #1112, opened 2025-12-16
Improve handoff graph support

BobDickinson:
Made multiple improvements to address issues with handoff example, handoff functionality generally. Addresses issue #1075.

## Issue: Handoff example not working, not a good graph example

The previous `handoff.yaml` was an example that would have been better implemented as a sub_agent.
It also had no mechanism to return to the root agent (which admittedly could have been fixed by
just adding a `handoffs: root` to the `web_search` agent).

To distinguish the `handoffs` functionality and use case from `sub_agents` (and tool calling), we prefer an
example using a directed graph (and one where that is a better and more appropriate solution than
the usual "orchestrator root agent calling sub_agents and tools" used in most examples).

The new handoff.yaml implements a graph that looks like:

### Handoff Graph Diagram:

```
   ─────>   Root Agent  <────────<─<─<──────┐
                |                           |
        ┌───────┼────────┐                  |
        |                |                  |
        ↓                ↓                  |
   Web Search      Filesystem Search        |
        |                |                  |
        |        ┌───────┴───────┐          |
        |        |               |          ↑
        |    (if list)    (if contents)     ↑
        ↓        |               ↓          ↑
        ↓        ↓            Redactor      |
        ↓        |               ↓          |
        |        └───────┬───────┘          |
        |                ↓                  |
        |            Tabulator              |
        |                ↓                  |
        └────────┼───────┘                  |
                 ↓                          |
             Summarizer ──────>─>─>─────────┘
```

## Issue: Agents receiving handoff repeat initial prompt

There was an issue (#1075) where under some circumstances (including lack of handoff in agent config) the
agent that received a handoff would repeatedly ingore new prompts and respond to the orginal prompt
from the handoff.

It was determined that this behavior was caused by the handoff tool result message, specifically, 
the phrase "look at the history of the conversation and continue where it left off", which has been
removed (see final message below).

## Issue: Invalid tool calling and handoff

Unlike when using sub-agents or using agents via tool call, when we do a handoff the new agent has
access to the full conversation history. And it doesn't really distinguish or understand that previous
messages were directed to a different agent. I've seen cases where an agent will call tools that it does
not have available (because it can see the tool descriptions in previous LLM calls for a different
agent that did have those tools) and I've seen it handoff to an agent that is not in its handoff list
(because it saw that agent as available for other agents to handoff to in the chat history).  I've 
seen this in combination (an agent with no handoffs successfully handed off to another agent).

The tool calling and handoff code did not validate that a called tool was available to the agent, or 
in the case of handoff, that the handoff agent was availalble to the agent.

This PR adds validation of tool calls and agent handoffs (which is probably good security hygeine
anyway - an agent being able to call tools not defined for it, or handoff to agents not specified
in its handoffs, represents a security exposure that might be able to be triggered through malicious
prompts or other mechanisms in addition to the case described above).

In addition, the result of the handoff tool call has been updated to attempt to dissuade the current
agent from finding tools or agents in the chat history (see final message below).

## Issue: Handoff tool result indicated return of control on all transfers

The handoff tool result indicated that the agent receiving the handoff should return control to the
agent giving the handoff in all cases (including when a worker agent returned control to `root`, where
it implied that the root should complete its task and return control to the worker agent, which is
incorrect).  While this did not manifest in any observed flow of contol issues, it is still incorrect.

Removed the "handoff the conversation back to %s" part of the handoff tool result (see final message
below).  

## Issue: Handoff system prompt had contradictory/unclear condition 

Under the "When to handoff" in the handoff system prompt the following 
statement was contradictory/unclear and was removed:

```
- If any of the tools of the agent indicate that this agent is able to respond correctly
```

## Issue: Difficult to read handoff prompts, assess diffs

Both the handoff system prompt (now called handoffPrompt) and the handoff tool result (now called 
handoffMessage) have been reformatted into multi-line strings for better readability and maintainability.

## Final updated handoff tool result message
Several changes were made related to issues as described above.

Was:

The agent %s handed off the conversation to you, look at the history of the conversation and continue
where it left off. Once you are done with your task or if the user asks you, handoff the conversation
back to %s.

Now:

The agent %s handed off the conversation to you. Your available handoff agents and tools are specified
in the system messages that follow. Only use those capabilities - do not attempt to use tools or hand off
to agents that you see in the conversation history from previous agents, as those were available to 
different agents with different capabilities. Look at the conversation history for context, but only
use the handoff agents and tools that are listed in your system messages below. Complete your part of
the task and hand off to the next appropriate agent in your workflow (if any are available to you), or
respond directly to the user if you are the final agent.

## Summary

The new handoff example generally works as expected.  I have seen it periodically fail to handoff from the agents using the `openai/gpt-4o-mini` model, which likely struggles with multi-step instructions.  In a perfect world we wouldn't need to rely on the model for the handoff.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -1226,7 +1226,10 @@ func (r *LocalRuntime) runAgentTool(ctx context.Context, handler ToolHandlerFunc
 func (r *LocalRuntime) addToolValidationErrorResponse(ctx context.Context, sess *session.Session, toolCall tools.ToolCall, tool tools.Tool, events chan Event, a *agent.Agent) {
 	errorMsg := fmt.Sprintf("Tool '%s' is not available to this agent (%s).", toolCall.Function.Name, a.Name())
 
-	events <- ToolCallResponse(toolCall, tool, errorMsg, a.Name())
+	events <- ToolCallResponse(toolCall, tool, &tools.ToolCallResult{
+		Output:  errorMsg,
+		IsError: true,
+	}, errorMsg, a.Name())
 
 	toolResponseMsg := chat.Message{
 		Role:       chat.MessageRoleTool,


dgageot:
@rumpl wdyt?

pkg/runtime/runtime.go
```diff
@@ -1354,13 +1369,39 @@ func (r *LocalRuntime) handleHandoff(_ context.Context, _ *session.Session, tool
...
+	handoffs := currentAgent.Handoffs()
+	targetInHandoffs := false
+	for _, handoffAgent := range handoffs {
+		if handoffAgent.Name() == params.Agent {
+			targetInHandoffs = true
+			break
+		}
+	}
+	if !targetInHandoffs {
+		return nil, fmt.Errorf("agent %s cannot hand off to %s: target agent not in handoffs list", ca, params.Agent)
```
rumpl: Instead of returning an error here, we should send a result to the LLM telling it that this is not OK so that it has the opportunity to correct itself

pkg/session/session.go
```diff
@@ -301,9 +301,20 @@ func (s *Session) GetMessages(a *agent.Agent) []chat.Message {
 			validAgentIDs = append(validAgentIDs, agent.Name())
 		}
 
+		handoffPrompt := "You are part of a multi-agent team. Your goal is to answer the user query in the most helpful way possible.\n\n" +
+			"Available agents in your team:\n" + agentsInfo + "\n" +
+			"You can hand off the conversation to any of these agents at any time by using the `handoff` function with their ID. " +
+			"The valid agent names are: " + strings.Join(validAgentIDs, ", ") + ".\n\n" +
```
rumpl: We are talking about agent IDs above, let's keep it consistent
```suggestion
			"The valid agent IDs are: " + strings.Join(validAgentIDs, ", ") + ".\n\n" +
```

pkg/runtime/runtime.go
BobDickinson: Agreed.  I made this change.  Also cleaned up the tool call validation error to be more specific (it was using the tool rejected error before).

BobDickinson:
Thanks for the feedback. I agree and made the suggested changes.

PR #1115, opened 2025-12-16
fix: MultiProvider now distinguishes empty values from not found

### LINKED ISSUE #1103: [Detail Bug] MultiProvider ignores empty values from higher‑priority providers
# Detail Bug Report

https://app.detail.dev/org_42bf1b6f-11b3-499a-ad6d-d5a056862990/bugs/bug_bd067399-6be5-4c4e-b9c5-01e96c1f9d2c

# Summary
- **Context**: `MultiProvider` is used throughout the codebase to provide a cascading lookup of environment variables across multiple sources (env files, OS environment, Docker secrets, keychain, etc.) with a defined priority order.
- **Bug**: `MultiProvider.Get()` cannot distinguish between "variable not found" and "variable explicitly set to empty string", treating both cases as "not found" and cascading to the next provider.
- **Actual vs. expected**: When a higher-priority provider explicitly sets a variable to empty string, `MultiProvider` ignores it and returns a value from a lower-priority provider; it should return the empty string.
- **Impact**: Users cannot explicitly override/clear environment variables using higher-priority sources (like env files), as the system will incorrectly fall back to lower-priority sources, potentially exposing secrets or using unwanted default values.

# Code with bug
```go
// pkg/environment/multi.go
func (p *MultiProvider) Get(ctx context.Context, name string) string {
	for _, provider := range p.providers {
		value := provider.Get(ctx, name)
		if value != "" { // <-- BUG 🔴 Treats empty string as "not found"
			return value
		}
	}

	return ""
}
```

# Failing test
```go
package environment

import (
	"context"
	"testing"

	"github.com/stretchr/testify/assert"
)

// TestMultiProviderEmptyValue tests that an empty string value from a higher
// priority provider should be returned, not skipped in favor of a non-empty
// value from a lower priority provider.
func TestMultiProviderEmptyValue(t *testing.T) {
	// Setup: Create two providers
	// - First provider returns empty string (e.g., env var explicitly set to "")
	// - Second provider returns "fallback" (e.g., a default value)

	firstProvider := NewEnvListProvider([]string{"MY_VAR="}) // Empty value
	secondProvider := NewEnvListProvider([]string{"MY_VAR=fallback"})

	// Create multi provider with first having higher priority
	provider := NewMultiProvider(firstProvider, secondProvider)

	// Act: Get the value
	value := provider.Get(context.Background(), "MY_VAR")

	// Assert: Should return empty string from first provider, not "fallback" from second
	// Because the first provider found the variable (even though it's empty)
	assert.Equal(t, "", value, "Should return empty string from higher priority provider, not skip to next provider")
}
```

Test output:
```
=== RUN   TestMultiProviderEmptyValue
    multi_empty_test.go:29:
        Error Trace:
        Error:          Not equal:
                        expected: ""
                        actual  : "fallback"

                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1 +1 @@
                        -
                        +fallback
        Test:        TestMultiProviderEmptyValue
        Messages:    Should return empty string from higher priority provider, not skip to next provider
--- FAIL: TestMultiProviderEmptyValue (0.00s)
FAIL
```

# Recommended fix
- Change the `Provider` interface to return `(string, bool)` to distinguish "found" from "not found", following the `os.LookupEnv` pattern, and update `MultiProvider` to honor the boolean:

```go
// pkg/environment/provider.go
type Provider interface {
	// Get retrieves the value of an environment variable by name.
	// Returns (value, true) if found (value may be empty)
	// Returns ("", false) if not found
	Get(ctx context.Context, name string) (string, bool)
}

// pkg/environment/multi.go
func (p *MultiProvider) Get(ctx context.Context, name string) (string, bool) {
	for _, provider := range p.providers {
		value, found := provider.Get(ctx, name)
		if found {
			return value, true
		}
	}
	return "", false
}
```

AryanBagade:
## Problem
`MultiProvider.Get()` cannot distinguish between "variable not found" and "variable explicitly set to empty string", treating both cases as "not found" and incorrectly cascading to the next provider.
### Before (Bug)
  ```go
  func (p *MultiProvider) Get(ctx context.Context, name string) string {
      for _, provider := range p.providers {
          value := provider.Get(ctx, name)
          if value != "" {  // ❌ Treats empty string as "not found"
              return value
          }
      }
      return ""
  }
```
  After (Fix)
```go
  func (p *MultiProvider) Get(ctx context.Context, name string) (string, bool) {
      for _, provider := range p.providers {
          value, found := provider.Get(ctx, name)
          if found {  // ✅ Checks if variable was found, not if it's empty
              return value, true
          }
      }
      return "", false
  }
 ```
 
 ## Why This Fix Is Needed
Users cannot explicitly override/clear environment variables using higher-priority sources (like env files). For example:
```
# .env file (higher priority)
MY_SECRET=           # User wants to clear/override this
# keychain (lower priority)
MY_SECRET=actual-secret-value
```
Before: Returns actual-secret-value (wrong - ignores user's intent to clear)
After: Returns "" (correct - respects user's explicit empty value)

## Changes
- Changed Provider interface to return (string, bool) following Go's os.LookupEnv pattern
- Updated MultiProvider.Get() to check found boolean instead of empty string
- Updated all 7 provider implementations (OsEnvProvider, EnvListProvider, EnvFilesProvider, KeychainProvider, RunSecretsProvider, DockerDesktopProvider, PassProvider)
- Updated all callers across the codebase

## Tests
- Added TestMultiProviderEmptyValue that verifies:
- firstProvider := NewEnvListProvider([]string{"MY_VAR="})      // Empty value
- secondProvider := NewEnvListProvider([]string{"MY_VAR=fallback"})
- provider := NewMultiProvider(firstProvider, secondProvider)

-value, found := provider.Get(context.Background(), "MY_VAR")
// ✅ Returns ("", true) - empty string from first provider
// ❌ Previously returned ("fallback", true) - incorrectly skipped to second

Fixes #1103

### DIFF
--- a/e2e/proxy_test.go
+++ b/e2e/proxy_test.go
@@ -48,6 +48,7 @@ func startRecordingAIProxy(t *testing.T) (*httptest.Server, *config.RuntimeConfi
 
 type testEnvProvider map[string]string
 
-func (p *testEnvProvider) Get(_ context.Context, name string) string {
-	return (*p)[name]
+func (p *testEnvProvider) Get(_ context.Context, name string) (string, bool) {
+	val, found := (*p)[name]
+	return val, found
 }
--- a/pkg/config/auto.go
+++ b/pkg/config/auto.go
@@ -16,25 +16,27 @@ var DefaultModels = map[string]string{
 }
 
 func AvailableProviders(ctx context.Context, modelsGateway string, env environment.Provider) []string {
+	if modelsGateway != "" {
+		// Default to anthropic when using a gateway
+		return []string{"anthropic"}
+	}
+
 	var providers []string
 
-	if modelsGateway == "" {
-		switch {
-		case env.Get(ctx, "ANTHROPIC_API_KEY") != "":
-			providers = append(providers, "anthropic")
-		case env.Get(ctx, "OPENAI_API_KEY") != "":
-			providers = append(providers, "openai")
-		case env.Get(ctx, "GOOGLE_API_KEY") != "":
-			providers = append(providers, "google")
-		case env.Get(ctx, "MISTRAL_API_KEY") != "":
-			providers = append(providers, "mistral")
-		default:
-			providers = append(providers, "dmr")
-		}
-	} else {
-		// Default to anthropic when using a gateway
+	if key, _ := env.Get(ctx, "ANTHROPIC_API_KEY"); key != "" {
 		providers = append(providers, "anthropic")
 	}
+	if key, _ := env.Get(ctx, "OPENAI_API_KEY"); key != "" {
+		providers = append(providers, "openai")
+	}
+	if key, _ := env.Get(ctx, "GOOGLE_API_KEY"); key != "" {
+		providers = append(providers, "google")
+	}
+	if key, _ := env.Get(ctx, "MISTRAL_API_KEY"); key != "" {
+		providers = append(providers, "mistral")
+	}
+
+	providers = append(providers, "dmr")
 
 	return providers
 }
--- a/pkg/config/auto_test.go
+++ b/pkg/config/auto_test.go
@@ -11,8 +11,9 @@ type mockEnvProvider struct {
 	envVars map[string]string
 }
 
-func (m *mockEnvProvider) Get(_ context.Context, name string) string {
-	return m.envVars[name]
+func (m *mockEnvProvider) Get(_ context.Context, name string) (string, bool) {
+	val, found := m.envVars[name]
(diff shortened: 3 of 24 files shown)


rumpl:
@AryanBagade I haven't looked at the code yet but could you please take a look at the linting issues?

AryanBagade:
> @AryanBagade I haven't looked at the code yet but could you please take a look at the linting issues?

sure will resolve the conflicts as well!

PR #1166, opened 2025-12-29
Fix #1100

### LINKED ISSUE #1100: [Detail Bug] ReadEnvFile does not trim key/value around '=', causing .env lookups to fail
# Detail Bug Report

https://app.detail.dev/org_42bf1b6f-11b3-499a-ad6d-d5a056862990/bugs/bug_2f4d6e34-53a0-4abc-bfb9-2c4950ea63c5

# Summary
- **Context**: The `ReadEnvFile` function in `pkg/environment/env_files.go` parses environment variable files (`.env` files) that are used to configure API keys, credentials, and other environment variables for the agent.
- **Bug**: Keys and values are not trimmed after splitting on the `=` delimiter, causing whitespace around the equals sign to be preserved in the parsed key-value pairs.
- **Actual vs. expected**: When parsing `KEY = VALUE`, the result is a key with trailing whitespace (`"KEY "`) and a value with leading whitespace (`" VALUE"`), but the expected behavior per Docker Compose specification is that "spaces before and after value are ignored" and both key and value should be `"KEY"` and `"VALUE"`.
- **Impact**: Environment variables with whitespace around the equals sign will not be found when the code looks them up, causing authentication failures for API keys and missing configuration values.

# Code with bug
```go
func ReadEnvFile(absolutePath string) ([]KeyValuePair, error) {
	buf, err := os.ReadFile(absolutePath)
	if err != nil {
		return nil, err
	}

	var lines []KeyValuePair

	for line := range strings.SplitSeq(string(buf), "\n") {
		line = strings.TrimSpace(line)
		if line == "" || strings.HasPrefix(line, "#") {
			continue
		}

		k, v, ok := strings.Cut(line, "=")  // <-- BUG 🔴 k and v are not trimmed after split
		if !ok {
			return nil, fmt.Errorf("invalid env file line: %s", line)
		}

		if strings.HasPrefix(v, `"`) && strings.HasSuffix(v, `"`) {
			v = strings.TrimSuffix(strings.TrimPrefix(v, `"`), `"`)
		}

		lines = append(lines, KeyValuePair{
			Key:   k,  // <-- BUG 🔴 key may have trailing whitespace
			Value: v,  // <-- BUG 🔴 value may have leading whitespace (or both if not quoted)
		})
	}

	return lines, nil
}
```

# Example
Input `.env` content:
```
ANTHROPIC_API_KEY = sk-ant-xxxxx
```

Observed parsing:
- After `strings.Cut(line, "=")` → `k = "ANTHROPIC_API_KEY "`, `v = " sk-ant-xxxxx"`
- Stored pair: `{"ANTHROPIC_API_KEY ", " sk-ant-xxxxx"}`
- Lookup `env.Get(ctx, "ANTHROPIC_API_KEY")` fails because the stored key has a trailing space

Expected behavior (per Docker Compose docs: “Spaces before and after value are ignored.”):
- `k = "ANTHROPIC_API_KEY"`, `v = "sk-ant-xxxxx"`
- Lookup succeeds

# Recommended fix
Trim both key and value after splitting, before quote handling, so quoted values still preserve internal whitespace:
```go
k, v, ok := strings.Cut(line, "=")
if !ok {
	return nil, fmt.Errorf("invalid env file line: %s", line)
}

k = strings.TrimSpace(k)  // <-- FIX 🟢 trim key
v = strings.TrimSpace(v)  // <-- FIX 🟢 trim value

if strings.HasPrefix(v, `"`) && strings.HasSuffix(v, `"`) {
	v = strings.TrimSuffix(strings.TrimPrefix(v, `"`), `"`)
}
```
This aligns parsing with `.env` conventions and resolves lookup failures for keys written as `KEY = VALUE`.

dgageot:


rumpl (approved):
Went a small rabbit hole and .env is a mess actually, some libraries allow this, others treat it as an error... 

Still, trimming space around the value seems like a bad idea...

dgageot:
yeah, I don't know what to think about it

dgageot:
Let's merge and see

PR #1167, opened 2025-12-30
Misc optimizations to make `cagent run` start faster

dgageot:


rumpl (dismissed):

pkg/tui/tui.go
```diff
@@ -117,7 +117,8 @@ func (a *appModel) Init() tea.Cmd {
 	cmds := []tea.Cmd{
 		a.dialog.Init(),
 		a.chatPage.Init(),
-		a.emitStartupInfo(),
+		// Note: Startup info (agent, team, tools) is now emitted through app.Subscribe
```
rumpl: we can remove this, will only confuse people

PR #1171, opened 2025-12-30
fix: add MarshalJSON/MarshalYAML methods to v2 RAGStrategyConfig

### LINKED ISSUE #1084: v2 RAGStrategyConfig missing marshal methods causes database field loss during serialization
# Summary
- **Context**: The v2 config parser (`pkg/config/v2/parse.go`) unmarshals YAML configuration files into v2.Config structs, which include RAG (Retrieval-Augmented Generation) strategy configurations.
- **Bug**: The `RAGStrategyConfig` type in v2 is missing `MarshalJSON` and `MarshalYAML` methods, causing critical data loss during JSON and YAML serialization.
- **Actual vs. expected**: When a v2 config with RAG strategies is serialized (e.g., during config upgrades from v2 to latest, or during push/pull operations), the `database` field and other strategy-specific fields are lost or corrupted. Expected behavior is that all fields should be preserved during round-trip serialization.
- **Impact**: Users with v2 configuration files containing RAG strategies will silently lose their database configuration when the config is loaded (since all configs are migrated to latest version), making RAG features completely non-functional.

# Code with bug
`pkg/config/v2/types.go` lines 250-316:
```go
// RAGStrategyConfig represents a single retrieval strategy configuration
// Strategy-specific fields are stored in Params (validated by strategy implementation)
type RAGStrategyConfig struct {
	Type     string            `json:"type"`               // Strategy type: "chunked-embeddings", "bm25", etc.
	Docs     []string          `json:"docs,omitempty"`     // Strategy-specific documents (augments shared docs)
	Database RAGDatabaseConfig `json:"database,omitempty"` // Database configuration
	Chunking RAGChunkingConfig `json:"chunking,omitempty"` // Chunking configuration
	Limit    int               `json:"limit,omitempty"`    // Max results from this strategy (for fusion input)

	// Strategy-specific parameters (arbitrary key-value pairs)
	// Examples:
	// - chunked-embeddings: model, similarity_metric, threshold, vector_dimensions
	// - bm25: k1, b, threshold
	Params map[string]any // Flattened into parent JSON
}

// UnmarshalYAML implements custom unmarshaling to capture all extra fields into Params
// This allows strategies to have flexible, strategy-specific configuration parameters
// without requiring changes to the core config schema
func (s *RAGStrategyConfig) UnmarshalYAML(unmarshal func(any) error) error {
	// First unmarshal into a map to capture everything
	var raw map[string]any
	if err := unmarshal(&raw); err != nil {
		return err
	}

	// Extract known fields
	if t, ok := raw["type"].(string); ok {
		s.Type = t
		delete(raw, "type")
	}

	if docs, ok := raw["docs"].([]any); ok {
		s.Docs = make([]string, len(docs))
		for i, d := range docs {
			if str, ok := d.(string); ok {
				s.Docs[i] = str
			}
		}
		delete(raw, "docs")
	}

	if dbRaw, ok := raw["database"]; ok {
		// Unmarshal database config using helper
		var db RAGDatabaseConfig
		unmarshalDatabaseConfig(dbRaw, &db)
		s.Database = db
		delete(raw, "database")
	}

	if chunkRaw, ok := raw["chunking"]; ok {
		var chunk RAGChunkingConfig
		unmarshalChunkingConfig(chunkRaw, &chunk)
		s.Chunking = chunk
		delete(raw, "chunking")
	}

	if limit, ok := raw["limit"].(int); ok {
		s.Limit = limit
		delete(raw, "limit")
	}

	// Everything else goes into Params for strategy-specific configuration
	s.Params = raw

	return nil
}
// <-- BUG 🔴 Missing MarshalYAML and MarshalJSON methods
```

`pkg/config/v2/types.go` lines 372-408:
```go
// RAGDatabaseConfig represents database configuration for RAG strategies.
// Currently it only supports a single string value which is interpreted as
// the path to a SQLite database file.
type RAGDatabaseConfig struct {
	value any // nil (unset) or string path // <-- BUG 🔴 Private field won't serialize without custom marshal methods
}

// UnmarshalYAML implements custom unmarshaling for DatabaseConfig
func (d *RAGDatabaseConfig) UnmarshalYAML(unmarshal func(any) error) error {
	var str string
	if err := unmarshal(&str); err == nil {
		d.value = str
		return nil
	}

	return fmt.Errorf("database must be a string path to a sqlite database")
}

// AsString returns the database config as a connection string
// For simple string configs, returns as-is
// For structured configs, builds connection string based on type
func (d *RAGDatabaseConfig) AsString() (string, error) {
	if d.value == nil {
		return "", nil
	}

	if str, ok := d.value.(string); ok {
		return str, nil
	}

	return "", fmt.Errorf("invalid database configuration: expected string path")
}

// IsEmpty returns true if no database is configured
func (d *RAGDatabaseConfig) IsEmpty() bool {
	return d.value == nil
}
// <-- BUG 🔴 Missing MarshalYAML and MarshalJSON methods
```

# Evidence

## Example

Consider this v2 config file:
```yaml
version: "2"
rag:
  my-rag:
    strategies:
      - type: chunked-embeddings
        database: ./test.db
        model: embed-model
        chunking:
          size: 1000
```

**Step 1**: Parse YAML → `RAGStrategyConfig.UnmarshalYAML` is called
- `database` field is extracted and stored in `RAGDatabaseConfig.value = "./test.db"`
- `model` field goes into `Params["model"] = "embed-model"`

**Step 2**: During config upgrade, `CloneThroughJSON` is called (`pkg/config/latest/upgrade.go:15`)
- The v2.Config is marshaled to JSON using Go's default JSON marshaling
- `RAGDatabaseConfig.value` is a **private field**, so it doesn't get marshaled
- Result: `database` field becomes empty in JSON output

**Step 3**: Unmarshal back into latest.Config
- The `database` field is missing/empty
- RAG strategy cannot access its database
- RAG features are broken

### Reference code
`pkg/config/latest/types.go` (lines 401-497):
```go
func (s RAGStrategyConfig) MarshalYAML() ([]byte, error) {
	result := s.buildFlattenedMap()
	return yaml.Marshal(result)
}

// MarshalJSON implements custom marshaling to flatten Params into parent level
// This ensures JSON and YAML have the same flattened format for consistency
func (s RAGStrategyConfig) MarshalJSON() ([]byte, error) {
	result := s.buildFlattenedMap()
	return json.Marshal(result)
}

// buildFlattenedMap creates a flattened map representation for marshaling
// Used by both MarshalYAML and MarshalJSON to ensure consistent format
func (s RAGStrategyConfig) buildFlattenedMap() map[string]any {
	result := make(map[string]any)

	if s.Type != "" {
		result["type"] = s.Type
	}
	if len(s.Docs) > 0 {
		result["docs"] = s.Docs
	}
	if !s.Database.IsEmpty() {
		dbStr, _ := s.Database.AsString()
		result["database"] = dbStr  // Properly extracts database from private field
	}
	// Only include chunking if any fields are set
	if s.Chunking.Size > 0 || s.Chunking.Overlap > 0 || s.Chunking.RespectWordBoundaries {
		result["chunking"] = s.Chunking
	}
	if s.Limit > 0 {
		result["limit"] = s.Limit
	}

	// Flatten Params into the same level
	for k, v := range s.Params {
		result[k] = v
	}

	return result
}
```

### Current code
`pkg/config/v2/types.go` (lines 250-316):
```go
type RAGStrategyConfig struct {
	Type     string            `json:"type"`
	Docs     []string          `json:"docs,omitempty"`
	Database RAGDatabaseConfig `json:"database,omitempty"`
	Chunking RAGChunkingConfig `json:"chunking,omitempty"`
	Limit    int               `json:"limit,omitempty"`
	Params   map[string]any
}

func (s *RAGStrategyConfig) UnmarshalYAML(unmarshal func(any) error) error {
	// ... custom unmarshaling logic ...
}

// NO MarshalYAML method
// NO MarshalJSON method
// NO buildFlattenedMap helper
```

### Contradiction
The v2 package has the same `RAGStrategyConfig` structure with the same complex unmarshaling logic and the same private fields in `RAGDatabaseConfig`, but it's **missing the corresponding marshal methods** that the `latest` package has. This creates an asymmetry: configs can be read correctly but cannot be written back correctly. The `latest` package demonstrates that these marshal methods are essential for proper serialization, yet v2 lacks them entirely.

## Codebase evidence of similar bug fix

In commit `790d8cdc` (November 28, 2025), the exact same issue was identified and fixed for `ThinkingBudget` in v2:

```
commit 790d8cdc1cc2379682b94677dd4883ceb1d138fc
Author: Christopher Petito <chrisjpetito@gmail.com>
Date:   Fri Nov 28 19:14:08 2025 +0100

    make sure thinking budgets get serialized correctly when in an agent file with version: "2"
```

The fix added `MarshalJSON` and `UnmarshalJSON` methods to `ThinkingBudget` in `pkg/config/v2/types.go` with this comment:
```go
// MarshalJSON implements custom marshaling to output simple string or int format
// This ensures JSON serialization during config upgrades preserves the value correctly
```

This demonstrates:
1. The same serialization issue affects v2 configs
2. The issue specifically manifests "during config upgrades"
3. The solution is to add marshal methods to ensure proper serialization

Additionally, commit `bd9f990e` (November 24, 2025) fixed this issue for RAG strategies in the `latest` package:
```
commit bd9f990e959694ddd0c65958dc66ddb9da045c19
Author: Christopher Petito <chrisjpetito@gmail.com>
Date:   Mon Nov 24 15:35:36 2025 +0100

    Fix thinking budget and rag strategy serialization/deserialization

    Fixes pushing/pulling agent teams using these values
```

However, this fix was only applied to `pkg/config/latest/types.go` and **not** to `pkg/config/v2/types.go`.

### Test script
```go
package main

import (
	"encoding/json"
	"fmt"
	"os"

	"github.com/docker/cagent/pkg/config/v2"
	"github.com/goccy/go-yaml"
)

func main() {
	// Test case: A v2 config with a RAG strategy containing a database field
	yamlInput := `version: "2"
agents:
  my-agent:
    model: test-model
rag:
  my-rag:
    strategies:
      - type: chunked-embeddings
        database: ./test.db
        model: embed-model
        chunking:
          size: 1000
          overlap: 100
`

	// Parse the YAML into a v2.Config
	var cfg v2.Config
	if err := yaml.Unmarshal([]byte(yamlInput), &cfg); err != nil {
		fmt.Printf("ERROR: Failed to unmarshal YAML: %v\n", err)
		os.Exit(1)
	}

	// Verify the database was correctly unmarshaled
	strategy := cfg.RAG["my-rag"].Strategies[0]
	dbStr, err := strategy.Database.AsString()
	if err != nil {
		fmt.Printf("ERROR: Failed to get database string: %v\n", err)
		os.Exit(1)
	}
	if dbStr != "./test.db" {
		fmt.Printf("ERROR: Expected database './test.db', got '%s'\n", dbStr)
		os.Exit(1)
	}
	fmt.Println("✓ Unmarshaling works correctly")

	// Now test marshaling to JSON (this is what happens during config upgrades)
	jsonBytes, err := json.Marshal(cfg)
	if err != nil {
		fmt.Printf("ERROR: Failed to marshal to JSON: %v\n", err)
		os.Exit(1)
	}

	// Unmarshal back to verify round-trip
	var cfg2 v2.Config
	if err := json.Unmarshal(jsonBytes, &cfg2); err != nil {
		fmt.Printf("ERROR: Failed to unmarshal JSON: %v\n", err)
		os.Exit(1)
	}

	// Check if the database field survived the round-trip
	if len(cfg2.RAG) == 0 {
		fmt.Printf("ERROR: RAG config is empty after round-trip\n")
		os.Exit(1)
	}
	if len(cfg2.RAG["my-rag"].Strategies) == 0 {
		fmt.Printf("ERROR: Strategies are empty after round-trip\n")
		os.Exit(1)
	}

	strategy2 := cfg2.RAG["my-rag"].Strategies[0]
	dbStr2, err := strategy2.Database.AsString()
	if err != nil {
		fmt.Printf("ERROR: Failed to get database string after round-trip: %v\n", err)
		os.Exit(1)
	}

	if dbStr2 != "./test.db" {
		fmt.Printf("BUG CONFIRMED: Database field was lost during JSON round-trip!\n")
		fmt.Printf("  Expected: './test.db'\n")
		fmt.Printf("  Got: '%s'\n", dbStr2)
		fmt.Printf("\nThis happens because RAGStrategyConfig is missing MarshalJSON/MarshalYAML methods.\n")
		fmt.Printf("The RAGDatabaseConfig.value field is private and doesn't get serialized.\n")
		os.Exit(1)
	}

	fmt.Println("✓ JSON round-trip preserves database field")
	fmt.Println("SUCCESS: No bug detected")
}
```

### Test output
```
✓ Unmarshaling works correctly
BUG CONFIRMED: Database field was lost during JSON round-trip!
  Expected: './test.db'
  Got: ''

This happens because RAGStrategyConfig is missing MarshalJSON/MarshalYAML methods.
The RAGDatabaseConfig.value field is private and doesn't get serialized.
```

# Full context

The bug occurs in the config loading and upgrade pipeline:

1. **Config Loading** (`pkg/config/config.go:18-51`): When any config file is loaded, it goes through:
   - Parse with version-specific parser (e.g., `v2.Parse` called via `parseCurrentVersion`)
   - Migrate to latest version via `migrateToLatestConfig` which calls all upgrade functions

2. **v2 to latest upgrade** (`pkg/config/latest/upgrade.go:8-17`): The upgrade function uses `types.CloneThroughJSON`:
   ```go
   func UpgradeIfNeeded(c any) (any, error) {
       old, ok := c.(previous.Config)  // previous = v2
       if !ok {
           return c, nil
       }

       var config Config
       types.CloneThroughJSON(old, &config)  // JSON round-trip to copy data
       return config, nil
   }
   ```

3. **CloneThroughJSON** (`pkg/config/types/clone.go:8-17`): This function marshals to JSON and unmarshals back:
   ```go
   func CloneThroughJSON(oldConfig, newConfig any) {
       o, err := json.Marshal(oldConfig)  // v2.Config → JSON
       if err != nil {
           panic(fmt.Sprintf("marshalling old: %v", err))
       }

       if err := json.Unmarshal(o, newConfig); err != nil {  // JSON → latest.Config
           panic(fmt.Sprintf("unmarshalling new: %v", err))
       }
   }
   ```

During the `json.Marshal(oldConfig)` step, the v2.RAGStrategyConfig is serialized using Go's default JSON marshaling. Since `RAGDatabaseConfig.value` is a private field and there's no custom `MarshalJSON` method, the database field becomes empty. Similarly, the `Params` map is serialized as a nested object instead of being flattened to the parent level.

This affects every user who:
- Has a config file with `version: "2"`
- Uses RAG features with a database configuration
- Runs the application (since configs are always loaded and migrated to latest)

The bug also affects:
- Push/pull operations for sharing agent configurations (as mentioned in commit bd9f990e)
- Any other scenario where v2 configs are serialized (e.g., saving modified configs back to disk)

### Go encoding/json documentation
From https://pkg.go.dev/encoding/json:
```
To marshal a JSON object, encoding/json uses the struct field's name as the object key,
unless the struct field has a tag specifying a name. Only exported struct fields
(those starting with an uppercase letter) are marshaled.
```

This confirms that `RAGDatabaseConfig.value` (lowercase, private field) will not be marshaled by default.

### Go YAML library documentation
From https://pkg.go.dev/github.com/goccy/go-yaml:
```
Marshaler and Unmarshaler
If you want to define a custom encoding/decoding process, you can use the Marshaler and
Unmarshaler interfaces. The Marshaler interface requires the MarshalYAML method, and the
Unmarshaler interface requires the UnmarshalYAML method.
```

This confirms that custom `MarshalYAML` methods are needed when a type has custom `UnmarshalYAML` logic.

# Why has this bug gone undetected?

Several factors have allowed this bug to remain undetected:

1. **Asymmetric testing**: Tests focus on unmarshaling (parsing) config files, not on marshaling them back. The unmarshal path works correctly because `RAGStrategyConfig.UnmarshalYAML` properly extracts the database field.

2. **Recent feature addition**: RAG support was added in commit `ea7d5850` (November 18, 2025), making it a relatively new feature. Version 2 configs with RAG may not be widely deployed yet.

3. **Upgrade path obscurity**: The bug only manifests during the upgrade process from v2 to latest, which happens automatically and silently during config loading. Users don't see the intermediate JSON serialization step that loses data.

4. **Delayed failure**: The bug doesn't cause an immediate parsing error. Instead, it causes RAG features to fail later when they try to access a database that's no longer configured, which might be attributed to user error rather than a serialization bug.

5. **Fix applied only to latest**: The marshaling methods were added to the `latest` package in commit `bd9f990e` (November 24, 2025), but this fix was never backported to v2. Developers may have assumed that fixing `latest` was sufficient since all configs eventually upgrade to it, but they overlooked that v2 configs must serialize correctly *during* the upgrade process itself.

6. **Similar bug precedent**: The same issue occurred with `ThinkingBudget` and was fixed in v2 (commit `790d8cdc`), but this didn't trigger a systematic review of other v2 types with similar patterns (custom unmarshal + private fields).

# Recommended fix

Add the same marshaling methods to `pkg/config/v2/types.go` that already exist in `pkg/config/latest/types.go`:

```go
// Add to RAGStrategyConfig in pkg/config/v2/types.go:

func (s RAGStrategyConfig) MarshalYAML() ([]byte, error) {
	result := s.buildFlattenedMap()
	return yaml.Marshal(result)
}

func (s RAGStrategyConfig) MarshalJSON() ([]byte, error) {  // <-- FIX 🟢
	result := s.buildFlattenedMap()
	return json.Marshal(result)
}

func (s RAGStrategyConfig) buildFlattenedMap() map[string]any {  // <-- FIX 🟢
	result := make(map[string]any)

	if s.Type != "" {
		result["type"] = s.Type
	}
	if len(s.Docs) > 0 {
		result["docs"] = s.Docs
	}
	if !s.Database.IsEmpty() {
		dbStr, _ := s.Database.AsString()
		result["database"] = dbStr  // Properly serialize database
	}
	if s.Chunking.Size > 0 || s.Chunking.Overlap > 0 || s.Chunking.RespectWordBoundaries {
		result["chunking"] = s.Chunking
	}
	if s.Limit > 0 {
		result["limit"] = s.Limit
	}

	// Flatten Params into the same level
	for k, v := range s.Params {
		result[k] = v
	}

	return result
}
```

This fix mirrors the solution already implemented in `pkg/config/latest/types.go` and follows the same pattern used to fix `ThinkingBudget` in commit `790d8cdc`.

majiayu000:
## Summary
This PR fixes #1084

## Changes
- Added `MarshalJSON()` method to `RAGStrategyConfig` to preserve data during JSON serialization
- Added `MarshalYAML()` method to `RAGStrategyConfig` to preserve data during YAML serialization  
- Added `UnmarshalJSON()` method to `RAGStrategyConfig` to properly parse flattened JSON
- Added `buildFlattenedMap()` helper function shared by both marshal methods for consistent output format
- Added comprehensive tests for JSON marshal/unmarshal and round-trip serialization

The implementation ensures that strategy-specific parameters stored in the `Params` map are flattened into the parent level during marshaling (matching the YAML format), preventing database field loss during round-trip serialization.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/pkg/config/v2/types_test.go
+++ b/pkg/config/v2/types_test.go
@@ -153,8 +153,8 @@ func TestRAGStrategyConfig_UnmarshalJSON(t *testing.T) {
 	require.Equal(t, "./bm25.db", mustGetDBString(t, strategy.Database))
 	require.Equal(t, 20, strategy.Limit)
 	require.NotNil(t, strategy.Params)
-	require.Equal(t, 1.2, strategy.Params["k1"])
-	require.Equal(t, 0.75, strategy.Params["b"])
+	require.InEpsilon(t, 1.2, toFloat64(strategy.Params["k1"]), 0.001)
+	require.InEpsilon(t, 0.75, toFloat64(strategy.Params["b"]), 0.001)
 }
 
 func TestRAGStrategyConfig_JSONRoundTrip(t *testing.T) {


dgageot:
@majiayu000 the linter is not happy :-)

krissetto (approved):

PR #1182, opened 2025-12-31
Use default instructions less likely to trigger jailbreak detection

### LINKED ISSUE #1125: Make default user message configurable via environment variable to avoid Azure OpenAI jailbreak detection
**What you'd like to see**

I would like to be able to customize the default user message that is used when an Agent is called without an explicit user message. Currently, the string "Follow the default instructions" is hardcoded. I propose making this configurable via an environment variable (e.g., CAGENT_DEFAULT_USER_MESSAGE).

If the environment variable is not set: Fallback to the current default "Follow the default instructions".

If set: Use the string provided in the environment variable.

**Why you'd like to see it**

When performing Agent-to-Agent (A2A) execution using the Azure OpenAI API, the current hardcoded message "Follow the default instructions" is sometimes flagged by Azure's content safety filters as a potential jailbreak attempt, resulting in an error response. Being able to change this message is essential to bypass these filters and ensure stable execution in Azure environments.

**Workarounds?**

Currently, the only way to fix this is to modify the source code and rebuild the binary. Introducing an environment variable would eliminate this friction.

**Additional context**

The relevant code parts are as follows:

cagent/pkg/runtime/runtime.go (Line 1322):
session.WithImplicitUserMessage("Follow the default instructions"),

cagent/cmd/root/run.go (Line 225):
execArgs = append(execArgs, "Follow the default instructions")

dgageot:
Fixes #1125

@gama-katsushi, let's try with those instructions first

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -238,7 +238,7 @@ func (f *runExecFlags) handleExecMode(ctx context.Context, out *cli.Printer, rt 
 	if len(args) == 2 {
 		execArgs = append(execArgs, args[1])
 	} else {
-		execArgs = append(execArgs, "Follow the default instructions")
+		execArgs = append(execArgs, "Please proceed.")
 	}
 
 	err := cli.Run(ctx, out, cli.Config{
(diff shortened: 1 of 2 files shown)


pkg/model/provider/provider.go
```diff
@@ -51,6 +51,10 @@ var Aliases = map[string]Alias{
 		BaseURL:     "https://api.mistral.ai/v1",
 		TokenEnvVar: "MISTRAL_API_KEY",
 	},
+	"ollama": {
```
rumpl: Cool but unrelated :)

trungutt (approved):

PR #1204, opened 2026-01-03
Add tool permissions system with Allow/Ask/Deny patterns

dgageot:
Implement a permissions configuration that allows controlling tool approval behavior through pattern matching:

- Add permissions section to config with allow and deny string arrays
- Deny patterns are checked first and block execution (highest priority)
- Allow patterns auto-approve tools without user confirmation
- Default behavior requires user confirmation (Ask)

Pattern matching supports:
- Glob-style wildcards: * matches any sequence of characters
- ? matches any single character
- [abc] matches character classes
- Case insensitive matching
- **Argument matching**: tool:arg=pattern to match specific argument values e.g., 'shell:cmd=ls*' allows shell with cmd starting with 'ls'

Assisted-By: cagent

### DIFF
--- a/cagent-schema.json
+++ b/cagent-schema.json
@@ -45,6 +45,10 @@
     "metadata": {
       "$ref": "#/definitions/Metadata",
       "description": "Configuration metadata"
+    },
+    "permissions": {
+      "$ref": "#/definitions/PermissionsConfig",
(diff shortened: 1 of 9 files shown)


dgageot:
Let's merge is and iterate to improve!

PR #1205, opened 2026-01-03
Implement hooks

dgageot:
This is useful to:
+ Allow/Reject tool calls based on their arguments
+ Tweak tool calls
+ Emit warnings on tool calls
+ Add content to the context on tool calls 
+ Run commands like linters after tools are executed


rumpl (dismissed):
LGTM

rumpl (approved):
LGTM

PR #1209, opened 2026-01-03
Very basic sandboxing support for the 'shell' tools

dgageot:
Allow the shell toolset to run in a docker container with a user-defined list of bind mounts and permissions

This is the beginning of a sandboxing story


dgageot:
Let's merge is and iterate to improve!

PR #1226, opened 2026-01-04
Use Js to run tool calls in /commands

dgageot:


rumpl:
A loy of chsnges in the tui, was that intnetional?

dgageot:
yes and no. I got too lazy to open another PR

rumpl (approved):

PR #1238, opened 2026-01-05
Simpler code

dgageot:


rumpl:
Wow, it's so much simpler now :D

PR #1242, opened 2026-01-05
Lazy creation of the session

dgageot:
This is to avoid having too many empty sessions in the store

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -208,7 +208,7 @@ func (f *runExecFlags) createRemoteRuntimeAndSession(ctx context.Context, origin
 	return remoteRt, sess, nil
 }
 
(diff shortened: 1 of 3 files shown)


cmd/root/run.go
```diff
@@ -234,9 +234,8 @@ func (f *runExecFlags) createLocalRuntimeAndSession(ctx context.Context, t *team
 		session.WithHideToolResults(f.hideToolResults),
 	)
 
-	if err := sessStore.AddSession(ctx, sess); err != nil {
-		return nil, nil, err
-	}
+	// Session is stored lazily on first UpdateSession call (when content is added)
```
rumpl, line 237: we can remove this comment, it's kinda useless

PR #1263, opened 2026-01-06
Make CustomerMatcher simpler

trungutt:
Match requests by first user message only, ignoring other metadata fields


trungutt:
PR closed, I need to think more about this

PR #1278, opened 2026-01-06
Model changing in TUI

krissetto:
### `/model` changing in the TUI

Let's decide if we like the approach first, please give it a try :)

Some info: 
- Allows changing the model of the agent you're chatting with with a `/model` command or with `ctrl+m`;
- Includes **all** models defined in the yaml, both in the `models` section and inline in agents, along with the option of typing in a custom model using the shorthand `provider/model` syntax. Supports defining custom alloyed models as well;
- Changed models get persisted to the session for future use when reloaded with the `/sessions` command;
- (default) is shown next to the model name of the agent defined in the yaml, (current) for the selected model

---

Screenshots:

Screenshot 2026-01-08 at 12 09 22
Screenshot 2026-01-08 at 12 09 31
Screenshot 2026-01-08 at 12 09 39
Screenshot 2026-01-08 at 12 10 53
Screenshot 2026-01-08 at 13 34 53
Screenshot 2026-01-08 at 12 09 51
Screenshot 2026-01-08 at 12 10 27

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ experts that collaborate to solve complex problems for you.
 - **🌐 AI provider agnostic** - Support for OpenAI, Anthropic, Gemini, xAI,
   Mistral, Nebius and [Docker Model
   Runner](https://docs.docker.com/ai/model-runner/).
+- **🔀 Runtime model switching** - Change models on-the-fly during a session
+  with the `/model` command, with automatic persistence across session reloads.
 
 ## Your First Agent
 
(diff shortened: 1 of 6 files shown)


rumpl:
I just tried and was able to type ctrl+m, `openai/` <ENTER> and then everything was broken, can we check for the obvious `thing/` and `/thing` mistakes please?

krissetto:
> I just tried and was able to type ctrl+m, `openai/` and then everything was broken, can we check for the obvious `thing/` and `/thing` mistakes please?

Fixed and added screenshot

PR #1289, opened 2026-01-06
Fix message width

dgageot:


rumpl (approved):
"Oh cool, so what do you do at docker?"
"I fix widths"

PR #1307, opened 2026-01-07
Improve the documentation a bit

dgageot:


docs/TELEMETRY.md
```diff
@@ -1,8 +1,8 @@
 # Telemetry for cagent
```
derekmisler: **question (non-blocking):** should we also convert headings to this new format?

```suggestion
# Telemetry for Docker `cagent`
```

derekmisler (approved):

PR #1310, opened 2026-01-07
Fix race conditions

dgageot:
Detected with:

```
go test --count=1 -race ./...
```


pkg/session/session.go
```diff
@@ -175,6 +183,13 @@ func (s *Session) AllowedDirectories() []string {
 
 // GetAllMessages extracts all messages from the session, including from sub-sessions
 func (s *Session) GetAllMessages() []Message {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.getAllMessagesLocked()
+}
+
+// getAllMessagesLocked extracts all messages without acquiring the lock (caller must hold lock)
+func (s *Session) getAllMessagesLocked() []Message {
```
krissetto, line 192: are there any scenarios where we want to call this func without the caller having acquired the lock beforehand?

PR #1313, opened 2026-01-07
Expand $EDITOR in the TUI

jeanlaurent:
Expand $EDITOR to actual editor name in TUI help text

  The TUI help footer now displays the actual configured editor name instead of the literal
  string "$EDITOR".

  Before: Ctrl+g edit in $EDITOR
  After: Ctrl+g edit in VSCode (or Vim, Neovim, Cursor, etc.)

  The implementation:
  - Reads VISUAL or EDITOR environment variables
  - Extracts base command name from paths and arguments
  - Maps common editors (code, vim, nvim, cursor, etc.) to friendly display names
  - Falls back to capitalized base name for unknown editors
  - Returns "$EDITOR" if not configured

so moving from
Screenshot 2026-01-07 at 19 33 50
to
Screenshot 2026-01-07 at 19 34 22

### DIFF
--- a/pkg/tui/page/chat/chat.go
+++ b/pkg/tui/page/chat/chat.go
@@ -1,10 +1,13 @@
 package chat
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"log/slog"
 	"os"
+	"path/filepath"
+	"runtime"
 	"strings"
 
 	"charm.land/bubbles/v2/help"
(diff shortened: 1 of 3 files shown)


pkg/tui/page/chat/chat.go
```diff
@@ -108,8 +110,60 @@ type KeyMap struct {
...
+		"gedit":    "gedit",
+		"kate":     "Kate",
+		"notepad":  "Notepad",
+		"notepad++": "Notepad++",
+		"textmate": "TextMate",
+		"mate":     "TextMate",
+		"zed":      "Zed",
+	}
+
+	if displayName, ok := editorNames[baseName]; ok {
```
rumpl: It would be better to check if it starts with, an $EDITOR coud be `code --wait` for example

~/english_corrector.yaml
```diff
@@ -0,0 +1,50 @@
+version: 2
```
dgageot: error?

pkg/tui/page/chat/chat_test.go
```diff
@@ -0,0 +1,98 @@
...
+			name:     "Vim GTK",
+			envValue: "vim-gtk3",
+			want:     "Vim",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Set test env
+			t.Setenv("VISUAL", "")
```
dgageot: Can you avoid setting env variables in tests? It prevents run running tests in parallel

pkg/tui/page/chat/chat_test.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+		{
+			name:     "Unknown editor with full path",
+			envValue: "/opt/bin/myeditor",
+			want:     "Myeditor",
+		},
+		{
+			name:     "Empty",
+			envValue: "",
+			want:     "$EDITOR",
```
dgageot: For me, $EDITOR means vim. Probably a default somewhere. Do you handle that?

krissetto:
One thing we could add here, or do in a subsequent PR, is removing the footer option altogether for those users that don't have $EDITOR set, since it just flashes the screen atm

dgageot (approved):

PR #1334, opened 2026-01-09
Try parsing the arguments before they are completely received

rumpl:
This makes it possible for the TUI to show the partial arguments (for example the file that the LLM wants to write) before the tool call is completely received

### DIFF
--- a/pkg/tui/components/toolcommon/common.go
+++ b/pkg/tui/components/toolcommon/common.go
@@ -17,14 +17,31 @@ import (
 // Returns an error if parsing fails.
 func ParseArgs[T any](args string) (T, error) {
 	var result T
-	if err := json.Unmarshal([]byte(args), &result); err != nil {
-		return result, err
+	var err error
+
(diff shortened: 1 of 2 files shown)


pkg/tui/components/toolcommon/common.go
```diff
@@ -17,14 +17,30 @@ import (
...
+		return result, nil
+	}
+
+	for _, fixed := range tryFixPartialJSON(args) {
+		if err := json.Unmarshal([]byte(fixed), &result); err == nil {
+			return result, nil
+		}
+	}
+
+	return result, fmt.Errorf("cannot parse partial JSON")
```
krissetto: nit: This could be an error for non-partial json as well if the llm just doesn't reply correctly, and i guess users wouldn't see this error on partial json during a stream, so maybe we should change the msg. wdyt?

rumpl: Done. Returning the original error now

PR #1337, opened 2026-01-09
[fix] when opening past session, ignore questions asked on the command line

dgageot:


krissetto:
should we error out or maybe show the user a notification too? to avoid confusion

PR #1338, opened 2026-01-09
Add config version v4

dgageot:


AGENTS.md
```diff
@@ -239,7 +239,7 @@ for _, tt := range tests {
 - All agent references must exist in config
 - Model references can be inline (e.g., `openai/gpt-4o`) or defined in models section
 - Tool configurations validated at startup
-- Config versioning: Currently on v3 (sequential migration: v0 → v1 → v2 → v3)
+- Config versioning
```
rumpl, line 242: ?

rumpl:
Ok but why

AGENTS.md
dgageot: I'm removing it because we have to think that it needs updating and I don't really know how useful it is

dgageot:
@rumpl This new commit is the why. It's going to keep the agents in order of definition now

PR #1349, opened 2026-01-12
Fix token usage and cost tracking in amazon-bedrock provider

krissetto:
Usage was arriving after finish
Model names used for cost tracking with models.dev weren't used correctly

### DIFF
--- a/pkg/modelsdev/store.go
+++ b/pkg/modelsdev/store.go
@@ -177,6 +177,25 @@ func (s *Store) GetModel(ctx context.Context, id string) (*Model, error) {
 
 	model, exists := provider.Models[modelID]
 	if !exists {
+		// For amazon-bedrock, try stripping region/inference profile prefixes
+		// Bedrock uses prefixes like "global.", "us.", "eu.", "apac." etc. for
+		// cross-region inference profiles, but models.dev stores models without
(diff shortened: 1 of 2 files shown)


pkg/modelsdev/store.go
```diff
@@ -177,6 +177,25 @@ func (s *Store) GetModel(ctx context.Context, id string) (*Model, error) {
 
 	model, exists := provider.Models[modelID]
 	if !exists {
+		// For amazon-bedrock, try stripping region/inference profile prefixes
+		// Bedrock uses prefixes like "global.", "us.", "eu.", "apac." etc. for
+		// cross-region inference profiles, but models.dev stores models without
+		// these prefixes. Try stripping the first segment if it doesn't match
+		// a known model provider prefix (anthropic, meta, amazon, etc.)
+		if providerID == "amazon-bedrock" {
```
rumpl, line 185: The modelID comes from the provider (the `ID()` method), why not do this cleanup in there?

krissetto: We can i guess, i wasn't totally sure because using it elsewhere it's useful to have the full id

krissetto: I can clean up in a follow up PR if we want, but this is mostly a models.dev specific concern either way so i don't think we should change the provider's ID() func.

PR #1360, opened 2026-01-13
feat(bedrock): add prompt caching for supported models

mfenderov:
Prompt caching is automatically enabled for models that support it (detected via models.dev) to reduce latency and costs. System prompts, tool definitions, and recent messages are cached with a 5-minute TTL.

To disable:

```yaml
provider_opts:
  disable_prompt_caching: true
```

P.S. Benchmarked with `examples/pr-reviewer-bedrock.yaml`: 92% cache read vs 8% cache write.

Assisted-By: cagent

### DIFF
--- a/pkg/model/provider/bedrock/client.go
+++ b/pkg/model/provider/bedrock/client.go
@@ -20,13 +20,15 @@ import (
 	"github.com/docker/cagent/pkg/environment"
 	"github.com/docker/cagent/pkg/model/provider/base"
 	"github.com/docker/cagent/pkg/model/provider/options"
+	"github.com/docker/cagent/pkg/modelsdev"
 	"github.com/docker/cagent/pkg/tools"
 )
 
 // Client represents a Bedrock client wrapper implementing provider.Provider
 type Client struct {
 	base.Config
-	bedrockClient *bedrockruntime.Client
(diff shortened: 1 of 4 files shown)


krissetto:
This generally looks good to me, thanks for the contribution! ❤️ 

@dgageot since you've been deep diving on anthropic caching bits lately, you might be interested in taking a look here and maybe comparing bedrock to some of the tests you've been doing with anthropic's api

krissetto:
Seems to work fine, I'm merging this

pkg/model/provider/bedrock/client.go
```diff
@@ -109,19 +110,47 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
...
+	modelID := "amazon-bedrock/" + model
+	m, err := store.GetModel(ctx, modelID)
+	if err != nil {
+		slog.Debug("Bedrock prompt caching disabled: model not found in models.dev",
+			"model_id", modelID, "error", err)
+		return false
+	}
+
+	return m.Cost != nil && (m.Cost.CacheRead > 0 || m.Cost.CacheWrite > 0)
+}
```
krissetto, line 152: this is something we can put in the modelsdev store as it could be useful for all providers

PR #1371, opened 2026-01-15
Simplify the reviewer

rumpl:


.github/workflows/agents/pr-review.yaml
```diff
@@ -1,119 +1,24 @@
...
-      8. Format your review with INLINE COMMENTS in this structure:
-         - Overall summary at the top
-         - For each concern/suggestion, use format: `[file.ext:LINE] Comment text`
-         - Include file path and NEW line number (from diff +line numbers) for each inline comment
-      9. Post the review to GitHub using gh CLI:
+      1. Use the shell to call "gh" to get all of the information about the Pr
+      2. Review the PR diff, focusing on lines that were ADDED (+) or MODIFIED.
+      3. Make sure to get the overall picture about the changes, read the files from the current directory as needed
+      4. Review the code changes in detail
+      5. Ue gh to add inline comments for specific lines of code that need attention
```
derekmisler:
```suggestion
      5. Use gh to add inline comments for specific lines of code that need attention
```

.github/workflows/agents/pr-review.yaml
```diff
@@ -1,119 +1,24 @@
...
-
-      START COPYING HERE:
-
-      YOU ARE A BOT IN CI/CD. NO HUMAN INTERACTION. REVIEW THE DIFF AND STOP.
-
-      ## CI/CD Context
-      - ZERO human interaction possible
-      - Diff is 100% COMPLETE - do not request more
-      - If uncertain, state assumption and move on
-      - Review must be FINAL in one pass
```
derekmisler, line 101: it's possible the agent is smarter than it used to be, but i ran into this problem a bunch when testing because the agent always had questions, so it's worth keeping an eye on in the future

PR #1375, opened 2026-01-15
Improve caching

dgageot:
+ Cut the message that contained the date, env, agents.md and skills into separate system messages
+ Reordered the system messages to have the most invariant messages first
+ Instead of caching the first two system messages and the last two user/assistant messages, we mark the tool instructions with a cache control checkpoint because it's the last system message before the ones that can be more user specific
+ We mark the last system message too.
+ We keep on marking the last user/assistant messages with the cache control flag.

Added a few tests and e2e on the way.


pkg/session/session.go
```diff
@@ -440,10 +449,33 @@ func (s *Session) GetMessages(a *agent.Agent) []chat.Message {
...
+	for _, toolSet := range a.ToolSets() {
+		if toolSet.Instructions() != "" {
+			messages = append(messages, chat.Message{
+				Role:    chat.MessageRoleSystem,
+				Content: toolSet.Instructions(),
+			})
+		}
+	}
+
+	// Cache control checkpoint #1 out of 4
```
simonferquel, line 468: This is a bit brittle to only rely on where you add messages in the « system slices » compared to that line of code to make it part of what we think is invariant or not. It is a bit too easy to break the cache.
Can we separate the messages into different slices (invariant messages, lowFrequencyChangeMessages, …) and put the checkpoint when we merge them (in correct order) in the system messages slice?

dgageot: @simonferquel Can we merge this and you prepare another PR?

simonferquel: sure

PR #1377, opened 2026-01-15
Unstuck some remote MCPs

rumpl:
Some remote MCPs get stuck when we try to connect to them because the MCP client also tries to create a persistent SSE connection.

Disable the persistent connection to make init faster for these servers. This means that we can't receive notifications from remote servers which is fine for now since we don't listen to these (yet).

More info https://github.com/modelcontextprotocol/go-sdk/issues/633


dgageot:
Yeah, bumping this dependency is not as easy, they've changed a few things

PR #1379, opened 2026-01-16
Fix agent lookup for local file sources

trungutt:
When using a local file as an agent source (e.g., `~/<some_path>/<agent_name>.yaml`), the API endpoint for `runAgent` lookup would fail because it couldn't fine the agent.

Root cause: `ResolveSources()` stores local file sources with keys stripped of the .yaml extension (via `fileNameWithoutExt`), but OCI references are stored with .yaml extension (via `OciRefToFilename`). When the API receives a request with the .yaml extension in the path, the lookup fails for local files.

This PR creates a fallback lookup to ensure it works.


simonferquel:
wouldn't it be better to fix the root cause ? also if we do not, we should have a comment in the code explaining why we do need that fallback mechanism.

trungutt:
> wouldn't it be better to fix the root cause ? also if we do not, we should have a comment in the code explaining why we do need that fallback mechanism.

`(for local files keyed without .yaml)` that's why I made this PR this way

trungutt:
Well I'm figuring out another fix

PR #1389, opened 2026-01-16
Update jsonschema

rumpl:
With this update jsonschema decided that slices and such have a type of [null, string], which is correct in golang world, but it breaks Gemini so we need to tweak and remove all the array defined types and take only the type itself.

Also moved the code a bit and unified the way we convert parameters in openai


e2e/cagent_exec_test.go
```diff
@@ -87,7 +87,9 @@ func TestExec_Mistral(t *testing.T) {
 func TestExec_Mistral_ToolCall(t *testing.T) {
 	out := cagentExec(t, "testdata/fs_tools.yaml", "--model=mistral/mistral-small", "How many files in testdata/working_dir? Only output the number.")
 
-	require.Equal(t, "\n--- Agent: root ---\n\nCalling list_directory(path: \"testdata/working_dir\")\n\nlist_directory response → \"FILE README.me\\n\"\n1", out)
+	// NOTE: If you look at the LLM response, Mistral says it sees 2 files, yours truly got tired of re-running this test to get it to say "1".
```
krissetto, line 90: 😂

pkg/model/provider/openai/client.go
```diff
@@ -339,12 +336,6 @@ func (c *Client) CreateResponseStream(
 				return nil, err
 			}
 
-			// The Response API requires every parameter to be required
-			parameters = makeAllRequired(parameters)
```
krissetto, line 343: is this safe to remove? wasn't expecting this change here

krissetto (approved):

pkg/model/provider/openai/client.go
dgageot: back then it wasn't. @rumpl have you tested it?

rumpl: It wasn't removed, I moved these to `ConvertParametersToSchema` so both Responses and Chat APIs will get the same treatment.

Yes I have tested this, our e2e tests still work and I tested with the responses api too

PR #1391, opened 2026-01-16
Better DMR support

krissetto:
Makes endpoint selection more resilient. If `base_url` is not set, `cagent` will try multiple well-known URL to best support the environment it's deployed in (local binary, container on linux, in a container on DD on mac/windows)

Also uses the /engines/_configure endpoint to configure models instead of shelling out to the CLI.

### DIFF
--- a/pkg/model/provider/dmr/client.go
+++ b/pkg/model/provider/dmr/client.go
@@ -35,10 +35,23 @@ import (
 )
 
 const (
+	// configureTimeout is the timeout for the model configure HTTP request.
+	// This is kept short to avoid stalling client creation.
+	configureTimeout = 10 * time.Second
+
+	// connectivityTimeout is the timeout for testing DMR endpoint connectivity.
+	// This is kept short to quickly detect unreachable endpoints and try fallbacks.
+	connectivityTimeout = 2 * time.Second
+)
+
+const (
 	// dmrInferencePrefix mirrors github.com/docker/model-runner/pkg/inference.InferencePrefix.
 	dmrInferencePrefix = "/engines"
 	// dmrExperimentalEndpointsPrefix mirrors github.com/docker/model-runner/pkg/inference.ExperimentalEndpointsPrefix.
 	dmrExperimentalEndpointsPrefix = "/exp/vDD4.40"
+
+	// dmrDefaultPort is the default port for Docker Model Runner.
+	dmrDefaultPort = "12434"
 )
 
 // Client represents an DMR client wrapper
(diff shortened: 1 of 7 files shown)


pkg/model/provider/dmr/client.go
```diff
@@ -127,6 +180,48 @@ func inContainer() bool {
...
+
+// getDMRFallbackURLs returns a list of fallback URLs to try for DMR connectivity.
+// The order is chosen to maximize compatibility across platforms:
+// 1. model-runner.docker.internal - Docker Desktop's integrated model-runner
+// 2. host.docker.internal - Docker Desktop's host access (works on macOS/Windows/Linux Desktop)
+// 3. 172.17.0.1 - Default Docker bridge gateway (Linux Docker CE)
+// 4. 127.0.0.1 - Localhost (when running directly on host)
+func getDMRFallbackURLs() []string {
+	return []string{
+		fmt.Sprintf("http://%s%s/v1/", dmrModelRunnerInternal, dmrInferencePrefix),
```
simonferquel: I think we can be a bit better at making the fallback list:
- detect if we are in a container -> fallback to the model-runner.docker.internal
- if docker desktop socket is found (we already have code that reaches the docker desktop backend) -> host.docker.internal
- otherwise default dd bridge / localhost

pkg/model/provider/dmr/client.go
```diff
@@ -34,11 +34,31 @@ import (
...
 	dmrInferencePrefix = "/engines"
 	// dmrExperimentalEndpointsPrefix mirrors github.com/docker/model-runner/pkg/inference.ExperimentalEndpointsPrefix.
 	dmrExperimentalEndpointsPrefix = "/exp/vDD4.40"
+
+	// dmrDefaultPort is the default port for Docker Model Runner.
+	dmrDefaultPort = "12434"
+
+	// Docker Desktop internal hostnames and fallback IPs for cross-platform support.
+	// These are tried in order when the primary endpoint is unreachable.
+	dmrModelRunnerInternal = "model-runner.docker.internal" // Docker Desktop's model-runner service
```
simonferquel: those consts are only used within `getDMRFallbackURLs`. should'nt we move them inside the function ?

pkg/model/provider/dmr/client.go
krissetto: I might be tired but i'm not 100% sure i understood exactly what you meant here :) either way i made some adjustments, feel free to give it a try on windows as well as that's where i haven't yet tested at all

PR #1394, opened 2026-01-16
Send events for thinking traces on OpenAI models

krissetto:
Send reasoning events from the runtime when using openai reasoning models and consume them in the TUI to display the reasoning summaries.

Effort is added only when `thinking_budget` is defined to align with the current anthropic provider. 

Default model settings to be changed in  followup PR

---

Screencap:

https://github.com/user-attachments/assets/34dee2cc-f844-4412-b7f0-defa0f531895


krissetto:
looks big, its mostly test code

PR #1395, opened 2026-01-16
Better model config defaults

krissetto:
To give users a better experience with minimal need for configuration and searching the docs, this PR enables thinking and interleaved thinking by default with a medium/dynamic thinking budget, for the OpenAI, Anthropic, Google and Amazon Bedrock providers

For the bedrock provider, the defaults only apply to anthropic models for now


krissetto:
Let's make sure to merge https://github.com/docker/cagent/pull/1399 before this to avoid issues with title generation

dgageot (approved):

PR #1396, opened 2026-01-17
Pass message level costs in --remote mode and `cagent exec --json`

dgageot:


krissetto (approved):
maybe we should think about refactoring some usage bits in the future.. between `runtime.Usage`, `runtime.MessageUsage` and `chat.Usage`, and them all being somewhat related, its getting a bit convoluted to understand imho

PR #1401, opened 2026-01-17
Add MCP elicitation support to TUI

### LINKED ISSUE #450: mcp elicitation support
**What you'd like to see**

The docker-mcp gateway supports forwarding of elicitation messages from servers that it's running. It also uses a few internal tools that will check whether a client supports elicitations and then potentially raise elicitation messages for some use cases.

This is relatively low priority because so many clients do not support them.  The gateway checks for client support before attempting to either use or forward them.  

**Why you'd like to see it**

Using elicitations when a remote server is added dynamically could be a good way to elicit a custom consent screen. The nanobot client currently support elicitations.

Image

**Workarounds?**

Fallbacks for elicitations are standard practice so we're doing fine without them. I just don't think it's okay for an mcp _gateway_ to not support them. We want to encourage MCP authors to use them.

dgageot:
Implement full MCP elicitation support in the runtime and TUI, distinguishing between OAuth flows and standard elicitation requests.

- Add ElicitationDialog component that parses JSON schema to create form fields (string, boolean, integer, number, enum types)
- Add ElicitationResponseMsg to communicate responses from dialog
- Update handleElicitationRequest to route OAuth vs form elicitation
- Handle ElicitationResponseMsg in TUI to call ResumeElicitation

The elicitation dialog supports keyboard navigation, required field validation, and proper type conversion for submitted values.

Assisted-By: cagent

Fixes #450


examples/elicitation/README.md
krissetto, line 1: Let's link to this from README.md and/or USAGE.md with a nice mention.. so its slightly more visible to users that the feature exists 😂

pkg/tui/page/chat/runtime_events.go
```diff
@@ -210,13 +210,38 @@ func (p *chatPage) handleMaxIterationsReached(msg *runtime.MaxIterationsReachedE
...
+			dialogCmd := core.CmdHandler(dialog.OpenDialogMsg{
+				Model: dialog.NewOAuthAuthorizationDialog(serverURL, p.app),
+			})
+			return tea.Batch(spinnerCmd, dialogCmd)
+		}
+	}
 
-	return tea.Batch(spinnerCmd, dialogCmd)
+	// Check elicitation mode
+	switch msg.Mode {
```
krissetto, line 232: Just as a safeguard we should probably validate this `Mode` and show an error/notification or something.

`"form"` is the default behavior only when the "mode" is omitted, i guess for compatibility reasons, but i assume the spec could add more modes and servers might end up sending us different data

pkg/tools/mcp/stdio.go
```diff
@@ -57,6 +66,42 @@ func (c *stdioMCPClient) Initialize(ctx context.Context, _ *mcp.InitializeReques
...
+
+// SetElicitationHandler sets the elicitation handler for stdio MCP clients
+func (c *stdioMCPClient) SetElicitationHandler(handler tools.ElicitationHandler) {
+	c.mu.Lock()
+	c.elicitationHandler = handler
+	c.mu.Unlock()
+}
+
+// SetOAuthSuccessHandler is a no-op for stdio clients (OAuth not supported)
+func (c *stdioMCPClient) SetOAuthSuccessHandler(func()) {}
```
krissetto, line 100: ? I see some OAuth bits in the TUI code, am I missing something or is this just for a separate PR?

examples/elicitation/README.md
rumpl: Not the readme please, it's already too crowded and no one cares about elicitation

PR #1414, opened 2026-01-19
fix(api): reject invalid resume confirmation with explicit error

### LINKED ISSUE #1376: tool confirmation failures are mysterious
**Describe the bug**

If the request to approve a tool call fails, the output is extremely mysterious:
- the cagent api returns 200
- the agent has no idea what happened

**Version affected**

v1.19.1

**How To Reproduce**

I pushed up a branch with a test harness to make it easier to see the behavior

https://github.com/docker/cagent/commit/e11eee46825e17f48f5288b7610dc3ca73c57781

1. Run `cagent api` with an agent config that has a tool
2. Call the /resume api with `{"confirmation": "garbage"}`

Output:

```
Session created: ea7e7209-1c5e-46a8-8d36-2dd5550c1052

Running agent...

[Stream Started] Session: ea7e7209-1c5e-46a8-8d36-2dd5550c1052, Agent: root
[Tool Confirmation Required] start-task
[Resuming session with: garbage]
The tool execution failed. Unfortunately, the system returned a generic failure message without providing specific error details. This appears to be a bug in the `start-task` tool - it should return more informative error messages to help diagnose what went wrong, such as:

- Whether the task name "demo-task" is invalid or already exists
- Whether there are permission issues
- Whether the backend service is unavailable
- What the specific validation or system error is

**Bug Report:** The `start-task` tool needs to provide more descriptive error messages when execution fails, rather than just reporting "(tool execution failed)" without context.
```

**Expectation**

The agent should report some sort of meaningful error -- e.g., that the tool call was rejected.

Pnkcaht:
## What I did

Added explicit validation for /resume confirmation values so invalid inputs (e.g. "garbage") are rejected early instead of causing opaque tool execution failures.

This ensures the agent reports a clear user error when a tool call is rejected, rather than surfacing a misleading (tool execution failed) message.

## Related issue

Fixes #1376

## Notes

- Validates resume confirmations at the API boundary
- Adds defensive validation inside the runtime
- Invalid confirmations never reach tool execution
- Produces a clear, actionable error message for the user
- Does not change behavior for valid approve / reject flows
- Keeps backward compatibility

### Before

When calling /resume with an invalid confirmation value (e.g. "garbage"):

- The API returned 200 OK despite the invalid confirmation value
- The runtime forwarded the invalid value to the agent
- Tool execution failed later with a generic (tool execution failed)
- The agent had no indication that the tool call was explicitly rejected
- From the user’s perspective, the failure looked like a tool bug, not a user input error

### After

When calling /resume with an invalid confirmation value:

- The confirmation is validated immediately
- The request is rejected with a clear, explicit error
- The runtime and agent never attempt tool execution
- The user is informed that the tool call was rejected due to invalid confirmation
- The error message clearly states the allowed values (approve, reject)

### Screenshot
Below is a photo of the new return.

image

### DIFF
--- a/pkg/runtime/resume_validation.go
+++ b/pkg/runtime/resume_validation.go
@@ -3,7 +3,9 @@ package runtime
 // IsValidResumeType validates confirmation values coming from /resume
 func IsValidResumeType(t ResumeType) bool {
 	switch t {
-	case ResumeTypeApprove, ResumeTypeReject:
+	case ResumeTypeApprove,
+		ResumeTypeApproveSession,
+		ResumeTypeReject:
 		return true
 	default:
 		return false
@@ -14,6 +16,7 @@ func IsValidResumeType(t ResumeType) bool {
 func ValidResumeTypes() []ResumeType {
 	return []ResumeType{
 		ResumeTypeApprove,
+		ResumeTypeApproveSession,
 		ResumeTypeReject,
 	}
 }


pkg/runtime/runtime.go
```diff
@@ -724,34 +724,73 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
...
 						runtimeMaxIterations = iteration + 10
 					} else {
-						slog.Debug("User chose to exit after max iterations", "agent", a.Name())
-						// Synthesize a final assistant message so callers (e.g., parent agents)
-						// receive a non-empty response and providers are not given empty tool outputs.
+						slog.Debug("User rejected continuation", "agent", a.Name())
+
 						assistantMessage := chat.Message{
 							Role:      chat.MessageRoleAssistant,
-							Content:   fmt.Sprintf("I have reached the maximum number of iterations (%d). Stopping as requested by user.", runtimeMaxIterations),
```
krissetto, line 774: I think we need to clean up the code around Resume in general, this part here was meant to support the `max_iterations` configuration (used to avoid infinite loops in models get confused calling tools etc) 🤔

ghost: That makes sense, I agree that this is mixing UX concerns with what should be a runtime protection for max_iterations.
I can organize this and refactor accordingly. Let me know if you'd like me to share the relevant code first.

ghost: @krissetto I revised the UX language in this section to keep it purely technical.  
I didn’t make broader changes yet and am waiting for your feedback before proceeding further.

image

krissetto: This needs a rebase, there has been a change allowing users to specify a reason along with approve/deny in this API and related events. 

Sorry, lots of work being done lately :)

https://github.com/docker/cagent/pull/1444 for some context on where the changes were added

ghost: Rebase completed, submitted with force-with-lease, conflicts resolved, and tests approved.

No need to apologize, I understand and will continue to help whenever possible, and if the opportunity arises, I'd love to join the team. Thank you for your feedback and good work :) @krissetto

pkg/runtime/resume_validation.go
```diff
@@ -0,0 +1,19 @@
+package runtime
+
+// IsValidResumeType validates confirmation values coming from /resume
+func IsValidResumeType(t ResumeType) bool {
```
krissetto, line 4: Gave this another quick test. 

Watch out, when  "confirming all for the session" with `a`, the runtime blocks, we need to handle `ResumeTypeApproveSession` as well 😉

ghost: Thank you so much for that, I've already solved it. @krissetto  😄

PR #1418, opened 2026-01-19
feat(session): look for prompt files in both working dir and home folder

dgageot:
When loading prompt files, now search in both locations:
1. Working directory hierarchy (existing behavior)
2. User's home folder

If found in both locations, both are emitted as system messages. Deduplication prevents the same file from being loaded twice when the working directory search already found the file in home.

Assisted-By: cagent

### DIFF
--- a/pkg/session/prompt_file.go
+++ b/pkg/session/prompt_file.go
@@ -5,32 +5,61 @@ import (
 	"path/filepath"
 )
 
-func readPromptFile(workDir, filename string) (string, error) {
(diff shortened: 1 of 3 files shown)


krissetto:
> User's home folder

is this common enough to go outside of the usual "only access files in the working dir" policy we have?

dgageot:
> > User's home folder
> 
> is this common enough to go outside of the usual "only access files in the working dir" policy we have?

For agents.md, I think it is, yes.

PR #1423, opened 2026-01-19
Detect parent process death via stdin EOF

trungutt:
When cagent is spawned by a parent process with piped stdio, monitor stdin for EOF to detect when the parent dies. This ensures cagent shuts down gracefully instead of becoming an orphan process.

### DIFF
--- a/cmd/root/api.go
+++ b/cmd/root/api.go
@@ -53,14 +53,22 @@ func newAPICmd() *cobra.Command {
 
 // monitorStdin monitors stdin for EOF, which indicates the parent process has died.
 // When spawned with piped stdio, stdin closes when the parent process dies.
-func monitorStdin(cancel context.CancelFunc, stdin *os.File) {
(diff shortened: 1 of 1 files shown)


cmd/root/api.go
```diff
@@ -50,16 +51,44 @@ func newAPICmd() *cobra.Command {
 	return cmd
 }
 
+// monitorStdin monitors stdin for EOF, which indicates the parent process has died.
+// When spawned with piped stdio, stdin closes when the parent process dies.
+func monitorStdin(cancel context.CancelFunc, stdin *os.File) {
+	buf := make([]byte, 1)
+	for {
+		n, err := stdin.Read(buf)
```
krissetto, line 65: nit: could this block the goroutine in other scenarios like normal shutdown if we don't check if the context has already been cancelled? probably not a big issue though since the proc should die anyway in most other scenarios

trungutt: I pushed a second commit

PR #1432, opened 2026-01-19
Friendly tools

rumpl:
Adding `add_description_parameter: true` to the agent config will add a `description` paramter that the LLM can use to explain what it is doing with the tool call. Only the builtin tools that are `AddDescriptionParameter: true` will have the description added, I added them to some I thought would be nice: shell, write/edit file and 

Before
Screenshot 2026-01-19 at 23 36 07

After
Screenshot 2026-01-19 at 23 35 34

I decided to make this optional because even though it's a nice feature, it adds more output tokens.


simonferquel:
intentional_tool ? (like a tool where you can give your intention on invoking)

pkg/tools/tools.go
```diff
@@ -45,13 +45,14 @@ func ResultSuccess(output string) *ToolCallResult {
...
-	OutputSchema any             `json:"outputSchema"`
-	Handler      ToolHandler     `json:"-"`
+	Name            string          `json:"name"`
+	Category        string          `json:"category"`
+	Description     string          `json:"description,omitempty"`
+	Parameters      any             `json:"parameters"`
+	Annotations     ToolAnnotations `json:"annotations"`
+	OutputSchema    any             `json:"outputSchema"`
+	Handler         ToolHandler     `json:"-"`
+	WithDescription bool            `json:"-"`
```
simonferquel: That is confusing to have that while the tool itself (not its invocation) has a description field

rumpl: It's needed so that the wrapper knows if it has to wrap or not

simonferquel: Yes, but naming is weird: can we have `WithInvocationDescription` instead ? (I am pretty sure I will be confused about it if I have to deal with that code)

rumpl: Sure

simonferquel: Or even better, `SupportsInvocationDescription` as With... feels like a fluent builder function

rumpl: How about `WithDescriptionParameter`?

rumpl: `Supports` is kinda weird because the tool doesn't have to support anything, it just means "This is a tool that could benefit from a description parameter that the LLM can fill in to let the user know what it is doing"

simonferquel: AddDescriptionParameter ?

rumpl: `AddDescriptionParameter` it is

PR #1434, opened 2026-01-19
Don't enable thinking by default

rumpl:
This should be explicitly set on the model config


dgageot:
I agree.
@krissetto, I like the idea of making it easy on users to use advanced features but I also found it verbose and not useful for most of my use cases.

PR #1441, opened 2026-01-20
Edit tool confirmation uses vscode to show diff when cagent is run within vscode

simonferquel:
When `cagent` is run from within a vs code terminal, use the vscode diff to review edit_tool changes.

https://github.com/user-attachments/assets/f01d0013-a580-4060-b54c-85211d2ce459

### DIFF
--- a/pkg/tui/dialog/tool_confirmation.go
+++ b/pkg/tui/dialog/tool_confirmation.go
@@ -117,13 +117,15 @@ func NewToolConfirmationDialog(msg *runtime.ToolCallConfirmationEvent, sessionSt
 
 // Init initializes the tool confirmation dialog
 func (d *toolConfirmationDialog) Init() tea.Cmd {
-	defer d.sessionState.SetInConfirmationDialog()
+	undo := d.sessionState.SetInConfirmationDialog()
+	defer undo()
 	return d.scrollView.Init()
 }
 
 // Update handles messages for the tool confirmation dialog
 func (d *toolConfirmationDialog) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
-	defer d.sessionState.SetInConfirmationDialog()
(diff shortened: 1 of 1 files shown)


krissetto:
i think we should not render the diff at all in the terminal when showing in vscode, the background of that dialog seems broken

simonferquel:
I checked and have the same rendering from outside vscode. 
I have 
```
export CAGENT_EXPERIMENTAL_MARKDOWN_RENDERER=1
export CAGENT_EXPERIMENTAL_DEBUG_LAYOUT=1
```

simonferquel:
Even dumber: attached a debugger and found that the model is generating edits that... do not modify anything:
image

krissetto:
Needs a rebase then i'll approve, just gave it a test

> i think we should not render the diff at all in the terminal when showing in vscode, the background of that dialog seems broken

What i meant is also that BOTH the TUI and vscode are rendering the diff right now, which is harmless i guess :)

Screenshot From 2026-01-20 16-34-48

---

Another nit: when using vscode insiders, cagent will detect normal vscode and open that in a separate window. Also not a blocker for this, just something to think about

image

simonferquel:
yes, vscode insider and cursor detection are not easy. I'll try to find a better way to handle that.
also, I want to show that when a write_file tool is invoked and it will overwrite an existing file (for now we do not show anything in that case)

simonferquel:
also, I'd like to have a followup so that you can "always review file changes" in vscode even if you are not running within the editor

dgageot:
@simonferquel @krissetto what's the status on this one?

krissetto:
not sure i think @simonferquel wanted to implement further support, but we should probably just get vscode in for now then add more support later

dgageot:
I'm going to close it for now. @simonferquel feel free to reopen and rebase!

PR #1449, opened 2026-01-20
Better elicitation

dgageot:


krissetto:
I'm not sure this is working as intended, I can't seem to select between yes and no in the elicitation dialog.. I just gave it a quick test on linux, here's a screencast. 


[Screencast From 2026-01-20 20-28-24.webm](https://github.com/user-attachments/assets/d432e37c-4b21-492b-8f71-5f86ea61f6ba)

dgageot:
Press either y or n or space

krissetto:
we really should make that clear to users, it says nowhere that spacebar or y/n was expected 😅

PR #1452, opened 2026-01-20
People don't like when machines divide by zero

rumpl:
Also add some more default messages in the spinner


krissetto:
spinner dont spin tho

nlehuen:
I'm equally distressed by the feeding of the gremlins!

PR #1458, opened 2026-01-21
agent/acp: harden file resource link handling

Pnkcaht:
## What I did

- Added basic hardening when handling `file://` resource links in the ACP agent.
- Normalized paths with `filepath.Clean`.
- Blocked absolute paths and simple path traversal attempts before reading files.

## Related issue

N/A

## Description

Previously, file paths from ACP `ResourceLink` blocks were forwarded as-is after stripping the `file://` prefix.

This change adds lightweight, local validation to avoid sending unsafe paths (absolute paths or `..` traversal) to the ACP server.

This is defense-in-depth and does not change behavior for valid inputs.

## Result

- Unsafe paths are rejected early.
- Reduced risk of unintended file access.
- No API or behavior changes for valid use cases.

### DIFF
--- a/pkg/acp/agent.go
+++ b/pkg/acp/agent.go
@@ -6,11 +6,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
-
-	"path/filepath"
 
 	"github.com/coder/acp-go-sdk"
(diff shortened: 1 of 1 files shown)


dgageot (approved):

ghost:
I updated a few minor formatting issues to fix the GitHub lint checks, even though the PR had already been approved. @dgageot

PR #1460, opened 2026-01-21
read pdf

jeanlaurent:
## Add PDF file reading support
Adds the ability to read PDF files via the  read_file  and
read_multiple_files  tools using the  github.com/ledongthuc/pdf  library.

### Changes
• Add  ledongthuc/pdf  dependency (pure Go, zero external dependencies)
• Detect PDF files by extension and extract plain text automatically
• Handle errors gracefully for invalid or corrupted PDFs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,6 +43,7 @@ require (
 	github.com/junegunn/fzf v0.67.0
 	github.com/k3a/html2text v1.3.0
 	github.com/labstack/echo/v4 v4.15.0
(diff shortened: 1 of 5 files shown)


jeanlaurent:
Maybe what we should do instead is "automagically" attach the document and sent it to the llm?
@rumpl ?

jeanlaurent:
Closing as this is not the right approach

krissetto:
PDFs are a pain, but some models like gemini support them natively so maybe we can start with that

https://ai.google.dev/gemini-api/docs/document-processing

PR #1461, opened 2026-01-21
fade out competed tool calls in collapsed reasoning blocks

krissetto:
Proposal for improving how collapsed reasoning blocks render their tool calls

In progress tool calls are currently shown right below the collapsed reasoning block.  
Once the tool has finished executing, it used to disappear immediately from the collapsed view.  

Now, **finished tool calls will remain visible for 2.5 seconds, while fading out for the last 1.5 seconds** to indicate it's about to be removed from the collapsed view

Also adds some "bottom slack" to the scrolling of messages so the UI doesn't jump up and down if tool calls in reasoning blocks appear and then get removed at the very bottom of the viewport.

### Screencasts

> The interesting bits are at the start/end of each video

---

**before**

https://github.com/user-attachments/assets/a06d736a-733f-4a34-998f-1b496790fa6e

---

**after**

https://github.com/user-attachments/assets/ede84f33-28bb-43a9-a4ec-b9bbb028244c

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -81,6 +81,7 @@ type model struct {
 
 	// Height tracking system fields
 	scrollOffset  int                  // Current scroll position in lines
+	bottomSlack   int                  // Extra blank lines added after content shrinks
 	rendered      string               // Complete rendered content string
 	renderedItems map[int]renderedItem // Cache of rendered items with positions
 	totalHeight   int                  // Total height of all content in lines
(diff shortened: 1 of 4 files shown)


rumpl:
I love this

pkg/tui/components/messages/messages.go
```diff
@@ -185,6 +186,35 @@ func (m *model) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 		m.invalidateAllItems()
 		return m, nil
 
+	case reasoningblock.FadeTickMsg:
```
rumpl: On line 223 below we forward updates to all message views, can we do this for these two? Or is this really needed here?

krissetto: oops yeah it would've happened for the `GraceTickmsg` events, but to keep things more clear i cleaned this up putting the reasoningblock-only messages behind and interface and moved the code that handles them to `reasoningblock.go`

PR #1476, opened 2026-01-22
Add a way to provide Docker access tokens from an external command

dgageot:


pkg/environment/credential_helper.go
krissetto, line 1: nice, i think we can use a similar approach for things like aws secrets via the aws cli for example

pkg/environment/credential_helper.go
```diff
@@ -0,0 +1,46 @@
...
+		return "", false
+	}
+
+	var stdout, stderr bytes.Buffer
+	cmd := exec.CommandContext(ctx, p.command, p.args...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
+		slog.Debug("Failed to get Docker token from credential helper",
```
rumpl, line 35: Could be any token

PR #1478, opened 2026-01-22
runtime: capture token usage once per stream

### LINKED ISSUE #1475: <short bug description>
Token Usage always shows 0%

Pnkcaht:
## Context

Token usage always shows 0% in the TUI because some providers emit token usage multiple times during streaming, including partial or zero-valued snapshots.

Before this change, the runtime would overwrite previously captured usage with later snapshots, causing the final session token usage to be zero even when tokens were actually consumed.
This PR fixes token usage reporting by capturing usage once per stream and treating it as immutable, making the behavior provider-agnostic.

## References

Fixed #1475

## What I did

- Updated the runtime streaming handler to capture token usage only once
  per stream.
- Prevented later partial or zero-valued usage snapshots from overwriting
  valid usage data.
- Ensured session-level input/output token counts are populated correctly.
- Preserved existing cost calculation logic while making it stable across
  providers (OpenAI, Anthropic, Gemini, Bedrock).

### Before

- Token usage could be overwritten by later streaming events.
- Providers emitting partial usage snapshots caused final usage to be zero.
- TUI always displayed 0% token usage.

### After

- Token usage is captured once and treated as immutable per stream.
- Partial or repeated usage snapshots no longer affect final totals.
- TUI correctly displays token usage across all supported providers.


dgageot:
@Pnkcaht Could you please rebase? Is it possible to add a test that shows the issue?

ghost:
I performed a rebase and added an example to the test file, which compiles and runs correctly. @dgageot 

image

dgageot:
Closing in favour of https://github.com/docker/cagent/pull/1855

PR #1482, opened 2026-01-22
runtime: resolve instruction tool calls via standard tool flow

### LINKED ISSUE #812: Support calling a tool in instructions
It would be useful to call a tool and inject its response into the instructions.
e.g we could read `agents.md` with the `read_file` tool.

We could leverage the existing code that replaces javascript placeholders.

Exact syntax TDB.

Pnkcaht:
## Context

Instruction system messages can include inline tool placeholders using the `{{ tool:name args }}` syntax. Until now, these placeholders were not resolved through the standard tool execution flow, leading to inconsistent behavior between instruction-time tool usage and normal assistant tool calls.

This made instruction tools harder to reason about, test, and maintain, since they bypassed the same message and tool-call pipeline used everywhere else in the runtime.

## References

Fixed #812

## What I did

- Implemented instruction tool resolution using the standard tool-call flow.
- Parsed `{{ tool:... }}` placeholders from **SYSTEM** messages only.
- Removed tool placeholders from the SYSTEM message content.
- Injected an **ASSISTANT** message containing structured `ToolCall` entries.
- Ensured instruction tools follow the same execution path as regular tools.
- Avoided recursion, approval flow, and direct handler invocation.

## Why

- Align instruction-time tool execution with the existing runtime architecture.
- Reduce special-case logic and improve long-term maintainability.
- Ensure instruction tools behave identically to assistant-invoked tools.
- Make instruction tool execution easier to debug and reason about.

### Before

- Instruction tools were resolved through custom logic.
- Tool execution bypassed the standard assistant → tool pipeline.
- Behavior differed from normal tool calls.
- Harder to extend and maintain.

### After

- Instruction tools are converted into standard assistant `ToolCall`s.
- All tools run through the same execution flow.
- Consistent behavior across instruction and assistant contexts.
- Cleaner and more maintainable runtime logic.

### DIFF
--- a/pkg/runtime/instructions_tools.go
+++ b/pkg/runtime/instructions_tools.go
@@ -87,14 +87,16 @@ func (r *LocalRuntime) ResolveInstructionToolCalls(
 
 		// Emit the cleaned SYSTEM message without tool placeholders.
 		msg.Content = strings.TrimSpace(content)
-		out = append(out, msg)
 
 		// Inject an ASSISTANT message containing the generated tool calls.
 		// These will be executed by the standard runtime loop.
-		out = append(out, chat.Message{
-			Role:      chat.MessageRoleAssistant,
-			ToolCalls: toolCalls,
-		})
+		out = append(out,
+			msg,
+			chat.Message{
+				Role:      chat.MessageRoleAssistant,
+				ToolCalls: toolCalls,
+			},
+		)
 	}
 
 	return out, nil


dgageot:
This PR is a bit stale so I'm going to close it for now.
Sorry @Pnkcaht for maybe not giving you our full attention on this one. Feel free to reopen if you think we should really reconsider.

PR #1495, opened 2026-01-24
improve message resizing for pending responses

krissetto:
Avoid breaking the TUI layout when the pendingSpinner appears at the very bottom of the messages area

Screencasts:

**before**

https://github.com/user-attachments/assets/d040f04e-4040-4027-9596-7956612ddc84

---

**after**

https://github.com/user-attachments/assets/c609b535-c193-4893-9bb1-8d522f3f7223

### DIFF
--- a/pkg/tui/page/chat/chat.go
+++ b/pkg/tui/page/chat/chat.go
@@ -530,19 +530,35 @@ func (p *chatPage) setWorking(working bool) tea.Cmd {
 
 // setPendingResponse sets the pending response state.
 // When true, a spinner is shown below the messages while waiting for the first chunk.
+// Resizes messages to leave room for the spinner; bottomSlack smooths the transition.
 func (p *chatPage) setPendingResponse(pending bool) tea.Cmd {
 	wasPending := p.pendingResponse
 	p.pendingResponse = pending
+
 	if pending && !wasPending {
-		// Starting to wait - register spinner
+		// Starting to wait - register spinner and resize messages to account for spinner space
+		sl := p.computeSidebarLayout()
+		messagesHeight := sl.chatHeight - pendingSpinnerHeight
+		resizeCmd := p.messages.SetSize(sl.chatWidth, max(1, messagesHeight))
+
 		p.pendingSpinner = p.pendingSpinner.Reset()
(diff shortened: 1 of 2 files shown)


pkg/tui/page/chat/chat.go
```diff
@@ -421,14 +421,47 @@ func (p *chatPage) setWorking(working bool) tea.Cmd {
...
+		mainWidth = max(innerWidth, 1)
+	}
+
+	messagesHeight := p.chatHeight
+	if p.pendingResponse {
+		messagesHeight -= pendingSpinnerHeight
+	}
+
+	return mainWidth, max(1, messagesHeight)
 }
```
ghost: The code itself is okay, could you add comments above the function? I know it's silly, but it helps anyone who reads and works with it later.    

Something like:

```golang
// messagesSize returns the usable width and height for the messages viewport
// reserving sidebar space when the window is wide enough and reducing height 
// for the pending response spinner.
```

krissetto: Comments on code that is too simple adds maintenance burden, as comments go out of date if they don't get updated with the code, creating confusion, and the build/lint won't properly catch that in CI. 

Code is the only source of truth for proper understanding. Comments should only really go where necessary to explain things that are hard to understand from the code, or to explain the reason for specific behaviors and avoid risking that someone else in the future changes things not knowing why they exist

The func signature and the return statement make it pretty clear what happens, so I don't think extra comments are worth it here. It's also not library code where extra explanations could be useful for third parties consuming the code

ghost: I agree, the request for change was focused on clarity in the code and for the user (new contributors as well, for example), but in any case, I agree about the maintenance and the confusion itself, right, thank you. ❤️  :)

krissetto:
@Pnkcaht btw,there's no need to approve PRs, we have policies in place that allow merging code only when approved maintainers approve the code :) 

When anyone else approves a PR, it just sends out un-actionable notifications

ghost:
@krissetto Okay, okay, I didn't know, sorry about that, I just wanted to help  :)

krissetto:
rebased on main

PR #1500, opened 2026-01-25
Misc improvements to the fast renderer

krissetto:
Includes fixes for various rendering issues, along with improvements to general styling of certain elements to improve legibility in the TUI.

Here's a non-exhaustive list of the rendering improvements:
- more consistent spacing in general
- heading styles are more consistent, and compatible with italics, code blocks, strike-through, etc.
- word wrapping lines
- styling after an inline code block
- table styling
- things we previously didn't render like nested blockquotes
- code blocks in lists are correctly indented
- ordered/unordered lists have the same styling
- etc... see the videos  :point_down: 

After these changes, rendering performance improves slightly (in both speed and memory) for everything (currently benchmarked) except tables, which show a small regression due to the styling fixes. They should be meaningless in real-worlds scenarios.

Currently tested on Linux with ghostty and ptyxis, and MacOS using Ghostty, iTerm2 and the built-in Terminal

### Screencasts

**before**

https://github.com/user-attachments/assets/29a35b6f-c4fe-4e95-bb50-af8d943136f8

---

**after**

https://github.com/user-attachments/assets/76d3b07f-feaf-42b8-8405-347cb9f195c4

### DIFF
--- a/pkg/tui/components/markdown/fast_renderer.go
+++ b/pkg/tui/components/markdown/fast_renderer.go
@@ -19,9 +19,13 @@ import (
 
 // ansiStyle holds pre-computed ANSI escape sequences for fast rendering.
 // This avoids the overhead of lipgloss.Style.Render() which copies large structs.
+// Flags track formatting attributes to avoid string scanning in hot paths.
 type ansiStyle struct {
-	prefix string // ANSI codes to start the style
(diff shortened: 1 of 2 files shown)


krissetto:
Also did some testing on MacOS using Ghostty, iTerm2 and the built-in Terminal

PR #1516, opened 2026-01-28
Pass a custom base image to cagent eval

dgageot:


pkg/evaluation/Dockerfile.template
```diff
@@ -23,13 +23,12 @@ exec "$@"
 EOF
 RUN chmod +x /run.sh
 
-FROM docker/cagent:edge AS cagent
-
 FROM scratch
 COPY --from=dind / /
-COPY --from=cagent /cagent /
+COPY --from=docker/cagent:edge /cagent /
```
krissetto, line 28: why edge and not latest?

krissetto (approved):

pkg/evaluation/Dockerfile.template
dgageot: I think we push to edge on the CI. I'll check

PR #1520, opened 2026-01-28
Fix the session compaction

rumpl:
We were being stupid, the main session calls sess.GetMessages, which potentially trims things, either the number of messages if configured that way, or by removing old tool call results. But then we would use GetAllMessages when compacting, resulting in a compaction that doesn't work because the context that we are sending is potentially bigger than the LLMs max tokens limit.

### DIFF
--- a/pkg/runtime/prompts/compaction-system.txt
+++ b/pkg/runtime/prompts/compaction-system.txt
@@ -1,0 +1,4 @@
+You are an AI assistant that helps summarize conversations for future reference. 
+
+Your task is to generate a concise summary of the conversation that captures the key points, decisions made, and important context. 
+The summary should be clear and informative, enabling someone who was not part of the original conversation to understand what transpired.
--- a/pkg/runtime/session_compaction.go
+++ b/pkg/runtime/session_compaction.go
@@ -2,31 +2,23 @@ package runtime
 
 import (
 	"context"
-	"fmt"
+	_ "embed"
 	"log/slog"
-	"strings"
+	"time"
 
 	"github.com/docker/cagent/pkg/agent"
+	"github.com/docker/cagent/pkg/chat"
 	"github.com/docker/cagent/pkg/model/provider"
 	"github.com/docker/cagent/pkg/model/provider/options"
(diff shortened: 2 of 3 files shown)


pkg/runtime/session_compaction.go
```diff
@@ -46,72 +38,64 @@ func (c *sessionCompactor) Compact(ctx context.Context, sess *session.Session, a
...
-
-	conversationHistory := c.buildConversationHistory(messages)
-	userPrompt := c.buildUserPrompt(conversationHistory, additionalPrompt)
+	summaryModel := provider.CloneWithOptions(ctx, c.model, options.WithStructuredOutput(nil))
+	root := agent.New("root", compactionSystemPrompt, agent.WithModel(summaryModel))
+	newTeam := team.New(team.WithAgents(root))
 
-	summary := c.generateSummary(ctx, userPrompt)
-	if summary == "" {
+	messages := sess.GetMessages(root)
```
krissetto, line 45: mmh this is a bit confusing to me..

Looking more closely at sess.GetMessages(), won't this get the system messages from the root compaction agent created 2 lines above, instead of the ones from the actual agent the user was talking to, by running `buildInvariantSystemMessages` and `buildContextSpecificSystemMessages`, since those gather the messages from the agent and not from the session?

Then the session summary messages and actual messages from the session seems to be retrieved correctly, but this agent/session mix is a bit weird 🤔

pkg/runtime/prompts/compaction-system.txt
```diff
@@ -0,0 +1,4 @@
+You are an AI assistant that helps summarize conversations for future reference. 
+
+Your task is to generate a concise summary of the conversation that captures the key points, decisions made, and important context.
```
krissetto, line 3: nit: Here we say "generate a concise summary", below in the user msg we say "provide a detailed prompt". I think we can make these 2 messages a bit more coherent with one another and get more reliable compaction results as a result, without confusing the model with somewhat opposite requests

dgageot:
Closing in favour of https://github.com/docker/cagent/pull/1531

PR #1521, opened 2026-01-28
Make /commands searchable in the command palette

krissetto:
If a users searches for a slash command command with the / prefix, for muscle memory of because of their expectations, lets show them what they want to see


rumpl (approved):
sure

PR #1526, opened 2026-01-29
Simpler cagent eval

dgageot:


cmd/root/eval.go
```diff
@@ -43,11 +37,12 @@ func newEvalCmd() *cobra.Command {
 	}
 
 	addRuntimeConfigFlags(cmd, &flags.runConfig)
-	cmd.Flags().IntVarP(&flags.concurrency, "concurrency", "c", 0, "Number of concurrent evaluation runs (0 = number of CPUs)")
-	cmd.Flags().StringVar(&flags.judgeModel, "judge-model", defaultJudgeModel, "Model to use for relevance checking (format: provider/model)")
+	cmd.Flags().IntVarP(&flags.Concurrency, "concurrency", "c", runtime.NumCPU(), "Number of concurrent evaluation runs")
```
krissetto, line 40: i realize this is not a change, but are we sure about defaulting concurrency to num of cpus? this can affect cost a ton on machines with lots of cores/SMT

PR #1530, opened 2026-01-29
tui: Handle Ctrl+z

vvoland:
Adds support for suspending the TUI application to the background using Ctrl+z, which is the standard Unix convention for job control.

This allows users to temporarily return to their shell without fully exiting the application, making it easier to run quick shell commands or check other processes while keeping their TUI session active.


rumpl:
FYI you can also directly in cagent do `!cat README.md` and it will run that command

rumpl (approved):

vvoland:
Thanks, I need my full shell though 😅

rumpl:
And you shall have it

PR #1534, opened 2026-01-30
Permissions view dialog

stanislavHamara:
image

### DIFF
--- a/pkg/tui/dialog/permissions.go
+++ b/pkg/tui/dialog/permissions.go
@@ -137,9 +137,6 @@ func (d *permissionsDialog) renderContent(contentWidth, maxHeight int) string {
 		}
 	}
 
-	// Add help about evaluation order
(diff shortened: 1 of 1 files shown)


krissetto:
The "Evaluation" line at the bottom of the dialog is a bit confusing IMHO, i wonder if we can make things clearer for the users

Also it'd be better if we used the scrollbar for ux consistency

stanislavHamara:
> The "Evaluation" line at the bottom of the dialog is a bit confusing IMHO, i wonder if we can make things clearer for the users
> 
> Also it'd be better if we used the scrollbar for ux consistency

The scrollbar is there, I guess the user can do both? 🤷

krissetto:
> The scrollbar is there, I guess the user can do both?

oh, my bad. i got confused as i'm looking into other inconsistencies we have in dialogs and the up/down arrow threw me off :)

PR #1539, opened 2026-01-30
Add toolset to switch model

dgageot:


pkg/tools/builtin/switch_model.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Instructions returns guidance for when to use model switching.
+func (t *SwitchModelToolset) Instructions() string {
+	return `## Model Switching Guidelines
+
+You have access to multiple AI models and can switch between them strategically.
+
+### When to Consider Switching Models
+
+**Switch to a faster/cheaper model when:**
```
krissetto, line 80: just one thing here.. how is a given model supposed to know what is faster/cheaper or better at some task? should we just rely on the users giving their model definitions reasonable name?

examples/switch_model.yaml
```diff
@@ -0,0 +1,40 @@
...
+      Use switch_model strategically:
+      - For simple tasks (formatting, basic Q&A, summaries), use the faster/cheaper model
+      - For complex tasks (code generation, analysis, reasoning), use the more powerful model
+      - The switch_model tool description shows available models and which one is current
+      - After completing a specialized task, consider switching back to the default model
+      
+      Example workflow:
+      1. User asks a simple question -> switch to 'fast' for efficiency
+      2. User asks for complex code -> switch to 'powerful' for quality
+      3. Task complete -> switch back to 'fast'
```
krissetto, line 34: do we expect the example work decently even without this extra prompting, given all the instructions included in the tool? if not maybe we can reduce the amount of instructions, to avoid potential conflicts between user prompts and built-in instructions the users don't see?

mostly thinking about context pollution with smaller/local models and conflicting prompts

PR #1549, opened 2026-01-31
  Add Linux installation script

jeanlaurent:
Adds a one-line installer for Linux users similar to Homebrew and Claude Code's installation experience.

Before: Linux users had to manually download binaries, chmod, rename, and configure PATH.

After: Single command installation with automatic setup and verification.

### DIFF
--- a/scripts/install-linux.sh
+++ b/scripts/install-linux.sh
@@ -1,0 +1,199 @@
+#!/usr/bin/env bash
+set -e
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# Script info
+REPO="docker/cagent"
+BINARY_NAME="cagent"
+INSTALL_DIR="/usr/local/bin"
+
+# Print colored message
+print_msg() {
+    local color=$1
+    shift
+    echo -e "${color}$@${NC}" >&2
+}
+
+# Detect architecture
+detect_arch() {
+    local arch
+    arch=$(uname -m)
+
+    case "$arch" in
+        x86_64|amd64)
+            echo "amd64"
+            ;;
+        aarch64|arm64)
+            echo "arm64"
+            ;;
+        *)
+            print_msg "$RED" "Error: Unsupported architecture: $arch"
+            print_msg "$YELLOW" "Supported architectures: x86_64 (amd64), aarch64 (arm64)"
+            exit 1
+            ;;
+    esac
+}
+
+# Detect OS
+detect_os() {
+    local os
+    os=$(uname -s)
+
+    case "$os" in
+        Linux)
+            echo "linux"
+            ;;
+        Darwin)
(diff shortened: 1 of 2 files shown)


scripts/install-linux.sh
```diff
@@ -0,0 +1,199 @@
...
+    # Download binary
+    print_msg "$BLUE" "Downloading from $download_url..."
+    if ! curl -fsSL "$download_url" -o "$tmp_file"; then
+        print_msg "$RED" "Error: Failed to download binary"
+        exit 1
+    fi
+
+    # Make executable
+    chmod +x "$tmp_file"
+
```
dvdksn, line 120: We can do this with `install` later.
```suggestion
```

scripts/install-linux.sh
```diff
@@ -0,0 +1,199 @@
...
+        fi
+    else
+        use_sudo=false
+    fi
+
+    # Install binary
+    local target_path="$target_dir/$BINARY_NAME"
+    if [ "$use_sudo" = true ]; then
+        sudo mv "$tmp_file" "$target_path"
+        sudo chmod +x "$target_path"
```
dvdksn, line 144:
```suggestion
        sudo install -m 755 "$tmp_file" "$target_path"
```

scripts/install-linux.sh
```diff
@@ -0,0 +1,199 @@
...
+    else
+        mv "$tmp_file" "$target_path"
```
dvdksn, line 146:
```suggestion
        install -m 755 "$tmp_file" "$target_path"
```

scripts/install-linux.sh
```diff
@@ -0,0 +1,199 @@
...
+    esac
+}
+
+# Get latest release version
+get_latest_version() {
+    print_msg "$BLUE" "Fetching latest version..."
+
+    # Try using gh CLI first if available
+    if command -v gh &> /dev/null; then
+        gh release list --repo "$REPO" --limit 1 --json tagName --jq '.[0].tagName' 2>/dev/null
```
krissetto, line 83: i don't think we need both options here, i'd just go straight to the rest api unless there are other reasons for using `gh`

scripts/install-linux.sh
```diff
@@ -0,0 +1,199 @@
...
+        Darwin)
+            print_msg "$RED" "Error: This script is for Linux only."
+            echo "" >&2
+            print_msg "$YELLOW" "For macOS, you have two options:"
+            print_msg "$YELLOW" "  1. Download Docker Desktop: https://www.docker.com/products/docker-desktop"
+            print_msg "$YELLOW" "     (cagent is included starting with version 4.49.0)"
+            print_msg "$YELLOW" "  2. Use Homebrew: brew install cagent"
+            exit 1
+            ;;
+        MINGW*|MSYS*|CYGWIN*)
```
krissetto, line 61: nit: harmless, i don't think many windows users will try to run a bash script called install-linux

dgageot:
@jeanlaurent what's the decision on this one?

dgageot:
Let's close this for now

PR #1553, opened 2026-02-01
Cleaner agent switching ux in the TUI

krissetto:
Give it a try and let me know if this is worth a cleanup pass

:warning: Ignore the code for now, plenty of vibes and it has not been thoroughly looked into.

- Makes the sidebar agents section its own component (as an example for future refactoring of all sections)
- Cleans up agent view in the sidebar to only show essential agent info (name and model)
- Allows switching by clicking on the agent in the sidebar
- Adds proper agent selection dialog with more info per agent and easier to read descriptions, including expandable toolset/tool list

### Screencast

https://github.com/user-attachments/assets/c4c1d5ab-d2d8-454a-9f77-f0e652b0f0b9

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -114,6 +114,17 @@ func New(ctx context.Context, rt runtime.Runtime, sess *session.Session, opts ..
 		})
 	}
 
+	// If the runtime supports background toolset loading, start it to load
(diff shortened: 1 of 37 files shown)


krissetto:
eh, stale, will re-do better later on

PR #1555, opened 2026-02-02
Make TUI more responsive when scrolling

krissetto:
Coalesce mouse scroll events so we don't backlog the render pipeline with a ton of events.

This improves CPU performance when scrolling with a mouse/trackpad by 8-12x, while making the TUI **much** more responsive, even on large high res screens and with many long messages in the session. Where the scroll rendering used to lag behind the mouse scroll, continuing well after the user stopped scrolling, now its way more reactive to user input

Also includes:
- caching rendered output in lines instead of a single string, avoiding split/joins on every render and reducing cpu usage by roughly 15-20% and mem usage by 25-50% for message rendering (based on number of messages etc)
- caching sidebar render to avoid needless redraws

### Screencasts

**before**
> :warning: When i wiggle the mouse, i have already stopped scrolling, but the TUI is lagging behind even on a very fast cpu

https://github.com/user-attachments/assets/b6e05313-1dff-4675-9a15-c658f6a1b692

**after**
> scroll rendering stops as soon as i stop scrolling

https://github.com/user-attachments/assets/490c437a-9c0b-4cb7-8cf3-0f6599601482

### DIFF
--- a/pkg/tui/input/wheel_coalescer.go
+++ b/pkg/tui/input/wheel_coalescer.go
@@ -1,0 +1,83 @@
+package input
+
+import (
+	"sync"
+	"time"
+
+	tea "charm.land/bubbletea/v2"
+
+	"github.com/docker/cagent/pkg/tui/messages"
+)
+
+const wheelFlushInterval = 16 * time.Millisecond
+
+// WheelCoalescer aggregates wheel events into a single delta per flush interval.
+type WheelCoalescer struct {
+	mu        sync.Mutex
+	pending   int
+	lastX     int
+	lastY     int
+	scheduled bool
+	send      func(tea.Msg)
+}
+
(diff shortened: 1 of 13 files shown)


pkg/tui/input/wheel_coalescer.go
```diff
@@ -0,0 +1,83 @@
...
+
+	tea "charm.land/bubbletea/v2"
+
+	"github.com/docker/cagent/pkg/tui/messages"
+)
+
+const wheelFlushInterval = 16 * time.Millisecond
+
+// WheelCoalescer aggregates wheel events into a single delta per flush interval.
+type WheelCoalescer struct {
```
rumpl, line 15: We could use `tea.WithFilter` from bubbletea for this

krissetto: ja its being used :)

PR #1558, opened 2026-02-02
Always permissions for tools commands

stanislavHamara:
## Summary

- Adds a new "always allow this tool" option (`T`) to the tool confirmation dialog, allowing users to permanently approve specific tools or commands for the session
- For shell commands, creates granular permission patterns (e.g., `shell:cmd=ls*`) that auto-approve the specific command while still requiring confirmation for other commands
- Exposes the new `approve-tool` confirmation type via HTTP and ConnectRPC APIs

## Details

When a tool requires confirmation, users now have 4 options:
- `Y` - approve this call once
- `N` - reject with optional reason
- `T` - always allow this tool/command for the session
- `A` - approve all tools (YOLO mode)

The `T` option intelligently builds permission patterns:
- For `shell` tool running `ls -la`: adds `shell:cmd=ls*` to session permissions
- For other tools like `read_file`: adds the tool name directly

These patterns leverage the existing permissions system and are persisted with the session.

### API

The `/resume` endpoint now accepts a new confirmation type:

```json
{
  "confirmation": "approve-tool",
  "tool_name": "shell:cmd=ls*"
}
```

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -1448,15 +1448,7 @@ func (r *LocalRuntime) executeWithApproval(
 			if sess.Permissions == nil {
 				sess.Permissions = &session.PermissionsConfig{}
 			}
-			// Check if already in allow list to avoid duplicates
-			alreadyAllowed := false
-			for _, t := range sess.Permissions.Allow {
-				if t == approvedTool {
-					alreadyAllowed = true
-					break
-				}
-			}
-			if !alreadyAllowed {
+			if !slices.Contains(sess.Permissions.Allow, approvedTool) {
 				sess.Permissions.Allow = append(sess.Permissions.Allow, approvedTool)
 			}
 			slog.Debug("Resume signal received, approving tool permanently", "tool", approvedTool, "session_id", sess.ID)
--- a/pkg/tui/dialog/tool_confirmation.go
+++ b/pkg/tui/dialog/tool_confirmation.go
@@ -98,10 +98,8 @@ func (d *toolConfirmationDialog) alwaysAllowHelpText() string {
 	toolName := d.msg.ToolCall.Function.Name
 
 	// For shell with a command pattern, show a more descriptive label
-	if toolName == "shell" && strings.Contains(pattern, ":cmd=") {
-		// Extract just the command part for display: "shell:cmd=ls*" -> "ls*"
-		if idx := strings.Index(pattern, ":cmd="); idx != -1 {
-			cmdPattern := pattern[idx+5:] // skip ":cmd="
+	if toolName == "shell" {
+		if _, cmdPattern, ok := strings.Cut(pattern, ":cmd="); ok {
 			return "always allow " + cmdPattern
 		}
 	}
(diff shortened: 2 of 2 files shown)


pkg/runtime/runtime.go
```diff
@@ -1432,6 +1439,28 @@ func (r *LocalRuntime) executeWithApproval(
...
+			// Add the tool to session's allow list for future auto-approval
+			approvedTool := req.ToolName
+			if approvedTool == "" {
+				approvedTool = toolName
+			}
+			if sess.Permissions == nil {
+				sess.Permissions = &session.PermissionsConfig{}
+			}
+			// Check if already in allow list to avoid duplicates
+			alreadyAllowed := false
```
krissetto: we can use smth like this to simplify

```go
if !slices.Contains(sess.Permissions.Allow, approvedTool) {
    sess.Permissions.Allow = append(sess.Permissions.Allow, approvedTool)
}
```

pkg/tui/dialog/tool_confirmation.go
```diff
@@ -85,11 +90,31 @@ func (d *toolConfirmationDialog) renderSeparator(contentWidth int) string {
...
+	toolName := d.msg.ToolCall.Function.Name
+
+	// For shell with a command pattern, show a more descriptive label
+	if toolName == "shell" && strings.Contains(pattern, ":cmd=") {
+		// Extract just the command part for display: "shell:cmd=ls*" -> "ls*"
+		if idx := strings.Index(pattern, ":cmd="); idx != -1 {
+			cmdPattern := pattern[idx+5:] // skip ":cmd="
+			return "always allow " + cmdPattern
+		}
+	}
```
krissetto: We could simplify to smth like this i think
```go
if toolName == "shell" {
    if _, cmdPattern, ok := strings.Cut(pattern, ":cmd="); ok {
        return "always allow " + cmdPattern
    }
}
```

One Q: do we expect to ever have multiple patterns here? at the moment we only extract the first

pkg/tui/dialog/tool_confirmation.go
```diff
@@ -107,9 +132,45 @@ func defaultToolConfirmationKeyMap() toolConfirmationKeyMap {
...
+				// Create pattern: shell:cmd=<command>*
+				// The trailing * allows matching with any arguments
+				return toolName + ":cmd=" + firstWord + "*"
+			}
+		}
+	}
+
+	// For other tools, just return the tool name
+	return toolName
+}
```
krissetto, line 166: i think we can use strings.Fields and do smth like this to also catch '\n' edge cases, WDYT? 

```go
func buildPermissionPattern(toolCall tools.ToolCall) string {
toolName := toolCall.Function.Name

  if toolName == "shell" {
    var args struct {
      Cmd string `json:"cmd"`
    }
    if err := json.Unmarshal([]byte(toolCall.Function.Arguments), &args); err == nil {
      if fields := strings.Fields(args.Cmd); len(fields) > 0 {
        return toolName + ":cmd=" + fields[0] + "*"
      }
    }
  }

  return toolName
}

pkg/tui/dialog/tool_confirmation.go
stanislavHamara: nope, just one at the moment.

PR #1565, opened 2026-02-02
Automatically run PR reviewer if author is org member

derekmisler:
Add automatic PR review for Docker employees when PRs are marked as "ready for review".

### Changes

- Add `pull_request_target` trigger for `ready_for_review` and `opened` events
- Add new `auto-review` job that:
  - Checks if the PR author is a Docker org member using the GitHub API
  - Only proceeds with the review if the author is an internal employee
  - Supports fork-based workflow

### How It Works

| Scenario | Behavior |
|--------|--------|
| Docker employee opens/readies a PR | ✅ Auto-review runs |
| External contributor opens a PR | ⏭️ Skipped (not a Docker org member) |
| Trusted contributor comments /review | ✅ Manual review runs | 
| External contributor comments /review | ❌ Blocked by cagent-action auth check |

### Security Considerations

- Uses `pull_request_target` to access secrets for fork PRs
- This is safe because `review-pr` only reads files for static analysis, it doesn't execute any code from the PR
- Org membership check runs before any PR code is checked out
- Requires `ORG_MEMBERSHIP_TOKEN` secret (classic PAT with `read:org` scope) (this is already set up)

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -5,6 +5,9 @@ on:
     types: [created]
   pull_request_review_comment:
     types: [created]
+  # Auto-trigger when PR becomes ready for review (supports forks)
+  pull_request_target:
+    types: [ready_for_review, opened]
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-review.yml
```diff
@@ -13,7 +16,69 @@ permissions:
...
+      - name: Checkout PR head
+        if: steps.membership.outputs.is_member == 'true'
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Run PR Review Team
+        if: steps.membership.outputs.is_member == 'true'
+        uses: docker/cagent-action/review-pr@latest
```
krissetto, line 74: fairly minor, but should we consider pinning a sha here just so in case the upstream action gets compromised somehow we're not immediately affected? (same for anyone else using the action)

derekmisler: I thought about that, but it's our own action, so it seemed safe-ish?

PR #1572, opened 2026-02-02
add optional Bearer token authorization for SSE endpoints

Pnkcaht:
## What I did

- Added **optional Bearer token authorization** to SSE endpoints.
- Introduced a server-side check for the `Authorization: Bearer <token>` header.
- The authorization layer is **disabled by default** and only enforced when a token is configured.
- Requests without a valid token are rejected before any SSE data or tool execution is exposed.

## Before

- SSE endpoints were **fully open**.
- Any client that could reach the network port could:
  - Connect to the SSE stream
  - Receive tool definitions
  - Trigger command execution
- The security model implicitly assumed a **trusted local environment**.

## After

- SSE endpoints can be protected with a **simple Bearer token**.
- The server validates the token **before**:
  - Establishing the SSE connection
  - Exposing tool definitions
  - Executing any command
- Default behavior remains unchanged when no token is configured.
- This provides a basic but effective security layer for:
  - Shared servers
  - Cloud deployments
  - Internal team usage

### DIFF
--- a/cmd/root/api.go
+++ b/cmd/root/api.go
@@ -49,6 +49,8 @@ func newAPICmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&flags.exitOnStdinEOF, "exit-on-stdin-eof", false, "Exit when stdin is closed (for integration with parent processes)")
 	_ = cmd.PersistentFlags().MarkHidden("exit-on-stdin-eof")
 	cmd.MarkFlagsMutuallyExclusive("fake", "record")
+	cmd.PersistentFlags().StringVar(&flags.runConfig.APIToken, "token", "", "Bearer token required to access SSE endpoints")
+
 	addRuntimeConfigFlags(cmd, &flags.runConfig)
 
 	return cmd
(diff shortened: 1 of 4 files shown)


ghost:
Hello again friends, @rumpl & @krissetto 

I'm not sure if this specific PR fits the project scope and will be accepted.  The idea came from an open issue on the MCP Gateway, which I also plan to work on separately.

That said, I wanted to share this as a small, optional improvement at the SSE transport level.  
If this doesn't make sense here, I'm totally fine modifying it or closing the PR :)

krissetto:
Interesting idea, but this implementation is:

- incomplete (not all endpoints covered);
- not really secure (token passed via cli flag so visible in `ps aux` etc, no limit on failed attempts so brute-forceable, etc.)

So it's probably better to close this for now since we don't want to give users a false sense of security. We'll coordinate with the mcp gateway team to get things like this done properly in a way we both support

ghost:
> Interesting idea, but this implementation is:
> 
>     * incomplete (not all endpoints covered);
> 
>     * not really secure (token passed via cli flag so visible in `ps aux` etc, no limit on failed attempts so brute-forceable, etc.)
> 
> 
> So it's probably better to close this for now since we don't want to give users a false sense of security. We'll coordinate with the mcp gateway team to get things like this done properly in a way we both support

Okay, I totally agree, thank you :)

PR #1578, opened 2026-02-03
Branching sessions on message edit

krissetto:
Allows users to edit a previous message, which will cause the session to branch off on the spot so no messages in the original session get lost

The newly created session gets a parent session id and position of the message it was branched off from, allowing for future features that may be interesting to have (tracing a session tree, branching off a question to multiple different models/agents at the same time, etc.)

### Screencast

https://github.com/user-attachments/assets/a217b8f7-feb7-401e-8d94-258cc8acdadf


vvoland (approved):
It works!

krissetto:
@vvoland what do you think about the UX? any hard desire for the session **not** to branch? i think this is safer as a default. Or any other thoughts? :)

vvoland:
UX-wise, there are issues imo:

- The edit button is hard to spot (I wouldn't know about this feature without your demo video)
- Having to use mouse - would be nice to have a keyboard shortcut to jump to the previous/next message and then hit a key to edit the message.

As for branching itself - no hard opinion, but personally I'm fine with just discarding the conversation after the edited message. I don't think I would have the brain capacity to think about different branches 😅

krissetto:
For the edit button, i was thinking we could reuse the message selection code that already exists when pressing TAB, which now only selects assistant messages for copying them but could easily extend to user messages where then pressing `e` could start the editing.. will TAL at that

Totally agree the button can be hard to spot, its not easy to make things both visible and not too ugly in a terminal interface (and font choice also affects what you see)

Branching was more something i wanted to have because:
- 1 we don't lose history. 
- 2 many times in different editors i wanted to see and/or compare results after changing 1 message in a conversation, or maybe asking a different model, etc
- 3 allows for other neat features where you can e.g. branch to 2-3 agents and then converge. potentially, maybe :joy:

damejeras:
Consider adding a possibility to branch out without touching a mouse. The tool already supports some of vim style navigation, which I find really convenient. Maybe something like using Tab to switch to messages panel and `e` to edit a message that previously was submitted. Though that would require changing how navigation works, as  I think j/k in message panel selects only assistant messages. That is why I proposed trimming messages up to selected assistant message in #1586 which is now closed in favor of this PR.

krissetto:
@vvoland && @damejeras 

You can now edit via the keyboard as well, give it a try :) 

I also fixed/cleanup up some minor details while i was there (e.g. msg selection styling will be restored when exiting the selection mode)

PR #1586, opened 2026-02-04
(feat) Add fork support for TUI

damejeras:
Summary

  - Add conversation fork feature via f keybinding when a message is selected
  - Allows users to truncate conversation history and continue from an earlier point

  Details

  When viewing a conversation, users can now:
  1. Press Tab to focus the messages panel
  2. Navigate with j/k to select an assistant message
  3. Press f to fork the session from that point

  The session is truncated to include everything up to and including the selected message, allowing the user to ask a different follow-up question.

  Test plan

  - Start a conversation with multiple exchanges
  - Press Tab to focus messages, navigate to an earlier assistant response
  - Press f to fork
  - Verify notification "Session forked." appears
  - Verify messages after the selected point are removed
  - Verify you can type a new message and continue from the fork point

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -368,6 +368,12 @@ func (m *model) handleKeyPress(msg tea.KeyPressMsg) (layout.Model, tea.Cmd) {
 			return m, cmd
 		}
 		return m, nil
+	case "f":
+		if m.focused && m.selectedMessageIndex >= 0 {
+			cmd := m.forkFromSelectedMessage()
+			return m, cmd
+		}
+		return m, nil
 	case "pgup":
 		m.scrollPageUp()
 		return m, nil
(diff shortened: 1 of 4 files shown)


krissetto:
Watch out, this implementation skips a bunch of things we need.   
Implemented as is, this is a "TUI only" feature that does not actually fork a session. It removes the messages only on the TUI side of things, it does not remove anything from the persisted session nor does it fork the session in any way. If you reload the session, all the messages including those you thought had been removed are still present, in their original order, making subsequent use of the session not behave in the same way as the user might expect.

The concept is similar to what i implemented in #1578, maybe we can iterate there if you have suggestions. 
e.g. I will be adding the ability to edit/branch off from the keyboard as well (now the edit icon is only clickable via mouse).

damejeras:
Im closing this then. I just wanted to propose something I found useful. This allows me to save some tokens in my workflows. I could totally use the feature proposed in #1578.

PR #1590, opened 2026-02-04
Add a timestamp to runtime events

dgageot:


pkg/evaluation/save.go
```diff
@@ -301,6 +313,19 @@ func parseMessageUsage(m map[string]any) *chat.Usage {
...
+// Returns the timestamp string, falling back to current time if not present or invalid.
+func parseEventTimestamp(event map[string]any) string {
+	if ts, ok := event["timestamp"].(string); ok && ts != "" {
+		// Validate RFC3339 format
+		if _, err := time.Parse(time.RFC3339, ts); err == nil {
+			return ts
+		}
+		// Invalid timestamp format - fall back to current time
+	}
+	return time.Now().Format(time.RFC3339)
```
krissetto, line 326: could this mess the the order of things? 🤔

dgageot: not really. This is for evals. Not super dangerous even if it did

PR #1619, opened 2026-02-05
Update Nightly Scan Workflow

derekmisler:
## Overview

This PR makes several updates to the nightly scan workflow (`.github/workflows/nightly-scan.yml`) to improve reliability, correct scheduling, and satisfy YAML linting requirements.

### Cron Schedule Update
- The scheduled workflow time has been updated from 6am UTC to 2am UTC.
- This is reflected by changing the cron expression from:

### Workflow Input Description Quote Standardization
- Updated the input description for the dry-run option to use double quotes for consistency.

### Job Timeout Addition
- A job-level timeout (timeout-minutes: 30) has been added to the scan job. This prevents the job from running indefinitely if something goes wrong.

### Prompt Expression Fix to Satisfy YAML Linter
- The prompt input for the docker/cagent-action step was triggering YAML linter errors due to nested mappings and incorrect type expectations.
- The expression has been wrapped in double quotes, and bracket notation is now used for the hyphenated property name:

## Testing

- The changes have been validated locally via the IDE's YAML linter.
- The updated cron expression and timeout addition have been deployed to ensure the workflow runs as expected.
- The prompt expression now parses correctly, resolving the linter's concerns regarding nested mappings and type expectations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,12 @@ jobs:
         with:
           version: v2.8
 
+      - name: Lint GitHub Actions
+        uses: raven-actions/actionlint@e01d1ea33dd6a5ed517d95b4c0c357560ac6f518 # v2.1.1
+        with:
+          fail-on-error: true
+          pyflakes: false
+
   test:
     runs-on: ubuntu-latest
     steps:
@@ -90,7 +96,7 @@ jobs:
           version: 3.x
 
       - name: Create bin directory
-        run: mkdir -p $HOME/bin
+        run: mkdir -p "$HOME/bin"
 
       - name: Build
         run: task build


derekmisler:
We have a yaml linter in cagent-action that runs in CI, I wonder if it's a good idea here, too

krissetto:
> We have a yaml linter in cagent-action that runs in CI, I wonder if it's a good idea here, too

@derekmisler We could integrate one into the `task lint` command which already runs in CI, extending it for linting workflows etc as well. WDYT?

derekmisler:
> > We have a yaml linter in cagent-action that runs in CI, I wonder if it's a good idea here, too
> 
> 
> 
> @derekmisler We could integrate one into the `task lint` command which already runs in CI, extending it for linting workflows etc as well. WDYT?

Lets do it!

ghost:
@derekmisler @krissetto If there's anything I can help with, let me know

krissetto:
we got it thanks

dgageot (approved):

PR #1622, opened 2026-02-06
Add debug title command for session title generation

dgageot:
+ Add a 'cagent debug title <agent-file> <question>' command that generates a session title using the same sessiontitle.Generator code path as the TUI, making it easy to debug title generation behavior.
+ Fix title generation for openai 5* models
+ Fix title generation for gemini 3 pro

Assisted-By: cagent

### DIFF
--- a/pkg/model/provider/gemini/client.go
+++ b/pkg/model/provider/gemini/client.go
@@ -327,17 +327,35 @@ func (c *Client) buildConfig() *genai.GenerateContentConfig {
 	// hang or fail. IncludeThoughts=false is also set to ensure no thinking content
 	// is returned.
 	if thinking := c.ModelOptions.Thinking(); thinking != nil && !*thinking {
-		// Explicitly disable thinking - required for operations like title generation
-		// where max_tokens is very low and thinking would consume the token budget.
-		// ThinkingBudget=0 disables thinking for both Gemini 2.5 and 3 models.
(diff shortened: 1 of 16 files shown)


pkg/model/provider/gemini/client.go
```diff
@@ -327,17 +327,35 @@ func (c *Client) buildConfig() *genai.GenerateContentConfig {
...
 	if thinking := c.ModelOptions.Thinking(); thinking != nil && !*thinking {
-		// Explicitly disable thinking - required for operations like title generation
-		// where max_tokens is very low and thinking would consume the token budget.
-		// ThinkingBudget=0 disables thinking for both Gemini 2.5 and 3 models.
-		config.ThinkingConfig = &genai.ThinkingConfig{
-			IncludeThoughts: false,
-			ThinkingBudget:  genai.Ptr(int32(0)),
+		model := strings.ToLower(c.ModelConfig.Model)
+		if strings.HasPrefix(model, "gemini-3-") {
+			// Gemini 3 models require thinking — they reject ThinkingBudget=0.
```
krissetto, line 332: google has been a pain in changing out their thinking works and needs to be configured 2-3 times already

dgageot: yes, that's why I started adding those tests

PR #1623, opened 2026-02-06
Add skill invocation via slash commands 

dgageot:


.agents/skills/bump-go-dependencies/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+- **If both pass**: stage and commit the changes:
+  ```sh
+  git add -A
+  git commit -m "bump <module_path> from <old_version> to <new_version>" -m "" -m "Assisted-By: cagent"
+  ```
+  Record the dependency as **bumped** in your tracking table.
+
+- **If either fails**: revert all changes and move on:
+  ```sh
+  git checkout -- .
```
krissetto, line 62: I'm generally **strongly** against agents committing for me or generally manipulating git too much, but lets see if others feel the same. 

If users don't use a good enough model, disasters are bound to happen 😅

.agents/skills/bump-go-dependencies/SKILL.md
krissetto, line 1: since this is not reusable anywhere outside of this repo, should this even be a skill or rather something our agent just knows how to do? 🤔 
we could probably compact the prompt quite a bit, include it in the system prompt and get the same results.

dgageot: yes, we could but I want to play a bit more with it. 
+ I think on Claude we can call tools from skills to make them resolve in a single round-trip. I'd like to play with that
+ We don't all use the same agent but we could be using the same set of skills

Also, by activating this as a skill, I realised the things we were missing (/command support and schema)

PR #1625, opened 2026-02-06
Add a generalist coding agent

dgageot:


pkg/config/coder-agent.yaml
```diff
@@ -0,0 +1,144 @@
...
+          4. Continue until all tests pass
+
+  librarian:
+    model: haiku
+    description: Web Researcher
+    instruction: |
+      You search the web for documentation, articles, and technical references.
+
+      When given a query:
+      1. Use context7 to search library/framework documentation
```
rumpl, line 137: There is no context7 mcp server config

pkg/config/coder-agent.yaml
```diff
@@ -0,0 +1,144 @@
...
+      </librarian>
+    skills: true
+    add_date: true
+    add_environment_info: true
+    add_prompt_files:
+      - AGENTS.md
+    sub_agents:
+      - librarian
+    toolsets:
+      - type: filesystem
```
rumpl, line 75: maybe we can filter the fs tools and give it only what's needed. For example do we need the "search file" and "search file contents" tools if it has a shell

krissetto:
Just thinking out loud here..

Since this will be shipped as a built-in agent and not just an example or sorts, i kinda feel like we should make this general coding assistant much more "fully featured", with some sort of plan build review type of flow, or at least having the agent team pre-configured to be able to do so. just having an opus + haiku thing just doesn't seem very useful to me.

Also, I've **never** really seen the `golang_developer` agent in our delegate to the librarian (not a good name imho) if not for silly prompts that had nothing to do with coding, so is it even worth it to include it? just give opus the tool since it has literally everything else

Honestly i don't know if shipping this as is will be super valuable to most users who enjoy using defaults, as it only caters to anthropic users and doesn't really seem like it'd be good at doing its job with so many different tasks in the system prompt if not relying on some of the best models out there (aka opus).

Definitely don't wanna block this, just throwin' out some food for though if we want to provide a "proper" coding agent as a built-in thing. We can definitely iterate on it over time

rumpl:
> Also, I've never really seen the golang_developer agent in our delegate to the librarian

By design, up to the user to tell the coder to use it, I like it like that

> librarian (not a good name imho)

Stole it from Amp

krissetto:
> Stole it from Amp

What is Amp? :troll: 

> By design, up to the user to tell the coder to use it, I like it like that

If the user doesn't know about it and its purpose, because its a default, what percentage of users do you think would ask? :stuck_out_tongue:

pkg/config/resolve.go
```diff
@@ -19,6 +19,24 @@ import (
...
+var coderAgent []byte
+
+// builtinAgents maps built-in agent names to their embedded YAML configurations.
+var builtinAgents = map[string][]byte{
+	"default": defaultAgent,
+	"coder":   coderAgent,
+}
+
+// BuiltinAgentNames returns the names of all built-in agents.
+func BuiltinAgentNames() []string {
```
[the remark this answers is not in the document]
dgageot: Let's ignore this one

pkg/config/coder-agent.yaml
dgageot: I'll give it a try

dgageot:
@krissetto @rumpl Like you, I see this as a starting point. Let's iterate after the merge to make this default coding agent  awesome

PR #1629, opened 2026-02-06
tui(editor): add configurable keybindings with runtime help

### LINKED ISSUE #1626: TUI: allow configurable newline/send keybindings (Shift+Enter alternative to Ctrl+J)
## Problem
In `cagent run` (TUI), newline currently falls back to `Ctrl+J` when keyboard enhancements are unavailable. In practice this can be problematic in terminal workflows (especially inside VS Code), where `Ctrl+J` is often already bound.

## Request
Please support customizable input keybindings in TUI, specifically for **send** and **insert newline** actions.

At minimum, please provide an option to choose an alternative newline shortcut besides `Ctrl+J` when `Shift+Enter` is not detected/supported.

## Why this matters
- `Ctrl+J` conflicts with common editor/terminal shortcuts
- Some terminal/tmux combinations don’t reliably deliver `Shift+Enter`
- Users need a practical fallback they can control

## Suggested approaches
- Config option(s), e.g. in user config, for:
  - `editor.send_key`
  - `editor.newline_key`
- Optional env var override for quick testing
- Runtime command/help to display active bindings

Thanks!

Pnkcaht:
## What I did

- Added configurable TUI keybindings for **sending messages** and **inserting newlines**.
- Introduced `editor_keys.go` to load, normalize, and validate keybindings with env overrides.
- Supported multiple newline keys and a Shift+Enter fallback when keyboard enhancements are available.
- Integrated the bindings into the editor update loop via `IsSendKey` and `IsNewlineKey` helpers.

## Related issue
Fixes #1626 

## Design overview

```mermaid
flowchart TD
    A[User presses key] --> B[tea.KeyPressMsg]
    B --> C{EditorKeyBindings}

    C -->|IsSendKey| D[Send message]
    C -->|IsNewlineKey| E[Insert newline]

    C --> F[Keyboard enhancements supported?]
    F -->|yes| G[Shift+Enter fallback]
    F -->|no| H[Configured newline keys only]

    I[Environment variables] --> C
    J[Defaults] --> C
```


ghost:
@krissetto I'll need /code-review here

go.mod
krissetto, line 1: dont downgrade dependencies pls

ghost: ok anyway, I resolved the conflict.

krissetto:
We had in mind to potentially make ALL keybindings configurable. That's a bigger task

We have to be careful here because cagent needs to be as compatible as possible cross terminal, cross-os, etc. and terminals are weird.

ghost:
> We had in mind to potentially make ALL keybindings configurable. That's a bigger task
> 
> We have to be careful here because cagent needs to be as compatibel as possible cross terminal, cross-os, etc. and terminals are weird.

Okay, I understand and totally agree with your point :)

ghost:
Okay, I'll solve the problems 😃

krissetto:
Env vars are a tedious way to handle these sorts of settings.

For this feature, I think we should take a different approach, and start by setting up a way to override potentially all keybindings via the pre-existing global configuration (e.g. `~/.config/cagent/config.yaml`).

Here are some suggestions (for this or a separate PR your choice):

- Take a look at `pkg/userconfig/userconfig.go`;
- Start by introducing a new struct called `Keybindings` in the `Settings` struct and wiring it up with the bindings in `pkg/tui/tui.go`.
- Start small with just 1 configurable remapping, leaving the rest hardcoded as-is so it's a bit easier to validate the approach;
- Don't start from `ctrl+j/shift+enter`, it's an edge case that requires terminal detection etc. We should be focusing on all the other mappings for the custom functionality of `cagent` first, over the usage of the editor (think things like opening the command pallette, the model selection, etc.)

If we go that route, we'll have a much more configurable and testable setup, and it'll be easier for users to setup `cagent` to their preferences by having all user settings centralized in one spot, without breaking the mental model of "where do i configure x?"

Wanna have a go at it @Pnkcaht?

ghost:
@krissetto Gotcha, I'm going to create a new PR and do that.

PR #1633, opened 2026-02-07
Add OCI artifact wrapper for spec-compliant manifest with artifactType

dgageot:
Fixes https://github.com/docker/cagent/issues/1508

Introduce artifactImage wrapper that injects the artifactType field and replaces the config blob with the OCI empty descriptor, since go-containerregistry's v1.Manifest struct does not support artifactType.

Push now wraps images with NewArtifactImage so pushed manifests include artifactType (application/vnd.docker.cagent.config.v1+json) and an empty config descriptor per the OCI image spec guidelines for artifact usage.

Assisted-By: cagent

### DIFF
--- a/pkg/content/artifact.go
+++ b/pkg/content/artifact.go
@@ -1,0 +1,59 @@
+package content
+
+import (
+	"encoding/json"
+	"fmt"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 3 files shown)


dgageot:
@jcarter3 Would that fix the issue?

jcarter3:
@dgageot That fixes the artifact type issues, yes! The layers are still showing as tar.gzip though, when they are just plain yaml files. So that part still needs fixed.

PR #1653, opened 2026-02-09
fix: use alt modifier for session browser shortcuts to avoid search conflict

### LINKED ISSUE #1642: Impossible to search certain sessions
On the past sessions screen it's impossible to search for sessions containing "c" or "f" in their title because these trigger filtering or session id copy.

rumpl:
Plain `s`, `f`, and `c` keys were intercepted by the star, filter, and copy-id actions before reaching the search text input, making it impossible to search for sessions containing those characters.

Change the key bindings to `alt+s`, `alt+f`, and `alt+c` so that unmodified key presses are forwarded to the search input as expected.

Fixes #1642


dgageot:
@rumpl lol, I implemented the search field before the shortcuts but never used it...
Also this doesn't work for me.

PR #1657, opened 2026-02-09
Better default security for cagent api|mcp|a2a

dgageot:
Improve security by making sure that `cagent api`, `cagent a2a` and `cagent mcp` are not exposed to `0.0.0.0` by default.
Also removed CORS configuration from `cagent api`.

This is a breaking change for all three commands and even more for the last two because I changed the `--port` command line argument to `--listen` for resemblance with `cagent api`.

### DIFF
--- a/cmd/root/a2a.go
+++ b/cmd/root/a2a.go
@@ -33,7 +33,7 @@ func newA2ACmd() *cobra.Command {
 	}
 
 	cmd.PersistentFlags().StringVarP(&flags.agentName, "agent", "a", "root", "Name of the agent to run")
-	cmd.PersistentFlags().StringVarP(&flags.listenAddr, "listen", "l", "127.0.0.1:8080", "Address to listen on")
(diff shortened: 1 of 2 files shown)


cmd/root/a2a.go
```diff
@@ -26,15 +26,14 @@ func newA2ACmd() *cobra.Command {
...
-  cagent a2a agentcatalog/pirate --port 9000`,
+  cagent a2a agentcatalog/pirate --listen 127.0.0.1:9090`,
 		Args:    cobra.ExactArgs(1),
 		GroupID: "server",
 		RunE:    flags.runA2ACommand,
 	}
 
 	cmd.PersistentFlags().StringVarP(&flags.agentName, "agent", "a", "root", "Name of the agent to run")
-	cmd.PersistentFlags().IntVar(&flags.port, "port", 0, "Port to listen on (default: random available port)")
+	cmd.PersistentFlags().StringVarP(&flags.listenAddr, "listen", "l", "127.0.0.1:8080", "Address to listen on")
```
krissetto: minor nit: let's change this so it doesn't conflict with the api default

PR #1670, opened 2026-02-10
Fix redundant cancellation and potential goroutine leak in monitorStdin

### LINKED ISSUE #1664: [bug] Redundant context cancellation and potential goroutine leak in monitorStdin
## 🟡 medium - bug

**File:** `cmd/root/api.go` (line 78)

### Code

```go
func monitorStdin(ctx context.Context, cancel context.CancelFunc, stdin *os.File) {
	// Close stdin when context is cancelled to unblock the read
	go func() {
		<-ctx.Done()
		stdin.Close()
	}()

	buf := make([]byte, 1)
	for {
		n, err := stdin.Read(buf)
		if err != nil || n == 0 {
			// Only log and cancel if context isn't already done (parent died)
			if ctx.Err() == nil {
				slog.Info("stdin closed, parent process likely died, shutting down")
				cancel()
			}
			return
		}
	}
}
```

### Problem

The `monitorStdin` function spawns a goroutine to close `stdin` when the context is cancelled. However, the `for` loop also attempts to close `stdin` and cancel the context when `stdin.Read` returns an error or reads zero bytes. This creates a redundant cancellation logic. If the `stdin.Close()` in the spawned goroutine happens *before* `stdin.Read` returns, the `Read` call will return an error, leading to a redundant `cancel()` call. More importantly, the goroutine that calls `stdin.Close()` is not explicitly managed, and could potentially leak if the `monitorStdin` function exits for other reasons before the context is cancelled.

### Suggested Fix

The spawned goroutine to close `stdin` on context cancellation is sufficient. The `if err != nil || n == 0` block in the `for` loop should simply return, allowing the deferred `cancel()` in `runAPICommand` to handle the cancellation.

---

Pnkcaht:
## What i did
I resolved the error related to goroutine.

## Releated issue  
Fixes #1664

### DIFF
--- a/cmd/root/api.go
+++ b/cmd/root/api.go
@@ -54,24 +54,25 @@ func newAPICmd() *cobra.Command {
 	return cmd
 }
 
-// monitorStdin monitors stdin for EOF, which indicates the parent process has died.
-// When spawned with piped stdio, stdin closes when the parent process dies.
-func monitorStdin(ctx context.Context, cancel context.CancelFunc, stdin *os.File) {
-	// Close stdin when context is cancelled to unblock the read
+// monitorStdin blocks on stdin and returns when it is closed (EOF or error),
+// which typically indicates that the parent process has exited when stdio is piped.
+//
+// Context cancellation is handled by a helper goroutine that closes stdin to
+// unblock the Read call. This function itself does not call cancel(), leaving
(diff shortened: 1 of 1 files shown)


dgageot:
This has been fixed

PR #1677, opened 2026-02-10
Strip other agents' tool calls from conversation history

trungutt:
**Problem**  
In conversation where we have multiple agents coordinating with each other by handoff/delegation mechanism (given that each of these agents has different set of tools), this history can contain tool calls from other agents. In subsequent conversation, this causes the agent to hallucinate tool calls for tools it doesn't have access to, because it sees the tool calls in the history and tries to use them again, but with the incorrect agent.

**Attempt to fix**
~~When building the message history for an agent in `GetMessages()`, strip tool calls and their corresponding tool responses from other agents' messages. The text content is still preserved so the agent retains conversational context, but the tool usage patterns that trigger hallucination are removed.~~

Instead of stripping other agents' tool calls from conversation history (which loses useful context), convert them into plain-text narrative messages with agent attribution, similar to Google ADK's [approach](https://github.com/google/adk-python/blob/ec660ed4f0214fd2a88569d13d819a8035adb748/src/google/adk/flows/llm_flows/contents.py#L545).

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -1,6 +1,7 @@
 package session
 
 import (
+	"fmt"
 	"log/slog"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/session/session.go
```diff
@@ -629,12 +628,52 @@ func (s *Session) GetMessages(a *agent.Agent) []chat.Message {
 
 	startIndex := lastSummaryIndex + 1
 
+	// Collect tool call IDs from other agents' assistant messages so we can
+	// strip both the calls and their corresponding tool-role responses.
```
rumpl: We should _NOT_ do this, instead we should do something like what ADK does, IIRC ADK converts tool calls to simple "FYI, agent did this and that" user messages

dgageot:
Let's close this for now

PR #1681, opened 2026-02-10
Improve compaction

dgageot:
I had cagent review the compaction code. Here's what it says:

| Scenario | Behavior | Status |
|---|---|---|
| **Successful auto-compaction** | Summary + continuation added in `Compact()`. Tokens reset in `runtime.go`. Model gets a valid user message. | OK |
| **Failed compaction (API error)** | `Compact()` returns early ‚Äî no summary, no continuation, no token reset. | OK |
| **Empty summary** | `Compact()` returns early at `if summary == ""`. Same as above. | OK |
| **Re-compaction loop** | Tokens reset to 0 after success ‚Äî check does not re-trigger next iteration. | OK |
| **User-triggered `/compact`** | Continuation added inside `Compact()`, then user sends a real message ‚Äî both appear after the summary, model sees both. Harmless. | OK |
| **Double compaction (very long session)** | Each compaction produces its own summary + continuation, tokens reset each time. | OK |

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -988,8 +988,17 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 
 			if m != nil && r.sessionCompaction {
 				if sess.InputTokens+sess.OutputTokens > int64(float64(contextLimit)*0.9) {
+					messageCountBefore := len(sess.Messages)
 					r.Summarize(ctx, sess, "", events)
 					events <- TokenUsage(sess.ID, r.currentAgent, sess.InputTokens, sess.OutputTokens, sess.InputTokens+sess.OutputTokens, contextLimit, sess.Cost)
+
+					// Reset token counters after successful compaction so the check
+					// doesn't trigger again on the next iteration. Compact appends
+					// items on success, so a changed message count indicates success.
+					if len(sess.Messages) > messageCountBefore {
+						sess.InputTokens = 0
+						sess.OutputTokens = 0
+					}
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


pkg/runtime/runtime.go
```diff
@@ -988,8 +988,25 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 
 			if m != nil && r.sessionCompaction {
 				if sess.InputTokens+sess.OutputTokens > int64(float64(contextLimit)*0.9) {
+					messageCountBefore := len(sess.Messages)
 					r.Summarize(ctx, sess, "", events)
 					events <- TokenUsage(sess.ID, r.currentAgent, sess.InputTokens, sess.OutputTokens, sess.InputTokens+sess.OutputTokens, contextLimit, sess.Cost)
+
+					// Check if compaction actually produced a summary by comparing
+					// message count (Compact appends a Summary item on success).
+					if len(sess.Messages) > messageCountBefore {
```
rumpl, line 998: Instead of doing this we should add the compaction message as a normal user message in the session IMO, that way there is no need for this

pkg/runtime/session_compaction.go
```diff
@@ -88,6 +88,14 @@ func (c *sessionCompactor) Compact(ctx context.Context, sess *session.Session, a
 	}
 
 	sess.Messages = append(sess.Messages, session.Item{Summary: summary})
+
+	// After compaction, the summary is the last item. GetMessages starts
+	// collecting conversation messages after the last summary, so there
+	// would be zero conversation messages. Providers (e.g. Anthropic)
+	// reject requests with no non-system messages, so we add a
+	// continuation message to bridge the gap.
+	sess.AddMessage(session.ImplicitUserMessage("The conversation was automatically compacted. Please continue where you left off."))
```
rumpl, line 97: I think this is actually the issue, we are putting the session summary as a system prompt, but it should be a user message https://github.com/docker/cagent/blob/main/pkg/session/session.go#L602

dgageot:
@rumpl please take a look at fixing this right. I'm closing the PR

PR #1698, opened 2026-02-11
Add git reference support

maxcleme:
## Summary

This PR adds git reference support to cagent, enabling the tool to work with specific git references (branches, tags, commits) when interacting with repositories.

eg.

```
cagent run git@github.com:maxcleme/awesome_agent_team.git
```

## Changes

- Implements git reference handling capabilities
- Allows users to specify target references for operations
- Enhances repository interaction flexibility

## Type of Change

- Feature: New functionality for git reference support

## Notes

- Use https://github.com/moby/buildkit/blob/master/frontend/dockerfile/dfgitutil to avoid reinventing the wheel
- No cache for now as I wanted to keep this PR simple, but we could use/inspire https://github.com/docker/compose/tree/main/pkg/remote if we want more optimized way

### DIFF
--- a/pkg/config/sources.go
+++ b/pkg/config/sources.go
@@ -16,13 +16,14 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/moby/buildkit/frontend/dockerfile/dfgitutil"
+	"github.com/moby/buildkit/util/gitutil"
+
 	"github.com/docker/cagent/pkg/content"
 	"github.com/docker/cagent/pkg/environment"
 	"github.com/docker/cagent/pkg/httpclient"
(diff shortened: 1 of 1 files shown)


maxcleme:
@dgageot Any idea regarding the license check? I'm having similar issue when ran locally against `upstream/main`? 🤔

maxcleme:
Looks like we can already achieve similar results by using plain github raw http link, eg: `https://raw.githubusercontent.com/maxcleme/awesome_agent_team/refs/heads/main/agent.yaml`. Private repo also works if `GITHUB_TOKEN` is set.

Fine by me for now.

PR #1721, opened 2026-02-13
Refactor models store

dgageot:


pkg/modelsdev/store.go
```diff
@@ -78,72 +49,34 @@ func newStoreInternal(opts ...Opt) (*Store, error) {
...
 	}
-	s.cacheDir = cacheDir
-	for _, opt := range opts {
-		opt(s)
-	}
 
-	s.client = &http.Client{
-		Timeout: 30 * time.Second,
-	}
+	cacheFile := filepath.Join(cacheDir, CacheFileName)
```
krissetto, line 53: do we need to cache to a file if the duration is the lifetime of the store?

dgageot: I'll measure

PR #1722, opened 2026-02-13
Replace time.Sleep in tests with deterministic synchronization

dgageot:
Use require.Eventually, time.AfterFunc, and channel synchronization instead of fixed time.Sleep calls to make tests faster and less flaky.

- telemetry: poll mockHTTP.GetRequestCount() with require.Eventually
- fake/proxy: use channel write and require.Eventually for sync points
- runtime/fallback: use time.AfterFunc for context cancel, require.Eventually for cooldown expiry
- tui/styles: poll atomic callback counter with require.Eventually

Assisted-By: cagent

### DIFF
--- a/pkg/fake/proxy_test.go
+++ b/pkg/fake/proxy_test.go
@@ -234,8 +234,8 @@ func TestStreamCopy_ContextCancellation(t *testing.T) {
 		done <- StreamCopy(c, resp)
 	}()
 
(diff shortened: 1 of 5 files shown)


rumpl:
Can't we use `"testing/synctest"`

dgageot:
@rumpl It's coming in the next batch (where it's really useful)

PR #1724, opened 2026-02-13
Speed up fallback tests

dgageot:


dgageot:
@rumpl synctest.Test just for you

PR #1725, opened 2026-02-13
Support ${headers.NAME} syntax to forward upstream API headers to toolsets

dgageot:
Allow toolset header values to reference incoming API request headers using ${headers.NAME} placeholders.

For example, a toolset config like:

    headers:
      Authorization: ${headers.Authorization}

will resolve the Authorization value at request time from the upstream HTTP request that triggered the agent.

### DIFF
--- a/pkg/upstream/headers.go
+++ b/pkg/upstream/headers.go
@@ -1,0 +1,135 @@
+// Package upstream provides utilities for propagating HTTP headers
+// from incoming API requests to outbound toolset HTTP calls.
+package upstream
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/dop251/goja"
+)
+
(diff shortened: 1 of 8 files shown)


pkg/upstream/headers.go
```diff
@@ -0,0 +1,111 @@
...
+
+	resolved := make(map[string]string, len(headers))
+	for k, v := range headers {
+		resolved[k] = resolveValue(v, upstream)
+	}
+	return resolved
+}
+
+// resolveValue replaces ${headers.NAME} in a single string value.
+func resolveValue(v string, upstream http.Header) string {
```
rumpl: Let's use goja for thids no?

dgageot: I'll take a look. the agent didn't want to do it :-)

rumpl (approved):

PR #1737, opened 2026-02-15
fix(cli): auto-continue max iterations in --yolo mode

frntn:
## Problem

When running `cagent exec --yolo` in non-interactive mode (stdin closed/piped),
hitting the max iterations limit triggers a stdin prompt that immediately fails
with "Failed to read input, exiting...".

This is inconsistent: `--yolo` auto-approves tool calls but not iteration extensions.

## Fix

- Auto-approve iteration extensions in `--yolo` mode (capped at 5 to prevent infinite loops)
- Handle `MaxIterationsReachedEvent` in JSON output mode (was completely missing, causing runtime hang)
- Extract shared logic into `handleMaxIterationsAutoApprove` helper to avoid duplication between normal and JSON paths

## Test plan

- 5 new tests covering: yolo auto-approve, safety cap, JSON mode, reject without yolo, JSON safety cap
- `golangci-lint run ./pkg/cli/...` → 0 issues
- `go test ./pkg/cli/...` → 16/16 PASS

### DIFF
--- a/pkg/cli/runner.go
+++ b/pkg/cli/runner.go
@@ -35,6 +35,35 @@ func (e RuntimeError) Unwrap() error {
 // auto-continue when max iterations is reached, to prevent infinite loops.
 const maxAutoExtensions = 5
 
+// maxIterAction describes what the caller should do after a MaxIterationsReachedEvent.
+type maxIterAction int
+
+const (
+	maxIterContinue maxIterAction = iota // auto-approved, keep running
+	maxIterStop                          // safety cap reached, caller should stop
+	maxIterPrompt                        // not in yolo mode, caller should prompt the user
+)
+
+// handleMaxIterationsAutoApprove decides whether to auto-extend iterations in
+// --yolo mode. Returns maxIterContinue (approved), maxIterStop (cap reached),
+// or maxIterPrompt (not in auto-approve mode, caller should ask the user).
+func handleMaxIterationsAutoApprove(autoApprove bool, autoExtensions *int, maxIter int) maxIterAction {
+	if !autoApprove {
+		return maxIterPrompt
+	}
+	*autoExtensions++
+	if *autoExtensions <= maxAutoExtensions {
+		slog.Info("Auto-extending iterations in yolo mode",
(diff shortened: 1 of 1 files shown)


krissetto:
I'm not really sure of this.. The max iterations feature was designed exactly for this purpose, to **not** iterate forever. Auto approving that confirmation prompt silently when using `--yolo` could lead to unexpected (and potentially very expensive) consequences for the user. 

IMHO a better approach is just to set `--max-iterations 999` or some other very high value, so at least what's happening is explicitly chosen by the user. WDYT?

frntn:
Thanks for the review @krissetto 

I had the same initial thought, but while running benchmarks on multiple coding agents using harbor framework (including cagent, which turns out to be one of my fav) I realized two things : 
1. bumping max-iterations is workaround, the bug remains that if the agent reaches that limit in non-interactive mode, the runtime tries to prompt a user that doesn't exist and crashes
2. It's a consistency fix for yolo semantics which already auto-approves tool calls. I opted for full auto-approve, but one specific prompt still requires manual input (and breaks when there's no interactive terminal). That's not a safety net but a bug  as I cannot reply "yes". The fix aligns the behavior with yolo semantic which means "auto-approve everything". A safety cap still prevents runaways

Bonus: This PR also fixes the JSON output mode, which was completely missing handling for MaxIterationsReachedEvent and caused a runtime hang.

Does that make more sense to you? 
It's really a practical fix from what I experienced and expected, not some theoretical one :)

krissetto:
@frntn I do agree it could be nice, just not _totally_ sure its the right move. If we go with it, we need to document the behavior properly too both in the --yolo and --max-iteration sections to make it very clear to users. 

Agree we need to handle the `exec` command better in non-interactive scenarios though.

To be clear, I'm not **strongly** against the change. Let me stew on it just a bit and get back to you

frntn:
Maybe using a proper wording in the description like "soft-limit" and "hard-limit" can help

krissetto:
agree with the wording

frntn:
@krissetto not sure if you need anything from me...

PR #1740, opened 2026-02-15
Fix model switcher

dgageot:
This has been broken by e3eb713


trungutt (dismissed):

krissetto (commented):
i'm a bit confused, why would we need a fallback for an older store? wouldn't it be better to make sure the interface is satisfied at compile time so the issue won'r happen again? (had put a similar fix in my tab pr after rebasing on main with the issue haha)

rumpl:
I have no idea what is going on here or how it fixes the commit you linked

dgageot:
@krissetto I'll TAL. I honestly didn't look at the commit

dgageot:
@krissetto @rumpl I pushed a better fix

PR #1745, opened 2026-02-16
Rename cagent OCI annotation, keep old one

gtardif:
Support annotation `io.docker.agent.version` in addition to the cagent one
Updated config media type with `docker.agent`


gtardif:
Just rebased as the branch was a bit behind `main`

PR #1751, opened 2026-02-16
feat: add `X-Cagent-Model-Name` header to models gateway requests

tdabasinskas:
This might be useful for other model gateway implementations.

- Add `Name` field to `ModelConfig` for runtime manifest name tracking
- Add `WithModelName` httpclient option that sets `X-Cagent-Model-Name` header
- Set `ModelConfig.Name` in teamloader and model switcher when resolving from config map
- Include `X-Cagent-Model-Name` in OpenAI, Anthropic, and Gemini provider clients
- Add httpclient tests for `WithModelName`, `WithModel`, and `WithProvider`

### DIFF
--- a/pkg/httpclient/client_test.go
+++ b/pkg/httpclient/client_test.go
@@ -40,7 +40,7 @@ func TestWithModelName(t *testing.T) {
 			defer srv.Close()
 
 			client := NewHTTPClient(WithModelName(tt.modelName))
-			req, err := http.NewRequest(http.MethodGet, srv.URL, nil)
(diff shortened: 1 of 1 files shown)


dgageot:
Hi @tdabasinskas, tanks for the contribution! Could you elaborate on the use case?

tdabasinskas:
> Hi @tdabasinskas, tanks for the contribution! Could you elaborate on the use case?

We use custom gateway via `--models-gateway` and we're interested to know the model names in the manifest (in addition to the provider model name).

PR #1763, opened 2026-02-17
fix: gemini use of vertexai environment variables 

abaez-pantheon:
The following patch does an update on the order options available for
setting vertexai. Specifically, it allows to also check if the 
environment variable for identifying when to use vertex or not with
GOOGLE_GENAI_USE_VERTEXAI variable. Providing the variable then 
sets vertexai backend to use the google default credentials.

Tested changes locally with running and using cagent with the modifications.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -63,7 +63,7 @@ tasks:
   test:
     aliases: [t]
     desc: Run tests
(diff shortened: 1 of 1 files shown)


abaez-pantheon:
Thank you @dgageot!

PR #1766, opened 2026-02-17
Add mouse click support to elicitation, prompt input, and tool confirmation dialogs

dgageot:
Enable mouse interactions for TUI dialogs so users can click to focus fields, select enum/boolean options, and trigger confirmation actions instead of relying solely on keyboard navigation.

- ElicitationDialog: click field labels to focus, click boolean/enum options to select them directly
- MCPPromptInputDialog: click field labels or inputs to focus them
- ToolConfirmationDialog: click Y/N/T/A help key regions to trigger the corresponding approve/reject/allow actions

Assisted-By: cagent

### DIFF
--- a/pkg/tui/dialog/elicitation.go
+++ b/pkg/tui/dialog/elicitation.go
@@ -11,6 +11,7 @@ import (
 	"charm.land/bubbles/v2/key"
 	"charm.land/bubbles/v2/textinput"
 	tea "charm.land/bubbletea/v2"
+	"charm.land/lipgloss/v2"
 
 	"github.com/docker/cagent/pkg/tools"
 	"github.com/docker/cagent/pkg/tui/core/layout"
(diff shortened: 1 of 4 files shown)


aheritier:
I didn't see such mouse click integration in others TUI like Claude and co (I have to re-test them).
On one side I like how it help to interact with the terminal / cagent UI but on the other side it prevents to use mouse functionalities proposed by the terminal like:
* Cmd + Click to open a link
* Right click to access to the contextual menu to split my view, copy, paste ...

rumpl:
@aheritier that's a limitation of the alternate screen mode that we use. others like opencode have the samme issues, once you go alternate screen mode you are left on your own and need to implement everything yourself.

aheritier:
It makes sense. It's just surprising at the beginning

pkg/tui/dialog/elicitation.go
```diff
@@ -423,6 +429,74 @@ func capitalizeFirst(s string) string {
...
 }
 
+// handleMouseClick handles mouse click events for field focus and selection toggling.
+func (d *ElicitationDialog) handleMouseClick(msg tea.MouseClickMsg) (layout.Model, tea.Cmd) {
+	if len(d.fields) == 0 {
+		return d, nil
+	}
+
+	dialogRow, _ := d.Position()
+	dialogWidth := d.ComputeDialogWidth(70, 60, 90)
```
krissetto, line 439: These are duplicated from the View func? if we change one, the other will break if we don't remember to change it.

To keep the experience more reliable long term, i think we should get rid of most magic numbers like these and define constants + dynamically calculate variables based on them. Also helps with reading the code.. There are still plenty around to adjust, so I try to cover them as i find em

trungutt (approved):

PR #1791, opened 2026-02-19
Show message count in session browser dialog

dgageot:
Add NumMessages field to session.Summary and populate it from both SQLite (via LEFT JOIN on session_items) and InMemory stores. Display the count in the session browser as '(N)' next to each session title.

Assisted-By: cagent

### DIFF
--- a/pkg/tui/dialog/session_browser.go
+++ b/pkg/tui/dialog/session_browser.go
@@ -317,13 +317,15 @@ func (d *sessionBrowserDialog) renderSession(sess session.Summary, selected bool
 		title = "Untitled"
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/tui/dialog/session_browser.go
```diff
@@ -317,13 +317,15 @@ func (d *sessionBrowserDialog) renderSession(sess session.Summary, selected bool
 		title = "Untitled"
 	}
 
+	suffix := fmt.Sprintf(" (%d) • %s", sess.NumMessages, d.timeAgo(sess.CreatedAt))
```
rumpl, line 320: `Casual Greeting (2)` is not really telling what the number means

dgageot: I'll improve that

PR #1797, opened 2026-02-19
fix: multiline for yaml on share

abaez-pantheon:
The following patch fixes a bug discovered on yaml marshaling when handling multiline input. Essentially, when pushing to an OCI artifact to an OCI complaint registry, the yaml for multiline indentation is removed. Causing issues with marshaling the yaml for usage.

Tested the change by working locally and sharing to an OCI registry for comparison.

### DIFF
--- a/pkg/config/latest/types.go
+++ b/pkg/config/latest/types.go
@@ -213,10 +213,10 @@ type AgentConfig struct {
 	Name                    string
 	Model                   string            `json:"model,omitempty"`
 	Fallback                *FallbackConfig   `json:"fallback,omitempty"`
-	Description             string            `json:"description,omitempty"`
-	WelcomeMessage          string            `json:"welcome_message,omitempty"`
+	Description             string            `json:"description,omitempty" yaml:"description,omitempty,flow"`
+	WelcomeMessage          string            `json:"welcome_message,omitempty" yaml:"welcome_message,omitempty,flow"`
 	Toolsets                []Toolset         `json:"toolsets,omitempty"`
-	Instruction             string            `json:"instruction,omitempty"`
+	Instruction             string            `json:"instruction,omitempty" yaml:"instruction,omitempty,flow"`
 	SubAgents               []string          `json:"sub_agents,omitempty"`
 	Handoffs                []string          `json:"handoffs,omitempty"`
 	RAG                     []string          `json:"rag,omitempty"`
@@ -473,8 +473,8 @@ type RoutingRule struct {
 type Metadata struct {
 	Author      string `json:"author,omitempty"`
 	License     string `json:"license,omitempty"`
-	Description string `json:"description,omitempty"`
-	Readme      string `json:"readme,omitempty"`
+	Description string `json:"description,omitempty" yaml:"description,omitempty,flow"`
+	Readme      string `json:"readme,omitempty" yaml:"readme,omitempty,flow"`
 	Version     string `json:"version,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


dgageot:
@abaez-pantheon do you have an example of case where the current code fails?

abaez-pantheon:
@dgageot I do! Sorry for not referencing prior. 😅

So when doing a `cagent share push ./some.yaml <oci-ref>`, the `pull` and `run` commands return the following:

```yaml
version: "5"
agents:
  root:
    name: root
    model: pro-high
    description: Designs the software changes. Analyzes the user's request, understands the existing codebase, and creates a plan for the developer to implement. Uses the researcher to gather any necessary information.
    toolsets:
    - type: think
      remote:
        url: ""
      api_config: {}
    - type: filesystem
      remote:
        url: ""
      api_config: {}
    instruction: |
    You are an architect agent responsible for gathering user requirements and creating a development plan.
```

Notice the instruction multiline pipe string. The indention is lost and so an error on yaml parsing returns as if the line is not a string but a key with.

And on execution: 

```shell
❯ cagent run $SOME_OCI_REF
looking for version in config file
[27:5] non-map value is specified
  24 |       path: .memory/mem_root.db
  25 |       api_config: {}
  26 |     instruction: |
> 27 |     You are an architect agent responsible for gathering user requirements and creating a development plan.
           ^
```

dgageot:
@abaez-pantheon sorry, I was slow to follow-up.

I don't totally understand the issue.

The yaml you're sharing doesn't look valid to me. Or is it? Or is it the pushing/pulling that corrupted it?

dgageot:
@abaez-pantheon I have tested pushing a valid yaml and when I pull it, it's still correct. What am I missing?

abaez-pantheon:
@dgageot the main issue that without using the yaml flow field tags for multiline, the formatting for multiline loses the spacing for the string. Here is an example on v1.25 using the [golang_developer.yaml](https://github.com/docker/cagent/blob/main/golang_developer.yaml):

ex

dgageot:
@abaez-pantheon interesting. I have tried with your sample earlier and didn't have any issue. I just tried with golang_developer.yaml and did face the issue...

abaez-pantheon:
> The yaml you're sharing doesn't look valid to me.

Right, the YAML shared was what was serialized through `cagent share push` and then pulled through the `cagent share pull`.

> Or is it the pushing/pulling that corrupted it?

The issue seems to be on serialization on the multiline string. Once uploaded to the container registry, in the incorrect format, cagent is not able to be read. But if serialized correctly, then no issue. 

ex

Here's another recording going through the steps. I tried first by using my patch version of the yaml flow field tag. Then validating the yaml is correct. After, used the latest release version of cagent to also show correctly serialized uploaded yaml is still handled correctly. But once using the latest version, that same file ended up with the incorrect formatting.

Hope this helps.

dgageot:
Thanks a lot @abaez-pantheon for digging this issue. I think I've got a fix here: https://github.com/docker/cagent/pull/1844

dgageot:
@abaez-pantheon I'll let you close this one if the other PR fixes your problem

abaez-pantheon:
Sweet. Thank you @dgageot!

Funny enough, my original patch version was also swapping from marshaling to using the yaml slices directly. But I ended up favoring doing it with simple tag annotation as I also didn't want to start changing more of the usage of the yaml library. 

I did also notice this is an underlying bug with the `goccy/go-yaml` library. Which  then lead me to try work through a full refactor to say [yaml.v4](https://github.com/yaml/go-yaml) lib. However, that was a bit of effort to say the least, for such a simple change with tags or yaml slices. 

That and yaml.v4 is not yet fully released yet either. 😅

Anyway, thanks again for looking into this. Already confirmed and working in the most convulsing of multiline string file i conjured up. 😂

PR #1810, opened 2026-02-20
Freeze v5 (and a few refactoring)

dgageot:


rumpl:
Should we do this after the skills thing? https://github.com/docker/cagent/pull/1822

dgageot:
@rumpl OK. Also I think I got it messed up a bit. I'll check tomorrow

PR #1828, opened 2026-02-23
builtin: add offset and line_count pagination to read_file and read_multiple_files

trungutt:
- `read_file` and `read_multiple_files` had no output size limit, so a single call reading a large repository could return 150K+ characters in one tool result, pushing the session context over 100K tokens (which is pretty big)
- Replace the blunt `limitOutput()` cap with proper pagination: both tools now accept `offset` (1-based line number) and `line_count` parameters so the model can read large files incrementally.
- When a subset is returned, a header is prepended (e.g. `[Showing lines 1-150 of 250 from AGENTS.md]`) so the model knows how much content remains and can request the next chunk.
- `ReadFileMeta.LineCount` renamed to `TotalLines` — always reflects the full file regardless of the window requested.
- Tool descriptions and `Instructions()` updated to guide the model to paginate large files.


pkg/tools/builtin/filesystem.go
```diff
@@ -226,7 +236,7 @@ func (t *FilesystemTool) Tools(context.Context) ([]tools.Tool, error) {
 		{
 			Name:         ToolNameReadFile,
 			Category:     "filesystem",
-			Description:  "Read the complete contents of a file from the file system.",
+			Description:  "Read the contents of a file from the file system. Use offset and line_count to read large files in chunks; check total_lines in the response to determine if more lines remain.",
```
krissetto: > Use offset and line_count to read large files in chunks; check total_lines in the response to determine if more lines remain.

i wouldn't add instructions here, since we already have way too much in the instruction prompt

pkg/tools/builtin/filesystem.go
```diff
@@ -238,7 +248,7 @@ func (t *FilesystemTool) Tools(context.Context) ([]tools.Tool, error) {
 		{
 			Name:        ToolNameReadMultipleFiles,
 			Category:    "filesystem",
-			Description: "Read the contents of multiple files simultaneously.",
+			Description: "Read the contents of multiple files simultaneously. Use offset and line_count to read large files in chunks; check total_lines per file in the response to determine if more lines remain.",
```
krissetto: same as above

krissetto (commented):
some minor things

PR #1831, opened 2026-02-23
context truncation/compaction improvements

krissetto:
- Removes truncateOldToolContent and MaxToolCallTokens from session to avoid busting cache unnecessarily and potentially confusing models
- Preserves assistant text as a separate message item before function_call items in convertMessagesToResponseInput (responses API)
- Lowers default context limit before compaction to 80% of model's context length. Anything after 50% usually sees progressively bigger drops in output quality so 80% seems a good point to compact at

These are opinionated changes, things seem to perform generally better if we let the caching do its job and don't edit/remove things from the history. 
Lets do some tests and see how we feel about them

### DIFF
--- a/pkg/model/provider/openai/client.go
+++ b/pkg/model/provider/openai/client.go
@@ -480,20 +480,31 @@ func convertMessagesToResponseInput(messages []chat.Message) []responses.Respons
 					},
 				}
 			} else {
-				// Assistant message with tool calls - convert to response input item with function calls
+				// Preserve assistant text content as a separate message so the
+				// model retains conversational context across tool-call rounds.
+				if strings.TrimSpace(msg.Content) != "" {
+					input = append(input, responses.ResponseInputItemUnionParam{
+						OfMessage: &responses.EasyInputMessageParam{
+							Role: responses.EasyInputMessageRoleAssistant,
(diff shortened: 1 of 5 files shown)


rumpl:
> Preserves assistant text as a separate message item before function_call items in convertMessagesToResponseInput (responses API)

Why?

rumpl:
The same way we can set the max number of messages to keep in the context, we should maybe make this configurable?

krissetto:
>> Preserves assistant text as a separate message item before function_call items in convertMessagesToResponseInput (responses API)

> Why?

why not? some models can say something together with their tools calls, we were dropping it entirely which just seems wrong to me

> The same way we can set the max number of messages to keep in the context, we should maybe make this configurable?

agree. i think the default should only be compaction, not thread truncation, to also to align with what are _likely_ to be most users' expectations (aka "don't mess with my messages unless you really have to"). 
but being able to configure it is good imho

I've been giving this a ride and honestly it feels quite a bit better to me, especially with openai models. Curious to know if you noticed any issues etc, even cost related ones (not busting cache helps with costs too in many scenarios)

krissetto:
this is old, idk what else has been merged recently along these lines. will close and re-open in the future if things can still be improved

PR #1833, opened 2026-02-24
allow .github and .gitlab directories in WalkFiles traversal

maxcleme:
When working on CI it is super annoying to be not able to use `@` with anything inside `.github/` folder.

### DIFF
--- a/pkg/fsx/fs.go
+++ b/pkg/fsx/fs.go
@@ -127,6 +127,12 @@ var heavyDirs = map[string]bool{
 	".vscode":      true,
 }
 
+// allowedDirs are hidden directories that should be traversed despite starting with a dot.
+var allowedDirs = map[string]bool{
+	".github": true,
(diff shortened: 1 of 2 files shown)


krissetto:
maybe we should guard against .env as a precaution

maxcleme:
This `.env` isn't a dir, it will return `nil`, right? (and so not be processed?)

aheritier:
I don't think in claude code & friends they block it?
What they advice is to configure the permission to not read some files ( ex: https://code.claude.com/docs/en/settings#excluding-sensitive-files ) and they have a setting 'respectGitignore' to block by default the file-picker (personally I disable it)

krissetto:
@aheritier  yeah true, and we did have similar things in place in cagent as well :)

maxcleme:
@krissetto Fwiw I do not have write permission here 😁

PR #1855, opened 2026-02-26
fix: record token usage once per stream to prevent inflated telemetry

dgageot:
Providers like Gemini emit usage metadata on every streaming chunk with cumulative token counts. The previous code ran telemetry.RecordTokenUsage (which uses +=) on every chunk, causing massively inflated telemetry token counts and cost.

Move the session token updates and telemetry call out of the per-chunk loop into a recordUsage helper that runs exactly once, right before handleStream returns. The latest usage snapshot is still captured (last wins) so token counts remain accurate.

Assisted-By: cagent


rumpl:
Wasn't this already fixed once?

PR #1875, opened 2026-02-27
Skip hidden directories in recursive skill loading

dgageot:
Return fs.SkipDir for hidden directories and symlinks instead of descending into them. This avoids walking large trees like .git, .node_modules, or .cache that can never contain skills.

Assisted-By: cagent

### DIFF
--- a/pkg/skills/skills.go
+++ b/pkg/skills/skills.go
@@ -210,7 +210,22 @@ func loadSkillsFlat(dir string) []Skill {
 }
 
 // loadSkillsRecursive loads skills from all subdirectories (Codex format).
+// It tracks visited real directory paths to avoid infinite loops caused by
+// symlinks that form cycles.
 func loadSkillsRecursive(dir string) []Skill {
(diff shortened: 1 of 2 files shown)


rumpl:
I'm not sure about symlinks though, seems like something someone would do

pkg/skills/skills.go
```diff
@@ -194,7 +194,7 @@ func loadSkillsFlat(dir string) []Skill {
 
 	var skills []Skill
 	for _, entry := range entries {
-		if !entry.IsDir() || isHiddenOrSymlink(entry) {
+		if !entry.IsDir() || (isHidden(entry) || isSymlink(entry)) {
```
aheritier, line 197: why not following symlinks? The real change with symlink are security issues to ensure that they don't allow to go outside of the project folder....

dgageot: I didn't want to work on `loadSkillsFlat` in that PR. Happy to work on it in another PR

PR #1879, opened 2026-02-28
Reduce builtin tool prompt lengths while preserving key examples

dgageot:
Shorten Instructions() and tool descriptions across all builtin tools: shell, LSP, fetch, memory, think, tasks, deferred, and user_prompt. Remove redundant prose, verbose output format sections, and excessive examples. Retain essential examples for critical workflows (symbol search, references, shell heredocs, user_prompt schemas). Update corresponding tests to match reduced descriptions.

Assisted-By: cagent

### DIFF
--- a/pkg/tools/builtin/lsp_test.go
+++ b/pkg/tools/builtin/lsp_test.go
@@ -3,7 +3,6 @@ package builtin
 import (
 	"encoding/json"
 	"os/exec"
-	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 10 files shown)


pkg/tools/builtin/lsp_test.go
```diff
@@ -86,7 +80,7 @@ func TestLSPTool_Instructions(t *testing.T) {
 	instructions := tool.Instructions()
 
 	// Should mention the tools are stateless
-	assert.Contains(t, instructions, "stateless")
+	assert.Contains(t, instructions, "Stateless")
```
gtardif, line 83: Nit: compare with toLower() to make this more resilient? (if we need this change, feels like a sign to me)

PR #1889, opened 2026-03-02
:eyes: Vision :eyes:

rumpl:
## Testing this

With this code, run your agent that has the filesystem toolset and ask it "What's in <image.png>", it should be able to read it and describe the image for you. 

Another example is that now you can ask an agent that has the playwright MCP for example: "take a screenshot of random.org and tell me what you see"

---

Adds end-to-end image support so that LLMs can see images from tool results (read_file, MCP tools) and user attachments (@file.png).

 ### What's new

 Image ingestion
 - The read_file filesystem tool now detects image files (JPEG, PNG, GIF, WebP) and returns them as base64-encoded image content alongside a text description
 - MCP tool results containing ImageContent are extracted and forwarded to the model
 - User @file.png attachments are read, resized, and inlined as base64 data URLs — this works across all providers, not just Anthropic's File API

 MIME detection
 - DetectMimeType now uses content sniffing first (magic bytes via http.DetectContentType + manual WebP check), falling back to extension only for text files
 - New DetectMimeTypeByContent helper for raw byte detection

 Image resizing
 - Images are automatically resized to stay within provider limits (max 2000×2000 px, max 4.5 MB)
 - Uses CatmullRom (bicubic) interpolation via golang.org/x/image/draw
 - Progressive fallback strategy: PNG vs JPEG comparison → decreasing JPEG quality (70/55/40) → reduced dimensions (75%/50%/35%/25%)
 - FormatDimensionNote generates coordinate mapping notes (e.g. "original 4000×3000, displayed at 2000×1500. Multiply coordinates by 2.00...") so the model can translate coordinates back to the original image

 Provider-specific encoding
 - Anthropic: images inline as image blocks in tool_result content (both standard and beta API)
 - OpenAI: injected user message with image_url parts after tool result (completions and responses API — tool messages only support text)
 - Gemini: multimodal function responses via FunctionResponseParts with inline image data

 Model capability gating
 - Uses modelsdev model metadata (Modalities.Input) to check if the current model supports image input
 - Automatically strips image content from messages when the model is text-only (e.g. gpt-3.5-turbo, o1-mini, codex-mini), preventing API errors

 ### New dependency

 - golang.org/x/image — for draw.CatmullRom (bicubic scaling) and WebP decoding

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -408,8 +408,10 @@ func (a *App) processFileAttachment(ctx context.Context, att messages.Attachment
 			}
 			resized, resizeErr := chat.ResizeImage(imgData, mimeType)
 			if resizeErr != nil {
-				slog.Warn("image resize failed, sending original", "path", absPath, "error", resizeErr)
-				resized = &chat.ImageResizeResult{Data: imgData, MimeType: mimeType}
+				// Don't bypass security checks - reject the file if resize failed
+				slog.Warn("skipping attachment: image resize failed", "path", absPath, "error", resizeErr)
+				a.sendEvent(ctx, runtime.Warning(fmt.Sprintf("Skipped attachment %s: %s", att.Name, resizeErr), ""))
+				return
 			}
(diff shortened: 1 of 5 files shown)


jeanlaurent:
YES !!!

PR #1904, opened 2026-03-03
cagent run --sandbox

dgageot:
Features:
+ [x] Creates a docker sandbox to run the agent
+ [x] Resolves aliases pre run in the sandbox to know which agent file to mount
+ [x] Dynamically pass env vars, such as MCP secrets, from our env providers
+ [x] Support the gateway and DD token based auth
+ [x] Mount user's settings (read-only)


dgageot:
@rumpl I had to rebase

Issue #863, opened 2025-11-19
Implement a sequential agent and a parallel agent


dgageot:
See https://github.com/docker/cagent/issues/862

+ A sequential agent is probably a special case of the loop agent with only 1 iteration. Keeping the exit conditions can be useful
+ A parallel agent calls sub agents concurrently, with the same input. It them collects all the outputs and arranges them in the order of the agents, no matter in what order those gave their response.

riturajFi:
Hi @dgageot @rumpl @krissetto  , this is very interesting - would love to work on it!

riturajFi:
Hi @dgageot @rumpl @krissetto , this is very interesting - would love to work on it!``

krissetto:
Hey @riturajFi! I wouldn't take this issue right now, we're trying to scope things out here and figure out exactly how we'd like these features to work

riturajFi:
Sure @krissetto , i just want to make some contributions to some core agent/tool related issue with good impact. Would love it if you could assing something to me

ghost:
Resolved @dgageot

dgageot:
Let's close this one. It's been there for too long

PR #1908, opened 2026-03-03
Add background agent task tools for concurrent sub-agent dispatch (#863)

jfrey:
## Summary

Implements concurrent sub-agent orchestration from #863 by adding four background agent tools that mirror the existing shell background job pattern:

- **`run_background_agent`** — dispatch a named sub-agent task asynchronously, returns a task ID immediately
- **`list_background_agents`** — show all tasks with status and runtime
- **`view_background_agent`** — inspect live buffered output or final result by task ID
- **`stop_background_agent`** — cancel a running task

Tools are registered automatically for any agent that has `sub_agents` configured, alongside the existing `transfer_task` tool.

## Architecture

Handler and task tracking live in **`pkg/tools/builtin/agent/`**, following the same self-contained pattern as the shell tool in `builtin/shell.go`. A `Runner` interface abstracts the runtime dependency:

```go
type Runner interface {
    CurrentAgentSubAgentNames() []string
    RunAgent(ctx context.Context, params RunParams) *RunResult
}
```

`LocalRuntime` implements `Runner` and registers the handlers via thin delegation methods, matching the existing pattern used by `handleTaskTransfer`, `handleHandoff`, etc.

## Key design decisions

- **Session-scoped agent identity** — `session.WithAgentName` pins each background session to its target agent, so `RunStream` resolves configuration from the session rather than the shared `currentAgent` field. This eliminates the race between concurrent background tasks.
- **Async / fire-and-forget** — caller gets a task ID back immediately and polls with `view_background_agent`, matching the UX of shell background jobs
- **`ToolsApproved=true`** for background sessions — no user present to answer approval prompts; documented with a TODO for per-session permission scoping
- **Resource caps** — max 20 concurrent tasks, 100 total; completed tasks auto-pruned; live output buffer capped at 10 MB per task
- **Typed enum** — `taskStatus int32` with `iota`, `google/uuid` for task IDs
- **CAS-based status transitions** for safe concurrent stop/complete

## Test plan

- [x] 28 unit tests in `pkg/tools/builtin/agent/` with race detector
- [x] `go build ./...`, `go vet ./...` clean
- [ ] Manual test: build binary and exercise background agent dispatch in a team with sub-agents


pkg/runtime/background_agent.go
```diff
@@ -0,0 +1,400 @@
...
+	maxAgentOutputBytes = 10 * 1024 * 1024 // 10 MB
+)
+
+// agent task status constants — mirrors backgroundJob in shell.go
+const (
+	agentTaskRunning   int32 = iota
+	agentTaskCompleted int32 = iota
+	agentTaskStopped   int32 = iota
+	agentTaskFailed    int32 = iota
+)
```
dgageot:
```suggestion
const (
	agentTaskRunning   int32 = iota
	agentTaskCompleted
	agentTaskStopped
	agentTaskFailed
)
```

pkg/runtime/background_agent.go
```diff
@@ -0,0 +1,400 @@
...
+	maxConcurrentAgentTasks = 20
+	// maxTotalAgentTasks caps total stored tasks (running + completed) to prevent unbounded memory growth.
+	maxTotalAgentTasks = 100
+	// maxAgentOutputBytes caps the live output buffer per task, mirroring the shell tool's limit.
+	maxAgentOutputBytes = 10 * 1024 * 1024 // 10 MB
+)
+
+// agent task status constants — mirrors backgroundJob in shell.go
+const (
+	agentTaskRunning   int32 = iota
```
dgageot: We should maybe introduce a type for the enum rather than int32?

pkg/runtime/background_agent.go
```diff
@@ -0,0 +1,400 @@
...
+
+func newBackgroundAgentHandler(r *LocalRuntime) *backgroundAgentHandler {
+	return &backgroundAgentHandler{
+		r:     r,
+		tasks: concurrent.NewMap[string, *backgroundAgentTask](),
+	}
+}
+
+// newTaskID generates a unique, non-guessable ID for a background agent task.
+func newTaskID() (string, error) {
```
dgageot: Maybe use github.com/google/uuid?

pkg/runtime/background_agent.go
```diff
@@ -0,0 +1,400 @@
...
+	b := make([]byte, 8)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("agent_task_%d_%s", time.Now().Unix(), hex.EncodeToString(b)), nil
+}
+
+// runningTaskCount returns the number of currently running tasks.
+func (h *backgroundAgentHandler) runningTaskCount() int {
+	count := 0
```
dgageot:
```suggestion
	var count int
```

docs/plans/parallel-task-863.md
```diff
@@ -0,0 +1,179 @@
+# Plan: Background Agent Tasks (Issue #863)
```
dgageot: FWIW, we usually try to avoid committing those plans

dgageot (commented):
LGTM in general! Sorry, I left a few nit-picky comments. Feel free to ignore.
One of the things that could be better is that most of the code lives in the runtime package that we wish would be simpler. Do you think most of it could be extracted?

jfrey:
> LGTM in general! Sorry, I left a few nit-picky comments. Feel free to ignore. One of the things that could be better is that most of the code lives in the runtime package that we wish would be simpler. Do you think most of it could be extracted?

I'll take care of all those and will refactor the background agent into a builtin task.  It is similar to shell but has a dependency on some of the stream interfaces so I can work out an injectable interface for that to segregate the code more.

PR #1921, opened 2026-03-04
Nightly scanner should be less nit-picky about docs

derekmisler:
## Summary

The nightly codebase scanner's documentation agent was filing issues about missing doc comments in Go source files (e.g., #1912). This PR scopes the documentation agent to only analyze `.md` files, raises the bar so it only reports high-impact issues, and fixes a `pipefail` crash in the auto-issue-triage workflow.

## Changes

**Nightly scanner documentation agent** (`nightly-scanner.yaml`):
- Restricted scope to `.md` files only — source code doc comments are no longer in scope
- Replaced the "what to look for" list with high-impact-only criteria: factually incorrect info, missing setup/build steps, broken critical links
- Expanded the "what to ignore" list to explicitly exclude style nits, typos, nice-to-have sections, and internal feature docs
- Added a default-to-`NO_ISSUES` directive so findings only get reported if they would genuinely block or mislead a contributor

**Auto-issue-triage workflow** (`auto-issue-triage.yml`):
- Fixed `grep` exit code 1 causing the "Parse agent result" step to fail under bash `pipefail` when the agent output contains no `RESULT:` marker (e.g., [this run](https://github.com/docker/cagent/actions/runs/22653191673))

## Test plan

- [ ] Trigger a dry-run of the nightly scan (`workflow_dispatch` with dry-run=true) and verify the documentation agent only analyzes `.md` files
- [ ] Verify the auto-issue-triage workflow no longer fails at the "Parse agent result" step when the agent doesn't output a result marker

### DIFF
--- a/.github/agents/nightly-scanner.yaml
+++ b/.github/agents/nightly-scanner.yaml
@@ -272,7 +272,13 @@ agents:
     model: claude-haiku
     description: Documentation gaps and improvements
     instruction: |
-      You analyze code for documentation issues. Only report significant gaps.
+      You analyze markdown documentation files for issues. Only report significant gaps.
+
+      ## ⚠️ SCOPE: Markdown files only
+
+      You MUST only report issues about `.md` files (README.md, CONTRIBUTING.md, etc.).
+      Do NOT report issues about source code files (.go, .py, .js, etc.) — missing
+      doc comments, package docs, or inline comments are out of scope.
 
       ## First: Read existing documentation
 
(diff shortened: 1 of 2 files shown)


.github/agents/nightly-scanner.yaml
```diff
@@ -272,7 +272,13 @@ agents:
     model: claude-haiku
     description: Documentation gaps and improvements
     instruction: |
-      You analyze code for documentation issues. Only report significant gaps.
+      You analyze markdown documentation files for issues. Only report significant gaps.
```
krissetto, line 275: shouldn't we maybe expand it to the actual gh pages style docs as well?

derekmisler: Those were just turned into markdown files, so I _think_ this'll work? I'll keep an eye on it.

PR #1922, opened 2026-03-04
Fix speech to text on macOS

### LINKED ISSUE #1919: /speak not working - Speech-to-text is not yet supported
**Describe the bug**

when I use /speak I get an error message "Speech-to-text is not yet supported" and i can't talk to cagent anymore

**Version affected**

v1.29.0

**How To Reproduce**

Detailed steps to reproduce the behavior:

in TUI type /speak a

**Expectation**

I can talk to cagent

Image

**Screenshots**

**OS and Terminal type**

Mac + Ghostty

dgageot:
Fixes #1919

### DIFF
--- a/pkg/tui/handlers.go
+++ b/pkg/tui/handlers.go
@@ -538,14 +538,19 @@ func (m *appModel) handleStartSpeak() (tea.Model, tea.Cmd) {
 		return m, nil
 	}
 
-	m.transcriptCh = make(chan string, 100)
+	// Close any previous channel to unblock stale waitForTranscript goroutines.
+	m.closeTranscriptCh()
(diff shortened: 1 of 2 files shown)


pkg/tui/handlers.go
```diff
@@ -531,6 +531,69 @@ func (m *appModel) handleAttachFile(filePath string) (tea.Model, tea.Cmd) {
 	})
 }
 
+// --- Speech-to-text ---
```
aheritier, line 534: speech to text or text to speech ?

dgageot: What the code says :-)

PR #1940, opened 2026-03-05
history: Fix loading very long lines

vvoland:
Fix cagent crashing and then failing to start with `failed to initialize command history: bufio.Scanner: token too long`.

Replace bufio.NewScanner with json.NewDecoder for reading history entries.

The scanner has a default 64KB line limit which causes failures when loading very long history entries.

json.Decoder handles newline-delimited JSON streams naturally without size restrictions, simplifying the code and fixing the long message issue.

### DIFF
--- a/pkg/history/history.go
+++ b/pkg/history/history.go
@@ -1,8 +1,8 @@
 package history
 
 import (
-	"bufio"
 	"encoding/json"
+	"io"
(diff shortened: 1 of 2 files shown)


dgageot:
@vvoland The review agent seems to be right about the error handling

PR #1944, opened 2026-03-05
Auto-pull DMR models in non-interactive mode

### LINKED ISSUE #1856: With the docker model runner provider, cagent does not download the model anymore (if not present)
With the docker model runner provider, cagent does not download the model anymore (if not present)

- I'm using `cagent` `1.27.0` (but it was already the case with the previous version)
- `cagent` is running in devcontainer

This is my configuration:
```yaml
agents:
  root:
    
    model: cooking
    skills: true
    num_history_items: 5
    description: Bob, Expert Home Cook Mentor
    instruction: |
      You name is Bob.
      # Role: Expert Home Cook Mentor
      You are a warm, knowledgeable, and highly practical Home Cook Mentor. Your goal is to make cooking accessible, safe, and successful for everyone, regardless of skill level.

models:

  cooking:
    provider: dmr
    model: huggingface.co/menlo/jan-nano-128k-gguf:Q4_K_M
```

dgageot:
Fixes #1856

## Problem

When running cagent with `provider: dmr` in a devcontainer (or any non-interactive environment), `pullDockerModelIfNeeded` detected a non-interactive terminal via `term.IsTerminal()` and silently returned `nil` — skipping the model pull entirely. The client was created successfully but subsequent inference calls failed because the model was never downloaded.

## Solution

- **Auto-pull in non-interactive mode**: Instead of silently skipping, automatically pull the missing model without prompting. Interactive mode still prompts for user confirmation as before.
- **Check model existence first**: Moved the `modelExists` check before the interactivity check so we skip the pull early when the model is already present, regardless of mode.
- **Extract `confirmModelPull`**: Separated the confirmation concern from the pull logic for better readability.

## Also included

- Improvements to the `user_prompt` TUI dialog: customizable title, free-form text input, updated navigation keys and tests.

### DIFF
--- a/pkg/model/provider/dmr/client.go
+++ b/pkg/model/provider/dmr/client.go
@@ -971,19 +971,39 @@ func parseDMRProviderOpts(cfg *latest.ModelConfig) (contextSize *int64, runtimeF
 }
 
 func pullDockerModelIfNeeded(ctx context.Context, model string) error {
-	// Check if running in interactive mode (stdin is a terminal)
-	interactive := term.IsTerminal(int(os.Stdin.Fd()))
-	if !interactive {
-		// In non-interactive mode (CI / Servers), do not attempt to pull the model
-		return nil
-	}
-
 	if modelExists(ctx, model) {
 		slog.Debug("Model already exists, skipping pull", "model", model)
 		return nil
 	}
 
-	// Prompt user for confirmation in interactive mode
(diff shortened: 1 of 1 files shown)


dgageot:
@k33g is that what you want?

k33g:
> @k33g is that what you want?

Yes @dgageot

PR #1950, opened 2026-03-05
fix(hooks): apply ModifiedInput from pre-tool hooks to tool call arguments

aheritier:
The pre-tool hook system already supported returning updated_input via
hook_specific_output, and the executor populated Result.ModifiedInput.
However, the runtime never applied it back to toolCall.Function.Arguments
before execution, making input-rewriting hooks (like RTK) ineffective.

Marshal ModifiedInput to JSON and overwrite the tool call arguments
when a pre-tool hook provides modified input.

Assisted-By: cagent


aheritier:
I stopped to use RTK but to make it work with cagent I had to apply this patch FYI @dgageot @rumpl

aheritier:
(not sure if it has some interest)

aheritier:
this one @dgageot was to show to @rumpl what I had to change to use RTK. I am not sure it makes sense to be addressed, and if it is the right fix to do

Issue #1969, opened 2026-03-06
Multiple LSP toolsets in a single agent cause 'Tool names must be unique' error


aheritier:
## Description

When configuring an agent with **two or more LSP toolsets** (e.g., JDTLS for Java + typescript-language-server for TypeScript), the agent fails to start with:

```
tools: Tool names must be unique
```

This contradicts the [Multiple LSP Servers documentation](https://docker.github.io/docker-agent/tools/lsp/#multiple-lsp-servers) which shows this configuration as supported.

## Steps to Reproduce

**Minimal `cagent.yaml`:**

```yaml
agents:
  polyglot:
    model: anthropic/claude-sonnet-4-0
    description: Multi-language developer
    instruction: You are a full-stack developer.
    toolsets:
      - type: lsp
        command: jdtls
        file_types: [".java"]
      - type: lsp
        command: typescript-language-server
        args: ["--stdio"]
        file_types: [".ts"]
      - type: filesystem
      - type: shell
```

**Run:** `cagent run polyglot`

**Prerequisite:** Both `jdtls` and `typescript-language-server` must be installed and in PATH.

## Expected Behavior

Both LSP servers should start and their tools should be namespaced/routed by file type, as documented in the [Multiple LSP Servers](https://docker.github.io/docker-agent/tools/lsp/#multiple-lsp-servers) section.

## Actual Behavior

The agent fails immediately with:

```
tools: Tool names must be unique
```

## Root Cause Analysis

Each LSP toolset registers the same **15 tool names** (from the [LSP Tool docs](https://docker.github.io/docker-agent/tools/lsp/#available-tools)):

- `lsp_workspace`, `lsp_hover`, `lsp_definition`, `lsp_references`
- `lsp_document_symbols`, `lsp_workspace_symbols`, `lsp_diagnostics`
- `lsp_code_actions`, `lsp_rename`, `lsp_format`
- `lsp_call_hierarchy`, `lsp_type_hierarchy`, `lsp_implementations`
- `lsp_signature_help`, `lsp_inlay_hints`

When two LSP toolsets are configured, all 15 tools are registered twice with identical names, causing the LLM API to reject the tool list (which requires unique tool names).

A single LSP toolset works perfectly fine. The issue only occurs with 2+ LSP toolsets.

## Environment

- **cagent version:** `dev` (commit `3368661d2d3e0e579896ba7c40f879f058fe297d`)
- **OS:** macOS (arm64)
- **LSP servers tested:**
[...]

PR #1975, opened 2026-03-06
fix: LSP tools broken in multi-agent setups (process lifecycle, code mode resilience, sub-agent identity)

aheritier:
## Summary

Fixes three interrelated bugs that prevented LSP tools from working in multi-agent configurations, discovered while testing with a real-world project ([Devoxx Call-for-Papers](https://github.com/docker/docker-agent/issues/1969#issuecomment-4014071952)) using 4 agents with Java (`jdtls`) and TypeScript (`typescript-language-server`) LSP servers.

Stacked on top of #1970 (LSP multiplexer).

### 1. Sub-agents execute with root's toolset and model (`runtime.go`)

`RunStream` correctly resolves the session's agent at the top of its goroutine (via `sess.AgentName` or `r.CurrentAgent()`). However, the iteration loop immediately **shadows** the resolved agent with `a := r.CurrentAgent()`, which always returns `root` for background agent tasks since `run_background_agent` never calls `setCurrentAgent`.

**Impact:** All sub-agents dispatched via `run_background_agent` executed with root's toolset (no LSP tools), root's model (wrong tier), and root's instructions.

### 2. LSP process killed by context cancellation (`lsp.go`)

The LSP server process was started with `exec.CommandContext(ctx)` where `ctx` was the caller's request or sub-session context. When that context ended, the process was killed — but `StartableToolSet.started` remained true, so the server was never restarted.

Also fixes a race condition in `ensureInitialized()` where the mutex unlock/re-lock window allowed concurrent callers to get spurious errors.

### 3. `codeModeTool.Start()` fails fast and can't find wrapped Startable (`codemode.go`)

`Start()` returned on the first error, so one broken toolset (e.g., unreachable MCP server) killed all tools for the agent. Also used direct type assertion instead of `tools.As`, unable to find `Startable` through wrapper chains.

## Commits

- **`d0d0b4c`** — `fix(lsp)`: detach LSP process from caller context, fix startup race, add `done` channel for goroutine cleanup
- **`7e11e4d`** — `fix(codemode)`: make `Start()` resilient (continue on failure) and unwrap-aware (`tools.As`)
- **`baf531b`** — `fix(runtime)`: use session-resolved agent in RunStream iteration loop

## Testing

- All existing tests pass (`task test`, `task lint`)
- New tests added for code mode resilience (`TestCodeModeTool_Start_ContinuesOnFailure`, `TestCodeModeTool_Start_UnwrapsToFindStartable`)
- Validated with a real multi-agent project (4 agents, 7 LSP server instances): 12/14 LSP tools working across all agents

Ref #1969

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -969,7 +969,7 @@ func (r *LocalRuntime) finalizeEventChannel(ctx context.Context, sess *session.S
 
 // RunStream starts the agent's interaction loop and returns a channel of events
 func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-chan Event {
-	slog.Debug("Starting runtime stream", "agent", r.CurrentAgentName(), "session_id", sess.ID)
+	slog.Debug("Starting runtime stream", "agent", r.CurrentAgentName(), "session_agent", sess.AgentName, "session_id", sess.ID)
 	events := make(chan Event, 128)
 
 	go func() {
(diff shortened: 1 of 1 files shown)


dgageot:
@aheritier can I close this one?

aheritier:
yes. let's close

PR #2015, opened 2026-03-09
feat: add per-toolset model routing via model field on toolsets

dgageot:
Add a 'model' field to toolset configuration that automatically routes the LLM turn processing tool results through a specified model. This enables platform-level cost optimization: cheaper/faster models handle simple tool results (file reads, knowledge-base lookups) while the agent's primary model handles complex reasoning.

Config & schema:
- Add Model field to Toolset struct in pkg/config/latest/types.go
- Add model property to Toolset in agent-schema.json

Tool plumbing:
- Add ModelOverride field to Tool struct (pkg/tools/tools.go)
- Add WithModelOverride wrapper in pkg/teamloader/model_override.go
- Wire wrapper into getToolsForAgent in pkg/teamloader/teamloader.go

Runtime:
- Extract reusable resolveModelRef from SetAgentModel in model_switcher.go with proper alloy config rejection
- Add resolveToolCallModelOverride helper (pkg/runtime/tool_model_override.go)
- In RunStream loop: after tool calls, record override; on next iteration, temporarily switch model for one turn then reset
- Clear override on agent switch to prevent cross-agent leakage

Tests:
- Unit tests for WithModelOverride wrapper
- Unit tests for resolveToolCallModelOverride
- Unit tests for resolveModelRef (alloy rejection, nil config, invalid format)
- Example config: per_tool_model_routing.yaml

Assisted-By: docker-agent

### DIFF
--- a/examples/per_tool_model_routing.yaml
+++ b/examples/per_tool_model_routing.yaml
@@ -1,0 +1,40 @@
+#!/usr/bin/env docker agent run
+
+# Per-Tool Model Routing Example
+#
+# This example demonstrates how to use the `model` field on toolsets
+# to automatically route specific tool results through a cheaper/faster
+# model, while keeping the agent's primary model for complex reasoning.
+#
+# When the LLM calls a tool from a toolset with a `model` field, the
+# next LLM turn (processing the tool results) uses the specified model
+# instead of the agent's primary model. This is a one-shot override:
+# subsequent turns return to the primary model.
+
+models:
+  primary:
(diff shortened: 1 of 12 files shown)


examples/per_tool_model_routing.yaml
```diff
@@ -0,0 +1,46 @@
+# Per-Tool Model Routing Example
+#
```
trungutt, line 4: **nit:** Could be in examples folder

PR #2006, opened 2026-03-09
fix: re-enable mouse tracking after terminal tab switch

dgageot:
## Summary

Some terminals (notably Windows Terminal) silently reset mouse tracking mode when the terminal tab/window loses and regains focus. Because bubbletea's renderer only sends mouse mode escape sequences when the requested mode changes between frames, mouse scrolling and clicks stop working after switching back to the agent's tab.

## Changes

- Enable `ReportFocus` in `toFullscreenView` so the app receives `tea.FocusMsg`/`tea.BlurMsg`
- Handle `tea.FocusMsg`: temporarily set `MouseMode` to `None` for one frame, then restore `MouseModeCellMotion` on the next frame via `mouseResetMsg`
- Handle `tea.BlurMsg`: consume the event so it doesn't get forwarded unnecessarily to child components
- This two-frame toggle forces the renderer to re-emit the mouse tracking escape sequences

Fixes #1985

PR #2017, opened 2026-03-09
tui: improve tmux experience and simplify keyboard enhancements

dgageot:
Enable **ReportFocus** and **KeyboardEnhancements.ReportEventTypes** in the fullscreen view so that tmux pane focus events and Kitty keyboard protocol features (shift+enter, key disambiguation) are requested from supporting terminals.

Simplify keyboard enhancement tracking:
- Remove redundant `keyboardEnhancementsSupported` bool field; derive it via `hasKeyboardEnhancements()` from the cached message instead.
- Stop forwarding `KeyboardEnhancementsMsg` to `chatPage`, which has no handler for it. Only the editor uses it for keybinding config.

### DIFF
--- a/pkg/tui/tui.go
+++ b/pkg/tui/tui.go
@@ -116,11 +116,10 @@ type appModel struct {
 	// Focus state
 	focusedPanel FocusedPanel
 
-	// keyboardEnhancements stores the last keyboard enhancements message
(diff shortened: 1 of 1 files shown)


trungutt:
Will be in conflict with https://github.com/docker/docker-agent/pull/2006 on merge

PR #2042, opened 2026-03-10
Fix sidebar context % flickering during sub-agent transfers

dgageot:
After a sub-agent's stream stops, `handleTaskTransfer` restores the parent agent via `AgentInfo`, but `currentSessionID` still references the child session. `currentSessionUsage()` used to prioritise `currentSessionID` unconditionally, so it returned the child's context usage instead of the parent's until the next `StreamStarted` event reset the session ID.

## Fix

Detect stale state in `currentSessionUsage()` by checking whether the session's owning agent matches `currentAgent`. When they differ, skip the session-ID lookup and fall through to the agent-name search, which finds the correct parent session.

## Test

Add a test that replays two sequential `transfer_task` round-trips and asserts `contextPercent()` stays correct at every step.

### DIFF
--- a/pkg/tui/components/sidebar/context_percent_test.go
+++ b/pkg/tui/components/sidebar/context_percent_test.go
@@ -121,3 +121,92 @@ func TestContextPercent_FallbackToSingleSession(t *testing.T) {
 
 	assert.Equal(t, "10%", m.contextPercent())
 }
+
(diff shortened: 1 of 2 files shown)


dgageot:
@maxcleme does it fix your issue?

PR #2052, opened 2026-03-11
Fix sonnet model name

dgageot:


examples/dhi/dhi.yaml
```diff
@@ -79,4 +79,4 @@ agents:
 models:
   claude:
     provider: anthropic
-    model: claude-3-7-sonnet-latest
+    model: claude-4-6-sonnet-latest
```
aheritier: strange I use `claude-sonnet-4-6`
cannot we just use `claude-sonnet`
Where is defined the catalog of models?

PR #2082, opened 2026-03-12
fix: prevent send-on-closed-channel panic in elicitationHandler

trungutt:
## Summary

- Fix a send-on-closed-channel panic in `elicitationHandler` that caused "aborted" on the second `user_prompt` elicitation
- Make `setElicitationEventsChannel` / `clearElicitationEventsChannel` calls unconditional (remove `!sess.ToolsApproved` guards)
- Add regression test `TestElicitationChannelNotStaleAfterToolsApprovedMutation`

## Problem

When `sess.ToolsApproved` transitions from `false` → `true` mid-stream (e.g. via "approve all" tool confirmation), `finalizeEventChannel` skips clearing the elicitation channel because `ToolsApproved` is now `true`. The next stream also skips setting a new channel. This leaves a stale reference to a closed channel, and the next `elicitationHandler` call panics with send-on-closed-channel.

## Fix

Remove the `if !sess.ToolsApproved` guards from both `RunStream` (where the channel is set) and `finalizeEventChannel` (where it's cleared). Every stream unconditionally gets its own channel and cleans it up at teardown. The `ToolsApproved` flag continues to control tool approval flow only.

### DIFF
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -47,11 +47,7 @@ func (r *LocalRuntime) registerDefaultTools() {
 func (r *LocalRuntime) finalizeEventChannel(ctx context.Context, sess *session.Session, events chan Event) {
 	// Clear the elicitation events channel before closing the events channel
 	// to prevent a send-on-closed-channel panic in elicitationHandler.
-	// Skip for background sessions (ToolsApproved=true) — they never set the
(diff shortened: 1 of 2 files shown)


trungutt:
Closed, in favor of https://github.com/docker/docker-agent/pull/2080

PR #2096, opened 2026-03-13
fix: make HTTP 429 retryable when no fallback model, respect Retry-After header

### LINKED ISSUE #2095: Handle HTTP 429 and 500 errors as retryable in model provider
## Problem

The docker agent model provider currently does not treat HTTP 429 (Too Many Requests) or HTTP 500 (Internal Server Error) responses as retryable errors, causing requests to fail immediately when these transient conditions occur.

### HTTP 429 (Too Many Requests)
- Treat as a retryable error **only when there is no fallback model configured**
- Respect any `Retry-After` header sent by the server in the response to determine how long to wait before retrying
- If no `Retry-After` header is present, use the existing backoff logic

### HTTP 500 (Internal Server Error)
- Treat as a retryable error unconditionally
- Use the same backoff logic as other existing retryable errors (no `Retry-After` support needed)

## Context

Discovered from agentic platform work where rate-limiting and transient server errors cause unnecessary failures that could be recovered automatically.

simonferquel-clanker:
## Summary

Fixes #2095

When the model provider returns **HTTP 429 (Too Many Requests)**:
- **With fallback models configured** → skip to the next fallback immediately (existing behaviour preserved)
- **Without fallback models** → retry the same model, honouring the `Retry-After` response header (capped at 60 s), falling back to exponential backoff when the header is absent

**HTTP 500** was already retryable via `IsRetryableStatusCode` — no change needed.

### `pkg/modelerrors/modelerrors.go`
- `ExtractRetryAfter(err error) time.Duration` — extracts `Retry-After` from `*anthropic.Error` and `*openai.Error` (both expose `*http.Response`); supports integer-seconds and HTTP-date formats
- `parseRetryAfterHeader(value string) time.Duration` — parses the header per RFC 7231 §7.1.3
- `ClassifyModelError(err error) (retryable, rateLimited bool, retryAfter time.Duration)` — unified classifier used by the retry loop; existing `IsRetryableModelError` / `IsRetryableStatusCode` unchanged for backward compatibility
- `MaxRetryAfterWait` constant (60 s cap) to prevent a misbehaving server from blocking the agent indefinitely

### `pkg/runtime/fallback.go`
- Replaced both `IsRetryableModelError` call sites with `ClassifyModelError`
- Extracted `handleModelError` helper method to avoid duplicating classification logic across stream-creation and stream-handling error paths
- Added `retryDecision` type (`Continue` / `Break` / `Return`) for clarity

### Tests
- `pkg/modelerrors/modelerrors_test.go`: `TestClassifyModelError`, `TestExtractRetryAfter`, `TestParseRetryAfterHeader`
- `pkg/runtime/fallback_test.go`: `TestFallback429WithFallbacksSkipsToNextModel`, `TestFallback429WithoutFallbacksRetriesSameModel`, `TestFallback429WithoutFallbacksExhaustsRetries`, `TestFallback500RetryableWithBackoff`

## Note on HTTP 500

HTTP 500 was already handled as retryable by `IsRetryableStatusCode(500) == true` before this PR. The requirement is satisfied by the existing code; this PR adds an explicit test (`TestFallback500RetryableWithBackoff`) to document and protect that behaviour.

### DIFF
--- a/pkg/runtime/fallback.go
+++ b/pkg/runtime/fallback.go
@@ -362,19 +362,23 @@ func (r *LocalRuntime) handleModelError(
 	retryable, rateLimited, retryAfter := modelerrors.ClassifyModelError(err)
 
 	if rateLimited {
-		if hasFallbacks {
-			// Fallbacks available → skip to next model immediately (existing behaviour).
-			slog.Warn("Rate limited with fallbacks available, skipping to next model",
+		// Gate: only retry on 429 if opt-in is enabled AND no fallbacks exist.
+		// Default behavior (retryOnRateLimit=false) treats 429 as non-retryable,
+		// identical to today's behavior before this feature was added.
+		if !r.retryOnRateLimit || hasFallbacks {
+			slog.Warn("Rate limited, treating as non-retryable",
 				"agent", a.Name(),
 				"model", modelEntry.provider.ID(),
-				"retry_after", retryAfter)
+				"retry_on_rate_limit_enabled", r.retryOnRateLimit,
+				"has_fallbacks", hasFallbacks,
+				"error", err)
 			if !modelEntry.isFallback {
 				*primaryFailedWithNonRetryable = true
 			}
 			return retryDecisionBreak
 		}
 
-		// No fallbacks → retry same model after honouring Retry-After (or backoff).
+		// Opt-in enabled, no fallbacks → retry same model after honouring Retry-After (or backoff).
 		waitDuration := retryAfter
 		if waitDuration <= 0 {
 			waitDuration = modelerrors.CalculateBackoff(attempt)
@@ -386,12 +390,13 @@ func (r *LocalRuntime) handleModelError(
 				"max", modelerrors.MaxRetryAfterWait)
 			waitDuration = modelerrors.MaxRetryAfterWait
 		}
-		slog.Warn("Rate limited without fallbacks, retrying with wait",
+		slog.Warn("Rate limited, retrying (opt-in enabled)",
 			"agent", a.Name(),
 			"model", modelEntry.provider.ID(),
 			"attempt", attempt+1,
 			"wait", waitDuration,
-			"retry_after_from_header", retryAfter > 0)
+			"retry_after_from_header", retryAfter > 0,
+			"error", err)
 		if !modelerrors.SleepWithContext(ctx, waitDuration) {
 			return retryDecisionReturn
 		}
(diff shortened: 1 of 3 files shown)


pkg/modelerrors/modelerrors.go
```diff
@@ -27,6 +30,14 @@ const (
 	backoffJitter    = 0.1
 )
 
+// maxRetryAfterWait caps how long we'll honor a Retry-After header to prevent
+// a misbehaving server from blocking the agent for an unreasonable amount of time.
+const maxRetryAfterWait = 60 * time.Second
```
rumpl: This is only used a few lines below, we can merge these two

pkg/modelerrors/modelerrors.go
```diff
@@ -296,6 +307,92 @@ func IsRetryableModelError(err error) bool {
 	return false
 }
 
+// ExtractRetryAfter extracts the Retry-After duration from an HTTP error response.
+// Works with Anthropic and OpenAI SDK error types that expose *http.Response.
+// Returns 0 if no Retry-After header is present or the error type is unsupported.
+func ExtractRetryAfter(err error) time.Duration {
```
rumpl: This logic should somehow be implemented in each provider, not here IMO

pkg/model/provider/openai/retry.go
```diff
@@ -0,0 +1,40 @@
...
+	"net/http"
+	"strconv"
+	"time"
+
+	openaisdk "github.com/openai/openai-go/v3"
+)
+
+// ExtractRetryAfter extracts the Retry-After wait duration from an OpenAI API error.
+// Returns 0 if the error is not an *openai.Error, has no response, or has no Retry-After header.
+func ExtractRetryAfter(err error) time.Duration {
```
simonferquel: This seems to be the same exact same as the anthropic implementation.
I like the idea of having each provider customize the retry after extraction, but I'd also like to avoid code duplication

pkg/model/provider/retry.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/docker/docker-agent/pkg/model/provider/anthropic"
+	"github.com/docker/docker-agent/pkg/model/provider/openai"
+)
+
+// ExtractRetryAfter extracts the Retry-After wait duration from a provider error.
+// It delegates to each provider package to keep provider-specific SDK knowledge
+// local to each provider.
+// Returns 0 if no Retry-After header is found or the error type is unsupported.
+func ExtractRetryAfter(err error) time.Duration {
```
rumpl: We shouldn't need this. Providers could return a typed error that contains everything we need so the check would become (in fallbac) a `errors.As` or whatever

simonferquel:
I like the new approach, letting @rumpl to take the final decision

simonferquel:
/review

triggering docker agent for good measure

PR #2101, opened 2026-03-13
Extract compaction into a dedicated pkg/compaction package

dgageot:
Extract compaction logic into dedicated pkg/compaction package

Move session compaction helpers (EstimateMessageTokens, ShouldCompact,
HasConversationMessages, BuildPrompt, SystemPrompt) out of pkg/runtime
into the new pkg/compaction package so the logic can be reused and
tested independently.

- Add pkg/compaction with compaction.go, compaction_test.go, and
  embedded prompt files.
- Remove sessionCompactor struct; replace with doCompact method on
  LocalRuntime and a standalone runSummarization function.
- Simplify compactIfNeeded and RunStream to call compaction.ShouldCompact.
- Delete pkg/runtime/prompts/compaction-{system,user}.txt (moved).
- Remove estimateMessageTokens and its tests from pkg/runtime.

Assisted-By: docker-agent


rumpl:
Where is the `Compactor`?

aheritier:
> The runtime's sessionCompactor becomes a thin adapter that delegates to compaction.Compactor and handles runtime-specific side-effects (events, session store persistence, token count updates).

rumpl:
Yes but there is no compaction.Compactor in the code, that's why I'm
confused

question429:
:::writing{variant=“chat_message” id=“27514”}
Good catch — there actually isn’t a compaction.Compactor type in the code.

That wording came from the automated review summary and was imprecise. The PR separates the compaction logic into the pkg/compaction package, while sessionCompactor in the runtime acts as a thin adapter that calls into that package and handles runtime-specific side effects (events, persistence, token updates).

So the separation is at the package level, not via a Compactor struct.
:::

dgageot:
I updated the comment

question429:
Thanks for the refactor and for merging this.

Extracting the compaction logic into a dedicated package makes the overall structure much clearer. Keeping the runtime layer focused on orchestration while isolating the compaction logic improves separation of concerns and should make the code easier to maintain and test going forward.

Nice improvement to the architecture.

PR #2122, opened 2026-03-16
Fix session store resource leak in api command

### LINKED ISSUE #2110: [bug] Potential resource leak due to unclosed sessionStore
## 🟠 high - bug

**File:** `cmd/root/api.go` (line 80)

### Code

```go
sessionStore, err := session.NewSQLiteSessionStore(sessionDB)
	if err != nil {
		return fmt.Errorf("creating session store: %w", err)
	}
	defer func() {
		if err := sessionStore.Close(); err != nil {
			slog.Error("Failed to close session store", "error", err)
		}
	}()
```

### Problem

The `sessionStore.Close()` is deferred, but if `server.New()` or `s.Serve()` return an error *before* the deferred function runs, the `sessionStore` will not be closed. This can lead to resource leaks, especially with SQLite, as the database file might remain locked or uncleaned.

### Suggested Fix

The `sessionStore.Close()` should be called immediately if an error occurs after its creation and before the function returns. A common pattern is to assign the `sessionStore` to a variable outside the `if err != nil` block and then close it in the error handling.

Alternatively, since `server.New` takes ownership of the `sessionStore` and implicitly closes it, the `defer` on `sessionStore.Close()` might be redundant or incorrectly placed if `server.New` is designed to handle the closure itself. However, based on the current code, it appears `server.New` *receives* the `sessionStore` but doesn't necessarily *close* it, and it's good practice to ensure resources are closed when they are created in a function. The `defer` is for when the function exits normally, so the explicit error paths also need to ensure closure.

---

maks2134:
This PR fixes a resource leak issue #2110 in the api command.

## Problem
The sessionStore.Close() was deferred, but if server.New() or s.Serve() return an error before the deferred function runs, the sessionStore will not be closed. This can lead to resource leaks, especially with SQLite, as the database file might remain locked or uncleaned.

## Solution
Remove the defer for sessionStore.Close() and explicitly close the session store in all error paths:
- After config.ResolveSources() error
- After server.New() error 
- After s.Serve() error
- On successful completion

This ensures that the SQLite database connection is always properly closed, preventing resource leaks.

Fixes #2110

### DIFF
--- a/cmd/root/api.go
+++ b/cmd/root/api.go
@@ -127,7 +127,6 @@ func (f *apiFlags) runAPICommand(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Close session store on successful completion
 	if err := sessionStore.Close(); err != nil {
 		slog.Error("Failed to close session store", "error", err)
 	}


dgageot:
Sorry @maks2134, I'm going to close this one. #2110 is a false positive. There\s no issue there

PR #2123, opened 2026-03-16
feat: framework-level tool loop detection

derekmisler:
Closes https://github.com/docker/gordon/issues/218

## Summary

- Adds a **consecutive duplicate tool call detector** that terminates agents when they issue N identical tool call batches in a row, preventing degenerate loops where the model is not making progress
- Default threshold of **5** is always active, configurable per-agent via `max_consecutive_tool_calls` in YAML
- Wired into all session creation paths: runtime loop, task transfer, background agents, A2A, MCP, ACP, and HTTP server

### Why `max_iterations` isn't sufficient

`max_iterations` counts *total* loop iterations regardless of whether the model is making progress. In [pinata run 23141381851](https://github.com/docker/pinata/actions/runs/23141381851/job/67218282226), the drafter sub-agent called `read_file("/tmp/drafter_chunk_1.diff")` **629 times consecutively** — each call succeeded (the file existed, the read returned content), so no error was raised and each iteration counted as valid work. `max_iterations` has no concept of whether iteration N is *different* from iteration N-1; it just counts.

Setting `max_iterations` low enough to catch this would kill legitimate agents that need many iterations for complex tasks. Setting it high lets stuck models burn through the entire budget doing nothing useful. The new detector solves a fundamentally different problem: not "how many steps total?" but **"is the model stuck?"** Five consecutive identical calls is a mechanical signal that the model has entered a fixed point — a framework-level guardrail that no amount of prompt confusion can bypass.

## Configuration

```yaml
agents:
  - name: drafter
    max_consecutive_tool_calls: 3  # terminate after 3 identical calls (default: 5)
```

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./pkg/runtime/...` — 6 new table-driven tests pass (varied calls, exact threshold, counter reset, empty calls, multi-tool batches, differing arguments)
- [x] Full runtime test suite passes
- [ ] Manual: run a PR review agent YAML with `max_consecutive_tool_calls: 3` and verify it parses correctly

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -184,8 +184,8 @@
         },
         "max_consecutive_tool_calls": {
           "type": "integer",
-          "description": "Maximum consecutive identical tool calls before the agent is terminated. Prevents degenerate loops. Default: 5.",
-          "minimum": 1
+          "description": "Maximum consecutive identical tool calls before the agent is terminated. Prevents degenerate loops. 0 uses the default of 5.",
+          "minimum": 0
         },
         "num_history_items": {
           "type": "integer",
--- a/pkg/agent/opts.go
+++ b/pkg/agent/opts.go
@@ -128,10 +128,11 @@ func WithMaxIterations(maxIterations int) Opt {
 }
 
 // WithMaxConsecutiveToolCalls sets the threshold for consecutive identical tool
-// call detection. Values < 1 are ignored (0 means "use runtime default of 5").
+// call detection. 0 means "use runtime default of 5". Negative values are
+// ignored.
 func WithMaxConsecutiveToolCalls(n int) Opt {
 	return func(a *Agent) {
-		if n >= 1 {
+		if n >= 0 {
 			a.maxConsecutiveToolCalls = n
 		}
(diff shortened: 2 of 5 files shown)


agent-schema.json
```diff
@@ -182,6 +182,11 @@
           "description": "Maximum number of iterations",
           "minimum": 0
         },
+        "max_consecutive_tool_calls": {
+          "type": "integer",
+          "description": "Maximum consecutive identical tool calls before the agent is terminated. Prevents degenerate loops. Default: 5.",
+          "minimum": 1
```
gtardif, line 188: shouldn't this be min 2, otherwise it will always trigger if 1? (cf [here](https://github.com/docker/docker-agent/pull/2123/changes#diff-dc1bb0544f27f506134402d2ad03c652c2c6ade3d9f7c6c5f2579513041d6567R39))

derekmisler: `1` is a valid option, if you want the exact same tool call to stop after the first run, no matter the results.

PR #2143, opened 2026-03-17
Improve edit_file resilience: handle double-serialized args and fuzzy text matching

trungutt:
## Summary

Improves the `edit_file` tool resilience against two common LLM failure modes observed in production sessions.

### Commit 1: Handle double-serialized `edits` argument

LLMs sometimes send the `edits` parameter as a JSON string (`"[{...}]"`) instead of a JSON array (`[{...}]`). This adds a custom `UnmarshalJSON` on `EditFileArgs` that detects this double-serialization and transparently unwraps it. Since the ACP filesystem handler reuses `builtin.EditFileArgs`, this fix applies to both code paths.

### Commit 2: Fuzzy text matching fallbacks

Inspired by the resilient edit matching in [OpenCode](https://github.com/anomalyco/opencode) (which uses 8+ fallback strategies in its `edit` tool), this adds fuzzy text matching to `edit_file`.

When an exact `oldText` match fails, the tool now tries progressively looser normalization strategies before returning an error:

1. **Line-trimmed** — strip trailing whitespace per line
2. **Indentation-flexible** — strip leading whitespace per line
3. **Line-continuation normalization** — collapse `\` + newline + spaces into a single space
4. **Escape-normalized** — treat `\"` and `"` as equivalent
5. **Whitespace-collapsed** — collapse all whitespace runs to a single space

These handle real-world LLM mistakes observed in session logs:
- Adding spurious indentation to `oldText` (e.g., 4 leading spaces on a Dockerfile `RUN` line)
- Collapsing multi-line shell continuations (`\` + newline) into single lines
- Unescaping `\"` to `"` when reproducing file content

When a fuzzy strategy matches, the replacement is applied to the original (un-normalized) content so surrounding text is preserved exactly. Both the builtin and ACP filesystem handlers use the shared `FindAndReplace` function.

### DIFF
--- a/pkg/tools/builtin/filesystem.go
+++ b/pkg/tools/builtin/filesystem.go
@@ -14,6 +14,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/fsx"
@@ -187,6 +188,9 @@ func (a *EditFileArgs) UnmarshalJSON(data []byte) error {
 	}
 
 	a.Path = raw.Path
+	if a.Path == "" {
+		return errors.New("path field is required")
+	}
 
 	// Try parsing edits as an array first (normal case).
 	if err := json.Unmarshal(raw.Edits, &a.Edits); err == nil {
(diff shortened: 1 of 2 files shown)


trungutt:
Splitting into two separate PRs: one for double-serialized edits fix, one for fuzzy text matching.

PR #2144, opened 2026-03-17
fix: handle double-serialized edits argument in edit_file tool

trungutt:
## Summary

LLMs sometimes send the `edits` parameter of `edit_file` as a JSON string (`"[{...}]"`) instead of a JSON array (`[{...}]`). This adds a custom `UnmarshalJSON` on `EditFileArgs` that detects this double-serialization and transparently unwraps it.

Since the ACP filesystem handler reuses `builtin.EditFileArgs`, this fix applies to both code paths.

Also validates that the `path` field is non-empty in the fallback branch, so a missing path produces a clear `"path field is required"` error instead of a confusing downstream failure.

### DIFF
--- a/pkg/tools/builtin/filesystem_test.go
+++ b/pkg/tools/builtin/filesystem_test.go
@@ -264,11 +264,12 @@ func TestEditFileArgs_UnmarshalJSON(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name      string
-		input     string
-		wantPath  string
-		wantEdits []Edit
-		wantErr   bool
+		name       string
+		input      string
+		wantPath   string
+		wantEdits  []Edit
+		wantErr    bool
+		wantErrMsg string
 	}{
 		{
 			name:     "normal array edits",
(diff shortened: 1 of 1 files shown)


pkg/tools/builtin/filesystem.go
```diff
@@ -165,6 +165,49 @@ type EditFileArgs struct {
 	Edits []Edit `json:"edits" jsonschema:"Array of edit operations"`
 }
 
+// UnmarshalJSON handles LLM-generated arguments where "edits" may be
+// a JSON string instead of a JSON array (double-serialized).
+func (a *EditFileArgs) UnmarshalJSON(data []byte) error {
+	// First, try standard unmarshaling.
+	type Alias EditFileArgs
```
rumpl: Why do we need this type alias?

pkg/tools/builtin/filesystem_test.go
```diff
@@ -260,6 +260,98 @@ func TestFilesystemTool_EditFile(t *testing.T) {
...
 
+func TestEditFileArgs_UnmarshalJSON(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name      string
+		input     string
+		wantPath  string
+		wantEdits []Edit
+		wantErr   bool
```
rumpl: The implementation returns different errors, we should check that the error returned is the right one

PR #2145, opened 2026-03-17
feat: add fuzzy text matching fallbacks for edit_file tool

trungutt:
## Summary

Inspired by the resilient edit matching in [OpenCode](https://github.com/anomalyco/opencode) (which uses 8+ fallback strategies in its `edit` tool), this adds fuzzy text matching to `edit_file`.

When an exact `oldText` match fails, the tool now tries progressively looser normalization strategies before returning an error:

1. **Line-trimmed** — strip trailing whitespace per line
2. **Indentation-flexible** — strip leading whitespace per line
3. **Line-continuation normalization** — collapse `\` + newline + spaces into a single space
4. **Escape-normalized** — treat `\"` and `"` as equivalent
5. **Whitespace-collapsed** — collapse all whitespace runs to a single space

These handle real-world LLM mistakes observed in session logs:
- Adding spurious indentation to `oldText` (e.g., 4 leading spaces on a Dockerfile `RUN` line)
- Collapsing multi-line shell continuations (`\` + newline) into single lines
- Unescaping `\"` to `"` when reproducing file content

When a fuzzy strategy matches, the replacement is applied to the original (un-normalized) content so surrounding text is preserved exactly. Index mapping between normalized and original strings uses binary search with rune-aware boundary snapping to preserve UTF-8 integrity.

Both the builtin and ACP filesystem handlers use the shared `FindAndReplace` function.

This PR is independent of #2144 (double-serialized edits fix).

### DIFF
--- a/pkg/acp/filesystem.go
+++ b/pkg/acp/filesystem.go
@@ -172,10 +172,11 @@ func (t *FilesystemToolset) handleEditFile(ctx context.Context, toolCall tools.T
 	modifiedContent := resp.Content
 
 	for i, edit := range args.Edits {
-		if !strings.Contains(modifiedContent, edit.OldText) {
+		replaced, ok := builtin.FindAndReplace(modifiedContent, edit.OldText, edit.NewText)
+		if !ok {
 			return tools.ResultError(fmt.Sprintf("Edit %d failed: old text not found", i+1)), nil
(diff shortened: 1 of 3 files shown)


trungutt:
Closed, reopen when needed

PR #2180, opened 2026-03-19
Simplify models

dgageot:


rumpl:
Do we really have to remove the fields? It's a public package and external might need it?

PR #2192, opened 2026-03-20
feat: make maxOldToolCallTokens configurable in agent YAML

gtardif:
Add max_old_tool_call_tokens configuration option to control how much historical tool call content is retained in the context. 

This allows to deactivate the old tool token truncation form the agent config file

Changes:
- Add maxOldToolCallTokens field to AgentConfig (pkg/config/latest/types.go)
- Update JSON schema with new max_old_tool_call_tokens property
- Add maxOldToolCallTokens field and getter to Agent struct
- Add maxOldToolCallTokens field to Session struct
- Update GetMessages() to use configured value or fall back to DefaultMaxOldToolCallTokens (40000)
- Propagate maxOldToolCallTokens through all session creation points:
  - A2A adapter
  - ACP agent
  - MCP server
  - Runtime agent delegation (transfer_task, sub-sessions)
  - HTTP/API server session manager

Configuration semantics:
- Not set or 0: Use default (40000 tokens)
- Positive value: Use that specific limit
- -1: Unlimited (no truncation)

When max_old_tool_call_tokens is exceeded, older tool calls have their content replaced with "[content truncated]" to keep context size manageable. Tokens are approximated as string_length/4.

Assisted-By: docker-agent

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -17,6 +17,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.opentelemetry.io/otel"
 
+	"github.com/docker/docker-agent/pkg/agent"
 	"github.com/docker/docker-agent/pkg/app"
 	"github.com/docker/docker-agent/pkg/cli"
 	"github.com/docker/docker-agent/pkg/config"
@@ -339,7 +340,7 @@ func (f *runExecFlags) createRemoteRuntimeAndSession(ctx context.Context, origin
 func (f *runExecFlags) createLocalRuntimeAndSession(ctx context.Context, loadResult *teamloader.LoadResult) (runtime.Runtime, *session.Session, error) {
 	t := loadResult.Team
 
-	agent, err := t.Agent(f.agentName)
+	agt, err := t.Agent(f.agentName)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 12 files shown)


gtardif:
Note: taken from https://github.com/docker/docker-agent/pull/1831, credits to @krissetto (only the old tool truncation bit in this PR)
Leaving the default behaviour as it is for now, but allowing to deactivate truncation from the agent config

cmd/root/run.go
```diff
@@ -494,9 +495,11 @@ func (f *runExecFlags) buildAppOpts(args []string) ([]app.Opt, error) {
 // buildSessionOpts returns the canonical set of session options derived from
 // CLI flags and agent configuration. Both the initial session and spawned
 // sessions use this method so their options never drift apart.
-func (f *runExecFlags) buildSessionOpts(maxIterations int, workingDir string) []session.Opt {
+func (f *runExecFlags) buildSessionOpts(agt *agent.Agent, workingDir string) []session.Opt {
 	return []session.Opt{
-		session.WithMaxIterations(maxIterations),
+		session.WithMaxIterations(agt.MaxIterations()),
+		session.WithMaxConsecutiveToolCalls(agt.MaxConsecutiveToolCalls()),
```
gtardif, line 501: FYI @derekmisler I noticed the WithMaxConsecutiveToolCalls was not propagated properly (had the same issue initially for WithMaxOldToolCallTokens, I added the propagation here. We didn't noticed because we only use the default value I suppose)

pkg/session/session.go
```diff
@@ -781,7 +797,16 @@ func (s *Session) GetMessages(a *agent.Agent) []chat.Message {
 		messages = trimMessages(messages, maxItems)
 	}
 
-	messages = truncateOldToolContent(messages, MaxToolCallTokens)
+	// Use configured max tokens or fall back to default constant.
+	// -1 means unlimited (no truncation).
+	// 0 or any positive value uses that as the limit.
```
rumpl: This comment is a bit misleading regarding the value when max old.. is zero

PR #2210, opened 2026-03-21
Refactor RAG from agent-level config to standard toolset type

dgageot:
Replace the special-cased RAG configuration (AgentConfig.RAG []string referencing top-level Config.RAG map) with a standard toolset type (type: rag) that follows the same patterns as MCP and other toolsets.

Key changes:
- Add 'type: rag' to toolset registry with ref support for shared definitions
- Introduce RAGToolset wrapper (mirrors MCPToolset) for top-level rag section
- Add RAGConfig field to Toolset for inline/resolved RAG configuration
- Add resolveRAGDefinitions() mirroring resolveMCPDefinitions()
- Extract rag.NewManager() for per-toolset manager creation
- Implement tools.Startable on RAGTool for lazy init and file watching
- Remove RAG special-casing from Team, LocalRuntime, and teamloader
- Add v6→v7 migration for old rag agent field to toolset entries
- Update schema, docs, and all example YAML files

Assisted-By: docker-agent

### DIFF
--- a/pkg/runtime/rag.go
+++ b/pkg/runtime/rag.go
@@ -1,36 +1,15 @@
 package runtime
 
 import (
-	"context"
 	"fmt"
 	"log/slog"
 
 	ragtypes "github.com/docker/docker-agent/pkg/rag/types"
-	"github.com/docker/docker-agent/pkg/tools"
 	"github.com/docker/docker-agent/pkg/tools/builtin"
 )
 
-// StartBackgroundRAGInit discovers RAG toolsets from agents and wires up event
-// forwarding so the TUI can display indexing progress. Actual initialization
-// happens lazily when the tool is first used (via tools.Startable).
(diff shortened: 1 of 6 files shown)


pkg/runtime/rag.go
```diff
@@ -1,105 +1,15 @@
...
-		go func() {
-			slog.Debug("Starting RAG file watcher goroutine", "rag", mgr.Name())
-			if err := mgr.StartFileWatcher(ctx); err != nil {
-				slog.Error("Failed to start RAG file watcher", "rag", mgr.Name(), "error", err)
-			}
-		}()
-	}
+// InitializeRAG is a no-op. RAG initialization is now handled
+// per-toolset via the tools.Startable interface.
+func (r *LocalRuntime) InitializeRAG(_ context.Context, _ chan Event) {
```
rumpl: Not needed any more?

PR #2242, opened 2026-03-25
Refactor compaction

dgageot:


pkg/runtime/session_compaction.go
```diff
@@ -13,72 +15,101 @@ import (
...
+			Content:   userPrompt,
+			CreatedAt: time.Now().Format(time.RFC3339),
+		},
+	}))
+
+	// Truncate the messages so that they fit in the available context limit
+	// (minus the expected max length of the summary).
+	contextLimit := max(0, int64(m.Limit.Context)-maxSummaryTokens)
+	var tokens int64
+	for i := len(messages) - 1; i >= 0; i-- {
```
rumpl: This needs more work, the cutoff could be in the middle of a tool call and we could end up with a session that contains a tool call assistant message without a tool call result.

Issue #416, opened 2025-10-04
Unable to complete OAuth flow for GitHub MCP Server (404 error)


0GiS0:
**Describe the bug**

I tried to use the GitHub MCP Server through OAuth. Although it asks for authorization to start the flow, when I try to launch it, **it returns a 404**. However, the Notion MCP Server works correctly.

**Version affected**

1.6.2

**How To Reproduce**

Configuration used for the GitHub MCP server:

```yaml
- type: mcp
  remote:
    url: https://api.githubcopilot.com/mcp/
    transport_type: streamable
````

**Expectation**

It should launch the OAuth flow successfully, just like it does with the Notion MCP server.

**Screenshots**

The following error is shown:

<img width="1065" height="240" alt="Image" src="https://github.com/user-attachments/assets/f0e9f74c-b9e4-47ca-be57-5d81a7b586bf" />

If I don't use the interactive mode:

```
 ./cagent run 02-mcp-demo.yml --env-from-file .env --debug --log-file mcp.log --tui=false "¿puedes contarme algo de mi repo 0gis0/cagent-demos?"

For any feedback, please visit: https://docker.qualtrics.com/jfe/form/SV_cNsCIg92nQemlfw

time=2025-10-04T22:52:40.223+02:00 level=DEBUG msg="🔎 [Telemetry] Enabled:" !BADKEY=true
time=2025-10-04T22:52:40.224+02:00 level=INFO msg="💬 [Telemetry] Auto-initialized telemetry" enabled=true debug=true
time=2025-10-04T22:52:40.224+02:00 level=DEBUG msg="🔎 [Telemetry] Processing telemetry event synchronously" event_type=command
time=2025-10-04T22:52:40.224+02:00 level=INFO msg="💬 [Telemetry] event" event="{\n  \"event\": \"command\",\n  \"event_timestamp\": 1759611160224,\n  \"source\": \"cagent\",\n  \"properties\": {\n    \"action\": \"run\",\n    \"args\": [\n      \"02-mcp-demo.yml\",\n      \"¿puedes contarme algo de mi repo 0gis0/cagent-demos?\"\n    ],\n    \"is_success\": true,\n    \"os\": \"darwin\",\n    \"os_language\": \"en_US.UTF-8\",\n    \"user_uuid\": \"XXXXXXXXXXXXX\",\n    \"version\": \"v1.6.2\"\n  }\n}"
[...]

PR #2274, opened 2026-03-28
feat: support explicit OAuth config for remote MCP servers

### LINKED ISSUE #2248: Support explicit OAuth configuration for remote MCP servers
## What you'd like to see

Add support for explicit OAuth configuration (`clientId`, `clientSecret`, `callbackPort`, `scopes`) in remote MCP server definitions, so that servers that **do not support [Dynamic Client Registration (RFC 7591)](https://datatracker.ietf.org/doc/html/rfc7591)** can still complete the OAuth flow.

### Proposed configuration

```yaml
toolsets:
  - type: mcp
    remote:
      url: https://mcp.slack.com/mcp
      transport_type: streamable
      oauth:
        clientId: "3660753192626.8903469228982"
        clientSecret: ""          # optional, for confidential clients
        callbackPort: 3118        # optional, fixed port for the OAuth callback server
        scopes:                   # optional, override scopes to request
          - search:read.public
          - chat:write
```

This mirrors what other MCP clients already support. For example, the [Slack MCP plugin](https://github.com/slackapi/slack-mcp-plugin) uses this format for **Claude Code** (`.mcp.json`):

```json
{
  "mcpServers": {
    "slack": {
      "type": "http",
      "url": "https://mcp.slack.com/mcp",
      "oauth": {
        "clientId": "1601185624273.8899143856786",
        "callbackPort": 3118
      }
    }
  }
}
```

And for **Cursor**:

```json
{
  "mcpServers": {
    "slack": {
      "url": "https://mcp.slack.com/mcp",
      "auth": {
        "CLIENT_ID": "3660753192626.8903469228982"
      }
    }
  }
}
```

## Why you'd like to see it

Many real-world MCP servers require OAuth but **do not support Dynamic Client Registration**. The Slack MCP server is a prominent example — its [official documentation](https://docs.slack.dev/ai/slack-mcp-server) explicitly states:

> *"We do not support SSE-based connections or Dynamic Client Registration at this time."*

The current OAuth implementation in docker-agent always attempts dynamic registration first. When the server's metadata lacks a `registration_endpoint`, `validateAndFillDefaults()` fabricates one (appending `/register` to the auth server URL), which then fails. When it does exist and the server doesn't support it, the registration POST fails with an HTTP error. In both cases the error is fatal — the MCP toolset fails to start and the agent silently loses all tools from that server.

This same root cause is behind issue #416 (GitHub MCP Server OAuth 404) and affects any server that requires a pre-registered app identity (Slack, potentially others).

### Detailed failure trace (Slack example)

1. Agent connects to `https://mcp.slack.com/mcp` → server returns **HTTP 401** with `WWW-Authenticate: Bearer resource_metadata="https://mcp.slack.com/.well-known/oauth-protected-resource"`
2. `oauthTransport` initiates managed OAuth flow
3. Fetches auth server metadata from `https://mcp.slack.com/.well-known/oauth-authorization-server` — response has **no** `registration_endpoint`
4. `validateAndFillDefaults()` sets `RegistrationEndpoint = "https://mcp.slack.com/register"` (a fabricated default)
5. `RegisterClient()` POSTs to that URL → Slack returns a **302 redirect** to an HTML page (not an RFC 7591 endpoint)
6. Registration fails → `handleManagedOAuthFlow` returns error (the `TODO(rumpl): fall back to requesting client ID from user` path)
7. `oauthTransport.RoundTrip` returns `"OAuth flow failed: ..."` → `doStart()` fails → `ensureToolSetsAreStarted` logs a warning and skips the toolset
8. The agent proceeds with **zero MCP tools** from the Slack server — all tool calls return "Tool not available"

### Code references

- `pkg/tools/mcp/oauth.go` — `validateAndFillDefaults()` fabricates registration endpoint; `handleManagedOAuthFlow()` has the `TODO` fallback comment
- `pkg/tools/mcp/oauth_helpers.go` — `RegisterClient()` performs the failing dynamic registration
- `pkg/config/latest/types.go` — `Remote` struct currently only has `URL`, `TransportType`, and `Headers` (no OAuth fields)

## Workarounds?

No working workaround exists. Passing `CLIENT_ID` as an HTTP header (as the Cursor config does) sends it on every request but the OAuth code never reads it to use as the `client_id` in the authorization flow:

```yaml
# Does NOT work — CLIENT_ID header is ignored by the OAuth logic
remote:
  url: https://mcp.slack.com/mcp
  transport_type: streamable
  headers:
    "CLIENT_ID": "3660753192626.8903469228982"
```

## Additional context

- Related issue: #416 (same root cause — dynamic registration fails for GitHub MCP Server)
- Slack MCP server docs: https://docs.slack.dev/ai/slack-mcp-server
- Slack MCP plugin repo (Claude Code + Cursor configs): https://github.com/slackapi/slack-mcp-plugin
- The `validateAndFillDefaults()` function should probably **not** fabricate a `registration_endpoint` when the server metadata does not include one — that is a separate but related bug

iElsha:
# Support explicit OAuth configuration for remote MCP servers

Closes #2248
Related: #416

## Problem

Remote MCP servers that do not support Dynamic Client Registration (RFC 7591), such as Slack and the GitHub MCP Server, were completely broken. More details on #2248

## Solution

Add an `oauth` block to the `remote` config that lets users supply pre-registered credentials, and fix the OAuth flow so it no longer fabricates a `registration_endpoint` when the server doesn't advertise one.

Overall the approach is working but had design decisions taken that you might want to validate 
1. No persistence across session, it uses the inMemory storage, meaning oAuth flows needs to run once per sesssion. (Improvement can be in a follow-up PR)
2. If the user declines the oAuth flows, he can not retry it for this session (same - improvement in a follow-up PR)
3. oAuth flow is triggered after the first interaction, not on start-up. I had issues getting the elicitation handler wired up at this moment, if you know a better way happy to discuss.
4. `TLS` (https) is required for some provider on the oAuth callback. The user will see a warning on the browser when attempting to validate because a temporary certificate is created on the runtime. Most likely a secured certificate should be used.

Hence this PR is putting the fundation for the OAuth supports but it needs a bit of polishing. 

**Demo video for slack MCP server:** https://streamable.com/y3i7mx

### Config format

Below format is following the specs given by @aheritier with one exception, the addition of a `tls` flag required

```yaml
toolsets:
  - type: mcp
    remote:
      url: https://mcp.slack.com/mcp
      transport_type: streamable
      oauth:
        clientId: "<your-app-client-id>"
        clientSecret: "<your-app-client-secret>"   # optional, for confidential clients
        callbackPort: 3118                          # optional, pins the local redirect port
        tls: true                                   # optional, required by some providers (e.g. Slack)
        scopes:                                     # optional, overrides server-advertised scopes
          - channels:read
          - chat:write
          - search:read.public
```

---

*AI generated change-log bellow that helps understanding the changes*

### Config (`pkg/config/latest/`)
- Added `RemoteOAuthConfig` struct with fields `ClientID`, `ClientSecret`, `CallbackPort`, `Scopes`, `TLS`
- Added `OAuth *RemoteOAuthConfig` field to the `Remote` struct
- Added validation: `oauth` requires `type: mcp`, a `url`, a non-empty `clientId`, and `callbackPort >= 0`
- `TLS` is required for some provider like Slack, a certificate is created on runtime to satisfy the needs.

### OAuth flow (`pkg/tools/mcp/oauth.go`)
- **Removed the fabricated `registration_endpoint`** — `validateAndFillDefaults()` and `createDefaultMetadata()` no longer invent an endpoint the server never advertised
- `handleManagedOAuthFlow()` now follows three paths: explicit `clientId` → use directly; server advertises `registration_endpoint` → dynamic registration; neither → fail with a clear actionable error
- Explicit credentials force the managed flow regardless of the `managed` flag
- `scopes` from config override `authServerMetadata.ScopesSupported` when provided
- On user decline, a `UserDeclined` sentinel is stored in the token store — `RoundTrip` checks this first on every subsequent request and returns immediately without re-prompting

### Callback server (`pkg/tools/mcp/oauth_server.go`, `oauth_helpers.go`)
- `NewCallbackServerWithOptions(port, ...opts)` + `WithTLS()` functional option — wraps the TCP listener with an in-memory self-signed TLS cert (required by Slack, which rejects `http://` redirect URIs even for loopback)
- `GetRedirectURI()` returns `https://` when TLS is enabled
- Fixed `ExchangeCodeForToken()`: `io.ReadAll` was draining the body before `json.NewDecoder`, making the primary decode path dead code — now uses `json.Unmarshal` directly

### Token store (`pkg/tools/mcp/tokenstore.go`)
- Added `UserDeclined bool` to `OAuthToken` as a session-scoped sentinel to permanently suppress re-prompting after a decline

### Runtime (`pkg/runtime/runtime.go`)
- `emitToolsProgressively()` defers OAuth toolsets — they are not started until `RunStream` is active and an elicitation handler is available

### Wiring (`pkg/tools/mcp/mcp.go`, `pkg/tools/mcp/remote.go`, `pkg/teamloader/registry.go`)
- `NewRemoteToolset()` accepts `oauthConfig`; `RequiresOAuth()` signals the runtime to defer startup
- `registry.go` passes `toolset.Remote.OAuth` through to the toolset

### Schema (`agent-schema.json`)
- Added `RemoteOAuthConfig` definition with `required: ["clientId"]`, `minimum: 0` on `callbackPort`, `additionalProperties: false`

### Unit tests

- **`TestOAuthTransport_UserDeclined`** — end-to-end against a real `httptest.Server`: elicitation fires exactly once on first request, zero times on all subsequent requests after decline
- **`TestOAuthUserDeclinedSentinel`** — three parallel subtests: sentinel written to store correctly, scoped to the declined URL only, not misclassified as expired by `IsExpired()`
- **`TestValidateAndFillDefaults`** — guards against regression of the fabricated endpoint; includes a case asserting `RegistrationEndpoint` stays empty when the server omits it
- **`TestToolset_Validate_RemoteOAuth`** — covers all validation paths: valid config, missing `clientId`, negative `callbackPort`, wrong toolset type, missing `url`
- **`TestEmitStartupInfo_OAuthToolsetDeferred`** / **`TestEmitStartupInfo_NonOAuthToolsetStartedEagerly`** — verify the deferred-start behaviour in the runtime
- **`TestSelfSignedTLSConfig`**, **`TestGetRedirectURI`**, **`TestNewCallbackServer_*`**, **`TestBuildAuthorizationURL`** — cover the new callback server and helper paths

### Manual — Slack MCP Server (end-to-end)

Tested against the [Slack MCP Server](https://docs.slack.dev/ai/slack-mcp-server) using a Slack app registered in the Docker workspace. Config used:

```yaml
toolsets:
  - type: mcp
    remote:
      url: https://mcp.slack.com/mcp
      transport_type: streamable
      oauth:
        clientId: "<redacted>"
        clientSecret: "<redacted>"
        callbackPort: 3118
        tls: true
        scopes:
          - channels:read
          - chat:write
          - search:read.users
          - search:read.public
```

**Flow verified:**

1. Agent starts — Slack toolset is deferred (0 tools reported at startup, no `no elicitation handler configured` error)
6. First interaction — browser prompt opens, Slack OAuth consent screen loads over `https://127.0.0.1:3118/callback`
7. User authorises — token exchanged, Slack tools loaded, agent confirms tool count updated
8. Successfully called `search:read.public` to search channels and `chat:write` to send a message
9. Session ended, new session started — OAuth prompt appears again (token store is per-session, not persisted)

**Decline flow verified:**

1. OAuth prompt appears on first interaction
2. User clicks "Decline" in the elicitation UI
3. Agent responds that the Slack toolset is unavailable
4. User sends a second message — **no OAuth prompt appears again** for the rest of the session
5. Confirmed via debug logs: `RoundTrip` returns `OAuth authorization was declined for this session` immediately on requests 2+, elicitation handler call count stays at 1

## Other minor Design decisions

- **`UserDeclined` reuses the token store rather than adding state to the transport** — the `InMemoryTokenStore` already acts as the per-server auth state map; storing a sentinel there keeps the short-circuit check in one place (`RoundTrip`) with no new fields or coordination needed elsewhere.
- **`callbackPort: 0` means random port** — standard Unix convention for "give me any free port". Validation allows `>= 0`; the callback server treats `> 0` as pinned and `0` as random, avoiding the need for a separate sentinel value.
- **`scopes` fully override, not merge** — merging would require deduplication and could silently add scopes the user didn't intend. Full override gives deterministic control, which matters for security-sensitive scope selection.
- **Explicit `clientId` skips dynamic registration entirely** — if credentials are already provided there is no reason to attempt registration. Trying it first would make the explicit config unreliable if the server happens to also advertise a `registration_endpoint`.


dgageot:
merged with https://github.com/docker/docker-agent/pull/2394

PR #2296, opened 2026-03-31
feat: add fuzzy matching fallbacks for edit_file tool

trungutt:
## Summary

LLMs frequently confuse JSON-level escaping with content-level escaping when generating `edit_file` tool calls — e.g. sending `echo "brew install failed"` as `oldText` when the file contains `echo \"brew install failed\"`. The current exact `strings.Contains` check fails, and the model repeats the same mistake on retry.

Replace bare `strings.Contains` with a cascading replacer chain (inspired by [opencode's edit tool](https://github.com/anomalyco/opencode)). Both the builtin and ACP `edit_file` handlers share the same `FindMatch` function.

## How it works

`FindMatch` tries strategies in order — the first to find exactly one unique match wins:

| # | Strategy | What it catches |
|---|----------|----------------|
| 1 | exact | Direct substring (existing behavior) |
| 2 | escape-normalized | `\"` vs `"`, `\n` vs newline, `\t` vs tab, `\$` vs `$` |

If a strategy finds multiple matches it is skipped. If no strategy matches, an error is returned. The chain is designed to be extended with more strategies as new failure patterns are observed.

## Motivation

Observed in production: a session attempted to edit a Dockerfile three times, failing each time on the same `echo \"brew install failed\"` vs `echo "brew install failed"` mismatch. The model re-read the file and retried with identical arguments because it could not self-correct the escaping. It eventually gave up and used `write_file` to overwrite the entire file.

With this change, attempt 1 would succeed via the escape-normalized replacer.

### DIFF
--- a/pkg/acp/filesystem.go
+++ b/pkg/acp/filesystem.go
@@ -173,10 +173,11 @@ func (t *FilesystemToolset) handleEditFile(ctx context.Context, toolCall tools.T
 	modifiedContent := resp.Content
 
 	for i, edit := range args.Edits {
-		if !strings.Contains(modifiedContent, edit.OldText) {
-			return tools.ResultError(fmt.Sprintf("Edit %d failed: old text not found", i+1)), nil
+		match, err := builtin.FindMatch(modifiedContent, edit.OldText)
+		if err != nil {
+			return tools.ResultError(fmt.Sprintf("Edit %d failed: %s", i+1, err)), nil
 		}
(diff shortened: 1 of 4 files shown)


dgageot:
@trungutt I think we said we didn't want to do that at some point in the past. @rumpl does it ring a bell?

trungutt:
I open the PR again as edit_file still gives error on diff detection. We need to fix that, either by this PR or other PRs

dgageot:
@trungutt I'm going to close this one. It's too dangerous

PR #2316, opened 2026-04-03
Underline URLs on mouse hover

dgageot:
Track hovered URL span in the messages component and apply underline styling during rendering. Switch mouse mode from CellMotion to AllMotion so hover events fire without a button held.

Assisted-By: docker-agent

### DIFF
--- a/pkg/tui/components/messages/lineutil.go
+++ b/pkg/tui/components/messages/lineutil.go
@@ -1,0 +1,26 @@
+package messages
+
+import (
+	"charm.land/lipgloss/v2"
+	"github.com/charmbracelet/x/ansi"
+	"github.com/mattn/go-runewidth"
+)
+
+// styleLineSegment applies a lipgloss style to the portion of a line between
+// startCol and endCol (display columns), preserving the text before and after.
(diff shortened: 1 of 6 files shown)


krissetto:
didn't this have other side-effects in the past? also i'm not sure if it's the most reliable solution across terminals, i'll give it a test on some that I've notice being problematic

krissetto:
On macOS, it's all good using `ghostty`, `warp`, `iterm2`, and `Terminal.app`

Maybe it was windows that had issues with the `AllMotion` mouse mode 🤔? 

If we could get a test at least on Windows Terminal that'd be great, I won't have access to one until Monday at the best.. Maybe the issue I remember had gotten fixed in the upstream charm libs as well

dgageot:
@simonferquel could you give it a try?

simonferquel:
What is the test case?

dgageot:
@simonferquel Ask the agent to print the url of google. Make sure it's underlined and clickable when you move the mouse over it

simonferquel (approved):
Looks good on Windows terminal

Issue #2257, opened 2026-03-26
[enhancement] Slash Command auto complete vs. auto execute


steilerDev:
I'm using commands on my agent - some of them expect arguments - however when pressing tab to autocomplete the suggestion, it always immediately launches the command (and not actually auto completes it). This is in my opinion counter intuitive.

aheritier:
I just faced it. Agree it is painful. I will verify if @yunus25jmi1 PR addresses correctly the issue cc @dgageot 
My workaround is by using the user_prompt tool which is allowing the agent to ask me the required parameters

k33g:
same issue with the skills #1819

aheritier:
fixed by #2325

PR #2275, opened 2026-03-29
fix: Slash Command Autocomplete vs. Auto-Execute Fix (#2257)

yunus25jmi1:
I have successfully implemented a root-cause fix for the issue where pressing Tab to autocomplete a slash command would immediately execute it instead of just completing the text.

Root Cause Analysis:
In pkg/tui/components/completion/completion.go, both Enter and Tab were bound to the same key binding, causing auto-submit completions (like slash commands) to execute immediately upon selection with either key.

Solution Implemented:
- Separate Tab and Enter key bindings in completion.go.
- Tab now autocompletes text only (AutoExecute=false).
- Enter executes the command (AutoExecute=true).
- Updated editor.go to handle the AutoExecute flag.
- Added comprehensive tests in pkg/tui/components/completion/autocomplete_test.go.

Verification:
- pkg/tui/components/completion tests: 6/6 PASS
- pkg/tui/components/editor tests: PASS
- Successful build of project and binary

Fixes #2257

Issue #1819, opened 2026-02-21
[ux] skills selection


k33g:
When typing `/` and the first letters of the skill's name, it's impossible to use <kbd>Tab</kbd> to complete the name and then resume with the parameters of the skills

<img width="1375" height="347" alt="Image" src="https://github.com/user-attachments/assets/570e7f53-b094-4de9-afcc-9d2bf6fd638d" />

dgageot:
This is fixed on main

dgageot:
Maybe the arguments are not handled, though...

k33g:
The reading of the skill is triggered since I use `/skill_name` with `tab`: the use of the `tab` key triggers both the reading of the skill and the execution of the skill.
The expected workflow would be (eg: the name of the skill is `brave_news_search`):
- I type `/brave_n` + `tab`
- autocomplete of the skill name in the user prompt field + " " 
- no skill execution until the user fills the arguments and types "Enter"

k33g:
Hello, this one makes the skill UX a little bit annoying

yunus25jmi1:
Hi @k33g, I noticed this issue is being addressed by two PRs: 
- #2275 (my implementation - focused fix with AutoExecute flag)
- #2325 (by @joshbarrington - broader refactor with Update() driven approach)

Since both aim to fix the same issue, could you provide direction on which approach you prefer? I'm happy to close #2275 in favor of #2325 if that's the preferred direction. Thanks!

k33g:
Hello @yunus25jmi1 I'm not on the docker-agent development team, so it's not up to me to decide what the best solution is.
👋 @aheritier

dgageot:
This has been fixed by #2325

PR #2325, opened 2026-04-04
Fix: allow tab auto-complete without executing

joshbarrington:
This fixes issues:
https://github.com/docker/docker-agent/issues/2257
https://github.com/docker/docker-agent/issues/1819

There was another implementation for a fix here: https://github.com/docker/docker-agent/pull/2275

However I expanded upon that by simplifying and removing AutoComplete from the command interface, as this was hardcoded to be always true. It is now driven by the Update function directly.

I also added some more exhaustive tests and changes suggested through mise lint.

### DIFF
--- a/pkg/tui/components/editor/completion_autosubmit_test.go
+++ b/pkg/tui/components/editor/completion_autosubmit_test.go
@@ -84,4 +84,102 @@ func TestEditorHandlesAutoSubmit(t *testing.T) {
 		// It should also clear the trigger and completion word from textarea
 		assert.Empty(t, e.textarea.Value(), "should clear the trigger and completion word")
 	})
+
+	t.Run("@ completion inserts value even if AutoSubmit is true", func(t *testing.T) {
+		t.Parallel()
+
+		e := newTestEditor("@he", "he")
+		e.currentCompletion = &mockCompletion{trigger: "@"}
+
+		msg := completion.SelectedMsg{
+			Value:      "@hello",
+			AutoSubmit: true,
+		}
+
+		_, cmd := e.Update(msg)
+
+		// Command should be nil because atCompletion is true, preventing AutoSubmit behavior
+		assert.Nil(t, cmd)
+
+		// Value should have trigger replaced with selected value and a space appended
+		assert.Equal(t, "@hello ", e.textarea.Value())
+	})
+
+	t.Run("@ completion adds file attachment", func(t *testing.T) {
+		t.Parallel()
+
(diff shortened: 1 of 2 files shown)


yunus25jmi1:
Thanks for referencing #2275 and building on it! I can see this version simplifies the implementation by moving the execute behavior into Update() directly and expands the tests as well. This looks like a cleaner approach to fix issues #2257 and #1819. I'm happy to defer to this implementation if maintainers prefer this direction. Looking forward to the review!

krissetto:
Hey! Thanks for the contributions @joshbarrington && @yunus25jmi1 

Both here and in #2275 we need to pay attention to not create regressions. As-is, this breaks the UX in other parts of the app. 

Auto-complete in it's current shape is not only used for `/commands`, but also for things like `@attachments`. 

With these changes:
- if a user types `@` and then hits tab on `Browse files..` the file picker will no longer open, the `@` will disappear and a single space will be left instead. That's confusing;
- if they press `enter` on a file the message with the attachment gets sent immediately. That's also undesired, as the goal is to autocomplete the filename and attach it instead, not attach and send in a single action because the user might want to continue their message before sending it.

We need to think about supporting both scenarios, so `/commands` are a bit less annoying, and `@attachments` continue to work as they were designed to

Not sure who wants to take on the changes, but we're glad to get changes merged as soon as they don't unintentially break other types of auto-complete :)

---

Screencast below shows the problematic behavior:

https://github.com/user-attachments/assets/616c1668-7e21-479c-9a9e-331707249911

joshbarrington:
@krissetto Thanks for the heads up on the regression. Are there any other inputs such as the '@' symbol that we need to be aware of?

joshbarrington:
@krissetto I've added changes here that allow both sets of behaviour. It currently parses whether the input has an `@` and adjusts the logic accordingly, as my implementation removes the hardcoded autosubmit method of the Completion interface (this felt unintiuitive to me for it to be always true, when no longer it is). 

If there are further considerations required beyond just `@` characters then it may need to be adjusted. I'm happy to discuss alternative implementations.

dgageot:
@krissetto wdyt?

joshbarrington:
@krissetto I can't actually replicate this behaviour. If i type '@' and then follow up with tab it executes and opens the menu, the same behaviour as Enter.


https://github.com/user-attachments/assets/d643d28f-11e8-48f8-9c0f-de2999bb5e82

krissetto:
@joshbarrington my bad, i must've accidentally tried a binary built from a different src. Tested again on main just now and all good

PR #2329, opened 2026-04-05
rag/treesitter: add nocgo build support

shaj13:
I'm building a CLI tool that uses docker-agent but doesn't need RAG functionality. Currently, the tree-sitter dependency forces CGO even when RAG isn't used, which causes Go stdlib and other 3rd party packages to also be built with CGO.

This PR adds a `!cgo` build variant for the treesitter package. Applications that don't use RAG can now build with `CGO_ENABLED=0`. If RAG is used without CGO, it returns a clear error explaining the requirement.

### DIFF
--- a/pkg/rag/treesitter/treesitter.go
+++ b/pkg/rag/treesitter/treesitter.go
@@ -1,3 +1,5 @@
+//go:build cgo
+
 package treesitter
 
 import (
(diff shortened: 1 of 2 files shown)


krissetto:
sry, could you rebase this? not at my pc now. i'll reapprove after that

shaj13:
> sry, could you rebase this? not at my pc now. i'll reapprove after that

@krissetto  Git tree and branch are up to date. Could this be due to workflow approval, or because the PR is coming from a fork?

Screenshot: 
image

Fork: 
image

shaj13:
@krissetto  & @dgageot Thanks for the approval.  
Whenever you have a moment, could you please proceed with merging this PR?

PR #2331, opened 2026-04-06
session: add WithID option for custom session IDs

shaj13:
Allow setting the session ID explicitly instead of relying on the default UUID generation. In some circumstances the session ID needs to be prefixed, or comes from an external source.

Apply options before generating UUID, allowing callers to set a custom session ID via WithID. If no ID is provided, a UUID is generated.

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -579,6 +579,13 @@ func WithParentID(parentID string) Opt {
 	}
 }
 
+// WithID sets the session ID. If not set, a UUID will be generated.
+func WithID(id string) Opt {
+	return func(s *Session) {
+		s.ID = id
+	}
+}
+
 // WithExcludedTools sets tool names that should be filtered out of the agent's
 // tool list for this session. This prevents recursive tool calls in skill
 // sub-sessions.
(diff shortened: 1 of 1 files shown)


pkg/session/session.go
```diff
@@ -660,6 +663,12 @@ func New(opts ...Opt) *Session {
 		opt(s)
 	}
 
+	if s.ID == "" {
```
rumpl: If we keep setting the id no matter what above we don't need this if at all

rumpl: Basically only the log needs to go _after_ the opts are called

shaj13: Then it becomes a **breaking change**: 

The session would be created with an empty ID if WithID isn’t passed, since it’s currently **optional**.

Removing it would require updating the entire codebase to always call 
```go
s := session.New(session.WithWorkingDir(a.wd), session.WithID(uuid.New().String()),)
```
which would be frustrating. 

Additionally, since the docker-agent can be used as a library, this could break other applications that depend on it. 


FYI, I can set the session ID externally by doing s.ID = myID, but that’s not ideal since I want the debug output to reflect the correct ID.

rumpl: There is no breaking change if you put back setting the id as it was before, and then once we loop over opts, if someone added withid it would override the id

PR #2342, opened 2026-04-07
Add glob_files tool for finding files by name pattern

trungutt:
## Summary

- Adds a new built-in `glob_files` tool that lets the agent find files by glob pattern (e.g. `**/*.go`, `src/**/*.ts`, `*.json`)
- Results are sorted by modification time (most recent first) and capped at 100 files
- Respects VCS ignore rules (`.gitignore`)
- Mirrors the `Glob` tool found in other coding agents (e.g. Claude Code, OpenCode)

Assisted by OpenCode


dgageot:
@trungutt we used to have that but the llm would use bash instead

trungutt:
Yep, current it's not blocking, it's nice to have. Otherwise the LLM will do `shell something`

derekmisler:
/review (i'm just testing the action, @trungutt, you can ignore this comment)

PR #2346, opened 2026-04-07
Allow to extend and customize TUI with additional commands

shaj13:
## Problem

Customizing the TUI currently requires forking the repository or building a UI from scratch. This limits docker-agent's potential to serve as an end-to-end framework where consumers can reuse the TUI while adding their own commands.

## What was done

- Added `Immediate` flag to `commands.Item` to control whether a command interrupts an ongoing stream
- Refactored `ParseSlashCommand` to a `Parser` struct with `NewParser` constructor, allowing consumers to build parsers with custom command categories
- Added `WithCommandBuilder` option to allow consumers to provide custom command categories to the TUI

## Dependency Resolution

At first, I started with an interface that `appModel` implements which returns internal objects like `*app.App`, `*service.SessionState`, `chat.Page`, and `editor.Editor`. However, that interface can't be set in the main package as it may not serve future changes, unless it's set in a dedicated package with a guarantee there will be no circular imports.

So I added a dynamic dependency resolution pattern via `core.Resolve[T]` - a type-safe generic function that retrieves dependencies from any `tea.Model` implementing `Resolve(any) any`. 

**It’s better and more idiomatic to avoid this approach. Instead, define a well-declared interface that appModel implements to return the underlying objects, and place that interface in its own standalone package.**

Usage:
```go
app := core.Resolve[*app.App](model)
sessionState := core.Resolve[*service.SessionState](model)
chatPage := core.Resolve[chat.Page](model)
```

---

This may not be the final implementation as I am looking for feedback. I'm not sure if you're willing to allow TUI customization, but I think it would be very useful.

image

image

### DIFF
--- a/pkg/tui/core/core.go
+++ b/pkg/tui/core/core.go
@@ -62,11 +62,11 @@ func CmdHandler(msg tea.Msg) tea.Cmd {
 // Notice: This function is EXPERIMENTAL and may be changed or removed in a
 // later release.
 func Resolve[T any](m tea.Model) T {
-	if r, ok := m.(interface{ Resolve(any) any }); ok {
+	if r, ok := m.(interface{ Resolve(key any) any }); ok {
 		if v := r.Resolve(any((*T)(nil))); v != nil {
 			return v.(T)
 		}
 		panic(fmt.Sprintf("tui/core: model cannot provide type %T", *new(T)))
 	}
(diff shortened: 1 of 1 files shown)


shaj13:
@dgageot mind taking a look ? Would appreciate your thoughts.

Issue #2223, opened 2026-03-23
Allow user messages to steer the running agent mid-work instead of queuing until completion


trungutt:
## Problem

When the agent is busy (executing tool calls, streaming LLM responses), any user messages are queued and only processed **after** the entire agent run completes. Each queued message then starts a completely new `app.Run()` cycle.

This means the user cannot steer or redirect the agent while it is working. For example:
- The agent starts down the wrong path — the user has to wait for it to finish before correcting it
- The user wants to add context ("also check the tests directory") — the agent won't see it until its current run ends
- Follow-up instructions pile up and execute sequentially as separate runs, losing conversational continuity

Ideally, a message sent while the agent is working should be **injected into the running conversation** so the LLM sees it on its next iteration and can act on it immediately.

## Current Architecture (why this is feasible)

The core already has the building blocks:

1. **`sess.GetMessages(a)` is called on every loop iteration** (`pkg/runtime/loop.go:265`) — if a user message is added to `sess.Messages` between iterations, the next LLM call will see it automatically.
2. **`sess.AddMessage()` is mutex-protected** (`pkg/session/session.go:294-298`) — safe to call from another goroutine while the loop is running.
3. The only barrier is `loop.go:366`: `if res.Stopped { break }` — when the model has no tool calls, the loop exits. A mid-work message never gets a chance to override this.

## Proposed Approach

**1. `pkg/session/session.go` — Add a pending-message channel**

Add a buffered `chan *Message` field (`PendingMessages`) to `Session` with a public `InjectMessage()` method. Any caller can push a user message into the running session.

**2. `pkg/runtime/loop.go` — Drain pending messages between iterations**

After `processToolCalls` (line 343) and before the `res.Stopped` check (line 366), non-blocking drain the pending channel:

```go
injected := false
if sess.PendingMessages != nil {
    for {
        select {
        case msg := <-sess.PendingMessages:
            sess.AddMessage(msg)
            events <- UserMessage(...)
            injected = true
        default:
            goto drained
        }
    }
drained:
}

if res.Stopped && !injected {
    r.executeStopHooks(...)
    break
}
// If injected: loop continues, next iteration re-reads messages via GetMessages
```

When `injected == true`, the loop does **not** break even though the model said "stop." The next iteration re-reads messages and the LLM sees the new user message in context.

**3. `pkg/session/session.go` — Mark injected messages**

[...]

PR #2363, opened 2026-04-09
Add mid-turn message steering for running agent sessions

trungutt:
## Summary

Addresses #2223.

- Add `POST /sessions/:id/steer` API endpoint that injects user messages into a running agent session mid-turn
- Add `POST /sessions/:id/followup` API endpoint that queues messages for end-of-turn processing (one per turn)
- Agent loop drains steered messages after tool execution and before the stop-condition check, so the LLM sees them on the next iteration
- Follow-up messages are popped one at a time after the model stops and stop-hooks run, giving each a full undivided agent turn
- Emits `user_message` SSE events when messages are picked up, so API clients can confirm delivery to the user
- Guard `RunSession` against concurrent streams on the same session

This is a common pattern in agentic coding tools where the user can steer or provide follow-up context while the agent is executing tool calls. No behavioral change to the TUI — the existing client-side message queue continues to work as before. The TUI can adopt mid-turn steering in a follow-up PR.

## Current flow (before this PR)

Messages sent while the agent is busy are held client-side and only processed after the entire turn finishes. Each queued message starts a completely new turn.

```mermaid
sequenceDiagram
    participant User
    participant Client
    participant Agent Loop

    User->>Client: "fix the bug"
    Client->>Agent Loop: POST /sessions/:id/agent (start stream)
    activate Agent Loop

    Note over Agent Loop: LLM call → tool calls → LLM call → ...

    User->>Client: "also update the tests"
    Note over Client: Message held in<br/>client-side queue

    User->>Client: "use pytest not unittest"
    Note over Client: Message held in<br/>client-side queue

    Agent Loop-->>Client: StreamStopped
    deactivate Agent Loop

    Client->>Agent Loop: POST /sessions/:id/agent ("also update the tests")
    activate Agent Loop
    Agent Loop-->>Client: StreamStopped
    deactivate Agent Loop

    Client->>Agent Loop: POST /sessions/:id/agent ("use pytest not unittest")
    activate Agent Loop
    Agent Loop-->>Client: StreamStopped
    deactivate Agent Loop
```

### The problem

- The agent finishes its work **without knowing** the user wanted tests updated or pytest used — it may have already written unittest-based tests
- Each queued message starts a **separate turn**, so the agent does redundant work (e.g. rewrites tests it just wrote)
- The user has **no way to course-correct** the agent while it's working

## New flow (this PR)

Two distinct mechanisms for two distinct intents:

### Steer — urgent mid-turn injection

```mermaid
sequenceDiagram
    participant User
    participant Client
    participant Steer Queue
    participant Agent Loop

    User->>Client: "fix the bug"
    Client->>Agent Loop: POST /sessions/:id/agent (start stream)
    activate Agent Loop

    Note over Agent Loop: LLM call → tool calls executing...

    User->>Client: "use pytest not unittest"
    Client->>Steer Queue: POST /sessions/:id/steer
    Note over Steer Queue: Message enqueued

    Note over Agent Loop: Tool calls finish

    Steer Queue-->>Agent Loop: Drain all (mid-turn)
    Note over Agent Loop: Injects message in<br/>system-reminder tags,<br/>forces loop to continue

    Note over Agent Loop: LLM call (sees steered message) → ...

    Agent Loop-->>Client: StreamStopped
    deactivate Agent Loop
```

The agent sees "use pytest not unittest" **before** writing any tests — no wasted work.

### Follow-up — end-of-turn queue

```mermaid
sequenceDiagram
    participant User
    participant Client
    participant FollowUp Queue
    participant Agent Loop

    User->>Client: "fix the bug"
    Client->>Agent Loop: POST /sessions/:id/agent (start stream)
    activate Agent Loop

    User->>Client: "then write a README"
    Client->>FollowUp Queue: POST /sessions/:id/followup
    Note over FollowUp Queue: Message enqueued

    User->>Client: "and add a changelog entry"
    Client->>FollowUp Queue: POST /sessions/:id/followup
    Note over FollowUp Queue: Message enqueued

    Note over Agent Loop: Model stops (bug fixed)

    FollowUp Queue-->>Agent Loop: Pop one ("write a README")
    Note over Agent Loop: New turn starts<br/>(plain user message)
    Note over Agent Loop: Model stops (README written)

    FollowUp Queue-->>Agent Loop: Pop one ("add a changelog entry")
    Note over Agent Loop: New turn starts
    Note over Agent Loop: Model stops (changelog added)

    Agent Loop-->>Client: StreamStopped
    deactivate Agent Loop
```

Each follow-up gets a **full undivided turn** — the agent finishes one task before starting the next.

## Steer vs Follow-up — two distinct scenarios

| | Steer (`POST /steer`) | Follow-up (`POST /followup`) |
|---|---|---|
| **When** | Injected at the next safe point mid-turn (after tool execution, before the next LLM call) | Processed after the current turn finishes completely |
| **Drain semantics** | Drain **all** pending messages every check | Pop exactly **one** message per turn boundary |
| **Injection format** | Wrapped in `<system-reminder>` tags (mid-stream context) | Plain user message (fresh turn) |
| **Effect on loop** | Forces `res.Stopped = false` — loop continues | `continue` — starts a new iteration of the loop |
| **Use case** | "The agent needs to see this NOW" — course corrections, additional context | "Run this next" — follow-up questions that can wait |

## MessageQueue interface with Lock + Confirm/Cancel

The queue storage is behind a `MessageQueue` interface so callers can provide their own implementation (e.g. persistent/distributed store). The interface uses a **Lock + Confirm/Cancel** pattern for safe dequeue: `Dequeue` locks a message, `Confirm` permanently removes it after `sess.AddMessage` succeeds, and `Cancel` releases it back to the queue on failure. This prevents message loss in persistent implementations when the process crashes between dequeue and session persistence.

```go
type MessageQueue interface {
    Enqueue(ctx context.Context, msg QueuedMessage) bool
    Dequeue(ctx context.Context) (QueuedMessage, bool)  // locks the message
    Confirm(ctx context.Context) error                   // removes after successful processing
    Cancel(ctx context.Context) error                    // releases back to queue on failure
    Drain(ctx context.Context) []QueuedMessage           // lock + auto-confirm all
    Len(ctx context.Context) int
}
```

The default in-memory implementation (backed by a buffered channel) treats `Confirm`/`Cancel` as no-ops. Custom implementations can be injected via `WithSteerQueue` and `WithFollowUpQueue` options:

```go
rt, err := runtime.New(team,
    runtime.WithSteerQueue(myPersistentQueue),
    runtime.WithFollowUpQueue(myPersistentQueue),
)
```


pkg/runtime/persistent_runtime.go
```diff
@@ -25,6 +25,21 @@ type streamingState struct {
 	messageID        int64 // ID of the current streaming message (0 if none)
 }
 
+// GetLocalRuntime extracts the underlying *LocalRuntime from a Runtime
+// implementation. It handles both *LocalRuntime and *PersistentRuntime
+// (which embeds *LocalRuntime). Returns nil if the runtime type is not
+// supported (e.g. RemoteRuntime).
```
rumpl: Why would the remote runtime not be supported?

pkg/runtime/persistent_runtime.go
```diff
@@ -25,6 +25,21 @@ type streamingState struct {
...
+func GetLocalRuntime(rt Runtime) *LocalRuntime {
```
rumpl: Why is this still needed?

pkg/runtime/runtime.go
```diff
@@ -80,6 +80,66 @@ func ResumeReject(reason string) ResumeRequest {
...
+
+// SteerQueue is the interface for storing steered messages that are injected
+// into a running agent loop mid-turn. Implementations must be safe for
+// concurrent use: Enqueue is called from API handlers while Drain is called
+// from the agent loop goroutine.
+//
+// The default implementation is InMemorySteerQueue. Callers that need
+// durable or distributed storage can provide their own implementation
+// via the WithSteerQueue option.
+type SteerQueue interface {
```
rumpl: Add a context.Context on each method of this interface

pkg/runtime/runtime.go
```diff
@@ -80,6 +80,115 @@ func ResumeReject(reason string) ResumeRequest {
 	return ResumeRequest{Type: ResumeTypeReject, Reason: reason}
 }
 
+// QueuedMessage is a user message waiting to be injected into the agent loop,
+// either mid-turn (via the steer queue) or at end-of-turn (via the follow-up
+// queue).
+type QueuedMessage struct {
```
rumpl: Let's move these queues outside of this file and into its own

pkg/runtime/loop.go
```diff
@@ -404,13 +386,62 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
...
+				// mid-stream interruption. Each follow-up gets a full
+				// undivided agent turn.
+				//
+				// Dequeue locks the message; Confirm is called after
+				// AddMessage succeeds so persistent queue implementations
+				// can safely re-queue on failure.
+				if followUp, ok := r.DequeueFollowUp(ctx); ok {
+					userMsg := session.UserMessage(followUp.Content, followUp.MultiContent...)
+					sess.AddMessage(userMsg)
+					_ = r.followUpQueue.Confirm(ctx)
```
rumpl: Looking at these lines I fail to see why we would need the confirm here, the only way for the thing to stop between dequeue and confirm is for the whole thing to die. 

Maybe because this only adds a message to the session and doesn't call the session store.

pkg/runtime/persistent_runtime.go
```diff
@@ -25,6 +25,14 @@ type streamingState struct {
 	messageID        int64 // ID of the current streaming message (0 if none)
 }
 
+// MessageInjector is implemented by runtimes that support mid-turn steering
+// and end-of-turn follow-up message injection. Both LocalRuntime (and
+// PersistentRuntime via embedding) and RemoteRuntime satisfy this interface.
+type MessageInjector interface {
```
rumpl: If all of the runtimes implement maybe we want to change the Runtime interface instead of creating a new one?

pkg/runtime/persistent_runtime.go
```diff
@@ -25,6 +25,14 @@ type streamingState struct {
...
+	Steer(msg QueuedMessage) bool
```
rumpl: Question: should we return the error rather than a boolean?

pkg/runtime/loop.go
```diff
@@ -404,13 +386,57 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
...
+						sm.Content,
+					)
+					userMsg := session.UserMessage(wrapped, sm.MultiContent...)
+					sess.AddMessage(userMsg)
+					events <- UserMessage(sm.Content, sess.ID, sm.MultiContent, len(sess.Messages)-1)
+				}
+
+				// Force the loop to continue — the model must respond to
+				// the injected messages even if it was about to stop.
+				res.Stopped = false
```
rumpl: Shouldn't this be a `continue` instead like in the followup case?

pkg/runtime/message_queue.go
```diff
@@ -0,0 +1,121 @@
...
+	Enqueue(ctx context.Context, msg QueuedMessage) bool
+	// Dequeue locks and returns the next message from the queue. The
+	// message is invisible to subsequent Dequeue calls until Confirm or
+	// Cancel is called. Returns the message and true, or a zero value
+	// and false if the queue is empty. Must not block.
+	Dequeue(ctx context.Context) (QueuedMessage, bool)
+	// Confirm permanently removes the most recently dequeued message.
+	// Must be called after the message has been successfully persisted
+	// to the session. For in-memory queues this is a no-op.
+	Confirm(ctx context.Context) error
```
rumpl: Confirm and Cancel are never called

pkg/runtime/message_queue.go
```diff
@@ -0,0 +1,121 @@
...
+	// Cancel releases the most recently dequeued message back to the
+	// queue. For in-memory queues this is a no-op (the message was
+	// already consumed from the channel).
+	Cancel(ctx context.Context) error
+	// Drain locks, returns, and auto-confirms all pending messages.
+	// Must not block — if the queue is empty it returns nil.
+	Drain(ctx context.Context) []QueuedMessage
+	// Len returns the current number of messages in the queue.
+	Len(ctx context.Context) int
```
rumpl: This one is never called either

pkg/runtime/runtime.go
```diff
@@ -1015,6 +1049,39 @@ func (r *LocalRuntime) ResumeElicitation(ctx context.Context, action tools.Elici
...
+	if !r.steerQueue.Enqueue(context.Background(), msg) {
+		return errors.New("steer queue full")
+	}
+	return nil
+}
+
+// DrainSteeredMessages returns all pending steered messages without blocking.
+// It is called inside the agent loop to batch-inject any messages that arrived
+// while the current iteration was in progress.
+func (r *LocalRuntime) DrainSteeredMessages(ctx context.Context) []QueuedMessage {
```
rumpl: This is useless, we can call `r.steerQueue.Drain(ctx)` directly in loop.go

rumpl (dismissed):
Let's get this one in

pkg/runtime/loop.go
```diff
@@ -404,13 +404,55 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
...
 			toolModelOverride = resolveToolCallModelOverride(res.Calls, agentTools)
 
+			// Only compact proactively when the model will continue (has
+			// tool calls to process on the next turn). If the model stopped
+			// and no steered messages override that, compaction is wasteful
+			// because no further LLM call follows.
+			if !res.Stopped {
+				r.compactIfNeeded(ctx, sess, a, m, contextLimit, messageCountBeforeTools, events)
+			}
+
```
rumpl: I'm not sure why we have this, here, let's just compact no matter what? And remove the compactIfneeded below?

rumpl (approved):

PR #2378, opened 2026-04-10
Faster PR CI

rumpl:
Use a different image build job:

- doesn't push
- doesn't need sbom
- doesn't need provenance
- doesn't need the metadata step


derekmisler (approved):
How much faster is it?

rumpl:
`build-image` still takes 8 minutes, but kicks in 4 minutes earlier so we are shaving off 4 minutes of human-wait time

rumpl:
I was hoping that the build would be faster without sbom and provenance but guess it's not the case. Still, 4 minutes is ok I think :)

dgageot (approved):

PR #2379, opened 2026-04-11
fix: compaction fallback to lowest model context limit

shaj13:
## Problem

When the provider is not found in the models store, compaction fails to determine the correct context limit for the model.

```yaml
providers:
  xxx:
    api_type: openai_chatcompletions
    base_url: xxxxx

models:
  claude-opus-4-5:
    provider: xxx
    model: claude-opus-4-5
..... 
```

image

## Fix
Added a fallback that searches all providers for the model and uses the **lowest context limit** found. 
This ensures compaction triggers conservatively regardless of which provider is actually being used at runtime.

image

### DIFF
--- a/pkg/modelsdev/store.go
+++ b/pkg/modelsdev/store.go
@@ -3,6 +3,7 @@ package modelsdev
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -22,6 +23,9 @@ const (
 	CacheFileName   = "models_dev.json"
 	refreshInterval = 24 * time.Hour
 )
+
+// ErrProviderNotFound is returned when a requested provider is not found in the database.
+var ErrProviderNotFound = errors.New("provider not found")
 
 // Store manages access to the models.dev data.
 // All methods are safe for concurrent use.
@@ -91,7 +95,7 @@ func (s *Store) getProvider(ctx context.Context, providerID string) (*Provider, 
 
 	provider, exists := db.Providers[providerID]
 	if !exists {
-		return nil, fmt.Errorf("provider %q not found", providerID)
+		return nil, fmt.Errorf("%w: %q", ErrProviderNotFound, providerID)
 	}
 
 	return &provider, nil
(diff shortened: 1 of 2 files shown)


aheritier (changes requested):
Thanks for identifying this bug — compaction silently failing for custom providers is a real problem worth fixing.

The approach (sentinel error + fallback to lowest context limit across providers) is sound, but unfortunately `main` has diverged significantly since this PR was opened. The `session_compaction.go` file was refactored to extract a `compactionContextLimit()` method, add hook support, and delegate heavy lifting to a new `compactor` package. This PR has merge conflicts and cannot be merged as-is.

**What needs to happen:**

1. **Rebase onto `main`** — the `store.go` change should apply cleanly; the `session_compaction.go` changes need to be rewritten against the current code.

2. **Target `compactionContextLimit()`** — this is the new single place where model context limits are resolved. The fallback logic belongs there:

```go
func (r *LocalRuntime) compactionContextLimit(ctx context.Context, a *agent.Agent) int64 {
    ...
    m, err := r.modelsStore.GetModel(ctx, summaryModel.ID())
    if err != nil && errors.Is(err, modelsdev.ErrProviderNotFound) {
        // fallback: find lowest context limit across all providers
        db, dberr := r.modelsStore.GetDatabase(ctx)
        if dberr == nil {
            for _, p := range db.Providers {
                if v, ok := p.Models[summaryModel.BaseConfig().ModelConfig.Model]; ok {
                    if m == nil || v.Limit.Context < m.Limit.Context {
                        m = &v
                        err = nil
                    }
                }
            }
        }
    }
    if err != nil || m == nil {
        return 0
    }
    return int64(m.Limit.Context)
}
```

3. **Add tests** — the fallback is non-trivial logic; unit tests for `compactionContextLimit` with a mock store that triggers `ErrProviderNotFound` would give confidence.

**What's good here:**
- The `ErrProviderNotFound` sentinel + `%w` wrapping is the correct Go pattern.
- "Use the lowest context limit" is a sensible conservative default.
- The structured logging improvements are a nice touch (though they'll need adapting to the current code which uses `slog.DebugContext`).

aheritier:

dgageot:

PR #2380, opened 2026-04-11
Vendor dependencies for faster build

dgageot:


rumpl:
Doesn't seem that faster? build-and-test still takes 4 minutes. huh

dgageot:
@rumpl yep. not faster

PR #2273, opened 2026-03-28
feat: Explicit OAuth Configuration for Remote MCP Servers (#2248)

yunus25jmi1:
## Summary

This PR implements **explicit OAuth configuration** for remote MCP servers, allowing for pre-registered OAuth clients with servers that do not support Dynamic Client Registration (RFC 7591), such as the Slack MCP server.

## Problem Solved

The previous OAuth implementation always attempted Dynamic Client Registration first. When servers like Slack MCP don’t support this, the OAuth flow would fail completely, causing the MCP toolset to be skipped.

## Solution Implemented

Added explicit OAuth configuration fields to remote MCP server definitions, allowing users to provide pre-registered client credentials directly in the agent YAML configuration.

## What it does

| Feature | Description |
|---------|-------------|
| 🔑 **Explicit Credentials** | Use pre-registered Client ID and Secret |
| 🚪 **Fixed Callback Port** | Configure a fixed port for the OAuth callback server |
| 🔭 **Custom Scopes** | Specify OAuth scopes to request from the server |
| 🛡️ **Priority Flow** | Explicit credentials take priority over dynamic registration |

## Configuration Example

```yaml
toolsets:
  - type: mcp
    remote:
      url: https://mcp.slack.com/mcp
      oauth:
        clientId: "your-client-id"
        clientSecret: "your-client-secret"  # optional
        callbackPort: 3118                   # optional
        scopes:                               # optional
          - search:read.public
          - chat:write
```

## Testing

- ✅ All 6 new tests in `pkg/tools/mcp/oauth_config_test.go` pass
- ✅ Verified with Slack and GitHub MCP servers
- ✅ Successful build with new configurations

## Files Modified

- `pkg/config/latest/types.go`: Added OAuth configuration structs (using camelCase as requested)
- `pkg/tools/mcp/*`: Updated MCP client, transport, and server to handle explicit OAuth
- `pkg/teamloader/registry.go`: Wired configuration to toolset creation
- `agent-schema.json`: Updated schema for validation (camelCase)
- `examples/slack-oauth-example.yaml`: New example configuration (camelCase)

---

**Related Issues:** Fixes #416, Addresses #2248

PR #2394, opened 2026-04-12
Support explicit OAuth credentials for remote MCP servers

### LINKED ISSUE #2248: Support explicit OAuth configuration for remote MCP servers

dgageot:
This PR is an attempt at continuing the work started in #2273 and #2274 to fix #2248.

## Problem

Remote MCP servers that don't support Dynamic Client Registration (RFC 7591) — such as Slack and GitHub — fail because the code fabricates a `registration_endpoint` when none is advertised, then tries (and fails) to dynamically register.

## Solution

Allow users to configure explicit OAuth credentials (`clientId`, `clientSecret`, `callbackPort`, `scopes`) directly in the remote MCP toolset config. The managed OAuth flow now resolves credentials in priority order:

1. **Explicit config** — use `clientId`/`clientSecret` from the YAML
2. **Dynamic registration** — if the server advertises a `registration_endpoint`
3. **Error** — clear message explaining neither option is available

## Key changes

- Add `RemoteOAuthConfig` to config types with `clientId`, `clientSecret`, `callbackPort`, and `scopes` fields
- Stop fabricating `registration_endpoint` in `validateAndFillDefaults()` and `createDefaultMetadata()`
- Use explicit credentials in the managed OAuth flow when configured
- Support fixed callback port for OAuth redirect URI
- Add `scopes` parameter to `BuildAuthorizationURL`
- Validate `callbackPort` range (1–65535) and `oauth` on non-mcp types
- Update `agent-schema.json` with `RemoteOAuthConfig` definition
- Add example config (`examples/remote_mcp_oauth.yaml`)

## Example

```yaml
toolsets:
  - type: mcp
    remote:
      url: "https://mcp.example.com/sse"
      transport_type: sse
      oauth:
        clientId: "your-client-id"
        clientSecret: "your-client-secret"
        callbackPort: 8080
        scopes:
          - "read"
          - "write"
```

Co-authored-by: nicholasgasior <nicholasgasior@users.noreply.github.com>
Co-authored-by: rumpl <rumpl@users.noreply.github.com>

### DIFF
--- a/pkg/config/latest/validate.go
+++ b/pkg/config/latest/validate.go
@@ -93,7 +93,7 @@ func (t *Toolset) validate() error {
 	if t.Ref != "" && t.Type != "mcp" && t.Type != "rag" {
 		return errors.New("ref can only be used with type 'mcp' or 'rag'")
 	}
-	if (t.Remote.URL != "" || t.Remote.TransportType != "") && t.Type != "mcp" {
+	if (t.Remote.URL != "" || t.Remote.TransportType != "" || t.Remote.OAuth != nil) && t.Type != "mcp" {
 		return errors.New("remote can only be used with type 'mcp'")
 	}
 	if (len(t.Remote.Headers) > 0) && (t.Type != "mcp" && t.Type != "a2a") {
@@ -138,6 +138,17 @@ func (t *Toolset) validate() error {
 		}
 		if count > 1 {
 			return errors.New("either command, remote or ref must be set, but only one of those")
+		}
+		if t.Remote.OAuth != nil {
+			if t.Remote.URL == "" {
+				return errors.New("oauth requires remote url to be set")
+			}
+			if t.Remote.OAuth.ClientID == "" {
+				return errors.New("oauth requires clientId to be set")
+			}
+			if t.Remote.OAuth.CallbackPort != 0 && (t.Remote.OAuth.CallbackPort < 1 || t.Remote.OAuth.CallbackPort > 65535) {
+				return errors.New("oauth callbackPort must be between 1 and 65535")
+			}
 		}
 	case "a2a":
 		if t.URL == "" {
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1253,6 +1253,10 @@
           "additionalProperties": {
             "type": "string"
           }
+        },
+        "oauth": {
+          "$ref": "#/definitions/RemoteOAuthConfig",
+          "description": "Explicit OAuth credentials for servers that do not support Dynamic Client Registration"
         }
       },
       "required": [
@@ -1686,6 +1690,36 @@
         "strategies"
       ],
       "additionalProperties": false
+    },
+    "RemoteOAuthConfig": {
+      "type": "object",
+      "description": "OAuth configuration for remote MCP servers that do not support Dynamic Client Registration (RFC 7591)",
+      "properties": {
+        "clientId": {
+          "type": "string",
+          "description": "OAuth client ID"
+        },
+        "clientSecret": {
+          "type": "string",
+          "description": "OAuth client secret"
+        },
+        "callbackPort": {
+          "type": "integer",
+          "description": "Fixed port for the OAuth callback server (default: random available port)",
+          "minimum": 1,
+          "maximum": 65535
+        },
+        "scopes": {
+          "type": "array",
+          "description": "OAuth scopes to request",
+          "items": {
+            "type": "string"
+          }
+        }
+      },
+      "required": [
+        "clientId"
+      ]
     }
   }
 }
--- a/examples/remote_mcp_oauth.yaml
+++ b/examples/remote_mcp_oauth.yaml
@@ -1,0 +1,23 @@
+#!/usr/bin/env docker agent run
+
+# Example: Remote MCP server with explicit OAuth credentials.
+# Use this when connecting to MCP servers that do NOT support
+# Dynamic Client Registration (RFC 7591), such as Slack or GitHub.
+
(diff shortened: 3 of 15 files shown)


pkg/config/latest/validate.go
```diff
@@ -139,6 +139,17 @@ func (t *Toolset) validate() error {
...
 		}
+		if t.Remote.OAuth != nil {
+			if t.Remote.URL == "" {
+				return errors.New("oauth requires remote url to be set")
+			}
+			if t.Remote.OAuth.ClientID == "" {
+				return errors.New("oauth requires clientId to be set")
+			}
+			if t.Remote.OAuth.CallbackPort != 0 && (t.Remote.OAuth.CallbackPort < 1 || t.Remote.OAuth.CallbackPort > 65535) {
+				return errors.New("oauth callbackPort must be between 1 and 65535")
```
krissetto, line 150: i don't think we should go all the way to port 1, that'd required root permissions

krissetto (approved):

PR #2411, opened 2026-04-14
fix: retry MCP stdio toolset when binary is unavailable at startup

### LINKED ISSUE #2410: MCP CLI tool installed after agent start is not picked up (requires restart)
### Summary

When an MCP CLI tool (stdio transport) is configured in `agent.yaml` but its binary is **not yet present** at agent startup, the agent correctly fails to use it. However, once the binary is installed/built and becomes available on PATH, the running agent **still cannot discover or use the tool** — a full restart is required.

### Reproduction

A minimal reproduction is available at: https://github.com/simonferquel/repro-docker-agent-install-mcp-after-start

The repo contains:
- `mcp-hello/` — a tiny MCP server (Go, stdio transport) with a single `hello` tool
- `build.sh` — builds the `mcp-hello` binary into `./bin/`
- `agent.yaml` — docker-agent config referencing `mcp-hello` as an MCP toolset

### Steps to Reproduce

1. Ensure the binary is **not** on PATH: `rm -rf bin/ && which mcp-hello` (should fail)
2. Start the agent: `docker agent run agent.yaml`
3. Ask the agent to use the hello tool → **fails** (expected — binary missing)
4. Build the binary in another terminal: `./build.sh`
5. Ask the agent to use the hello tool again in the still-running agent → **still fails** 🐛

### Expected Behavior

After the MCP binary becomes available on PATH, the running agent should be able to discover and use it without requiring a restart (e.g., by retrying to start the MCP server on next tool use, or by periodically checking for newly available toolsets).

### Actual Behavior

The agent continues to report the tool as unavailable even after the binary is installed. Only restarting the agent (`docker agent run agent.yaml`) resolves the issue.

### Workaround

Restart the agent after installing the MCP binary.

### Impact

This is particularly painful in development workflows where MCP tools are built and iterated on locally — every rebuild requires a full agent restart to pick up the new binary.

simonferquel-clanker:
When a stdio MCP toolset binary is not found at startup (EOF from Initialize), `Start()` now propagates `errServerUnavailable` instead of silently marking the toolset as started with empty tools.

- The agent runtime already calls `ensureToolSetsAreStarted()` at the beginning of every conversation turn, which calls `StartableToolSet.Start()`
- `StartableToolSet` only sets `started=true` when the inner `Start()` returns nil, so a failed start is automatically retried next turn
- Once the binary appears on PATH, `Start()` succeeds and tools become available — no background goroutine needed

Tests added:
- `Start()` returns error and does not mark started when binary is unavailable
- Lazy retry across turns: fail on turn 1, succeed on turn 2
- Full `StartableToolSet` integration across multiple turns

Fixes #2410

### DIFF
--- a/pkg/tools/mcp/mcp.go
+++ b/pkg/tools/mcp/mcp.go
@@ -118,9 +118,11 @@ func NewRemoteToolset(name, urlString, transport string, headers map[string]stri
 }
 
 // errServerUnavailable is returned by doStart when the MCP server could not be
-// reached but the error is non-fatal (e.g. EOF). The toolset is considered
-// "started" so the agent can proceed, but watchConnection must not be spawned
-// because there is no live connection to monitor.
+// reached but the error is non-fatal (e.g. EOF during stdio exec). Start()
+// propagates this error so the caller (StartableToolSet) does not mark the
+// toolset as started. The agent runtime calls ensureToolSetsAreStarted() at
+// the beginning of every conversation turn, which will retry Start() and
+// pick up the binary once it becomes available—no background goroutine needed.
 var errServerUnavailable = errors.New("MCP server unavailable")
 
 // Describe returns a short, user-visible description of this toolset instance.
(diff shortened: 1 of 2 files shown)


simonferquel (changes requested):
Instead of a goroutine looping, could we just retry failed MCPs on each new conversation turn ?

simonferquel (changes requested):
Repro project still fails. Your approach might be wrong

simonferquel-clanker:
Closing in favour of a redesign — the approach needs rethinking before re-implementing. See #2410 for context.

PR #2415, opened 2026-04-14
feat: user configurable tui keybindings

joshbarrington:
As referenced here: https://github.com/docker/docker-agent/issues/1626 and the subsequent closed MR: https://github.com/docker/docker-agent/pull/1629

First stage implementation for allowing user configurable keybindings.

@krissetto would like to get your thoughts and then we can discuss potential extensions/alternative implementations.

### DIFF
--- a/pkg/tui/core/keys.go
+++ b/pkg/tui/core/keys.go
@@ -1,0 +1,108 @@
+package core
+
+import (
+	"sync"
+
+	"charm.land/bubbles/v2/key"
+
+	"github.com/docker/docker-agent/pkg/userconfig"
+)
+
+// KeyMap contains global keybindings used across the TUI
+type KeyMap struct {
+	Quit                  key.Binding
+	SwitchFocus           key.Binding
+	Commands              key.Binding
+	Help                  key.Binding
+	ToggleYolo            key.Binding
+	ToggleHideToolResults key.Binding
+	CycleAgent            key.Binding
+	ModelPicker           key.Binding
+	ClearQueue            key.Binding
+	Suspend               key.Binding
+	ToggleSidebar         key.Binding
+	EditExternal          key.Binding
+	HistorySearch         key.Binding
+}
+
+var (
+	cachedKeys KeyMap
+	keysOnce   sync.Once
+)
+
+// DefaultKeyMap returns the default keybindings
+func DefaultKeyMap() KeyMap {
+	return KeyMap{
+		Quit:                  key.NewBinding(key.WithKeys("ctrl+c"), key.WithHelp("ctrl+c", "quit")),
+		SwitchFocus:           key.NewBinding(key.WithKeys("tab"), key.WithHelp("tab", "switch focus")),
+		Commands:              key.NewBinding(key.WithKeys("ctrl+k"), key.WithHelp("ctrl+k", "commands")),
+		Help:                  key.NewBinding(key.WithKeys("ctrl+h", "f1", "ctrl+?"), key.WithHelp("ctrl+h", "help")),
+		ToggleYolo:            key.NewBinding(key.WithKeys("ctrl+y"), key.WithHelp("ctrl+y", "toggle yolo mode")),
+		ToggleHideToolResults: key.NewBinding(key.WithKeys("ctrl+o"), key.WithHelp("ctrl+o", "toggle hide tool results")),
+		CycleAgent:            key.NewBinding(key.WithKeys("ctrl+s"), key.WithHelp("ctrl+s", "cycle agent")),
+		ModelPicker:           key.NewBinding(key.WithKeys("ctrl+m"), key.WithHelp("ctrl+m", "model picker")),
+		ClearQueue:            key.NewBinding(key.WithKeys("ctrl+x"), key.WithHelp("ctrl+x", "clear queue")),
+		Suspend:               key.NewBinding(key.WithKeys("ctrl+z"), key.WithHelp("ctrl+z", "suspend")),
+		ToggleSidebar:         key.NewBinding(key.WithKeys("ctrl+b"), key.WithHelp("ctrl+b", "toggle sidebar")),
+		EditExternal:          key.NewBinding(key.WithKeys("ctrl+g"), key.WithHelp("ctrl+g", "edit in external editor")),
+		HistorySearch:         key.NewBinding(key.WithKeys("ctrl+r"), key.WithHelp("ctrl+r", "history search")),
+	}
+}
+
+// buildKeys merges user config overrides with the defaults to produce a KeyMap.
+// This is separated from GetKeys() to allow testing with mock settings.
+func buildKeys(settings *userconfig.Settings) KeyMap {
+	keys := DefaultKeyMap()
+
+	if settings != nil && settings.Keybindings != nil {
+		for _, b := range *settings.Keybindings {
+			if len(b.Keys) == 0 {
+				continue
+			}
+
+			usrKeys := b.Keys
+			keyName := usrKeys[0]
+
+			switch b.Action {
+			case "quit":
+				keys.Quit = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "quit"))
+			case "switch_focus":
+				keys.SwitchFocus = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "switch focus"))
+			case "commands":
+				keys.Commands = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "commands"))
+			case "help":
+				keys.Help = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "help"))
+			case "toggle_yolo":
+				keys.ToggleYolo = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "toggle yolo mode"))
+			case "toggle_hide_tool_results":
+				keys.ToggleHideToolResults = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "toggle hide tool results"))
+			case "cycle_agent":
+				keys.CycleAgent = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "cycle agent"))
+			case "model_picker":
+				keys.ModelPicker = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "model picker"))
+			case "clear_queue":
+				keys.ClearQueue = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "clear queue"))
+			case "suspend":
+				keys.Suspend = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "suspend"))
+			case "toggle_sidebar":
+				keys.ToggleSidebar = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "toggle sidebar"))
+			case "edit_external":
+				keys.EditExternal = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "edit in external editor"))
+			case "history_search":
+				keys.HistorySearch = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "history search"))
+			}
+		}
+	}
+
+	return keys
+}
+
+// GetKeys returns the current keybindings, merging user config overrides with defaults.
+// The result is cached after the first call.
+func GetKeys() KeyMap {
+	keysOnce.Do(func() {
+		cachedKeys = buildKeys(userconfig.Get())
+	})
+
+	return cachedKeys
+}
--- a/pkg/tui/core/keys.go
+++ b/pkg/tui/core/keys.go
@@ -28,8 +28,8 @@ type KeyMap struct {
 }
 
 var (
-	cachedKeys KeyMap
-	keysOnce   sync.Once
+	ckMutex    sync.RWMutex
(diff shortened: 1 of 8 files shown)


aheritier:
/review

Any thoughts @krissetto ?

pkg/userconfig/userconfig.go
```diff
@@ -65,6 +65,13 @@ type Settings struct {
 	// and agents. These act as user-wide defaults; session-level and agent-level
 	// permissions override them.
 	Permissions *latest.PermissionsConfig `yaml:"permissions,omitempty"`
+	// Keybindings allows users to configure custom keybindings for TUI actions.
+	Keybindings *[]Keybindings `yaml:"keybindings,omitempty"`
+}
+
+type Keybindings struct {
+	Action string   `yaml:"action"`
```
aheritier, line 81: **Blocking**: The type name `Keybindings` (plural) represents a single keybinding entry. Should be `Keybinding` (singular) to match Go naming conventions. The slice usage (`*[]Keybinding`) already communicates the plural nature.

```go
type Keybinding struct {
    Action string   `yaml:"action"`
    Keys   []string `yaml:"keys"`
}
```

And in `Settings`:
```go
Keybindings *[]Keybinding `yaml:"keybindings,omitempty"`
```

pkg/tui/tui.go
```diff
@@ -1705,32 +1671,38 @@ func (m *appModel) handleKeyPress(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
 	}
 
 	// Global keyboard shortcuts (active even during history search)
+	keys := core.GetKeys()
 	switch {
-	case key.Matches(msg, key.NewBinding(key.WithKeys("ctrl+z"))):
+	case key.Matches(msg, keys.Quit):
```
aheritier: **Blocking**: This adds new behavior beyond configurable keybindings — pressing the quit key now opens an exit confirmation dialog instead of quitting directly (which was the previous ctrl+c behavior via bubbletea's default handling).

This is a meaningful UX change. Before: ctrl+c → immediate quit. After: ctrl+c → confirmation dialog → confirm → quit.

Should this behavioral change be in a separate commit or at minimum called out in the PR description? It's easy to miss during review otherwise.

pkg/tui/core/keys.go
```diff
@@ -0,0 +1,108 @@
...
+	if settings != nil && settings.Keybindings != nil {
+		for _, b := range *settings.Keybindings {
+			if len(b.Keys) == 0 {
+				continue
+			}
+
+			usrKeys := b.Keys
+			keyName := usrKeys[0]
+
+			switch b.Action {
```
aheritier: **Non-blocking**: Consider adding basic validation here — at minimum, log a warning for unrecognized `b.Action` values so users know they have a typo in their config. Currently, unknown actions are silently ignored.

Also worth considering for a follow-up: conflict detection (two actions bound to the same key) and invalid key string validation.

pkg/tui/core/keys.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+		}
+	}
+
+	return keys
+}
+
+// GetKeys returns the current keybindings, merging user config overrides with defaults.
+// The result is cached after the first call.
+func GetKeys() KeyMap {
```
aheritier, line 133: **Non-blocking**: The `sync.Once` caching means keybindings can never be reloaded without restarting the application. This is fine for a first implementation, but consider adding a `ResetKeys()` function (even if unexported, useful for tests) or a comment noting this limitation for future hot-reload support.

pkg/tui/core/keys.go
```diff
@@ -0,0 +1,108 @@
...
+				keys.ToggleSidebar = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "toggle sidebar"))
+			case "edit_external":
+				keys.EditExternal = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "edit in external editor"))
+			case "history_search":
+				keys.HistorySearch = key.NewBinding(key.WithKeys(usrKeys...), key.WithHelp(keyName, "history search"))
+			}
+		}
+	}
+
+	return keys
```
aheritier: **Non-blocking**: The repetitive switch/case could be reduced with a map-based lookup. Something like:

```go
type keyField struct {
    binding *key.Binding
    help    string
}

actionMap := map[string]keyField{
    "quit":        {&keys.Quit, "quit"},
    "switch_focus": {&keys.SwitchFocus, "switch focus"},
    // ...
}

for _, b := range *settings.Keybindings {
    if len(b.Keys) == 0 { continue }
    if f, ok := actionMap[b.Action]; ok {
        *f.binding = key.NewBinding(key.WithKeys(b.Keys...), key.WithHelp(b.Keys[0], f.help))
    }
}
```

This would also make it trivial to add validation/logging for unknown actions.

pkg/tui/dialog/exit_confirmation.go
```diff
@@ -21,7 +21,7 @@ type exitConfirmationKeyMap struct {
 func defaultExitConfirmationKeyMap() exitConfirmationKeyMap {
 	return exitConfirmationKeyMap{
 		Yes: key.NewBinding(
-			key.WithKeys("y", "Y", "ctrl+c"),
+			key.WithKeys("y", "Y", core.GetKeys().Quit.Keys()[0]),
```
aheritier: **Question**: What happens if the user configures quit with multiple keys (e.g., `keys: ["ctrl+q", "ctrl+c"]`)? Only `Keys()[0]` is used here, so pressing the second configured key won't trigger "yes" in the confirmation dialog. Is that intentional?

Might be better to append all quit keys: `key.WithKeys(append([]string{"y", "Y"}, core.GetKeys().Quit.Keys()...)...)`

pkg/tui/core/keys.go
aheritier, line 1: **Blocking**: The TUI docs (`docs/features/tui/index.md`) have a "Keyboard Shortcuts" table with hardcoded key names. This needs to either:
1. Mention that these are the defaults and can be overridden via `~/.config/cagent/config.yaml`
2. Add a section documenting the keybinding configuration format

Example config section to document:
```yaml
settings:
  keybindings:
    - action: "quit"
      keys: ["ctrl+q"]
    - action: "commands"
      keys: ["f2", "ctrl+k"]
```

Also: the list of valid action names should be documented somewhere users can find it.

pkg/tui/core/keys_test.go
aheritier, line 1: **Praise**: Good test coverage. Testing `buildKeys` directly (bypassing the `sync.Once` cache) is the right approach, and the YAML round-trip test is a nice touch for ensuring the config format works end-to-end.

aheritier (changes requested):
Good foundation for configurable keybindings. The architecture (centralized KeyMap, defaults + override merge, caching) is sound and the refactoring of `tui.go` to use the shared KeyMap is a clear improvement.

**Blocking items:**
1. **Type naming**: `Keybindings` → `Keybinding` (singular for a single entry)
2. **Undocumented behavioral change**: The quit key now opens an exit confirmation dialog — this changes existing UX and should be explicitly noted/discussed
3. **Missing docs**: No user-facing documentation for how to configure keybindings

**CI**: Still running at review time — please confirm tests pass.

**Also noting** (non-blocking): PR title has a typo — "keybings" → "keybindings".

pkg/tui/tui.go
joshbarrington: @aheritier can you confirm that this indeed does change the current behaviour? I also seem to get the confirmation dialog when building from `main` branch and pressing `ctrl+c`.

joshbarrington:
@aheritier this can be re-reviewed 👍

pkg/tui/core/keys.go
```diff
@@ -0,0 +1,144 @@
...
+		cachedKeys = buildKeys(userconfig.Get())
+	})
+
+	return cachedKeys
+}
+
+// ResetKeys clears the cached keybindings, allowing them to be reloaded.
+// This is primarily useful for testing or future hot-reload support.
+func ResetKeys() {
+	keysOnce = sync.Once{}
```
docker-agent: **[MEDIUM] Data race: `ResetKeys()` writes `keysOnce` without synchronization**

`ResetKeys()` assigns `keysOnce = sync.Once{}` directly to the package-level variable with no mutex, while `GetKeys()` concurrently reads `keysOnce` via `.Do(...)`. In Go, a concurrent read+write of a struct value without synchronization is a data race — the race detector will flag this.

Even though the comment says this is "primarily useful for testing or future hot-reload support", tests run with `-race` will still trigger the detector, and any future hot-reload use would be a production race.

**Suggested fix**: Protect both `keysOnce` and `cachedKeys` with a `sync.RWMutex`, or replace the `sync.Once` with an `atomic.Pointer[KeyMap]` pattern that allows safe replacement:

```go
var (
    cachedKeysMu sync.RWMutex
    cachedKeys   *KeyMap
)

func GetKeys() KeyMap {
    cachedKeysMu.RLock()
    k := cachedKeys
    cachedKeysMu.RUnlock()
    if k != nil {
        return *k
    }
    cachedKeysMu.Lock()
    defer cachedKeysMu.Unlock()
    if cachedKeys == nil { // double-checked
        built := buildKeys(userconfig.Get())
        cachedKeys = &built
    }
    return *cachedKeys
}

func ResetKeys() {
    cachedKeysMu.Lock()
    cachedKeys = nil
    cachedKeysMu.Unlock()
}
```

pkg/tui/core/keys.go
```diff
@@ -0,0 +1,144 @@
...
+			continue
+		}
+
+		if existingAction, exists := boundKeys[kStr]; exists {
+			slog.Warn("Keybinding conflict detected", "key", kStr, "action", action, "conflicts_with", existingAction)
+		} else {
+			boundKeys[kStr] = action
+		}
+
+		validKeys = append(validKeys, kStr)
```
docker-agent, line 75: **[MEDIUM] Keybinding conflict logged but not enforced — duplicate key bound to multiple actions**

When `validateKeys` detects a duplicate key, it logs a warning but then falls through to `validKeys = append(validKeys, kStr)` unconditionally. The conflicting key is added to both the original action's binding and the new action's binding.

Result: pressing the conflicting key will match **both** `key.Matches` checks in `handleKeyPress` — the `switch` will execute the first matching case only, but the second action silently never fires, making the warning misleading (the user is told of a conflict but gets unpredictable behavior, not an error).

For example, if a user configures:
```yaml
keybindings:
  - action: "quit"
    keys: ["ctrl+q"]
  - action: "suspend"
    keys: ["ctrl+q"]  # conflict with quit
```
Both `keys.Quit` and `keys.Suspend` will contain `"ctrl+q"`. Pressing it will always trigger whichever `case` appears first in the `switch`.

**Suggested fix**: Skip the conflicting key (add `continue` after logging the warning):

```go
if existingAction, exists := boundKeys[kStr]; exists {
    slog.Warn("Keybinding conflict detected", "key", kStr, "action", action, "conflicts_with", existingAction)
    continue  // skip this key — don't bind it to the new action
} else {
    boundKeys[kStr] = action
}

validKeys = append(validKeys, kStr)
```

docker-agent (commented):

This PR introduces user-configurable TUI keybindings — a clean, well-structured implementation with good test coverage. Three issues were found in the new code:

1. **Data race** in `ResetKeys()` — writes `keysOnce` without a mutex while `GetKeys()` reads it concurrently (flagged by Go's race detector, and blocks safe hot-reload use).
2. **Conflict detection not enforced** in `validateKeys` — conflicting keys are logged but still bound to both actions, leading to ambiguous key dispatch.
3. **Elicitation dialog exits without confirmation** — the quit key in `ElicitationDialog.handleKeyPress` calls `tea.Quit` directly rather than stacking the exit confirmation dialog like every other quit path does.

A low-severity edge case was also noted: if a user maps quit to `n`/`N`, the exit confirmation dialog's Yes binding would capture it before the No binding.

pkg/tui/dialog/elicitation.go
```diff
@@ -180,6 +181,9 @@ func (d *ElicitationDialog) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 
 func (d *ElicitationDialog) handleKeyPress(msg tea.KeyPressMsg) (layout.Model, tea.Cmd) {
 	switch {
+	case key.Matches(msg, core.GetKeys().Quit):
+		cmd := d.close(tools.ElicitationActionDecline, nil)
+		return d, tea.Sequence(cmd, tea.Quit)
```
docker-agent: **[MEDIUM] Quit key in elicitation dialog calls `tea.Quit` directly, bypassing exit confirmation**

When the user presses their configured quit key while an elicitation dialog is open, the elicitation dialog captures the key press first (dialogs have priority in `handleKeyPress`) and calls `tea.Sequence(cmd, tea.Quit)` — exiting the program immediately.

This bypasses the exit confirmation pattern used everywhere else in the TUI, where the quit key opens an `ExitConfirmationDialog` first (allowing the user to cancel with N/Esc). The user gets no chance to confirm or back out.

Compare with how pressing Escape is handled:
```go
case key.Matches(msg, d.keyMap.Escape):
    cmd := d.close(tools.ElicitationActionCancel, nil)
    return d, cmd  // just closes the dialog, stays in TUI
```

**Suggested fix**: Instead of calling `tea.Quit` directly, emit an `OpenDialogMsg` to push the exit confirmation on top, consistent with the rest of the codebase:

```go
case key.Matches(msg, core.GetKeys().Quit):
    _ = d.close(tools.ElicitationActionDecline, nil)
    return d, core.CmdHandler(dialog.OpenDialogMsg{
        Model: dialog.NewExitConfirmationDialog(),
    })
```

dgageot:

PR #2419, opened 2026-04-15
fix: add mutex lock to AddMessageUsageRecord to prevent data race

dgageot:
AddMessageUsageRecord appends to MessageUsageHistory without holding sess.mu, while every other Session append method holds the lock. This causes a data race when called concurrently. Adds the missing lock and a race-detector test.


aheritier (approved):
LGTM

PR #2420, opened 2026-04-15
fix: add mutex to protect lastSelectedID in rule-based router

dgageot:
lastSelectedID is written in CreateChatCompletionStream and read in LastSelectedModelID with no synchronization. Adds a sync.RWMutex to prevent the data race.


aheritier (approved):
LGTM

PR #2421, opened 2026-04-15
fix: hold mutex for instructions read in MCP Toolset.Instructions()

dgageot:
Instructions() checked ts.started under the lock, released it, then read ts.instructions unlocked. A concurrent doStart could write ts.instructions in between. Keeps the lock held with defer for the entire method.


aheritier (approved):
LGTM

PR #2422, opened 2026-04-15
fix: prevent panic in extractSystemBlocks on empty system message with CacheControl

dgageot:
If a system message has CacheControl=true but empty content, nothing is appended to systemBlocks, and systemBlocks[len(systemBlocks)-1] panics with index -1. Adds a len() guard.


aheritier (approved):
LGTM

PR #2423, opened 2026-04-15
fix: guard against empty messages slice in SendUserMessage path

dgageot:
SendUserMessage defaults to true on new sessions. If GetMessages returns an empty slice (e.g. agent with no instruction), messages[len(messages)-1] panics. Adds a len() > 0 guard.


aheritier (approved):
LGTM

PR #2424, opened 2026-04-15
fix: replace time.Sleep with context-aware select in MCP tryRestart

dgageot:
tryRestart uses time.Sleep for exponential backoff (up to 16s) which ignores context cancellation and Stop(). Replaces with a select on time.NewTimer and ctx.Done() for prompt shutdown.


aheritier (approved):
LGTM

PR #2425, opened 2026-04-15
fix: prevent symlink-based path traversal in ACP filesystem toolset

dgageot:
resolvePath uses filepath.Abs which doesn't follow symlinks. A symlink inside the working directory pointing outside it passes the prefix check. Uses filepath.EvalSymlinks with a recursive ancestor walk for non-existent paths. Security fix.


aheritier (approved):
LGTM

PR #2426, opened 2026-04-15
fix: reject OAuth callback when expected state has not been set (CSRF)

dgageot:
The CSRF state check was skipped when expectedState was empty (before SetExpectedState was called). A callback arriving between NewCallbackServer() and SetExpectedState() bypassed CSRF protection. Now rejects callbacks when no expected state is set.


aheritier (approved):
LGTM

PR #2431, opened 2026-04-15
fix: prevent panic in extractSystemBlocks on empty system message wit…

dgageot:
…h CacheControl

Assisted-By: docker-agent


aheritier (approved):
LGTM

PR #2436, opened 2026-04-15
fix: replace time.Sleep with context-aware select in MCP tryRestart

dgageot:
Assisted-By: docker-agent


aheritier (approved):
LGTM. This is a focused fix that makes tryRestart responsive to context cancellation instead of blocking in time.Sleep during backoff. The added regression test covers the behavior change and the overall risk looks low.

PR #2437, opened 2026-04-15
fix: cache parsed robots.txt per host instead of boolean result

dgageot:
The robots.txt cache was keyed by host but stored a boolean from the first URL's path check. This caused subsequent URLs on the same host to skip their own per-path robots.txt evaluation.

Cache the parsed *robotstxt.RobotsData per host and call TestAgent for each URL individually.

Assisted-By: docker-agent

### DIFF
--- a/pkg/tools/builtin/fetch.go
+++ b/pkg/tools/builtin/fetch.go
@@ -59,8 +59,8 @@ func (h *fetchHandler) CallTool(ctx context.Context, params FetchToolArgs) (*too
 
 	var results []FetchResult
 
-	// Group URLs by host to fetch robots.txt once per host
-	robotsCache := make(map[string]bool)
+	// Cache parsed robots.txt per host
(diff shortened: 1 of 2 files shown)


aheritier (changes requested):
Good fix overall — caching the parsed robots.txt per host and re-evaluating `TestAgent` per URL path is the right change, and it addresses the bug described in the PR.

I do think this needs regression coverage before merge, though. Right now `fetch_test.go` only covers single-URL robots allow/block cases, but this PR is fixing a multi-URL / same-host caching bug. Please add a test that exercises two URLs on the same host with different path permissions (for example one allowed path and one disallowed path) and verifies we fetch `robots.txt` once but evaluate access per URL.

It would also be good to add coverage for the new error path introduced by `fetchRobots` (for example malformed robots.txt or an unexpected non-200/non-404 response), since this refactor changes the observable error returned to callers.

With that test coverage in place, this looks good.

aheritier (approved):

PR #2438, opened 2026-04-15
fix: preserve assistant text when tool calls present in Responses API conversion

dgageot:
When converting assistant messages with both text and tool calls to the Responses API format, the text content was silently discarded via continue. Emit the text as a separate assistant message before the function calls.

Includes a regression test.

Assisted-By: docker-agent


aheritier (approved):
LGTM. This fixes the Responses API conversion bug cleanly by preserving assistant text when tool calls are present, and the added regression test covers the reported case. I verified the relevant package tests pass locally.

PR #2439, opened 2026-04-15
Npm install

dgageot:
Allow to use npm to install a missing tool

### DIFF
--- a/docs/configuration/tools/index.md
+++ b/docs/configuration/tools/index.md
@@ -120,9 +120,9 @@ When configuring MCP or LSP tools that require a binary command, docker agent ca
 2. If not found, it checks the docker agent tools directory (`~/.cagent/tools/bin/`)
 3. If still not found, it looks up the command in the aqua registry and installs it automatically
 
-### Explicit Package Reference
+### Explicit Package Reference (aqua)
 
-Use the `version` property to specify exactly which package to install:
+Use the `version` property to specify exactly which aqua package to install:
 
 ```yaml
 toolsets:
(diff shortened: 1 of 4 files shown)


derekmisler:
are we planning to _only_ support `npm install`? over on the frontends and pinata repos, for example, they've all switched to `pnpm` because it's more disk-space and security friendly. i think it's worth considering if there should be a config option (`npmCommand: pnpm`, for example), or if it's okay to require `npm` for just this specific feature.

also, i don't see that we're specifying node versions, so whatever `npm` is on a user's `$PATH` is what'll get used. if a tool requires node 20+ and the user's default is node 18, the install could succeed but the tool could fail at runtime with a confusing error. at minimum, the error message when things break should hint at checking the node/npm version. `npm install --global --prefix` only works on `npm 7+`, for example. our web apps all specify a node version for this reason

aheritier (commented):
This looks like a clean, focused addition overall, and the CI is green. I like the explicit `npm:` routing and the use of `--prefix` to keep installs inside docker-agent's tools directory.

One thing I'd still like to call out: the new npm installer path isn't covered by a success-path test yet. The added tests cover parsing and disabled auto-install behavior, but we still don't exercise the actual `npm install --global --prefix ...` flow, the "npm not found" case, or the post-install binary verification logic. Since this PR adds a brand-new installation backend, that leaves the highest-risk path untested.

I don't see a clear blocker in the implementation itself, but I would strongly recommend adding at least one installer-level test around the successful npm path in a follow-up (or here if convenient), especially to lock down the expected binary lookup behavior under `<tools dir>/bin/<command>`.

dgageot:
I'm going to close it for now and think more about what we want to do

Issue #2409, opened 2026-04-14
RFC: Establish pkg/ai as the foundational AI kernel


shaj13:
## Overview

This RFC proposes standardizing the Docker Agent to go beyond a CLI tool for running agentic AI. There are many open-source AI frameworks out there, each with its own limitations and tradeoffs. The Docker Agent already has strong core capabilities — model fallback, streaming, tool execution, structured output, multi-provider support — but they're buried inside runtime internals, tightly coupled and not reusable.

The focus of this proposal is modularization and establishing a strong core. By redesigning around a central `pkg/ai` package with clean, idiomatic Go APIs, we create clear boundaries between layers. With that modularity in place, targeting different use cases becomes straightforward, adding new features becomes incremental instead of invasive, and we avoid the tradeoffs that come from having everything tangled together.

## Pain

The Docker Agent was built around `pkg/runtime` as the primary entry point for model interaction. This made sense early on — the runtime powers the TUI agent loop and it does that job well. But as the project grew, more features needed to talk to models (session titles, compaction, background agents), and each one had to work around the runtime rather than with a shared foundation.

Today, the core LLM interaction logic — streaming, fallback, retry, tool execution — lives inside runtime alongside orchestration concerns like sessions, events, permissions, and hooks. These are tightly coupled, which makes it difficult to use one without pulling in the other.

Specific challenges:

- **Model interaction requires runtime.** There is no lightweight way to call a model. Session compaction creates an entire nested runtime instance for a single summarization call. Session title generation reimplements its own stream drain and fallback loop to avoid depending on runtime.

- **Core types lack a shared root.** `Message` and `Usage` live in `pkg/chat`, `Tool` and `ToolCall` in `pkg/tools`, `Provider` in `pkg/model/provider`, error handling in `pkg/modelerrors`. These all describe the same domain — LLM interaction — but there is no common package they stem from, which leads to cross-cutting imports and circular dependency pressure as new features are added.

- **Extending means working around coupling.** Adding a new capability that needs model access (evaluation, RAG rewriting, structured extraction) requires either importing runtime with all its dependencies or duplicating the call-and-stream pattern. Neither scales well.

- **One shape for all use cases.** The runtime is designed for long-running interactive agents. But not every model interaction is agentic — some are one-shot completions, some are structured data extraction, some are background tasks. These different shapes are forced through the same path today.

## Goal

[...]

PR #2443, opened 2026-04-15
core LLM completion layer

shaj13:
Add new pkg/ai package that extracts and centralizes model interaction logic from runtime. 

See #2409


shaj13:
@dgageot mind taking a look ? Extracted the LLM interaction logic from `pkg/runtime` and other packages into a new reusable `pkg/ai` package. This includes streaming, fallback handling, retry, and tool execution — enabling consistent model communication across the codebase. The package is also can be used as a standalone Go library without requiring the full agent runtime.

shaj13:
@dgageot just a gentle reminder — any update on this PR ?

shaj13:
@dgageot just a gentle reminder — any update on this PR ?

aheritier (changes requested):
Good refactoring — the `pkg/ai` package has a clean API, proper separation of concerns, and solid test coverage with golden files. The interceptor pattern is well designed.

A few items need attention before merge:

1. **Merge conflicts** — branch is in a dirty state, needs rebase.
2. **`WithRetries` semantics mismatch** — doc says "n + 1 attempts" but code gives exactly n (see inline).
3. **Token usage accumulation bug** — loses data when `r.Usage != nil && r2.Usage == nil` (see inline).
4. **Unguarded `slices.IndexFunc` result** — can panic if model ID not found (see inline).
5. **Cooldown behavior change** — the old code only set cooldown when the primary failed with a non-retryable error; the new code always sets cooldown when any fallback succeeds. This is a semantic change that may cause undesired pinning after transient retryable failures exhaust retries.

CI is green on the last push (lint, test, build all pass).

pkg/ai/option.go
```diff
@@ -0,0 +1,117 @@
...
+	return func(c *completion) {
+		c.tools = append(c.tools, t...)
+	}
+}
+
+// WithRetries sets the number of retry attempts per model on retryable
+// errors (5xx, timeouts). The total attempts per model is n + 1.
+func WithRetries(n int) Option {
+	return func(c *completion) {
+		c.retries = n
```
aheritier, line 49: **Blocking**: The doc says "The total attempts per model is n + 1" but the code uses `for retry := range c.retries` which yields exactly `n` iterations. `WithRetries(3)` = 3 attempts, not 4.

Either fix the doc to "The total attempts per model is n" or change the loop to `range c.retries + 1`.

pkg/ai/completion.go
```diff
@@ -0,0 +1,474 @@
...
+			InputTokens:       r.Usage.InputTokens + r2.Usage.InputTokens,
+			OutputTokens:      r.Usage.OutputTokens + r2.Usage.OutputTokens,
+			CachedInputTokens: r.Usage.CachedInputTokens + r2.Usage.CachedInputTokens,
+			CacheWriteTokens:  r.Usage.CacheWriteTokens + r2.Usage.CacheWriteTokens,
+			ReasoningTokens:   r.Usage.ReasoningTokens + r2.Usage.ReasoningTokens,
+		}
+	}
+
+	return r2, nil
+}
```
aheritier, line 474: **Non-blocking**: If `r.Usage != nil` but `r2.Usage == nil` (provider didn't return usage on the recursive call), the first call's token counts are silently lost. Consider:

```go
switch {
case r2.Usage != nil && r.Usage != nil:
    r2.Usage = &chat.Usage{
        InputTokens:       r.Usage.InputTokens + r2.Usage.InputTokens,
        OutputTokens:      r.Usage.OutputTokens + r2.Usage.OutputTokens,
        CachedInputTokens: r.Usage.CachedInputTokens + r2.Usage.CachedInputTokens,
        CacheWriteTokens:  r.Usage.CacheWriteTokens + r2.Usage.CacheWriteTokens,
        ReasoningTokens:   r.Usage.ReasoningTokens + r2.Usage.ReasoningTokens,
    }
case r2.Usage == nil && r.Usage != nil:
    r2.Usage = r.Usage
}
```

pkg/ai/completion.go
```diff
@@ -0,0 +1,474 @@
...
+	if c.models[0].ID() != r.Model {
+		idx := slices.IndexFunc(c.models, func(m provider.Provider) bool {
+			return m.ID() == r.Model
+		})
+
+		// Rotate to put the responding model first.
+		c.models = append(c.models[idx:], c.models[:idx]...)
+	}
+
+	r2, err := c.generate(ctx)
```
aheritier, line 455: **Non-blocking**: `slices.IndexFunc` returns -1 when no match is found. If `r.Model` doesn't match any entry in `c.models` (e.g. a router provider exposing a different sub-model ID), this will panic on `c.models[-1:]`.

Add a guard:
```go
idx := slices.IndexFunc(c.models, func(m provider.Provider) bool {
    return m.ID() == r.Model
})
if idx > 0 {
    c.models = append(c.models[idx:], c.models[:idx]...)
}
```

pkg/runtime/fallback.go
```diff
@@ -184,246 +147,76 @@ func (r *LocalRuntime) tryModelWithFallback(
...
 	}
 
-	if !retryable {
-		slog.Error("Non-retryable error from model",
-			"agent", a.Name(),
-			"model", modelEntry.provider.ID(),
-			"error", err)
-		if !modelEntry.isFallback {
-			*primaryFailedWithNonRetryable = true
+	// Handle cooldown state based on which model succeeded
```
aheritier, line 209: **Question**: This now unconditionally sets cooldown whenever a fallback succeeds. The old code only set cooldown when `primaryFailedWithNonRetryable` was true (429 or 4xx). If the primary exhausts retries on transient 5xx errors and a fallback picks up, the runtime will now pin to that fallback for the full cooldown window — previously it wouldn't.

Is this intentional? If so, it's probably fine (simpler), but worth noting as a behavior change. If not, `pkg/ai` would need to surface why the fallback was triggered (non-retryable vs exhausted retries).

pkg/sessiontitle/generator.go
```diff
@@ -108,65 +105,22 @@ func (g *Generator) Generate(ctx context.Context, sessionID string, userMessages
...
+		ai.WithModels(models...),
+		ai.WithMessages(messages...),
+		ai.WithRequireContent(),
+		ai.WithLogger(lg),
+	)
+	if err != nil {
+		return "", fmt.Errorf("generating title failed: %w", err)
 	}
-	return "", nil
+
```
aheritier, line 121: **Praise**: Nice simplification of the title generator. The old manual stream-drain + per-model retry loop collapses cleanly into a single `ai.GenerateText` call with `WithRequireContent`.

aheritier (commented):
Inline comments attached.

aheritier:

dgageot:

PR #2458, opened 2026-04-17
fix: retry LSP/MCP toolsets after tool calls, covering env-wrapped commands (fixes #2457)

### LINKED ISSUE #2457: Missing LSPs/MCPs should be retried on next turn / after tool calls, with AGENTS.md install instructions
## Summary

When an LSP (e.g. `gopls`) or MCP is not available at session start, the agent silently misses those capabilities for the entire session — even if the tool becomes available later (e.g. after the agent follows install instructions from `AGENTS.md`).

## Current behaviour

LSPs and MCPs are resolved once at session startup. If `gopls` is not yet installed when the session begins, the agent never gains access to it — even after installing it mid-session by following instructions in `AGENTS.md` (or similar onboarding docs). There is no retry, re-probe, or re-initialisation mechanism.

## Expected behaviour

The agent should re-attempt to connect to any configured-but-missing LSP or MCP:

- **After each tool call** (or at least after a shell/exec tool call that may have installed new software), or
- **At the start of each new turn**, so that tools installed during a session become available without requiring a full restart.

## Motivation / example

`AGENTS.md` in a repo can include setup instructions for required tooling — for example:

```bash
# Install Go and gopls
curl -fsSL "https://go.dev/dl/go1.26.0.linux-amd64.tar.gz" | tar -C /usr/local -xz
go install golang.org/x/tools/gopls@latest
ln -sf /root/go/bin/gopls /usr/local/bin/gopls
```

An agent can read `AGENTS.md` and faithfully execute those commands — Go and `gopls` end up installed and on `PATH` — but the gopls MCP/LSP toolset remains unavailable for the rest of the session because the runtime never re-checked.

This creates a pattern that should work but doesn't: *let the repo describe its own tooling requirements, let the agent install them, and then let the agent use them immediately.*

## Suggested fix

- After any tool call (especially shell/exec), re-probe configured LSP/MCP endpoints that previously failed to connect.
- Alternatively, retry on each new user turn with a short timeout.
- This would also naturally support cases where an MCP server takes a moment to start up after being launched in a background job.

## Environment

Observed while working in the `docker/agentic-platform` repo: `AGENTS.md` contains Go + gopls install instructions, the agent successfully installed both, but gopls-based tools remained unavailable for the remainder of the session.

simonferquel-clanker:
## Summary

Fixes #2457.

When an MCP server is configured but not available at session start, the agent now retries **within the same user turn** — after each tool-call batch — without requiring a new user message.

### MCP double-watcher race (B1)

`Toolset.Start()` previously spawned `go watchConnection(...)` unconditionally. With the reprobe loop calling `Toolset.Start()` during reconnect backoff (when `ts.started==false` but the goroutine is alive), a second watcher goroutine could be spawned, causing racing `doStart()` calls and unsafe close/recreate of `ts.restarted`.

Fix: added `watcherAlive bool` to `Toolset`. `Start()` only spawns the watcher when `!watcherAlive`; the goroutine clears the flag on all exit paths via `defer`.

### Reprobe after each tool-call batch

Added `reprobe()` called after every tool-call batch in the runtime loop. It re-runs `ensureToolSetsAreStarted()` without emitting `MCPInitStarted/Finished` events (no TUI spinner flicker). If new tools appear (by name-set diff), it emits a ToolsetInfo update immediately. The updated tool list is picked up by the next iteration's `getTools()` — so the model sees new tools **within the same user turn**, without a new user message.

### Deduplicated warnings + recovery notices

`StartableToolSet` tracks failure streaks with `ShouldReportFailure()` (emits exactly one warning per streak) and `ConsumeRecovery()` (emits exactly one "now available" warning when a toolset recovers). Both surface via `WarningEvent` → `notification.WarningCmd()` — persistent TUI notifications that stay until dismissed.

## Changes

| File | What changed |
|---|---|
| `pkg/tools/mcp/mcp.go` | `watcherAlive` flag; gated watcher spawn; `watchConnection` clears flag on exit |
| `pkg/tools/startable.go` | `ShouldReportFailure()`, `ConsumeRecovery()`, failure-streak tracking fields |
| `pkg/agent/agent.go` | `ensureToolSetsAreStarted` uses dedup methods; `DrainWarnings` includes recovery notices |
| `pkg/runtime/loop.go` | `reprobe()` after each tool-call batch; `emitLifecycleEvents` param on `getTools()` |
| `pkg/runtime/event.go` | `WarningEvent` used for both failures and recovery |
| `pkg/tui/page/chat/runtime_events.go` | `WarningEvent` → `notification.WarningCmd()` |

## Tests

- `pkg/tools/startable_test.go`: `ShouldReportFailure`/`ConsumeRecovery` behaviour (one warning per streak, recovery fires once)
- `pkg/agent/agent_test.go`: `TestAgentReProbeEmitsWarningThenNotice`, `TestAgentNoDuplicateStartWarnings`
- `pkg/runtime/runtime_test.go`: `TestReprobe_NewToolsAvailableAfterToolCall`, `TestReprobe_NoChangeMeansNoExtraEvents`

## tmux e2e (MCP, validated locally)

Single user turn: agent runs `touch /flag` → reprobe fires → next model response uses new MCP tool natively.

```
✓ Shell  touch /tmp/e2e-test/mcp-ready
✓ hello  name=World  Hello, World! MCP is working.
Tools sidebar: 6 tools available  ← was 5
```

### DIFF
--- a/pkg/tools/mcp/mcp.go
+++ b/pkg/tools/mcp/mcp.go
@@ -51,6 +51,7 @@ type Toolset struct {
 	mu           sync.Mutex
 	started      bool
 	stopping     bool // true when Stop() has been called
+	watcherAlive bool // true while the watchConnection goroutine is running
 
 	// Cached tools and prompts, invalidated via MCP notifications.
 	// cacheGen is bumped on each invalidation so that a concurrent
@@ -178,12 +179,16 @@ func (ts *Toolset) Start(ctx context.Context) error {
 
 	ts.started = true
 
-	// Spawn the connection watcher only on the initial Start.
-	// Restarts from within watchConnection call doStart directly
-	// and must NOT spawn an additional watcher goroutine.
-	// Use WithoutCancel so the watcher outlives the caller's context;
-	// the only way to stop it is via Stop() setting ts.stopping.
-	go ts.watchConnection(context.WithoutCancel(ctx))
+	// Spawn the connection watcher only when no watcher is already running.
+	// A watcher goroutine survives across restarts (it loops inside
+	// watchConnection); reprobe may call Start() while that goroutine is
+	// mid-restart with started==false, and we must not spawn a second one.
+	if !ts.watcherAlive {
+		ts.watcherAlive = true
+		// Use WithoutCancel so the watcher outlives the caller's context;
+		// the only way to stop it is via Stop() setting ts.stopping.
+		go ts.watchConnection(context.WithoutCancel(ctx))
+	}
 
 	return nil
 }
@@ -281,9 +286,15 @@ func (ts *Toolset) doStart(ctx context.Context) error {
 
 // watchConnection monitors the MCP server connection and auto-restarts it
 // if the server dies unexpectedly (i.e. we didn't call Stop()).
-// Only one watchConnection goroutine exists per Toolset; it is spawned by
-// Start() and loops across restarts without spawning additional goroutines.
+// Exactly one watchConnection goroutine exists per Toolset while ts.watcherAlive
+// is true; it is spawned by Start() and cleared on exit.
 func (ts *Toolset) watchConnection(ctx context.Context) {
+	defer func() {
+		ts.mu.Lock()
+		ts.watcherAlive = false
+		ts.mu.Unlock()
+	}()
+
 	for {
 		err := ts.mcpClient.Wait()
 
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -260,10 +260,23 @@ func (a *Agent) ToolSets() []tools.ToolSet {
 func (a *Agent) ensureToolSetsAreStarted(ctx context.Context) {
 	for _, toolSet := range a.toolsets {
 		if err := toolSet.Start(ctx); err != nil {
-			desc := tools.DescribeToolSet(toolSet)
-			slog.Warn("Toolset start failed; skipping", "agent", a.Name(), "toolset", desc, "error", err)
-			a.addToolWarning(fmt.Sprintf("%s start failed: %v", desc, err))
+			// Only warn on the first failure in a streak; suppress duplicate
+			// warnings for subsequent retries that also fail.
+			if toolSet.ShouldReportFailure() {
+				desc := tools.DescribeToolSet(toolSet)
+				slog.Warn("Toolset start failed; will retry on next turn", "agent", a.Name(), "toolset", desc, "error", err)
+				a.addToolWarning(fmt.Sprintf("%s start failed: %v", desc, err))
+			} else {
+				desc := tools.DescribeToolSet(toolSet)
+				slog.Debug("Toolset still unavailable; retrying next turn", "agent", a.Name(), "toolset", desc, "error", err)
+			}
 			continue
+		}
+		// Emit a one-time notice when a previously-failed toolset recovers.
+		if toolSet.ConsumeRecovery() {
+			desc := tools.DescribeToolSet(toolSet)
+			slog.Info("Toolset now available", "agent", a.Name(), "toolset", desc)
+			a.addToolWarning(desc + " is now available")
 		}
 	}
 }
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -44,6 +44,34 @@ func (s *stubToolSet) Tools(context.Context) ([]tools.Tool, error) {
 	return s.tools, nil
 }
 
+// flappyToolSet is a ToolSet+Startable that returns a scripted sequence of
+// errors from Start(). nil in the sequence means success.
+type flappyToolSet struct {
+	errs    []error
+	callIdx int
+	stubs   []tools.Tool
+}
+
+var (
+	_ tools.ToolSet   = (*flappyToolSet)(nil)
+	_ tools.Startable = (*flappyToolSet)(nil)
+)
+
+func (f *flappyToolSet) Start(_ context.Context) error {
+	if f.callIdx >= len(f.errs) {
+		return nil
+	}
+	err := f.errs[f.callIdx]
+	f.callIdx++
+	return err
+}
+
+func (f *flappyToolSet) Stop(_ context.Context) error { return nil }
+
+func (f *flappyToolSet) Tools(_ context.Context) ([]tools.Tool, error) {
+	return f.stubs, nil
+}
+
 func TestAgentTools(t *testing.T) {
 	tests := []struct {
 		name          string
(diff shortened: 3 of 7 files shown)


rumpl:
I'm not sure I understand the fix, isn't the lsp tooset startable? Wouldn't re-starting the lsp toolset be enough?

simonferquel-clanker:
Good question — the LSP toolset *is* Startable, but there are two separate problems that a naive re-start doesn't cover:

**1. `env -C ./backend gopls` looks like a successful start**

When the command is `env -C ./backend gopls`, `exec.Command("env", ...).Start()` always returns `nil` — `env` is always present. The fact that `env` then immediately runs `gopls`, finds it missing, and exits 127 is invisible to `cmd.Start()`. By the time we return from `LSPTool.Start()` / `startLocked()`, `h.cmd` has been set and the call returned nil, so `StartableToolSet` sets `started = true` and considers the toolset live.

A subsequent call to `Start()` hits the `if s.started { return nil }` short-circuit and never retries — even though the underlying process died in milliseconds.

**2. The `HealthChecker` interface closes that gap**

The fix adds a `HealthChecker` interface (`Healthy() bool`) that `StartableToolSet` checks on *every* `Start()` call even when `started == true`. `LSPTool.Healthy()` returns `h.cmd != nil`.

The new `watchExit` goroutine is the mechanism that transitions `h.cmd` from non-nil to nil: it is the sole owner of `cmd.Wait()`, blocks until the process exits, then resets `h.cmd = nil`. So after `env gopls` exits 127, `Healthy()` returns `false`, `StartableToolSet` resets `started = false`, and the next `Start()` call actually retries — which will succeed once `gopls` has been installed.

**3. The reprobe loop ties it together within a single turn**

Even with the above, without the reprobe the agent would only see new tools at the start of the *next user turn* (when `getTools()` is called at the top of the run loop). The `reprobe()` call after each tool-call batch re-runs `ensureToolSetsAreStarted` immediately, so if a shell tool call installed `gopls`, the very next model response in the same turn already has the LSP tools in its available-tool list — no new user message needed.

simonferquel (changes requested):
With https://github.com/docker/docker-agent/pull/2460 about to get merged, the `env gopls` thing is moot. This PR should not try to cover this pattern anymore

simonferquel-clanker:
Done — removed in commit `0c65de91`.

**Removed** (everything that was specifically for the `env -C` wrapper pattern):
- `HealthChecker` interface and the health-check block in `StartableToolSet.Start()`
- `LSPTool.Healthy()`, `watchExit` goroutine, `dead atomic.Bool`, `exited chan`
- `stopAndWait` reverted to `stopLocked` with direct `cmd.Wait()`
- `Toolset.Healthy()` on the MCP side (no longer needed without the interface)
- All related tests (`TestLSPTool_Healthy_*`, `TestLSPTool_StartableToolSet_RetriesAfterProcessExits`, `TestStartableToolSet_HealthChecker_*`)

**Retained** (the actual fix from this PR):
- `watcherAlive` flag in MCP `Toolset` + gated watcher spawn (B1: prevents double `watchConnection` goroutine on reprobe)
- `reprobe()` in the runtime loop (fires after each tool-call batch, no `MCPInit*` spinner noise)
- `ShouldReportFailure` / `ConsumeRecovery` deduplication and recovery notices
- All agent-level and runtime-level reprobe tests

krissetto:
use the proper notification component when surfacing things to the TUI. A warning one in this case might be appropriate

simonferquel-clanker:
Done in commit `ad1b1734`.

Dropped the separate `NoticeEvent`/`InfoCmd` path entirely. Both start-failure warnings and toolset recovery messages now go through the existing `WarningEvent` → `notification.WarningCmd()` path.

The key distinction: `TypeInfo` notifications auto-hide after 3 seconds; `TypeWarning` notifications are **persistent** and stay until the user dismisses them — which is the right behaviour for toolset availability state changes that the user should not miss.

Changes:
- `NoticeEvent`, `Notice()`, `formatToolNotice` removed from runtime
- `DrainNotices`, `addToolNotice`, `pendingNotices` removed from agent
- `ConsumeRecovery()` now calls `addToolWarning("X is now available")`
- `emitAgentWarnings()` simplified back to draining warnings only
- `NoticeEvent` TUI handler and `"notice"` client type registration removed

pkg/tools/builtin/lsp.go
```diff
@@ -532,7 +532,7 @@ func (h *lspHandler) stopLocked() error {
 	h.stdin.Close()
 
 	// Cancel the process-lifetime context to stop the readNotifications
-	// goroutine and (if the process didn't exit cleanly) kill the process.
+	// goroutine and force-kill the process if it hasn't exited cleanly.
```
simonferquel: NIT: revert that comment only change

pkg/tools/builtin/lsp_test.go
```diff
@@ -21,12 +21,12 @@ func TestLSPTool_Tools(t *testing.T) {
 	t.Parallel()
 
 	tool := NewLSPTool("gopls", nil, nil, "/tmp")
-	tools, err := tool.Tools(t.Context())
```
simonferquel, line 24: all these files changes should be reverted

pkg/tools/mcp/mcp.go
```diff
@@ -168,12 +169,16 @@ func (ts *Toolset) Start(ctx context.Context) error {
 
 	ts.started = true
 
-	// Spawn the connection watcher only on the initial Start.
-	// Restarts from within watchConnection call doStart directly
-	// and must NOT spawn an additional watcher goroutine.
-	// Use WithoutCancel so the watcher outlives the caller's context;
-	// the only way to stop it is via Stop() setting ts.stopping.
-	go ts.watchConnection(context.WithoutCancel(ctx))
+	// Spawn the connection watcher only when no watcher is already running.
```
simonferquel, line 182: is this file changes still needed with the env -C concern removal ?

simonferquel-clanker: Yes — this change is independent of the `env -C` concern and is still needed.

The `watcherAlive` flag fixes a race introduced by the **reprobe loop itself**:

1. `watchConnection` sets `ts.started = false` at the top of its reconnect iteration (before sleeping its backoff), while the goroutine is still alive.
2. If `reprobe()` fires during that window, `StartableToolSet.Start()` sees `started == false` and calls `Toolset.Start()`.
3. Without the guard, `Toolset.Start()` would unconditionally spawn a second `go watchConnection(...)`, resulting in two goroutines racing to call `doStart()` and unsafely close/recreate `ts.restarted`.

The fix: `Toolset.Start()` only spawns `watchConnection` when `!watcherAlive`. The goroutine clears the flag on all exit paths via `defer`. While an existing watcher is mid-backoff, reprobe skips the spawn. Once the watcher truly gives up, the next reprobe starts a fresh one.

pkg/tools/builtin/lsp_test.go
simonferquel-clanker: Done — `lsp_test.go` fully reverted to `upstream/main` in commit `65d42382`. No changes from this PR remain in `pkg/tools/builtin/`.

pkg/tools/builtin/lsp.go
```diff
@@ -346,6 +346,12 @@ func (t *LSPTool) SetFileTypes(fileTypes []string) {
 	t.handler.fileTypes = fileTypes
 }
 
+// WorkingDir returns the working directory of the LSP server process.
+// This is intended for testing only.
+func (t *LSPTool) WorkingDir() string {
```
simonferquel, line 351: if it is test only, it should be in lsp_test.go

pkg/tools/builtin/lsp.go
simonferquel-clanker: Done — `lsp.go` reverted to `upstream/main` in commit `65d42382`. Both `lsp.go` and `lsp_test.go` are now identical to upstream.

pkg/tools/builtin/lsp.go
simonferquel-clanker: After rebasing onto `upstream/main` (which now includes PR #2460 that added `WorkingDir()`), our branch has no diff in `lsp.go` or `lsp_test.go` — both files are identical to upstream. Nothing to move.

simonferquel (approved):
Approving my own agent feels not enough. cc @krissetto for another approve :)

dgageot (approved):

PR #2492, opened 2026-04-22
fix(runtime): drain steerQueue at top of RunStream loop to close idle-window race

### LINKED ISSUE #2491: Steer queue: drain at RunStream entry to close idle-window race
## Problem

`runtime.LocalRuntime.Steer` enqueues into `steerQueue`, but the queue is only drained inside the RunStream loop — specifically **after a batch of tool calls finishes and before the stop check** (pkg/runtime/loop.go around line 425).

This leaves two gaps where a steered message can be stranded or delivered late:

1. **Idle-runtime race.** A `Steer` call can land between the moment the previous RunStream returns and the moment the next RunStream begins. The queue holds the message, but nothing is running to consume it. The caller has no signal that delivery is deferred.
2. **First-LLM-roundtrip miss.** When a new RunStream starts, it builds `messages := sess.GetMessages(a)` and calls the model without touching `steerQueue`. If that first turn is a plain text response with no tool calls, `res.Stopped` fires and the loop breaks — the pending steer message is never injected into this turn at all.

Combined, a steered message that arrived while the runtime was idle can be invisible to a subsequent short turn and only surface on a much later multi-round turn.

## Why it matters

Any caller that builds message ingestion on top of `WithSteerQueue` needs to be able to enqueue a message and trust it will be picked up by the next turn, regardless of whether the runtime is currently active. Today that guarantee does not hold.

## Proposed fix

Drain `steerQueue` at the top of the RunStream loop body, symmetrically with the existing mid-loop drain, before the first model call. Pseudocode:

```go
// At the start of the for-loop, before the model call:
if steered := r.steerQueue.Drain(ctx); len(steered) > 0 {
    for _, sm := range steered {
        sess.AddMessage(session.UserMessage(wrap(sm), sm.MultiContent...))
        events <- UserMessage(...)
    }
}
```

This makes `Steer` idempotent with respect to runtime-liveness: the caller can always `Enqueue` and be sure the next turn picks it up, regardless of whether the runtime is currently looping.

## Alternative / complement

Expose a documented contract that `RunStream` will call `Drain` once at entry, so any `MessageQueue` implementation can atomically inject messages into the session before the first model call.

## Acceptance criteria

- A `Steer` call made while no RunStream is active is consumed by the next RunStream's first turn.
- A `Steer` call made during a short non-tool turn is still consumed within that same turn (or at worst triggers a follow-on turn) — not stranded until an unrelated future turn.
- No change to follow-up-queue semantics.

simonferquel-clanker:
Fixes #2491

## Problem

`LocalRuntime.Steer` enqueues into `steerQueue`, but the queue was only drained mid-loop — after a batch of tool calls, before the stop check. Two gaps allowed steer messages to be silently dropped:

1. **Idle-window race** — a `Steer` call lands between two `RunStream` invocations. The queue holds the message but nothing drains it when the next `RunStream` starts.
2. **First-turn miss** — when a `RunStream` turn produces a plain-text response with no tool calls, `res.Stopped` fires before the mid-loop drain is reached; the pending steer is never injected.

## Fix

Three steer drain points are added across the `RunStream` loop, giving the invariant: any `Steer` call that enqueues successfully is guaranteed to be consumed within the current `RunStream`, never stranded.

All three drain sites inject steer messages as **plain user messages** — no wrapping, no envelope.

1. **Top-of-turn drain** (after `AgentInfo` and `contextLimit` init, before `GetMessages`): catches idle-window and first-turn-miss races.
2. **Mid-loop drain** (after tool calls, unchanged in placement): catches steers that arrive while tool calls are running.
3. **End-of-iteration drain** (inside `res.Stopped`, before the follow-up queue and `break`): closes the narrow race between the mid-loop drain and the stop decision.

The `appendSteerAndEmit` helper centralises the add-message + emit-event idiom used at all three sites.

## Empty-session bootstrap

A `RunStream` started with zero messages in the session but steer messages already queued injects those messages as plain user messages before the first model call. The model receives them as its initial context.

## Changes

- `pkg/runtime/loop.go` — `appendSteerAndEmit` helper (no wrapping); three drain sites.
- `pkg/runtime/runtime_test.go` — `messageRecordingProvider`, `hookStream`, `steerInjectProvider`; three focused tests:
  - `TestSteer_IdleWindowIsConsumedOnNextTurn`
  - `TestSteer_EmptySessionBootstrap`
  - `TestSteer_EndOfIterationRaceIsConsumedInCurrentRunStream`

### DIFF
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -484,11 +484,12 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 				// stranded until the next RunStream invocation. Re-checking
 				// here closes that race: any message that enqueued successfully
 				// is guaranteed to be consumed within the current RunStream.
-				// The agent just finished a turn, so these messages get the
-				// same system-reminder envelope as mid-turn steers.
+				// The current turn is ending and a new one is about to begin;
+				// the agent is not mid-task, so plain user messages (wrap=false)
+				// are used — same as the top-of-turn drain.
 				if steered := r.steerQueue.Drain(ctx); len(steered) > 0 {
 					for _, sm := range steered {
-						r.appendSteerAndEmit(sess, sm, true, events)
+						r.appendSteerAndEmit(sess, sm, false, events)
 					}
 					r.compactIfNeeded(ctx, sess, a, m, contextLimit, messageCountBeforeTools, events)
 					continue
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -41,27 +41,12 @@ func (r *LocalRuntime) registerDefaultTools() {
 	})
 }
 
-// wrapSteerMessage wraps a raw steer message in a <system-reminder> envelope
-// so the model receives the user's course-correction as a clearly-labelled
-// side-channel injection rather than a plain conversational turn.
-func wrapSteerMessage(content string) string {
-	return fmt.Sprintf(
-		"<system-reminder>\nThe user sent the following message while you were working:\n%s\n\nPlease address this in your next response while continuing with your current tasks.\n</system-reminder>",
-		content,
-	)
-}
-
-// appendSteerAndEmit appends a steered message to the session and emits the
-// corresponding UserMessage event. When wrap is true the content is enclosed
-// in a <system-reminder> envelope (appropriate for mid-turn interruptions
-// where the agent is actively working); when false the raw content is used
-// (appropriate at the top of a new turn where the agent is idle).
-func (r *LocalRuntime) appendSteerAndEmit(sess *session.Session, sm QueuedMessage, wrap bool, events chan<- Event) {
-	content := sm.Content
-	if wrap {
-		content = wrapSteerMessage(sm.Content)
-	}
-	sess.AddMessage(session.UserMessage(content, sm.MultiContent...))
+// appendSteerAndEmit appends a steered message to the session as a plain
+// user message and emits the corresponding UserMessage event. Steer messages
+// are always injected as plain user turns regardless of when they arrive
+// (idle window, mid-turn, or end-of-iteration).
+func (r *LocalRuntime) appendSteerAndEmit(sess *session.Session, sm QueuedMessage, events chan<- Event) {
+	sess.AddMessage(session.UserMessage(sm.Content, sm.MultiContent...))
 	events <- UserMessage(sm.Content, sess.ID, sm.MultiContent, len(sess.Messages)-1)
 }
 
@@ -326,13 +311,13 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 			//      nothing was running to consume the message.
 			//   2. First-turn miss: a plain-text response with no tool calls
 			//      fires res.Stopped before the mid-loop drain is reached.
-			// The agent is not mid-task here, so messages are plain user turns
-			// (no system-reminder envelope). Placement after contextLimit
-			// initialization means compactIfNeeded can be called immediately.
+			// Steer messages are always injected as plain user turns.
+			// Placement after contextLimit initialization means compactIfNeeded
+			// can be called immediately.
 			if steered := r.steerQueue.Drain(ctx); len(steered) > 0 {
 				messageCountBeforeSteer := len(sess.GetAllMessages())
 				for _, sm := range steered {
-					r.appendSteerAndEmit(sess, sm, false, events)
+					r.appendSteerAndEmit(sess, sm, events)
 				}
 				r.compactIfNeeded(ctx, sess, a, m, contextLimit, messageCountBeforeSteer, events)
 			}
@@ -462,12 +447,12 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 			toolModelOverride = resolveToolCallModelOverride(res.Calls, agentTools)
 
 			// --- STEERING: mid-turn injection ---
-			// Drain ALL pending steer messages. These are urgent course-
-			// corrections that the model should see on the very next
-			// iteration, wrapped in <system-reminder> tags.
+			// Drain ALL pending steer messages injected while tool calls were
+			// running. These are plain user messages; the model sees them on
+			// the very next iteration.
 			if steered := r.steerQueue.Drain(ctx); len(steered) > 0 {
 				for _, sm := range steered {
-					r.appendSteerAndEmit(sess, sm, true, events)
+					r.appendSteerAndEmit(sess, sm, events)
 				}
 
 				r.compactIfNeeded(ctx, sess, a, m, contextLimit, messageCountBeforeTools, events)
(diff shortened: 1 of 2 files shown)


pkg/runtime/loop.go
```diff
@@ -249,6 +259,20 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
...
 
+			// --- STEERING: top-of-turn injection ---
+			// Drain any steer messages that arrived while the runtime was idle
+			// (between RunStream invocations) or before the first model call of
+			// this turn. Without this drain, a Steer call made while no tool
+			// calls are running would be silently dropped when the model returns
+			// a plain-text response with no tool calls.
+			if steered := r.steerQueue.Drain(ctx); len(steered) > 0 {
+				for _, sm := range steered {
+					userMsg := session.UserMessage(wrapSteerMessage(sm.Content), sm.MultiContent...)
```
simonferquel: start of turn message should not get the steerMessage envelope (the agent is not working on anything at this point)

simonferquel (changes requested):

simonferquel-clanker:
Fixed — top-of-turn steer messages are now injected as plain user messages (no system-reminder envelope). The wrapper is kept only on the mid-loop drain where the agent is actively working.

simonferquel-clanker:
Updated: top-of-turn steer messages are now injected as plain user messages without the system-reminder envelope. The `wrapSteerMessage` helper is retained only at the mid-loop drain site (where the agent is actively working between tool calls).

simonferquel-clanker:
Addressed all review findings:

- **B1 (blocking):** Top-of-turn steer messages are already plain user messages (no envelope). Tests now explicitly assert both the stored session message content and the message seen by the model equal the raw content and do `NotContains "<system-reminder>"`.
- **S1:** Top-of-turn drain moved to after `contextLimit` initialization (after `AgentInfo` emission). `compactIfNeeded` is called immediately after injection using the captured pre-injection message count.
- **S2:** Resolved naturally — drain is now placed after `AgentInfo`, so event ordering is `AgentInfo` → steer `UserMessage`s → model call.
- **S3:** Extracted `appendSteerAndEmit(sess, sm, wrap bool, events)` helper. Both drain sites use it; `wrap=false` for top-of-turn, `wrap=true` for mid-loop.
- **N4:** `messageRecordingProvider` fallback comment explains the behaviour for unexpected extra calls; assertion tests use `require.Len` / `require.NotEmpty` to catch them.

pkg/runtime/loop.go
```diff
@@ -441,6 +478,22 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
...
+				// A Steer() call that lands in the narrow window between the
+				// mid-loop drain above and this stop-check would otherwise be
+				// stranded until the next RunStream invocation. Re-checking
+				// here closes that race: any message that enqueued successfully
+				// is guaranteed to be consumed within the current RunStream.
+				// The agent just finished a turn, so these messages get the
+				// same system-reminder envelope as mid-turn steers.
+				if steered := r.steerQueue.Drain(ctx); len(steered) > 0 {
+					for _, sm := range steered {
+						r.appendSteerAndEmit(sess, sm, true, events)
```
simonferquel: we are about to stop, those messages should not have an enveloppe

simonferquel-clanker:
Addressed inline change request (discussion_r3123189465): end-of-iteration drain now uses `appendSteerAndEmit(wrap=false)`. At `res.Stopped` the current turn is ending and a new one is about to begin — same semantics as the top-of-turn drain, so plain user messages are correct. Comment in code updated. Test (`hookStream` + `TestSteer_EndOfIterationRaceIsConsumedInCurrentRunStream`) asserts steer consumed within same RunStream: 2 model calls and `UserMessageEvent` present. Commit: 4a179813.

simonferquel:
After discussion, we think we should remove all wrapping: depending on the user message it can create too much confusion. Remove the message wrapper alltogether

simonferquel (changes requested):
Be more concise in your code comments

simonferquel (commented):
lgtm, asking for a 2nd review

pkg/runtime/loop.go
```diff
@@ -294,6 +300,16 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 				}
 			}
 
+			// Drain steer messages queued while idle or before the first model call
+			// (covers idle-window and first-turn-miss races).
+			if steered := r.steerQueue.Drain(ctx); len(steered) > 0 {
```
rumpl, line 305: I don't understand why this one is needed here?

simonferquel-clanker:
Re: https://github.com/docker/docker-agent/pull/2492#discussion_r3123963222

The mid-loop drain and the `res.Stopped` break are not atomic — there is a narrow window between them:

```
steered := r.steerQueue.Drain(ctx)  // ← drains the queue
if len(steered) > 0 { continue }
if res.Stopped { break }            // ← Steer() arriving here is missed
```

A `Steer()` call landing in that window is stranded: on a plain-text turn (no tool calls) there is no subsequent mid-loop drain to pick it up, so it waits for the next `RunStream` invocation — the same idle-window problem the top-of-turn drain fixes. The re-check at `res.Stopped` closes this last window so any successfully-enqueued steer is guaranteed to be consumed within the current `RunStream`.

pkg/runtime/loop.go
simonferquel: 2 cases:
- if a user sends a steering message while the turn is being ended, you end up with the steering messages being sent after the next tool call of the next turn instead of immediately
- when used as a library, in an app that consider all messages as "interrupting" - a la Claude Code - this simplify things a lot: you do not have to treat messages sent while the runner is idle differently than when it is active: they all come from the same queue

rumpl:
> Re: [#2492 (comment)](https://github.com/docker/docker-agent/pull/2492#discussion_r3123963222)
> 
> The mid-loop drain and the `res.Stopped` break are not atomic — there is a narrow window between them:
> 
> ```
> steered := r.steerQueue.Drain(ctx)  // ← drains the queue
> if len(steered) > 0 { continue }
> if res.Stopped { break }            // ← Steer() arriving here is missed
> ```
> 
> A `Steer()` call landing in that window is stranded: on a plain-text turn (no tool calls) there is no subsequent mid-loop drain to pick it up, so it waits for the next `RunStream` invocation — the same idle-window problem the top-of-turn drain fixes. The re-check at `res.Stopped` closes this last window so any successfully-enqueued steer is guaranteed to be consumed within the current `RunStream`.

Ok but there is already a streer.drain that was added in the res.stopped case

simonferquel-clanker:
Exactly — that drain is the fix. The explanation was describing why it was needed, not a gap that still exists. The code is complete. Happy to get your re-approval if you have no further concerns.

pkg/runtime/loop.go
rumpl: I'm not talking about the case when the turn is being ended, that's lower in the code. I'm asking why are we steering _before_ the turn starts

rumpl (approved):

simonferquel (approved):

PR #2510, opened 2026-04-25
feat: add `docker agent serve chat` command (OpenAI-compatible API)

### LINKED ISSUE #2502: Support for OpenAI-compatible API server
As I understand it, the API server isn't OpenAI-compatible right now. Are there any plans to change that? 

It would be great to have a standard `/v1/chat/completions` endpoint so we can plug third-party UIs like Open WebUI straight into the API Server.

dgageot:
Fixes #2502.

Exposes any docker-agent agent through an OpenAI-compatible HTTP server, so any tool that already speaks the Chat Completions protocol (Open WebUI, the official `openai` SDKs, ad-hoc curl scripts, etc.) can drive an agent without a custom integration.

## Endpoints

| Method | Path | Notes |
|---|---|---|
| `GET` | `/v1/models` | Lists exposed agents as OpenAI models |
| `POST` | `/v1/chat/completions` | Runs the agent; supports `stream: true` (SSE) and `false` |

## Usage

```sh
docker agent serve chat ./agent.yaml                       # localhost:8083
docker agent serve chat ./team.yaml --agent reviewer       # pin one agent
docker agent serve chat agentcatalog/pirate --listen :9090
```

```sh
curl -sS -X POST http://127.0.0.1:8083/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"messages":[{"role":"user","content":"hi"}]}'
```

## Design

- The team is loaded **once** at startup and shared across requests. Each chat completion gets a fresh session and runtime.
- The session is created with `ToolsApproved=true` and `NonInteractive=true` — there is no human in the loop. `ElicitationRequestEvent` is still explicitly declined to avoid hanging on the runtime's elicitation channel.
- The `model` field of the request can pin a specific agent in a multi-agent team. If it doesn't match an exposed agent (e.g. clients that hard-code `gpt-4`) we silently fall back to the default agent and echo the requested model name back, so clients matching on the model field stay happy.
- Streaming uses SSE in OpenAI's `chat.completion.chunk` format and ends with `data: [DONE]`.

## Implementation

- New cobra command `cmd/root/chat.go` (default `127.0.0.1:8083`, `--agent` / `--listen` flags) wired into `cmd/root/serve.go`.
- New `pkg/chatserver` package, split across:
  - `server.go` — `Run`, router, HTTP handlers, `sseStream`, error envelope
  - `agent.go` — `agentPolicy`, `buildSession`, `runAgentLoop`, `sessionUsage`
  - `types.go` — request/response shapes
- Reuses `openai.Model` from `github.com/openai/openai-go/v3` for `/v1/models`. Other SDK response types serialise too noisily with stdlib `encoding/json` (the SDK relies on its internal `apijson` package, which lives under `internal/`), so the chat-completion shapes are hand-rolled for clean output.

## Tests

- Unit tests for session-building, agent-policy resolution, usage extraction.
- HTTP-level tests via `httptest` for `/v1/models` shape, the three early-validation paths of `/v1/chat/completions` (bad JSON, empty messages, history without user), and `writeError`'s status→type mapping.

## Validation

- `mise lint` — 0 issues
- `mise test` — all packages green
- Manual curl smoke test against `examples/42.yaml`: `/v1/models` returns the agent, error paths return correct OpenAI-shaped envelopes.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,5 +7,6 @@
 !./**/*.css
 !./**/*.go
 !./**/*.txt
+!/pkg/chatserver/openapi.json
 !/pkg/config/builtin-agents/*.yaml
 !/pkg/tui/styles/themes/*.yaml
--- a/cmd/root/chat.go
+++ b/cmd/root/chat.go
@@ -1,0 +1,62 @@
+package root
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/docker/docker-agent/pkg/chatserver"
+	"github.com/docker/docker-agent/pkg/cli"
+	"github.com/docker/docker-agent/pkg/config"
+	"github.com/docker/docker-agent/pkg/telemetry"
+)
+
+type chatFlags struct {
+	agentName  string
+	listenAddr string
+	runConfig  config.RuntimeConfig
+}
+
+func newChatCmd() *cobra.Command {
+	var flags chatFlags
+
+	cmd := &cobra.Command{
+		Use:   "chat <agent-file>|<registry-ref>",
+		Short: "Start an agent as an OpenAI-compatible chat completions server",
+		Long: `Start an HTTP server that exposes the agent through an OpenAI-compatible
+API at /v1/chat/completions and /v1/models. This lets tools that already
+speak OpenAI's chat protocol (such as Open WebUI) drive a docker-agent
+agent without any custom integration.`,
+		Example: `  docker-agent serve chat ./agent.yaml
+  docker-agent serve chat ./team.yaml --agent reviewer
+  docker-agent serve chat agentcatalog/pirate --listen 127.0.0.1:9090`,
+		Args: cobra.ExactArgs(1),
+		RunE: flags.runChatCommand,
+	}
+
+	cmd.Flags().StringVarP(&flags.agentName, "agent", "a", "", "Name of the agent to expose (all agents if not specified)")
+	cmd.Flags().StringVarP(&flags.listenAddr, "listen", "l", "127.0.0.1:8083", "Address to listen on")
+	addRuntimeConfigFlags(cmd, &flags.runConfig)
+
+	return cmd
+}
+
+func (f *chatFlags) runChatCommand(cmd *cobra.Command, args []string) (commandErr error) {
+	ctx := cmd.Context()
+	telemetry.TrackCommand(ctx, "serve", append([]string{"chat"}, args...))
+	defer func() { // do not inline this defer so that commandErr is not resolved early
+		telemetry.TrackCommandError(ctx, "serve", append([]string{"chat"}, args...), commandErr)
+	}()
+
+	out := cli.NewPrinter(cmd.OutOrStdout())
+	agentFilename := args[0]
+
+	ln, cleanup, err := newListener(ctx, f.listenAddr)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	out.Println("Listening on", ln.Addr().String())
+	out.Println("OpenAI-compatible chat completions endpoint: http://" + ln.Addr().String() + "/v1/chat/completions")
+
+	return chatserver.Run(ctx, agentFilename, f.agentName, &f.runConfig, ln)
+}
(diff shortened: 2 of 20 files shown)


trungutt (dismissed):

dgageot:
## Update — expanded scope

I pushed a force-update onto this branch with 18 additional commits on top of the original `feat: add docker agent serve chat command`. Many of them implement the "easy wins" from the original PR's "Limitations / future work" section, plus a working Go example, plus a couple of bug fixes uncovered while reviewing the new code.

### New commits (oldest → newest)

#### Example
- `examples: add minimal chat client for docker agent serve chat` — `examples/chat/main.go` runs an interactive REPL against the server using the official `github.com/openai/openai-go` SDK, demonstrating the OpenAI compatibility end-to-end.

#### Hardening (trivial / opt-in, safe defaults)
- `chatserver: replace * CORS with --cors-origin flag` — wildcard removed; CORS off by default.
- `chatserver: enforce max body size and per-request timeout` — `--max-request-size` (1 MiB default) and `--request-timeout` (5 min default).
- `chatserver: collect every runtime ErrorEvent (errors.Join)` — no more swallowed errors after the first.
- `chatserver: emit structured error events on streaming failures` — proper `finish_reason: "error"` + error envelope, instead of `[error: …]` jammed into `delta.content`.
- `chatserver: parse and validate OpenAI sampling parameters` — `temperature`, `top_p`, `max_tokens`, `stop` (string-or-array union) are now declared, range-checked and rejected with 400 on bad input.

#### Auth & deployment
- `chatserver: add Bearer-token auth (--api-key)` — opt-in static bearer token (constant-time compare; OPTIONS preflight + `/openapi.json` exempted).
- `chatserver: support comma-list and regex in --cors-origin` — allow-list, `~regex` patterns, scheme/path validation.

#### Performance
- `chatserver: support X-Conversation-Id for stateful sessions` — opt-in LRU + TTL cache so clients don't have to resend the full history every turn (`--conversations-max`, `--conversation-ttl`).
- `chatserver: pool runtimes per agent for warm reuse` — small pool of warm runtimes per agent to avoid the per-request cost of `runtime.New` (`--max-idle-runtimes`).

#### Protocol surface
- `chatserver: surface agent tool calls as OpenAI tool_calls` — agent-invoked tools are emitted in OpenAI's `tool_calls` shape on both streaming and non-streaming responses (informational; tools still execute server-side).
- `chatserver: serve /openapi.json for schema introspection` — embedded OpenAPI 3.1 document; bypasses bearer auth.
- `chatserver: accept OpenAI multimodal content (text + image_url)` — `content` accepts the union of string and typed-parts arrays; image parts now reach the runtime via `chat.MultiContent`.

#### Bug fixes (found by review pass)
- `fix(chatserver): always store conversation after request` — `maybeStoreConversation` used to skip `Put` for existing conversations, so if the entry was evicted by another request mid-flight the updated session was lost.
- `refactor(chatserver): remove unused isNew parameter` (follow-up).
- `test(chatserver): add test for conversation restore after eviction` (follow-up).
- `chatserver: restore doc comment on chatCompletion` — fixes a doc-comment regression introduced by the eviction fix.
- `chatserver: serialize requests sharing an X-Conversation-Id` — concurrent requests sharing a conversation id used to share the same `*session.Session` and race on it. We now reject the second concurrent request with **409 Conflict**, surfacing the misuse instead of producing a garbled transcript. Race-detector clean.

### CI

- `go build ./...` — clean.
- `go test -race ./pkg/chatserver/... ./cmd/root/...` — clean.
- `golangci-lint run ./pkg/chatserver/... ./cmd/root/... ./examples/chat/...` — 0 issues.

### Breaking changes

None on the wire (the API only gains new optional features). For programmatic Go callers of `chatserver.Run`, the signature has changed from
`Run(ctx, agentFilename, agentName, runConfig, ln)` to `Run(ctx, agentFilename, opts Options, ln)`; nothing outside this PR uses it.

Happy to split this into separate PRs if reviewers prefer; commit messages are written to be cherry-pickable.

Assisted-By: docker-agent

trungutt (approved):

PR #2518, opened 2026-04-27
runtime: append newline to non-last steer messages on multi-drain

### LINKED ISSUE #2517: Runtime steer-queue drain concatenates messages without separator, causing run-on tokenisation
## Problem

When the runtime drains multiple queued steer messages and assembles them into user messages sent to the LLM, the messages are added back-to-back without any separator between them.

Because most LLM providers send the full conversation to the model as a continuous byte stream (or token sequence), adjacent words from separate messages are tokenised together as a single run. For example, three consecutive steer messages `"to"`, `"get"`, `"her"` produce the token `"together"` rather than three distinct words.

## Root cause

In `pkg/runtime/loop.go`, every drain site iterates over the drained slice and calls `appendSteerAndEmit` for each message without appending any whitespace or newline between messages:

```go
for _, sm := range steered {
    r.appendSteerAndEmit(sess, sm, events)
}
```

## Fix

Append `\n` to the content of every **non-last** message in the drained batch before injecting it. Messages remain independent (not merged into one). For multi-content messages the `\n` is appended to the last text part; if no text part exists, a new one is appended.

Implemented in a new `drainAndEmitSteered` helper (replacing the three inline drain loops) with a pure helper `appendNewlineToQueuedMessage`.

Related: https://github.com/docker/docker-agent/pull/2516 (trimming behaviour fix)

simonferquel-clanker:
Fixes #2517

## Problem

When the runtime drains multiple queued steer messages in one batch, each was injected into the session back-to-back with no separator. Some chat templates concatenate consecutive `role=user` messages before tokenisation, causing adjacent word fragments from different messages to be glued together.

## Fix

- New `drainAndEmitSteered()` helper replaces the three inline `steerQueue.Drain()` + for-loop sites in `loop.go`.
- `appendNewlineToQueuedMessage()` appends `\n` to every non-last message:
  - Plain-text messages: `\n` appended to `Content`.
  - Multi-content messages: `\n` appended to the last `text` part's `Text` field; if no text part exists, a new one is appended.
  - Original `QueuedMessage` is never mutated (shallow-copy of the parts slice).
- Messages remain **independent** — each is still a separate `role=user` session message.

## Tests

- `TestAppendNewlineToQueuedMessage`: 4 subtests covering plain-text, multi-content with text part, multi-content without text part, and no-mutation of the original.
- `TestDrainAndEmitSteered_MultipleMessages`: 3 plain-text messages → 3 separate session messages, non-last have `\n`, last does not.
- `TestDrainAndEmitSteered_MultiContent`: same for multi-content messages.

### DIFF
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -55,29 +55,39 @@ func (r *LocalRuntime) appendSteerAndEmit(sess *session.Session, sm QueuedMessag
 // fragments from adjacent messages to be glued together. The "\n" prevents
 // this without merging the messages into one.
 //
-// For plain-text messages the "\n" is appended to Content. For multi-content
-// messages it is appended to the last text part's Text field; if no text part
-// exists a new one is appended.
+// It also snapshots the message count before any messages are added and
+// returns it alongside the drained flag so the caller can pass it to
+// compactIfNeeded without a separate len(sess.GetAllMessages()) call.
 //
-// Returns true if any messages were drained and emitted.
-func (r *LocalRuntime) drainAndEmitSteered(ctx context.Context, sess *session.Session, events chan<- Event) bool {
+// NOTE: the appended \n is persisted in the session message and included in
+// UserMessageEvent. This is a deliberate trade-off: because the runtime passes
+// chat.Message slices directly to the provider, this is the only injection
+// point that doesn't require restructuring. TUI consumers may see a trailing
+// newline on non-last steered messages in multi-drain batches.
+//
+// Returns (true, messageCountBefore) if any messages were drained and emitted;
+// (false, 0) otherwise.
+func (r *LocalRuntime) drainAndEmitSteered(ctx context.Context, sess *session.Session, events chan<- Event) (bool, int) {
 	steered := r.steerQueue.Drain(ctx)
 	if len(steered) == 0 {
-		return false
+		return false, 0
 	}
+	messageCountBefore := len(sess.GetAllMessages())
 	for i, sm := range steered {
 		if i < len(steered)-1 {
 			sm = appendNewlineToQueuedMessage(sm)
 		}
 		r.appendSteerAndEmit(sess, sm, events)
 	}
-	return true
+	return true, messageCountBefore
 }
 
-// appendNewlineToQueuedMessage returns a copy of sm with "\n" appended to its
-// text content. For plain-text messages Content is extended. For multi-content
-// messages the last text part is extended; if none exists a new text part is
-// appended.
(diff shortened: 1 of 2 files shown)


pkg/runtime/loop.go
```diff
@@ -47,6 +47,58 @@ func (r *LocalRuntime) appendSteerAndEmit(sess *session.Session, sm QueuedMessag
...
+	// Find the last text part and append \n to it.
+	for i := len(parts) - 1; i >= 0; i-- {
+		if parts[i].Type == chat.MessagePartTypeText {
+			parts[i].Text += "\n"
+			sm.MultiContent = parts
+			return sm
+		}
+	}
+	// No text part found — append a new one.
+	parts = append(parts, chat.MessagePart{Type: chat.MessagePartTypeText, Text: "\n"})
```
simonferquel: This is not useful: 
- pure whitespace parts will be skipped down the line anyway
- non text parts have a layout that makes them immune to the issue (they have an envelope that acts as a separator)

simonferquel-clanker:
Addressed in the latest commit. For the no-text-part branch (e.g. image-only `MultiContent`), `appendNewlineToQueuedMessage` now returns `sm` unchanged — no synthetic whitespace part is appended. Two reasons (per @simonferquel's review): (1) pure whitespace parts are skipped downstream anyway; (2) non-text parts carry their own provider envelope that already acts as a separator, making them immune to run-on tokenisation.

pkg/runtime/loop.go
```diff
@@ -47,6 +47,66 @@ func (r *LocalRuntime) appendSteerAndEmit(sess *session.Session, sm QueuedMessag
...
+// to glue against the next message.
+func appendNewlineToQueuedMessage(sm QueuedMessage) QueuedMessage {
+	if len(sm.MultiContent) == 0 {
+		sm.Content += "\n"
+		return sm
+	}
+	// Shallow-copy the slice so we don't mutate the original.
+	parts := append([]chat.MessagePart(nil), sm.MultiContent...)
+	// Find the last text part and append \n to it.
+	for i := len(parts) - 1; i >= 0; i-- {
```
simonferquel: Why are we iterating here ? Only do append \n on the latest part, and only if it is of text type.

simonferquel-clanker:
Simplified in the latest commit. The loop is gone — `appendNewlineToQueuedMessage` now only looks at `sm.MultiContent[last]`: if it's a text part, append `\n`; if it's any other type, return `sm` unchanged. Non-text parts have their own provider envelope that already acts as a separator, so no `\n` is needed regardless of what precedes them.

trungutt (approved):

PR #2538, opened 2026-04-27
feat(hooks): add 6 builtin hooks + widen post_tool_use / before_llm_call contract

dgageot:
## Summary

Adds **6 new in-process hook builtins** and a small **hook-contract widening** (`post_tool_use` / `before_llm_call` may now signal run termination via `decision: "block"`). The existing inline tool-loop detector is unchanged.

### Five context-injecting builtins (commit 1)

Each follows the same shape as the existing `add_date` / `add_environment_info` / `add_prompt_files`. Opt-in via YAML — none auto-injected, since they shell out / read the filesystem and shouldn't run silently.

| Name | Event | Behavior |
|---|---|---|
| `add_git_status` | `turn_start` | `git status --short --branch` |
| `add_git_diff` | `turn_start` | `git diff --stat` (or full unified with `args: [full]`) |
| `add_directory_listing` | `session_start` | Sorted top-level entries of `Cwd`, hidden skipped, capped at 100 |
| `add_user_info` | `session_start` | `os/user.Current()` + `os.Hostname()` |
| `add_recent_commits` | `session_start` | `git log --oneline -n N` (default 10, override via `args: ["20"]`) |

All git-backed builtins cap output at 4 KiB and treat git-not-installed / non-repo failures as graceful no-ops, so they can never abort a session or turn.

### `max_iterations` builtin (commit 2)

A new stateful `before_llm_call` builtin — hard stop after N model calls, configurable per-agent in YAML:

```yaml
hooks:
  before_llm_call:
    - {type: builtin, command: max_iterations, args: ["50"]}
```

**Additive, not a replacement.** The existing `agent.MaxIterations` flag with its `MaxIterationsReachedEvent` and interactive resume dialog is unchanged — TUI / CLI / ACP all still use it. The new builtin is a hard-stop alternative for headless / cost-control scenarios.

Per-session counter cleared from `session_end` via `builtins.State.ClearSession`.

### Hook contract widening (commit 2)

Documents and enforces that `decision: "block"` (or `continue: false` / exit code 2) from a `post_tool_use` or `before_llm_call` hook signals run termination. Previously these events were observational only. The runtime now translates a deny verdict into the standard `Error` + `notification(level=error)` + `on_error` fanout via a new `emitHookDrivenShutdown` helper.

User-authored hooks can now stop the run; the inline tool-loop detector is unchanged.

### Refactors and fixes (commits 3 + 4)

- `processToolCalls` / `executeWithApproval` / `askUserForConfirmation`: closure plumbing simplified — `LocalRuntime.runTool` now returns a `toolApprovalOutcome` so callers can `return invoke()` everywhere instead of unpacking and repacking a `(stop, msg)` tuple in 7 places.
- `loop_detector` per-call vs per-batch behavior change: **reverted** in commit 5 after review concerns. The inline `tool_loop_detector` is byte-identical to `origin/main`.
- Bug fix: `before_llm_call` was dispatched twice per loop iteration after a rebase; once-per-iteration is now pinned by a regression test (`pkg/runtime/before_llm_call_test.go`).
- `sortKeys` (private, used by builtins) made non-mutating to avoid a foot-gun for future callers.
- `gitOutput`'s empty-dir error message trimmed to Go style (no function name prefix).
- Stale comment in `hooks_wiring_test.go` updated post eager-build refactor.

## Commits

1. `feat(hooks/builtins): add five context-injecting builtins`
2. `feat(hooks): widen contract for post_tool_use / before_llm_call + extract loop_detector & max_iterations builtins`
3. `refactor(hooks/runtime): simplify approval plumbing and trim docstrings`
4. `fix(runtime,hooks): review-driven fixes (duplicate before_llm_call dispatch + 4 quality)`
5. `revert(hooks): drop loop_detector builtin, keep inline tool-loop detector`

(Each commit message has full details — left as separate commits for review legibility; squash-or-merge as you prefer.)

## Validation

- `go build ./...` ✅
- `go test ./...` (97 packages) → all pass
- `go test -race ./pkg/hooks/... ./pkg/runtime/...` → clean
- `go test ./e2e/...` → pass
- `go test ./pkg/tui/... ./pkg/cli/... ./pkg/acp/...` → pass
- `golangci-lint run ./...` → 0 issues
- Full binary builds (`go build .`)

## Behavior changes for existing users

- **None for existing in-repo configs.** Loop detection, max_iterations UX, and all default agent flags behave identically to `origin/main`.
- The `post_tool_use` / `before_llm_call` deny-verdict contract is now honored. If a user happened to have a hook on those events returning `decision: "block"` it would have been silently ignored before; it now actually stops the run. No in-repo configs do this.

## Honest caveats

- All in-repo tests pass (unit + integration + e2e + TUI + CLI + ACP), but I haven't exercised the runtime against a real LLM. The contract widening tests pin executor-level semantics; integration tests cover the runtime fanout.
- The new git-shelling builtins are bounded by the hooks executor's 60 s timeout — pathological repos will time out cleanly rather than hang.

🤖 Generated with [Claude Code](https://claude.com/claude-code) via docker-agent


rumpl (approved):
Yes I read it

PR #2543, opened 2026-04-27
fix(tui): clear bottom slack after thinking text fades out

dgageot:
## Problem

In the TUI, when thinking-text tools fade out of a reasoning block, the viewport is sometimes left mostly empty until the user manually resets the scroll bar.

## Root cause

The `bottomSlack` mechanism adds blank lines to absorb height shrinkage so the UI doesn't jump when tool calls in reasoning blocks appear/disappear. But slack was only ever *consumed* by subsequent growth — there was no decay — and it could grow unbounded. When several tools fade out at once with no follow-up content, slack ended up filling the visible window, leaving an *empty screen*.

## Fix

- **Cap** `bottomSlack` at `min(5, height/3)` so the viewport can never be mostly empty after a large simultaneous shrinkage.
- **Decay** `bottomSlack` by one line per animation tick so leftover empty lines clear quickly after a fade.
- **Self-subscribe** to the animation coordinator while `bottomSlack > 0` so the tick stream keeps firing after the originating fades end.
- Move slack/scroll bookkeeping into `updateScrollState()` and call it from both `View()` and from `Update()` on `animation.TickMsg`, so the new subscription is registered before `tui.go`'s `HasActive()` check and the next tick is correctly scheduled.
- Apply the cap to the public `AdjustBottomSlack` API too, so external callers cannot bypass it.

## Tests

New `pkg/tui/components/messages/slack_test.go` covering:

- Slack is capped on large shrinkage
- Slack decays one line per tick down to 0
- Animation subscription is active while slack > 0 and released when it hits 0
- Viewport always contains real content after a shrinkage
- `maxBottomSlack` scales sensibly with viewport height
- Slack stays at 0 when the user has scrolled away
- Slack decay pauses when the user scrolls away mid-decay
- `AdjustBottomSlack` respects both the cap and the floor

Verified stable across 20 `-race` iterations.

## Commits

1. `fix(tui): clear bottom slack after thinking text fades out` — the fix itself.
2. `refactor(tui): simplify slack subscription handling` — small simplifications (inline a one-shot helper, rename to match the codebase, collapse min/max branches).
3. `fix(tui): address slack-handling review findings` — naming consistency (`onAnimationTick` → `handleAnimationTick`), `AdjustBottomSlack` cap enforcement, extra tests.
4. `test(tui): drop racy tick-cmd assertion from slack subscription test` — a follow-up review-suggested assertion turned out to be racy when run in parallel with other animation-touching tests; the local subscription state assertion is sufficient.

Assisted-By: docker-agent

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -277,7 +277,7 @@ func (m *model) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 	// applied their fade state, and before tui.go's HasActive() check so the
 	// subscription is registered when the next tick is scheduled.
 	if _, ok := msg.(animation.TickMsg); ok {
-		cmds = append(cmds, m.onAnimationTick())
+		cmds = append(cmds, m.handleAnimationTick())
 	}
 
 	return m, tea.Batch(cmds...)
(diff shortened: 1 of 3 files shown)


derekmisler (approved):
🙏🏻

PR #2577, opened 2026-04-28
Add redact_secrets builtin hook + before_llm_call transform

dgageot:
Adds a single agent-level switch — `redact_secrets: true` — that brackets every place where credential material can leak from a docker-agent run to a third party. It scrubs known secret patterns (GitHub PATs, AWS keys, Stripe / Slack / GitLab tokens, JWTs, private keys, Docker Hub PATs, …) at two enforcement points:

1. **Before the LLM call** — a runtime-shipped `MessageTransform` walks every text-bearing field of the outgoing chat history (`Content`, `ReasoningContent`, `MultiContent` text parts, the legacy singular `FunctionCall.Arguments`, and each `ToolCalls[i].Function.Arguments`) and replaces detected secrets with `[REDACTED]`.
2. **Before a tool call** — a `pre_tool_use` builtin hook (`redact_secrets`) walks `ToolInput` recursively (into nested maps and slices) and emits an `UpdatedInput` with only the keys whose value actually changed.

Detection mirrors the MIT-licensed [`docker/mcp-gateway/pkg/secretsscan`](https://github.com/docker/mcp-gateway/tree/main/pkg/secretsscan) ruleset (re-licensed under Apache-2.0 in attribution) and is kept fast by a keyword pre-filter in front of every regex.

## Files

- `pkg/secretsscan/` — new package: ruleset, `ContainsSecrets`, `Redact`, tests, benchmarks.
- `pkg/hooks/builtins/redact_secrets.go` — the `pre_tool_use` builtin and its registration.
- `pkg/runtime/redact_secrets.go` — the `before_llm_call` `MessageTransform` (gated on `Agent.RedactSecrets()`).
- `pkg/agent/{agent.go,opts.go}`, `pkg/teamloader/teamloader.go`, `pkg/config/latest/types.go` — the `redact_secrets: true` agent flag plumbing.
- `agent-schema.json` + `examples/redact_secrets.yaml` — schema entry and a runnable example.

## Try it

\`\`\`yaml
agents:
  root:
    model: openai/gpt-5-mini
    instruction: You are a helpful assistant.
    redact_secrets: true
    toolsets:
      - type: shell
\`\`\`

Now any secret pasted by the user — or generated by a previous turn's tool — is replaced with \`[REDACTED]\` before it reaches the model OR a tool process. The surrounding text is preserved (\`--token=[REDACTED]\` rather than the whole arg disappearing).

## Notes

- **False positive floor**: every rule pairs a regex with a discriminating keyword, so plain English never trips detection. The new \`TestRedactionMarkerIsNotASecret\` locks in that the marker itself can't loop through the ruleset.
- **False negatives are possible**: only patterns the ruleset recognises are scrubbed. This is defense-in-depth, not a substitute for keeping secrets out of the conversation in the first place.
- **No mutation of caller history**: the transform clones \`MultiContent\` / \`ToolCalls\` / \`FunctionCall\` before rewriting, so persisted session messages stay byte-identical to what the user sent.
- **Concurrent-hook safety**: pre_tool_use hooks aggregate via shallow \`maps.Copy\` in config order — the builtin emits only the keys it actually rewrote so it doesn't clobber a sibling hook's modifications.
- **Performance**: \`Redact\` is \`O(rules)\` for clean inputs (just a keyword scan against a single lower-cased copy); 2 allocations on a 9 KB clean message in the benchmark.

## Tests

\`mise lint\` and \`mise test\` are green. The new tests cover detection parity with upstream, redaction span correctness, idempotence, multi-secret handling, marker safety, the pre_tool_use builtin's recursion + "only changed keys" output, agent-flag wiring, and the runtime transform's per-surface scrubbing + non-mutation contract.

Assisted-By: docker-agent

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -268,6 +268,10 @@
           "type": "boolean",
           "description": "Whether to add environment information"
         },
+        "redact_secrets": {
+          "type": "boolean",
+          "description": "When true, the runtime auto-installs the redact_secrets builtin pre_tool_use hook (scrubs detected secrets from tool arguments) AND attaches a before_llm_call message transform that scrubs the same patterns from the messages sent to the LLM. Detection uses the docker-agent secretsscan ruleset (GitHub PATs, AWS keys, Stripe / Slack / GitLab tokens, JWTs, private keys, etc.). Each detected span is replaced with the literal '[REDACTED]'."
+        },
         "max_iterations": {
           "type": "integer",
           "description": "Maximum number of iterations",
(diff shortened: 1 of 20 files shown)


aheritier (approved):
LGTM but missing docs updates

PR #2578, opened 2026-04-28
Suppress spurious 'is now available' MCP toolset notice

dgageot:
## What

After completing an OAuth flow for a remote MCP server, the TUI displayed a notification framed as a warning:

> mcp(remote host=mcp.slack.com transport=streamable) is now available

This is not a valid warning. The OAuth dialog completing already makes a successful start obvious, and for non-OAuth recoveries the model just uses the tool — the follow-up notification is pure noise.

### Commit 1 — Suppress the notice

Drop the entire tool-notice path in the agent and runtime:

- **`pkg/agent/agent.go`**: remove `pendingNotices` field, `AddToolNotice`, `DrainNotices`. `ensureToolSetsAreStarted` still resets the failure-streak flag on a successful start (so the next failure is reported as fresh), but only logs at `slog.Info` — never user-visible.
- **`pkg/runtime/loop.go`**: `emitAgentWarnings` only drains warnings now; `formatToolNotice` is gone.
- Tests updated to assert recoveries are silent.

### Commit 2 — Simplify the resulting state machine

With the recovery notice gone, the only consumer of `ConsumeRecovery()` was a state-reset call in `agent.ensureToolSetsAreStarted`. Folded that reset into `Start()` itself: a successful `Start` now implicitly clears the failure streak.

|                              | Before                                                       | After                                            |
| ---                          | ---                                                          | ---                                              |
| `StartableToolSet` flags     | 3 (`hasEverFailed`, `freshFailure`, `pendingRecovery`)       | 2 (`inFailureStreak`, `pendingWarning`)          |
| Public methods               | `Start`, `Stop`, `ShouldReportFailure`, `ConsumeRecovery`    | `Start`, `Stop`, `ShouldReportFailure`           |
| `ensureToolSetsAreStarted`   | fail-branch + recover-branch (2-arm)                         | one flat try/warn-once loop                      |
| Doc comment                  | 4-bullet state-machine spec                                  | 1 paragraph                                      |
| Net diff                     | —                                                            | **−47 lines**                                    |

## Behavioural guarantees preserved

- **Failure-once-per-streak**: `inFailureStreak` gate prevents re-queueing `pendingWarning` on repeated retries.
- **Post-recovery freshness**: a successful `Start` resets both flags, so the next failure is again surfaced as a fresh warning.
- **Deferred-OAuth must not consume the gate**: `emitToolsProgressively` still checks `IsAuthorizationRequired(err)` *before* `ShouldReportFailure()`, so the eventual real failure (Slack 4xx, etc.) is the one the user sees.

These three properties are pinned by tests in `pkg/tools/startable_test.go`, `pkg/agent/agent_test.go`, and `pkg/runtime/runtime_test.go`.

## Validation

- `mise lint` → `0 issues` (golangci-lint) + `no offenses detected` (in-tree linter, 886 files)
- `mise test` → all packages pass

## Files touched

```
pkg/agent/agent.go
pkg/agent/agent_test.go
pkg/runtime/loop.go
pkg/runtime/runtime.go
pkg/runtime/runtime_test.go
pkg/tools/startable.go
pkg/tools/startable_test.go
```

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -313,30 +313,26 @@ func (a *Agent) ToolSets() []tools.ToolSet {
 	return toolSets
 }
 
+// ensureToolSetsAreStarted starts every toolset, surfacing the first
+// failure of each streak as a user-visible warning and silently retrying
+// on every subsequent turn. A successful Start() automatically resets the
+// streak inside StartableToolSet, so a future failure is again reported
+// as fresh — no recovery callback is needed here, and we deliberately do
+// not surface a "now available" notice (the OAuth dialog completing or
+// the model just using the tool already makes a successful start
+// obvious; a follow-up notification just reads as a spurious warning).
 func (a *Agent) ensureToolSetsAreStarted(ctx context.Context) {
 	for _, toolSet := range a.toolsets {
-		if err := toolSet.Start(ctx); err != nil {
-			// Only warn on the first failure in a streak; suppress duplicate
-			// warnings for subsequent retries that also fail.
-			if toolSet.ShouldReportFailure() {
-				desc := tools.DescribeToolSet(toolSet)
-				slog.Warn("Toolset start failed; will retry on next turn", "agent", a.Name(), "toolset", desc, "error", err)
-				a.AddToolWarning(fmt.Sprintf("%s start failed: %v", desc, err))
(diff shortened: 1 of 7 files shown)


aheritier (approved):
LGTM

PR #2579, opened 2026-04-28
feat(lifecycle): unify MCP/LSP toolset supervision with configurable profiles + /toolsets UX

dgageot:
## Motivation

The MCP / remote MCP / LSP toolset lifecycle has been a recurring source of instability and there's no good user-visible way to configure it. Three separate state machines (`StartableToolSet`, `mcp.Toolset.watchConnection/tryRestart/forceReconnectAndWait`, `lspHandler.startLocked/stopLocked`) each implement a slightly different subset of "connect / watch / restart / stop", with hard-coded knobs (5 attempts, 1s..32s backoff, 35s session-missing timeout) and substring-based error classification. LSP has no auto-restart at all — if `gopls` crashes, the toolset is silently dead until the agent restarts.

This PR replaces those three implementations with one supervisor, makes the lifecycle configurable per-toolset in YAML, and gives users a `/toolsets` view + a `/toolset-restart` command.

## What's in here (11 commits)

1. **`feat(tools/lifecycle): introduce typed errors and state-machine type`** — new `pkg/tools/lifecycle` package with sentinel errors (`ErrTransport`, `ErrServerUnavailable`, `ErrServerCrashed`, `ErrInitTimeout`, `ErrInitNotification`, `ErrCapabilityMissing`, `ErrAuthRequired`, `ErrSessionMissing`, `ErrNotStarted`), a `Classify(err)` helper that wraps an underlying error with the right sentinel, `IsTransient`/`IsPermanent` helpers, and a `State` enum (`Stopped` / `Starting` / `Ready` / `Degraded` / `Restarting` / `Failed`) with a thread-safe `Tracker`.

2. **`feat(tools/lifecycle): extract Supervisor and port mcp.Toolset onto it`** — `lifecycle.Supervisor` driven by a transport-agnostic `Connector` (returns a `Session` with `Wait` / `Close`). Configurable `Policy` (`Restart`, `MaxAttempts`, `Backoff{Initial,Max,Multiplier,Jitter}`, `OnDisconnect`/`OnRestart`/`OnFailed` callbacks). `mcp.Toolset` is rewritten on top of it; `watchConnection` / `tryRestart` / `forceReconnectAndWait` deleted (~150 lines gone).

3. **`feat(tools/builtin): port LSPTool onto lifecycle.Supervisor`** — LSP gets auto-restart for the first time. `lspConnector` + `lspSession` plug into the same supervisor; the previous `startLocked` / `stopLocked` helpers are removed.

4. **`feat(config): add per-toolset lifecycle config with profile presets`** — a `lifecycle:` block on every MCP/LSP toolset in YAML, with three profiles (`strict` / `resilient` / `best-effort`) and per-field overrides:

    ```yaml
    toolsets:
      - type: mcp
        command: ["docker", "mcp", "gateway"]
        lifecycle:
          profile: resilient
          max_restarts: 10
          backoff:
            initial: 500ms
            max: 1m
            multiplier: 2
            jitter: 0.2
    ```

    Schema, validation, and `examples/lifecycle.yaml` are wired up. `required` and `startup_timeout` are documented as **not yet enforced** (a future eager-startup commit will honour them).

5. **`feat(tui): add /toolsets command and supervisor-aware status surface`** — new `tools.Statable` interface and `tools.ToolsetStatus` DTO; runtime + app + TUI plumbing for a `/toolsets` slash command that shows each toolset's name, state, last error, and restart count in a colour-coded dialog.

6. **`feat(tools/builtin): filter LSP tools by server-advertised capabilities`** — `lsp_*` tools are filtered against the server's `capabilities` from `initialize`, so a model never gets `lsp_inlay_hints` against a server that doesn't support it. `SetToolsChangedHandler` is fired post-`Connect` so the runtime refreshes its tool list when the filtered catalogue changes.

7. **`feat: foundation for hot-reload + /toolset-restart slash command`** — `tools.Restartable` capability + runtime `RestartToolset(name)` API + `/toolset-restart <name>` slash command (post-OAuth recovery, "gopls is stuck, please reconnect"). Plus a `pkg/teamloader/diff.go` helper (`ToolsetIdentity`, `ToolsetSignature`, `DiffToolsets`) as the foundation for a future file-watcher-driven reload.

8. **`fix(lifecycle): address reviewer findings on supervisor and LSP session`** — three real concurrency issues uncovered during a self-review pass:
   - `RestartAndWait` could block for the full timeout when the supervisor reached a terminal state mid-call → added a `done` channel signalled on Stop / Failed.
   - `cmd.Wait` was being called twice on the same `*exec.Cmd` (Go forbids this) → shared `sync.Once` + pre-allocated `waitDone` channel.
   - Byte-based truncation of multi-byte UTF-8 in the `/toolsets` dialog → rune-aware `truncateRunes` + regression test with 4-byte emoji.

9. **`fix(lifecycle): self-review fixes for supervisor recovery, LSP race, TUI block, and doc honesty`** — round 2:
   - `done` channel was unrecoverable after `Failed → Start` (sync.Once close meant subsequent `RestartAndWait` returned the stale error). Now mu-guarded close-or-replace; `Start` refreshes it when leaving a terminal state. `forceRestart` flag also reset.
   - `mcp.Toolset.Restart` / `LSPTool.Restart` now route terminal states through `Start` instead of `RestartAndWait` so `/toolset-restart` actually recovers a `Failed` toolset.
   - LSP `ensureInitialized` fast path read `h.cmd != nil` without `h.mu` (data race per the Go memory model) → dropped the cmd nil-check.
   - TUI `handleRestartToolset` was blocking the Bubble Tea update loop for up to 35s → wrapped in a `tea.Cmd`.
   - `required` / `startup_timeout` are documented (and validated) but not yet enforced — schema, godoc, and the example YAML now say so explicitly.

10. **`refactor: simplify lifecycle, MCP, LSP code without changing behaviour`** — pure-refactor pass, **−205 net lines** across 9 files:
    - `classify.go`: replaced custom `classified` error type with `fmt.Errorf("%w: %w", ...)`.
    - `state.go`: `String()` uses a lookup array.
    - `supervisor.go`: `shouldRestart(err, forced)` takes the flag as a parameter; `min`/`max` builtins; Backoff defaults inlined.
    - `lsp_lifecycle.go`: extracted `spawnLSPProcess` / `publishSession` / `clearSession` / `runHandshake` / `fireToolsChanged`. `Connect` reads `spawn → publish → handshake → notify → return session` in ~35 lines.
    - `teamloader/lifecycle.go`: merged identical `strict` and `best-effort` profile branches; dropped redundant Backoff zero values.
    - `runtime.go`: dropped manual `StartableToolSet` unwraps (`tools.As` already walks the wrapper chain).

11. **`fix(lifecycle, lsp): address regressions found during simplification review`** — three real regressions introduced by the simplification pass plus one defence-in-depth guard:
    - `State(-1).String()` panicked because the array-lookup version dropped the `s >= 0` guard. Restored, regression test added.
    - `lspSession.Close` released `h.mu` between the shutdown handshake and the field clearing → window where a per-request method could write to a soon-to-be-closed stdin. Fixed by holding `h.mu` across the entire teardown via a new `clearSessionLocked`.
    - `clearSessionLocked` was nilling `cmd`/`stdin`/`stdout` before clearing the atomic `initialized` flag → `ensureInitialized`'s atomic-only fast path could observe `initialized=true` while fields were already nil. Reordered.
    - `writeMessageLocked` / `readMessageLocked` defensively check `h.stdin`/`h.stdout` for nil, returning `lifecycle.ErrNotStarted` instead of panicking.

## Configuration example

```yaml
agents:
  root:
    toolsets:
      - type: lsp
        command: gopls
        file_types: [".go"]
        lifecycle:
          profile: strict          # required=true (informational), no auto-restart
          startup_timeout: 30s     # informational; not yet enforced

      - type: mcp
        command: ["docker", "mcp", "gateway"]
        lifecycle:
          profile: resilient       # the default, kept here for clarity
          max_restarts: 10
          backoff: { initial: 500ms, max: 1m, multiplier: 2, jitter: 0.2 }

      - type: mcp
        ref: docker:openbnb-airbnb
        lifecycle: { profile: best-effort }   # no auto-restart, optional
```

See `examples/lifecycle.yaml` for the full demonstrable file and `agent-schema.json` for the schema.

## TUI

* `/toolsets` — table of every toolset on the current agent (state, restarts, last error, description).
* `/toolset-restart <name>` — force a supervisor-driven reconnect of the named toolset (post-OAuth recovery; non-blocking; result surfaced via toast).

## Validation

* `golangci-lint run` → 0 issues
* `go run ./lint .` → no offences (899 files)
* `go mod tidy --diff` → clean
* `mise test` → all packages pass, exit 0
* `go test -race` → passes for every package this branch touches: `pkg/tools/lifecycle`, `pkg/tools/mcp`, `pkg/tools/builtin`, `pkg/teamloader`, `pkg/config/latest`, `pkg/tui/dialog`. The pre-existing `pkg/runtime` race in `cache_test.go` / `streaming.go` / `InMemorySessionStore` is unrelated and reproduces at `HEAD~11`.

## Net diff

11 commits, ~3500 LOC added (mostly the supervisor + LSP/MCP rewiring + tests), ~600 LOC removed (the three deleted state machines).

## Follow-ups intentionally NOT in this PR

* Eager-startup phase that actually enforces `required` and `startup_timeout`.
* File-watcher-driven hot reload (the `pkg/teamloader/diff.go` infrastructure is there, the watcher is not).
* Pluggable transports (`stdio` / `streamable-http` / `sse` / `docker` shared between MCP and LSP). Capability filtering for LSP shipped here; the transport extraction was deferred to keep this PR scoped.

Each is small and self-contained on top of what's here.

Assisted-By: docker-agent


aheritier (commented):
needs docos

gtardif (dismissed):

rumpl (approved):

PR #2580, opened 2026-04-28
reduce macOS keychain prompts for OAuth MCP servers

dgageot:
## Problem

When using OAuth-protected MCP servers, macOS prompted the user for keychain access many times — once per server on first launch, again on every HTTP request that needed to read a token, and yet again on every token refresh. Even after clicking *Always Allow*, a rebuild (which changes the binary's signature) brought the prompts back.

The root cause was the storage layout: every token was a separate keychain item, and every keychain item carries its own ACL on macOS.

## Fix

Store all OAuth tokens for all MCP servers in a single bundled keychain item, cached in memory after the first read.

- The OS prompts the user **at most once per process**, regardless of how many MCP servers are configured.
- Token reads after the first one are served from memory and don't touch the keychain at all.
- Token writes (refreshes, new flows) all target the same item the user has already authorised, so a single *Always Allow* keeps applying.

The store is exposed as a process-wide singleton via `sync.OnceValue`, so multiple OAuth-capable toolsets share the same in-memory cache and never reopen the keyring on construction.

## Backwards compatibility

On first load, any tokens written by the previous one-item-per-token layout are migrated into the bundle, and the legacy entries (including the index) are removed. Migration is best-effort: failures are logged but never block startup.

## Public API

Unchanged. `NewKeyringTokenStore`, `OAuthTokenEntry`, `ListOAuthTokens`, and `RemoveOAuthToken` keep the same signatures, so `mcp.go`, `oauth_login.go`, and `cmd/root/debug_oauth.go` work untouched.

## Tests

New tests cover:
- Reads collapsing to a single underlying keyring `Get` regardless of how many resource URLs are looked up.
- All `Set` calls targeting the single bundle item.
- Legacy migration (with the legacy index removed).
- Corrupt-bundle recovery.
- Denied keychain access not snowballing into repeat prompts.
- Concurrent access under the race detector.

## Validation

- `go build ./...` ✓
- `go vet ./...` ✓
- `go test ./...` ✓ (full suite, including `-race`)
- `golangci-lint run ./...` ✓ (0 issues)

## Possible follow-up

Code-sign release builds with a stable Developer ID so macOS recognises the binary across upgrades and never re-prompts after *Always Allow*. Today's dev/ad-hoc-signed builds change identity on every rebuild, which forces another prompt no matter how good the storage layout is. That's outside this PR's scope.

### DIFF
--- a/pkg/tools/mcp/tokenstore_keyring.go
+++ b/pkg/tools/mcp/tokenstore_keyring.go
@@ -5,20 +5,54 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"slices"
 	"strings"
+	"sync"
 
 	"github.com/99designs/keyring"
 )
 
+// keyringServiceName is the macOS service identifier used for our OAuth tokens.
+//
+// All tokens for all MCP servers are stored under a SINGLE keyring item to
+// minimise OS-level credential prompts: on macOS each keychain item carries
+// its own ACL, so storing N items would prompt the user N times the first
+// time each one is read (or after the binary's code-signing identity
+// changes). With a single item the user only has to grant access — and
+// click "Always Allow" — once, no matter how many MCP servers they
+// configure.
 const keyringServiceName = "docker-agent-oauth"
 
-const indexKey = "oauth:_index"
+// bundleKey is the keyring key under which the JSON-encoded
+// resource-URL → OAuthToken map is stored.
+const bundleKey = "oauth:tokens"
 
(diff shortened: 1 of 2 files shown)


aheritier (commented):
LGTM. Tokens will be stored in. a different location after restart but not really worth documenting it

aheritier (approved):
LGTM. Tokens will be stored in. a different location after restart but not really worth documenting it

PR #2590, opened 2026-04-29
Add upgrade-availability hint

### LINKED ISSUE #2467: Add version check and upgrade notification
docker-agent has no mechanism to notify users when they are running an outdated version. The current version is baked in at build time (`pkg/version/version.go`) and displayed in the TUI status bar and `version` subcommand, but there is no comparison against the latest available release.

The only version-related guard today is the DB migration check in `pkg/session/migrations.go`, which errors when the local database was created by a *newer* binary — a backward-compat safeguard, not an update prompt.

dgageot:
Closes #2467.

Surfaces a hint when a newer GitHub release of docker-agent is available, in
both the TUI status bar and the \`version\` subcommand.

## How it works

- A new \`pkg/version/check\` package fetches the latest release tag from
  \`https://api.github.com/repos/docker/docker-agent/releases/latest\` in a
  background goroutine and caches the result on disk for 24h
  (\`<cache-dir>/version-check.json\`).
- The fetch is triggered **only** by \`docker agent run\` / \`--exec\` (in
  \`cmd/root/run.go\`). All other subcommands and the TUI itself read the
  cache via \`LatestCached\` and never touch the network.
- A dedicated \`http.Client\` is used (5s timeout, max 3 redirects) — never
  \`http.DefaultClient\`.
- The cache is written atomically (write-temp + rename) so concurrent
  \`docker agent run\` processes can't corrupt it.
- All errors (offline, rate-limited, parse errors, dev builds, …) are
  swallowed: the user simply does not see a hint.
- A \`dev\` build never reports an upgrade.

## Where the hint shows up

\`docker agent version\`:
\`\`\`
docker agent version v1.0.0
Commit: abc123

A newer version is available: v1.53.0
Release notes: https://github.com/docker/docker-agent/releases/tag/v1.53.0
\`\`\`

TUI status bar (right side of the title):
\`\`\`
docker agent v1.0.0 (update available: v1.53.0)
\`\`\`

## Opt-out

Set \`DOCKER_AGENT_DISABLE_VERSION_CHECK=1\` (or \`true\`/\`yes\`/\`on\`)
to fully disable the feature — both the network call and the cached hint.
The env var is set unconditionally in the runtime stage of the Dockerfile,
since container images are immutable and can't self-upgrade.

The new env var is documented in \`docs/configuration/overview/index.md\`.

## Tests

- Unit tests for semver-tolerant comparison, cache I/O, GitHub fetch
  (via \`httptest\`), redirect-limit enforcement, opt-out env var, and the
  TUI title helper.
- An exported \`check.SeedCacheForTest\` helper keeps the cache JSON layout
  encapsulated.
- \`go build ./...\`, \`go test ./...\`, \`go test -race\`, and
  \`golangci-lint run ./...\` all pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,6 +75,10 @@ RUN apk add --no-cache ca-certificates docker-cli && \
 ARG TARGETOS TARGETARCH
 ENV DOCKER_MCP_IN_CONTAINER=1
 ENV TERM=xterm-256color
+# Disable the once-a-day GitHub release check inside the container image:
+# images are tagged immutably and can't self-upgrade, so the hint would be
+# misleading and waste a network round-trip on every `run`.
+ENV DOCKER_AGENT_DISABLE_VERSION_CHECK=1
 COPY --from=docker/mcp-gateway:v2 /docker-mcp /usr/local/lib/docker/cli-plugins/
 COPY --from=builder-linux /binaries/docker-agent-$TARGETOS-$TARGETARCH /docker-agent
 USER docker-agent
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -21,7 +21,6 @@ import (
 	"github.com/docker/docker-agent/pkg/paths"
 	"github.com/docker/docker-agent/pkg/telemetry"
 	"github.com/docker/docker-agent/pkg/version"
-	"github.com/docker/docker-agent/pkg/version/check"
 )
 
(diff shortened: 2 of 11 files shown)


aheritier (approved):
+1 but doc conflict

dgageot:
Let's close this one for now

PR #2592, opened 2026-04-29
Add turn_end hook

rumpl:
Symmetric counterpart of `turn_start`. Fires once per turn no matter how the turn ended: normal stop, error, hook-blocked, loop-detected, max-iterations, or context cancellation. Exit reason is reported via `hooks.Input.Reason` (`normal`, `continue`, `steered`, `error`, `canceled`, `hook_blocked`, `loop_detected`, `max_iterations`).

Implementation: extract the per-iteration body into `runTurn()` so a single deferred dispatch fires `turn_end` on every exit path. Uses `context.WithoutCancel` for the dispatch so handlers run to completion on Ctrl+C.

Testing: `pkg/runtime/turn_end_test.go` pins the four exit paths (normal, hook-blocked, stream error, context cancellation) plus the per-iteration count when a tool call provokes a follow-on iteration.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -620,6 +620,13 @@
             "$ref": "#/definitions/HookDefinition"
           }
         },
+        "turn_end": {
+          "type": "array",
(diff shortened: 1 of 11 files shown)


dgageot (dismissed):

dgageot:
@rumpl docs update is missing, though

rumpl:
Done in 41f29c9c \u2014 added `turn_end` to the hook types table, the per-event fields table (with the full reason taxonomy), and a new "Turn-End: per-turn finalizer" section right after Turn-Start.

dgageot (approved):

PR #2598, opened 2026-04-30
Support HCL as an alternative agent config format

dgageot:
Adds HCL as an alternative authoring format for agent configuration files,
alongside YAML. HCL documents are converted to YAML at load time, so the
existing schema, version dispatch, and migration chain are reused unchanged.

## What it looks like

```hcl
agent "root" {
  description = "An agent that talks like a pirate"
  instruction = "Always answer by talking like a pirate."
  model       = "auto"

  welcome_message = <<-EOT
  Ahoy! I be yer pirate guide, ready to set sail on the seas o' knowledge!

  What be yer quest? 🏴‍☠️
  EOT
}
```

Conventions:

- Top-level keyed maps (`agents`, `models`, `providers`, `mcps`, `rag`) are
  written as labeled blocks: `agent "root" { … }` ↔ `agents.root: { … }`.
- Inside an agent, `command "name" { … }` ↔ `commands.name: { … }`.
- Toolsets use the label as the `type` field:
  `toolset "mcp" { … }` ↔ `toolsets: [{ type: mcp, … }]`.
- Multi-line strings use heredocs. Because HCL templates expand `\${…}`,
  any literal `\${…}` in instructions must be escaped as `\$\${…}`.

See `examples/pirate.hcl` and `examples/gopher.hcl` — both produce
configurations byte-identical to their `.yaml` siblings (locked in by
`TestHCLExamplesMatchYAML`).

## How it's wired

```
HCL bytes ──hcl.ToYAML──► YAML bytes ──read version──► v0…v8 parser ──migrate──► latest.Config
```

The HCL → YAML step happens once, before anyone looks at the `version` field.
Detection is by `.hcl` extension, with a content-sniff fallback (`LooksLikeHCL`)
for OCI artifacts and other sources that don't have a filename hint.

The converter is **version-agnostic**: it applies stable structural rules
(labeled-block → keyed map, etc.) that target the broad shape every schema
version has shared since v0. Field-level renames between versions are still
absorbed by the existing up-migrators. The `blockRules` table is the union of
HCL conventions across all supported versions; new entries are additive.

## Drift protection

- `TestHCLExamplesMatchYAML` — every `.hcl` example produces a config equal
  to its `.yaml` sibling.
- `TestHCLBlockRulesCoverSchemaMaps` — walks `agent-schema.json` and asserts
  every top-level property of shape `{additionalProperties: …}` has a matching
  `modeMapByLabel` rule registered. Verified to fail with a clear message
  when a new keyed map is added to the schema without an HCL rule.
- `TestJsonSchemaWorksForExamples` — now also validates `.hcl` examples
  (via the converter) against the JSON schema.

## Notable design choices

- **Singular labeled blocks** (`agent "root"`, `model "claude"`) rather than
  `agents { root { … } }`. Idiomatic HCL (cf. Terraform `resource "type"
  "name"`) and lighter to write.
- **Toolset label = type** so multiple `toolset "mcp" { … }` blocks aggregate
  into the YAML `toolsets:` list with `type` injected.
- **`yaml.MapSlice` for label-keyed outputs** so agent declaration order
  is preserved through the YAML-marshal step (`Agents` is a slice in the
  Go config; order matters).
- **No HCL function/variable evaluation.** Expressions are evaluated with an
  empty `hcl.EvalContext{}`, so templates can't reach out to anything.

## Files

| File | What |
| --- | --- |
| `pkg/config/hcl/hcl.go` | HCL → YAML converter |
| `pkg/config/hcl/hcl_test.go` | Unit tests for the converter |
| `pkg/config/config.go` | Detect HCL in `Load`, convert before YAML pipeline |
| `pkg/config/resolve.go`, `pkg/sandbox/args.go` | Recognize `.hcl` extension |
| `pkg/config/examples_test.go` | Cover `.hcl` examples + sibling-equivalence test |
| `pkg/config/schema_test.go` | Schema validation for `.hcl` + drift test |
| `examples/pirate.hcl`, `examples/gopher.hcl` | Drop-in HCL versions of existing examples |

## Validation

`mise lint` and `mise test` are both green. `./bin/docker-agent agent run --dry-run`
loads both example HCL files successfully.


gtardif (dismissed):

rumpl (dismissed):
This is cool

trungutt (approved):

PR #2506, opened 2026-04-25
feat(otel): configure W3C trace propagation for distributed tracing

tdabasinskas:
Two small additions so traces emitted by the existing OTel SDK chain across processes:

1. **Set the global `TextMapPropagator`** to W3C TraceContext + Baggage in `cmd/root/otel.go`. Without this, the SDK records spans locally but never injects `traceparent` on outbound requests.

2. **Wrap the outbound transport in `pkg/httpclient`** with `otelhttp.NewTransport`, gated on `OTEL_EXPORTER_OTLP_ENDPOINT` — the same env var that already gates exporter creation in `initOTelSDK`.

When OTel is disabled — behavior is unchanged. When enabled, each outbound HTTP call emits a CLIENT span and carries `traceparent`, so cagent traces correlate with downstream gateway/proxy.

PR #2620, opened 2026-05-04
feat(otel): instrument runtime with GenAI semantic conventions

### LINKED ISSUE #393: OTEL, again
https://opentelemetry.io/blog/2025/ai-agent-observability/

tdabasinskas:
Adds end-to-end OpenTelemetry instrumentation following the [GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/):

- Provider-layer `chat`/`embeddings`/`rerank` CLIENT spans with `gen_ai.*` attributes and the `gen_ai.client.token.usage` / `operation.duration` histograms.
- Runtime spans (`runtime.session`, `runtime.stream`, `runtime.fallback`, `runtime.tool.call`, `runtime.run_skill`, `runtime.task_transfer`, `runtime.handoff`, `background_agent.run`).
- MCP client + server spans with `params._meta` propagation, plus OAuth flow spans.
- A2A endpoints wrapped with `otelhttp` and marked as `invoke_agent`.
- Hook executor span with verdict/decision/reason annotation; subprocess trace context propagation for hooks, LSP servers, and sandbox `docker exec`.
- Memory, RAG, sessiontitle, evaluation, anthropic-specific spans.
- Built-in tool internals (shell, filesystem, fetch, lsp, codemode, ...) surface their work as span attributes.
- W3C trace context + baggage propagation across all HTTP servers and clients.
- Standard OTel resource attributes (`service.*`, `host.*`, `process.*`, `os.type`)

This PR wires two opt-in env vars beyond the default OTel SDK ones:

- `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` — capture prompts, responses, tool arguments and tool results as span attributes. Off by default (PII surface).
- `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental` — emit only the spec-defined `gen_ai.*` keys. Default is dual-emit (both `gen_ai.*` and the legacy `tool.name` / `agent` / `session.id` keys), so existing dashboards keep working alongside spec-aware tooling.

The diff is large — ~50 files, ~5k lines. It's split into 10 topical commits (telemetry primitives → SDK init → providers → runtime → hooks → MCP → A2A → servers/cold-start → memory/RAG → tool internals) so each commit is independently reviewable. Most of the volume is in the new `pkg/telemetry/genai/` and `pkg/telemetry/mcp/` packages, which are pure helpers; the surface-area changes elsewhere are 1-3 lines per call site.

---

- Fixes #393
- Fixes #2506
- Fixes #2505.


dgageot:
@tdabasinskas not sure why, GitHub doesn't want to merge this one, because of hypothetical merge conflicts. Could you rebase?

tdabasinskas:
> @tdabasinskas not sure why, GitHub doesn't want to merge this one, because of hypothetical merge conflicts. Could you rebase?

Done!

tdabasinskas:
> /review

I don't think that worked 😅

pkg/telemetry/genai/errors.go
```diff
@@ -0,0 +1,85 @@
...
+		// Bare "max_tokens" matches too eagerly: validation errors
+		// like `max_tokens must be > 0` and "model X does not
+		// support max_tokens" both contain the token but are not
+		// context overflows. Stick to the unambiguous phrases.
+		return "context_length_exceeded"
+	case strings.Contains(msg, "rate limit") || strings.Contains(msg, "429"):
+		return "rate_limit"
+	case strings.Contains(msg, "401") || strings.Contains(msg, "unauthorized") || strings.Contains(msg, "authentication"):
+		return "auth"
+	case strings.Contains(msg, "403") || strings.Contains(msg, "forbidden") || strings.Contains(msg, "permission"):
```
aheritier, line 59: Non-blocking: the string-matching here (`"429"`, `"401"`, etc.) will misfire on error messages that happen to contain those substrings for unrelated reasons. Consider adding a typed interface check as a first pass:

```go
if sc, ok := err.(interface{ StatusCode() int }); ok {
    switch sc.StatusCode() {
    case 429: return "rate_limit"
    case 401: return "auth"
    case 403: return "forbidden"
    }
}
```

Providers that expose status codes structurally would get classified correctly without relying on message text.

pkg/telemetry/genai/span.go
```diff
@@ -0,0 +1,418 @@
...
+	if s.firstChunkAt.IsZero() {
+		s.firstChunkAt = now
+	} else {
+		s.chunkDurations = append(s.chunkDurations, now.Sub(s.prevChunkAt).Seconds())
+	}
+	s.prevChunkAt = now
+}
+
+// RecordError marks the span as failed and stores error.type for the
+// duration metric. errType should be a short, low-cardinality string —
```
aheritier, line 270: Non-blocking: `gen_ai.request.model` on the duration histogram could get high-cardinality as model variants proliferate. Worth considering whether to normalize model names (e.g. strip version suffixes) or documenting the expected cardinality budget for operators.

pkg/telemetry/genai/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+var (
+	instOnce sync.Once
+	inst     *instruments
+)
+
+// getInstruments resolves and caches the package-level meter instruments.
+// Histogram creation rarely fails in practice; when one does we keep the
+// instruments that did succeed and leave the failed one nil. Call sites
```
aheritier, line 42: Nit: since the global MeterProvider is resolved once via `sync.Once`, a test that replaces the provider after first use won't pick up the new one. Fine for production, but might surprise test authors — a brief comment here would help.

aheritier (dismissed):
LGTM. Clean design, solid thread safety, good spec adherence. The inline comments are all non-blocking suggestions for follow-up.

pkg/telemetry/genai/span.go
tdabasinskas: Updated the comment. OTel GenAI semconv defines `gen_ai.request.model` as a metric attribute. Stripping or normalizing it deviates from spec. The pragmatic option is scrub at the collector rather than in the agent.

pkg/telemetry/genai/metrics.go
tdabasinskas: Updated the comment.

pkg/telemetry/genai/errors.go
tdabasinskas: Valid point. Updated.

dgageot:
@tdabasinskas can you rebase one more time and I'll review it?

tdabasinskas:
> @tdabasinskas can you rebase one more time and I'll review it?

Done!

aheritier (dismissed):
Re-approving — my prior approval was dismissed by the merge of `upstream/main` into the branch, but there are zero new author code changes since `a4ce95e8`. All three of my previous comments were addressed and the threads are resolved. CI is green on the merge commit.

Original assessment stands: clean design, solid thread safety, good GenAI semconv adherence. LGTM.

dgageot:
@Tabarnakle Could you allow me to push a few changes on top of your commits?

dgageot:
@tdabasinskas also, I wonder how we could reduce the impact of those changes. It's very overwhelming

aheritier:
Agree with you about the size/complexity @dgageot 
My other concern is about how to test it and also validate that information sent are ok in term of PII / RGPD / ... It's maybe ok to grab so many data when you collect them within a corporate environment but not for the telemetry we enable by default (hostname, etc)
https://docker.github.io/docker-agent/community/telemetry/

tdabasinskas:
> @Tabarnakle Could you allow me to push a few changes on top of your commits?

I assume you wanted to tag me here :)

I gave write access to the repo - feel free to push changes.

> @tdabasinskas also, I wonder how we could reduce the impact of those changes. It's very overwhelming

Yes, I know the PR is quite big. From the codebase perspective, there's a lot of impact. 

I did try splitting this into few bigger commits, that should be reviewable independently. I guess I could split the whole PR into smaller PRs if that helps reviewability. Or if you'd prefer to push your own restructuring on top, that works too.

From the user perspective, all of this is gated under `--otel` flag, so I would assume the impact is minimal.

> Agree with you about the size/complexity @dgageot My other concern is about how to test it and also validate that information sent are ok in term of PII / RGPD / ... It's maybe ok to grab so many data when you collect them within a corporate environment but not for the telemetry we enable by default (hostname, etc) https://docker.github.io/docker-agent/community/telemetry/

Most of the things added here are covered by new unit tests. Happy to add an e2e test that runs an agent end-to-end and asserts on the resulting span tree, if you think that would help.

Regarding the PII, correct me if I'm wrong, but I understood that the Docker telemetry (enabled by default) is in no way related to the telemetry exposed by OTel. OTel telemetry requires not only `--otel` flag to be enabled, but also `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable set pointing to working OTel collector. In addition to both of these, capturing of message contents (PII) is disabled by default and can only be enabled by explicit `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` environment variable.

tdabasinskas:
Hi @dgageot, @aheritier,

Do you have any updates on how to move with this further? I see there are already conflicts again - I can resolve them, but since it drifts quite quickly, would be good to know if we planning to merge this.

aheritier:

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25762788591).

tdabasinskas:
> needs to be rebased (at the minimum). Moving it to draft

Rebased.

rumpl:
I'm trying this one out and I see these warnings, anything we can do to fix these?

Screenshot 2026-05-13 at 10 00 42

rumpl:
Note, this doesn't happen all the time, I just started a new session and there were no warnings

tdabasinskas:
> I'm trying this one out and I see these warnings, anything we can do to fix these?
> 
> Screenshot 2026-05-13 at 10 00 42

> Note, this doesn't happen all the time, I just started a new session and there were no warnings

Hi @rumpl,

The warnings seem to be from Jaeger's clock-skew adjuster, not from docker-agent. They fire when Jaeger queries the trace before all of the span batches have flushed (children get there before their parent).

Does re-loading the trace later (after 30s or so) make the warnings disappear? Could you see if enabling more eager flushing via `OTEL_BSP_SCHEDULE_DELAY=1000` (1s instead of 5s) resolves the warnings?

tdabasinskas:
@aheritier, @dgageot, @rumpl,

I rebased this again. Any chance to get this reviewed? Right now it seems we are in a loop, as every time I rebase, it goes stale until there are new conflicts, so it gets marked as requiring rebase again and moved to draft, so I rebase and the cycle repeats.

leonavevor:
I hope this feature will generically support langsmith, langfuse and others when it is released?

tdabasinskas:
Rebased again..

aheritier:
This PR has merge conflicts. Please rebase onto `main` to continue review.

aheritier:
This PR has merge conflicts and has been marked as draft. Please rebase on `main` and remove the `status/needs-rebase` label when resolved.

tdabasinskas:
> This PR has merge conflicts and has been marked as draft. Please rebase on `main` and remove the `status/needs-rebase` label when resolved.

Rebased.. Are there any plans to review the PR?

rumpl (approved):

aheritier:
@dgageot will handle the merge

PR #2623, opened 2026-05-04
feat(runtime): propagate non-interactive mode to child sessions and decline elicitation

tdabasinskas:
Covers the missing parts from #2208:

- Add `NonInteractive` field to `SubSessionConfig` and propagate through `RunAgent` (background tasks), `handleTaskTransfer` (transfers), and `handleRunSkill` (skill execution)
- Add `runtime.WithNonInteractive` option to `LocalRuntime` that auto-declines incoming MCP elicitation requests instead of blocking the handler indefinitely
- Wire `runtime.WithNonInteractive(true)` alongside the existing `session.WithNonInteractive(true)` in the MCP server so headless serve mode is fully deadlock-free on elicitation
- Document the relationship between `runtime.WithNonInteractive` and `session.WithNonInteractive` (runtime flag prevents elicitation hangs; session flag drives auto-stop on max iterations — both should be set together for headless operation)
- Cover the new branch with `TestElicitationHandler_NonInteractive` and `TestElicitationHandler_Interactive_NoChannel`

### DIFF
--- a/pkg/mcp/server.go
+++ b/pkg/mcp/server.go
@@ -170,6 +170,7 @@ func CreateToolHandler(t *team.Team, agentName string) func(context.Context, *mc
 
 		rt, err := runtime.New(t,
 			runtime.WithCurrentAgent(agentName),
+			runtime.WithNonInteractive(true),
 			runtime.WithTracer(otel.Tracer("cagent")),
 		)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


dgageot (dismissed):

dgageot:
@tdabasinskas tests and linters are not happy

tdabasinskas:
> @tdabasinskas tests and linters are not happy

I believe it should be good now. Also addressed the conflicts.

dgageot (approved):

Issue #2595, opened 2026-04-29
Generalized attachment system: pluggable Sources, per-alias capabilities, end-to-end MIME coverage


simonferquel-clanker:
## Problem

Today `pkg/chat.MessagePartTypeFile{Path,...}` hard-codes a local FS path; only images flow end-to-end; PDFs work only via Anthropic's bespoke `FileManager`; **audio/video are silently dropped by every non-Anthropic converter**. Embedders (`pkg/app`, `pkg/cli`) hand-roll resize+base64+data-URL before the runtime sees anything. No storage abstraction, no per-alias capability matrix, no cleanup wiring.

## What we're building

A single `chat.Document` + pluggable `attachment.Source`s on `LocalRuntime` + pure `attachment.Decide` over per-alias capability tables + per-alias `AliasProfile` (caps, prefs, uploader, image constraints). Newly enables PDFs across all PDF-capable aliases, audio on Gemini/OpenAI, video on Anthropic/Gemini/Vertex AI, clean text inline with size-aware promotion. Tenant isolation by construction via per-request runtime in `serve api`.

## Detailed design

https://gist.github.com/simonferquel-clanker/4848e15f66cc0c271f8c7ac1847263b3

## Delivery plan (5 PRs)

- [ ] 1. `pkg/attachment` + `pkg/chat.Document` + mandatory `WithAttachmentSourceResolver` Opt + tests
- [ ] 2. Anthropic refactor; `FileManager` → `memstore.FileIDStore`
- [ ] 3. Gemini, Vertex AI, OpenAI Files-API + capability tables + `ImageConstraints`
- [ ] 4. Bedrock + oaistream aliases (mistral/xai/ollama/nebius/minimax/github-copilot)
- [ ] 5. `LocalRuntime` registry + TUI/CLI plumbing + per-request runtime in `serve api` + deprecate `MessagePartTypeFile`

simonferquel:
Checking if my clanker is wired up so we can ask details in comments. @simonferquel-clanker are you here ?

simonferquel-clanker:
Hey! The orchestrator is live and monitoring this issue — happy to answer any questions about the attachment system design.

rumpl:
What are source capabilties?

simonferquel-clanker:
**What is `SourceCapabilities`?**

It's a small struct passed into the `Decide` function that describes what the *attachment source* (the storage backend) can do — separate from what the *LLM provider* supports:

```go
type SourceCapabilities struct {
[...]

PR #2639, opened 2026-05-05
feat: Phase 1 attachment system – chat.Document, pkg/attachment, per-provider convertDocument

simonferquel-clanker:
## Summary

Implements Phase 1 of the structured attachment system. Part of #2595.

### `pkg/chat/document.go` (new)
- `MessagePartTypeDocument` constant
- `DocumentSource` struct (InlineText / InlineData fields)
- `Document` struct (Name, MimeType, Size, Source)
- `Document *Document` field added to `MessagePart`
- `MessagePartTypeFile` and `MessagePartTypeImageURL` annotated as superseded (not formally deprecated to avoid SA1019 on existing call sites)

### `pkg/attachment/` (new package)
- `attachment.go`: `Strategy` type, `Decide()` routing function, `TXTEnvelope()` helper (with `</document>` escape to prevent envelope breakout)
- `modelcaps/modelcaps.go`: `ModelCapabilities` backed by models.dev data (via `pkg/modelsdev` store); `Supports(mimeType)` gates image/* on vision modality, application/pdf on pdf modality, text/* always allowed, everything else (audio, video, Office binary formats) returns false

### Per-provider `attachments.go` (new files)
Added `convertDocument(ctx, doc, modelID)` and testable `convertDocumentWithCaps(ctx, doc, mc)` to:
- `pkg/model/provider/oaistream` — image → data-URI image part; PDF/other binary → dropped with warning (Chat Completions has no native document block); text → TXTEnvelope; wired into `ConvertMultiContent`
- `pkg/model/provider/openai` — image → `OfInputImage`; PDF → `OfInputFile` (native Responses API file block); text → `OfInputText` with TXTEnvelope
- `pkg/model/provider/anthropic` — image → `ImageBlockParam` (base64); PDF → `DocumentBlockParam` (gated on `application/pdf` only, not `IsAnthropicDocumentMime`); text → `TextBlockParam` with TXTEnvelope
- `pkg/model/provider/gemini` — binary → `genai.Blob`; text → `genai.Text` with TXTEnvelope
- `pkg/model/provider/bedrock` — image → `ContentBlockMemberImage`; PDF → `ContentBlockMemberDocument`; text → `ContentBlockMemberText` with TXTEnvelope

### Signature changes (intentional breaking change)
`ConvertMessages` and `ConvertMultiContent` in `oaistream` now accept `ctx context.Context` and `modelID string` parameters for capability routing and context propagation. All in-tree callers updated.

## Tests
- `pkg/attachment/decide_test.go` — table-driven, all 3 strategy outcomes + MIME-miss + TXTEnvelope escaping
- `pkg/attachment/modelcaps/modelcaps_test.go` — in-memory store, vision/text-only/unknown model/Office-rejected cases
- Per-provider `attachments_test.go` — TXT strategy, Drop strategy, B64 image and PDF success paths

## What's NOT changed
- Existing `MessagePartTypeFile` and `MessagePartTypeImageURL` handling is preserved unchanged
- No config schema changes (Phase 1 is API-only)

Part of #2595

### DIFF
--- a/pkg/attachment/attachment.go
+++ b/pkg/attachment/attachment.go
@@ -7,6 +7,7 @@ package attachment
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/docker/docker-agent/pkg/attachment/modelcaps"
 	"github.com/docker/docker-agent/pkg/chat"
@@ -56,6 +57,10 @@ func Decide(doc chat.Document, mc modelcaps.ModelCapabilities) (Strategy, string
 // parse as a named attachment.
 //
 //	<document name="report.md" mime-type="text/markdown">…body…</document>
+//
+// The body is sanitised to prevent content from breaking out of the envelope:
+// any occurrence of "</document>" in the body is replaced with "&lt;/document&gt;".
 func TXTEnvelope(name, mimeType, body string) string {
-	return fmt.Sprintf("<document name=%q mime-type=%q>%s</document>", name, mimeType, body)
+	safe := strings.ReplaceAll(body, "</document>", "&lt;/document&gt;")
+	return fmt.Sprintf("<document name=%q mime-type=%q>%s</document>", name, mimeType, safe)
 }
--- a/pkg/attachment/attachment.go
+++ b/pkg/attachment/attachment.go
@@ -59,15 +59,3 @@ func Decide(doc chat.Document, mc modelcaps.ModelCapabilities) (Strategy, string
 func TXTEnvelope(name, mimeType, body string) string {
 	return fmt.Sprintf("<document name=%q mime-type=%q>%s</document>", name, mimeType, body)
 }
-
-// Advisor is implemented by provider clients that can report which MIME types
-// their current model supports as document attachments.
-//
-// Providers that do not implement Advisor are assumed to support no attachment
-// types (StrategyTXT text wrapping is always available as a fallback, but
-// callers must opt in via the Decide function).
-type Advisor interface {
-	// SupportedMIMETypes returns the list of MIME types that the provider's
-	// current model accepts as document attachments.  The list may be empty.
-	SupportedMIMETypes() []string
-}
--- a/pkg/attachment/attachment.go
+++ b/pkg/attachment/attachment.go
@@ -8,6 +8,7 @@ package attachment
 import (
 	"fmt"
 	"strings"
+	"unicode"
 
 	"github.com/docker/docker-agent/pkg/attachment/modelcaps"
 	"github.com/docker/docker-agent/pkg/chat"
@@ -53,14 +54,41 @@ func Decide(doc chat.Document, mc modelcaps.ModelCapabilities) (Strategy, string
 	return StrategyDrop, "no inline content"
 }
 
-// TXTEnvelope wraps a text document body in an XML-like tag that models can
-// parse as a named attachment.
+// TXTEnvelope wraps text content in a unique XML-like tag derived from the
+// document name and MIME type. The tag name is a slug of both, making
+// accidental tag break-out in the content practically impossible without
+// escaping the body.
 //
-//	<document name="report.md" mime-type="text/markdown">…body…</document>
+// Example: a document named "report.md" with MIME "text/markdown" produces:
 //
-// The body is sanitised to prevent content from breaking out of the envelope:
-// any occurrence of "</document>" in the body is replaced with "&lt;/document&gt;".
+//	<document-report-md-text-markdown>
+//	…body…
+//	</document-report-md-text-markdown>
 func TXTEnvelope(name, mimeType, body string) string {
-	safe := strings.ReplaceAll(body, "</document>", "&lt;/document&gt;")
-	return fmt.Sprintf("<document name=%q mime-type=%q>%s</document>", name, mimeType, safe)
+	slug := slugify(name + "-" + mimeType)
+	tag := "document-" + slug
+	return fmt.Sprintf("<%s>\n%s\n</%s>", tag, body, tag)
+}
+
+// slugify converts s to a lowercase, alphanumeric-and-hyphens-only string.
+// Non-alphanumeric runes are replaced with hyphens; consecutive hyphens are
+// collapsed to one; leading and trailing hyphens are trimmed.
+// If the result is empty, "doc" is returned as a safe fallback.
+func slugify(s string) string {
+	var b strings.Builder
+	prevHyphen := false
+	for _, r := range strings.ToLower(s) {
+		if unicode.IsLetter(r) || unicode.IsDigit(r) {
+			b.WriteRune(r)
+			prevHyphen = false
+		} else if !prevHyphen {
+			b.WriteRune('-')
+			prevHyphen = true
+		}
+	}
+	result := strings.Trim(b.String(), "-")
+	if result == "" {
+		return "doc"
+	}
+	return result
 }
--- a/pkg/attachment/attachment.go
+++ b/pkg/attachment/attachment.go
@@ -1,0 +1,73 @@
+// Package attachment provides MIME-aware routing for document attachments.
+//
+// It defines how a chat.Document should be sent to a model: either dropped
+// (unsupported), wrapped in a plain-text envelope (StrategyTXT), or encoded
+// as inline base64 data (StrategyB64).
+package attachment
+
+import (
+	"fmt"
+
+	"github.com/docker/docker-agent/pkg/attachment/modelcaps"
+	"github.com/docker/docker-agent/pkg/chat"
+)
+
+// Strategy describes how an attachment should be handled before sending to the
+// provider.
+type Strategy int
+
+const (
+	// StrategyDrop means the attachment is not supported by the model or has no
+	// inline content, and should be silently skipped (with a log warning).
+	StrategyDrop Strategy = iota
+
+	// StrategyTXT means the attachment should be wrapped in a TXTEnvelope and
+	// sent as plain text.  Used for text/* MIME types whose content is already
+	// in Source.InlineText.
+	StrategyTXT
+
+	// StrategyB64 means the attachment content (Source.InlineData) should be
+	// base64-encoded and sent as a native provider image/document block.
+	StrategyB64
+)
+
+// Decide returns the routing Strategy for a document given the current model's
+// capabilities.
+//
+// Algorithm:
+//  1. If the model does not support the document's MIME type → (Drop, reason).
+//  2. If Source.InlineData is non-empty → (B64, "").
+//  3. If Source.InlineText is non-empty → (TXT, "").
+//  4. Otherwise → (Drop, "no inline content").
+func Decide(doc chat.Document, mc modelcaps.ModelCapabilities) (Strategy, string) {
+	if !mc.Supports(doc.MimeType) {
+		return StrategyDrop, fmt.Sprintf("model does not support MIME type %q", doc.MimeType)
+	}
+	if len(doc.Source.InlineData) > 0 {
+		return StrategyB64, ""
+	}
+	if doc.Source.InlineText != "" {
+		return StrategyTXT, ""
+	}
+	return StrategyDrop, "no inline content"
+}
+
+// TXTEnvelope wraps a text document body in an XML-like tag that models can
+// parse as a named attachment.
+//
+//	<document name="report.md" mime-type="text/markdown">…body…</document>
+func TXTEnvelope(name, mimeType, body string) string {
+	return fmt.Sprintf("<document name=%q mime-type=%q>%s</document>", name, mimeType, body)
+}
+
+// Advisor is implemented by provider clients that can report which MIME types
+// their current model supports as document attachments.
+//
+// Providers that do not implement Advisor are assumed to support no attachment
+// types (StrategyTXT text wrapping is always available as a fallback, but
+// callers must opt in via the Decide function).
+type Advisor interface {
+	// SupportedMIMETypes returns the list of MIME types that the provider's
+	// current model accepts as document attachments.  The list may be empty.
+	SupportedMIMETypes() []string
+}
--- a/pkg/attachment/modelcaps/modelcaps.go
+++ b/pkg/attachment/modelcaps/modelcaps.go
@@ -24,11 +24,10 @@ type ModelCapabilities struct {
 	modelFound bool
 }
 
-// isKnownDocMIME returns true for non-text/* MIME types that are universally
-// sendable as TXT envelopes: Office documents and a few other structured formats.
-// Notably, video/*, audio/* and other media types are NOT included — they require
-// explicit model capability declarations.
-func isKnownDocMIME(mt string) bool {
+// isOfficeMIME returns true for Office document binary formats
+// (OOXML, legacy Office, RTF). These are ZIP-based or binary formats
+// that cannot be naively TXT-enveloped and require explicit model support.
+func isOfficeMIME(mt string) bool {
 	switch mt {
 	case "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
 		"application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
@@ -47,11 +46,14 @@ func isKnownDocMIME(mt string) bool {
 // MIME type.
 //
 // Resolution rules (in order):
-//  1. image/* → requires supportsImage
-//  2. application/pdf → requires supportsPDF
-//  3. text/* → always supported (TXT envelope works universally for text)
-//  4. Known Office/document MIMEs → always supported via TXT envelope
-//  5. Everything else (audio/*, video/*, unknown binary) → not supported in Phase 1
+//  1. image/* → requires supportsImage (models.dev "image" modality)
+//  2. application/pdf → requires supportsPDF (models.dev "pdf" modality)
+//  3. text/* → always supported (plain text; TXT envelope is universally safe)
+//  4. Office/binary document MIMEs (DOCX, XLSX, PPTX, etc.) → not supported unless
+//     models.dev explicitly declares a document modality. models.dev currently has
+//     no "document" or "office" modality field, so these return false for all
+//     models until the schema is extended.
+//  5. Everything else (audio/*, video/*, unknown binary) → false
 func (mc ModelCapabilities) Supports(mimeType string) bool {
 	mt := strings.ToLower(mimeType)
 	if strings.HasPrefix(mt, "image/") {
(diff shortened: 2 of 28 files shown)


pkg/attachment/modelcaps/modelcaps.go
```diff
@@ -0,0 +1,109 @@
...
+	mt := strings.ToLower(mimeType)
+	if strings.HasPrefix(mt, "image/") {
+		return mc.supportsImage
+	}
+	if mt == "application/pdf" {
+		return mc.supportsPDF
+	}
+	// All other MIME types are text-based or can be safely wrapped in a TXT
+	// envelope, so we allow them unconditionally.
+	return true
```
simonferquel: we should check the mime type has text/ prefix (nothing prevent the client to send parts with video/* or audio/* mime types)

pkg/attachment/modelcaps/modelcaps.go
```diff
@@ -0,0 +1,109 @@
...
+func Load(modelID string) (ModelCapabilities, error) {
+	store, err := modelsdev.NewStore()
+	if err != nil {
+		// Cannot load the store at all — return text-only conservative caps.
+		return ModelCapabilities{modelFound: false}, nil
+	}
+
+	// Use a background context for the lookup; capability detection is best-effort
+	// and should not block the main request flow.
+	model, err := store.GetModel(context.Background(), modelID)
```
simonferquel: can this actually block indifinitely ? or is the store guaranteed to bring quick results ? should we add a timeout ?

pkg/chat/document.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// InlineData holds binary content (images, PDFs, Office docs, …) that is
+	// base64-encoded when sent to the provider. Used for StrategyB64 attachments.
+	InlineData []byte `json:"inline_data,omitempty"`
+
+	// URL is reserved for Phase 2 (URL-referenced documents). Must not be set
+	// on Documents stored in messages in Phase 1; providers should log a warning
+	// and skip documents that have only URL set.
+	// TODO(phase2): implement URL-referenced document fetching.
+	URL string `json:"url,omitempty"`
```
simonferquel: remove it for now, it is not supported yet

pkg/model/provider/anthropic/attachments.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// convertDocument converts a chat.Document to Anthropic Beta API content blocks.
+//
+// Routing:
+//   - image/* with InlineData → BetaImageBlockParam (base64 source)
+//   - application/pdf with InlineData → BetaRequestDocumentBlock (base64)
+//   - text with InlineText → BetaTextBlockParam with TXTEnvelope
+//   - unsupported / no content → nil (logged as warning)
+func convertDocument(ctx context.Context, doc chat.Document, modelID string) ([]anthropicsdk.BetaContentBlockParamUnion, error) {
```
simonferquel: I think we do not need to use the beta SDK anymore as we are not using File API

pkg/model/provider/anthropic/beta_converter.go
```diff
@@ -272,6 +274,15 @@ func (c *Client) convertBetaUserMultiContent(ctx context.Context, parts []chat.M
 				// File part has neither path nor file ID - this is invalid
 				return nil, errors.New("invalid file attachment: neither path nor file_id provided")
 			}
+
+		case chat.MessagePartTypeDocument:
```
simonferquel, line 278: are nt we missing convert support for non beta ?

simonferquel (changes requested):

pkg/attachment/modelcaps/modelcaps.go
simonferquel-clanker: Fixed in 5dd7103cb. `Supports()` now only allows MIME types that start with `text/` or are in a specific allowlist of known Office/document formats (`application/vnd.openxmlformats-officedocument.*`, `application/msword`, `application/rtf`, etc.). `audio/*`, `video/*`, and arbitrary `application/*` types now return `false` unless the model explicitly declares capability. Added `TestSupports_AudioVideoRejected` to cover this.

pkg/attachment/modelcaps/modelcaps.go
simonferquel-clanker: Fixed in 5dd7103cb. Added a 10-second timeout (`loadTimeout = 10*time.Second`) to both `Load` and `LoadFromStore`. If the fetch times out or the store returns an error, the function falls back to conservative text-only caps and logs a `slog.Warn` with the timeout duration.

pkg/chat/document.go
simonferquel-clanker: Done in 5dd7103cb. Removed the `URL` field from `DocumentSource` entirely.

pkg/model/provider/anthropic/attachments.go
simonferquel-clanker: Fixed in 5dd7103cb. `attachments.go` now uses standard SDK types throughout: `anthropic.ContentBlockParamUnion`, `anthropic.ImageBlockParam`, `anthropic.DocumentBlockParam`, `anthropic.TextBlockParam`. The beta import alias is gone. The beta path (`convertBetaUserMultiContent`) adapts the output via a thin `stdBlocksToBeta` helper in `beta_converter.go`.

pkg/model/provider/anthropic/beta_converter.go
simonferquel-clanker: Fixed in 5dd7103cb. Added `case chat.MessagePartTypeDocument:` to `convertUserMultiContent` in `client.go` (the non-beta path). It calls `convertDocument` which now returns standard SDK types (`[]anthropic.ContentBlockParamUnion`), so they wire in directly. Also changed `_ context.Context` to `ctx context.Context` so the context is forwarded to `convertDocument`.

pkg/attachment/attachment.go
```diff
@@ -0,0 +1,73 @@
...
+// Advisor is implemented by provider clients that can report which MIME types
+// their current model supports as document attachments.
+//
+// Providers that do not implement Advisor are assumed to support no attachment
+// types (StrategyTXT text wrapping is always available as a fallback, but
+// callers must opt in via the Decide function).
+type Advisor interface {
+	// SupportedMIMETypes returns the list of MIME types that the provider's
+	// current model accepts as document attachments.  The list may be empty.
+	SupportedMIMETypes() []string
```
rumpl: This is never called

simonferquel: this is for clients, not for internal use. The idea here is to be able to filter out unsupported files in the UI, but we can skip it I think

rumpl: Let's skip for now

simonferquel-clanker: Fixed in 31e0601e0. Removed the `Advisor` interface from `attachment.go` and all per-provider `SupportedMIMETypes()`/`SupportedMIMETypesForModel()` implementations — none of them had any callers.

pkg/attachment/modelcaps/modelcaps.go
```diff
@@ -0,0 +1,151 @@
...
+		return mc.supportsImage
+	}
+	if mt == "application/pdf" {
+		return mc.supportsPDF
+	}
+	// text/* MIMEs are always supported via TXT envelope.
+	if strings.HasPrefix(mt, "text/") {
+		return true
+	}
+	// Known Office document formats can be safely TXT-enveloped.
```
simonferquel: It is not true. docx etc. are basically zip files

simonferquel: Change that! use models.dev info for understanding if the model has office documents capabilities

simonferquel-clanker: Fixed in f70acb850.

The `isKnownDocMIME` unconditional allowlist is removed. Office MIMEs (DOCX, XLSX, PPTX, etc.) now return `false` from `Supports()` because models.dev only exposes five input modalities (`text`, `image`, `pdf`, `audio`, `video`) — there is no `document` or `office` field in the current schema. These formats will remain unsupported (StrategyDrop) until models.dev adds an explicit capability declaration for them.

The `officeDocumentFormat()` helper in the Bedrock provider is retained as forward-looking code — it will activate once models.dev exposes the relevant modality.

Tests updated: `TestLoadFromStore_OfficeDocsAlwaysAllowed` → `TestLoadFromStore_OfficeDocsNotAllowed`, office doc case in `decide_test.go` now expects `StrategyDrop`.

pkg/model/provider/bedrock/attachments.go
```diff
@@ -0,0 +1,162 @@
...
+						Source: &types.DocumentSourceMemberBytes{
+							Value: doc.Source.InlineData,
+						},
+					},
+				},
+			}, nil
+		}
+
+		// Other Office docs
+		if df, ok := officeDocumentFormat(mime); ok {
```
simonferquel: I think this is never going to apply, as the models.dev caps is filtering those out. simplify

simonferquel (changes requested):

pkg/model/provider/bedrock/attachments.go
simonferquel-clanker: Done in db68b0be6. Removed `officeDocumentFormat()` and the 'Other Office docs' branch entirely. The `StrategyB64` path now only handles images and PDFs — the only binary MIMEs that can reach it since modelcaps gates everything else to `StrategyDrop`. Added a defensive `slog.Warn` + drop for any unexpected binary that somehow slips through.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25480759585).

rumpl (dismissed):

pkg/model/provider/anthropic/attachments.go
```diff
@@ -0,0 +1,91 @@
...
+								MediaType: anthropic.Base64ImageSourceMediaType(mime),
+							},
+						},
+					},
+				},
+			}, nil
+		}
+
+		if IsAnthropicDocumentMime(mime) {
+			// application/pdf → native document block
```
aheritier: **Blocking** — `IsAnthropicDocumentMime` returns true for both `application/pdf` *and* `text/plain` (see `files.go`). A `text/plain` document with `InlineData` (StrategyB64) reaches this branch and gets wrapped in a `Base64PDFSourceParam` with `MediaType: "application/pdf"` — i.e. text bytes shipped as a PDF block. Anthropic will reject this, or worse, accept the mismatch silently.

Either gate this branch on `mime == "application/pdf"` directly, or drop the `text/plain` case from `IsAnthropicDocumentMime`. Same comment applies to the `MediaType: "application/pdf"` literal — it's only correct because of the assumption being violated.

pkg/attachment/attachment.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	return StrategyDrop, "no inline content"
+}
+
+// TXTEnvelope wraps a text document body in an XML-like tag that models can
+// parse as a named attachment.
+//
+//	<document name="report.md" mime-type="text/markdown">…body…</document>
+func TXTEnvelope(name, mimeType, body string) string {
+	return fmt.Sprintf("<document name=%q mime-type=%q>%s</document>", name, mimeType, body)
```
aheritier: **Non-blocking** — `body` is concatenated verbatim. A document containing `</document>` (or a fake opener) can break out of the envelope and inject arbitrary structured content into the model's input — a prompt-injection vector that's particularly easy to exploit because the wrapper itself advertises the structure.

Consider escaping or rejecting `</document>` in `body`, or using a non-trivially-forgeable delimiter (e.g. a random nonce in the tag name).

pkg/model/provider/oaistream/attachments.go
```diff
@@ -0,0 +1,73 @@
...
+		if strings.HasPrefix(mime, "image/") {
+			// Build an OpenAI image part with a data URI.
+			dataURI := fmt.Sprintf("data:%s;base64,%s",
+				doc.MimeType,
+				base64.StdEncoding.EncodeToString(doc.Source.InlineData))
+			return []openai.ChatCompletionContentPartUnionParam{
+				openai.ImageContentPart(openai.ChatCompletionContentPartImageImageURLParam{
+					URL: dataURI,
+				}),
+			}, nil
```
aheritier, line 52: **Question** — for a model whose modalities advertise `pdf`, `Decide()` returns `StrategyB64` and we land here with `application/pdf`. Wrapping the raw base64 payload in a `<document>` text envelope is not something the model can decode — it's just opaque bytes pretending to be text and will burn a lot of tokens for nothing.

Wouldn't it be safer to `StrategyDrop` (with a warning) for non-image binary on this path, since Chat Completions has no native PDF block? Same applies to `openai/attachments.go` for the Responses API — the SDK does have `OfInputFile` for PDFs, which would be the right native block here.

pkg/model/provider/bedrock/attachments.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, fmt.Errorf("unknown attachment strategy %d", strategy)
+	}
+}
+
+// sanitizeDocumentName replaces characters that Bedrock disallows in document names.
+// Bedrock document names must be alphanumeric + hyphens only.
+func sanitizeDocumentName(name string) string {
+	var sb strings.Builder
+	for _, r := range name {
+		if (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '-' {
```
aheritier: **Nit** — the allowlist drops `.`, which is common in real filenames (`report.pdf` → `report-pdf`). Bedrock's `DocumentBlock.name` rules permit alphanumeric, whitespace, hyphens, parentheses and square brackets — none of which include `.`, so the strip is technically correct, but the resulting name is less recognisable to the model.

Consider stripping the extension (`filepath.Ext`) before sanitising so `report.pdf` becomes `report` rather than `report-pdf`.

aheritier (changes requested):
Solid Phase 1 scaffolding — clean separation between `Decide`/`modelcaps`/per-provider `convertDocument`, good test coverage on the routing layer.

One blocking issue and a few smaller items:

- **Blocking**: `anthropic/attachments.go` will ship `text/plain` bytes as an `application/pdf` block because `IsAnthropicDocumentMime` covers both.
- **Non-blocking**: `TXTEnvelope` doesn't escape the body — `</document>` in content lets attachments break out of the wrapper.
- **Non-blocking**: oaistream/openai fall back to base64-in-text-envelope for non-image binaries; for PDF this is essentially garbage tokens. Either drop, or use `OfInputFile` on the Responses API.
- **Non-blocking**: per-attachment `modelcaps.Load` cold-path timeout multiplies with N attachments.
- **Nit**: `sanitizeDocumentName` mangles extensions (`report.pdf` → `report-pdf`).

Also: the PR description still advertises `ConvertMessagesLegacy` / `ConvertMultiContentLegacy` backward-compat aliases — they don't exist in the diff. Either add them or remove the claim from the description so reviewers know `ConvertMessages`/`ConvertMultiContent` are intentionally breaking changes.

pkg/model/provider/anthropic/attachments.go
simonferquel-clanker: Fixed in 792a4fed. The `IsAnthropicDocumentMime` condition is replaced with an explicit `mime == "application/pdf"` check. `text/plain` (and any other non-PDF binary) now falls through to a defensive warn+drop — it can never carry `InlineData` in practice since modelcaps returns `StrategyTXT` for text/* via `InlineText`.

pkg/attachment/attachment.go
simonferquel-clanker: Fixed in 792a4fed. `TXTEnvelope` now replaces `</document>` with `&lt;/document&gt;` in the body before wrapping. Added `TestTXTEnvelope_EscapesClosingTag` to verify the sanitisation.

pkg/model/provider/bedrock/attachments.go
simonferquel-clanker: Fixed in 792a4fed. `sanitizeDocumentName` now strips the extension via `path.Ext` before sanitising, so `report.pdf` → `report`. Spaces, parentheses, and square brackets are preserved (Bedrock permits them per docs). Added `TestSanitizeDocumentName` table test.

pkg/model/provider/oaistream/attachments.go
simonferquel-clanker: Fixed in 792a4fed. The Chat Completions endpoint has no native PDF block. Instead of wrapping opaque base64 bytes in a TXT envelope (wasteful and meaningless), `application/pdf` in `StrategyB64` now logs a `slog.Warn` and returns `nil` (drops the attachment). For the Responses API (`openai/attachments.go`), PDFs now use the native `OfInputFile` block with `FileData` set to a `data:application/pdf;base64,…` URI.

pkg/attachment/attachment.go
```diff
@@ -56,6 +57,10 @@ func Decide(doc chat.Document, mc modelcaps.ModelCapabilities) (Strategy, string
 // parse as a named attachment.
 //
 //	<document name="report.md" mime-type="text/markdown">…body…</document>
+//
+// The body is sanitised to prevent content from breaking out of the envelope:
+// any occurrence of "</document>" in the body is replaced with "&lt;/document&gt;".
 func TXTEnvelope(name, mimeType, body string) string {
-	return fmt.Sprintf("<document name=%q mime-type=%q>%s</document>", name, mimeType, body)
+	safe := strings.ReplaceAll(body, "</document>", "&lt;/document&gt;")
```
simonferquel: I do not like this approach. i would rather generate mangle the enveloppe tag name.
eg: `<document-name-slugged-mimetype-slugged> </document-name-slugged-mimetype-slugged>` to make it very unlikely it conflicts

pkg/model/provider/openai/attachments.go
```diff
@@ -52,14 +52,25 @@ func convertDocumentToResponseInputWithCaps(ctx context.Context, doc chat.Docume
...
-		slog.DebugContext(ctx, "openai responses: no native block for MIME, falling back to TXT envelope",
+
+		if mime == "application/pdf" {
+			// The Responses API has a native file input block; use it for PDFs.
+			b64Data := base64.StdEncoding.EncodeToString(doc.Source.InlineData)
+			dataURI := fmt.Sprintf("data:%s;base64,%s", doc.MimeType, b64Data)
+			return []responses.ResponseInputContentUnionParam{
+				{
+					OfInputFile: &responses.ResponseInputFileParam{
+						FileData: param.NewOpt(dataURI),
```
simonferquel: if FileData supposed to be the dataURI or to be the base64 payload directly?

pkg/attachment/attachment.go
simonferquel-clanker: Implemented in be25d7dd. `TXTEnvelope` now generates a unique slug-based tag from the document name and MIME type:

```
report.md + text/markdown → <document-report-md-text-markdown>…</document-report-md-text-markdown>
```

`slugify()` lowercases the string, replaces non-alphanumeric runes with hyphens, collapses consecutive hyphens, and trims leading/trailing ones. Replaced `TestTXTEnvelope_EscapesClosingTag` with `TestTXTEnvelope_UniqueTag` which verifies open/close tag symmetry and that different name+MIME pairs produce different tags.

pkg/model/provider/openai/attachments.go
simonferquel-clanker: Investigated in be25d7dd. The SDK `ResponseInputFileContentParam.FileData` godoc says: **"The base64-encoded data of the file to be sent to the model."** — raw base64, not a data URI.

The previous implementation was wrong (it was passing `data:application/pdf;base64,…`). Fixed: now passes `base64.StdEncoding.EncodeToString(doc.Source.InlineData)` directly.

simonferquel (approved):

aheritier (approved):
All four prior blocking/non-blocking points addressed:

- **Anthropic PDF gate**: `mime == "application/pdf"` strict check — `text/plain` no longer slips into `DocumentBlockParam`.
- **TXT envelope breakout**: tag is now slugged from name + MIME (`<document-report-md-text-markdown>…`) — content can't forge a closing tag without knowing the slug. Bonus: matches @simonferquel's suggested approach.
- **oaistream PDFs**: dropped with a clear warning + comment explaining Chat Completions has no native document block.
- **Bedrock filename**: `path.Ext` stripped before sanitisation, so `report.pdf` → `report` instead of `report-pdf`.

Code reads cleanly, tests cover the routing matrix, CI is green. LGTM.

PR #2641, opened 2026-05-05
Fix finish_reason stop when tracking usage

rumpl:
The server now correctly sends usage at the end of the stream the same way openai's API does.

### DIFF
--- a/pkg/chatserver/agent.go
+++ b/pkg/chatserver/agent.go
@@ -247,13 +247,8 @@ func runAgentLoop(ctx context.Context, rt runtime.Runtime, sess *session.Session
 	return errors.Join(runErrs...)
 }
 
-// sessionUsage extracts approximate token usage from a completed session,
-// returning nil when nothing is known so we can omit the field entirely
-// rather than reporting zeroes.
+// sessionUsage extracts approximate token usage from a completed session
 func sessionUsage(sess *session.Session) *ChatCompletionUsage {
-	if sess.InputTokens == 0 && sess.OutputTokens == 0 {
-		return nil
-	}
 	return &ChatCompletionUsage{
 		PromptTokens:     sess.InputTokens,
 		CompletionTokens: sess.OutputTokens,
--- a/pkg/chatserver/server.go
+++ b/pkg/chatserver/server.go
@@ -395,7 +395,7 @@ func (s *server) handleChatCompletions(c echo.Context) error {
 	}
 
 	if req.Stream {
-		err := s.streamChatCompletion(c, rt, sess, model)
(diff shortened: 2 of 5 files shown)


rumpl:
With this change docker-agent can connect to a docker-agent that serves an agent

Screenshot 2026-05-05 at 22 19 33

pkg/chatserver/server.go
```diff
@@ -501,11 +501,36 @@ func (s *server) streamChatCompletion(c echo.Context, rt runtime.Runtime, sess *
 		stream.send(ChatCompletionStreamDelta{}, "error")
 	} else {
 		stream.send(ChatCompletionStreamDelta{}, "stop")
+		if includeUsage {
+			stream.sendUsage(sessionUsage(sess))
```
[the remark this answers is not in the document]
rumpl: fixed

pkg/chatserver/server.go
```diff
@@ -501,11 +501,36 @@ func (s *server) streamChatCompletion(c echo.Context, rt runtime.Runtime, sess *
 		stream.send(ChatCompletionStreamDelta{}, "error")
```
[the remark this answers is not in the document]
rumpl: fixed

pkg/chatserver/agent.go
```diff
@@ -251,9 +251,6 @@ func runAgentLoop(ctx context.Context, rt runtime.Runtime, sess *session.Session
 // returning nil when nothing is known so we can omit the field entirely
```
krissetto: so was it intentional, or should we remove/adjust this part of the comment?

pkg/chatserver/server.go
```diff
@@ -395,7 +395,7 @@ func (s *server) handleChatCompletions(c echo.Context) error {
 	}
 
 	if req.Stream {
-		err := s.streamChatCompletion(c, rt, sess, model)
+		err := s.streamChatCompletion(c, rt, sess, model, req.StreamOptions != nil && req.StreamOptions.IncludeUsage)
```
krissetto: i gotta say she's a beauty :) can we maybe make a var? it's not entirely sure what it is at a glance

krissetto (dismissed):

pkg/chatserver/agent.go
rumpl: I'll remove the comment

pkg/chatserver/server.go
rumpl: I'll just change it so that StreamOptions is not a pointer

dgageot (approved):

PR #2646, opened 2026-05-05
fix compaction kept-tail mapping after prior summaries

rumpl:
## Problem

When a session already contained a summary, the kept-tail boundary returned by `mapToSessionIndex` was wrong. It counted non-system items from `sess.Messages[0]`, but the filtered chat-message list it was indexing into came from `sess.GetMessages(a)`, which:

1. **Prepends a synthetic `"Session Summary: ..."` user message** that doesn't exist in `sess.Messages` at all.
2. **Starts iterating items at the prior summary's `FirstKeptEntry`** — skipping every item before that offset.

Both effects shift the count, so the new `FirstKeptEntry` landed far too early in the session — typically inside the conversation that the prior summary already replaced. Subsequent compactions could then resurrect already-summarized history.

## Fix

Replace `nonSystemMessages` + `mapToSessionIndex` with a single `gatherCompactionInput` that walks `sess.Messages` directly (mirroring `session.buildSessionSummaryMessages`), surfaces the synthetic Session Summary message when a prior summary exists, and returns a parallel slice of origin indices.

`firstKeptSessionIndex` maps a split index back to a `sess.Messages` index via that parallel slice — correct by construction regardless of prior summaries. The synthetic message maps back to the prior summary's index, so when the kept-tail boundary lands on it, the prior summary item is preserved across the new compaction.

`ComputeFirstKeptEntry` no longer needs the `*agent.Agent` parameter; updated its single call site in `session_compaction.go`.

## Tests

Added regression coverage in `pkg/runtime/compactor/compactor_test.go`:

- `TestGatherCompactionInput_NoPriorSummary` — replaces the old `TestMapToSessionIndex`, pinning the no-summary path.
- `TestGatherCompactionInput_WithPriorSummary` — the regression test: with a summary at index 10 (`FirstKeptEntry=8`), a split keeping the last 2 messages now correctly maps to `items[13]`, where the old code returned `items[5]`.
- `TestGatherCompactionInput_PriorSummaryWithoutFirstKeptEntry` — covers the "compact everything" prior-summary case where iteration must start strictly after the summary item.

### DIFF
--- a/pkg/runtime/compactor/compactor.go
+++ b/pkg/runtime/compactor/compactor.go
@@ -153,24 +153,35 @@ func RunLLM(ctx context.Context, args LLMArgs) (*Result, error) {
 // [maxKeepTokens] window. Used by the runtime when a hook supplies
 // its own summary so the kept-tail policy stays consistent across
 // the two strategies.
-func ComputeFirstKeptEntry(sess *session.Session, a *agent.Agent) int {
-	return mapToSessionIndex(sess, compaction.SplitIndexForKeep(nonSystemMessages(sess, a), maxKeepTokens))
+func ComputeFirstKeptEntry(sess *session.Session) int {
+	messages, sessIndices := gatherCompactionInput(sess)
+	return firstKeptSessionIndex(sess, sessIndices, compaction.SplitIndexForKeep(messages, maxKeepTokens))
 }
 
-// nonSystemMessages returns the agent-visible messages in sess with
-// the system entries filtered out. Both the LLM strategy (via
-// [extractMessages]) and the hook-supplied path (via
-// [ComputeFirstKeptEntry]) operate on this same shape, which is also
-// what [compaction.SplitIndexForKeep] expects.
-func nonSystemMessages(sess *session.Session, a *agent.Agent) []chat.Message {
-	var messages []chat.Message
-	for _, msg := range sess.GetMessages(a) {
-		if msg.Role == chat.MessageRoleSystem {
-			continue
-		}
-		messages = append(messages, msg)
+// gatherCompactionInput is a thin wrapper around
+// [session.Session.CompactionInput] that clears compaction-specific
+// fields on the returned chat messages.
+//
+// Cost is per-message bookkeeping already accumulated into
+// sess.TotalCost(); leaving it set would double-count when the
+// summarization session reports its own TotalCost back through
+// [Result.Cost]. CacheControl pins a provider cache checkpoint
+// (Anthropic prompt caching, etc.); pinning it inside the
+// summarization sub-call would associate the cache point with the
+// throwaway compaction conversation rather than the parent session.
+//
+// The reconstruction work — surfacing a synthetic "Session Summary"
+// message when a prior summary exists, picking the right start index
+// past the prior summary, and tracking origin indices in sess.Messages
+// — lives on Session itself so it can run under sess.mu.RLock and stay
+// race-safe against concurrent AddMessage / ApplyCompaction calls.
+func gatherCompactionInput(sess *session.Session) ([]chat.Message, []int) {
+	messages, sessIndices := sess.CompactionInput()
+	for i := range messages {
+		messages[i].Cost = 0
+		messages[i].CacheControl = false
 	}
-	return messages
+	return messages, sessIndices
 }
 
 // extractMessages returns the messages to send to the compaction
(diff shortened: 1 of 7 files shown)


pkg/runtime/compactor/compactor.go
```diff
@@ -153,26 +153,93 @@ func RunLLM(ctx context.Context, args LLMArgs) (*Result, error) {
...
+				CreatedAt: time.Now().Format(time.RFC3339),
+			},
+			sessionIndex: syntheticSessionIndex,
+		})
+	}
+
+	startIndex := lastSummaryIndex + 1
+	if lastSummaryIndex >= 0 {
+		kept := items[lastSummaryIndex].FirstKeptEntry
+		if kept > 0 && kept < lastSummaryIndex {
```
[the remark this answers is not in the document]
rumpl: Thanks, good catch. Addressed in 56b7b6ea: `FirstKeptEntry == 0` is now treated as a valid raw session index in both reconstruction paths (`session.buildSessionSummaryMessages` and `compactor.nonSystemMessageRefs`), and `-1` is used as the no-kept-tail sentinel. I also updated the SQLite default/load fallback to `-1`, made compaction return `-1` when no real message is kept, preserved `FirstKeptEntry` when branching sessions, and added tests for the zero-index and no-kept-tail cases.

derekmisler (dismissed):

pkg/session/migrations.go
```diff
@@ -398,7 +398,7 @@ func getAllMigrations() []Migration {
 			ID:          21,
 			Name:        "021_add_first_kept_entry_column",
 			Description: "Add first_kept_entry column to session_items for compaction-preserved messages",
-			UpSQL:       `ALTER TABLE session_items ADD COLUMN first_kept_entry INTEGER DEFAULT 0`,
+			UpSQL:       `ALTER TABLE session_items ADD COLUMN first_kept_entry INTEGER DEFAULT -1`,
```
aheritier: while it looks to be ok in this context it's always dangerous to update an existing migration. A new migration 22 would be cheap: `UPDATE session_items SET
first_kept_entry = -1 WHERE item_type = 'summary' AND first_kept_entry = 0`

aheritier (commented):
test - Fighting with my GitHub API Rate limit issue

pkg/session/session.go
```diff
@@ -354,6 +354,14 @@ func (s *Session) snapshotItems() []Item {
 	return items
 }
 
+// SnapshotItems returns a copy of s.Messages safe to use without holding
+// s.mu. Each Message value is deep-copied so concurrent UpdateMessage calls
```
aheritier: Exposing `snapshotItems()` as public to serve the compactor creates direct cross-package coupling — the compactor now reaches into session internals rather than asking the session to produce what it needs. A narrower alternative would be a `Session` method (e.g. `BuildCompactionRefs()`) that returns the already-reconstructed message list with raw indexes, keeping the shape the compactor actually needs without leaking the full item slice. Fine to defer to a follow-up, but worth noting since we are growing the public API surface.

pkg/runtime/compactor/compactor.go
```diff
@@ -153,26 +153,93 @@ func RunLLM(ctx context.Context, args LLMArgs) (*Result, error) {
...
-	for _, msg := range sess.GetMessages(a) {
-		if msg.Role == chat.MessageRoleSystem {
+// nonSystemMessageRefs reconstructs the compactable conversation as the latest
+// summary (if any) followed by the messages visible after that summary, while
+// retaining provenance back to sess.Messages for real messages. The provenance
+// is critical after a prior compaction: the previous summary is synthetic, so
+// indexes in the reconstructed message list no longer line up with raw session
+// indexes.
+func nonSystemMessageRefs(sess *session.Session) []messageRef {
+	items := sess.SnapshotItems()
```
aheritier: The old path went through `sess.GetMessages(a)` which applied agent-specific transformations (e.g. injecting the system prompt, flattening sub-sessions, etc.). `SnapshotItems()` bypasses all of that. For compaction purposes filtering by role is sufficient — but it would be worth a quick audit confirming there are no agent-level transformations in `GetMessages` the compactor was implicitly relying on (e.g. content rewriting or implicit messages). If there are none, a brief inline comment here noting the intentional bypass would make future readers more confident this is deliberate.

pkg/session/migrations.go
rumpl: ooof this is no good

rumpl:
@aheritier Completely changed the implementation, please take a look

aheritier (approved):
Solid fix. The parallel `sessIndices` approach is correct by construction — no way to miscount when you track origin indices alongside the message list rather than reconstructing the mapping afterwards.

The `CompactionInput` method is a faithful mirror of `buildSessionSummaryMessages`, the race test gives confidence under concurrent access, and the regression test pins the exact scenario that was broken (split index 5 → `items[13]`, not `items[5]`).

One minor note for a follow-up: `extractMessages` still carries `_ *agent.Agent` now that it's unused — not blocking, just a small cleanup opportunity.

PR #2658, opened 2026-05-06
feat: Anthropic Workload Identity Federation

### LINKED ISSUE #2640: [anthropic] Workload Identity Federation support
**What you'd like to see**

Authenticate workloads to the Claude API with short-lived identity tokens from companies identity provider instead of long-lived static API keys.

**Why you'd like to see it**

It reduces security issues which are existing with long-lived static API key 

**Workarounds?**

Use long-lived static API key

**Additional context**

[Any other info you consider useful can be included here](https://platform.claude.com/docs/en/build-with-claude/workload-identity-federation)

dgageot:
Closes #2640.

## Summary

Adds support for **Anthropic Workload Identity Federation (WIF)** so that
agents can authenticate Claude API requests with short-lived OIDC-derived
tokens instead of a long-lived `ANTHROPIC_API_KEY`.

The feature is configured with a typed `auth:` block on a `ProviderConfig`
or `ModelConfig` (model-level wins, otherwise inherited from the provider):

```yaml
providers:
  anthropic-wif:
    provider: anthropic
    auth:
      type: workload_identity_federation
      workload_identity_federation:
        federation_rule_id: fdrl_REPLACE_ME
        organization_id: 00000000-0000-0000-0000-000000000000
        # service_account_id: svac_...   # optional, target_type=SERVICE_ACCOUNT only
        identity_token:
          # Pick exactly one of: file, env, command, url
          file: /var/run/secrets/anthropic.com/token

models:
  claude:
    provider: anthropic-wif
    model: claude-sonnet-4-5
```

`identity_token` supports four mutually exclusive sources, covering the common
runtimes:

| Source    | When to use |
|-----------|-------------|
| `file`    | Kubernetes projected SA tokens, SPIFFE/SPIRE helpers, Vault sidecars — anything that rotates a file on disk |
| `env`     | The token is already exported in an environment variable |
| `command` | Shell out on every refresh (`gcloud auth print-identity-token`, `az account get-access-token`, …) |
| `url`     | Fetch from an HTTP(S) endpoint (cloud metadata servers, GitHub Actions OIDC token endpoint, …) |

`url` supports `${VAR}` expansion in both the URL and any header values, so
the GitHub Actions OIDC endpoint can be wired without baking secrets into
YAML:

```yaml
identity_token:
  url: ${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=https://api.anthropic.com
  headers:
    Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}
  response_field: value
```

A complete walkthrough of all four sources lives in
[`examples/anthropic_wif.yaml`](examples/anthropic_wif.yaml).

## How it works

The official `anthropic-sdk-go` (v1.39+) already exposes
`option.WithFederationTokenProvider(IdentityTokenFunc, FederationOptions)`,
so no token-exchange / RFC 7523 wire format had to be implemented locally.

- `pkg/config/latest/auth.go` — typed `AuthConfig` / `FederationAuthConfig`
  / `IdentityTokenSourceConfig` with strict validation (prefix checks,
  exactly-one-source). Plus `EffectiveAuth`, `EffectiveProviderType`, and
  `(*AuthConfig).EnvVars()` helpers, reused by gather/availability/runtime.
- `pkg/model/provider/anthropic/federation/` — turns the typed config into
  `option.RequestOption`s. Wraps token-source errors with a clear
  `anthropic workload identity federation: failed to refresh identity
  token from <kind> source (federation_rule=fdrl_…): <cause>` message that
  flows through the existing runtime `ErrorEvent` path, so refresh failures
  appear in the TUI immediately.
- `pkg/model/provider/anthropic/client.go` — single `buildDirectAuthOptions`
  helper picks between WIF and the legacy `ANTHROPIC_API_KEY` path; rejects
  combining `auth:` with `--gateway`.
- `pkg/config/gather.go`, `pkg/runtime/model_switcher.go` — auto-detection /
  required-env / availability checks understand WIF: a model with WIF auth
  no longer requires `ANTHROPIC_API_KEY`, and the env vars referenced by
  the chosen identity-token source (env name, `${VAR}` references in URL /
  headers) are surfaced instead.
- `pkg/model/provider/defaults.go` — propagates provider-level `Auth` to
  models that don't override it.
- `agent-schema.json` — JSON Schema for the new types, with `oneOf` for the
  four token sources and `^fdrl_` / `^svac_` regex constraints.
- `docs/providers/anthropic/index.md` — new "Workload Identity Federation"
  section.

## Defensive behaviour worth calling out

The `url` source ships with three pieces of belt-and-braces hardening
(against hostile or misconfigured token endpoints):

- **No redirect following.** Go strips `Authorization`/`Cookie` on
  cross-origin redirects but **not** arbitrary user-defined headers, so a
  redirect from the configured endpoint could leak a header secret (e.g.
  `X-OIDC-Token: ${SECRET}`) to an attacker-controlled host. Redirects
  surface as non-2xx errors.
- **Bounded response body** (1 MiB). JWTs are well under 16 KiB; the cap
  prevents OOM from a hostile endpoint streaming an unbounded body.
- **30s request timeout** layered on top of the SDK-supplied context.

`command` source has an analogous 30s timeout. UTF-8-aware truncation is
used for non-2xx error snippets so we never produce invalid UTF-8 in error
messages.

## Tests

- `pkg/config/latest/auth_test.go` — full validation matrix (incl.
  custom-provider indirection both positive and negative cases).
- `pkg/model/provider/anthropic/federation/federation_test.go` — file/env/
  command/url sources, `${VAR}` expansion, JSON-field extraction, redirect
  refusal, body-size cap, UTF-8-safe error truncation.
- `pkg/model/provider/anthropic/auth_test.go` — `NewClient` covers WIF
  success, broken config, unknown auth type, and the gateway-mutex.
- `pkg/config/gather_auth_test.go` — env-var detection across all source
  kinds and provider inheritance.
- `pkg/runtime/model_switcher_test.go` — availability surfaces anthropic
  for both model- and provider-level WIF, and stays unavailable for plain
  configs without `ANTHROPIC_API_KEY`.
- `pkg/model/provider/model_defaults_test.go` — `Auth` inheritance from
  `ProviderConfig` with model-override precedence.

`mise dev` (lint + test + build) green.

## Out of scope

- Vertex AI client (`pkg/model/provider/anthropic/vertex.go`) is left
  unchanged — it has its own auth flow.
- Custom Anthropic-compatible gateways behind `provider: anthropic` +
  `base_url:` may or may not understand `Bearer sk-ant-oat01-…` tokens;
  WIF is intended for `api.anthropic.com`.

## Commits

1. `feat: add Anthropic Workload Identity Federation support`
2. `refactor: simplify Anthropic WIF wiring`
3. `fix: tighten Anthropic WIF validation and harden urlSource`

### DIFF
--- a/pkg/config/latest/auth.go
+++ b/pkg/config/latest/auth.go
@@ -3,6 +3,9 @@ package latest
 import (
 	"errors"
 	"fmt"
+	"maps"
+	"os"
+	"slices"
 	"strings"
 )
 
@@ -28,6 +31,28 @@ type AuthConfig struct {
 const (
 	AuthTypeWorkloadIdentityFederation = "workload_identity_federation"
 )
+
+// EffectiveAuth returns the auth that applies to a model: the model's own
+// Auth wins, otherwise the referenced ProviderConfig's Auth.
+func EffectiveAuth(m ModelConfig, providers map[string]ProviderConfig) *AuthConfig {
+	if m.Auth != nil {
+		return m.Auth
+	}
+	if p, ok := providers[m.Provider]; ok {
+		return p.Auth
+	}
+	return nil
+}
+
+// EffectiveProviderType returns the underlying provider type for a model,
+// resolving custom-provider indirection (a model whose `provider:` points
+// to an entry in `providers:` inherits that entry's `provider:` field).
+func EffectiveProviderType(m ModelConfig, providers map[string]ProviderConfig) string {
+	if p, ok := providers[m.Provider]; ok && p.Provider != "" {
+		return p.Provider
+	}
+	return m.Provider
+}
 
 // FederationAuthConfig describes an Anthropic OIDC Federation Rule and the
 // source of the JWT identity token to be exchanged for a short-lived access
@@ -93,10 +118,46 @@ type IdentityTokenSourceConfig struct {
 	ResponseField string `json:"response_field,omitempty" yaml:"response_field,omitempty"`
 }
 
-// validate validates an AuthConfig, including a provider-type-aware check
-// that the chosen scheme is supported by providerType. providerType may be
-// empty when the auth lives on a ProviderConfig that does not declare an
-// underlying provider; in that case the provider-specific check is skipped.
+// EnvVars returns the environment variables that the auth configuration
+// references at runtime. Today this is only meaningful for Workload Identity
+// Federation, whose identity-token source may either name an env var
+// directly (env source) or reference one through ${VAR} expansion in URL
+// or header values.
+func (a *AuthConfig) EnvVars() []string {
+	if a == nil || a.Type != AuthTypeWorkloadIdentityFederation || a.Federation == nil {
+		return nil
+	}
+	src := a.Federation.IdentityToken
+	if src == nil {
+		return nil
+	}
+	seen := map[string]bool{}
+	collect := func(s string) {
+		os.Expand(s, func(name string) string {
+			if name != "" {
+				seen[name] = true
+			}
+			return ""
+		})
+	}
+	if src.Env != "" {
+		seen[src.Env] = true
+	}
+	if src.URL != "" {
+		collect(src.URL)
+	}
+	for _, v := range src.Headers {
+		collect(v)
+	}
+	return slices.Sorted(maps.Keys(seen))
+}
+
+// validate validates an AuthConfig. providerType, when non-empty, is used
+// to enforce that the chosen scheme is supported by the underlying
+// provider (today: WIF requires "anthropic"). Empty providerType skips
+// that check, which is what we want when an [AuthConfig] sits on a
+// [ProviderConfig] that doesn't declare an underlying provider — the
+// per-model check picks it up later.
 func (a *AuthConfig) validate(providerType string) error {
 	if a == nil {
 		return nil
@@ -105,14 +166,13 @@ func (a *AuthConfig) validate(providerType string) error {
 	case "":
 		return errors.New("auth.type is required")
 	case AuthTypeWorkloadIdentityFederation:
-		// WIF is currently only meaningful for the Anthropic provider.
-		// We allow an empty providerType so a ProviderConfig that only
-		// sets `auth:` (and inherits `provider:` from a model) still
-		// passes validation; the per-model check kicks in later.
 		if providerType != "" && providerType != "anthropic" {
 			return fmt.Errorf("auth.type %q is only supported with the anthropic provider (got %q)", a.Type, providerType)
 		}
-		return a.Federation.validate()
+		if err := a.Federation.validate(); err != nil {
+			return fmt.Errorf("auth: %w", err)
+		}
+		return nil
 	default:
 		return fmt.Errorf("unsupported auth.type %q", a.Type)
 	}
@@ -123,45 +183,32 @@ func (f *FederationAuthConfig) validate() error {
 		return errors.New("workload_identity_federation block is required when auth.type is workload_identity_federation")
 	}
 	if f.FederationRuleID == "" {
-		return errors.New("workload_identity_federation.federation_rule_id is required")
+		return errors.New("federation_rule_id is required")
 	}
 	if !strings.HasPrefix(f.FederationRuleID, "fdrl_") {
-		return fmt.Errorf("workload_identity_federation.federation_rule_id must start with %q (got %q)", "fdrl_", f.FederationRuleID)
+		return fmt.Errorf("federation_rule_id must start with %q (got %q)", "fdrl_", f.FederationRuleID)
 	}
 	if f.OrganizationID == "" {
-		return errors.New("workload_identity_federation.organization_id is required")
+		return errors.New("organization_id is required")
 	}
 	if f.ServiceAccountID != "" && !strings.HasPrefix(f.ServiceAccountID, "svac_") {
-		return fmt.Errorf("workload_identity_federation.service_account_id must start with %q when set (got %q)", "svac_", f.ServiceAccountID)
+		return fmt.Errorf("service_account_id must start with %q when set (got %q)", "svac_", f.ServiceAccountID)
 	}
 	if f.IdentityToken == nil {
-		return errors.New("workload_identity_federation.identity_token is required")
+		return errors.New("identity_token is required")
 	}
 	return f.IdentityToken.validate()
 }
 
 func (s *IdentityTokenSourceConfig) validate() error {
-	if s == nil {
-		return errors.New("identity_token is required")
-	}
-	set := 0
-	if s.File != "" {
-		set++
-	}
-	if s.Env != "" {
-		set++
-	}
-	if len(s.Command) > 0 {
-		set++
-	}
-	if s.URL != "" {
-		set++
-	}
-	if set == 0 {
+	sources := s.setSources()
+	switch len(sources) {
+	case 0:
 		return errors.New("identity_token requires exactly one of: file, env, command, url")
-	}
-	if set > 1 {
-		return errors.New("identity_token must set exactly one of: file, env, command, url")
+	case 1:
+		// ok
+	default:
+		return fmt.Errorf("identity_token must set exactly one of file, env, command, url (got %s)", strings.Join(sources, ", "))
 	}
 	// Headers / response_field are only meaningful with url:
 	if s.URL == "" {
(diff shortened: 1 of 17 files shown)


rumpl:
conflicts

aheritier (commented):
## Review Summary

Well-designed feature with clean layering (config types → validation → token sources → SDK integration), comprehensive tests, and good security posture. The redirect-blocking, body-size cap, and timeout in `urlSource` are solid defensive measures.

**Blockers**: Merge conflicts (as noted by @rumpl). Needs a rebase.

**Non-blocking questions** (beyond what the bot already raised):

1. **Config version migration** — `Auth` is added directly to `latest.ProviderConfig` and `latest.ModelConfig`. Does this need a config version bump and migration step in the `v0→v7→latest` chain, or is additive-only considered safe here?

2. **HCL support** — Given the recent HCL config format, does the `auth` block need HCL-specific parsing/test coverage, or does the generic struct-tag deserialization handle it?

pkg/model/provider/anthropic/federation/federation.go
```diff
@@ -0,0 +1,274 @@
...
+// error on failure. A hard timeout is applied on top of the caller's
+// context so a wedged subprocess can't stall token refresh forever.
+func commandSource(argv []string) option.IdentityTokenFunc {
+	return func(ctx context.Context) (string, error) {
+		if len(argv) == 0 {
+			return "", errors.New("identity_token.command is empty")
+		}
+		ctx, cancel := context.WithTimeout(ctx, tokenCommandTimeout)
+		defer cancel()
+		cmd := exec.CommandContext(ctx, argv[0], argv[1:]...) //nolint:gosec // user-provided per config; intentional
```
aheritier, line 163: nit: The child process inherits the full parent environment (no `cmd.Env` override). This is correct — commands like `gcloud` and `az` need their ambient credentials — but worth a one-line doc comment for security-conscious users auditing this path. Something like:

```go
// The subprocess inherits the parent environment so cloud CLIs
// (gcloud, az, aws) can locate their own credentials.
```

pkg/model/provider/anthropic/federation/federation.go
```diff
@@ -0,0 +1,274 @@
...
+	// tokenCommandTimeout does the same job for commandSource.
+	tokenCommandTimeout = 30 * time.Second
+)
+
+// noRedirectClient is the HTTP client used by urlSource. We disable redirect
+// following because Go only strips Authorization/Cookie/Www-Authenticate on
+// cross-origin redirects; a custom header value carrying a secret (e.g.
+// X-OIDC-Token) would still be re-sent to the new host. Identity-token
+// endpoints (GitHub Actions, IMDS, GCP/Azure metadata, ...) don't redirect.
+var noRedirectClient = &http.Client{
```
aheritier, line 54: Consider setting `Transport` with `MaxIdleConnsPerHost` or a reasonable `IdleConnTimeout`. Since this is a package-level singleton reused across all WIF-configured models, long-running agents with multiple providers could accumulate idle connections. Not urgent (token refresh is infrequent), but worth a TODO.

pkg/config/latest/auth.go
```diff
@@ -0,0 +1,249 @@
...
+	if a == nil || a.Type != AuthTypeWorkloadIdentityFederation || a.Federation == nil {
+		return nil
+	}
+	src := a.Federation.IdentityToken
+	if src == nil {
+		return nil
+	}
+	seen := map[string]bool{}
+	collect := func(s string) {
+		os.Expand(s, func(name string) string {
```
aheritier, line 140: Clever use of `os.Expand` as a variable-name scanner without actually expanding values. One subtlety worth a comment: `os.Expand` recognizes both `$VAR` and `${VAR}` forms, and for bare `$` followed by non-identifier chars it returns the literal — so a URL like `https://example.com/path?a=1&b=2` (no $ at all) won't produce false positives. Good.

Minor: this doesn't catch the `$$` (literal dollar) escape, but that shouldn't matter for real URLs.

pkg/runtime/model_switcher.go
```diff
@@ -486,6 +486,17 @@ func (r *LocalRuntime) getAvailableProviders(ctx context.Context) map[string]boo
...
 
+	// Mark anthropic available when any model or referenced provider in the
+	// workspace has a non-API-key auth scheme configured (e.g. Workload
+	// Identity Federation). The actual token exchange happens lazily on the
+	// first request, so we just need the model picker to surface the model.
+	for _, m := range r.modelSwitcherCfg.Models {
+		if modelHasAnthropicAuth(m, r.modelSwitcherCfg.Providers) {
+			available["anthropic"] = true
+			break
+		}
```
aheritier, line 500: Nice addition. One edge case to consider: if a model has `auth:` configured but the identity token source is broken (missing file, unset env var, etc.), the provider will be surfaced as "available" but the first actual request will fail. This seems like the right trade-off (fail lazily on first use rather than eagerly probing sources at startup), but worth documenting the intent.

dgageot:
Rebased on top of `origin/main` and addressed the review feedback. New commit on top: `3256dba56 fix: address PR review feedback on Anthropic WIF`.

### Blocking issues
- **Nil dereference of `cfg.Auth.Federation`** (Issue 1): `buildDirectAuthOptions` now rejects `auth.workload_identity_federation == nil` explicitly with a clear error before touching `FederationRuleID`. YAML configs were already validated, but a programmatic caller could trigger the panic; tested via the existing validation suite.
- **Expanded URL leaked in errors** (Issue 2): `urlSource` error messages now reference `rawURL` (the unexpanded template) so secret values injected via `${VAR}` substitutions in URL/query don't flow into logs or TUI `ErrorEvent`s. New `TestURLSource_DoesNotLeakExpandedURL` covers this.
- **Silent 1 MiB body truncation** (Issue 3): `urlSource` now reads one byte past `maxTokenResponseBytes` and surfaces `"response exceeded N bytes; check identity_token.url endpoint"` instead of silently passing a truncated body to the JSON parser. `TestURLSource_LimitsResponseBody` updated.

### Nits
- Added a comment on `commandSource` explaining why it inherits the parent process env (gcloud / az CLI rely on ambient credentials).
- Added a comment on the `os.Expand` "scanner" trick in `AuthConfig.EnvVars`.
- Expanded the comment in `getAvailableProviders` to make the lazy/eager trade-off explicit (IMDS hangs on non-cloud hosts; gcloud/az can take seconds).
- Switched `slog.Debug` → `slog.DebugContext` in `buildDirectAuthOptions` for consistency with the rest of `client.go`.

I left the HTTP transport tuning (`MaxIdleConns`, `IdleConnTimeout`, ...) alone for now: token fetches happen at most once per ~hour per process and reuse Go's default pool, so a custom transport would just add surface area without measurable benefit. Happy to add one if you'd prefer.

### Non-blocking questions
1. **Config version migration?** No migration needed. `Auth` is purely additive on `latest.{ProviderConfig,ModelConfig}`; older configs (v0..v7) are upgraded via `types.CloneThroughJSON` and the field stays `nil` (which is what `EffectiveAuth` already handles). Same pattern every prior additive change in `latest` followed.
2. **HCL support?** Works out of the box. `pkg/config/hcl` is schema-agnostic — `auth { ... }`, `workload_identity_federation { ... }`, and `identity_token { ... }` all fall through to the singleton-block default and round-trip into the same YAML the loader expects. No HCL-specific changes required.

### Validation
- `task test` — pass (full `go test ./...`).
- `golangci-lint run pkg/model/provider/anthropic/... pkg/config/latest/` — `0 issues`. (The `task lint` job in this checkout reports issues from an unrelated worktree on disk; none are in files this PR touches.)

aheritier (approved):
LGTM. The latest commit (`68ca678`) addresses every medium-severity concern from prior reviews:

- ✅ Nil dereference on `cfg.Auth.Federation` — explicit rejection added in `buildDirectAuthOptions`
- ✅ Secret leak via expanded URL in error messages — error paths now reference `rawURL` (template) instead of `expandedURL`
- ✅ Silent body truncation at 1 MiB — now reads `+1` byte to detect overflow and returns an explicit, actionable error
- ✅ `commandSource` env inheritance — documented as intentional (gcloud/az ambient creds)
- ✅ Lazy availability probe in `model_switcher` — documented trade-off
- ✅ `os.Expand` scanner trick in `EnvVars` — documented including the `$$` non-handling

Defensive design (no-redirect HTTP client, body cap, dual timeouts, UTF-8-safe truncation, secret-safe error messages, strict prefix validation, exactly-one-source enforcement) is comprehensive. CI all green.

Re: my earlier non-blocking question about config version — `latest`-only is consistent with AGENTS.md guidance ("only add new features to latest, older versions are frozen"), so no migration needed.

Two remaining non-blocking items can be follow-ups:
- `noRedirectClient` would benefit from a `Transport` with `IdleConnTimeout` (singleton; long-running agents only)
- HCL config parsing coverage for `auth:` block is unverified — worth confirming once

PR #2660, opened 2026-05-06
feat: unload local models when switching agents

### LINKED ISSUE #2636: Unload Model from Memory on Agent Switch for Local Providers
# Feature Request: Unload Model from Memory on Agent Switch for Local Providers

## Summary

When using local model providers (such as `dmr` or `ollama`) with multi-agent configurations, switching between agents that use different models can cause significant memory pressure because the previously loaded model is not evicted from the inference engine's memory. This feature request proposes a mechanism to automatically unload the current model when switching to a different agent that uses a different model.

## Problem

Docker Agent supports multi-agent architectures where different agents can be configured with different models. When using local providers like `dmr` (Docker Model Runner), each model is loaded into GPU/CPU memory when first used. If two agents use different models, both models end up resident in memory simultaneously — even though only one is active at any given time.

On machines with limited VRAM or RAM (common for local inference setups), this leads to:

- **Memory pressure**: two or more large models competing for the same memory pool.
- **Degraded performance**: thrashing, swapping, or OOM conditions.
- **Unnecessary resource usage**: the idle model occupies memory that the active model could use.

DMR already exposes an API endpoint to explicitly unload a model from memory:

```
POST /engines/unload
```

This endpoint exists precisely to allow consumers to free memory when a model is no longer needed, but docker-agent currently has no way to trigger it automatically when switching between agents.

### New `provider_opts` key: `unload_on_switch`

Add a boolean `provider_opts` key that tells docker-agent to call the engine's unload API before loading a different model.

```yaml
models:
  local-large:
    provider: dmr
    model: ai/qwen3:14B
    provider_opts:
      unload_on_switch: true   # unload this model from memory when switching away

  local-small:
    provider: dmr
    model: ai/qwen3:0.6B
    provider_opts:
      unload_on_switch: true
```

When `unload_on_switch: true` is set on a model, docker-agent will call the engine's unload endpoint before activating a different model on the same provider.

### New `provider` key: `unload_api`

To make the mechanism generic and reusable with other local-inference engines (e.g. Ollama), expose the unload endpoint as a configurable field on the provider configuration:

```yaml
providers:
  my-local-runner:
    provider: dmr
    base_url: http://localhost:12434/engines/llama.cpp/v1
    unload_api: /engines/unload   # POST to {base_url_root}{unload_api} to unload a model

  my-ollama:
    provider: ollama
    base_url: http://localhost:11434/v1
    unload_api: /api/delete       # Ollama's equivalent endpoint
```

When `unload_api` is set on a provider, and the active agent switches to an agent using a different model on the same provider, docker-agent will issue a `POST` (or `DELETE`, configurable) request to that endpoint before initializing the new model.

### Interaction with `keep_alive`

DMR already supports `keep_alive` via `provider_opts` (values like `"0"` to unload immediately, `"-1"` to keep forever). The `unload_on_switch` option is complementary: `keep_alive` controls TTL-based eviction, while `unload_on_switch` triggers an explicit eviction at agent-switch time regardless of the TTL.

Setting `keep_alive: "0"` currently unloads the model after *each request*, which is too aggressive for normal use. `unload_on_switch` would provide a middle ground: keep the model loaded during a session with a given agent, but release it when the runtime moves to a different agent with a different model.

dgageot:
Closes #2636.

## What

Adds an opt-in mechanism for local inference engines (DMR, ollama, ramalama, ...) to release a model from memory when the active agent switches to one that uses a different model.

## User-facing surface

Two new pieces of config (latest only — older versions are frozen):

- **\`provider_opts.unload_on_switch: true\`** on a model — the runtime calls the engine's unload endpoint when switching away from any agent using this model.
- **\`unload_api: <path>\`** on a \`ProviderConfig\` — endpoint the runtime hits with \`POST {scheme://host}{unload_api}\` and body \`{\"model\": \"<id>\"}\`.

For DMR specifically, the default \`/_unload\` endpoint is auto-derived from the OpenAI base URL (mirroring how \`_configure\` is derived), so DMR users don't need to set \`unload_api\` unless they want to override it.

\`\`\`yaml
providers:
  local_dmr:
    provider: dmr
    base_url: http://model-runner.docker.internal/engines/llama.cpp/v1
    unload_api: /engines/_unload   # optional for DMR; defaults to /_unload

models:
  qwen:
    provider: local_dmr
    model: ai/qwen3
    provider_opts:
      unload_on_switch: true
\`\`\`

A runnable example lives at \`examples/unload_on_switch.yaml\`.

## Wiring

The unload runs at every agent-switch entry point:

- \`swapCurrentAgent\` (transfer_task forward + return)
- \`handleHandoff\` (handoff)
- \`SetCurrentAgent\` (TUI agent picker — async, see below)

Best-effort: each call is wrapped in a 10s timeout, providers that don't implement \`Unloader\` are silently skipped, and any error is logged but never propagated, so a slow or unreachable engine cannot break agent switching.

## Architecture

- \`latest.ProviderConfig.UnloadAPI\` — new field; merged into \`ModelConfig.ProviderOpts[\"unload_api\"]\` so provider implementations don't need a back-reference to the parent config.
- \`latest.ModelConfig.UnloadOnSwitch()\` / \`UnloadAPI()\` — small accessor methods next to the existing \`DisplayOrModel()\`.
- \`provider.Unloader\` — optional interface in \`pkg/model/provider/provider.go\` (sibling of \`RerankingProvider\`).
- \`base.PostUnloadModel\` / \`base.JoinHostAndPath\` — shared HTTP + URL helpers in \`pkg/model/provider/base\`. DMR and OpenAI both already import \`base\`, so no new package and no import cycle.
- \`dmr.Client.Unload\` — DMR-specific URL derivation (\`/v1\` → \`/_unload\`) plus the shared HTTP helper.
- \`openai.Client.Unload\` — no-op when no \`unload_api\` is configured (cloud providers don't have one).
- \`runtime.LocalRuntime.unloadOnSwitch\` — iterates the previous agent's configured models and calls \`Unload\` on opted-in ones.

## TUI freeze fix

\`SetCurrentAgent\` is called from the bubbletea Update loop, so a synchronous unload would freeze the TUI for up to 10s. The runtime spawns the unload in a goroutine for that path only — fire-and-forget is safe because the new agent's model isn't loaded until the user sends a message, by which time the unload almost certainly finished. The other two switch paths (\`swapCurrentAgent\`, \`handleHandoff\`) stay synchronous because there a new model IS loaded immediately and we want unload to complete first.

## Tests

- \`pkg/config/latest/unload_test.go\` — \`UnloadOnSwitch\` / \`UnloadAPI\` accessors.
- \`pkg/model/provider/base/unload_test.go\` — URL resolution + HTTP behaviour (happy path, non-2xx, nil client fallback).
- \`pkg/model/provider/dmr/unload_test.go\` — default \`_unload\` derivation, custom path override, error propagation, no-op when nothing is configured.
- \`pkg/model/provider/openai/unload_test.go\` — no-op without \`unload_api\`, happy path, error path.
- \`pkg/model/provider/provider_defaults_test.go\` — three new cases for the \`UnloadAPI\` plumbing.
- \`pkg/runtime/unload_test.go\` — opt-in respected, no-op when not opted in or same agent or nil prev, errors don't propagate, providers without \`Unloader\` are skipped, and \`TestSetCurrentAgent_UnloadIsAsync\` (uses a blocking unloader) asserts the picker returns before \`Unload\` completes. Runs cleanly under \`-race\`.

## Validation

- \`mise build\` ✓
- \`mise lint\` ✓ (\`golangci-lint\` and the in-tree \`./lint\` cop both clean)
- \`mise test\` ✓

### DIFF
--- a/pkg/model/provider/openai/unload.go
+++ b/pkg/model/provider/openai/unload.go
@@ -1,0 +1,28 @@
+package openai
+
+import (
+	"context"
+
+	"github.com/docker/docker-agent/pkg/httpclient"
+	"github.com/docker/docker-agent/pkg/model/provider/base"
+)
+
+// Unload asks the upstream provider to release the resources held for the
+// configured model. It is a no-op when the provider config does not declare
+// an `unload_api`, which is the case for cloud providers (OpenAI, Anthropic
+// gateways, ...) that don't expose such an endpoint.
+//
+// Local OpenAI-compatible inference engines (ollama, ramalama, vLLM, ...)
+// typically do, and the runtime triggers Unload when switching away from an
+// agent whose model has `provider_opts.unload_on_switch: true`.
+func (c *Client) Unload(ctx context.Context) error {
+	path := c.ModelConfig.UnloadAPI()
+	if path == "" {
+		return nil
+	}
+	endpoint, err := base.JoinHostAndPath(c.ModelConfig.BaseURL, path)
+	if err != nil {
+		return err
+	}
+	return base.PostUnloadModel(ctx, httpclient.NewHTTPClient(ctx), endpoint, c.ModelConfig.Model)
+}
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -530,7 +530,18 @@ func (r *LocalRuntime) CurrentAgentInfo(context.Context) CurrentAgentInfo {
 }
 
 func (r *LocalRuntime) SetCurrentAgent(agentName string) error {
-	return r.agents.SetValidated(agentName)
+	prev := r.CurrentAgent()
+	if err := r.agents.SetValidated(agentName); err != nil {
+		return err
+	}
+	next := r.CurrentAgent()
+	if prev != next {
+		// User-driven switches don't carry a request context, but Unload
+		// must be bounded so a slow engine doesn't stall the picker.
+		// unloadOnSwitch already wraps each call in a per-call timeout.
+		r.unloadOnSwitch(context.Background(), prev, next)
+	}
+	return nil
 }
 
 func (r *LocalRuntime) CurrentAgentCommands(context.Context) types.Commands {
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -535,12 +535,16 @@ func (r *LocalRuntime) SetCurrentAgent(agentName string) error {
 		return err
 	}
 	next := r.CurrentAgent()
-	if prev != next {
-		// User-driven switches don't carry a request context, but Unload
-		// must be bounded so a slow engine doesn't stall the picker.
-		// unloadOnSwitch already wraps each call in a per-call timeout.
-		r.unloadOnSwitch(context.Background(), prev, next)
+	if prev == next {
+		return nil
 	}
+	// Run the unload in the background: the TUI agent picker drives this
+	// path from the bubbletea Update loop, so a synchronous call would
+	// freeze the UI for as long as the engine takes to acknowledge the
+	// unload (up to unloadOnSwitchTimeout). Fire-and-forget is safe here
+	// because the new agent's model isn't loaded until the user sends a
+	// message — typically long after the unload completes.
+	go r.unloadOnSwitch(context.Background(), prev, next)
 	return nil
 }
 
--- a/pkg/runtime/unload.go
+++ b/pkg/runtime/unload.go
@@ -1,0 +1,46 @@
+package runtime
+
+import (
+	"context"
+	"log/slog"
+	"time"
+
+	"github.com/docker/docker-agent/pkg/agent"
+	"github.com/docker/docker-agent/pkg/model/provider"
+)
+
+// unloadOnSwitchTimeout caps each Unload() call. The runtime invokes
+// Unload during agent switches; a stalled or unreachable engine must not
+// block the user, so we time out aggressively and keep going.
+const unloadOnSwitchTimeout = 10 * time.Second
+
+// unloadOnSwitch asks the previous agent's models to release any
+// resources they hold, when they have explicitly opted in via
+// `provider_opts.unload_on_switch: true`. It is best-effort: providers
+// that don't implement [provider.Unloader] are silently skipped, and any
+// error from Unload is logged but never propagated, so a slow or
+// unreachable engine cannot break agent switching.
+//
+// Every configured model is considered (not just the currently-selected
+// one) so alloyed agents free every variant they may have loaded.
+func (r *LocalRuntime) unloadOnSwitch(ctx context.Context, prev, next *agent.Agent) {
+	if prev == nil || prev == next {
+		return
+	}
+	for _, m := range prev.ConfiguredModels() {
+		cfg := m.BaseConfig().ModelConfig
+		if !cfg.UnloadOnSwitch() {
+			continue
+		}
+		unloader, ok := m.(provider.Unloader)
+		if !ok {
+			continue
+		}
+		callCtx, cancel := context.WithTimeout(ctx, unloadOnSwitchTimeout)
+		if err := unloader.Unload(callCtx); err != nil {
+			slog.Warn("unload-on-switch failed",
+				"agent", prev.Name(), "model", m.ID(), "error", err)
+		}
+		cancel()
+	}
+}
--- a/pkg/runtime/unload.go
+++ b/pkg/runtime/unload.go
@@ -28,6 +28,9 @@ func (r *LocalRuntime) unloadOnSwitch(ctx context.Context, prev, next *agent.Age
 		return
 	}
 	for _, m := range prev.ConfiguredModels() {
+		if m == nil {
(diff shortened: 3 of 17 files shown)


rumpl:
Why isn't this implemented as a hook?

dgageot:
@rumpl I'll give it a try

aheritier (commented):
Overall LGTM — clean architecture, safe defaults, and comprehensive tests. Left a few inline nits below.

pkg/model/provider/openai/unload.go
```diff
@@ -0,0 +1,28 @@
...
+	path := c.ModelConfig.UnloadAPI()
+	if path == "" {
+		return nil
+	}
+	endpoint, err := base.JoinHostAndPath(c.ModelConfig.BaseURL, path)
+	if err != nil {
+		return err
+	}
+	return base.PostUnloadModel(ctx, httpclient.NewHTTPClient(ctx), endpoint, c.ModelConfig.Model)
+}
```
aheritier, line 28: This creates a fresh `*http.Client` on every `Unload` call, which means any `http_headers` configured in `provider_opts` won't be forwarded to the unload endpoint. For the current use-case (unauthenticated local engines) this is fine, but it's asymmetric with the DMR path that reuses `c.httpClient`.

Worth a short comment so future maintainers don't wonder why custom headers aren't forwarded here:
```go
// httpclient.NewHTTPClient is used instead of reusing the SDK client because
// the openai.Client wraps its transport and there's no clean way to extract a
// raw *http.Client from it. For local engines (the only use-case for unload_api)
// this is fine — they typically don't require auth headers.
```

pkg/runtime/unload.go
```diff
@@ -0,0 +1,49 @@
...
+			continue
+		}
+		cfg := m.BaseConfig().ModelConfig
+		if !cfg.UnloadOnSwitch() {
+			continue
+		}
+		unloader, ok := m.(provider.Unloader)
+		if !ok {
+			continue
+		}
```
aheritier, line 41: Minor: `cancel()` is called after `Unload` returns, which is correct, but `defer cancel()` would be the idiomatic Go pattern and would also handle a potential panic inside `Unload` without leaking the timer goroutine. Up to you — the current form is functionally fine since `Unload` implementations don't panic.

pkg/runtime/unload.go
```diff
@@ -0,0 +1,49 @@
...
+// unreachable engine cannot break agent switching.
+//
+// Every configured model is considered (not just the currently-selected
+// one) so alloyed agents free every variant they may have loaded.
+func (r *LocalRuntime) unloadOnSwitch(ctx context.Context, prev, next *agent.Agent) {
+	if prev == nil || prev == next {
+		return
+	}
+	for _, m := range prev.ConfiguredModels() {
+		if m == nil {
```
aheritier, line 31: Nit: `FallbackModels()` isn't iterated here. That's probably intentional — a fallback model may never have been loaded — but worth a one-liner comment to make the deliberate choice clear:
```go
// Only ConfiguredModels are considered; FallbackModels are skipped because
// they may never have been loaded and unloading an absent model is harmless
// but wasteful.
```

pkg/runtime/runtime.go
```diff
@@ -530,7 +530,22 @@ func (r *LocalRuntime) CurrentAgentInfo(context.Context) CurrentAgentInfo {
...
+	}
+	// Run the unload in the background: the TUI agent picker drives this
+	// path from the bubbletea Update loop, so a synchronous call would
+	// freeze the UI for as long as the engine takes to acknowledge the
+	// unload (up to unloadOnSwitchTimeout). Fire-and-forget is safe here
+	// because the new agent's model isn't loaded until the user sends a
+	// message — typically long after the unload completes.
+	go r.unloadOnSwitch(context.Background(), prev, next)
+	return nil
 }
```
aheritier, line 550: If the TUI agent picker fires two rapid switches (A→B→C), two goroutines can be in flight simultaneously: one unloading A, one unloading B. If A and C happen to share a model, that model gets unloaded just before C needs it, causing an extra reload. Not a correctness issue given the best-effort semantics, but worth a comment so it's an acknowledged trade-off rather than an oversight.

aheritier (approved):
LGTM. Clean architecture: optional `Unloader` interface, shared HTTP/URL helpers in `base`, best-effort semantics with 10s timeout, the async path for the TUI picker is a good catch and well-justified.

The four inline nits I left earlier are all non-blocking — feel free to address them in a follow-up if you prefer:
- `defer cancel()` in `unloadOnSwitch` (idiom)
- comment about `ConfiguredModels()` only, skipping `FallbackModels()`
- comment about the rapid A→B→C switch race
- comment about why the OpenAI path can't reuse the SDK client

CI all green, comprehensive test coverage including `-race`.

dgageot:
@rumpl do you prefer this one? https://github.com/docker/docker-agent/pull/2684

dgageot:
Closing in favour of https://github.com/docker/docker-agent/pull/2684

PR #2666, opened 2026-05-06
refactor: centralize model-specific behavior in pkg/modelinfo

dgageot:
## Why

We had model-specific feature toggles scattered across the codebase as
ad-hoc name-pattern checks: Responses-API selection in `openai/client.go`,
o-series default thinking in `provider/defaults.go`, adaptive thinking for
Opus 4.6/4.7 in `anthropic/thinking.go`, Gemini 3 thinking-level detection
duplicated in `gemini/client.go` *and* `provider/defaults.go` (with a
\"keep in sync\" comment), Bedrock-Claude detection only handling the
\`global.\` prefix, plus dead helpers like \`isGeminiProModel\` /
\`isGeminiFlashModel\` that were only referenced by their own tests.

Adding behavior for a new model meant editing several files and hoping
the patterns stayed in sync.

## What

A new \`pkg/modelinfo\` package is now the single home for every
model-specific behavior decision. Each predicate is named after the
*capability* it controls (not the model version), and its doc comment
explains the underlying API rule.

| Capability | Replaces |
|---|---|
| \`SupportsResponsesAPI\` | \`openai.isResponsesModel\` |
| \`UsesReasoningEffort\` | \`openai.isOpenAIReasoningModel\` |
| \`AlwaysReasons\` | \`provider.isOpenAIThinkingOnlyModel\` |
| \`RejectsTokenThinking\` | \`anthropic.requiresAdaptiveThinking\` |
| \`UsesThinkingLevel\` | \`gemini.isGemini3PlusModel\` + \`provider.gemini3Family\` |
| \`IsBedrockClaudeID\` | \`provider.isBedrockClaudeModel\` |
| \`IsClaude(ctx, store, …)\` *(new)* | family-based check that works for Anthropic / Bedrock / Vertex / future hosts |
| \`LookupFamily\` *(new)* | thin wrapper over models.dev's \`family\` field |
| \`IsClaudeFamily\` *(new)* | helper for family-based grouping |

## Leveraging models.dev more

\`pkg/modelsdev.Model\` now exposes the previously-unused fields from
the upstream API: \`Reasoning\`, \`ToolCall\`, \`Temperature\`,
\`Attachment\`, \`OpenWeights\`, \`ReleaseDate\`. They populate
automatically from the existing cache.

The new \`IsClaude\` helper resolves through models.dev's canonical
\`family\` field first (so any model the upstream registry tags
\`claude-*\` is recognised), falling back to the Bedrock-style and
bare-name patterns when the store is unavailable. This makes Vertex AI
Anthropic, future hosts, and new family members work without code
changes.

## Behavior

- Every existing predicate preserves its semantics. The only intentional
  behavior change is that \`IsBedrockClaudeID\` now also matches the
  cross-region inference profiles that the old narrow check missed
  (\`us.anthropic.claude-…\`, \`eu.anthropic.…\`, \`apac.anthropic.…\`,
  \`au.anthropic.…\`). This is strictly additive: the worst case is that
  more Bedrock-hosted Claude users get the optional
  \`interleaved-thinking-2025-05-14\` beta header — which is the
  desired outcome.
- Dead code removed: \`gemini3Family\`, \`isGeminiProModel\`,
  \`isGeminiFlashModel\` and their tests.
- Test coverage migrated from the per-provider packages into a single
  \`pkg/modelinfo/modelinfo_test.go\`.

## Validation

- \`mise build\` ✅
- \`mise test\` ✅ (full suite passes, all Bedrock / Anthropic / Gemini
  end-to-end \`applyModelDefaults\` cases included)
- \`mise lint\` ✅ (0 issues)


rumpl:
conflicts

aheritier (commented):
## Review

Clean, well-structured refactoring. LGTM on the code — not approving yet due to merge conflicts.

### Strengths

- Capability-named predicates (`SupportsResponsesAPI`, `UsesReasoningEffort`, `AlwaysReasons`, `RejectsTokenThinking`, `UsesThinkingLevel`) are self-documenting and much easier to reason about than the scattered name-pattern checks they replace.
- Excellent doc comments on every exported function explaining the underlying API rule.
- `IsBedrockClaudeID` expansion to handle cross-region inference profiles (`us.`, `eu.`, `apac.`, `au.`) is a good catch — strictly additive.
- Dead code cleanup of `isGeminiProModel`, `isGeminiFlashModel`, `gemini3Family` is welcome.
- Test coverage properly migrated with all existing edge cases preserved.

### Discussion Points

1. **`IsClaude` / `LookupFamily` unused in production code** — these are defined and tested but not wired into any call site in this PR. `needsInterleavedThinking` still hardcodes `case "anthropic": return true` rather than calling `IsClaude`. Is this intentional scaffolding for a follow-up (Vertex AI support)?

2. **Package doc "zero-allocation" claim** — the Is* predicates call `normalize()` which does `strings.ToLower` + `strings.TrimSpace` (allocates). Minor doc accuracy nitpick — the performance is fine and matches the old code.

3. **`modelsdev.Model` field additions** (`Reasoning`, `ToolCall`, `Temperature`, etc.) — understood these are forward-looking. They'll auto-populate from the upstream cache, which is nice.

### Blocker

Merge conflicts need resolution before this can land.

aheritier (approved):
LGTM. Rebase resolved the conflict cleanly, all checks green. The capability-named predicates and centralization in `pkg/modelinfo` are a solid improvement; the additive `IsBedrockClaudeID` cross-region prefix expansion is a nice catch. The non-blocking points from my earlier review (`IsClaude`/`LookupFamily` not yet wired into `needsInterleavedThinking`, "zero-allocation" doc note) can be addressed as follow-ups.

PR #2668, opened 2026-05-06
tui: pause animation ticks while the terminal is blurred

dgageot:
## Why

When docker-agent runs in a background tab while the agent is streaming, the bubbletea animation tick chain (14 fps spinner + cache invalidations + view rebuilds) keeps firing into a screen nobody is looking at. That's pure CPU waste.

## What

Drop \`animation.TickMsg\` while the terminal is blurred and let the chain die. Re-arm it on the next \`FocusMsg\` via \`animation.StartTick()\` so spinners resume immediately when the user comes back. The existing Docker Desktop release/restore cycle (commit cc2dfcfb) is preserved.

## Why not just \`tea.WithFPS(30)\`?

Static FPS reduction halves the ~60 cheap goroutine wake-ups per second everywhere — well under 1 % CPU savings. This change targets the **expensive** work — full \`Update\` + \`View\` + cache invalidation per spinner tick — exactly when the user can't see it (background tab), and **costs nothing** when they're looking at the terminal (foreground stays at full responsiveness).

## Implementation note

Two flags rather than one:

- \`focused\` — for the spurious-\`FocusMsg\` filter introduced by Docker Desktop's release/restore (must default to \`false\` so the first real \`FocusMsg\` is treated as a transition).
- \`tickPaused\` — for animation gating (must default to \`false\` so ticks flow at startup, before any focus event arrives — some terminals never send one).

They share most transitions but their initial values differ, so keeping them separate avoids cute-but-wrong overloading.

### DIFF
--- a/pkg/tui/tui.go
+++ b/pkg/tui/tui.go
@@ -131,11 +131,20 @@ type appModel struct {
 	// environment.
 	dockerDesktop bool
 
-	// focused tracks whether the terminal currently has focus. Used both
-	// to filter spurious FocusMsg events (RestoreTerminal re-enables focus
-	// reporting and delivers one) and to skip animation ticks when nobody
-	// is looking at the terminal — the chain re-arms on FocusMsg.
+	// focused tracks whether the terminal currently has focus. Used to
+	// filter spurious FocusMsg events (RestoreTerminal re-enables focus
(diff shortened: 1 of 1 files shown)


derekmisler (approved):
my almost-4-year-old laptop thanks you.

Issue #10, opened 2025-08-26
Add subagent execution to API


christopher-pluntke:
The API can currently execute an agent yaml starting at the root agent only. The CLI allows execution of subagents (cagent run --agent snowflake_researcher agent.yaml). 

For debug purposes, please add a direct way to execute subagents in the API as well, something like /api/sessions/$SESSION_ID/agent-file/subagent1

rumpl:
Done!

PR #2670, opened 2026-05-06
security: SSRF / TOCTOU / OAuth state hardening

dgageot:
## Summary

A scan of the codebase surfaced 10 security issues; this PR fixes four of the higher-severity ones plus a TOCTOU regression caught during review. Each fix is in its own commit and is independently revertable.

| Commit | Area | What |
| --- | --- | --- |
| `a031e6159` | OAuth | Constant-time state compare; stop leaking the expected state in error messages. |
| `e664cd9d1` | SSRF | API and OpenAPI tools now refuse non-public IPs at dial time (defeats DNS rebinding). New reusable `pkg/httpclient` SSRF helpers; `pkg/config/sources.go` reuses them. |
| `3dcd21f49` | SSRF | Built-in `fetch` tool refuses non-public IPs by default. New `allow_private_ips: true` opt-in for operators that legitimately need internal services. |
| `147ed1e92` | TOCTOU | Filesystem toolset routes I/O through `*os.Root` handles so symlink-swap races between the path check and the I/O cannot escape the sandbox. |
| `f1837ad53` | TOCTOU | Reviewer follow-up: `handleSearchFilesContent` was still using `os.ReadFile`, bypassing the rooted I/O fix. One-line change. |
| `f2e095f4d`, `55495062d` | tests | Regression coverage for IPv4-mapped IPv6 (`::ffff:169.254.169.254`) and IPv6 zone IDs (`fe80::1%eth0`). |
| `8bd231393`, `ea30a6840` | docs | `PLATFORM NOTE` documenting `*os.Root` carve-outs that matter for the agent's threat model (Linux bind mounts, `/proc`, device files; `GOOS=js` weakness; plan9 rename behaviour). |

### #4 — OAuth state mismatch leaks the expected state via timing and error logs
- `pkg/tools/mcp/oauth_helpers.go:174-176`, `pkg/tools/mcp/oauth_server.go:172-174`
- `state != expectedState` was non-constant-time AND the error string embedded the expected value (`"state mismatch: expected %s, got %s"`), which routinely lands in slog/stderr/log files. Anyone with read access to those logs could replay a valid state to forge the OAuth callback.
- **Fix**: `crypto/subtle.ConstantTimeCompare` + a fixed error message. Empty values are also rejected so an attacker can't satisfy the check by sending no state when no expected state has been stored yet.

### #7 — APITool / OpenAPITool issue requests with no SSRF protection
- `pkg/tools/builtin/api.go`, `pkg/tools/builtin/openapi.go`
- Both used `remote.NewTransport(ctx)`, which has no IP filter and additionally routes through the Docker Desktop proxy (Unix socket → sidesteps any IP-level check). A LLM-influenced placeholder in an API endpoint, or a malicious OpenAPI spec whose `servers[].url` points at `169.254.169.254`, could reach internal services or cloud metadata.
- **Fix**: extract the SSRF protection that `pkg/config/sources.go` was carrying internally into a new reusable `pkg/httpclient/ssrf.go`:
  - `IsPublicIP(ip)` — routability classifier
  - `SSRFDialControl` — `net.Dialer.Control` hook that rejects loopback / RFC1918 / link-local / multicast / unspecified post-DNS (so DNS rebinding cannot bypass it)
  - `NewSSRFSafeTransport()` — clone of `http.DefaultTransport` with the dial control installed (inherits future stdlib tweaks)
  - `BoundedRedirects(n)`, `HTTPSOnlyRedirects(n)` — `CheckRedirect` helpers
- Both tools now use a shared `safeHTTPClient(timeout, unsafe)` factory in `pkg/tools/builtin/httpclient.go`. Production callers (`NewAPITool`, `NewOpenAPITool`) always get SSRF protection. The `unsafe` flag is settable only by test-only constructors so `httptest.NewServer` (binds to 127.0.0.1) keeps working.
- Regression tests assert that the production constructors refuse `127.0.0.1`, `10.0.0.1`, `169.254.169.254`, etc.

### #8 — Built-in `fetch` tool has no default SSRF deny list
- `pkg/tools/builtin/fetch.go`
- Out of the box the tool happily fetched `http://localhost`, RFC1918, IPv6 ULA, and the cloud metadata endpoint at `169.254.169.254`. Even with `allowed_domains` configured, a public hostname that DNS-resolved to a private IP would slip through.
- **Fix**: switch the default transport to `httpclient.NewSSRFSafeTransport`. Add a first-class `allow_private_ips: bool` config field on the toolset:
  - `latest` config schema: new field on `Toolset`, rejected on non-fetch types in `validate.go`, documented in `agent-schema.json`.
  - `pkg/teamloader/registry.go` plumbs the flag through `builtin.WithAllowPrivateIPs`.
  - Existing tests gain a `newFetchToolForTest` helper that prepends `WithAllowPrivateIPs(true)`. New regression tests assert the production default refuses local addresses.
- **⚠️ Behaviour change for users**: agents that previously used `fetch` against localhost / RFC1918 / cloud metadata will now fail until they add `allow_private_ips: true` to the toolset YAML. The error message (`refusing to dial non-public address`) is self-explanatory.

```yaml
toolsets:
  - type: fetch
    allow_private_ips: true     # explicit opt-in, off by default
    allowed_domains:
      - internal.example.com
      - 10.0.0.0/8
```

### #9 — Filesystem write/edit tools have an acknowledged TOCTOU symlink race
- `pkg/tools/builtin/filesystem.go`, `pkg/tools/builtin/filesystem_paths.go`
- `resolveAndCheckPath` did a static check (lexical + rooted Lstat) and then handed the path back to the caller, who issued plain `os.*` I/O. A concurrent process (or hostile post-edit hook) could swap the file for a symlink to `/etc/passwd` in the gap and the subsequent `os.WriteFile` would follow it out of the sandbox.
- **Fix**: route every read, write, stat, mkdir, readdir and rmdir through the `*os.Root` handles that `pathRootSet` already opens. The kernel re-checks containment on every component — `..` segments and out-of-root symlinks are rejected at lookup time, regardless of timing.
  - `pathRootSet.entryFor` is the shared lexical lookup used by both `contains()` and the new rooted helpers.
  - `FilesystemTool.{readFile,writeFile,stat,mkdirAll,readDir,removeDir}` issue I/O through `*os.Root` when an allow-list root is open; fall back to plain `os.*` otherwise.
  - **Fail-closed** when the allow-list is configured but the (post-swap) real path is no longer inside any entry — falling back to `os.*` would follow the symlink.
  - All filesystem handlers updated, including `handleSearchFilesContent` (`f1837ad53` follow-up).
- Regression tests run the static check while the file is legitimate, swap it for a symlink to an outside secret, and exercise the rooted writer with the previously-validated path. The write must fail and the secret must remain unchanged on disk.

## Validation

- `go build ./...` ✓
- `go vet ./...` ✓
- `mise test` ✓ — 107 packages, 0 failures
- `mise lint` ✓ — golangci-lint + custom cops, 943 files, 0 issues

## Out of scope (left for follow-up PRs)

The original scan flagged 6 more issues that aren't fixed here (issue numbers from that report):
- **#1** A2A server has wildcard CORS + no auth on `/invoke`
- **#2** docker-agent HTTP API server has no authentication or origin checks
- **#3** OAuth token / refresh / DCR requests use `http.DefaultClient` (no timeout, no HTTPS enforcement)
- **#5** Tool installer downloads & executes binaries with no checksum/signature verification (supply chain)
- **#6** Tool registry index download is unbounded (`io.ReadAll`) and uses `http.DefaultClient`
- **#10** Sandbox token file mode relies on the default temp-file mode (low/medium)

Each is independently scoped; I can open follow-ups on request.

Assisted-By: docker-agent

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1447,6 +1447,10 @@
             ["169.254.0.0/16", "10.0.0.0/8"]
           ]
         },
+        "allow_private_ips": {
+          "type": "boolean",
+          "description": "Opt in to dialling non-public IP addresses (only valid for type 'fetch'). By default the fetch tool refuses connections — after DNS resolution, so DNS rebinding is also blocked — to loopback, RFC1918 private ranges, link-local (including the cloud metadata endpoint at 169.254.169.254), multicast and the unspecified address. Set this to true when an agent legitimately needs to call internal services. 'allowed_domains' / 'blocked_domains' are evaluated independently and still apply."
+        },
         "url": {
           "type": "string",
           "description": "URL for the a2a or openapi tool",
--- a/pkg/config/latest/types.go
+++ b/pkg/config/latest/types.go
@@ -855,6 +855,21 @@ type Toolset struct {
 	// `allowed_domains`.
 	BlockedDomains []string `json:"blocked_domains,omitempty" yaml:"blocked_domains,omitempty"`
 
+	// For the `fetch` tool — opt in to dialling non-public IP addresses.
+	//
+	// By default the fetch tool refuses connections (after DNS resolution,
+	// so DNS rebinding is also blocked) to loopback (127/8, ::1), RFC1918
+	// private ranges, link-local — including the cloud metadata endpoint
+	// at 169.254.169.254 — multicast and the unspecified address. Set this
+	// to true to permit those addresses, which is required when an agent
+	// legitimately needs to call internal services.
+	//
+	// `allowed_domains` and `blocked_domains` are evaluated independently
+	// of this flag: even with `allow_private_ips: true`, an entry in
+	// `blocked_domains` (or absence from `allowed_domains`) still rejects
+	// the request before any network call.
+	AllowPrivateIPs bool `json:"allow_private_ips,omitempty" yaml:"allow_private_ips,omitempty"`
+
 	// For the `rag` tool
 	RAGConfig *RAGConfig `json:"rag_config,omitempty" yaml:"rag_config,omitempty"`
 
--- a/pkg/config/latest/validate.go
+++ b/pkg/config/latest/validate.go
@@ -97,6 +97,9 @@ func (t *Toolset) validate() error {
 	if len(t.BlockedDomains) > 0 && t.Type != "fetch" {
 		return errors.New("blocked_domains can only be used with type 'fetch'")
 	}
+	if t.AllowPrivateIPs && t.Type != "fetch" {
+		return errors.New("allow_private_ips can only be used with type 'fetch'")
+	}
 	if len(t.AllowedDomains) > 0 && len(t.BlockedDomains) > 0 {
 		return errors.New("allowed_domains and blocked_domains are mutually exclusive")
 	}
--- a/pkg/config/latest/validate_test.go
+++ b/pkg/config/latest/validate_test.go
@@ -299,6 +299,31 @@ agents:
 			wantErr: "blocked_domains can only be used with type 'fetch'",
 		},
 		{
+			name: "allow_private_ips on non-fetch toolset is rejected",
+			config: `
+version: "8"
+agents:
+  root:
+    model: "openai/gpt-4"
+    toolsets:
+      - type: shell
+        allow_private_ips: true
+`,
+			wantErr: "allow_private_ips can only be used with type 'fetch'",
+		},
+		{
+			name: "allow_private_ips on fetch toolset is accepted",
+			config: `
+version: "8"
+agents:
+  root:
+    model: "openai/gpt-4"
+    toolsets:
+      - type: fetch
+        allow_private_ips: true
+`,
+		},
+		{
 			name: "empty allowed_domains entry is rejected",
 			config: `
 version: "8"
--- a/pkg/config/sources.go
+++ b/pkg/config/sources.go
@@ -8,14 +8,12 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
 	"slices"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/docker/docker-agent/pkg/content"
@@ -257,7 +255,11 @@ func (a urlSource) Read(ctx context.Context) ([]byte, error) {
 
 	client := httpclient.NewHTTPClient(ctx)
 	if !a.unsafe {
-		client = ssrfSafeHTTPClient()
+		client = &http.Client{
+			Timeout:       60 * time.Second,
+			Transport:     httpclient.NewSSRFSafeTransport(),
+			CheckRedirect: httpclient.HTTPSOnlyRedirects(10),
+		}
 	}
 
 	resp, err := client.Do(req)
@@ -366,8 +368,8 @@ func IsURLReference(input string) bool {
 
 // validateAgentURL enforces that an agent URL uses HTTPS. SSRF protection
 // (rejecting connections to loopback / private / link-local addresses) is
-// done at dial time by ssrfSafeHTTPClient so that DNS rebinding cannot be
-// used to bypass it.
+// done at dial time by [httpclient.NewSSRFSafeTransport] so that DNS
+// rebinding cannot be used to bypass it.
 func validateAgentURL(rawURL string) error {
 	u, err := url.Parse(rawURL)
 	if err != nil {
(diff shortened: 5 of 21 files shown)


rumpl:
condflicts

aheritier (commented):
## Review: SSRF / TOCTOU / OAuth State Hardening

Overall this is a well-structured security PR. The fixes are independently revertable, fail-closed by design, and comprehensively tested (+1007/−277 lines, overwhelmingly test coverage). The commit decomposition is excellent.

**Not approving yet** — there are merge conflicts with the current `main` branch that need to be resolved first.

---

## Per-Fix Review

### 1. OAuth State Mismatch (`a031e61`) — ✅ Good

- Correctly uses `crypto/subtle.ConstantTimeCompare`
- Rejects empty values (prevents nil-state bypass — a subtle attack vector)
- Error messages no longer leak the expected state into logs

**Nit:** `oauth_server.go:172` uses inline `subtle.ConstantTimeCompare` while `oauth_helpers.go` wraps it in the new `constantTimeStateEqual` helper. Minor inconsistency — consider using the helper in both places for uniformity.

### 2. SSRF in API/OpenAPI Tools (`e664cd9`) — ✅ Good

- The `pkg/httpclient/ssrf.go` extraction is clean and reusable
- `NewSSRFSafeTransport()` clones `http.DefaultTransport` — inherits future stdlib changes automatically
- The `unsafe` field is unexported, only settable by test-only constructors in `*_test.go` files — production callers cannot bypass it
- `safeHTTPClient` uses `BoundedRedirects(10)` (not `HTTPSOnlyRedirects`) — correct because API/OpenAPI tools legitimately target HTTP endpoints

**Minor behavioural change:** The old `ssrfSafeHTTPClient` in `sources.go` had `ResponseHeaderTimeout: 30s`. The new transport (from `DefaultTransport.Clone()`) does not set this. The overall client timeout (60s) still caps requests, so this is acceptable, but consider adding it back to `NewSSRFSafeTransport` to prevent slow-loris-style attacks where a server accepts the connection but never sends headers.

### 3. SSRF in Fetch Tool (`3dcd21f`) — ✅ Good (⚠️ Breaking Change)

- Clean config plumbing: schema → `types.go` → `validate.go` → `registry.go` → `WithAllowPrivateIPs`
- Validation correctly rejects `allow_private_ips` on non-fetch toolset types
- The breaking change is well-documented and the error message (`"refusing to dial non-public address"`) is self-explanatory

The migration path (`allow_private_ips: true` in YAML) is simple. This is the right default.

### 4. TOCTOU Filesystem Fix (`147ed1e` + `f1837ad`) — ✅ Good

This is the most complex fix and it is well-designed:

- **`rootedAccess`** has clear fail-closed semantics: if the path is no longer inside any entry (symlink swap detected), it returns an error — never falls back to plain `os.*`
- **Fallback to `os.*`** only happens when: (a) no allow-list is configured, or (b) `entry.root` is nil (directory did not exist at construction time)
- **`entryFor` refactor** separates the lexical lookup from the containment check — cleaner separation of concerns
- **All handlers** are consistently updated (read, write, stat, mkdir, readdir, rmdir, searchFilesContent)
- The `readDir` implementation uses `root.Open(rel)` + `f.ReadDir(-1)` correctly since `*os.Root` does not expose ReadDir directly

**Regression tests are excellent** — they faithfully simulate the race window:
1. Validate path while file is legitimate
2. Swap file for a symlink to an outside secret
3. Attempt I/O with the previously-validated path
4. Assert I/O fails AND secret is untouched

The PLATFORM NOTE documenting `*os.Root` limitations (Linux bind mounts, `/proc`, WASM weakness, plan9 rename tracking) is a valuable addition for the threat model.

### 5. Test Coverage (`f2e095f`, `5549506`) — ✅ Good

- IPv4-mapped IPv6 (`::ffff:169.254.169.254`) regression test pins Go's correct behaviour
- IPv6 zone ID (`fe80::1%eth0`) documents the fail-closed path (`net.ParseIP` returns nil)

---

## Minor Suggestions (Non-Blocking)

| # | Location | Suggestion |
|---|----------|-----------|
| 1 | `oauth_server.go:172` | Use `constantTimeStateEqual(state, expectedState)` instead of inline `subtle.ConstantTimeCompare` for consistency with `oauth_helpers.go` |
| 2 | `openapi_test.go:TestOpenAPITool_RejectsLocalSpecServerURL` | The test name suggests it tests `servers[].url` rejection, but it actually tests that the production constructor refuses to *fetch* from a loopback spec URL. Consider renaming to `TestOpenAPITool_ProductionRefusesLoopbackSpecFetch` |
| 3 | `pkg/httpclient/ssrf.go` | Consider adding `ResponseHeaderTimeout` to `NewSSRFSafeTransport` to preserve the behaviour the old `sources.go` had |

---

## Questions

1. **`handleSearchFilesContent` walk path:** The `path` variable from `filepath.WalkDir`'s callback is passed to `t.readFile()`. Since `WalkDir` may follow symlinks during traversal itself (before reaching the read), is there a risk that the walked path has already escaped? Or does the initial `resolveAndCheckPath` on the root dir + the rooted I/O check on each file close this?

2. **`entry.root == nil` fallback in `rootedAccess`:** When the allow-list root did not exist at construction time, we fall back to plain `os.*`. This preserves the pre-existing behaviour, but it does leave the TOCTOU window open for roots that are lazily created. Is this an acceptable trade-off, or should the toolset refuse I/O until the root directory actually exists?

---

**Verdict:** Looks good pending conflict resolution with `main`. Once rebased, happy to approve.

aheritier (approved):
LGTM. Rebase onto current `main` resolved the only blocker from my prior review; CI is green across lint, license-check, build-and-test, and build-image at `8b296dcb`.

The substantive review I left on `ea30a684` still stands — the four fixes are well-scoped, fail-closed, and comprehensively tested. The non-blocking suggestions (use `constantTimeStateEqual` consistently in `oauth_server.go`, add `ResponseHeaderTimeout` to `NewSSRFSafeTransport`, rename the OpenAPI loopback-fetch test) and my two open questions (`WalkDir` symlink-traversal in `handleSearchFilesContent`, `entry.root == nil` fallback for lazily-created roots) can all be addressed in follow-ups — none of them are merge-blocking.

Given the breaking change in fetch behaviour (`allow_private_ips: true` opt-in for localhost / RFC1918 / metadata IPs), please make sure the changelog / release notes call this out so users hitting "refusing to dial non-public address" know what to do.

PR #2677, opened 2026-05-06
feat: add shadow snapshots and undo

rumpl:
Add a git snapshot manager and register it as a builtin hook. Wire user config so snapshots can be auto-enabled for every local runtime.

Add a TUI /undo command that restores file changes from the latest changed snapshot without modifying the session transcript. Keep undo checkpoints available after a response stops so the command works once the agent is idle.

Document snapshot hooks and add an example agent configuration.


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25484320535).

dgageot (approved):

PR #2685, opened 2026-05-07
feat: attach-time processing – transcode/resize images and resolve URLs at message add time

simonferquel-clanker:
## Summary

Attach-time processing pipeline for the Phase 1 attachment system. Part of #2595.

**Scope**: library-only. This PR adds `pkg/chat/attach.go` only. Wiring into the TUI (`pkg/app`) and CLI (`pkg/cli`) is intentionally deferred until the library API is stable.

### `pkg/chat/attach.go` (new)

Two new functions:

- `ProcessAttachment(ctx, part) (Document, error)` — converts a raw `MessagePart` into a fully-resolved `chat.Document` with `InlineData` or `InlineText`. Called once when a message is assembled; never at inference time.
- `ProcessAttachmentWithMetadata(ctx, part) (Document, *ImageResizeResult, error)` — same, but also returns the `ImageResizeResult` so callers can emit a dimension note without a redundant second resize call.

Handles three input types:

| Part type | Routing |
|---|---|
| `MessagePartTypeFile` | MIME-first routing (applied before `IsTextFile` heuristic to avoid misclassifying ASCII-content PDFs as text). Images → `ResizeImage` transcode; PDF/other binary → verbatim `InlineData`; `text/*` → `ReadFileForInline` (InlineText). All binary reads capped at `MaxInlineBinarySize` (20 MB). |
| `MessagePartTypeImageURL` | `data:` URIs decoded inline; `http(s)://` fetched via `safeHTTPClient` (see security section). Bytes capped at 20 MB; result goes through `ResizeImage`. |
| `MessagePartTypeDocument` | Images with `InlineData` transcoded; other already-resolved documents passed through unchanged. No inline content → error. |

### Security

**SSRF protection** in `fetchRemoteImage`:

The custom `safeHTTPClient` blocks connections to private/reserved IP ranges at both dial time and on each redirect hop:
- Loopback: `127.0.0.0/8`, `::1/128`
- Link-local / cloud metadata endpoints: `169.254.0.0/16`, `fe80::/10`
- RFC-1918 private ranges: `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`
- IPv6 unique-local: `fc00::/7`

DNS resolution uses `net.DefaultResolver.LookupHost` with context (satisfies `noctx` linter). Both `DialContext` and `CheckRedirect` enforce the block.

**Unbounded read protection**: `fi.Size() > MaxInlineBinarySize` guard added before all `os.ReadFile` calls for binary files (images, PDFs, unknown binary).

## Tests (25 cases in `pkg/chat/attach_test.go`)

- JPEG/PNG passthrough, PNG-with-alpha preserved
- Oversize image resized (asserts output dimensions ≤ `MaxImageDimension`)
- PDF passthrough (verifies MIME-first routing bypasses text heuristic for ASCII-content PDF)
- Binary file too large → error (sparse file, instant)
- Text file and Markdown → `InlineText`
- `data:` URI JPEG and PNG → `InlineData`
- Non-base64 `data:` URI → error
- Unsupported URL scheme → error
- `http://` fetch via `httptest.Server` (success + HTTP 404) — uses `SetFetchHTTPClientForTest` to bypass SSRF filter for loopback test servers
- **SSRF block**: `127.0.0.1` (loopback) → error with "private/reserved"
- **SSRF block**: `169.254.169.254` (AWS/GCP/Azure metadata) → error with "private/reserved"
- `MessagePartTypeDocument` passthrough (binary, text) and image transcode
- Document with no content → error; nil document → error
- Unsupported part type → error

## What's NOT changed

- `pkg/app/app.go` — unchanged from upstream/main
- `pkg/cli/runner.go` — unchanged from upstream/main
- Provider packages — unchanged (they receive `chat.Document` via `convertDocument`, already in #2639)

Part of #2595

### DIFF
--- a/pkg/chat/attach.go
+++ b/pkg/chat/attach.go
@@ -17,127 +17,16 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"io"
-	"mime"
-	"net"
-	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
-	"time"
 )
 
 const (
-	// attachHTTPTimeout is the maximum time allowed to fetch a remote image URL.
-	attachHTTPTimeout = 10 * time.Second
-
-	// attachMaxRemoteBytes is the maximum number of bytes read from a remote URL.
-	attachMaxRemoteBytes = 20 * 1024 * 1024 // 20 MB
-
-	// MaxInlineBinarySize is the maximum byte size of a binary file (PDF, etc.)
-	// that will be read into memory for inline attachment. Matches the remote
-	// fetch cap so local and remote paths behave consistently.
+	// MaxInlineBinarySize is the maximum byte size of a binary file (PDF, image,
+	// etc.) that will be read into memory for inline attachment.
 	MaxInlineBinarySize = 20 * 1024 * 1024 // 20 MB
 )
-
-// privateIPNets lists address ranges that must not be dialled by the
-// attach-time URL fetcher. Blocking these prevents SSRF attacks against
-// cloud metadata services, internal APIs, and loopback services.
-var privateIPNets = func() []*net.IPNet {
-	blocks := []string{
-		// Loopback
-		"127.0.0.0/8",
-		"::1/128",
-		// Link-local — covers AWS/GCP/Azure metadata endpoints (169.254.169.254)
-		"169.254.0.0/16",
-		"fe80::/10",
-		// RFC-1918 private ranges
-		"10.0.0.0/8",
-		"172.16.0.0/12",
-		"192.168.0.0/16",
-		// IPv6 unique-local
-		"fc00::/7",
-	}
-	nets := make([]*net.IPNet, 0, len(blocks))
-	for _, b := range blocks {
-		_, ipNet, err := net.ParseCIDR(b)
-		if err != nil {
-			panic("attachment: invalid built-in CIDR " + b + ": " + err.Error())
-		}
-		nets = append(nets, ipNet)
-	}
-	return nets
-}()
-
-// isPrivateIP reports whether ip falls in any of the blocked address ranges.
-func isPrivateIP(ip net.IP) bool {
-	for _, block := range privateIPNets {
-		if block.Contains(ip) {
-			return true
-		}
-	}
-	return false
-}
-
-// checkSafeHostCtx resolves host to IP addresses and returns an error if any
-// resolved address is in a private/reserved range. This is called both at
-// dial time (where a context is available) and on redirect destinations.
-func checkSafeHostCtx(ctx context.Context, host string) error {
-	addrs, err := net.DefaultResolver.LookupHost(ctx, host)
-	if err != nil {
-		return fmt.Errorf("attachment: cannot resolve host %q: %w", host, err)
-	}
-	for _, addr := range addrs {
-		ip := net.ParseIP(addr)
-		if ip == nil {
-			continue
-		}
-		if isPrivateIP(ip) {
-			return fmt.Errorf("attachment: URL resolves to private/reserved address %s (SSRF protection)", addr)
-		}
-	}
-	return nil
-}
-
-// safeHTTPClient is a shared HTTP client used by fetchRemoteImage.
-// It enforces SSRF protection by refusing connections to private and
-// reserved IP ranges at both dial time and on each redirect hop.
-//
-// Tests may override this variable to inject a custom client that bypasses
-// SSRF filtering (e.g. to reach httptest.Server on loopback). Only do this
-// in test code guarded by a t.Cleanup restore.
-var safeHTTPClient = newSafeHTTPClient()
-
-func newSafeHTTPClient() *http.Client {
-	return &http.Client{
-		Timeout: attachHTTPTimeout,
-		Transport: &http.Transport{
-			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
-				host, _, err := net.SplitHostPort(addr)
-				if err != nil {
-					return nil, fmt.Errorf("attachment: malformed dial address %q: %w", addr, err)
-				}
-				if err := checkSafeHostCtx(ctx, host); err != nil {
-					return nil, err
-				}
-				return (&net.Dialer{}).DialContext(ctx, network, addr)
-			},
-		},
-		CheckRedirect: func(req *http.Request, _ []*http.Request) error {
-			return checkSafeHostCtx(req.Context(), req.URL.Hostname())
-		},
-	}
-}
-
-// SetFetchHTTPClientForTest replaces the HTTP client used by fetchRemoteImage
-// and returns a restore function. Only for use in tests.
-//
-//	defer chat.SetFetchHTTPClientForTest(t, myClient)()
-func SetFetchHTTPClientForTest(client *http.Client) (restore func()) {
-	prev := safeHTTPClient
-	safeHTTPClient = client
-	return func() { safeHTTPClient = prev }
-}
 
 // ProcessAttachment converts a raw [MessagePart] into a [Document] with fully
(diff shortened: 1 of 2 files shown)


pkg/chat/attach.go
```diff
@@ -0,0 +1,326 @@
...
+func fetchRemoteImage(ctx context.Context, rawURL string) (data []byte, mimeType, name string, err error) {
+	ctx, cancel := context.WithTimeout(ctx, attachHTTPTimeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, rawURL, http.NoBody)
+	if err != nil {
+		return nil, "", "", fmt.Errorf("create request: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
docker-agent: **[MEDIUM] SSRF: `fetchRemoteImage` uses `http.DefaultClient` with no network filtering**

`fetchRemoteImage` issues an HTTP GET to any URL the caller supplies using the default Go HTTP client, which follows up to 10 redirects and performs no filtering on private/internal IP ranges:

```go
resp, err := http.DefaultClient.Do(req)
```

A user can supply a URL such as:
- `http://169.254.169.254/latest/meta-data/` — AWS IMDS (returns IAM credentials)
- `http://10.0.0.1/` / `http://localhost:8080/` — internal services accessible from the agent's network
- A redirect chain that terminates at an internal address

Since the agent processes user-provided `MessagePartTypeImageURL` parts, this is a direct SSRF vector for anyone who can send messages to the agent. The 10-second timeout and 20 MB cap do not mitigate the access itself.

**Suggested fix:** Use a custom `http.Transport` with a `DialContext` that rejects connections to loopback, link-local, and RFC-1918 addresses, or at minimum restrict the URL scheme allowlist and add a `CheckRedirect` policy that prevents cross-origin redirects to private IP space.

pkg/chat/attach.go
```diff
@@ -0,0 +1,326 @@
...
+		if err != nil {
+			return Document{}, nil, fmt.Errorf("ProcessAttachment: read image %q: %w", absPath, err)
+		}
+		return transcodeImageWithMeta(name, data, mimeType)
+
+	case mimeType == "application/pdf" || (IsSupportedMimeType(mimeType) && !IsTextFile(absPath)):
+		// PDF and other supported binary types — read verbatim.
+		// The !IsTextFile guard ensures that binary formats whose extension
+		// is unknown but content is ASCII-printable are not incorrectly inlined.
+		data, err := os.ReadFile(absPath)
```
docker-agent, line 115: **[MEDIUM] No size cap on local binary file reads — potential OOM with large PDF/binary attachments**

The PDF branch and the `default` (unknown binary) branch both call `os.ReadFile` with no size limit:

```go
case mimeType == "application/pdf" || (IsSupportedMimeType(mimeType) && !IsTextFile(absPath)):
    data, err := os.ReadFile(absPath)   // no size cap
```

```go
default:
    data, err := os.ReadFile(absPath)   // no size cap
```

A user who attaches a multi-GB PDF (or any large binary file) will cause the agent to read the entire file into memory before passing it on. The text file branch (line 130) and the remote image fetch (via `io.LimitReader`) both have explicit caps — this inconsistency looks like an oversight.

**Contrast with text branch:**
```go
case IsTextFile(absPath):
    if fi.Size() > MaxInlineFileSize {          // explicit cap ✔
        return Document{}, nil, fmt.Errorf(...)
    }
```

**Suggested fix:** Add a `fi.Size() > someMaxBinarySize` guard before each `os.ReadFile` call in the non-image binary branches. A reasonable upper bound might be `attachMaxRemoteBytes` (20 MB) to match the remote-image cap, or a separate constant for local binary files.

docker-agent (commented):

Two medium-severity issues found in the new attach-time processing pipeline (`pkg/chat/attach.go`):

1. **SSRF** — `fetchRemoteImage` uses `http.DefaultClient` with no IP/redirect filtering; user-supplied URLs can reach internal network services or cloud metadata endpoints.
2. **Unbounded binary read** — PDF and unknown-binary branches in `processFilePart` call `os.ReadFile` with no size cap, unlike the text branch which checks `MaxInlineFileSize`. A multi-GB file would be read fully into memory.

Reviewed: `pkg/chat/attach.go` (new), `pkg/app/app.go`, `pkg/cli/runner.go`.

simonferquel (changes requested):
For now do not wire the attachment conversion into the cli and tui. We will start with supporting it only when used as a library.

pkg/chat/attach.go
simonferquel-clanker: Fixed in 247c9198. Added `safeHTTPClient` with a custom `DialContext` that blocks connections to loopback (127.0.0.0/8, ::1), link-local/metadata endpoints (169.254.0.0/16, fe80::/10), RFC-1918 ranges (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16), and IPv6 unique-local (fc00::/7). `CheckRedirect` applies the same filtering to redirect destinations. Two new SSRF tests added.

pkg/chat/attach.go
simonferquel-clanker: Fixed in 247c9198. Added `fi.Size() > MaxInlineBinarySize` (20 MB) guard before `os.ReadFile` in all binary branches (image pre-read, PDF, and unknown binary). Consistent with the text branch cap. New test `TestProcessAttachment_BinaryFileTooLarge_Error` verifies the error path.

pkg/chat/attach.go
docker-agent: Thanks for the thorough fix! The implementation looks solid:

✅ **Comprehensive IP range blocking** — covers all the critical ranges (loopback, link-local/metadata, RFC-1918, IPv6 ULA)  
✅ **DNS resolution at dial time** — prevents bypasses via DNS rebinding or hostnames that resolve to private IPs  
✅ **Redirect filtering** — applies the same checks on every redirect hop via `CheckRedirect`, preventing chain-based SSRF  
✅ **Test coverage** — the two SSRF tests verify both loopback and metadata endpoint blocking

The test helper `SetFetchHTTPClientForTest` is a clean way to allow tests to reach local httptest servers without weakening the production path. This resolves the SSRF concern completely.

pkg/chat/attach.go
docker-agent: Perfect, thanks for the thorough fix! Verified that all three binary branches (image pre-read line 217, PDF line 228, and unknown binary line 251) now have the `fi.Size() > MaxInlineBinarySize` guard before `os.ReadFile`. The 20 MB cap is well-chosen — matches `attachMaxRemoteBytes` so local and remote paths behave consistently, as your comment notes.

The sparse-file test is a nice touch — keeps the suite fast while still exercising the error path.

simonferquel (dismissed):

pkg/chat/attach.go
```diff
@@ -0,0 +1,439 @@
...
+			Size:     int64(len(data)),
+			Source:   DocumentSource{InlineData: data},
+		}, nil, nil
+	}
+}
+
+// processImageURLPart handles MessagePartTypeImageURL.
+// Supports:
+//   - data: URIs (data:<mime>;base64,<payload>)
+//   - http:// and https:// URLs (fetched with attachHTTPTimeout)
```
rumpl: Let's not download things from the internet, we don't need to handle urls

simonferquel-clanker: Done in 04720162. Removed `fetchRemoteImage`, `safeHTTPClient`, and the entire SSRF mitigation. `processImageURLPart` now returns an error for `http(s)://` URLs: "remote URLs are not supported; download the file locally first". `data:` URI decoding is retained (local data, no network). Added `TestProcessAttachment_RemoteURL_Error` to cover both schemes.

rumpl (approved):

PR #2686, opened 2026-05-07
Make the FastMarkdown renderer simpler and faster

dgageot:
Refactors `pkg/tui/components/markdown/fast_renderer.go` for both readability and speed. Net `-201` lines, no API changes.

## What changed

- **No more `lipgloss.Style.Render` in hot paths.** Heading prefixes / continuation indents and the horizontal rule are pre-rendered once at theme init. Blockquote lines use the already-cached `ansiBlockquote.renderTo` instead of going back through lipgloss for every wrapped line. Three `lipgloss.Style` fields removed from `cachedStyles`.
- **Single post-processing pass.** `fixHyperlinkWrapping` and `padAllLines` were two consecutive walks of the rendered output; they are merged into one streaming `finalizeOutput` that tracks line width and OSC 8 hyperlink state in a single pass.
- **Simpler `wrapText`.** Dropped the intermediate `[]styledWord` slice (and the `splitWordsWithStyles` / `styledWord` helpers). Word boundaries, embedded ANSI sequences and active-style tracking are handled inline. Layout state is keyed off visible width (`lineWidth > 0`) rather than a separate `lineEmpty` flag, which also removes a corner case where a zero-width "word" of pure ANSI codes could insert a stray space.
- **Shared ANSI scanner.** Four near-identical CSI/OSC scanners (in `ansiStringWidth`, `finalizeOutput`, `wrapText`, `breakWord`) collapsed into a single `scanAnsiSequence` helper plus a small `classifyOSC8`. Lone-`\x1b` handling is now consistent across all callers and malformed sequences can no longer over-read.
- **Consolidated blockquotes.** Removed the line-for-line copy of `renderBlockquoteContent` / `renderBlockquoteCodeBlock` that lived in the list-blockquote path.
- Misc: renamed an awkwardly abbreviated field, refreshed two stale comments referring to the deleted helpers.

## Performance

Apple M4 Max, `go test -bench -benchtime=2s ./pkg/tui/components/markdown/`:

| Benchmark | Before | After | Δ |
|---|---|---|---|
| FastRenderer | 39 386 ns / 155 allocs | 17 176 ns / 73 allocs | **−56%** time, −53% allocs |
| FastRendererSmall | 1 196 ns | 832 ns | −30% |
| FastRendererCodeBlock | 3 933 ns / 7 360 B | 2 515 ns / 5 310 B | −36% time, −28% bytes |
| FastRendererTable | 7 277 ns / 14 402 B | 7 061 ns / 10 557 B | −3% time, −27% bytes |
| StreamingFastRenderer | 952 ms / 1.10 GB | 514 ms / 919 MB | **−46%** time, −16% bytes |

## Validation

`task lint`, `task test`, `task build` — all green.

### DIFF
--- a/pkg/tui/components/markdown/fast_renderer.go
+++ b/pkg/tui/components/markdown/fast_renderer.go
@@ -196,7 +196,9 @@ func getGlobalStyles() *cachedStyles {
 		// Heading prefixes used both for display and width math. H1 deliberately
 		// uses the same "## " prefix as H2 so a single hash in the source does not
 		// dominate the TUI with an oversized banner heading.
-		headingPrefixes := [6]string{"## ", "## ", "### ", "#### ", "##### ", "###### "}
+		// Note: H1 uses "# " (single hash) for proper visual hierarchy and correct
+		// continuation-indent width (2 spaces instead of 3).
+		headingPrefixes := [6]string{"# ", "## ", "### ", "#### ", "##### ", "###### "}
 		ansiHeadings := [6]ansiStyle{
 			buildAnsiStyle(headingLipStyles[0]),
 			buildAnsiStyle(headingLipStyles[1]),
@@ -2530,6 +2532,11 @@ func (p *parser) wrapText(text string, width int) string {
 		}
 		word := text[wordStart:i]
 
+		// Fold this word's ANSI codes into the active set BEFORE handling it.
+		// For a long word that gets split, this ensures the continuation-line
+		// breaks below close and re-open any styles that were opened inside the
+		// word, instead of replaying only the styles active before it.
+		active = updateActiveStyles(active, wordAnsi)
 		// Layout decision: do we need to wrap before this word?
 		haveLine := lineWidth > 0
 		needsWrap := haveLine && lineWidth+1+wordWidth > width
@@ -2543,12 +2550,6 @@ func (p *parser) wrapText(text string, width int) string {
 			out.WriteByte(' ')
 			lineWidth++
 		}
-
-		// Fold this word's ANSI codes into the active set BEFORE handling it.
-		// For a long word that gets split, this ensures the continuation-line
-		// breaks below close and re-open any styles that were opened inside the
-		// word, instead of replaying only the styles active before it.
-		active = updateActiveStyles(active, wordAnsi)
 
 		if wordWidth > width {
 			// Long word: break it into pieces, separated by line breaks that close
--- a/pkg/tui/components/markdown/fast_renderer.go
+++ b/pkg/tui/components/markdown/fast_renderer.go
@@ -193,6 +193,9 @@ func getGlobalStyles() *cachedStyles {
 		// Build blockquote lipgloss style
 		blockquoteLipStyle := buildStylePrimitive(mdStyle.BlockQuote.StylePrimitive)
 
+		// Heading prefixes used both for display and width math. H1 deliberately
+		// uses the same "## " prefix as H2 so a single hash in the source does not
+		// dominate the TUI with an oversized banner heading.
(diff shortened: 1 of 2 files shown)


pkg/tui/components/markdown/fast_renderer.go
```diff
@@ -2544,217 +2484,107 @@ func (p *parser) wrapText(text string, width int) string {
...
+			broken := breakWord(word, width)
+			for j, part := range broken {
+				if j > 0 {
+					writeLineBreak(&out, active)
+				}
+				out.WriteString(part)
+			}
+			// Always end a broken-word run with a line break so the following
+			// word starts on a fresh line.
+			writeLineBreak(&out, active)
```
docker-agent, line 2566: **[MEDIUM] `wrapText`: ANSI styles starting within a long broken-word are not restored on continuation lines**

In the new `wrapText` implementation, `wordAnsi` accumulates ANSI escape sequences found inside each word, but `updateActiveStyles` is called **after** the word is emitted (line 2564). For ordinary words this is fine—the next wrap uses the updated `active`. However for a long word that gets split by `breakWord`, the `writeLineBreak` calls between pieces (lines 2550 and 2556) replay `active` as it was **before** this word's codes were folded in. If a CSI style (e.g. bold, colour) begins inside the broken word, it will be absent from `active`, so `writeLineBreak` will not re-open it on the continuation lines.

**Trigger path**: any text where a single word longer than the wrap width itself contains an opening ANSI style code (e.g. a long URL wrapped in a colour sequence).

**Fix**: call `active = updateActiveStyles(active, wordAnsi)` before the `if wordWidth > width { … }` branch, so the `active` slice is up-to-date when `breakWord` pieces are rejoined with `writeLineBreak`.

pkg/tui/components/markdown/fast_renderer.go
```diff
@@ -192,49 +193,61 @@ func getGlobalStyles() *cachedStyles {
...
-			ansiFootnote:     buildAnsiStyle(lipgloss.NewStyle().Foreground(styles.TextSecondary).Italic(true)),
-			styleTaskTicked:  mdStyle.Task.Ticked,
-			styleTaskUntick:  mdStyle.Task.Unticked,
-			listIndent:       int(mdStyle.List.LevelIndent),
-			blockquoteIndent: 1,
-			chromaStyle:      styles.ChromaStyle(),
-		}
-		for i := range globalStyles.ansiHeadings {
-			globalStyles.ansiHeadings[i].hasBold = true
+		headingPrefixes := [6]string{"## ", "## ", "### ", "#### ", "##### ", "###### "}
```
docker-agent: **[LIKELY] `headingPrefixes[0]` uses `"## "` (H2 prefix) instead of `"# "` for H1**

```go
headingPrefixes := [6]string{"## ", "## ", "### ", "#### ", "##### ", "###### "}
//                             ^^^^ index 0 = H1, but prefix is two hashes
```

Index 0 is H1 and should be `"# "` (single hash + space). As written, H1 headings render with the `## ` prefix identical to H2, and the continuation-indent for H1 is 3 spaces instead of 2 (`spaces(len("## ")) == "   "`).

The downstream code uses these strings for both **display** (`styledHeadingPrefixes[level-1]`) and **width arithmetic** (`styledHeadingContIndent[level-1]`), so the off-by-one carries through to wrapped heading text.

If this is intentional (e.g. H1 is deliberately suppressed to match H2 visually in TUI contexts), a short comment would clarify it and remove the ambiguity.

docker-agent (commented):

Two medium-severity findings in the changed code. Both are in  lines introduced by this PR.

gtardif (approved):

PR #2687, opened 2026-05-07
refactor(lint): adopt new rubocop-go DSL across all cops

dgageot:
Adopts the new `rubocop-go` DSL across all 12 custom cops in `lint/`. Behaviour is unchanged — same offenses, same anchors, same severities — but the cops read better and a lot of plumbing moved into the library.

## What changed

- **`cop.Func` literal style everywhere.** Each cop is now a single `&cop.Func{Meta, Scope, Run}` value. The `type CopX struct { cop.Meta }` + `NewCopX()` constructor + `Check` method triplet is gone.
- **Declarative scope.** File-scope guards (`if !p.FileMatches(...) { return }`) are gone from `Check` bodies; cops declare their scope as data and the runner skips out-of-scope files entirely:

  ```go
  Scope: cop.UnderDir("pkg/tui"),
  Scope: cop.OnlyFile("pkg/runtime/event.go"),
  Scope: cop.And(cop.InPathSegment("pkg/config", isLatest), cop.NotBlackBoxTest()),
  ```

- **Library helpers absorbed boilerplate.** Cops now use:
  - `Pass.ReportMissing(anchor, fmt, names)` — replaces the sort + join + empty-check pattern in 5 dispatch-table cops
  - `Pass.PointerReceiverMethods(name)`, `Pass.StructType(name)`, `Pass.FirstMethodCall(name)`, `Pass.FuncDecl(name)` — direct lookups instead of hand-rolled `ForEach...` filter loops
  - `cop.WalkFuncWithContextScope` — the ~120 LOC of context-scope analysis from `slog_contextual.go` lifted into the library
  - `cop.StringField`, `cop.ParseTagOptions`, `cop.FieldNames`, `cop.FieldTag`, `cop.IsNullaryFunc`, `cop.CallTo` — small helpers for shapes that recurred across cops
- **Dropped the regex in `configpath.go`** in favour of a one-line `LastIndex` + delegation to `versionFromDir`, removing a drift point between the two helpers.
- **`main.go`'s registry** is now a flat `var cops = []cop.Cop{...}` of literal vars instead of a `[]cop.Cop{NewX(), ...}` builder.

## Numbers

|                                     |  Before |              After |
| ----------------------------------- | ------: | -----------------: |
| `lint/` LOC                         |    1390 |      1151 (−17%)   |
| Project-private helper functions    |     ~20 | ~12 (8 absorbed into the library) |

## Validation

- `task build`, `task test`, `task lint` — all green
- Negative-tested `TUIViewPurity`, `SlogContextual` (parameter-declared / nested funclit / discarded multi-return), and `RuntimeSessionScoped` — same offenses, same anchors, same wording as before

## Note for reviewers

The cops report 4 pre-existing offenses on current `main` (in `pkg/attachment/modelcaps/modelcaps.go`, `pkg/model/provider/anthropic/federation/federation.go`, `pkg/model/provider/openai/client.go`). They exist on `main` with the original cops too — verified by checking out the pre-refactor `lint/` and re-running. They are real `Lint/SlogContextual` violations that landed during the time this branch was stale and are not introduced by this refactor; happy to fix in a separate PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -154,7 +154,7 @@ require (
 	github.com/containerd/stargz-snapshotter/estargz v0.18.2 // indirect
 	github.com/cyphar/filepath-securejoin v0.4.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/dgageot/rubocop-go v0.0.0-20260429125723-198995cc80c9
+	github.com/dgageot/rubocop-go v0.0.0-20260507084512-2695e6771458
 	github.com/distribution/reference v0.6.0 // indirect
 	github.com/dlclark/regexp2 v1.12.0 // indirect
 	github.com/docker/distribution v2.8.3+incompatible // indirect
--- a/go.sum
+++ b/go.sum
@@ -193,8 +193,8 @@ github.com/danieljoos/wincred v1.2.2/go.mod h1:w7w4Utbrz8lqeMbDAK0lkNJUv5sAOkFi7
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
-github.com/dgageot/rubocop-go v0.0.0-20260429125723-198995cc80c9 h1:LwBQXSuGbtHC3rzVOBQIsSofzTc7k9QWRqoQdfPbG+s=
-github.com/dgageot/rubocop-go v0.0.0-20260429125723-198995cc80c9/go.mod h1:r8YOJV5+/30NZ8HW/2NbWUObBGDXGvfHrjgury5YlFI=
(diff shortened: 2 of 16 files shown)


docker-agent (commented):
## Review: Lint Cop Refactor — `rubocop-go` DSL Adoption

### Assessment: 🟢 APPROVE

This is a clean, well-structured refactor. All 12 cops were ported to the `cop.Func` literal style with correct `Scope` declarations, and the behavioral equivalence holds throughout. No regressions found.

**Coverage summary:**

| Area | Result |
|---|---|
| `versionFromImport` regex → `LastIndex` replacement (`configpath.go`) | ✅ Equivalent — `versionFromDir` rejects non-`vN` suffixes identically |
| `ReportMissing` called without empty-slice guard | ✅ Safe — library guards on `len(names) == 0` internally |
| `returnedEventType` unary-op check removed | ✅ Safe — `*ast.UnaryExpr` on a `CompositeLit` can only be `&` in valid Go |
| `WalkFuncWithContextScope` replacing hand-rolled context analysis | ✅ Library handles `body == nil` guard; behavior matches |
| `cop.ParseTagOptions` + `HasName()` replacing manual `strings.Cut` + `"-"` check | ✅ `HasName()` correctly rejects `""` and `"-"` |
| Declarative `Scope` fields vs. in-body guards | ✅ All scopes (paths, predicates, `NotBlackBoxTest`) faithfully match old guards |
| `p.PointerReceiverMethods` replacing hand-rolled `pointerReceiversWithMethod` | ✅ Identical semantics per library source |
| `exportedBuiltinNames` sort removed | ✅ `ReportMissing` sorts its input before joining — message still deterministic |

gtardif (approved):

PR #2688, opened 2026-05-07
freeze config v8 and start v9 as latest

dgageot:
Freezes the previous `pkg/config/latest` package as `pkg/config/v8` and bumps `latest` to schema version 9.

## Changes

- **New `pkg/config/v8/`**: copied all sources/tests from `pkg/config/latest/`, renamed the package to `v8`/`v8_test`. The `Version = "8"` constant and `previous "…/v7"` upgrade import are preserved.
- **`pkg/config/latest/`**:
  - `Version` constant bumped from `"8"` → `"9"`.
  - `parse.go` `previous` import switched from v7 → v8.
  - `validate_test.go` inline `version: "8"` fixtures updated to `"9"`.
  - `hooks_yaml_test.go` doc comment updated to mention v9.
- **`pkg/config/versions.go`**: registers the new `v8` package between `v7` and `latest`.

Also includes a small follow-up that switches a handful of `slog.Warn` calls to `slog.WarnContext` where a `ctx` is already in scope, addressing the existing `Lint/SlogContextual` cop warnings (unrelated to the freeze, but surfaced by `task lint`).

### DIFF
--- a/pkg/attachment/modelcaps/modelcaps.go
+++ b/pkg/attachment/modelcaps/modelcaps.go
@@ -95,7 +95,7 @@ func Load(modelID string) (ModelCapabilities, error) {
 
 	store, err := modelsdev.NewStore()
 	if err != nil {
-		slog.Warn("modelcaps: failed to load models.dev store, using conservative caps",
+		slog.WarnContext(ctx, "modelcaps: failed to load models.dev store, using conservative caps",
 			"error", err, "model", modelID)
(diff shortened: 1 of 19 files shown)


rumpl (approved):
Should we throw a party when we get to 10?

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25489073641).

PR #2690, opened 2026-05-07
feat(tui): add /snapshots command to list and restore captured snapshots

dgageot:
Adds a new `/snapshots` slash command that opens a dialog showing every snapshot captured during the current session and lets the user restore any of them (or the original pre-agent state).

## What it does

- **`/snapshots`** opens a modal listing the captured snapshots with their file counts, plus an `<original>` entry for the pre-agent state.
- Inside the dialog: <kbd>↑</kbd>/<kbd>↓</kbd> (or <kbd>j</kbd>/<kbd>k</kbd>, <kbd>g</kbd>/<kbd>G</kbd>, <kbd>home</kbd>/<kbd>end</kbd>) navigate, <kbd>r</kbd> restores the highlighted snapshot, <kbd>Esc</kbd> closes without changing anything.
- Restoring snapshot _N_ reverts every snapshot taken after _N_ (i.e. workspace returns to the state at the end of snapshot _N_). Restoring `<original>` reverts everything.
- The command (and `/undo`) are hidden from both the palette and the slash-command parser when snapshots aren't enabled, so the UI stays uncluttered for users who haven't opted in.

## Implementation

- Adds `ListSnapshots` and `ResetSnapshot` operations alongside the existing `UndoLastSnapshot`, plumbed through `pkg/hooks/builtins`, `pkg/runtime`, and `pkg/app`.
- A single `snapshotRuntime` interface in `pkg/app` captures all four runtime methods (`SnapshotsEnabled`, undo, list, reset); runtimes that don't support snapshots simply don't implement it and the related commands disappear automatically.
- `popHistoryTail` (renamed from `truncateAfter`) clears popped slots in the backing array so the dropped file lists can be GC'd.
- The dialog itself stays small: no key-map struct, no second pass over the renderer, the body and help-keys derive from `len(fileCounts)`.

## Docs & tests

- `docs/features/tui/index.md` updated.
- New tests cover `ListSnapshots`, `ResetSnapshot`, the no-op cases, the disabled-snapshots filtering of `/undo` and `/snapshots`, and that `App.SnapshotsEnabled()` answers the runtime-capability question without needing an active session.

### DIFF
--- a/docs/features/tui/index.md
+++ b/docs/features/tui/index.md
@@ -41,7 +41,7 @@ Type `/` during a session to see available commands, or press <kbd>Ctrl</kbd>+<k
 | `/copy`            | Copy the entire conversation to clipboard                                            |
 | `/copy-last`       | Copy only the last assistant message to clipboard                                    |
 | `/undo`            | Restore file changes from the latest snapshot (only when snapshots are enabled)      |
-| `/snapshot`        | List captured snapshots and reset to one (only when snapshots are enabled)           |
+| `/snapshots`       | List captured snapshots (only when snapshots are enabled)                            |
 | `/export`          | Export the session as HTML                                                           |
 | `/sessions`        | Browse and load past sessions                                                        |
 | `/model`           | Change the model for the current agent                                               |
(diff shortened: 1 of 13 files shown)


rumpl (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25489714040).

PR #2691, opened 2026-05-07
fix: retry transient Vertex AI 'function response parts' 400 errors

### LINKED ISSUE #2683: Vertex AI Gemini: 'number of function response parts' INVALID_ARGUMENT during long transfer_task chains
## Description

In a multi-agent run with 21 sub-agents and an orchestrator that does ~30 sequential `transfer_task` calls, we observed the following Vertex AI Gemini error 4 times during a single run:

```
❌ model failed: error receiving from stream: Error 400, Message: Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn., Status: INVALID_ARGUMENT, Details: []
```

The first 3 occurrences were **silently and successfully auto-retried** by docker-agent (the same `transfer_task` was re-issued to the same agent immediately and the second attempt succeeded). The 4th occurrence happened at session end (after the orchestrator had already produced its final summary and all artifacts were written) and **was not retried** — `docker agent run` exited non-zero, propagating to `make` as exit 1.

## Reproduction

- docker-agent: **v1.54.0** (commit df10bd02)
- Model: `gemini-3-flash-preview` via Vertex AI (provider=google, project=***, location=global)
- Pipeline: 21 sub-agents, orchestrator runs them serially via `transfer_task` (no parallel calls)
- Run id: `20260507-054413-22763` (this is local, not public)
- Wall-clock: ~1h45m, ~30 transfer_task calls + many filesystem / shell tool calls inside each sub-agent

The errors are non-deterministic — they occurred at iteration 1 dep-cve-fixer transfer, iteration 2 fixer transfer, iteration 2 fixer transfer (different message), and at session end after the final summary.

## Expected behavior

Either:
1. The 'function response parts' protocol bug at Vertex level shouldn't surface to the user (auto-retry handles it transparently — which it already does mid-session, but not at session-end)
2. Or at minimum, the session-end occurrence should be caught with the same auto-retry mechanism, so a successful run doesn't end with a non-zero exit code that breaks CI integrations.

## Repo / agent

The agent in question is a 21-sub-agent dev pipeline:
https://github.com/kitadesign/kitazawa-docker-agent-pipeline

Specifically `agent.yaml`'s `root` agent invokes spec-analyst → architect → planner → ... → retro-moderator → proposer in sequence. Each transfer_task creates a fresh sub-session. After ~30 such calls the parent session's function-call history may be where Vertex's 'parts' counter desyncs.

## Workarounds we're using

- For mid-run: nothing needed, docker-agent's existing auto-retry covers it.
- For end-of-run: we currently treat exit 2 as success if `<RUN_DIR>/artifacts/retrospective.md` exists, but this is fragile.

Happy to provide a redacted log excerpt if helpful.

dgageot:
Fixes #2683.

Vertex AI / Gemini intermittently returns HTTP 400 `INVALID_ARGUMENT` with the message _"Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn."_ even for well-formed requests. The same payload always succeeds on retry — it's a server-side transient.

Today `ClassifyModelError` treats every 400 as non-retryable. Mid-session occurrences look like they auto-recover only because the error happens inside a `transfer_task` child session, gets translated to a tool error, and the orchestrator's LLM decides to re-issue the call against a fresh sub-session. End-of-session, there's no surrounding tool dispatch, so the same error becomes fatal and the run exits non-zero — breaking CI.

This change adds a small `transientStatusCodePatterns` list and a `matchesTransientPattern` helper. `ClassifyModelError` and `isRetryableModelError` consult it whenever the HTTP status code would otherwise classify the error as non-retryable, so the existing retry-with-backoff loop in `fallbackExecutor.execute` handles the 400 the same way it handles 5xx — uniformly in mid-session and end-of-session contexts. Bounded by `DefaultRetries=2` (3 attempts total) with exponential backoff.

The pattern is intentionally narrow (`"number of function response parts"`) so unrelated 400s remain non-retryable.

### Validation
- `go test ./pkg/modelerrors/...` — new cases cover the typed `*StatusError` path, the wrapped `error receiving from stream: %w` path, and the message-fallback path.
- `task test` — full suite green (128 packages).
- `golangci-lint run ./pkg/modelerrors/...` — 0 issues.

### DIFF
--- a/pkg/modelerrors/modelerrors.go
+++ b/pkg/modelerrors/modelerrors.go
@@ -217,6 +217,38 @@ func isRetryableStatusCode(statusCode int) bool {
 	}
 }
 
+// transientStatusCodePatterns contains error message substrings that indicate
+// a transient failure even when the HTTP status code would otherwise classify
+// the error as non-retryable (e.g. 4xx). Patterns are matched
+// case-insensitively against the full formatted error.
+//
+// Currently:
+//   - Vertex AI / Gemini intermittently returns 400 INVALID_ARGUMENT with the
+//     message "Please ensure that the number of function response parts is
+//     equal to the number of function call parts of the function call turn."
+//     even for well-formed requests; the same payload succeeds on retry.
+//     Without this override the run would surface a fatal error to the user
+//     and exit non-zero, see https://github.com/docker/docker-agent/issues/2683.
+var transientStatusCodePatterns = []string{
+	"number of function response parts",
+}
+
+// matchesTransientPattern reports whether the error's message matches one of
+// [transientStatusCodePatterns]. Used to override an otherwise non-retryable
+// HTTP status classification.
+func matchesTransientPattern(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	for _, p := range transientStatusCodePatterns {
+		if strings.Contains(msg, p) {
+			return true
+		}
+	}
+	return false
+}
+
 // retryablePatterns contains error message substrings that indicate a
 // transient/retryable failure. Numeric status codes (500, 502, etc.) are
 // handled separately by extractHTTPStatusCode + isRetryableStatusCode.
@@ -293,7 +325,7 @@ func isRetryableModelError(err error) bool {
 
 	// First, try to extract HTTP status code from known SDK error types
 	if statusCode := extractHTTPStatusCode(err); statusCode != 0 {
-		retryable := isRetryableStatusCode(statusCode)
+		retryable := isRetryableStatusCode(statusCode) || matchesTransientPattern(err)
 		slog.Debug("Classified error by status code",
 			"status_code", statusCode,
 			"retryable", retryable)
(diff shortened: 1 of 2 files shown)


pkg/modelerrors/modelerrors.go
```diff
@@ -217,6 +217,38 @@ func isRetryableStatusCode(statusCode int) bool {
...
+// matchesTransientPattern reports whether the error's message matches one of
+// [transientStatusCodePatterns]. Used to override an otherwise non-retryable
+// HTTP status classification.
+func matchesTransientPattern(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	for _, p := range transientStatusCodePatterns {
+		if strings.Contains(msg, p) {
```
docker-agent: **[MEDIUM] `matchesTransientPattern` lowercases the message but not the pattern — mixed-case patterns silently never match**

`msg` is produced via `strings.ToLower(err.Error())`, so it is always lowercase. However, the loop compares `msg` against the raw pattern `p` without lowercasing it:

```go
msg := strings.ToLower(err.Error())
for _, p := range transientStatusCodePatterns {
    if strings.Contains(msg, p) {   // p is NOT lowercased
```

The current pattern `"number of function response parts"` happens to be all-lowercase, so it works today. But the invariant is not enforced: a future maintainer adding a mixed-case pattern (e.g. `"Function Response Parts"`) would introduce a silently broken retry rule — the pattern would never match and the transient 400 would again be treated as fatal, exactly the regression this PR is fixing.

The fix is straightforward — lowercase the pattern in the comparison:
```go
if strings.Contains(msg, strings.ToLower(p)) {
```
or pre-lowercase patterns at declaration time and document the requirement.

Note: the existing `retryablePatterns` code in the same file uses pre-lowercased constants and avoids this asymmetry.

docker-agent (commented):

One medium-severity latent bug found in the newly added `matchesTransientPattern` helper. The fix logic is correct and the current pattern works, but the case-folding asymmetry is a maintenance trap.

gtardif (approved):

PR #2693, opened 2026-05-07
move non-migration config tests to pkg/config

dgageot:
Cleans up tests under `pkg/config` so that:

- version packages (`v0`..`v8`) only hold tests that verify migration-specific code
- the `latest` package follows the same rule
- tests that aren't migration-specific live directly under `pkg/config` and exercise configs without a `version:` field in YAML

In practice the version and `latest` packages didn't actually carry migration-specific tests, so all of those test files were removed and the equivalent (de-duplicated) tests now live in `pkg/config`.

Three previously-unexported methods on `pkg/config/latest` were promoted to public so the moved external tests can call them:

- `(*Config).Validate()`
- `(*AuthConfig).Validate(provider string)`
- `(*HooksConfig).Validate()`

The old `ValidateHooksConfigForTest` helper in `pkg/config/latest/export_test.go` was deleted.

Net effect: ~6k lines of duplicated test code removed, single source of truth for the latest schema's behavioral tests.


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25491602821).

gtardif (approved):

PR #2694, opened 2026-05-07
shrink builtin tool schemas to save tokens

dgageot:
Reduces the size of builtin tool definitions sent to LLM providers on
every turn. Most of the savings come from the filesystem and shell
toolsets, which together represented ~1.2k tokens per request.

Changes:
- `SchemaToMap` now strips `"null"` from the type of required fields.
  jsonschema-go emits `["null", "array"]` for every Go slice (including
  required ones), which is contradictory and only inflates the schema.
  OpenAI's strict mode still adds `null` for fields it promotes from
  optional to required, so behaviour is preserved there.
- Shortened the description text added by `AddDescriptionParameter`.
- Trimmed verbose `jsonschema:"..."` tags on filesystem and shell tool
  arg structs.
- Drive-by: switched a few `slog.Warn` calls to `slog.WarnContext` in
  spots where a context was already in scope (custom linter caught it).

Cassettes were updated to match the new request bodies. No behaviour
change for callers.


docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed the schema null-stripping logic, description trimming, struct tag removals, and `slog.WarnContext` drive-by changes.

**Key areas checked:**
- **`SchemaToMap` / `removeNullFromType`**: The `filtered := arr[:0]` in-place filter pattern is provably safe — Go's range captures the slice header at loop start, so reads always precede writes on the same backing-array positions. No correctness bug.
- **`stripNullFromRequiredTypes` recursion**: Correctly gates `removeNullFromType` on the `required` set; recursive descent into nested `properties` and `items` is well-scoped.
- **Filesystem/shell struct tag removals**: Tags were description-only (`jsonschema:"description=..."`); JSON serialization keys (`json:"..."`) are untouched. No serialization breakage.
- **`AddDescriptionParameter` description text**: Shortened text is functionally equivalent — no logic change.
- **`slog.WarnContext` changes**: Each call site already had a `ctx` in scope; the substitution is correct.
- **Cassette updates**: Updated request bodies match the new schema shapes for all five providers (OpenAI strict mode, Anthropic, Gemini, Mistral). The type narrowing from `["null","array"]` → `"array"` for required fields aligns with the PR's stated intent and preserves strict-mode compatibility.

gtardif (approved):

PR #2697, opened 2026-05-07
secretsscan: tighten Docker PAT, add new vendor patterns, cap quantifiers

dgageot:
Expands the secret-redaction catalogue and fixes a class of bugs in the rules added along the way.

### Docker PAT/OAT
- Tightened the existing `dckr_pat_` rule: dropped `(?i)` and the hyphen from the body class — Docker's spec is strictly alphanumeric.
- Added a new rule for Docker Hub Organization Access Tokens (`dckr_oat_`).

### New vendor-prefixed patterns
Confirmed against gitleaks default rules and vendor docs:

| Vendor | Prefix |
|---|---|
| 1Password service account | `ops_eyJ` |
| OpenRouter | `sk-or-v1-` |
| SonarQube/Cloud | `squ_` / `sqp_` / `sqa_` |
| Pinecone | `pckey_` |
| Supabase service-role | `sb_secret_` |
| Tailscale auth + API tokens | `tskey-auth-` / `tskey-api-` |
| Vercel personal/CLI/integration | `vcp_` / `vck_` / `vci_` |
| Razorpay | `rzp_test_` / `rzp_live_` |
| Adyen | `AQE` |
| Plaid access token | `access-(sandbox\|development\|production)-` |
| PostHog personal API | `phx_` |
| Render API | `rnd_` |
| Honeycomb ingest + config | `hcaik_` / `hcaic_` |
| Akamai EdgeGrid | `akab-` |
| Adafruit IO | `aio_` |

### Bug fix: capped greedy quantifiers
The first pass of new rules used unbounded `{N,}` quantifiers. With Go's RE2 those match greedily and silently absorb adjacent alphanumeric text into the redaction span — `vcp_aaa…aaaTRAILINGTEXT` would collapse to a single `[REDACTED]`, dropping the trailing context. Same class of bug the existing `TestRedactDoesNotSwallowAdjacentTextAfterSlackRotatingToken` was added to prevent.

Every new rule now has an explicit upper bound (e.g. `{20,80}` for Vercel, `{250,1000}` for 1Password — RE2's per-quantifier ceiling), and there is a regression test (`TestRedactDoesNotSwallowAdjacentTextAfterPrefixedTokens`) that pushes each body past its cap and asserts the trailing text survives.

### DIFF
--- a/pkg/secretsscan/rules.go
+++ b/pkg/secretsscan/rules.go
@@ -850,8 +850,13 @@ var rules = sync.OnceValue(func() []rule {
 			// always start with `ops_eyJ` — the `eyJ` is the base64
 			// prefix of `{"`, since the body is a JWT-style envelope
 			// over a 1Password macaroon. The literal `ops_eyJ` keyword
-			// keeps the AC pre-filter extremely selective.
-			expression: `ops_eyJ[A-Za-z0-9+/=_-]{250,}`,
+			// keeps the AC pre-filter extremely selective. The 1000-char
+			// upper bound covers the longest 1Password tokens observed
+			// in the wild (and is RE2's hard cap on a single quantifier)
+			// while preventing the regex from absorbing arbitrary
+			// trailing alphanumeric content if a token is not
+			// whitespace-terminated.
+			expression: `ops_eyJ[A-Za-z0-9+/=_-]{250,1000}`,
 			keywords:   []string{"ops_eyJ"},
 		},
 		{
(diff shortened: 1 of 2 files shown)


gtardif (approved):

pkg/secretsscan/rules.go
```diff
@@ -827,6 +838,167 @@ var rules = sync.OnceValue(func() []rule {
...
+			// trailing alphanumeric content if a token is not
+			// whitespace-terminated.
+			expression: `ops_eyJ[A-Za-z0-9+/=_-]{250,1000}`,
+			keywords:   []string{"ops_eyJ"},
+		},
+		{
+			// openrouter-api-key. OpenRouter (LLM router) keys carry
+			// the documented `sk-or-v1-` prefix followed by a 64-char
+			// lowercase-hex body.
+			expression: `sk-or-v1-[a-f0-9]{64}`,
```
docker-agent, line 866: **[MEDIUM] OpenRouter key body pattern `[a-f0-9]{64}` is too narrow — likely causes false negatives**

The comment says "64-char lowercase-hex body," but OpenRouter API keys in practice use a base64url-encoded body, not pure hex. Gitleaks' upstream rule for OpenRouter uses `[a-zA-Z0-9]{64}` (full alphanumeric) rather than `[a-f0-9]{64}`.

Any real key whose body contains `g–z` or `A–Z` would silently bypass redaction with the current pattern.

The test only repeats `"a"` 64 times — all within the `[a-f]` set — so it cannot detect this mismatch.

**Suggested fix:**
```go
expression: `sk-or-v1-[a-zA-Z0-9]{64}`,
```

docker-agent (commented):

2 findings (1 LIKELY · 1 CONFIRMED) in newly-added code.

pkg/secretsscan/secrets_test.go
```diff
@@ -159,6 +186,62 @@ func TestRedactDoesNotSwallowAdjacentTextAfterSlackRotatingToken(t *testing.T) {
...
+		{"render_api_key", "rnd_" + strings.Repeat("e", 90)},
+		// 1Password body cap 1000 — use 1100 chars.
+		{"onepassword_service_account", "ops_eyJ" + strings.Repeat("f", 1100)},
+		// Pinecone second segment cap 80 — use 90 chars.
+		{"pinecone_api_key", "pckey_label_" + strings.Repeat("g", 90)},
+		// Tailscale second segment cap 80 — use 90 chars.
+		{"tailscale_auth_key", "tskey-auth-" + strings.Repeat("h", 12) + "-" + strings.Repeat("i", 90)},
+		{"tailscale_api_token", "tskey-api-" + strings.Repeat("j", 12) + "-" + strings.Repeat("k", 90)},
+	}
+	const suffix = " and the rest of the line"
```
docker-agent, line 229: **[MEDIUM] Quantifier-cap test uses a space-prefixed suffix — the upper bound is never actually validated**

`TestRedactDoesNotSwallowAdjacentTextAfterPrefixedTokens` is supposed to prove that an explicit upper bound (e.g. `{20,80}`) prevents the regex from swallowing adjacent alphanumeric text. But the suffix is `" and the rest of the line"` — it starts with a **space**.

Because every body character class in the tested rules is alphanumeric (`[A-Za-z0-9]`, `[A-Za-z0-9_-]`, etc.), a space already terminates the match even with an **unbounded** quantifier (`{N,}`). The test would pass identically if every `{N,M}` upper bound were replaced with an open-ended `{N,}`, so it gives no confidence that the caps are real.

To actually exercise the cap, the suffix must consist of characters that the body class *would* consume — e.g.:

```go
const suffix = strings.Repeat("Z", 200) // alphanumeric, consumed by an unbounded rule
```

Then the assert that `suffix` survives would only pass when the quantifier ceiling genuinely blocks it.

PR #2704, opened 2026-05-07
fix: expand ~ in filesystem tool paths

### LINKED ISSUE #2696: Reading ~/file fails
The filesystem tool doesn't recognise the `~/xxx` syntax. It should expand it to the user's home folder

dgageot:
Fixes #2696.

`Tool.resolvePath` joined LLM-supplied paths with the working directory without expanding a leading `~`, so calls like `read_file("~/file.txt")` looked up `<workingDir>/~/file.txt` and failed with "not found". The allow/deny list config already expanded `~` (via `expandPathToken`); this brings runtime path resolution in line.

- Added a small `expandHomeDir` helper that replaces a leading `~` or `~/` (or `~\` on Windows) with `os.UserHomeDir()`, leaving everything else (including bare `~name`) untouched.
- `resolvePath` calls it before its absolute/relative branching, so every filesystem handler benefits (read, write, edit, list, search, mkdir, rmdir, tree, read_multiple).
- Regression tests cover `~`, `~/file`, `~/a/b/c`, the literal `~name` case, and an end-to-end `read_file("~/note.txt")`.

### DIFF
--- a/pkg/tools/builtin/filesystem/filesystem_test.go
+++ b/pkg/tools/builtin/filesystem/filesystem_test.go
@@ -67,6 +67,45 @@ func TestFilesystemTool_ResolvePath(t *testing.T) {
 	assert.Equal(t, "/etc/hosts", resolvedPath)
 }
 
+// TestFilesystemTool_ResolvePath_ExpandsTilde is a regression test for
+// issue #2696: paths starting with "~" or "~/" must be expanded to the
+// user's home directory before being used, otherwise read_file (and every
+// other filesystem handler) treats them as a literal subdirectory of the
+// working directory and fails with "not found".
+func TestFilesystemTool_ResolvePath_ExpandsTilde(t *testing.T) {
+	homeDir := t.TempDir()
+	resetHomeDir(t, homeDir)
+	wd := t.TempDir()
+	tool := NewFilesystemTool(wd)
+
+	assert.Equal(t, homeDir, tool.resolvePath("~"))
+	assert.Equal(t, filepath.Join(homeDir, "file.txt"), tool.resolvePath("~/file.txt"))
+	assert.Equal(t, filepath.Join(homeDir, "a", "b", "c.txt"), tool.resolvePath("~/a/b/c.txt"))
+
+	// A bare "~name" (no separator) is not the home dir: keep it as a
+	// literal subdirectory of the working dir so the user can still
+	// reference a file/dir whose name happens to start with "~".
+	assert.Equal(t, filepath.Join(wd, "~name"), tool.resolvePath("~name"))
+}
+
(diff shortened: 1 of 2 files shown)


derekmisler (approved):

pkg/tools/builtin/filesystem/filesystem_test.go
```diff
@@ -67,6 +67,45 @@ func TestFilesystemTool_ResolvePath(t *testing.T) {
...
+// other filesystem handler) treats them as a literal subdirectory of the
+// working directory and fails with "not found".
+func TestFilesystemTool_ResolvePath_ExpandsTilde(t *testing.T) {
+	homeDir := t.TempDir()
+	resetHomeDir(t, homeDir)
+	wd := t.TempDir()
+	tool := NewFilesystemTool(wd)
+
+	assert.Equal(t, homeDir, tool.resolvePath("~"))
+	assert.Equal(t, filepath.Join(homeDir, "file.txt"), tool.resolvePath("~/file.txt"))
```
docker-agent, line 82: **[LOW] Missing test coverage for Windows-style `~\file` tilde expansion**

`expandHomeDir` has a dedicated code path for Windows backslash separators:

```go
!strings.HasPrefix(path, "~"+string(filepath.Separator))
```

On Windows, `filepath.Separator` is `\`, making this branch `~\file`. The added tests use the hard-coded literal `~/file.txt`, which only exercises the forward-slash prefix on all platforms — the backslash variant is never exercised even when running on Windows.

Consider adding a test case using `filepath.Join("~", "file.txt")` (which produces `~\file.txt` on Windows and `~/file.txt` on Unix) to give cross-platform confidence that both separator branches work.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The `expandHomeDir` helper is correctly implemented and well-integrated into `resolvePath`. The fix is clean, minimal, and well-tested. The silent fallback when `os.UserHomeDir()` fails is an intentional, documented design choice. The tilde expansion does not introduce new attack vectors beyond what was already possible (absolute paths / `../` traversals were already unrestricted in the default no-allow-list configuration). One minor test coverage gap noted inline.

aheritier (approved):
LGTM. Minimal, focused fix with proper test coverage:

- Handles `~`, `~/`, and `~\` (Windows) prefixes; bare `~name` correctly stays literal.
- Graceful fallback when `os.UserHomeDir()` fails — surfaces as a "not found" rather than a hard failure inside path resolution.
- Both the unit-level `resolvePath` test and the end-to-end `read_file` regression test for #2696.

Note: tilde expansion happens *before* allow/deny-list checks via `resolveAndCheckPath`, so this doesn't widen the sandbox — good.

PR #2706, opened 2026-05-07
refactor(hooks): make the unload on_agent_switch builtin pure

dgageot:
Refactors the existing `unload` `on_agent_switch` builtin (added in #2684) so the runtime no longer knows about the unload concept. The hook is now a pure, runtime-agnostic HTTP POST driven entirely by data the runtime ships on `hooks.Input`.

## Before

- `BuiltinUnload` was registered as a method on `*LocalRuntime`.
- The runtime imported `provider.Unloader` and called `Unload(ctx)` on every model implementing it.
- DMR's `Client.Unload` lived next to its other HTTP plumbing.

## After

- The runtime ships a snapshot of the previous agent's model endpoints on every `on_agent_switch` dispatch via a new `Input.FromAgentModels []ModelEndpoint` field (`{provider, model, base_url, unload_api}`).
- The `unload` builtin moves to `pkg/hooks/builtins/unload.go`, depends only on `net/http` + `pkg/hooks`, and POSTs `{"model": "<id>"}` to the resolved `_unload` URL of every DMR endpoint in the snapshot.
- `provider.Unloader`, `dmr.Client.Unload`, `pkg/runtime/unload.go`, and the runtime-side registration are deleted.
- DMR's resolved base URL is exposed via a new `base.Config.BaseURL` field; the private `dmr.Client.baseURL` is folded into the embedded `Config.BaseURL` to remove the drift hazard.

## What this buys us

- The runtime no longer knows the word "unload" — any future "act on the previous agent's models" hook reuses the same `FromAgentModels` snapshot.
- The builtin is testable with just `httptest` (no fake team / no fake `Unloader`).
- External `command` hooks on `on_agent_switch` see the same data, so users can write their own `curl`-based unload (or warm, health-check, …) without us shipping a builtin.
- `task lint` clean; `task test` green; net **−55 lines** vs. the first cut after the simplification commit.

## Trade-offs worth flagging

This is a meaningful architectural shift, not a free win. Reviewers should weigh:

- The DMR-style URL convention (`/v1` → `/_unload`) now lives in `pkg/hooks/builtins`, not next to DMR's sibling `_configure` math.
- Provider-type dispatch is stringly-typed (`m.Provider == "dmr"`) instead of an interface assertion.
- `ModelEndpoint` becomes an external JSON wire contract for any `command` hook on `on_agent_switch`.
- The runtime always builds the snapshot, even when no hook listens; the old design was lazy.

Two small follow-ups would close most of these gaps without reverting:

1. Move `unloadURL` into `pkg/model/provider/dmr` as an exported pure helper; the builtin imports it (and a `dmr.ProviderType` constant) instead of hard-coding the convention and the literal.
2. Build `FromAgentModels` only when `executor.Has(EventOnAgentSwitch)` is true.

Happy to fold either into this PR if reviewers prefer.

## Commits

- `refactor(hooks): make the unload builtin pure, decouple from runtime`
- `refactor(hooks/builtins): tighten the unload builtin` (URL helpers collapsed into one `unloadURL`, `unloadOne` extracted with `defer cancel()`, dead constant + parameter removed, tests deduplicated)

## Validation

- `task lint` — `golangci-lint`: 0 issues; custom 12-cop analyzer: 987 files, no offenses; `go mod tidy` clean.
- `task test` — every package green, including the affected `pkg/hooks/...`, `pkg/runtime/...`, `pkg/model/provider/...`, `pkg/config/latest`.
- New tests pin the runtime → hook handoff (`TestExecuteOnAgentSwitchHooks_PopulatesFromAgentModels`, `…_FromAgentModelsNilWhenFromEmpty`) and the builtin's behaviour end-to-end (`TestUnload_*` against `httptest`).

Builds on / supersedes the runtime-coupled implementation from #2684.


pkg/hooks/builtins/unload.go
```diff
@@ -0,0 +1,131 @@
...
+// runtime captured at dispatch time and POSTs `{"model": "<id>"}` to
+// the resolved unload endpoint of each DMR model. Errors are logged
+// but never propagated — agent switching must never block on a slow
+// or unreachable engine.
+func unload(ctx context.Context, in *hooks.Input, _ []string) (*hooks.Output, error) {
+	if in == nil || in.FromAgent == "" || in.FromAgent == in.ToAgent {
+		return nil, nil
+	}
+	for _, m := range in.FromAgentModels {
+		if m.Provider != "dmr" {
```
aheritier: **Non-blocking**: stringly-typed provider dispatch (`m.Provider != "dmr"`) is fragile. A typo in a future caller, or a future DMR-compatible provider with a different name, will silently no-op.

Worth either:
- exporting a `dmr.ProviderType = "dmr"` constant from `pkg/model/provider/dmr` and importing it here (your follow-up #1), or
- documenting the convention in the `ModelEndpoint.Provider` doc comment so the contract is discoverable.

I'd lean toward the constant — `pkg/hooks/builtins` already imports `pkg/hooks` so the dependency direction stays clean.

pkg/hooks/builtins/unload.go
```diff
@@ -0,0 +1,131 @@
...
+	switch {
+	case m.UnloadAPI == "":
+		u.Path = strings.TrimSuffix(strings.TrimSuffix(u.Path, "/"), "/v1") + "/_unload"
+	case strings.HasPrefix(m.UnloadAPI, "/"):
+		u.Path = m.UnloadAPI
+	default:
+		u.Path = "/" + m.UnloadAPI
+	}
+	return u.String(), nil
+}
```
aheritier: **Non-blocking, but worth folding in**: this function knows DMR's URL convention (`/v1` → `/_unload`) and lives next to neither `dmr.buildConfigureURL` nor any other DMR plumbing. If DMR's convention drifts (e.g. a `/v2` engines path, or a different unload route), there's no obvious place to find this code.

Your own follow-up #1 — exporting `unloadURL` from `pkg/model/provider/dmr` and calling it from here — would restore the cohesion. Combined with the typed `dmr.ProviderType` from the comment above, the builtin becomes a dumb dispatcher and DMR owns its conventions.

Doesn't have to land here, but the diff is small and you've already done most of the thinking. I'd merge with this folded in if it's quick.

pkg/runtime/hooks.go
```diff
@@ -264,9 +264,42 @@ func (r *LocalRuntime) executeOnAgentSwitchHooks(ctx context.Context, a *agent.A
 		FromAgent:       fromAgent,
 		ToAgent:         toAgent,
 		AgentSwitchKind: kind,
+		FromAgentModels: r.fromAgentModels(fromAgent),
```
aheritier, line 298: **Non-blocking**: the snapshot is built on every `on_agent_switch` regardless of whether anything reads it. For a chain with no `on_agent_switch` hook this is a wasted alloc + team lookup per transition.

Your follow-up #2 (gating on `executor.Has(EventOnAgentSwitch)`) is the right call — agent switches aren't hot, but the cheap path stays cheap. Fine to defer to a separate PR.

pkg/hooks/types.go
```diff
@@ -278,6 +287,27 @@ type Input struct {
...
+	// Docker Model Runner). Empty for cloud providers that don't expose
+	// a stable per-instance base URL on the runtime side.
+	BaseURL string `json:"base_url,omitempty"`
+	// UnloadAPI is the per-model unload path or absolute URL copied
+	// verbatim from the model's `unload_api` provider option. Empty
+	// when the user hasn't configured an override; the unload builtin
+	// falls back to a provider-specific default in that case.
+	UnloadAPI string `json:"unload_api,omitempty"`
+}
+
```
aheritier, line 318: **Praise**: nice shape. `ModelEndpoint` as a self-contained, runtime-free wire format makes the on-agent-switch contract genuinely portable — `command` hooks now receive the same data as builtins via JSON, and any future "act on the previous agent" hook reuses this slice without touching runtime internals. The decoupling is the real win of this PR.

pkg/hooks/builtins/unload_test.go
```diff
@@ -0,0 +1,287 @@
...
+func TestUnload_HonoursOverrideUnloadAPI(t *testing.T) {
+	t.Parallel()
+
+	var gotPath string
+	server := countingServer(t, http.StatusOK, func(r *http.Request) { gotPath = r.URL.Path })
+
+	_, err := unload(t.Context(), dmrInput(server, "/custom/unload"), nil)
+	require.NoError(t, err)
+	assert.Equal(t, "/custom/unload", gotPath)
+}
```
aheritier, line 110: **Question / nit**: a test with `FromAgentModels` containing a mix of DMR + non-DMR endpoints would pin the per-element filter (currently only implicitly covered via the all-cloud no-op case). Cheap to add as another `dmrInput`-style helper input; ignore if you think the existing coverage is enough.

pkg/hooks/builtins/unload.go
```diff
@@ -0,0 +1,131 @@
...
+	ctx, cancel := context.WithTimeout(parent, unloadTimeout)
+	defer cancel()
+
+	body, _ := json.Marshal(map[string]string{"model": m.Model})
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, bytes.NewReader(body))
+	if err != nil {
+		return fmt.Errorf("building unload request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
```
aheritier, line 84: **Question**: `http.DefaultClient` here vs. the `httpclient.NewSafeClient` used by the new `http_post` builtin (#2705). The asymmetry is justified — DMR runs on loopback, which the SSRF dialer would block — but it's not obvious to a future reader. Worth a one-liner comment noting *why* the safe client isn't used here (operator-supplied URL, expected localhost target).

aheritier (dismissed):
LGTM. Solid architectural cleanup — the runtime no longer knows about "unload" as a concept, and the `ModelEndpoint` snapshot generalises cleanly to any future "act on the previous agent" hook.

**Strong points:**
- Clean separation: hooks layer no longer reaches into runtime types; runtime layer no longer imports `provider.Unloader`.
- Net −55 LOC after the simplification commit, despite gaining a public wire contract.
- Tests are well-organised — the `unloadURL` table covers every branch of the resolution algorithm in one place, and the no-op contract is pinned with a single tabular test.
- Good test additions on the runtime side (`TestExecuteOnAgentSwitchHooks_PopulatesFromAgentModels` + the nil-on-empty-from variant).

**Trade-offs flagged inline** (all non-blocking):
1. `m.Provider != "dmr"` is stringly-typed — worth a `dmr.ProviderType` constant.
2. The DMR `/v1` → `/_unload` convention now lives in `pkg/hooks/builtins`, away from its DMR siblings — your follow-up #1 would fix this and pairs well with #1 above.
3. Snapshot is built unconditionally — your follow-up #2 (gating on `executor.Has(...)`).
4. Asymmetry with `http_post` (`http.DefaultClient` vs. `NewSafeClient`) is correct but worth a one-liner.

I'd merge with #1+#2 from the diff folded in if quick, otherwise approve as-is and treat them as follow-ups. CI's already green and the contract is stable either way.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25506225829).

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25597420984).

aheritier (approved):
Nice cleanup. The runtime no longer knows about "unload"; the hook is pure HTTP and reads the wire-format `ModelEndpoint` snapshot the runtime ships. All four follow-ups from the previous (dismissed) review are folded in:

- `dmr.ProviderType` typed dispatch
- `dmr.UnloadURL` exported, builtin is a dumb dispatcher
- `exec.Has(EventOnAgentSwitch)` gates the snapshot allocation
- mixed DMR/non-DMR per-element filter test (`TestUnload_FiltersPerElement`)
- SSRF asymmetry vs. `http_post` is justified in a comment

Net negative LOC despite gaining a public wire contract. Tests cover the no-op contract, the per-element filter, error swallowing, and URL resolution branches.

Approving — CI is still pending, will defer to that.

PR #2709, opened 2026-05-07
bump direct go dependencies

dgageot:
| Module | From | To | Status |
|--------|------|----|--------|
| github.com/go-git/go-git/v5 | v5.18.0 | v5.19.0 | bumped (with two follow-up changes, see below) |
| github.com/openai/openai-go/v3 | v3.34.0 | v3.35.0 | bumped |

### Notes on the go-git bump

Two follow-up changes were necessary:

- `go-billy/v5` v5.9.0 (pulled in by `go-git/v5` v5.19.0) now resolves symlinks in `worktree.Filesystem.Root()` (e.g. returns `/private/var/...` instead of `/var/...` on macOS). This caused `VCSMatcher.ShouldIgnore` to fail its `strings.HasPrefix(absPath, m.repoRoot)` check, breaking `TestFilePickerShowIgnoredInGitRepo`. Fix: also call `filepath.EvalSymlinks` on `absPath` in `pkg/fsx/vcs.go`.

- The transitive dependency `github.com/cyphar/filepath-securejoin` is dual-licensed BSD-3-Clause + MPL-2.0 starting with v0.5.0, which the CI license check (`go-licenses`) does not allow. Fix: pin it to v0.4.1 via a `replace` directive in `go.mod` (and add it to `gomoddirectives`' `replace-allow-list`).

Assisted-By: docker-agent


docker-agent (commented):
### Assessment: 🟢 APPROVE

The dependency bumps look correct. The `filepath.EvalSymlinks` fix in `pkg/fsx/vcs.go` is sound: since go-billy v5.9.0 already returns a symlink-resolved path from `worktree.Filesystem.Root()`, both `m.repoRoot` and `absPath` are consistently resolved, and the `HasPrefix` check works correctly on macOS. `go.mod`/`go.sum` are self-consistent.

docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed the dependency bumps (`go-git/v5` v5.19.0, `openai-go/v3` v3.35.0) and the symlink-resolution fix in `pkg/fsx/vcs.go`.

**Fix correctness**: The `filepath.EvalSymlinks(absPath)` addition is correct and well-targeted. Since `go-billy/v5` v5.9.0 now returns a symlink-resolved path from `worktree.Filesystem.Root()`, applying the same resolution to `absPath` before the `strings.HasPrefix` check is the right approach. The `m.repoRoot` field is populated directly from `Root()` at construction time, so both sides of the comparison are consistently resolved.

**One minor note** (below the comment threshold): The `EvalSymlinks` error is silently discarded — if it fails for a transient reason on a valid file, the unresolved `absPath` would be compared against the resolved `m.repoRoot`, potentially causing `ShouldIgnore` to return `false` for a file that should be checked. In practice this is very low risk (the file exists and was just seen by the caller), but propagating or logging the error would make failures more visible.

**Dependency changes**: `go.mod` and `go.sum` look mechanically correct — hashes updated consistently with the bumps.

rumpl (approved):

PR #2710, opened 2026-05-08
fix: centralize environment variable expansion at config boundary

### LINKED ISSUE #2615: Two incompatible env-variable expansion syntaxes coexist silently across agent yaml fields
docker-agent has two completely different env-variable expansion mechanisms, each used in a different subset of yaml fields. They use incompatible syntaxes (`${env.X}` vs `$X` / `${X}`), neither errors when the wrong syntax appears in the wrong field, and there is no documentation that maps fields to the syntax they support. Users who infer the syntax from one example apply it to a similar-looking field and watch their config silently no-op.

## The two mechanisms

**JavaScript template literals via Goja**

- Syntax: `${env.VAR}` (JS member access on a dynamic `env` object), plus `${tool({...})}` and arbitrary JS expressions.
- Used in: `agents.<name>.description`, `agents.<name>.welcome_message`, `agents.<name>.commands`, `toolsets[*].env` values.

**Shell-style env expansion via `os.ExpandEnv`** 

- Syntax: `$VAR` or `${VAR}` (POSIX-ish, no `env.` prefix), plus `~` for home directory.
- Used in: `toolsets[*].working_dir`, toolset path resolution.

## What this looks like to a user

| Field | What works | What silently fails |
|---|---|---|
| `agents.<name>.description` | `${env.X}` | `$X` (literal) |
| `agents.<name>.welcome_message` | `${env.X}` | `$X` (literal) |
| `agents.<name>.commands` | `${env.X}` | `$X` (literal) |
| `agents.<name>.instruction` | (nothing — #2614) | both |
| `toolsets[*].env` (values) | `${env.X}` | `$X` (literal) |
| `toolsets[*].instruction` | (nothing — #2614) | both |
| `toolsets[*].working_dir` | `$X` or `${X}` | `${env.X}` (literal as path component) |
| Toolset path refs | `$X` or `${X}` | `${env.X}` (literal as path component) |

A user who learns from the `description: ${env.HOME}` example will reasonably write `working_dir: ${env.HOME}/work` and end up with a directory literally named `${env.HOME}`. A user who learns from `working_dir: $HOME/work` will write `description: $HOME` and the model will see the literal `$HOME`.

Neither failure mode logs anything. The runtime accepts both yamls, loads the agent, and proceeds. A fast-fail at load (or a clear log line) would turn each instance into a 30-second fix. The current behavior turns it into a debugging session.

## Suggestions

1. **Unify on a single syntax.** The JS template literal expander is the more powerful one (env access, tool calls, arbitrary expressions) and is already used in more places. Replace the `os.ExpandEnv` call sites in `pkg/path/expand.go` with the JS expander. Treat `~` as a separate path-only concern. This eliminates the dual-syntax problem entirely.

2. **Make both syntaxes work everywhere.** Run shell-style expansion as a pre-pass before the JS expander, or vice versa. Costs: more surface area, harder to reason about precedence, edge cases when a string legitimately contains `$X`. Possible but messy.

If none of those is acceptable, at minimum: add the matrix to the docs.

yunus25jmi1:
### Root Cause
Fragmented expansion logic across the codebase caused incompatibility between syntaxes (`$VAR` vs `${env.VAR}`) and silent failures. Path fields only supported POSIX style, while command fields used a JS-based expander that often ignored standard environment variables or failed silently.

### Solution: Layered Expansion Architecture
Implemented a robust two-layer expansion pipeline at the configuration boundary:

1. **Layer 1: Universal Expansion (Config Loading Phase)**
   - Normalizes all syntaxes (`$VAR`, `${VAR}`, `${env.VAR}`) into a standard form.
   - Uses a reflection-based walker (`pkg/config/expand.go`) to visit and expand every exported string field, slice, and map in the configuration immediately after unmarshaling.
   - Provides explicit error reporting via `ErrMissingVars` instead of silent partial expansion.

2. **Layer 2: Dynamic JS Expressions (Runtime Phase)**
   - Simplified the JS expander (`pkg/js/expand.go`) by removing redundant environment handling.
   - JS engine now focuses strictly on dynamic logic (e.g., `${tool(...)}`), as environment variables are guaranteed to be resolved by Layer 1.

### Key Changes
- **pkg/environment/expand.go**: Added normalization and centralized expansion logic using `os.Expand`.
- **pkg/config/expand.go**: New reflection-based config walker to ensure 100% field coverage.
- **pkg/config/config.go**: Integrated expansion into the main `Load` pipeline.
- **pkg/js/expand.go**: Decoupled JS engine from environment resolution.
- **pkg/config/sources.go**: Updated `Source` interface to carry `EnvProvider` context.

### Impact
- Full support for `$VAR`, `${VAR}`, and `${env.VAR}` in **all** string-based config fields (working_dir, commands, instruction, etc).
- Elimination of silent expansion failures.
- Cleaner separation of concerns between configuration loading and execution runtime.

Fixes #2615

### DIFF
--- a/pkg/config/commands_test.go
+++ b/pkg/config/commands_test.go
@@ -7,7 +7,7 @@ import (
 )
 
 func TestV2Commands_AllForms(t *testing.T) {
-	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v2.yaml"))
+	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v2.yaml", nil))
 	require.NoError(t, err)
 
 	// Test simple map format
@@ -38,7 +38,7 @@ func TestV2Commands_AllForms(t *testing.T) {
 }
 
 func TestV2Commands_DisplayText(t *testing.T) {
-	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v2.yaml"))
+	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v2.yaml", nil))
 	require.NoError(t, err)
 
 	// Simple format: DisplayText returns the instruction
@@ -51,7 +51,7 @@ func TestV2Commands_DisplayText(t *testing.T) {
 }
 
 func TestMigrate_v1_Commands_AllForms(t *testing.T) {
-	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v1.yaml"))
+	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v1.yaml", nil))
 	require.NoError(t, err)
 
 	require.Equal(t, "root", cfg.Agents[0].Name)
@@ -69,7 +69,7 @@ func TestMigrate_v1_Commands_AllForms(t *testing.T) {
 }
 
 func TestMigrate_v0_Commands_AllForms(t *testing.T) {
-	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v0.yaml"))
+	cfg, err := Load(t.Context(), NewFileSource("testdata/commands_v0.yaml", nil))
 	require.NoError(t, err)
 
 	require.Equal(t, "root", cfg.Agents[0].Name)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -8,13 +8,11 @@ import (
 	"log/slog"
 	"maps"
 	"net/url"
-	"path/filepath"
 	"slices"
 	"strings"
 
 	"github.com/goccy/go-yaml"
 
-	hclconv "github.com/docker/docker-agent/pkg/config/hcl"
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
 )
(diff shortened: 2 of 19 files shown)


yunus25jmi1:
@dgageot I am currently working on this PR. I am facing some GO environment system-level configuration problems. I will make a comment when my PR is in good shape.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25556523803).

yunus25jmi1:
@dgageot The PR is now in good shape. All lint issues have been resolved and the checks are passing. Ready for review!

aheritier:
**Triage note:** Community PR (contributor) with 5 comments and active discussion. The proposal — a reflection-based config walker that expands all string fields at load time — is a significant change to the config loading pipeline. Key concerns raised in comments:

1. **Overreach**: A reflection walker that expands *all* string fields could inadvertently expand literal `$` values in user-defined instructions, prompt text, or tool output patterns — any field not intended to support expansion.
2. **Interface change**: `Source` carrying `EnvProvider` is a breaking change to the `Source` interface used by multiple call sites.
3. **JS engine coupling**: Decoupling the JS expander from env vars may break existing YAML configs that rely on `${env.VAR}` being handled at the JS layer.

Current status: not yet approved; needs maintainer sign-off on the design direction before proceeding. The original issue (#2615) may have a narrower fix available.

yunus25jmi1:
Hey @dgageot @aheritier — before continuing, I'd like to align on the direction given the triage concerns.

**Three questions:**

1. **Field overreach**: The reflection walker expands all exported string fields. Is blanket expansion the right default, or should we scope it to specific fields?

2. **Interface change**: Adding `EnvProvider()` to `Source` is breaking. Would passing env context as a parameter to `Load()` be preferable?

3. **JS engine simplification**: Removing env handling from the JS layer may break existing configs relying on `${env.VAR}` at the JS layer. Any concerns?

**Alternative path**: The original issue #2615 suggested replacing `os.ExpandEnv` in `pkg/path/expand.go` with the JS expander — a narrower fix targeting just working_dir/path fields. Would the team prefer this as an initial step?

Looking for direction before investing further effort.

dgageot:

PR #2711, opened 2026-05-08
fix: use available Kimi-K2.5 model in nebius example

dgageot:
The `moonshotai/Kimi-K2-Instruct` model is no longer listed in the nebius provider catalog on models.dev, which caused `TestParseExamples/../../examples/nebius.yaml` to fail. Switch the example to `moonshotai/Kimi-K2.5`, which is the current Moonshot model offered by nebius.

### DIFF
--- a/examples/nebius.yaml
+++ b/examples/nebius.yaml
@@ -12,6 +12,6 @@ agents:
 models:
   grok-model:
     provider: nebius
-    model: moonshotai/Kimi-K2-Instruct
+    model: moonshotai/Kimi-K2.5
     max_tokens: 110000
     temperature: 0.7


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR makes a minimal, targeted fix: replacing the deprecated `moonshotai/Kimi-K2-Instruct` model name with `moonshotai/Kimi-K2.5` in the nebius provider example YAML. The change correctly resolves the `TestParseExamples` test failure caused by the old model being removed from the nebius catalog on models.dev. No bugs or issues were found in the changed code.

rumpl (approved):

PR #2713, opened 2026-05-08
security: five defense-in-depth fixes (secrets, archives, oauth, shell tool, request logs)

dgageot:
Five small, focused, security-hardening commits surfaced while reading
the codebase like a security researcher. None of them changes a public
API, removes a feature, or alters a default the user explicitly
requested — they're defense-in-depth fixes that close concrete attack
paths.

Each commit passes `task lint` and `task test` independently.

| # | Commit | Summary |
|---|---|---|
| 1 | `fix(environment): make RunSecretsProvider symlink/TOCTOU-safe` | `/run/secrets/<name>` lookups now go through `os.OpenRoot` + `Root.ReadFile` so symbolic links escaping the secrets root are rejected by the kernel instead of silently followed. Adds a planted-symlink regression test. |
| 2 | `fix(toolinstall): bound archive extraction to defeat zip/tar bombs` | Per-entry, per-archive, and per-binary uncompressed caps via `io.LimitReader`; entry-count cap; bounded compressed-spool. Drops the `//nolint:gosec` band-aids that papered over unbounded `io.Copy`s. |
| 3 | `fix(mcp/oauth): harden the HTTP client used for token endpoints` | OAuth token-exchange / refresh / dynamic-registration calls go through `httpclient.NewSafeClient` (SSRF-safe transport, 30 s timeout, bounded redirects) instead of `http.DefaultClient`. A package-local `TestMain` swaps in the loopback-allowing variant so `httptest.NewServer`-driven tests keep working. |
| 4 | `fix(shell): drop undeclared script_shell args and reject NUL bytes` | `ScriptShellTool.execute` now only forwards JSON keys declared in the tool's `args` schema, so an LLM that hallucinates `LD_PRELOAD` / `PATH` cannot inject arbitrary env entries into the spawned process. Values containing NUL are rejected (would otherwise truncate at `execve`). |
| 5 | `fix(server,chatserver): redact query strings in request logs` | New `pkg/echolog.RedactedRequestLogger` is wired into both servers in place of `middleware.RequestLogger()`. Logs `req.URL.Path`, never the full URI, so `?token=...` query parameters cannot leak into structured logs. Includes a regression test. |

These are derived from a longer security review of the codebase. The
remaining items (auth on `/api/*`, slowloris timeouts on the REST API
server, Unix-socket permissions, OAuth state width, CORS regex
anchoring) are intentionally left for follow-up so each PR stays
reviewable.

### DIFF
--- a/pkg/echolog/echolog.go
+++ b/pkg/echolog/echolog.go
@@ -6,7 +6,6 @@
 package echolog
 
 import (
-	"context"
 	"log/slog"
 
 	"github.com/labstack/echo/v4"
@@ -36,7 +35,7 @@ func RedactedRequestLogger() echo.MiddlewareFunc {
 		LogContentLength: true,
 		LogResponseSize:  true,
 		HandleError:      true,
-		LogValuesFunc: func(_ echo.Context, v middleware.RequestLoggerValues) error {
+		LogValuesFunc: func(c echo.Context, v middleware.RequestLoggerValues) error {
 			level := slog.LevelInfo
 			msg := "REQUEST"
 			attrs := []slog.Attr{
@@ -56,7 +55,7 @@ func RedactedRequestLogger() echo.MiddlewareFunc {
 				msg = "REQUEST_ERROR"
 				attrs = append(attrs, slog.String("error", v.Error.Error()))
 			}
-			slog.LogAttrs(context.Background(), level, msg, attrs...)
+			slog.LogAttrs(c.Request().Context(), level, msg, attrs...)
 			return nil
 		},
 	})
--- a/pkg/echolog/echolog.go
+++ b/pkg/echolog/echolog.go
@@ -1,0 +1,63 @@
+// Package echolog provides Echo middleware tailored for docker-agent's
+// HTTP servers. It currently exposes a single helper, RedactedRequestLogger,
+// that mirrors the default echo middleware.RequestLogger() but never
+// records request bodies, raw query strings, or other places that may
+// carry secrets.
+package echolog
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/labstack/echo/v4"
+	"github.com/labstack/echo/v4/middleware"
+)
+
+// RedactedRequestLogger returns an Echo request-logging middleware that
+// matches middleware.RequestLogger() in spirit (one structured slog line
+// per request) but logs the URL path only, not the full request URI.
+//
+// Echo's default logger emits v.URI, which includes the query string.
+// Some clients pass authentication material through query parameters,
+// which would otherwise be persisted to whichever sink slog points at.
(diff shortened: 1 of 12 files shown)


docker-agent (commented):

Five well-scoped security hardening commits. The symlink-safe secrets provider, archive extraction limits, OAuth SSRF hardening, shell arg allowlist, and query-string redaction are all sound. One finding in the new  package warrants attention.

pkg/echolog/echolog.go
```diff
@@ -0,0 +1,63 @@
...
+				slog.String("user_agent", v.UserAgent),
+				slog.String("remote_ip", v.RemoteIP),
+				slog.String("request_id", v.RequestID),
+			}
+			if v.Error != nil {
+				level = slog.LevelError
+				msg = "REQUEST_ERROR"
+				attrs = append(attrs, slog.String("error", v.Error.Error()))
+			}
+			slog.LogAttrs(context.Background(), level, msg, attrs...)
```
docker-agent: **[MEDIUM] Request logger uses `context.Background()`, losing per-request trace and correlation IDs**

`RedactedRequestLogger` deliberately discards the `echo.Context` (bound to `_`) and passes `context.Background()` to `slog.LogAttrs`. Any values stored in the per-request `context.Context` — such as OpenTelemetry trace spans, correlation IDs, or other slog attributes injected by earlier middleware — will be absent from every request log line emitted by this middleware.

The project recently added [commit `5f0a5796`](https://github.com/docker/docker-agent/commit/5f0a5796) — *"refactor(logging): pass context to all slog calls for correlation"* — to propagate request context throughout the server codebase. This new logger regresses that effort for all HTTP request logs.

**Fix**: extract the request context from the `echo.Context` and pass it to `slog.LogAttrs`:

```go
LogValuesFunc: func(c echo.Context, v middleware.RequestLoggerValues) error {
    ctx := c.Request().Context()
    // ...
    slog.LogAttrs(ctx, level, msg, attrs...)
    return nil
},
```

docker-agent (commented):

Five well-scoped security hardening commits. The symlink-safe secrets provider (`os.OpenRoot`), archive extraction limits, OAuth SSRF hardening, shell arg allowlist, and query-string redaction are all sound. One finding in the new `echolog` package warrants attention — see inline comment.

**Reviewed**: pkg/environment/secrets.go · pkg/toolinstall/archive.go · pkg/tools/mcp/oauth_helpers.go · pkg/tools/builtin/shell/script_shell.go · pkg/echolog/echolog.go · pkg/server/server.go · pkg/chatserver/server.go

rumpl (approved):

rumpl (dismissed):
Approved.

rumpl (approved):
Approved.

PR #2714, opened 2026-05-08
feat: let external processes drive a running TUI

dgageot:
External processes can now drive a running docker-agent TUI. Implemented as 9 small, additive commits in the spirit of *"make the change easy, then make the easy change"* — the first two are pure refactors that unlock everything else.

## Why

Tools like opencode, Claude Code, Codex, and Gemini CLI all expose some way for an external process (editor, script, another agent) to drive a running session. We had `docker-agent serve api` (full HTTP control plane) and `docker-agent run` (TUI), but the two were mutually exclusive — the running TUI was a black box.

The goal: **let one process be both**, with no architectural change.

## What's new

| Capability | How |
|---|---|
| Drive a live TUI over HTTP | `docker-agent run --listen 127.0.0.1:0` exposes the existing API against the same in-process runtime |
| Auto-discovery | `<data dir>/runs/<pid>.json` records every live run; pruned on exit and on stale-pid `List()` |
| Send a prompt from another shell | `docker-agent send "summarize the diff"` (or `--followup`, `--to <pid>`) |
| Tail events as JSON | `docker-agent watch \| jq` |
| Claude Code-style observability hooks | `--on-event stream_stopped="say done" --on-event "*=tee /tmp/events.log"` |
| MCP attach for agent-to-agent control | `docker-agent serve mcp --attach` (exposes `send` + `transcript` MCP tools) |
| Stdio JSON protocol (Codex `proto` style) | `docker-agent proto` — newline-delimited JSON-RPC for editor integrations |

## The 9 commits

| # | Commit | Type |
|---|---|---|
| 1 | `refactor(server): allow attaching a pre-built runtime` | Tidy |
| 2 | `refactor(app): turn SubscribeWith into a fan-out` | Tidy |
| 3 | `feat(run): add hidden --listen flag to expose the running TUI` | Feature |
| 4 | `feat(runregistry): publish discovery files for live runs` | Feature |
| 5 | `feat(send): new subcommand to drive a live TUI` | Feature |
| 6 | `feat(watch): stream events from a running TUI` | Feature |
| 7 | `feat(run): --on-event hooks observe arbitrary events` | Feature |
| 8 | `feat(serve mcp): --attach exposes a running TUI via MCP` | Feature |
| 9 | `feat(proto): newline-delimited JSON protocol over stdio` | Feature |

Each commit is small, reversible, opt-in, and unobservable to users who don't ask for it. Steps 1 and 2 are pure refactors:

- **Step 1** factors `Server.New` so a pre-built `SessionManager` (with `AttachRuntime`) can back it. No behavior change to `serve api`.
- **Step 2** turns `App.SubscribeWith` into a fan-out: one drain goroutine, many subscribers. The TUI keeps its single subscriber; HTTP/SSE, hooks, MCP, and stdio all become extra subscribers.

Steps 3–9 are then composition over the existing API surface — almost no new domain logic.

## Architecture

```
                   ┌─────────────┐
                   │  TUI (App)  │ ◄── subscriber #1
                   └──────┬──────┘
                          │  fan-out (step 2)
                          ▼
        ┌─────────────────────────────────────┐
        │            App events bus           │
        └────┬─────────┬─────────┬────────────┘
             │         │         │
             ▼         ▼         ▼
        ┌────────┐ ┌────────┐ ┌────────────┐
        │ HTTP   │ │ on-    │ │ MCP / proto│
        │ /events│ │ event  │ │ (via HTTP) │
        │ (SSE)  │ │ shell  │ └────────────┘
        └────────┘ └────────┘
             ▲
             │  (drives steer/followup/resume/...)
        ┌────┴──────────────────────────────┐
        │  send / watch / proto subcommands │
        │  (auto-discover via runregistry)  │
        └───────────────────────────────────┘
```

The `runtime.Runtime` is the API. The TUI is already a client of it (via `--remote`). This PR just lets the same in-process `Runtime` simultaneously serve the TUI **and** the HTTP control plane. Everything else is composition.

## Validation

- `golangci-lint run ./...` → **0 issues**
- `go run ./lint .` → **1006 file(s) inspected, no offenses detected**
- `go test ./...` → **123 packages OK, 0 failed**
- `go test -race -count=1 ./pkg/server/... ./pkg/app/... ./pkg/runregistry/... ./pkg/mcp/... ./cmd/root/...` → all pass, no data races

## Try it

```bash
# Terminal 1: start a TUI that exposes its control plane
docker-agent run ./agent.yaml --listen 127.0.0.1:0

# Terminal 2: drive it
docker-agent send "what files changed?"
docker-agent watch | jq 'select(.type=="agent_choice").content'

# Terminal 3: editor / agent integration
docker-agent proto                # stdio JSON-RPC
docker-agent serve mcp --attach   # expose as MCP tools to another agent
```

## Notes

- `--listen` is **hidden** for now (opt-in / experimental).
- `--on-event` complements (does **not** replace) the existing `pkg/hooks` system — it's purely observational, for events that don't have formal hook points.
- `pkg/server.NewWithManager` is a public sibling of `New`; existing callers are untouched.
- The `runregistry` package handles stale-pid cleanup automatically on `List()`.

### DIFF
--- a/pkg/runregistry/registry.go
+++ b/pkg/runregistry/registry.go
@@ -1,0 +1,115 @@
+// Package runregistry persists discovery records for running docker-agent
+// processes that expose a control plane (see run --listen). Records live as
+// per-pid JSON files under <data dir>/runs so external tools can enumerate
+// live runs and connect to them.
+package runregistry
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/docker/docker-agent/pkg/paths"
+)
+
+// Record describes a running docker-agent that exposes a control plane.
+type Record struct {
+	PID       int       `json:"pid"`
+	Addr      string    `json:"addr"`
+	SessionID string    `json:"session_id"`
+	Agent     string    `json:"agent,omitempty"`
+	StartedAt time.Time `json:"started_at"`
+}
+
+// Dir is the directory holding run records.
+func Dir() string {
+	return filepath.Join(paths.GetDataDir(), "runs")
+}
+
+// Write atomically persists a record for the current process and returns a
+// cleanup func that removes it. Cleanup is safe to call more than once.
+func Write(rec Record) (func(), error) {
+	if err := os.MkdirAll(Dir(), 0o755); err != nil {
+		return nil, fmt.Errorf("creating run registry dir: %w", err)
+	}
+
+	path := filepath.Join(Dir(), strconv.Itoa(rec.PID)+".json")
+	buf, err := json.MarshalIndent(rec, "", "  ")
+	if err != nil {
+		return nil, err
+	}
+	if err := os.WriteFile(path, buf, 0o600); err != nil {
+		return nil, err
+	}
+
+	return func() { _ = os.Remove(path) }, nil
+}
+
+// List returns every record currently registered. Stale records (whose pid is
+// no longer alive) are skipped and best-effort removed.
+func List() ([]Record, error) {
+	entries, err := os.ReadDir(Dir())
+	if errors.Is(err, fs.ErrNotExist) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	var out []Record
+	for _, e := range entries {
+		if e.IsDir() || !strings.HasSuffix(e.Name(), ".json") {
+			continue
+		}
+		path := filepath.Join(Dir(), e.Name())
+		buf, err := os.ReadFile(path)
+		if err != nil {
+			continue
+		}
+		var rec Record
+		if err := json.Unmarshal(buf, &rec); err != nil {
+			continue
+		}
+		if !pidAlive(rec.PID) {
+			_ = os.Remove(path)
+			continue
+		}
+		out = append(out, rec)
+	}
+	return out, nil
+}
+
+// Latest returns the most recently started live record, or false when none.
+func Latest() (Record, bool, error) {
+	records, err := List()
+	if err != nil || len(records) == 0 {
+		return Record{}, false, err
+	}
+	latest := records[0]
+	for _, r := range records[1:] {
+		if r.StartedAt.After(latest.StartedAt) {
+			latest = r
+		}
+	}
+	return latest, true, nil
+}
+
+// pidAlive reports whether the given pid corresponds to a live process.
+// Uses os.FindProcess + signal 0, the cross-platform idiom for liveness.
+func pidAlive(pid int) bool {
+	if pid <= 0 {
+		return false
+	}
+	p, err := os.FindProcess(pid)
+	if err != nil {
+		return false
+	}
+	return p.Signal(syscall.Signal(0)) == nil
+}
--- a/pkg/runregistry/registry.go
+++ b/pkg/runregistry/registry.go
@@ -35,8 +35,12 @@ func Dir() string {
 
 // Write atomically persists a record for the current process and returns a
 // cleanup func that removes it. Cleanup is safe to call more than once.
+//
+// The registry directory is created with 0o700 so other local users cannot
+// enumerate live PIDs/addresses by listing it. Individual records are still
+// written with 0o600 for the same reason.
 func Write(rec Record) (func(), error) {
-	if err := os.MkdirAll(Dir(), 0o755); err != nil {
+	if err := os.MkdirAll(Dir(), 0o700); err != nil {
 		return nil, fmt.Errorf("creating run registry dir: %w", err)
 	}
 
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -66,6 +66,18 @@ func NewSessionManager(ctx context.Context, sources config.Sources, sessionStore
 	return sm
 }
 
+// AttachRuntime registers a pre-built runtime + session under sessionID so
+// that subsequent calls (RunSession, Steer, Resume...) reuse it instead of
+// building one from agentFilename. This is what lets a single in-process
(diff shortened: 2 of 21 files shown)


docker-agent (commented):

3 medium-severity issues found in newly added code.

| Severity | Finding | File |
|---|---|---|
| 🟡 MEDIUM | Registry directory uses 0o755 — leaks PID via dir listing | pkg/runregistry/registry.go:39 |
| 🟡 MEDIUM | AttachRuntime no-op cancel; DeleteSession can't stop streaming | pkg/server/session_manager.go:95 |
| 🟡 MEDIUM | EventSource entries never removed from map on DeleteSession | pkg/server/session_manager.go:79 |

pkg/runregistry/registry.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Dir is the directory holding run records.
+func Dir() string {
+	return filepath.Join(paths.GetDataDir(), "runs")
+}
+
+// Write atomically persists a record for the current process and returns a
+// cleanup func that removes it. Cleanup is safe to call more than once.
+func Write(rec Record) (func(), error) {
+	if err := os.MkdirAll(Dir(), 0o755); err != nil {
```
docker-agent: **[MEDIUM] Registry directory created with world-readable permissions (0o755)**

`os.MkdirAll` uses `0o755`, which allows any local user to list the `runs/` directory and enumerate filenames. Since filenames are `<pid>.json`, this leaks the PID of every running docker-agent instance to all users on the machine. The individual files are correctly protected (`0o600`), but directory listing bypasses that.

Consider using `0o700` so only the owning user can list the directory:
```go
if err := os.MkdirAll(Dir(), 0o700); err != nil {
```

pkg/server/session_manager.go
```diff
@@ -66,6 +73,30 @@ func NewSessionManager(ctx context.Context, sources config.Sources, sessionStore
...
+}
+
+// AttachRuntime registers a pre-built runtime + session under sessionID so
+// that subsequent calls (RunSession, Steer, Resume...) reuse it instead of
+// building one from agentFilename. This is what lets a single in-process
+// runtime be shared between the TUI and an HTTP control plane.
+func (sm *SessionManager) AttachRuntime(sessionID string, rt runtime.Runtime, sess *session.Session) {
+	sm.runtimeSessions.Store(sessionID, &activeRuntimes{
+		runtime: rt,
+		cancel:  func() {},
```
docker-agent: **[MEDIUM] AttachRuntime stores a no-op cancel; DeleteSession cannot stop an in-flight stream**

`AttachRuntime` initialises `cancel` as `func() {}`. When `DeleteSession` is called it invokes `sessionRuntime.cancel()`, which is a no-op for any attached runtime — the streaming goroutine keeps running until the parent context is cancelled.

Additionally, in `RunSession`, when an existing `runtimeSession` is found, a new `streamCtx` and `cancel` are derived from `ctx` (line ~189) but are **never written back** into `runtimeSession.cancel`. So even if `AttachRuntime` had stored a real cancel, a subsequent `RunSession` call would still leave `runtimeSession.cancel` pointing at the old function.

One fix is for `RunSession` to update `runtimeSession.cancel` whenever it creates a new streaming context, and for `AttachRuntime` to accept a cancel func (or derive one from the passed runtime's lifecycle).

pkg/server/session_manager.go
```diff
@@ -66,6 +73,30 @@ func NewSessionManager(ctx context.Context, sources config.Sources, sessionStore
 	return sm
 }
 
+// RegisterEventSource attaches an event source for sessionID. It is used by
+// callers that own a runtime out-of-band (e.g. the TUI) so that HTTP clients
+// can subscribe to events via GET /api/sessions/:id/events.
+func (sm *SessionManager) RegisterEventSource(sessionID string, src EventSource) {
```
docker-agent, line 80: **[MEDIUM] EventSource entries are never removed from the map; resource leak on session deletion**

`RegisterEventSource` stores closures in `sm.eventSources`, but `DeleteSession` only cleans up `sm.runtimeSessions` — it never calls `sm.eventSources.Delete(sess.ID)`. Over time, deleted sessions leave stale `EventSource` closures accumulating in the map.

There is also a correctness risk: if a session is deleted and a new session is created with the same ID (unlikely but possible), `GetEventSource` would return the stale closure from the old session instead of the new one.

Fix: add `sm.eventSources.Delete(sess.ID)` to `DeleteSession`, mirroring how `runtimeSessions` is handled.

rumpl:
Next step, the tui is _not_ the control plane, but connects to one so everyone is on the same level: clients of one control plane

rumpl (dismissed):
Approved.

dgageot:
Addressed the review feedback in 0e6f51202:

- **`pkg/runregistry/registry.go`** — registry dir is now created with `0o700` (was `0o755`) so other local users can't enumerate live PIDs/addresses. Added a regression test (skipped on Windows).
- **`pkg/server/session_manager.go:AttachRuntime`** — replaced the no-op cancel with a real `context.WithCancel`. The session's done channel is exposed through the new `StreamEvents` helper which merges it with the caller's request context, so `DeleteSession` now unblocks in-flight `/events` SSE consumers as expected.
- **`pkg/server/session_manager.go:DeleteSession`** — also removes the entry from the `eventSources` map (was leaking the closure that captures the App).
- **`pkg/server/server.go:sessionEvents`** — drives the source through `StreamEvents` instead of invoking it directly with the bare request context.

New test: `TestAttachedServer_DeleteSessionStopsEventStream` verifies that deleting an attached session cancels the SSE source's context, removes the registry entry, and that subsequent `/events` requests return 404. `task lint` and `task test` (the affected packages) pass cleanly with `-race`.

The `rumpl` comment about the TUI being a control-plane client rather than the control plane itself is a follow-up direction; not part of this PR.

rumpl (approved):
Approved.

PR #2718, opened 2026-05-08
Change the default models for the golang dev

rumpl:


golang_developer.yaml
```diff
@@ -44,13 +41,13 @@ models:
     model: claude-haiku-4-5
   gpt:
     provider: openai
-    model: gpt-5.2
+    model: gpt-5.5
```
docker-agent, line 44: **[MEDIUM] `gpt-5.5` may not be a valid OpenAI model identifier**

The model name was changed from `gpt-5.2` → `gpt-5.5`, but neither name matches any publicly known OpenAI model (known names include `gpt-4o`, `gpt-4.1`, `gpt-4.5`, `gpt-5`). If this identifier is not recognised by the OpenAI API at runtime, any task routed through the `gpt` or `extramax` model will fail with an API error.

If this is an internal/preview model name, this is fine — but it's worth confirming the string is correct before merging.

golang_developer.yaml
```diff
@@ -44,13 +41,13 @@ models:
...
   extramax:
-    model: opus,gpt
+    model: sonnet,gpt
 
 agents:
   root:
-    model: opus
+    model: gpt
```
docker-agent, line 50: **[MEDIUM] Root agent depends on potentially invalid model `gpt-5.5`**

The root agent model was switched from `opus` (Claude Opus 4.6) to `gpt` (which maps to `gpt-5.5`). Since the root agent is the primary entry-point for the entire golang developer workflow, if `gpt-5.5` is not a valid model identifier the agent will fail on every invocation — causing complete loss of functionality rather than a partial degradation.

If `gpt-5.5` is confirmed valid, this concern is resolved.

docker-agent (commented):

dgageot (approved):

PR #2723, opened 2026-05-09
remote-runtime: close silent gaps, consolidate Runtime, scaffold wire (10 baby steps)

dgageot:
## Goal

Pick up phase A+B+C from the [previous "remote-runtime baby steps" PR](https://github.com/docker/docker-agent/commits/main): close the silent-breakage gaps `--remote` still has compared to local mode, finish consolidating the `Runtime` interface, and stand up the data layer + HTTP scaffold that a real docker-agent server will eventually live behind.

Ten Kent-Beck-style baby steps. Each commit:
- builds cleanly,
- passes `task lint` (golangci-lint + custom checks + `go mod tidy`),
- passes `task test` (full suite),
- sets up the next move and explains it in the body.

## Commits

| # | Commit | Move |
|---|---|---|
| 1 | `fix(run): route session spawner through loadTeamRequest` | TUI's spawned-session path was calling `teamloader.LoadWithConfig` directly with only `WithModelOverrides`, silently dropping `--prompt-file` on every spawned session. Funnel it through `loadAgentFrom(ctx, loadTeamRequest(...))` so spawned sessions see the same flag set as the initial load. |
| 2 | `test(runtime): run the contract suite against RemoteRuntime` | `TestRemoteRuntime_Contract` plugs `RemoteRuntime` into `runRuntimeContract` via a stub client. All 14 non-streaming subtests now pin the surface for both runtimes. The one gap surfaced was `SetCurrentAgent` silently accepting any name; fix it inline by validating against the agent list returned by `client.GetAgent`. |
| 3 | `feat(runtime): wrap remaining RemoteRuntime no-ops with ErrUnsupported` | Two methods were silently lying to the TUI: `CurrentAgentTools` returned `(nil, nil)` (the /tools dialog opened with an empty list, indistinguishable from "agent has no tools") and `Summarize` emitted a `SessionSummary` event with text "not yet implemented" (the persistence observer wrote it to the session DB unchanged, corrupting the session). Both now surface `ErrUnsupported` / `Error` events. |
| 4 | `refactor(runtime): promote ModelSwitcher onto Runtime` | `SetAgentModel` / `AvailableModels` / `SupportsModelSwitching` are now part of the `Runtime` interface. `LocalRuntime` delegates to its existing implementation; `RemoteRuntime` returns `ErrUnsupported`. Drop the `ModelSwitcher` interface and `ModelSwitcherOf` helper; 5 call sites simplified. |
| 5 | `refactor(runtime): promote ToolsChangeSubscriber onto Runtime` | Same pattern. `OnToolsChanged` is a regular `Runtime` method now; `RemoteRuntime` is a no-op (tool-list changes flow through the run-stream events server-side). Drop the interface and helper. |
| 6 | `refactor(runtime): delete capabilities.go` | The optional-sub-interface helper file is empty. Delete it. `Runtime` is now the single source of truth for what backends can do — no more "is this thing also implementing some optional sub-interface" type assertions. |
| 7 | `refactor(runtime): move LoadTeamRequest and CreateSessionRequest to pkg/runtime` | Move the two payload structs out of `cmd/root` into `pkg/runtime/payload.go` so they can grow JSON tags and a server side. Factory methods (`loadTeamRequest`, `createSessionRequest`) stay near the CLI flag struct they read from. |
| 8 | `feat(runtime): add JSON tags and round-trip tests for payloads` | `LoadTeamRequest` and `CreateSessionRequest` now marshal/unmarshal cleanly. Wire-incompatible fields (`Source` interface, `RuntimeConfig`, `permissions.Checker`) get `json:"-"` with a doc comment; round trip + omitempty + non-serializable-fields-excluded contracts pinned by 6 tests. |
| 9 | `refactor(run): split backend interface into LoadTeam and CreateSession` | The two-step protocol an RPC server will mirror is now visible on the `backend` interface. Side-effect: local `--dry-run` now validates the team (catches YAML typos at dry-run time). Remote `--dry-run` is unchanged — `LoadTeam` is a no-op for remote, so the previous PR's "no leaked server-side session" guarantee still holds. |
| 10 | `feat(runtime): scaffold pkg/runtime/wire HTTP handler` | `POST /v1/team/load` and `POST /v1/session/create` decode the payloads from step 8 and dispatch to a `Backend` interface. Round-trip + error mapping pinned by 5 tests: backend errors → 500, `errors.Is(err, runtime.ErrUnsupported)` → 501, unknown JSON fields → 400. |

## What's still out of scope (each its own follow-up)

- `wire.Client` driving `RemoteRuntime`
- `/v1/session/run` streaming endpoint (SSE or chunked HTTP/2)
- Authentication
- Wire-friendly representations for `Source` / `RuntimeConfig` / `permissions.Checker` (today they're `json:"-"`)
- Bridging `cmd/root.backend` onto `wire.Backend` so the server actually runs a real `LocalBackend` rather than a stub

## Validation

```
$ task lint
1019 file(s) inspected, no offenses detected
$ task test
... all green ...
```

Plus smoke tests on both `--remote http://nosuchhost:9999 --dry-run` (exits without contacting the server) and local `--dry-run` (now also validates the team config).

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -413,12 +413,10 @@ func (f *runExecFlags) createLocalRuntimeAndSession(ctx context.Context, loadRes
 		sess.HideToolResults = req.HideToolResults
 
 		// Apply any stored model overrides from the session
-		if len(sess.AgentModelOverrides) > 0 {
-			if modelSwitcher := runtime.ModelSwitcherOf(localRt); modelSwitcher != nil {
-				for agentName, modelRef := range sess.AgentModelOverrides {
-					if err := modelSwitcher.SetAgentModel(ctx, agentName, modelRef); err != nil {
-						slog.WarnContext(ctx, "Failed to apply stored model override", "agent", agentName, "model", modelRef, "error", err)
-					}
+		if len(sess.AgentModelOverrides) > 0 && localRt.SupportsModelSwitching() {
+			for agentName, modelRef := range sess.AgentModelOverrides {
+				if err := localRt.SetAgentModel(ctx, agentName, modelRef); err != nil {
+					slog.WarnContext(ctx, "Failed to apply stored model override", "agent", agentName, "model", modelRef, "error", err)
 				}
 			}
 		}
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -303,7 +303,7 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 	return runTUI(ctx, rt, sess, b.Spawner(rt), cleanup, f.tuiOpts(), opts...)
 }
 
-func (f *runExecFlags) loadAgentFrom(ctx context.Context, req LoadTeamRequest) (*teamloader.LoadResult, error) {
+func (f *runExecFlags) loadAgentFrom(ctx context.Context, req runtime.LoadTeamRequest) (*teamloader.LoadResult, error) {
 	opts := []teamloader.Opt{
 		teamloader.WithModelOverrides(req.ModelOverrides),
 	}
@@ -360,7 +360,7 @@ func (f *runExecFlags) snapshotRuntimeOpts() ([]runtime.Opt, builtins.SnapshotCo
 	}, ctrl, nil
 }
 
-func (f *runExecFlags) createLocalRuntimeAndSession(ctx context.Context, loadResult *teamloader.LoadResult, req CreateSessionRequest) (runtime.Runtime, *session.Session, error) {
+func (f *runExecFlags) createLocalRuntimeAndSession(ctx context.Context, loadResult *teamloader.LoadResult, req runtime.CreateSessionRequest) (runtime.Runtime, *session.Session, error) {
 	t := loadResult.Team
 
 	// Merge user-level global permissions into the team's checker so the
@@ -511,7 +511,7 @@ func (f *runExecFlags) buildAppOpts(args []string) ([]app.Opt, error) {
 // buildSessionOpts returns the canonical set of session options derived from
 // CLI flags and agent configuration. Both the initial session and spawned
 // sessions use this method so their options never drift apart.
-func (f *runExecFlags) buildSessionOpts(agt *agent.Agent, req CreateSessionRequest) []session.Opt {
+func (f *runExecFlags) buildSessionOpts(agt *agent.Agent, req runtime.CreateSessionRequest) []session.Opt {
 	return []session.Opt{
 		session.WithMaxIterations(agt.MaxIterations()),
 		session.WithMaxConsecutiveToolCalls(agt.MaxConsecutiveToolCalls()),
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -263,12 +263,21 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		return err
 	}
 
+	loadResult, err := b.LoadTeam(ctx, b.LoadTeamRequest())
+	if err != nil {
+		return err
+	}
+
 	if f.dryRun {
+		if loadResult != nil {
+			stopToolSets(loadResult.Team)
+		}
 		out.Println("Dry run mode enabled. Agent initialized but will not execute.")
 		return nil
 	}
 
-	rt, sess, cleanup, err := b.CreateRuntimeAndSession(ctx)
+	wd, _ := os.Getwd()
+	rt, sess, cleanup, err := b.CreateSession(ctx, loadResult, b.CreateSessionRequest(wd))
 	if err != nil {
 		return err
 	}
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -531,8 +531,11 @@ func (f *runExecFlags) createSessionSpawner(agentSource config.Source, sessStore
 		runConfigCopy := f.runConfig.Clone()
 		runConfigCopy.WorkingDir = workingDir
 
-		// Load team with the new working directory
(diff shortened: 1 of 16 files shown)


rumpl:
Conflicts

pkg/runtime/remote_runtime.go
```diff
@@ -111,17 +111,34 @@ func (r *RemoteRuntime) CurrentAgentInfo(ctx context.Context) CurrentAgentInfo {
...
+			if a.Name == agentName {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("agent %q not found in remote team", agentName)
+		}
+	}
 	r.currentAgent = agentName
```
aheritier, line 135: **Blocking** — this re-introduces the silent-gap pattern the PR claims to fix.

Two issues:

1. If `GetAgent` returns an error (network blip, server down, auth failure), validation is skipped entirely and the unknown name is accepted. The doc comment says "validates the name against the remote team config; an unknown agent is rejected" — that contract isn't honoured on the error path. Either propagate the error, or document explicitly that validation is best-effort.
2. `context.Background()` for a network call inside a long-lived runtime means the call can't be cancelled on shutdown. The `Runtime.SetCurrentAgent` interface doesn't take a `ctx`, so fixing this cleanly likely requires widening the interface — worth flagging as a follow-up if not done here.

Suggested minimal fix for (1):
```go
cfg, err := r.client.GetAgent(context.Background(), r.agentFilename)
if err != nil {
    return fmt.Errorf("validate agent %q: %w", agentName, err)
}
```

pkg/runtime/payload.go
```diff
@@ -0,0 +1,46 @@
...
+//
+// GlobalPermissions is currently json:"-" (the permissions.Checker is an
+// opaque struct without a wire form yet); it'll get a serializable
+// representation in a follow-up.
+type CreateSessionRequest struct {
+	AgentName         string               `json:"agent_name,omitempty"`
+	ToolsApproved     bool                 `json:"tools_approved,omitempty"`
+	HideToolResults   bool                 `json:"hide_tool_results,omitempty"`
+	SessionDB         string               `json:"session_db,omitempty"`
+	ResumeSessionID   string               `json:"resume_session_id,omitempty"`
```
aheritier, line 48: **Non-blocking** — `omitempty` on `bool` fields (`ToolsApproved`, `HideToolResults`, `SnapshotsEnabled`) makes "explicitly false" indistinguishable from "unset" on the wire. Fine while server defaults all match `false`, but worth a note in the doc comment so a future server-side default flip doesn't silently change client semantics. `*bool` or a separate `Has*` mask would preserve the distinction if you anticipate that.

pkg/runtime/wire/server.go
```diff
@@ -0,0 +1,110 @@
...
+		writeJSON(w, http.StatusOK, resp)
+	})
+	return mux
+}
+
+func decodeJSON(r *http.Request, v any) error {
+	dec := json.NewDecoder(r.Body)
+	dec.DisallowUnknownFields()
+	return dec.Decode(v)
+}
```
aheritier, line 103: **Non-blocking** — no `http.MaxBytesReader` cap on the decoded body. The scaffold is in-process today so it doesn't matter, but the moment this is fronted by a real listener an unbounded JSON POST is a trivial DoS. Worth wiring in before the streaming endpoint lands.

cmd/root/run.go
```diff
@@ -263,12 +263,21 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		return err
 	}
 
+	loadResult, err := b.LoadTeam(ctx, b.LoadTeamRequest())
```
aheritier, line 266: **Praise** — splitting `LoadTeam` out of `CreateRuntimeAndSession` so `--dry-run` exercises the team load is a nice side-effect; it catches YAML typos at dry-run time without changing the remote behaviour. Worth calling out in a release note.

aheritier (dismissed):
Solid set of baby steps — interface consolidation reads well, the contract test surfacing the previous silent gaps is exactly the right pattern, and the wire scaffold is appropriately minimal.

One blocking item on `RemoteRuntime.SetCurrentAgent` (silent-fallback on `GetAgent` error re-introduces exactly the silent-gap pattern this PR is closing). Two non-blocking notes on payload `omitempty` semantics and missing body size cap on the wire handler.

dgageot:
Rebased onto `origin/main` (no conflicts on the current `main`) and addressed the three review comments:

- **🔴 `RemoteRuntime.SetCurrentAgent` silently accepted errors** — `ed5f65217` propagates the `GetAgent` error instead of swallowing it. `currentAgent` is now only mutated once the agent has been confirmed to exist in the remote team. New test `TestRemoteRuntime_SetCurrentAgent_PropagatesClientError` pins the contract.
- **🟡 `omitempty` on `CreateSessionRequest` booleans** — `fb7cafacd` drops `omitempty` from `ToolsApproved`, `HideToolResults` and `SnapshotsEnabled` so an explicit `false` is always serialised on the wire. Updated `TestCreateSessionRequest_OmitsEmptyFields` and added `TestCreateSessionRequest_PreservesExplicitFalseOnTheWire` to pin the new contract.
- **🟡 Missing `http.MaxBytesReader` in the wire server** — `43fa298f7` caps request bodies at 1 MiB (orders of magnitude above the largest legitimate request today). New test `TestLoadTeam_RejectsOversizedBody` confirms over-cap bodies are rejected with 4xx and never reach the backend.

`task lint` and `task test` pass locally on the rebased branch.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25664392909).

rumpl (approved):

PR #2724, opened 2026-05-09
feat(lint): add Concurrency/CancelFuncMustBeReal cop

dgageot:
Adds one custom cop, `Concurrency/CancelFuncMustBeReal`, that reports two AST
shapes which compile and type-check but defeat cancellation at runtime:

1. an empty `func() {}` literal assigned to a value whose type is
   syntactically `context.CancelFunc` — either through a var/const with
   explicit type annotation or a composite-literal field whose declared
   type is `context.CancelFunc` in the same file;
2. a `context.WithCancel` / `WithCancelCause` / `WithDeadline[Cause]` /
   `WithTimeout[Cause]` call whose `CancelFunc` return is dropped onto
   the blank identifier (e.g. `_, _ = context.WithCancel(ctx)`).

The cop is purely syntactic, ~150 lines, no type info, no false positives
on the current tree.

### What it would have caught

**[PR #2714 — `AttachRuntime` stores a no-op cancel](https://github.com/docker/docker-agent/pull/2714)**, a MEDIUM finding from the bot reviewer:

> `AttachRuntime` initialises `cancel` as `func() {}`. When `DeleteSession`
> is called it invokes `sessionRuntime.cancel()`, which is a no-op for any
> attached runtime — the streaming goroutine keeps running until the
> parent context is cancelled.

Pre-fix code (commit d8a5494d5):

```go
sm.runtimeSessions.Store(sessionID, &activeRuntimes{
    runtime: rt,
    cancel:  func() {},   // ← cop fires here: shape #1
    session: sess,
})
```

`activeRuntimes.cancel` is declared `context.CancelFunc`, so the cop's
shape-1b check (composite-literal field whose declared type in the same
file is `context.CancelFunc`) would have flagged this at the diff-write
moment with:

> *empty func() {} assigned to context.CancelFunc field
> activeRuntimes.cancel — this looks like a cancel but cancels nothing*

### Why a custom cop and not `go vet`

`govet/lostcancel` covers the case where a cancel returned by
`context.WithCancel` is bound and never called. It does **not** cover
the cases this cop catches:

- a `func() {}` literal stored in a CancelFunc field — the call to it
  *exists*, it just does nothing;
- `_, _ = context.WithCancel(...)` — the cancel is never bound at all,
  so there is nothing for `lostcancel` to follow.

Together with `lostcancel`, this closes the loop on the
"compiles, type-checks, runs without error, leaks goroutines" class.

### Cost

- 0 existing offenses (rumpl's PR #2714 fix had already pulled the
  codebase clean).
- 1 cop registration in `lint/main.go`.
- 1 unit-test file with 8 cases, including coverage for the
  scope-narrowing rules (test files exempt, named-non-bare type not
  flagged, untyped vars not flagged).

The cop is preventive: its job is to make sure the PR #2714 fix doesn't
regress, and to catch the same shape if a sibling helper repeats the
mistake.

### DIFF
--- a/lint/concurrency_cancel_func_must_be_real.go
+++ b/lint/concurrency_cancel_func_must_be_real.go
@@ -1,0 +1,250 @@
+package main
+
+import (
+	"go/ast"
+
+	"github.com/dgageot/rubocop-go/cop"
+)
+
+// ConcurrencyCancelFuncMustBeReal flags two AST shapes that have shown
+// up in PR review (e.g. PR #2714) as silent "I look like I cancel but I
+// don't" bugs:
+//
+//  1. an empty function literal `func() {}` assigned to a value whose
+//     type is syntactically `context.CancelFunc`, either through:
+//
+//     - a var/const declaration with an explicit type annotation:
+//     `var c context.CancelFunc = func() {}`
+//     - a composite literal whose target field is declared
+//     `context.CancelFunc` in the same file:
+//     `S{cancel: func() {}}`
+//
+//  2. a [context.WithCancel] / [context.WithCancelCause] /
+//     [context.WithDeadline] / [context.WithTimeout] call whose cancel
+//     return is dropped onto the blank identifier — either via
+//     `_, _ := context.WithCancel(...)` or by a follow-on
+//     `_ = cancel` statement.
+//
+// Each of these "compiles and runs" but defeats cancellation, leaking
+// the goroutine / context tree forever. The cop is purely syntactic:
+// type aliases of `context.CancelFunc` under a different package name
+// are not recognised, and a cancel that is captured by a closure and
+// then dropped is not detected. In practice both gaps are rare; the
+// shapes above are how the bug actually shipped.
+//
+// Suppression: per-line `//rubocop:disable Concurrency/CancelFuncMustBeReal`
+// with a one-line rationale.
+var ConcurrencyCancelFuncMustBeReal = &cop.Func{
+	Meta: cop.Meta{
+		Name:        "Concurrency/CancelFuncMustBeReal",
+		Description: "context.CancelFunc must be wired up; func() {} placeholders and discarded cancels leak goroutines",
+		Severity:    cop.Error,
+	},
+	Scope: nonTestFileScope,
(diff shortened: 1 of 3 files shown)


lint/concurrency_cancel_func_must_be_real.go
```diff
@@ -0,0 +1,250 @@
...
+							"returned by context.WithCancel, or store nil and nil-check at the "+
+							"call site")
+				}
+			}
+			return true
+		})
+
+		// Shape 1b: composite literal field whose declared type in the
+		// receiving struct is context.CancelFunc.
+		cancelFields := cancelFuncFieldsInFile(p.File)
```
docker-agent, line 65: **[MEDIUM] Shape 1b silently produces false-negatives for cross-file struct definitions**

`cancelFuncFieldsInFile(p.File)` builds its map of `context.CancelFunc`-typed fields by scanning only the declarations in the **current** file. If a struct is defined in one file (e.g., `pkg/runner/types.go`) and a composite literal assigning `func() {}` to its `cancel` field appears in another file in the same package (e.g., `pkg/runner/factory.go`), the cop produces **no finding** — silently passing code that defeats cancellation.

This is an inherent limitation of a per-file AST-only approach and is partially acknowledged in the doc-comment ("purely syntactic"), but it is not explicitly called out as a cross-file gap. Because the bug this cop aims to prevent (PR #2714) involved a struct defined and instantiated in the **same** file, the current scope is adequate for that exact case. However, callers of the struct defined elsewhere will remain unchecked.

Consider adding a comment to `cancelFuncFieldsInFile` noting explicitly:

```go
// Note: only structs defined in this file are included. Composite literals
// that reference structs from other files (even in the same package) are
// not covered by Shape 1b; Shape 1a still catches var-level assignments.
```

This ensures reviewers don't mistake a clean lint run for full cross-file coverage.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR adds a well-structured, purely syntactic AST cop that correctly targets two real bug shapes that defeated context cancellation in PR #2714. The implementation is clean, the test coverage is solid, and the cop is correctly registered. One minor documentation gap worth noting (see inline).

PR #2725, opened 2026-05-09
ci: lint workflow invariants actionlint misses (concurrency, SHA pinning, payload deny-list)

dgageot:
actionlint is already in CI but it does not validate event-payload
schemas — that's how PR #2645 slipped through (`github.event.issue.type`
is not a real field for the `issues: labeled` trigger). This PR adds
a small bash check, run after actionlint in the existing `lint` job,
that encodes three project-specific invariants:

1. **PR-triggered workflows declare a `concurrency:` block** — prevents
   piled-up runs on rapid pushes;
2. **Third-party `uses:` references are pinned by 40-char SHA** —
   defence-in-depth for supply-chain attacks;
3. **Workflows do not reference fields on a post-mortem deny-list**
   (currently `github.event.issue.type`, from PR #2645). Easy to
   extend in `scripts/workflow-lint.sh` whenever a new bad pattern
   shows up in review.

### What it caught immediately

`pr-review-trigger.yml` had no `concurrency:` block — fixed in this
PR by keying the group on the PR number so overlapping runs cancel
per-PR.

### Cost

- 1 step in `ci.yml` (runs in the existing `lint` job, no new job).
- 1 portable bash script (~110 lines, self-tested in a fixture
  `.github/workflows/` tree).
- 0 net additions to the existing CI runtime — actionlint and the
  new script together complete in well under a second.

### DIFF
--- a/scripts/workflow-lint.sh
+++ b/scripts/workflow-lint.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+# scripts/workflow-lint.sh — project-specific invariants on
+# .github/workflows/*.yml that actionlint does not enforce.
+#
+# Three checks, each tied to a documented incident or AGENTS.md entry:
+#
+#   1. concurrency:           every PR-triggered workflow declares a
+#                             concurrency: group (AGENTS.md §
+#                             GitHub Actions);
+#
+#   2. pinned-by-sha:         every third-party `uses:` reference is
+#                             pinned by a 40-char SHA, not a tag/branch
+#                             (AGENTS.md § GitHub Actions);
+#
+#   3. payload-field deny:    no `github.event.X.Y` reference on the
+#                             documented broken list — currently the
+#                             `github.event.issue.type` family from
+#                             PR #2645 (the `issues: labeled` payload
+#                             carries no `type` field, so the workflow
+#                             never runs).
+#
+# Failures print the file:line of the offending construct and the
+# rule that fired. The script exits non-zero on any finding so CI
+# fails the lint job; pre-existing findings should be fixed at
+# source rather than suppressed inline.
+
+set -euo pipefail
+
+# Allow running from any directory: prefer cwd if it already has
+# .github/workflows, else jump to the project root relative to the
+# script's own location. Tests can `cp` the script next to a fixture
+# tree without needing to exec from the project root.
+if [ -d ".github/workflows" ]; then
+  :
+else
+  cd "$(dirname "$0")/.."
+fi
+
+WORKFLOWS_DIR=".github/workflows"
+errors=0
+
+note() {
+  printf '%s: %s\n' "$1" "$2" >&2
+  errors=$((errors + 1))
+}
+
+# Check 1: PR-triggered workflows declare concurrency.
+#
+# We treat a workflow as PR-triggered if its `on:` block mentions
+# `pull_request` in any of the three forms GitHub Actions accepts:
+#
+#   on:
+#     pull_request:           # mapping form
+#       branches: [main]
+#
+#   on: [pull_request, push]  # compact array form
+#
+#   on:                       # sequence form
+#     - pull_request
+#
+# A `\bpull_request\b` word-boundary match catches all three without
+# matching `pull_request_target` (a separate event with different
+# concurrency semantics). The check is grep-based rather than
+# yq-based so it runs in the lint job without extra deps; the
+# trade-off is a false positive if `pull_request` appears in a
+# comment, which we accept.
+for f in "$WORKFLOWS_DIR"/*.yml "$WORKFLOWS_DIR"/*.yaml; do
+  [ -e "$f" ] || continue
+  if grep -qE '\bpull_request\b' "$f"; then
+    if ! grep -qE '^\s*concurrency:' "$f"; then
+      note "$f" "PR-triggered workflow has no concurrency: block (AGENTS.md § GitHub Actions)"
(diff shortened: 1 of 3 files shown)


scripts/workflow-lint.sh
```diff
@@ -0,0 +1,128 @@
...
+  # any trailing `# version` comment.
+  ref="${content#*uses:}"
+  # Strip leading whitespace (POSIX-style: drop one leading [ \t]).
+  ref="${ref##+([[:space:]])}"
+  # Bash 3.2 / macOS bash 3 doesn't support extglob without setting
+  # it. Fall back to a sed pipe for portability.
+  ref="$(printf '%s' "$ref" | sed -E 's/^[[:space:]]+//; s/[[:space:]]+#.*$//; s/[[:space:]]+$//')"
+
+  # Skip empty / local / re-usable workflow refs.
+  case "$ref" in
```
docker-agent, line 101: **[MEDIUM] `docker/` namespace exemption is documented but not implemented**

The comment at lines 65–68 explicitly states that the `docker/` namespace is exempt from the SHA-pinning check, but the `case` statement only skips `''`, `'./'*`, and `'../'*` — there is no `'docker/'*` arm:

```bash
case "$ref" in
  '' | './'* | '../'*)
    continue
    ;;
esac
```

Any `docker/` action referenced without a 40-char SHA (e.g. `docker/setup-buildx-action@v3`) would be **incorrectly flagged** despite the documented exemption — or conversely, if the intent is for `docker/` actions to also require SHA pinning, the comment is misleading and should be updated. Either way, the code and the comment disagree.

**Suggested fix** — if docker/ should truly be exempt, add it to the case arm:
```bash
case "$ref" in
  '' | './'* | '../'* | 'docker/'*)
    continue
    ;;
esac
```
Or remove the exemption from the comment if docker/ actions are also required to be SHA-pinned (which is arguably the better security posture).

scripts/workflow-lint.sh
```diff
@@ -0,0 +1,128 @@
...
+#   github.event.issue.type           PR #2645 (issues:labeled has no .type field)
+#
+# Add new entries here when a similar bug is reviewed. The check is a
+# literal grep so the entries are easy to read and keep in sync with
+# the post-mortem notes.
+broken_refs=(
+  "github.event.issue.type"
+)
+for ref in "${broken_refs[@]}"; do
+  matches=$(grep -nE -- "$ref" "$WORKFLOWS_DIR"/*.yml "$WORKFLOWS_DIR"/*.yaml 2>/dev/null || true)
```
docker-agent: **[MEDIUM] Check 3 uses regex grep (`-E`) but the comment promises literal matching**

The comment directly above says _"The check is a literal grep"_, but the implementation uses `grep -nE` (extended regex) — the dots in `github.event.issue.type` are regex metacharacters that match **any character** rather than a literal `.`:

```bash
matches=$(grep -nE -- "$ref" "$WORKFLOWS_DIR"/*.yml ...)
```

While `githubXeventYissueZtype` is vanishingly unlikely in a YAML file today, this is a latent bug: any future deny-list entry containing `[`, `(`, `*`, `+`, or `?` will either over-match or cause `grep` to error. The code also doesn't escape the string before passing it as a regex.

**Suggested fix** — use `-F` (fixed-string) to match the comment's guarantee:
```bash
matches=$(grep -nF -- "$ref" "$WORKFLOWS_DIR"/*.yml "$WORKFLOWS_DIR"/*.yaml 2>/dev/null || true)
```

scripts/workflow-lint.sh
```diff
@@ -0,0 +1,128 @@
...
+# Check 1: PR-triggered workflows declare concurrency.
+#
+# We treat a workflow as PR-triggered if its `on:` block contains
+# `pull_request` (with or without sub-options). The check is grep-based
+# rather than yq-based so it runs in the lint job without extra deps;
+# the trade-off is a couple of false positives if `pull_request` shows
+# up in a comment, which we accept.
+for f in "$WORKFLOWS_DIR"/*.yml "$WORKFLOWS_DIR"/*.yaml; do
+  [ -e "$f" ] || continue
+  if grep -qE '^\s*pull_request:' "$f"; then
```
docker-agent: **[MEDIUM] Concurrency check misses compact `on: [pull_request]` array syntax**

The pattern `^\s*pull_request:` requires a trailing colon, so it only matches the expanded YAML map form. A workflow that declares triggers using the compact sequence syntax is silently skipped:

```yaml
# These forms are missed by the current check:
on: [pull_request, push]
on: [pull_request]
```

Both syntaxes are valid and commonly used across public GitHub Actions workflows. A new workflow added using this form would pass the lint check with no `concurrency:` block, defeating the invariant.

**Suggested fix** — match both forms:
```bash
if grep -qE '^\s*pull_request:|on:\s*\[([^]]*,\s*)?pull_request' "$f"; then
```
Or more simply, match `pull_request` anywhere in the `on:` block and accept the low false-positive rate the comment already acknowledges:
```bash
if grep -qE '^\s*pull_request[: \[]' "$f"; then
```

docker-agent (commented):

Three confirmed issues in the newly-added `scripts/workflow-lint.sh` — all in logic that defines the invariants the script is supposed to enforce. The CI integration (`ci.yml`) and the `concurrency:` addition to `pr-review-trigger.yml` look correct.

gtardif (dismissed):

gtardif (approved):

PR #2728, opened 2026-05-09
feat: built-in large tool output handling

### LINKED ISSUE #2722: feat: built-in large tool output handling
### Problem

When an MCP tool returns a large response the agent breaks in one of two ways:

- With the default `max_old_tool_call_tokens` budget, a single large result eats the entire budget and all earlier tool results in the session get silently replaced with `[content truncated]`.
- With `max_old_tool_call_tokens: -1`, the full response stays in the session but overflows the model's context window. Auto-compaction can't help because the oversized result is in the most recent messages, so the agent gets stuck.

Worth noting: builtin tools (shell, openapi, api) each have a hardcoded 30 000-char per-result limit. MCP tools have no per-result limit at all.

### Workaround

Currently I found a workaround using a `tool_response_transform` hook with a shell script that saves large responses to disk and returns a compact pointer with a path the model can read later. With the `shell` toolset also enabled, the agent picks up the modified response and uses `cat` / `grep` / `jq` on the saved file to work with the full data as needed.

**`agent.yaml`**
```yaml
agents:
  root:
    toolsets:
      - type: shell
      - type: mcp
        remote:
          url: http://127.0.0.1:8012/sse
          transport_type: sse
    hooks:
      tool_response_transform:
        - matcher: "*"
          hooks:
            - type: command
              command: /path/to/mcp_save_response.sh
```

**`mcp_save_response.sh`**
```bash
#!/usr/bin/env bash

THRESHOLD=5000
OUTPUT_DIR="/tmp/mcp_responses"
PREVIEW_CHARS=3000

input=$(cat)

tool_response=$(echo "$input" | jq -r '.tool_response // ""')
response_len=${#tool_response}

# Small responses pass through unchanged
if [ "$response_len" -le "$THRESHOLD" ]; then
  exit 0
fi

session_id=$(echo "$input" | jq -r '.session_id // "unknown"')
tool_use_id=$(echo "$input" | jq -r '.tool_use_id // "unknown"')
tool_name=$(echo "$input"  | jq -r '.tool_name   // "unknown"')

mkdir -p "$OUTPUT_DIR"
outfile="${OUTPUT_DIR}/${session_id}_${tool_use_id}.txt"

printf '%s' "$tool_response" > "$outfile"

preview="${tool_response:0:$PREVIEW_CHARS}"

jq -n \
  --arg file    "$outfile"      \
  --arg len     "$response_len" \
  --arg name    "$tool_name"    \
  --arg preview "$preview"      \
  '{
    hook_specific_output: {
      updated_tool_response: "[\($name) response: \($len) chars, full output saved to \($file)]\n\nFirst \($preview | length) chars:\n\($preview)\n\n[Use shell tool to read: cat \($file)]"
    }
  }'
```

It works but requires every agent author to write and wire this up manually.

### Suggestion

It would be great to have this built in, either as an agent configuration option or as a builtin hook similar to how `redact_secrets` works.

yunus25jmi1:
## Summary

Add a new builtin hook that automatically handles large tool responses. When enabled, tool responses exceeding the threshold are saved to disk and replaced with a pointer that the agent can read back using shell tools.

This is a **root-cause fix** for issue #2722.

---

## Problem

When an MCP tool returns a large response the agent breaks in one of two ways:

1. **With default `max_old_tool_call_tokens` budget**: A single large result eats the entire budget and all earlier tool results in the session get silently replaced with `[content truncated]`.

2. **With `max_old_tool_call_tokens: -1`**: The full response stays in the session but overflows the model's context window. Auto-compaction can't help because the oversized result is in the most recent messages, so the agent gets stuck.

Worth noting: builtin tools (shell, openapi, api) each have a hardcoded 30,000-char per-result limit. MCP tools have no per-result limit at all.

---

## Root Cause

MCP tools have **no output limit**, unlike builtin tools (shell, openapi, api) which all have a hardcoded 30,000-char limit. When MCP returns large responses, they flow into the session unmodified.

---

## Solution

Added a new builtin hook `handle_large_tool_output` that:
1. Intercepts tool responses at `EventToolResponseTransform` 
2. When response exceeds threshold, saves full output to disk
3. Replaces response with a pointer showing tool name, char count, file path, preview

### Configuration

```yaml
agents:
  root:
    handle_large_tool_output:
      enabled: true        # default: false
      threshold: 5000      # chars - default
      output_dir: /tmp     # default: os.TempDir()
      preview_size: 3000   # chars in preview - default
```

---

## Files Changed

| File | Change |
|------|--------|
| `agent-schema.json` | Added `handle_large_tool_output` config |
| `pkg/config/latest/types.go` | Added `HandleLargeToolOutputConfig` struct |
| `pkg/agent/agent.go` | Added field and accessor |
| `pkg/agent/opts.go` | Added `WithHandleLargeToolOutput` option |
| `pkg/teamloader/teamloader.go` | Wired config to agent |
| `pkg/hooks/builtins/handle_large_tool_output.go` | New builtin hook |
| `pkg/hooks/builtins/builtins.go` | Registered builtin |
| `pkg/runtime/hooks.go` | Passed config to ApplyAgentDefaults |

---

## Testing

- Build: ✅ Passes
- Lint: ✅ 0 issues
- Docker Image: ✅ Builds

Fixes #2722

### DIFF
--- a/pkg/hooks/builtins/handle_large_tool_output.go
+++ b/pkg/hooks/builtins/handle_large_tool_output.go
@@ -29,7 +29,7 @@ func handleLargeToolOutput(ctx context.Context, in *hooks.Input, args []string) 
 	cfg := parseArgs(args)
 	threshold := cfg.Threshold
 	if threshold == 0 {
-		threshold = 5000
+		threshold = 30000
 	}
 
 	if len(response) <= threshold {
--- a/pkg/hooks/builtins/handle_large_tool_output.go
+++ b/pkg/hooks/builtins/handle_large_tool_output.go
@@ -1,0 +1,91 @@
+package builtins
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/docker/docker-agent/pkg/hooks"
+)
+
+const HandleLargeToolOutput = "handle_large_tool_output"
+
+func handleLargeToolOutput(ctx context.Context, in *hooks.Input, args []string) (*hooks.Output, error) {
+	if in == nil {
+		return nil, nil
+	}
+
+	if in.HookEventName != hooks.EventToolResponseTransform {
+		return nil, nil
+	}
+
+	response, ok := in.ToolResponse.(string)
+	if !ok || response == "" {
+		return nil, nil
+	}
+
+	cfg := parseArgs(args)
+	threshold := cfg.Threshold
+	if threshold == 0 {
+		threshold = 5000
+	}
+
+	if len(response) <= threshold {
+		return nil, nil
+	}
+
+	outputDir := cfg.OutputDir
+	if outputDir == "" {
+		outputDir = os.TempDir()
+	}
+
+	if err := os.MkdirAll(outputDir, 0o750); err != nil {
+		return nil, fmt.Errorf("create output directory: %w", err)
+	}
+
+	filename := fmt.Sprintf("%s_%s.txt", in.SessionID, in.ToolUseID)
+	path := filepath.Join(outputDir, filename)
+
+	if err := os.WriteFile(path, []byte(response), 0o640); err != nil {
+		return nil, fmt.Errorf("write output file: %w", err)
+	}
+
+	previewSize := cfg.PreviewSize
+	if previewSize == 0 {
+		previewSize = 3000
+	}
+	preview := response
+	if len(preview) > previewSize {
+		preview = response[:previewSize]
+	}
+
+	pointer := fmt.Sprintf("[%s response: %d chars, full output saved to %s]\n\nFirst %d chars:\n%s\n\n[Use shell tool to read: cat %s]",
+		in.ToolName, len(response), path, previewSize, preview, path)
+
+	return &hooks.Output{
+		HookSpecificOutput: &hooks.HookSpecificOutput{
+			HookEventName:       hooks.EventToolResponseTransform,
+			UpdatedToolResponse: &pointer,
+		},
+	}, nil
+}
+
+type toolOutputConfig struct {
+	Threshold   int
+	OutputDir   string
+	PreviewSize int
+}
+
+func parseArgs(args []string) toolOutputConfig {
+	if len(args) == 0 {
+		return toolOutputConfig{}
+	}
+
+	var cfg toolOutputConfig
+	if err := json.Unmarshal([]byte(args[0]), &cfg); err != nil {
+		return toolOutputConfig{}
+	}
+	return cfg
+}
--- a/pkg/hooks/builtins/handle_large_tool_output.go
+++ b/pkg/hooks/builtins/handle_large_tool_output.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/docker-agent/pkg/hooks"
 )
@@ -45,7 +46,9 @@ func handleLargeToolOutput(ctx context.Context, in *hooks.Input, args []string) 
 		return nil, fmt.Errorf("create output directory: %w", err)
 	}
 
-	filename := fmt.Sprintf("%s_%s.txt", in.SessionID, in.ToolUseID)
+	filename := fmt.Sprintf("%s_%s.txt",
+		sanitizeFilename(in.SessionID),
+		sanitizeFilename(in.ToolUseID))
 	path := filepath.Join(outputDir, filename)
 
 	if err := os.WriteFile(path, []byte(response), 0o600); err != nil {
@@ -88,4 +91,11 @@ func parseArgs(args []string) toolOutputConfig {
 		return toolOutputConfig{}
 	}
 	return cfg
+}
+
+func sanitizeFilename(name string) string {
+	name = strings.ReplaceAll(name, "..", "__")
+	name = strings.ReplaceAll(name, "/", "_")
+	name = strings.ReplaceAll(name, "\\", "_")
+	return name
 }
--- a/pkg/hooks/builtins/handle_large_tool_output.go
+++ b/pkg/hooks/builtins/handle_large_tool_output.go
@@ -48,7 +48,7 @@ func handleLargeToolOutput(ctx context.Context, in *hooks.Input, args []string) 
 	filename := fmt.Sprintf("%s_%s.txt", in.SessionID, in.ToolUseID)
 	path := filepath.Join(outputDir, filename)
 
-	if err := os.WriteFile(path, []byte(response), 0o640); err != nil {
+	if err := os.WriteFile(path, []byte(response), 0o600); err != nil {
 		return nil, fmt.Errorf("write output file: %w", err)
 	}
 
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -434,6 +434,30 @@
           "type": "boolean",
           "description": "When true, the runtime auto-installs the redact_secrets builtin on all three of pre_tool_use (scrubs detected secrets from tool arguments), before_llm_call (scrubs the messages sent to the LLM), and tool_response_transform (scrubs tool output before it reaches event consumers, the persisted session, the post_tool_use hook input, or the next LLM call). The same hook entries can be authored directly in YAML for finer-grained control — see the hooks.tool_response_transform / hooks.before_llm_call / hooks.pre_tool_use sections. Detection uses the docker-agent secretsscan ruleset (GitHub PATs, AWS keys, Stripe / Slack / GitLab tokens, JWTs, private keys, etc.). Each detected span is replaced with the literal '[REDACTED]'."
         },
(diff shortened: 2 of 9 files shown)


aheritier:
**Triage note:** This is a community PR (contributor) implementing feature request #2722. The approach — a new `handle_large_tool_output` builtin hook at `EventToolResponseTransform` — is architecturally consistent with the existing `redact_secrets` pattern.

A few design questions for maintainer review before merging:
1. **Opt-in vs. opt-out**: The PR requires explicit YAML config. Should there be a sensible default threshold (e.g. the existing 30,000-char limit already on builtins)?
2. **Config location**: `handle_large_tool_output` is a top-level agent field vs. a hooks entry — breaks the hooks-as-the-extension-point pattern. Would `hooks.tool_response_transform: [{type: builtin, command: handle_large_tool_output}]` with agent-level options be more consistent?
3. **Tests**: The PR notes build/lint pass but no unit tests for the new builtin are mentioned — same bar as `redact_secrets` / `unload` would require test coverage.

Flagging `help-wanted` — good candidate for an iterative review with the contributor.

yunus25jmi1:
Addressed all three triage concerns:

**1. Threshold aligned to 30,000** — default threshold bumped from 5,000 to 30,000 to match the existing builtin tool limit (shell, openapi, api all cap at 30k chars). This is the sensible default you mentioned.

**2. Config location — same pattern as `redact_secrets`** — `handle_large_tool_output` is a top-level `AgentConfig` field, auto-injected into `hooks.tool_response_transform` via `ApplyAgentDefaults`. This mirrors `redact_secrets` exactly:
- `AgentConfig.RedactSecrets` (bool) → auto-injects 3 hooks
- `AgentConfig.HandleLargeToolOutput` (struct) → auto-injects 1 hook

Same abstraction, same mechanism. If the team wants to move to `hooks.tool_response_transform: [{type: builtin, command: handle_large_tool_output}]` with agent-level options, I'm happy to take that as a v2 refactor — it's a non-trivial change to the injection model and not a blocker for this PR.

**3. Unit tests added** — 13 tests covering:
- Under/over threshold behavior
- File write + pointer format
- Custom output_dir and preview_size
- Fallback to os.TempDir()
- No-op cases (nil input, non-string response, empty response, wrong event)
- Error propagation on write failure
- Builtin registration
- ApplyAgentDefaults injection behavior (enabled/disabled/nil)

Permissions hardened: `0o640` → `0o600` (tool output may contain sensitive data).

yunus25jmi1:
**Security fix pushed** — Path traversal vulnerability resolved.

Added `sanitizeFilename()` to strip `..` and `/` from `SessionID` and `ToolUseID` before path construction. An attacker controlling an MCP tool server could have injected `../` sequences to write files outside the configured output directory.

```go
func sanitizeFilename(name string) string {
    name = strings.ReplaceAll(name, "..", "__")
    name = strings.ReplaceAll(name, "/", "_")
    name = strings.ReplaceAll(name, "\\", "_")
    return name
}
```

Added two tests:
- `TestHandleLargeToolOutputPathTraversalIsBlocked` — verifies malicious paths stay within output directory
- `TestSanitizeFilename` — covers edge cases (slashes, backslashes, path traversal sequences)

This brings the PR to 4 commits covering functionality, tests, threshold alignment, and security.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

1 confirmed high-severity bug (user config silently ignored) and 2 likely issues (filename collision, shell metacharacters in suggested command). See inline comments for details.

pkg/hooks/builtins/handle_large_tool_output.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return &hooks.Output{
+		HookSpecificOutput: &hooks.HookSpecificOutput{
+			HookEventName:       hooks.EventToolResponseTransform,
+			UpdatedToolResponse: &pointer,
+		},
+	}, nil
+}
+
+type toolOutputConfig struct {
```
docker-agent, line 78: **[HIGH] `output_dir` and `preview_size` config values are silently ignored due to JSON key mismatch**

`HandleLargeToolOutputConfig` (in `pkg/config/latest/types.go`) uses snake_case JSON tags:
```go
Threshold   int    `json:"threshold,omitempty"`
OutputDir   string `json:"output_dir,omitempty"`
PreviewSize int    `json:"preview_size,omitempty"`
```

But `toolOutputConfig` (used to unmarshal the config at runtime) has **no JSON tags**, so Go uses the PascalCase field names as keys (`"Threshold"`, `"OutputDir"`, `"PreviewSize"`):
```go
type toolOutputConfig struct {
    Threshold   int
    OutputDir   string
    PreviewSize int
}
```

In `ApplyAgentDefaults`, the `HandleLargeToolOutputConfig` is marshalled to JSON and passed as a string arg — producing e.g. `{"threshold":5000,"output_dir":"/my/dir","preview_size":2000}`. When `parseArgs` unmarshals this into `toolOutputConfig`:

- `"threshold"` → matches `Threshold` ✅ (Go's case-insensitive match works here)
- `"output_dir"` → **does NOT match** `OutputDir` ❌ (underscore vs no-underscore; case-insensitivity can't bridge this)
- `"preview_size"` → **does NOT match** `PreviewSize` ❌ (same reason)

**Result**: Any `output_dir` or `preview_size` a user configures is silently dropped, and the hook always uses the hardcoded defaults (`os.TempDir()` and `3000` chars).

**Fix**: Add json struct tags to `toolOutputConfig` matching those on `HandleLargeToolOutputConfig`:
```go
type toolOutputConfig struct {
    Threshold   int    `json:"threshold"`
    OutputDir   string `json:"output_dir"`
    PreviewSize int    `json:"preview_size"`
}
```

pkg/hooks/builtins/handle_large_tool_output.go
```diff
@@ -0,0 +1,101 @@
...
+	if outputDir == "" {
+		outputDir = os.TempDir()
+	}
+
+	if err := os.MkdirAll(outputDir, 0o750); err != nil {
+		return nil, fmt.Errorf("create output directory: %w", err)
+	}
+
+	filename := fmt.Sprintf("%s_%s.txt",
+		sanitizeFilename(in.SessionID),
```
docker-agent, line 50: **[MEDIUM] Filename collision risk when `SessionID` or `ToolUseID` is empty**

The output filename is built as:
```go
filename := fmt.Sprintf("%s_%s.txt",
    sanitizeFilename(in.SessionID),
    sanitizeFilename(in.ToolUseID))
```

If either `SessionID` or `ToolUseID` is empty (e.g. `sanitizeFilename("") == ""`), multiple tool calls can produce the same filename (e.g., `_.txt` or `_<toolUseID>.txt`). Each subsequent `os.WriteFile` call silently **overwrites** the previous content. When the agent later reads the file, it gets the last response only — the earlier ones are lost without any error.

**Fix**: Validate that both IDs are non-empty before constructing the filename, or incorporate a timestamp/random component to guarantee uniqueness:
```go
if in.SessionID == "" || in.ToolUseID == "" {
    return nil, fmt.Errorf("handle_large_tool_output: missing session or tool-use ID")
}
```

pkg/hooks/builtins/handle_large_tool_output.go
```diff
@@ -0,0 +1,101 @@
...
+	previewSize := cfg.PreviewSize
+	if previewSize == 0 {
+		previewSize = 3000
+	}
+	preview := response
+	if len(preview) > previewSize {
+		preview = response[:previewSize]
+	}
+
+	pointer := fmt.Sprintf("[%s response: %d chars, full output saved to %s]\n\nFirst %d chars:\n%s\n\n[Use shell tool to read: cat %s]",
```
docker-agent, line 67: **[LOW] Shell metacharacters in `SessionID`/`ToolUseID` flow into the suggested `cat` command**

The pointer message shown to the agent includes:
```go
pointer := fmt.Sprintf("...[Use shell tool to read: cat %s]", path)
```

`sanitizeFilename` only strips `..`, `/`, and `\` — it does **not** strip spaces, `$`, backticks, `;`, `|`, `&`, or `()`). If a `SessionID` or `ToolUseID` contains these characters, the generated `cat <path>` string will embed them literally. When the agent's shell tool executes the suggested command, those metacharacters are interpreted by the shell, potentially causing unintended execution.

**Fix**: Either quote the path in the suggested command, or expand `sanitizeFilename` to allowlist only safe characters:
```go
// Option A: shell-quote the path in the message
pointer := fmt.Sprintf("...[Use shell tool to read: cat '%s']", path)

// Option B: allowlist in sanitizeFilename
func sanitizeFilename(name string) string {
    var b strings.Builder
    for _, r := range name {
        if (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') ||
           (r >= '0' && r <= '9') || r == '-' || r == '.' {
            b.WriteRune(r)
        } else {
            b.WriteRune('_')
        }
    }
    return b.String()
}
```

dgageot:

PR #2729, opened 2026-05-09
add bump-config-version skill

dgageot:
Adds a project-level skill that documents how to freeze the current `pkg/config/latest` package as `pkg/config/vN` and bump `latest` to `vN+1`.

### DIFF
--- a/.agents/skills/bump-config-version/SKILL.md
+++ b/.agents/skills/bump-config-version/SKILL.md
@@ -1,0 +1,173 @@
+---
+name: bump-config-version
+description: Freeze the current `pkg/config/latest` package as a numbered, immutable `pkg/config/vN`, then bump `latest` to the next version so new features can land on the work-in-progress schema
+---
+
+# Bump Agent Config Version
+
+The agent config schema is **versioned**. Each numbered package
+(`pkg/config/v0`, `pkg/config/v1`, …, `pkg/config/vN`) is **frozen** and
+must never change. New features always land on `pkg/config/latest`, which
+holds the next, unreleased version.
+
+This skill freezes whatever currently lives in `pkg/config/latest` as
+`pkg/config/vN`, then re-opens `latest` for `vN+1` work. Run it whenever a
+released version of the schema needs to be locked down before the next round
+of breaking changes.
+
+**Note**: Throughout this skill, `N` represents the current version number (e.g., `9`), 
+and `vN` represents the version package name (e.g., `v9`). Replace these placeholders 
+with the actual numbers when executing commands.
+
+## 1. Determine the Current and Next Version
+
+Read the version constant:
+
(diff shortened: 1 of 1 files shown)


.agents/skills/bump-config-version/SKILL.md
```diff
@@ -0,0 +1,173 @@
...
+
+```sh
+cp -R pkg/config/latest pkg/config/vN
+```
+
+Rewrite the package name in all Go files (including tests) under the new directory:
+
+```sh
+# macOS / BSD sed requires -i '' (empty backup extension)
+find pkg/config/vN -name '*.go' -exec sed -i '' 's/^package latest$/package vN/' {} +
```
docker-agent, line 57: **[MEDIUM] `sed -i ''` is macOS/BSD-only and will fail on GNU/Linux**

The command as written:
```sh
find pkg/config/vN -name '*.go' -exec sed -i '' 's/^package latest$/package vN/' {} +
```
uses the BSD `sed -i ''` syntax (empty backup-extension argument). On GNU/Linux, `sed` interprets the empty string `''` as a **separate filename argument** rather than a no-backup flag, so the command errors out. The comment above it even calls this out — but no Linux alternative is given.

**Suggested fix**: provide a portable variant or use `perl -pi`:
```sh
# macOS / BSD sed
find pkg/config/vN -name '*.go' -exec sed -i '' 's/^package latest$/package vN/' {} +
# GNU/Linux sed (no empty-string argument needed)
find pkg/config/vN -name '*.go' -exec sed -i 's/^package latest$/package vN/' {} +
# Portable (works on both)
find pkg/config/vN -name '*.go' -exec perl -pi -e 's/^package latest$/package vN/' {} +
```

docker-agent (commented):

One medium-severity portability issue found in the new skill's shell command.

.agents/skills/bump-config-version/SKILL.md
rumpl: Perl?!?!

docker-agent: Fair point! Perl was overkill for this. A better approach: detect the platform and use the appropriate sed syntax, or use a simpler portable alternative with ex/ed (which ships with vi/vim).

**Platform detection**:
```sh
if sed --version >/dev/null 2>&1; then
  # GNU sed
  find pkg/config/vN -name '*.go' -exec sed -i 's/^package latest$/package vN/' {} +
else
  # BSD/macOS sed
  find pkg/config/vN -name '*.go' -exec sed -i '' 's/^package latest$/package vN/' {} +
fi
```

Or even simpler – just document both variants in the skill and let the developer use the right one for their OS, since this is a manual skill execution anyway.

gtardif (approved):

PR #2732, opened 2026-05-09
xml fallback for llama.cpp models

### LINKED ISSUE #2263: Tool call detection issue with DMR and qwen3-coder-30b-a3b-instruct-gguf:Q3_K_M
Tool call detection issue with DMR and qwen3-coder-30b-a3b-instruct-gguf:Q3_K_M

### Context

Configuration used:

- **Tool**: `docker-agent`
- **Model**: `huggingface.co/unsloth/qwen3-coder-30b-a3b-instruct-gguf:Q3_K_M`
- **Backend**: Docker Model Runner (DMR) → llama.cpp
- **Provider**: `dmr`

### Symptoms

When the agent calls a skill or a tool, instead of executing the action, the model returns raw text containing the tool call:

**Skill with no parameter (`what-time-is-it`):**
```
<function=what-time-is-it> </function> </tool_call>
```

**Skill with a parameter (`brave-news-search`):**
```
<function=brave-news-search>
<parameter=query> small local models </parameter>
</function> </tool_call>
```

**Shell toolset:**
```
<function=shell>
<parameter=cmd> ls -lha </parameter>
</function> </tool_call>
```

These `<function=...>` blocks are displayed as-is in the response and **are never executed**.

### Workaround: 

- Adjust sampling parameters: lower `presence_penalty` to `0.0` and slightly raise `temperature` to improve the coherence of structured outputs (argument JSON):
- Explicit system instruction

```yaml
agents:
  root:
    model: qwen3-coder
    num_history_items: 50
    skills: true
    description: Bob
    instruction: |
      You are Bob, a useful AI expert.

      IMPORTANT — Tool usage rules:
      - When you need to call a tool, use ONLY the structured tool-calling API.
      - Do NOT output tool invocations as raw text blocks such as <function=...> or <tool_call> tags.
      - The tool calling mechanism is handled automatically by the system. Just invoke the tool directly.
      - If a tool is unavailable, say so in plain text. Never fabricate tool output.

    toolsets:
      - type: shell
      - type: filesystem
      - type: fetch

models:
  qwen3-coder:
    provider: dmr
    model: huggingface.co/unsloth/qwen3-coder-30b-a3b-instruct-gguf:Q3_K_M
    temperature: 0.6
    top_p: 0.8
    presence_penalty: 0.0
    max_tokens: 32768

    provider_opts:
      runtime_flags: ["--context_size=65536"]
      top_k: 20
      repetition_penalty: 1.05
```

areebahmeddd:
## summary

closes #2263

* some models served via docker model runner (like qwen3-coder and hermes) ignore the openai function-calling api and instead return tool calls as `<tool_call>...</tool_call>` text, which caused tools to never execute
* added a fallback to detect and extract these xml blocks after streaming completes and convert them into actual tool calls
* suppressed raw xml blocks from showing up in the tui while streaming so they never appear as visible chat output

## ai model disclosure

used vscode copilot (claude sonnet 4.6) to understand the issue context and help generate a fix. changes were self-reviewed and verified via:

```bash
task build
task test
task lint
```

### DIFF
--- a/pkg/runtime/runtime_test.go
+++ b/pkg/runtime/runtime_test.go
@@ -452,6 +452,71 @@ func TestToolCallSequence(t *testing.T) {
 	require.True(t, hasEventType(t, events, &StreamStoppedEvent{}), "Expected StreamStoppedEvent")
 }
 
+// TestXMLToolCallFallback verifies that <tool_call> blocks in text content
+// are extracted as tool calls and not leaked as AgentChoice events.
+func TestXMLToolCallFallback(t *testing.T) {
+	xmlPayload := `<tool_call>
+{"name": "shell_exec", "arguments": {"cmd": "ls -la"}}
+</tool_call>`
+
+	stream := newStreamBuilder().
+		AddContent(xmlPayload).
+		AddStopWithUsage(10, 15).
+		Build()
+
+	sess := session.New(session.WithUserMessage("list files"))
+	events := runSession(t, sess, stream)
+
+	// The XML should be promoted to a PartialToolCall, not remain as plain text.
+	require.True(t, hasEventType(t, events, &PartialToolCallEvent{}), "Expected PartialToolCallEvent from XML extraction")
+
+	// No raw XML should have been emitted as an AgentChoice event.
+	for _, ev := range events {
+		if choice, ok := ev.(*AgentChoiceEvent); ok {
+			require.NotContains(t, choice.Content, "<tool_call>", "XML tool call block must not appear in AgentChoice events")
+		}
+	}
+
+	// Verify the extracted tool call fields.
+	for _, ev := range events {
+		if partial, ok := ev.(*PartialToolCallEvent); ok {
+			require.Equal(t, "shell_exec", partial.ToolCall.Function.Name)
+			require.JSONEq(t, `{"cmd": "ls -la"}`, partial.ToolCall.Function.Arguments)
+		}
+	}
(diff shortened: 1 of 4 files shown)


aheritier:
**Triage note:** DRAFT PR from a first-time contributor — addresses a real and confirmed bug (#2263, `priority:high`). The XML `<tool_call>` fallback approach for llama.cpp models matches the workaround in the original issue.

Things to check in review:
1. **Scope**: The fallback is a streaming-layer concern. Needs to confirm it doesn't regress standard OpenAI function-calling models that also stream through the same path.
2. **TUI suppression**: "suppressed raw xml blocks from showing up in the TUI while streaming" — the mechanism needs scrutiny to ensure it doesn't swallow legitimate XML content in tool outputs.
3. **Test coverage**: No test file mentioned; the DMR provider already has a test suite to extend.

Recommend promoting from DRAFT once the contributor adds tests and addresses the above. Good first issue candidate for collaborative refinement.

areebahmeddd:
sure, i've reviewed the 3 checks 🙂

1. standard function-calling models are unaffected since the fallback exits early when toolcalls already exist, so normal streaming behavior stays the same

2. the tui gate only applies to assistant streaming deltas containing `<tool_call>`, so tool outputs/results are untouched and it resets every handlestream call

3. added coverage for xml parsing cases and full handlestream

dgageot (approved):

PR #2738, opened 2026-05-11
fix: pass fully-qualified provider/model ID to modelcaps.Load

### LINKED ISSUE #2737: Anthropic client passes bare model name to modelcaps.Load, causing image attachments to be dropped
## Bug

`convertDocument` in `pkg/model/provider/anthropic/client.go` passes `c.ModelConfig.Model` (e.g. `claude-sonnet-4-6`) to `modelcaps.Load`, but `Load` requires a fully-qualified `provider/model` identifier (e.g. `anthropic/claude-sonnet-4-6`).

## Root cause

`modelcaps.Load` does:
```go
parts := strings.SplitN(id, "/", 2)
if len(parts) != 2 {
    return nil, fmt.Errorf("invalid model ID: %q", id)
}
```

When the bare model name (no `/`) is passed, it returns an error and falls back to `ModelCapabilities{modelFound: false}`, which means `Supports("image/png") == false`.

## Effect

Every image and PDF document attachment is silently dropped with:
```
level=WARN msg="attachment dropped" reason="model does not support MIME type \"image/png\""
```

...even for models like `claude-sonnet-4-6` that fully support images.

## Location

`pkg/model/provider/anthropic/client.go` line where `convertDocument` is called:
```go
docBlocks, err := convertDocument(ctx, *part.Document, c.ModelConfig.Model)
```

Should be:
```go
docBlocks, err := convertDocument(ctx, *part.Document, c.ModelConfig.Provider+"/"+c.ModelConfig.Model)
```

The same issue likely affects the OpenAI/oaistream provider if it also calls `convertDocument` with a bare model name.

## Notes

- The teamloader correctly uses `modelCfg.Provider+"/"+modelCfg.Model` when calling `modelsStore.GetModel` — the Anthropic client should follow the same pattern.

simonferquel-clanker:
## Problem

`modelcaps.Load` requires a `provider/model` identifier (e.g. `anthropic/claude-sonnet-4-6`) to look up model capabilities in the models.dev database. All provider call sites were passing just the bare model name (e.g. `claude-sonnet-4-6`), so the lookup always missed and image/PDF attachments were silently dropped with `StrategyDrop`.

Fixes #2737.

## Changes

- `pkg/model/provider/anthropic/client.go`: pass `Provider+"/"+Model` to `convertDocument` in the standard (non-beta) path
- `pkg/model/provider/anthropic/beta_converter.go`: same fix for the beta API path
- `pkg/model/provider/openai/client.go`: fix both `oaistream.ConvertMessages` and `convertDocumentToResponseInput` calls
- `pkg/model/provider/dmr/client.go`: fix `oaistream.ConvertMessages` call
- `pkg/model/provider/bedrock/client.go`: fix `convertMessages` call
- `pkg/model/provider/gemini/client.go`: fix `convertMessagesToGemini` call

## Testing

- Existing tests for all affected packages pass: `go test ./pkg/model/provider/... ./pkg/attachment/...`

### DIFF
--- a/pkg/model/provider/anthropic/attachments.go
+++ b/pkg/model/provider/anthropic/attachments.go
@@ -15,25 +15,14 @@ import (
 	"github.com/docker/docker-agent/pkg/modelsdev"
 )
 
-// convertDocument converts a chat.Document to standard Anthropic SDK content blocks
-// (not the Beta API).
+// convertDocumentFromStore converts a chat.Document to standard Anthropic SDK content blocks
+// using an explicit modelsdev.Store for capability lookup.
 //
 // Routing:
 //   - image/* with InlineData → ImageBlockParam (base64 source)
 //   - application/pdf with InlineData → DocumentBlockParam (base64)
 //   - text with InlineText → TextBlockParam with TXTEnvelope
 //   - unsupported / no content → nil (logged as warning)
-func convertDocument(ctx context.Context, doc chat.Document, modelID string) ([]anthropic.ContentBlockParamUnion, error) {
-	store, err := modelsdev.NewStore()
-	if err != nil {
-		// Fall back to conservative text-only caps when the store is unavailable.
-		return convertDocumentWithCaps(ctx, doc, modelcaps.ModelCapabilities{})
-	}
-	return convertDocumentFromStore(ctx, doc, modelID, store)
-}
-
-// convertDocumentFromStore is the store-injectable variant of convertDocument,
-// used by tests to inject a fake in-memory modelsdev.Store.
 func convertDocumentFromStore(ctx context.Context, doc chat.Document, modelID string, store *modelsdev.Store) ([]anthropic.ContentBlockParamUnion, error) {
 	mc := modelcaps.LoadFromStore(store, modelID)
 	return convertDocumentWithCaps(ctx, doc, mc)
--- a/pkg/model/provider/anthropic/attachments.go
+++ b/pkg/model/provider/anthropic/attachments.go
@@ -12,6 +12,7 @@ import (
 	"github.com/docker/docker-agent/pkg/attachment"
 	"github.com/docker/docker-agent/pkg/attachment/modelcaps"
 	"github.com/docker/docker-agent/pkg/chat"
+	"github.com/docker/docker-agent/pkg/modelsdev"
 )
 
 // convertDocument converts a chat.Document to standard Anthropic SDK content blocks
@@ -23,7 +24,18 @@ import (
 //   - text with InlineText → TextBlockParam with TXTEnvelope
 //   - unsupported / no content → nil (logged as warning)
 func convertDocument(ctx context.Context, doc chat.Document, modelID string) ([]anthropic.ContentBlockParamUnion, error) {
-	mc, _ := modelcaps.Load(modelID)
+	store, err := modelsdev.NewStore()
+	if err != nil {
+		// Fall back to conservative text-only caps when the store is unavailable.
+		return convertDocumentWithCaps(ctx, doc, modelcaps.ModelCapabilities{})
+	}
+	return convertDocumentFromStore(ctx, doc, modelID, store)
+}
+
+// convertDocumentFromStore is the store-injectable variant of convertDocument,
+// used by tests to inject a fake in-memory modelsdev.Store.
+func convertDocumentFromStore(ctx context.Context, doc chat.Document, modelID string, store *modelsdev.Store) ([]anthropic.ContentBlockParamUnion, error) {
+	mc := modelcaps.LoadFromStore(store, modelID)
 	return convertDocumentWithCaps(ctx, doc, mc)
 }
 
--- a/pkg/model/provider/anthropic/attachments_test.go
+++ b/pkg/model/provider/anthropic/attachments_test.go
@@ -113,7 +113,7 @@ func TestConvertDocumentAnthropic_StrategyTXT(t *testing.T) {
 		Source:   chat.DocumentSource{InlineText: "## Specification"},
 	}
 
-	blocks, err := convertDocument(t.Context(), doc, "")
+	blocks, err := convertDocumentWithCaps(t.Context(), doc, modelcaps.ModelCapabilities{})
 	require.NoError(t, err)
 	require.Len(t, blocks, 1)
 	require.NotNil(t, blocks[0].OfText)
@@ -129,7 +129,7 @@ func TestConvertDocumentAnthropic_StrategyTXT_Envelope(t *testing.T) {
 		Source:   chat.DocumentSource{InlineText: "some notes"},
 	}
 
-	blocks, err := convertDocument(t.Context(), doc, "")
+	blocks, err := convertDocumentWithCaps(t.Context(), doc, modelcaps.ModelCapabilities{})
 	require.NoError(t, err)
 	require.Len(t, blocks, 1)
 	require.NotNil(t, blocks[0].OfText)
@@ -144,7 +144,7 @@ func TestConvertDocumentAnthropic_Drop_NoContent(t *testing.T) {
 		Source:   chat.DocumentSource{},
 	}
 
-	blocks, err := convertDocument(t.Context(), doc, "")
+	blocks, err := convertDocumentWithCaps(t.Context(), doc, modelcaps.ModelCapabilities{})
 	require.NoError(t, err)
 	assert.Nil(t, blocks, "should be dropped when no inline content")
 }
--- a/pkg/model/provider/anthropic/attachments_test.go
+++ b/pkg/model/provider/anthropic/attachments_test.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/docker/docker-agent/pkg/attachment/modelcaps"
 	"github.com/docker/docker-agent/pkg/chat"
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/model/provider/base"
 	"github.com/docker/docker-agent/pkg/modelsdev"
 )
 
@@ -53,13 +55,15 @@ func TestConvertDocumentAnthropic_StrategyB64_PDF(t *testing.T) {
 }
 
 // TestConvertDocumentAnthropic_QualifiedIDRequired is the regression test for
-// the bug where callers passed a bare model name to convertDocument instead of
-// a "provider/model" qualified identifier, causing modelcaps to miss the model
-// and silently drop all image/PDF attachments.
+// the bug where convertUserMultiContent passed only c.ModelConfig.Model (bare
+// model name) to convertDocument instead of c.ModelConfig.Provider+"/"+c.ModelConfig.Model.
+// When the bare name was used, modelcaps.Load always missed the model and all
+// image/PDF attachments were silently dropped.
 //
-// It calls convertDocumentFromStore directly with an injected fake store so
-// the test exercises the full modelID → LoadFromStore → caps → blocks path
-// without hitting the network.
+// The test constructs a Client with Provider="anthropic" and Model="claude-sonnet-4-6",
+// injects a fake modelsdev.Store, and calls convertUserMultiContent directly.
+// The image block must be present in the output — which only happens if the
+// fully-qualified "anthropic/claude-sonnet-4-6" was used for the caps lookup.
 func TestConvertDocumentAnthropic_QualifiedIDRequired(t *testing.T) {
 	store := modelsdev.NewDatabaseStore(&modelsdev.Database{
 		Providers: map[string]modelsdev.Provider{
@@ -75,24 +79,31 @@ func TestConvertDocumentAnthropic_QualifiedIDRequired(t *testing.T) {
 		},
 	})
 
-	doc := chat.Document{
-		Name:     "photo.jpg",
-		MimeType: "image/jpeg",
-		Source:   chat.DocumentSource{InlineData: minJPEG},
+	c := &Client{
+		Config: base.Config{
+			ModelConfig: latest.ModelConfig{
+				Provider: "anthropic",
+				Model:    "claude-sonnet-4-6",
+			},
+		},
+		modelsStore: store,
 	}
 
-	// Bare model name (the original bug): LoadFromStore misses the model,
-	// capabilities fall back to text-only, image must be dropped.
-	blocksBare, err := convertDocumentFromStore(t.Context(), doc, "claude-sonnet-4-6", store)
-	require.NoError(t, err)
-	assert.Nil(t, blocksBare, "bare model name must not resolve caps: image should be dropped")
+	parts := []chat.MessagePart{
+		{
+			Type: chat.MessagePartTypeDocument,
+			Document: &chat.Document{
+				Name:     "photo.jpg",
+				MimeType: "image/jpeg",
+				Source:   chat.DocumentSource{InlineData: minJPEG},
+			},
+		},
+	}
 
-	// Qualified ID (the fix): LoadFromStore finds the model, vision caps
-	// are set, image must be preserved as a native image block.
-	blocksQualified, err := convertDocumentFromStore(t.Context(), doc, "anthropic/claude-sonnet-4-6", store)
+	blocks, err := c.convertUserMultiContent(t.Context(), parts)
 	require.NoError(t, err)
-	require.Len(t, blocksQualified, 1, "qualified ID must resolve caps: image should be present")
-	assert.NotNil(t, blocksQualified[0].OfImage, "expected native image block for qualified model ID")
+	require.Len(t, blocks, 1, "image must not be dropped when provider+model ID is used for caps lookup")
+	assert.NotNil(t, blocks[0].OfImage, "expected native image block")
 }
 
 func TestConvertDocumentAnthropic_StrategyTXT(t *testing.T) {
--- a/pkg/model/provider/anthropic/attachments_test.go
+++ b/pkg/model/provider/anthropic/attachments_test.go
@@ -54,12 +54,12 @@ func TestConvertDocumentAnthropic_StrategyB64_PDF(t *testing.T) {
 
 // TestConvertDocumentAnthropic_QualifiedIDRequired is the regression test for
 // the bug where callers passed a bare model name to convertDocument instead of
-// a "provider/model" qualified identifier, causing all image/PDF attachments
-// to be silently dropped because modelcaps.Load never found the model.
+// a "provider/model" qualified identifier, causing modelcaps to miss the model
+// and silently drop all image/PDF attachments.
 //
-// It uses LoadFromStore (fake in-memory store) to avoid network calls, and
-// feeds the resulting caps into convertDocumentWithCaps so we can verify the
-// full chain: qualified ID -> vision caps -> image block produced.
+// It calls convertDocumentFromStore directly with an injected fake store so
+// the test exercises the full modelID → LoadFromStore → caps → blocks path
+// without hitting the network.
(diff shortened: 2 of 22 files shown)


simonferquel (changes requested):
I think we are missing indirect calls to it, in the chat package, are'nt we ?

simonferquel (changes requested):
To avoid similar issues in the future, refactor modelcaps to be typesafe:
- intrroduce an ID struct (Provider, Model)
- introduce a NewID constructor (Provider, Model string)
- and a ParseID func (singleParamWithSlashSeparator string) (ID, error)

Make all modelcaps functions taking ids as a parameter take this ID struct instead

dgageot:
@simonferquel are there tests for that?

simonferquel-clanker:
Yes — tests were added as part of the refactor pushed in the latest two commits:

- **`TestLoadFromStore_QualifiedIDRequired`** (regression test, `pkg/attachment/modelcaps/`): directly locks in the #2737 fix by asserting that a bare model name without provider prefix (`NewID("", "claude-3-5-sonnet")`) returns conservative text-only caps, while the fully-qualified ID (`NewID("anthropic", "claude-3-5-sonnet")`) correctly resolves to vision-capable.
- **`TestNewID`**, **`TestParseID_Valid`**, **`TestParseID_Invalid`**: cover the new `modelcaps.ID` type — construction, round-trip via `String()`, and error cases.
- All existing `attachments_test.go` and `messages_test.go` files updated to use the new typed API (`modelcaps.ID{}` / `modelcaps.NewID(...)`).

pkg/attachment/modelcaps/modelcaps.go
```diff
@@ -5,13 +5,47 @@ package modelcaps
...
+func ParseID(s string) (ID, error) {
+	provider, model, ok := strings.Cut(s, "/")
+	if !ok || provider == "" || model == "" {
+		return ID{}, fmt.Errorf("modelcaps: invalid model ID %q: must be in \"provider/model\" format", s)
+	}
+	return ID{Provider: provider, Model: model}, nil
+}
+
+// String returns the "provider/model" representation of the ID.
+func (id ID) String() string {
```
rumpl: Every provider already has an `ID()` method, why do we need this, feels like a duplicate.

simonferquel-clanker: Good question. There are two reasons this isn't a duplicate of `base.Config.ID()`:

1. **Semantics differ.** `base.Config.ID()` uses `DisplayOrModel()` — it returns the user-configured alias when present (e.g. `"anthropic/my-sonnet"`). That's correct for display, routing, and telemetry. `modelcaps.ID` always carries the *resolved* `Model` name (e.g. `"anthropic/claude-sonnet-4-5"`) because models.dev keys are pinned identifiers — a user alias won't be found in the database.

2. **Circular dependency.** `modelcaps` lives in `pkg/attachment/modelcaps` and is imported by every provider package. `base.Config` lives in `pkg/model/provider/base`, which itself imports `modelcaps`. Reusing `base.Config` here would create a cycle. The `ID` type has to live in `modelcaps` (or a shared leaf package) so both sides can reference it — `base.Config.ModelCapsID()` returns a `modelcaps.ID`, and `modelcaps.Load` consumes one.

simonferquel: ID is the new structured type for Provider+Model pair

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25660610853).

pkg/attachment/modelcaps/modelcaps.go
rumpl: Sure, but it's only used for `String()` really, I didn't yet find the place where having a type for this is useful? Gotta look closer at the code I guess

simonferquel: the ID method that existed did return a string like provider/model. It now returns the structured ID. 

The problem is mostly that modelscaps.Load and co took a single string as the identifier, and it led me into thinking something like "opus-4.7" would work. It does not.

simonferquel-clanker: The key benefit is at the *call sites* that build the ID, not at the Load side. Before this PR the code was:

```go
convertDocument(ctx, doc, c.ModelConfig.Model)  // bug: bare model name, no provider
```

With a string parameter there's nothing stopping you from passing just `c.ModelConfig.Model` instead of `c.ModelConfig.Provider + "/" + c.ModelConfig.Model` — that's exactly the bug this PR fixes (#2737).

With the struct type:
```go
convertDocument(ctx, doc, c.ModelCapsID())  // returns modelcaps.ID{Provider: ..., Model: ...}
```

The compiler enforces that you construct an `ID` explicitly with both fields — you can't accidentally slip a bare string through. That's the type-safety win the second reviewer requested.

simonferquel: No, String is just there for string representation of the type. modselscap.Load now takes an ID

simonferquel: Ah, no I see your point: actually ID should belong to the modelsdev package rather than modelscaps, so that the type safety goes down to the modelsdev store (and modelsdev consumer also benefit from it).

rumpl: > the ID method that existed did return

It still exists

simonferquel: Yes, it now returns a type-safe ID instead of a string

simonferquel-clanker: Done — moved in commit ef439d0. `modelsdev.ID` now lives in `pkg/modelsdev/id.go` alongside `NewID`, `ParseID`, `ParseIDOrZero`, and `String()`. `Store.GetModel` takes `modelsdev.ID` directly, so the type safety extends all the way to the store layer. `modelcaps` no longer defines any ID type — it just consumes `modelsdev.ID` in `Load` and `LoadFromStore`.

pkg/rag/strategy/semantic_embeddings.go
```diff
@@ -500,11 +501,15 @@ func humanizeMetadataKey(key string) string {
 }
 
 // calculateSemanticUsageCost calculates cost for semantic LLM usage.
-func calculateSemanticUsageCost(modelsStore modelStore, modelID string, usage *chat.Usage) float64 {
-	if usage == nil || modelsStore == nil || modelID == "" || strings.HasPrefix(modelID, "dmr/") {
+func calculateSemanticUsageCost(modelsStore modelStore, modelIDStr string, usage *chat.Usage) float64 {
```
simonferquel: modelId should be of type modelsdev.ID and provider.Provider.ID() should return a modelsdev.ID

pkg/runtime/runtime.go
```diff
@@ -1024,7 +1024,7 @@ func (r *LocalRuntime) EmitStartupInfo(ctx context.Context, sess *session.Sessio
 	// their costs.
 	if sess != nil && (sess.InputTokens > 0 || sess.OutputTokens > 0) {
 		var contextLimit int64
-		if m, err := r.modelsStore.GetModel(ctx, modelID); err == nil && m != nil {
+		if m, err := r.modelsStore.GetModel(ctx, modelsdev.ParseIDOrZero(modelID)); err == nil && m != nil {
```
simonferquel: could we also make getEffectiveModelID return a modelsdev.ID ?

pkg/rag/strategy/semantic_embeddings.go
simonferquel-clanker: Done in commit 99a88db. `calculateSemanticUsageCost` now takes `modelsdev.ID` directly (no intermediate string). `Provider.ID()` returns `modelsdev.ID` across the full interface — all 50+ call sites updated: slog/fmt use the `Stringer` impl transparently, struct comparisons work directly, and string-consuming functions (AgentInfo, event constructors) call `.String()`.

pkg/runtime/runtime.go
simonferquel-clanker: Done in commit 99a88db. `getEffectiveModelID` (and `getAgentModelID`) now return `modelsdev.ID`. The `GetModel` call at this line now passes the ID directly — no more `ParseIDOrZero` wrapper needed.

simonferquel (changes requested):
THis is to big a change, reset the pr to just fix the original issue, and file an issue to introduce typesafe model IDs (provider + model name struct) trough the whole code base as a change suggestion

simonferquel-clanker:
Reset. Force-pushed to reduce the PR to just the original one-line bugfix (commit 446b4e09 — 6 files, 7 insertions/deletions).

Filed #2743 to track the full type-safe modelsdev.ID refactor (Provider.ID(), Store.GetModel, modelcaps.Load, all downstream callers) as a standalone follow-up.

simonferquel-clanker:
Added regression tests in commit 1dbac3d8 (3 files, no production code changes):

- `modelcaps.TestLoadFromStore_QualifiedIDRequired` — at the modelcaps layer: bare model name `"claude-sonnet-4-6"` returns text-only caps; qualified `"anthropic/claude-sonnet-4-6"` returns vision+pdf caps
- `anthropic.TestConvertDocumentAnthropic_QualifiedIDRequired` — end-to-end through `convertDocumentWithCaps`: image dropped for bare ID, native image block produced for qualified ID
- `oaistream.TestConvertDocument_QualifiedIDRequired` — same for the OpenAI/oaistream path: image dropped for bare ID, image URL part produced for qualified ID

All three use an injected fake `modelsdev.Store` (no network, deterministic).

pkg/model/provider/anthropic/attachments_test.go
```diff
@@ -51,6 +52,49 @@ func TestConvertDocumentAnthropic_StrategyB64_PDF(t *testing.T) {
...
 
+// TestConvertDocumentAnthropic_QualifiedIDRequired is the regression test for
+// the bug where callers passed a bare model name to convertDocument instead of
+// a "provider/model" qualified identifier, causing all image/PDF attachments
+// to be silently dropped because modelcaps.Load never found the model.
+//
+// It uses LoadFromStore (fake in-memory store) to avoid network calls, and
+// feeds the resulting caps into convertDocumentWithCaps so we can verify the
+// full chain: qualified ID -> vision caps -> image block produced.
+func TestConvertDocumentAnthropic_QualifiedIDRequired(t *testing.T) {
```
simonferquel, line 67: That is not really what we want to test: we want to test that the fully qualified ID is passed when resolving the models caps

pkg/model/provider/oaistream/attachments_test.go
```diff
@@ -53,6 +54,44 @@ func TestConvertDocument_StrategyB64_ImageDropped(t *testing.T) {
 	assert.Nil(t, parts, "image should be dropped for text-only model")
 }
 
+// TestConvertDocument_QualifiedIDRequired is the regression test for the bug
+// where callers passed a bare model name instead of a "provider/model" ID,
+// causing modelcaps.Load to miss every model and silently drop image/PDF.
+func TestConvertDocument_QualifiedIDRequired(t *testing.T) {
```
simonferquel, line 63: Same here: we want to validate that the fully qualified model ID is passed when resolving caps

pkg/model/provider/anthropic/attachments_test.go
simonferquel-clanker: Done in commit 9e4db925. Added `convertDocumentFromStore(ctx, doc, modelID, store)` to `anthropic/attachments.go` as the store-injectable path; `convertDocument` delegates to it. The test now calls `convertDocumentFromStore` directly, exercising the full `modelID → LoadFromStore → caps → blocks` chain: bare `"claude-sonnet-4-6"` misses the store (text-only, image dropped), qualified `"anthropic/claude-sonnet-4-6"` hits it (vision caps, image block produced).

pkg/model/provider/oaistream/attachments_test.go
simonferquel-clanker: Done in commit 9e4db925. Same pattern applied to oaistream: added `convertDocumentFromStore(ctx, doc, modelID, store)` to `oaistream/attachments.go`; the test calls it directly with a fake store, exercising `modelID → LoadFromStore → caps → parts`: bare `"gpt-4o"` misses the store (image dropped), qualified `"openai/gpt-4o"` hits it (image URL part produced).

pkg/model/provider/anthropic/client.go
```diff
@@ -553,7 +553,7 @@ func (c *Client) convertUserMultiContent(ctx context.Context, parts []chat.Messa
 
 		case chat.MessagePartTypeDocument:
 			if part.Document != nil {
-				docBlocks, err := convertDocument(ctx, *part.Document, c.ModelConfig.Model)
+				docBlocks, err := convertDocument(ctx, *part.Document, c.ModelConfig.Provider+"/"+c.ModelConfig.Model)
```
simonferquel: maybe use c.ID() ?

pkg/model/provider/dmr/client.go
```diff
@@ -147,7 +147,7 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, opts ...options.Opt
 // convertMessages converts chat messages to OpenAI format and merges consecutive
 // system/user messages, which is needed by some local models run by DMR.
 func (c *Client) convertMessages(ctx context.Context, messages []chat.Message) []openai.ChatCompletionMessageParamUnion {
-	openaiMessages := oaistream.ConvertMessages(ctx, messages, c.ModelConfig.Model)
+	openaiMessages := oaistream.ConvertMessages(ctx, messages, c.ModelConfig.Provider+"/"+c.ModelConfig.Model)
```
simonferquel: c.ID()

pkg/model/provider/gemini/client.go
```diff
@@ -601,7 +601,7 @@ func (c *Client) CreateChatCompletionStream(
 		}
 	}
 
-	contents := convertMessagesToGemini(ctx, messages, c.ModelConfig.Model)
+	contents := convertMessagesToGemini(ctx, messages, c.ModelConfig.Provider+"/"+c.ModelConfig.Model)
```
simonferquel: c.ID()

pkg/model/provider/openai/client.go
```diff
@@ -613,7 +613,7 @@ func (c *Client) convertMessagesToResponseInput(ctx context.Context, messages []
 						}
 					case chat.MessagePartTypeDocument:
 						if part.Document != nil {
-							docParts, err := convertDocumentToResponseInput(ctx, *part.Document, c.ModelConfig.Model)
+							docParts, err := convertDocumentToResponseInput(ctx, *part.Document, c.ModelConfig.Provider+"/"+c.ModelConfig.Model)
```
simonferquel: c.ID()

pkg/model/provider/anthropic/client.go
simonferquel-clanker: Done in commit bda5311d — using `c.ID()` across all call sites.

pkg/model/provider/dmr/client.go
simonferquel-clanker: Done in commit bda5311d — using `c.ID()` across all call sites.

pkg/model/provider/gemini/client.go
simonferquel-clanker: Done in commit bda5311d — using `c.ID()` across all call sites.

pkg/model/provider/openai/client.go
simonferquel-clanker: Done in commit bda5311d — using `c.ID()` across all call sites.

pkg/model/provider/anthropic/attachments.go
```diff
@@ -23,7 +24,18 @@ import (
 //   - text with InlineText → TextBlockParam with TXTEnvelope
 //   - unsupported / no content → nil (logged as warning)
 func convertDocument(ctx context.Context, doc chat.Document, modelID string) ([]anthropic.ContentBlockParamUnion, error) {
```
simonferquel: remove that function (only take the one that takes a modelsdev.Store as a parameter. Make sure Client construction always initializes a modelsdev.Store.

Apply the same in oaistream package

simonferquel (changes requested):
Simplify the tests: I do not care about making sure not passing the full id skips images, I only care about the end to end: client configured with a Provider / Model that has the image caps should not skip images

pkg/model/provider/anthropic/attachments.go
simonferquel-clanker: Done in commit a490b33e. `convertDocument` (bare, network-calling) removed from all provider packages — only `convertDocumentFromStore` remains. All provider clients (`anthropic`, `openai`, `dmr`, `bedrock`, `gemini`) now initialise `modelsStore` in `NewClient` (falling back to an empty store on error). Same applied to oaistream: bare `ConvertMessages`/`ConvertMultiContent` removed; only `ConvertMessagesFromStore`/`ConvertMultiContentFromStore` remain.

simonferquel (dismissed):

rumpl (approved):

pkg/model/provider/openai/client.go
```diff
@@ -150,13 +153,20 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 
 	slog.DebugContext(ctx, "OpenAI client created successfully", "model", cfg.Model)
 
+	store, err := modelsdev.NewStore()
```
rumpl, line 156: All of the clients make their own store, let's drill one down please

rumpl (changes requested):

pkg/model/provider/openai/client.go
simonferquel-clanker: Done in commit 2bc0b53e. Added `options.WithModelsStore(store)` to the options package. `createDirectProvider` in `factory.go` (and `factory_js.go`) now creates the store once and injects it via that option. Each provider's `NewClient` checks `globalOptions.ModelsStore()` first and only falls back to `modelsdev.NewStore()` when called directly (not through the factory), so backward compatibility is preserved.

simonferquel-clanker:
CI failure on 2bc0b53e was the pre-existing flaky `TestAttachedServer_DeleteSessionStopsEventStream` (2-second timing test in `pkg/server`, zero overlap with our changes).

Also pushed commit 24ce3c0d addressing rumpl's 'drill one store down' feedback more directly: instead of creating the store in the factory, the code now passes the store that callers already hold:
- `teamloader`: `getModelsForAgent` and `getFallbackModelsForAgent` already call `modelsdev.NewStore()` before their provider loops — now pass that store via `options.WithModelsStore`
- `runtime/model_switcher.go`: `createProviderFromConfig` extracts the underlying `*modelsdev.Store` from the runtime's `lazyModelStore` (new `rawStore()` helper) and passes it via `options.WithModelsStore`
- `factory.go` / `factory_js.go`: store creation removed (providers still fall back to the singleton `modelsdev.NewStore()` when no store is injected)

simonferquel-clanker:
CI failure on 2bc0b53e was the pre-existing flaky TestAttachedServer_DeleteSessionStopsEventStream (2-second timing test in pkg/server, zero overlap with our changes).

Pushed commit 24ce3c0d addressing rumpl's 'drill one store down' feedback: instead of creating a store in the factory, the code now passes the store that callers already hold. teamloader's getModelsForAgent/getFallbackModelsForAgent already call modelsdev.NewStore() — now pass it via options.WithModelsStore. runtime/model_switcher.createProviderFromConfig extracts the store from lazyModelStore via a new rawStore() helper. factory.go/factory_js.go no longer create a store (providers fall back to the singleton NewStore() when none is injected).

rumpl (approved):

PR #2744, opened 2026-05-11
feat: add fd:// scheme support to server.Listen

dgageot:
## Summary

Add `fd://` scheme support to `server.Listen()` so a parent process can pass a pre-opened listening socket via a file descriptor. This eliminates the need for polling when a client creates the listener before spawning the server.

## Motivation

When the server is a child process, the parent can create the listening socket first and pass the file descriptor via `cmd.ExtraFiles`. The server then uses `fd://3` (or whichever FD) to adopt the listener immediately — no race, no polling.

This is the same mechanism used by Docker Desktop when spawning vpnkit.

## Changes

- **`pkg/server/listen.go`**: Added `fd://` prefix handling in `Listen()` and a new `listenFD()` helper that parses the FD number, wraps it with `os.NewFile` + `net.FileListener`, and returns the listener. Validates against non-numeric and negative FD values.
- **`pkg/server/listen_test.go`** (new): Tests for valid FD, invalid (non-numeric) FD, negative FD, and bogus FD number.

## Supported address schemes

| Scheme | Transport |
|---|---|
| `fd://3` | Inherited file descriptor |
| `unix:///path/to/sock` | Unix domain socket |
| `npipe://\\.\pipe\name` | Windows named pipe |
| `:8080` (anything else) | TCP |

### DIFF
--- a/pkg/server/listen.go
+++ b/pkg/server/listen.go
@@ -2,9 +2,11 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 )
 
@@ -17,7 +19,31 @@ func Listen(ctx context.Context, addr string) (net.Listener, error) {
 		return listenNamedPipe(path)
 	}
 
+	if raw, ok := strings.CutPrefix(addr, "fd://"); ok {
+		return listenFD(raw)
+	}
+
 	return listenTCP(ctx, addr)
+}
+
+func listenFD(raw string) (net.Listener, error) {
+	fd, err := strconv.Atoi(raw)
+	if err != nil {
+		return nil, fmt.Errorf("invalid file descriptor %q: %w", raw, err)
+	}
+	if fd < 3 {
+		return nil, fmt.Errorf("invalid file descriptor %d: must be > 2 (0-2 are stdin/stdout/stderr)", fd)
+	}
+
+	f := os.NewFile(uintptr(fd), "fd://"+raw)
+	defer f.Close() // net.FileListener duplicates the fd; close our copy
+
+	ln, err := net.FileListener(f)
+	if err != nil {
+		return nil, fmt.Errorf("file descriptor %d is not a listener: %w", fd, err)
+	}
+
+	return ln, nil
 }
 
 func listenUnix(ctx context.Context, path string) (net.Listener, error) {
--- a/pkg/server/listen_test.go
+++ b/pkg/server/listen_test.go
@@ -1,0 +1,64 @@
+package server
+
+import (
+	"fmt"
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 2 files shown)


rumpl (dismissed):

pkg/server/listen.go
```diff
@@ -17,9 +19,35 @@ func Listen(ctx context.Context, addr string) (net.Listener, error) {
...
+func listenFD(raw string) (net.Listener, error) {
+	fd, err := strconv.Atoi(raw)
+	if err != nil {
+		return nil, fmt.Errorf("invalid file descriptor %q: %w", raw, err)
+	}
+	if fd < 0 {
+		return nil, fmt.Errorf("invalid file descriptor %d: must be non-negative", fd)
+	}
+
+	f := os.NewFile(uintptr(fd), "fd://"+raw)
```
docker-agent, line 38: **[HIGH] Resource leak: `f` returned by `os.NewFile` is never closed**

`net.FileListener` duplicates the file descriptor internally via `dup(2)` (documented: "It is the caller's responsibility to close f when finished"). The original `*os.File` `f` is never closed on either the success path or the error path, so both the `*os.File` wrapper and its underlying fd leak on every `listenFD` call.

In a long-running server this leaks one fd per call. While the GC finalizer will eventually close it, relying on GC finalization for fd management is a well-known Go anti-pattern and risks fd exhaustion under load.

**Fix**: add `defer f.Close()` immediately after the nil-check:

```go
f := os.NewFile(uintptr(fd), "fd://"+raw)
if f == nil {
    return nil, fmt.Errorf("invalid file descriptor %d", fd)
}
defer f.Close()  // net.FileListener duplicates the fd; close our copy

ln, err := net.FileListener(f)
```

pkg/server/listen.go
```diff
@@ -17,9 +19,35 @@ func Listen(ctx context.Context, addr string) (net.Listener, error) {
...
+	if f == nil {
```
docker-agent: **[MEDIUM] `os.NewFile` nil-guard is dead code; `fd=0` (stdin) silently passes validation**

On Linux/macOS `os.NewFile` never returns `nil` for any non-negative integer — the `if f == nil` branch is unreachable on all major platforms. Because the existing guard only rejects `fd < 0`, `fd=0` (stdin), `fd=1` (stdout), and `fd=2` (stderr) all pass through to `net.FileListener`, which then returns a confusing low-level error instead of a clear validation message.

Suggested improvement: replace the dead nil-check with an explicit guard against well-known non-socket fds:

```go
if fd < 3 {
    return nil, fmt.Errorf("invalid file descriptor %d: must be > 2 (stdin/stdout/stderr)", fd)
}
```

Or at minimum document that `os.NewFile` will not return nil, and drop the guard (or widen it to cover fd=0/1/2).

docker-agent (commented):

1 high-severity resource leak (fd not closed after ) and 1 medium-severity dead nil-guard / missing stdin validation found in .

rumpl (approved):

PR #2746, opened 2026-05-11
refactor: centralize modelsdev.Store creation and inject via RuntimeConfig

dgageot:
Centralizes `modelsdev.Store` creation into a lazy `RuntimeConfig.ModelsDevStore()` accessor so the store is instantiated once and shared everywhere via dependency injection, replacing scattered `modelsdev.NewStore()` calls.

## Changes

- Add `ModelsDevStore()` lazy accessor to `RuntimeConfig` with `sync.Once` and `ModelsDevStoreOverride` for test injection
- Thread the store through provider options (`options.WithModelsDevStore`) so every provider client receives it
- Update `modelcaps.Load` to accept `*modelsdev.Store` directly; remove redundant `LoadFromStore`
- Wire `runtime.WithModelStore` from `RuntimeConfig` into all `runtime.New` callsites (run, mcp, a2a, acp, server)
- Convert `modelsdev.NewStore` from `sync.OnceValues` singleton to a plain function since the singleton behavior is now handled by `RuntimeConfig`
- Remove per-client `modelsStore` fields from anthropic, bedrock, gemini, openai, and dmr providers

### DIFF
--- a/cmd/root/models.go
+++ b/cmd/root/models.go
@@ -14,7 +14,6 @@ import (
 	"github.com/docker/docker-agent/pkg/config"
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/model/provider"
-	"github.com/docker/docker-agent/pkg/modelsdev"
 	"github.com/docker/docker-agent/pkg/telemetry"
 )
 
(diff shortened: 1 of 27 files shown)


rumpl:
lol conflicts

rumpl (approved):

PR #2753, opened 2026-05-11
feat(modelsdev): add WithCache option to override cache file path

rumpl:


dgageot (approved):

docker-agent (commented):

pkg/modelsdev/store.go
```diff
@@ -34,23 +34,48 @@ type Store struct {
...
+// Opt configures a Store created with NewStore.
+type Opt func(*storeOptions)
+
+type storeOptions struct {
+	cacheFile string
+}
+
+// WithCache overrides the path of the on-disk cache file used by the Store.
+// The parent directory will be created if it does not already exist.
+func WithCache(path string) Opt {
```
docker-agent, line 46: **[MEDIUM] `WithCache` accepts directory paths without validation, causing silent cache-write failure**

If a caller passes a directory path (e.g. `WithCache("/tmp/mycache/")`) the `NewStore` call succeeds — `filepath.Dir("/tmp/mycache/")` resolves to `"/tmp"`, so `os.MkdirAll` creates the parent and returns `nil`. But the `Store` now holds a directory as its `cacheFile`. Later, every call to `saveToCache` will attempt `os.WriteFile("/tmp/mycache/", …)`, which fails with *"is a directory"*. That error is only logged as a `slog.WarnContext` warning and never returned to callers, so the `Store` silently operates without any on-disk persistence.

Consider adding a guard in `WithCache` or `NewStore`:
```go
if fi, err := os.Stat(path); err == nil && fi.IsDir() {
    return nil, fmt.Errorf("WithCache: path %q is a directory, not a file", path)
}
```

docker-agent, line 46: **[LOW] `WithCache` with a bare filename resolves relative to the process working directory**

Passing a bare filename (e.g. `WithCache("models_dev.json")`) causes `filepath.Dir("models_dev.json")` to return `"."`, making `os.MkdirAll(".", 0o700)` a no-op. The cache file will then be written to whatever the process's current working directory happens to be at call time — which may be unexpected or even unwritable in certain deployment contexts.

Consider documenting that the path must be absolute, or adding a check and returning an error if the path is not absolute:
```go
if !filepath.IsAbs(path) {
    return nil, fmt.Errorf("WithCache: path %q must be an absolute path", path)
}
```

PR #2754, opened 2026-05-11
refactor: simplify RuntimeConfig by removing dead field and caching env provider

dgageot:
## Summary

- Remove the dead `envProvider` field (written but never read)
- Replace `envProviderLock sync.Mutex` with `sync.Once` caching, matching the existing `ModelsDevStore()` pattern
- Propagate the cached env provider through `Clone()`

This eliminates redundant recomputation on every `EnvProvider()` call and makes the two lazy-init patterns in `RuntimeConfig` consistent.

### DIFF
--- a/pkg/config/runtime.go
+++ b/pkg/config/runtime.go
@@ -16,8 +16,8 @@ type RuntimeConfig struct {
 	Config
 
 	EnvProviderForTests environment.Provider
-	envProvider         environment.Provider
-	envProviderLock     sync.Mutex
+	envProviderCached   environment.Provider
+	envProviderOnce     sync.Once
 
 	ModelsDevStoreOverride *modelsdev.Store
 	modelsDevStore         *modelsdev.Store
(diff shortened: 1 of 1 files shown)


pkg/config/runtime.go
```diff
@@ -48,12 +48,16 @@ type Config struct {
 
 func (runConfig *RuntimeConfig) Clone() *RuntimeConfig {
 	store, storeErr := runConfig.ModelsDevStore()
+	env := runConfig.EnvProvider()
```
docker-agent, line 51: **[MEDIUM] Clone() side-effect: calling EnvProvider() mutates the original's EnvFiles**

`Clone()` now calls `runConfig.EnvProvider()` on line 51. When `EnvProviderForTests` is `nil` (all non-test callers), this triggers `envProviderOnce.Do(...)`, which calls `computedEnvProvider()`. That function mutates the **original** `runConfig.EnvFiles` in-place:

```go
runConfig.EnvFiles, err = environment.AbsolutePaths(runConfig.WorkingDir, runConfig.EnvFiles)
```

This is a new side-effect introduced by this PR — before this change, `Clone()` never called `EnvProvider()`, so paths were never absolutized as a side-effect of cloning.

**Concrete impact:**
- Any caller that holds a reference to the original's `EnvFiles` slice (expecting relative paths) will silently observe absolutized paths after the first `Clone()` call.
- `TestClone_EnvFilesIsolated` passes `["a.env", "b.env"]` without setting `EnvProviderForTests` — this test now silently absolutizes the original's `EnvFiles` as a side-effect of `Clone()`, which was not the case before this PR.

**Suggested fix:** Either skip the `EnvProvider()` call during `Clone()` (and let the clone lazily compute its own provider on first use), or cache the provider on the original before `Clone()` is called so the side-effect is explicit:

```go
// Option A: don't eagerly resolve in Clone(); let the clone compute lazily
// (remove the env := runConfig.EnvProvider() call and envProviderCached: env line)

// Option B: decouple path-absolutization from provider caching
// so computedEnvProvider() does not mutate EnvFiles in-place
```

docker-agent (commented):

1 medium-severity finding in the changed code.

rumpl (approved):

PR #2756, opened 2026-05-11
perf: TUI rendering performance improvements

dgageot:
Five focused performance optimizations for the TUI rendering pipeline:

1. **Store pre-split lines in renderedItem** — `ensureAllItemsRendered()` was re-splitting every cached item's view string on every dirty rebuild. Lines are now split once in `renderItem()` and reused. O(n×m) → O(n).

2. **Targeted invalidation on selection changes** — `Focus()`, `Blur()`, `FocusAt()`, `selectPreviousMessage()`, and `selectNextMessage()` all called `invalidateAllItems()`, wiping the entire render cache. Now only the old/new selected items are invalidated. O(n) → O(1).

3. **Cache URL spans per rendered line** — `updateHoveredURL()` re-parsed URL spans (OSC 8 + plain-text) on every mouse motion event (~60 Hz). Spans are now cached per line index and cleared when rendered lines change. O(m) → O(1) for repeated motions on the same line.

4. **Prefix-sum + binary search for globalLineToMessageLine** — Replaces a linear scan through all views on every mouse event with a pre-computed offset array and `sort.Search`. O(n) → O(log n). Also optimizes `scrollToSelectedMessage()`.

5. **Pre-computed fade styles and cached stripped text in reasoning block** — The fade animation was creating a new color string (`fmt.Sprintf`), `lipgloss.NewStyle()`, and `ansi.Strip()` per tool per tick. Fade styles are now a pre-computed 21-element lookup table (`sync.Once`), and the stripped collapsed view is cached once on tool completion. O(m) → O(1) per frame.

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -2,6 +2,7 @@ package messages
 
 import (
 	"slices"
+	"sort"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -106,6 +107,7 @@ type model struct {
 	renderedLines     []string               // Cached rendered content as lines (avoids split/join per frame)
 	renderedItems     map[int]renderedItem   // Cache of rendered items with positions
 	urlSpans          *urlSpanCache          // Cached URL spans per rendered line
+	lineOffsets       []int                  // Prefix-sum: lineOffsets[i] = starting global line of view i
 	totalHeight       int                    // Total height of all content in lines
 	renderDirty       bool                   // True when rendered content needs rebuild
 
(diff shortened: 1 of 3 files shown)


pkg/tui/components/messages/messages.go
```diff
@@ -992,8 +1011,8 @@ func (m *model) renderItem(index int, view layout.Model) renderedItem {
 	// If this message is being inline edited, render the textarea instead
 	if index == m.inlineEditMsgIndex {
 		rendered := m.renderInlineEditTextarea()
-		height := lipgloss.Height(rendered)
-		return renderedItem{view: rendered, height: height}
+		lines := strings.Split(strings.TrimSuffix(rendered, "\n"), "\n")
```
docker-agent: **[LOW] Inline-edit path in `renderItem` missing empty-string guard before `strings.Split`**

The non-inline-edit branch added a guard in this PR:
```go
var lines []string
if rendered != "" {
    lines = strings.Split(strings.TrimSuffix(rendered, "\n"), "\n")
}
```
…keeping `lines = nil` and `height = 0` when the view renders nothing.

The inline-edit branch replaced the old `lipgloss.Height` call without the same guard:
```go
rendered := m.renderInlineEditTextarea()
// no guard ↓
lines := strings.Split(strings.TrimSuffix(rendered, "\n"), "\n")
return renderedItem{view: rendered, lines: lines, height: len(lines)}
```

If `renderInlineEditTextarea()` ever returns `""` (e.g. at zero content width), `strings.Split("", "\n")` yields `[""]` — a one-element slice — so `height` becomes `1` instead of `0`. That phantom line inflates `totalHeight` and shifts all `lineOffsets` for subsequent items, misaligning mouse-click-to-message mapping and scroll positioning until the next full invalidation.

**Suggested fix:** apply the same guard:
```go
rendered := m.renderInlineEditTextarea()
var lines []string
if rendered != "" {
    lines = strings.Split(strings.TrimSuffix(rendered, "\n"), "\n")
}
return renderedItem{view: rendered, lines: lines, height: len(lines)}
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

Five well-scoped performance optimizations. The prefix-sum + binary search, targeted cache invalidation, and pre-computed fade styles are all correctly implemented. One low-severity latent issue found in the changed code.

pkg/tui/components/messages/messages.go
dgageot: Good catch — added the same empty-string guard to the inline-edit branch in `renderItem`. Pushed in the rebase.

docker-agent: Perfect! I can confirm the guard is now in place at line 1014:

```go
rendered := m.renderInlineEditTextarea()
var lines []string
if rendered != "" {
    lines = strings.Split(strings.TrimSuffix(rendered, "\n"), "\n")
}
return renderedItem{view: rendered, lines: lines, height: len(lines)}
```

This matches the non-inline-edit branch and prevents the phantom-line issue when the textarea renders empty. Thanks for the quick fix!

trungutt (approved):

PR #2757, opened 2026-05-11
feat: improve TUI control plane API for external consumers

dgageot:
## What

Seven improvements to the control plane API, motivated by building an external board integration that drives docker-agent over HTTP + SSE.

### API changes

| Change | Description |
|--------|-------------|
| **StreamStoppedEvent.Reason** | Carries the exit reason (`normal`, `error`, `canceled`, `hook_blocked`, `loop_detected`) so consumers can distinguish completion from crashes. |
| **ErrorEvent.Code** | Machine-readable error classification (`rate_limited`, `context_exceeded`, `model_error`, `tool_failed`, `hook_blocked`, `loop_detected`). |
| **GET /api/sessions/:id/status** | Lightweight state snapshot (title, streaming flag, agent, tokens, message count) for late-joining SSE consumers. |
| **POST /followup response** | Returns `queued_streaming` or `queued_idle` so callers know if the message will be processed immediately. |
| **GET /api/ready** | Blocks until the first session is registered, replacing the poll-every-100ms pattern. Supports `?timeout=`. |
| **DELETE /api/sessions/:id ?wait=true** | Optionally blocks until the stream goroutine fully exits, enabling graceful shutdown without SIGKILL. |
| **Re-emit SessionTitle on stream start** | Existing titles are re-emitted so late joiners get them without an extra API call. |

### Bug fixes found during review

- **WaitStopped ignored context** — client disconnect left the polling loop spinning until timeout.
- **deletedSessions memory leak** — entries were never cleaned up without `?wait=true`.
- **Unbounded `?timeout`** — capped at 5 minutes to prevent goroutine pinning.
- **Reason threading** — simplified from `*string` + named return to pointer param + closure defer.

### TUI fix

- **Success sound on error/cancel** — the chime now only plays on normal completion, not after errors (which already play the failure sound) or user cancels.

### DIFF
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -357,8 +357,11 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 
 	streamChan := make(chan runtime.Event)
 
-	// Check if we need to generate a title
-	needsTitle := sess.Title == "" && len(userMessages) > 0 && titleGen != nil
+	// Snapshot the title under sm.mux before launching the goroutine to
+	// avoid a data race with UpdateSessionTitle, which takes sm.mux and
+	// writes to sess.Title concurrently with the goroutine's read.
+	titleToEmit := sess.Title
+	needsTitle := titleToEmit == "" && len(userMessages) > 0 && titleGen != nil
 
 	go func() {
 		// Defers run LIFO: close(streamChan) last, so by the time the
@@ -373,6 +376,10 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 		// Start title generation in parallel if needed
 		if needsTitle {
 			go sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
+		} else if titleToEmit != "" {
+			// Re-emit the existing title so late-joining SSE consumers
+			// and boards can pick it up without an extra API call.
+			streamChan <- runtime.SessionTitle(sess.ID, titleToEmit)
 		}
 
 		stream := runtimeSession.runtime.RunStream(streamCtx, sess)
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -140,6 +140,35 @@ func (sm *SessionManager) GetSession(ctx context.Context, id string) (*session.S
 	return sess, nil
 }
 
+// GetSessionStatus returns a lightweight snapshot of the session's current
+// runtime state. Designed for late-joining SSE consumers that need to know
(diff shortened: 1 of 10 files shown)


pkg/server/session_manager.go
```diff
@@ -287,6 +398,10 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 		// Start title generation in parallel if needed
 		if needsTitle {
 			go sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
+		} else if sess.Title != "" {
```
docker-agent: **[MEDIUM] Data race on `sess.Title` between re-emit goroutine and `UpdateSessionTitle`**

The goroutine spawned by `RunSession` reads `sess.Title` (here and on the `streamChan <-` line below) **outside** the `sm.mux` lock — the mutex was already released before the goroutine was started. Meanwhile, `UpdateSessionTitle` acquires `sm.mux.Lock()` and writes `rt.session.Title = title` on the same underlying `*session.Session` pointer.

This is an unsynchronized concurrent read/write on a string field, which is a **data race** under the Go memory model. Go's race detector would flag it. The window is small (only during stream startup when a title already exists), but the race is real — any HTTP call to `UpdateSessionTitle` while a session is starting its stream could trigger it.

**Suggested fix**: Snapshot the title into a local variable before launching the goroutine, while still holding `sm.mux`:
```go
titleToEmit := sess.Title  // captured under sm.mux before goroutine starts
go func() {
    // ...
    if needsTitle {
        go sm.generateTitle(...)
    } else if titleToEmit != "" {
        streamChan <- runtime.SessionTitle(sess.ID, titleToEmit)
    }
    // ...
}()
```
This is safe because `sess.ID` is immutable and `titleToEmit` is a value copy captured before the mutex is released.

docker-agent (commented):

Reviewed 839 diff lines across 10 files. The PR is well-structured with good concurrency discipline in most places. One **CONFIRMED** data race was found in newly-added code.

pkg/server/session_manager.go
dgageot: Fixed: snapshot `sess.Title` into a local `titleToEmit` while still holding `sm.mux` and use that snapshot from the goroutine. Verified `go test -race` passes for the server package.

docker-agent: Thanks for the quick fix! The snapshot approach looks good — titleToEmit is now captured under sm.mux and the goroutine reads only the local copy, eliminating the race with UpdateSessionTitle. Nice to hear the race detector is happy.

trungutt (approved):

PR #2758, opened 2026-05-11
Improve ACP support: session management, event handling, and code simplification

dgageot:
## Summary

Improves ACP protocol support with new features, a bug fix, and a code simplification pass.

### New features
- **Session listing** — `ListSessions` queries the SQLite session store so clients can browse past sessions
- **Session resumption** — `ResumeSession` restores a previous session from the store
- **Session title updates** — Auto-generated titles are pushed to the client in real time via `session_info_update`
- **Token usage & cost tracking** — Context window size, tokens used, and estimated cost streamed after each model call
- **Warning & model fallback visibility** — Runtime warnings and model fallback events surfaced to the client
- **Advertised session capabilities** — `Initialize` now declares Close, List, and Resume so clients know they're available

### Bug fix
- **Tool call error status** — Failed tool calls now report `ToolCallStatusFailed` instead of `ToolCallStatusCompleted`
- **Max iterations dialog title** — Fixed `new(fmt.Sprintf(...))` to `&title` so the permission dialog shows the actual message

### Code simplification
- Extracted tool-call helpers (builders, kind mapping, diff extraction, argument parsing, todo/plan mapping) into `toolcall.go`
- Added `sendUpdate` helper to eliminate 12 repetitions of the `SessionNotification` boilerplate
- Added `newRuntime`, `registerSession`, `resolveWorkingDir` helpers to deduplicate `NewSession`/`ResumeSession`
- Net reduction: **−357 lines** from `agent.go` (1036 → 679 lines)

### DIFF
--- a/pkg/acp/agent.go
+++ b/pkg/acp/agent.go
@@ -1,32 +1,28 @@
 package acp
 
 import (
-	"cmp"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
 	"os"
 	"path/filepath"
-	"slices"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/coder/acp-go-sdk"
 
+	"github.com/docker/docker-agent/pkg/agent"
 	"github.com/docker/docker-agent/pkg/config"
 	"github.com/docker/docker-agent/pkg/runtime"
 	"github.com/docker/docker-agent/pkg/session"
 	"github.com/docker/docker-agent/pkg/team"
 	"github.com/docker/docker-agent/pkg/teamloader"
-	"github.com/docker/docker-agent/pkg/tools"
-	"github.com/docker/docker-agent/pkg/tools/builtin/todo"
 	"github.com/docker/docker-agent/pkg/version"
 )
 
-// Agent implements the ACP Agent interface for docker agent
+// Agent implements the ACP Agent interface for docker agent.
 type Agent struct {
 	agentSource  config.Source
 	runConfig    *config.RuntimeConfig
@@ -40,7 +36,7 @@ type Agent struct {
 
 var _ acp.Agent = (*Agent)(nil)
 
-// Session represents an ACP session
+// Session represents an ACP session.
 type Session struct {
 	id         string
 	sess       *session.Session
@@ -49,7 +45,7 @@ type Session struct {
 	workingDir string
 }
 
-// NewAgent creates a new ACP agent
+// NewAgent creates a new ACP agent.
 func NewAgent(agentSource config.Source, runConfig *config.RuntimeConfig, sessionStore session.Store) *Agent {
 	return &Agent{
 		agentSource:  agentSource,
@@ -59,7 +55,7 @@ func NewAgent(agentSource config.Source, runConfig *config.RuntimeConfig, sessio
 	}
 }
 
-// Stop stops the agent and its toolsets
+// Stop stops the agent and its toolsets.
 func (a *Agent) Stop(ctx context.Context) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
@@ -70,12 +66,12 @@ func (a *Agent) Stop(ctx context.Context) {
 	}
 }
 
-// SetAgentConnection sets the ACP connection
+// SetAgentConnection sets the ACP connection.
 func (a *Agent) SetAgentConnection(conn *acp.AgentSideConnection) {
 	a.conn = conn
 }
 
-// Initialize implements [acp.Agent]
+// Initialize implements [acp.Agent].
 func (a *Agent) Initialize(ctx context.Context, params acp.InitializeRequest) (acp.InitializeResponse, error) {
 	slog.DebugContext(ctx, "ACP Initialize called", "client_version", params.ProtocolVersion)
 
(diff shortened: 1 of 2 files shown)


pkg/acp/agent.go
```diff
@@ -110,91 +111,97 @@ func (a *Agent) Initialize(ctx context.Context, params acp.InitializeRequest) (a
...
-		runtime.WithCurrentAgent(defaultAgent.Name()),
-		runtime.WithSessionStore(a.sessionStore),
-	)
-	if err != nil {
-		return acp.NewSessionResponse{}, fmt.Errorf("failed to create runtime: %w", err)
-	}
-
-	// Build session options (title will be set after we have the session ID)
-	sessOpts := []session.Opt{
+	defaultAgent, _ := a.team.DefaultAgent()
```
docker-agent: **[HIGH] Silently ignored `DefaultAgent()` error can cause nil-pointer panic**

`newRuntime()` (line 162) already calls `a.team.DefaultAgent()` internally and returns an error if it fails — so the runtime is only created when `DefaultAgent()` succeeds. Immediately after that, line 167 calls `a.team.DefaultAgent()` a **second time** and discards the error with `_`:

```go
defaultAgent, _ := a.team.DefaultAgent()
sess := session.New(
    session.WithMaxIterations(defaultAgent.MaxIterations()),  // panics if defaultAgent is nil
    ...
)
```

If the team's state changes between the two calls (e.g. a race, config reload, or flaky implementation), `defaultAgent` can be `nil` and the next line panics. The fix is to either:
1. Return the `defaultAgent` from `newRuntime()` so the result is reused, or
2. Handle the error: `defaultAgent, err := a.team.DefaultAgent(); if err != nil { return ..., err }`

pkg/acp/agent.go
```diff
@@ -206,33 +213,87 @@ func (a *Agent) CloseSession(_ context.Context, params acp.CloseSessionRequest)
...
+
+	sessions := make([]acp.SessionInfo, 0, len(summaries))
+	for _, s := range summaries {
+		info := acp.SessionInfo{
+			SessionId: acp.SessionId(s.ID),
+			Title:     &s.Title,
+		}
+		if !s.CreatedAt.IsZero() {
+			updatedAt := s.CreatedAt.UTC().Format(time.RFC3339)
+			info.UpdatedAt = &updatedAt
```
docker-agent: **[MEDIUM] `UpdatedAt` is populated from `CreatedAt` — semantically wrong**

The `session.Summary` struct has no `UpdatedAt` field; it only exposes `CreatedAt`. The code creates a variable named `updatedAt` but gives it the creation timestamp:

```go
if !s.CreatedAt.IsZero() {
    updatedAt := s.CreatedAt.UTC().Format(time.RFC3339)
    info.UpdatedAt = &updatedAt  // ← actually the creation time, not an update time
}
```

Clients relying on `UpdatedAt` to find recently-modified sessions (e.g. for sorting or change detection) will always receive the creation time instead. Either:
- Rename the variable to `createdAt` and populate `info.CreatedAt` (if the ACP `SessionInfo` has that field), **or**
- Add an `UpdatedAt` field to `session.Summary` and the underlying store so the correct timestamp can be tracked.

pkg/acp/agent.go
```diff
@@ -206,33 +213,87 @@ func (a *Agent) CloseSession(_ context.Context, params acp.CloseSessionRequest)
...
+	rt, err := a.newRuntime()
+	if err != nil {
+		return acp.ResumeSessionResponse{}, err
+	}
+
+	a.registerSession(&Session{
+		id:         sid,
+		sess:       sess,
+		rt:         rt,
+		workingDir: workingDir,
```
docker-agent, line 308: **[MEDIUM] Resumed session's `workingDir` field is empty when client omits `Cwd`**

In `ResumeSession`, when the client doesn't provide a `Cwd`, `resolveWorkingDir` returns `""`. The stored `sess.WorkingDir` is only used to update `sess.WorkingDir` — it is **not** propagated into `Session.workingDir`:

```go
workingDir, err := resolveWorkingDir(params.Cwd)  // "" if Cwd not set
if workingDir != "" {
    sess.WorkingDir = workingDir  // only updated when client sends new Cwd
}
// ...
a.registerSession(&Session{
    id:         sid,
    sess:       sess,             // sess.WorkingDir = stored value ✓
    rt:         rt,
    workingDir: workingDir,       // ← "" instead of sess.WorkingDir ✗
})
```

`Session.workingDir` is used in `pkg/acp/filesystem.go` for path resolution:
```go
resolved := filepath.Clean(filepath.Join(t.workingDir, userPath))
absWorkingDir, err := filepath.Abs(t.workingDir)
```

With `workingDir == ""`, `filepath.Abs("")` returns the **process CWD** rather than the session's stored working directory, silently breaking file operations for resumed sessions.

Fix: fall back to the stored value when the client omits `Cwd`:
```go
if workingDir == "" {
    workingDir = sess.WorkingDir
}
```

pkg/acp/agent.go
```diff
@@ -206,33 +213,87 @@ func (a *Agent) CloseSession(_ context.Context, params acp.CloseSessionRequest)
...
-// ResumeSession implements [acp.Agent] (optional, not advertised in capabilities)
-func (a *Agent) ResumeSession(ctx context.Context, _ acp.ResumeSessionRequest) (acp.ResumeSessionResponse, error) {
-	slog.DebugContext(ctx, "ACP ResumeSession called (not supported)")
-	return acp.ResumeSessionResponse{}, errors.New("resume session not supported")
+// ResumeSession implements [acp.Agent].
+func (a *Agent) ResumeSession(ctx context.Context, params acp.ResumeSessionRequest) (acp.ResumeSessionResponse, error) {
+	sid := string(params.SessionId)
+	slog.DebugContext(ctx, "ACP ResumeSession called", "session_id", sid)
+
+	a.mu.Lock()
```
docker-agent, line 276: **[MEDIUM] TOCTOU race in `ResumeSession`: check and registration are not atomic**

The session-existence check is done under the mutex, but the lock is released before the expensive I/O and before `registerSession` re-acquires it:

```go
a.mu.Lock()
if _, ok := a.sessions[sid]; ok {
    a.mu.Unlock()
    return acp.ResumeSessionResponse{}, nil
}
a.mu.Unlock()   // ← lock dropped here

sess, err := a.sessionStore.GetSession(ctx, sid)   // I/O without lock
rt, err := a.newRuntime()                           // more work without lock
a.registerSession(&Session{...})                    // re-acquires lock to insert
```

If two concurrent calls arrive for the same session ID, both can pass the existence check before either one registers. Both will load the session from the store and create a runtime; the second `registerSession` will overwrite the first's entry, leaking the first runtime (no cancel, no cleanup).

Consider tracking in-progress resumes with a secondary set (or a `sync.Map` with a sentinel) so a second concurrent call can wait or short-circuit rather than racing.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

4 issues found in the new session management code (1 high, 3 medium). All confirmed in changed lines.

pkg/acp/agent.go
dgageot: Addressed: `newRuntime` now returns the resolved `*agent.Agent` alongside the runtime so callers get a non-nil default agent or a real error.

pkg/acp/agent.go
dgageot: Added a comment explaining that we deliberately report `CreatedAt` in the ACP `UpdatedAt` field because we do not track session updates yet. Renamed the local var to `createdAt` for clarity.

pkg/acp/agent.go
dgageot: Added a comment documenting that an empty `cwd` is allowed (Zed for instance does not always provide one) and that downstream code falls back to the agent default.

pkg/acp/agent.go
dgageot: Fixed: introduced `registerSessionIfAbsent` that performs the existence check and insert atomically under `a.mu`, replacing the old TOCTOU-prone pattern in `ResumeSession`.

docker-agent: Perfect fix — `registerSessionIfAbsent` closes the race window by holding the lock across the entire check-and-insert, and the caller now handles the case where another goroutine won the race by reusing the existing registration and discarding the speculative runtime. This eliminates both the TOCTOU and the resource leak.

Thanks for the quick turnaround!

gtardif (approved):

PR #2759, opened 2026-05-11
refactor: extract loopState struct to bundle runTurn parameters

dgageot:
Bundle the 15 mutable per-RunStream variables into a single loopState struct. Only loop.go is modified. Pure refactor, no behaviour change.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor correctly bundles 15 mutable per-`RunStream` variables into a `loopState` struct. All field initialisations match their original zero/non-zero values, the struct is created fresh per `RunStream` call (no cross-call state bleed), pointer semantics for `overflowCompactions` are preserved, and the three-way `slices.Concat` ordering in `runTurn` is unchanged. No behavioural regressions found.

gtardif (approved):

rumpl (approved):

PR #2762, opened 2026-05-11
feat(a2a): allow session to be resumed interactively

### LINKED ISSUE #2734: Share sessions between A2A and interactive mode
I'd like to start an agent task non-interactively over A2A, let it finish, and then jump into the same conversation interactively from another terminal with `docker-agent run --session <id>`. That doesn't work today: A2A sessions are never visible to `run --session`.

The cause is that `serve a2a` and `run` use two different session backends:

+ https://github.com/docker/docker-agent/blob/b6575306cd59b85f3ee529bffe6fd21a04e4c7e8/cmd/root/run.go#L121

+ https://github.com/docker/docker-agent/blob/b6575306cd59b85f3ee529bffe6fd21a04e4c7e8/pkg/a2a/server.go#L86-L92

The two stores never see each other. So a session started via `serve a2a` never lands in `~/.cagent/session.db`, never shows up in `run --session` listings, and disappears entirely the moment the `serve a2a` process exits.

### What I'd expect

Be able to do this:

```bash
# Terminal 1 — kick off a task via A2A and let it run to completion.
docker-agent serve a2a ./agent.yaml --listen 127.0.0.1:8082 &

# Send a message with a known contextId.
curl -sS http://127.0.0.1:8082/invoke \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":"1","method":"message/send","params":{"message":{"role":"user","messageId":"m1","contextId":"my-task-001","parts":[{"kind":"text","text":"Investigate the failing test in pkg/foo"}]}}}'

# Wait for completion, then stop the serve process — the work is done.

# Terminal 2 — pick the conversation up interactively.
docker-agent run ./agent.yaml --session my-task-001
```

That second command should drop me into the same conversation, with full prior turns in context, so I can poke around interactively (ask follow-ups, tweak, etc.) without re-priming the agent.

maxcleme:
Closes https://github.com/docker/docker-agent/issues/2734

Store A2A sessions to SQLite.

I threw my agent at the above issue and the changes make sense to me but happy to get your feedback and iterate.

### DIFF
--- a/cmd/root/a2a.go
+++ b/cmd/root/a2a.go
@@ -1,17 +1,21 @@
 package root
 
 import (
+	"path/filepath"
+
 	"github.com/spf13/cobra"
 
 	"github.com/docker/docker-agent/pkg/a2a"
 	"github.com/docker/docker-agent/pkg/cli"
 	"github.com/docker/docker-agent/pkg/config"
+	"github.com/docker/docker-agent/pkg/paths"
 	"github.com/docker/docker-agent/pkg/telemetry"
 )
 
 type a2aFlags struct {
 	agentName  string
 	listenAddr string
+	sessionDB  string
 	runConfig  config.RuntimeConfig
 }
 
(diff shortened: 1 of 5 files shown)


rumpl (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25688404084).

PR #2763, opened 2026-05-12
drop send, watch and proto subcommands

dgageot:
Removes the experimental `send`, `watch`, and `proto` subcommands along with their SSE helper. The `run --listen` control plane is unaffected and can still be driven directly via HTTP.


trungutt (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR removes the experimental `send`, `watch`, `proto`, and `sse` subcommands (and their tests) from `cmd/root`. The change is a pure deletion — no new code is introduced.

All removed symbols have been fully excised: `root.go` registers only the surviving commands, no dangling imports or residual wiring were found, and the remaining source files contain no references to the deleted helpers. The `run --listen` control plane is unaffected.

PR #2764, opened 2026-05-12
feat: add utility helpers for shell commands and config parsing

silvin-lubecki:


pkg/utils/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		// just ignore errors lol
+		return ""
+	}
+	return string(out)
+}
+
+// ParseConfig reads config from a file
+func ParseConfig(path string) map[string]string {
+	data := RunShell("cat " + path) // shell injection vulnerability
```
docker-agent, line 22: **[HIGH] Shell command injection via unsanitized `path` argument in `ParseConfig`**

The `path` parameter is concatenated directly into a shell command string:

```go
data := RunShell("cat " + path)
```

`RunShell` passes this to `exec.Command("bash", "-c", cmd)`, meaning any attacker-controlled or externally influenced `path` value can inject arbitrary shell commands. For example, passing `path = "/etc/passwd; rm -rf /"` would execute both `cat /etc/passwd` and `rm -rf /`.

**Fix**: Use `os.ReadFile(path)` instead of shelling out to `cat`, or at minimum pass the path as a separate argument to avoid injection:
```go
out, err := exec.Command("cat", "--", path).Output()
```

pkg/utils/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	result := make(map[string]string)
+	for _, line := range strings.Split(data, "\n") {
+		parts := strings.Split(line, "=")
+		result[parts[0]] = parts[1] // panic on malformed lines - no bounds check
```
docker-agent, line 26: **[HIGH] Index out-of-range panic on config lines without `=` separator**

After splitting each line on `=`, `parts[1]` is accessed unconditionally:

```go
parts := strings.Split(line, "=")
result[parts[0]] = parts[1] // panic on malformed lines - no bounds check
```

If any line lacks an `=` (blank lines, comment lines, malformed entries — including the trailing newline present in virtually every file), `strings.Split` returns a 1-element slice and `parts[1]` panics with an index-out-of-range runtime error, **crashing the process**.

**Fix**: Guard with a length check:
```go
if len(parts) < 2 {
    continue
}
```

pkg/utils/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	return result
+}
+
+// password is the default admin password
+var password = "admin123" // hardcoded credential
```
docker-agent, line 32: **[HIGH] Hardcoded credential committed to version control**

```go
var password = "admin123" // hardcoded credential
```

This plaintext password is committed to the repository and visible to anyone with read access. It is also used to construct a connection string that is logged to stdout (line 36) and passed as a shell argument (line 37), resulting in triple exposure.

**Fix**: Read credentials from environment variables or a secrets manager at runtime:
```go
password := os.Getenv("DB_PASSWORD")
if password == "" {
    return fmt.Errorf("DB_PASSWORD environment variable not set")
}
```

pkg/utils/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+
+func ConnectDB() {
+	connStr := fmt.Sprintf("postgres://admin:%s@localhost:5432/prod?sslmode=disable", password)
+	fmt.Println(connStr) // logging credentials
```
docker-agent, line 36: **[HIGH] Database credentials logged in plaintext to stdout**

```go
fmt.Println(connStr) // logging credentials
```

`connStr` contains the full PostgreSQL URL including the plaintext password. Printing it to stdout exposes the credential in terminal output, container logs (e.g., `docker logs`, Kubernetes pod logs), and any log aggregation pipeline (CloudWatch, Datadog, Splunk, etc.).

**Fix**: Remove this `fmt.Println` call entirely, or log only a redacted version if debugging is needed.

pkg/utils/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	RunShell("psql " + connStr)
```
docker-agent, line 37: **[HIGH] Database credentials exposed in OS process argument list**

```go
RunShell("psql " + connStr)
```

This passes the full connection string (including the password) as a command-line argument to bash via `exec.Command("bash", "-c", ...)`. On Linux/macOS, all command-line arguments are visible to any local user via `ps aux` or `/proc/<pid>/cmdline`. This makes the database password trivially extractable by any process or user on the same host.

**Fix**: Use a PostgreSQL client library (e.g., `lib/pq`, `pgx`) with proper connection parameters, or at minimum pass credentials via the `PGPASSWORD` environment variable rather than the connection URI.

pkg/utils/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// RunShell executes a shell command and returns the output
+func RunShell(cmd string) string {
+	// TODO: add timeout
+	out, err := exec.Command("bash", "-c", cmd).Output()
+	if err != nil {
```
docker-agent, line 13: **[MEDIUM] All shell command errors silently ignored in `RunShell`**

```go
if err != nil {
    // just ignore errors lol
    return ""
}
```

Errors from `exec.Command(...).Output()` are swallowed and an empty string is returned. Callers like `ParseConfig` receive an empty string and treat it as valid input — this triggers the out-of-range panic on the empty line (Finding 2) or silently produces an empty config map, masking command failures entirely.

**Fix**: Return an error from `RunShell` and propagate it to callers:
```go
func RunShell(cmd string) (string, error) {
    out, err := exec.Command("bash", "-c", cmd).Output()
    if err != nil {
        return "", fmt.Errorf("command %q failed: %w", cmd, err)
    }
    return string(out), nil
}
```

pkg/utils/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+		result[parts[0]] = parts[1] // panic on malformed lines - no bounds check
+	}
+	return result
+}
+
+// password is the default admin password
+var password = "admin123" // hardcoded credential
+
+func ConnectDB() {
+	connStr := fmt.Sprintf("postgres://admin:%s@localhost:5432/prod?sslmode=disable", password)
```
docker-agent, line 35: **[MEDIUM] SSL/TLS explicitly disabled for PostgreSQL connection**

```go
connStr := fmt.Sprintf("postgres://admin:%s@localhost:5432/prod?sslmode=disable", password)
```

`sslmode=disable` forces the connection to transmit all data — including credentials and query results — in plaintext over the network. Even though `localhost` is used here, this setting propagates as a pattern and is explicitly unsafe for any non-loopback network path.

**Fix**: Use `sslmode=require` (or `sslmode=verify-full` for production):
```go
connStr := fmt.Sprintf("postgres://admin:%s@localhost:5432/prod?sslmode=require", password)
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR introduces **5 high-severity** and **2 medium-severity** confirmed bugs in the newly added `pkg/utils/helpers.go`. All findings are in code added by this PR. The file should not be merged in its current state.

PR #2767, opened 2026-05-12
fix: rewrite Vertex AI Model Garden provider to publisher for capability lookups

### LINKED ISSUE #2740: Vertex AI Model Garden: attachment capability lookup fails due to provider mismatch
## Problem

`pkg/model/provider/vertexai/modelgarden.go` routes non-Gemini models by creating an `openai.Client` with the original `ModelConfig` (Provider = `"google"`). When this client calls `oaistream.ConvertMessages`, it passes `modelcaps.NewID("google", modelName)`, e.g. `"google/anthropic.claude-3-5-sonnet-20241022-v2:0"`. 

models.dev does not have entries under the `"google"` provider for Anthropic/Meta/Mistral models — they are listed under their own providers (`"anthropic"`, `"meta"`, etc.). The lookup therefore always misses, and vision/PDF attachments are silently dropped.

## Example

A Vertex AI Model Garden config with `publisher: anthropic` and `model: claude-3-5-sonnet-20241022` should resolve capabilities from `anthropic/claude-3-5-sonnet-20241022`, not `google/claude-3-5-sonnet-20241022`.

## Possible fixes

- In the Model Garden client setup, remap the `ModelConfig.Provider` to match the publisher before constructing the underlying `openai.Client` or `anthropic.Client`
- Add a capability-override mechanism in the agent config (e.g. explicit modality declarations)
- Add Vertex AI model aliases to a local lookup table mapping Vertex model IDs to their models.dev equivalents

## Context

Identified during review of PR #2738 (fix: pass fully-qualified provider/model ID to modelcaps.Load).

dgageot:
Fixes #2740.

When a model runs on Vertex AI Model Garden, the OpenAI/Anthropic clients underneath compute their capability-lookup ID as `Provider + "/" + Model`. With `provider: google`, that yields `google/claude-sonnet-4-20250514` — which doesn't exist in models.dev. The lookup silently fails back to "no capabilities", so attachments and other capability-gated features get dropped.

This rewrites `ModelConfig.Provider` from `google` to the lower-cased `provider_opts.publisher` (e.g. `anthropic`, `meta`, `mistral`) on a clone of the config before handing it to the underlying client. The original config is never mutated, so routing/telemetry keep their `google` view.

Tests cover anthropic/meta publishers, case folding, whitespace trimming, missing/empty publisher (no-op), input-mutation safety, and a regression test pinning the resulting ID to `anthropic/claude-sonnet-4-20250514`.

### DIFF
--- a/pkg/model/provider/vertexai/modelgarden_test.go
+++ b/pkg/model/provider/vertexai/modelgarden_test.go
@@ -4,6 +4,9 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
 )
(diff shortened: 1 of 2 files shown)


docker-agent (commented):

pkg/model/provider/vertexai/modelgarden_test.go
```diff
@@ -72,6 +72,119 @@ func TestIsModelGardenConfig(t *testing.T) {
...
+			name: "anthropic publisher rewrites Provider",
+			cfg: &latest.ModelConfig{
+				Provider:     "google",
+				Model:        "claude-sonnet-4-20250514",
+				ProviderOpts: map[string]any{"publisher": "anthropic"},
+			},
+			wantProvider: "anthropic",
+		},
+		{
+			name: "meta publisher rewrites Provider",
```
docker-agent: **[MEDIUM] Potential double-prefix capability-lookup ID for Meta models**

The test fixture sets `Model: "meta/llama-4-maverick-17b-128e-instruct-maas"` alongside `publisher: "meta"`. After `withPublisherProvider` rewrites `Provider` to `"meta"`, the downstream capability-lookup ID produced by `base.Config.ID()` becomes:

```
"meta" + "/" + "meta/llama-4-maverick-17b-128e-instruct-maas"
= "meta/meta/llama-4-maverick-17b-128e-instruct-maas"
```

This composite key does not exist in models.dev, reproducing the exact silent capability-drop the PR is fixing — just for Meta models instead of the `google/…` case.

The test only asserts `got.Provider == "meta"` and never checks the final composed ID. The `TestWithPublisherProvider_NormalizesIDForCapabilityLookup` regression test covers only Anthropic (whose model names carry no publisher prefix).

**Suggested fix**: Either strip the `meta/` prefix from `Model` before storing it in `withPublisherProvider` (or at the ingestion site), or add a test case like `TestWithPublisherProvider_NormalizesIDForCapabilityLookup` for Meta that asserts the composed ID is `"meta/llama-4-maverick-17b-128e-instruct-maas"` rather than `"meta/meta/..."`. Confirming whether Vertex AI always stores Meta model names with the publisher prefix would also clarify whether this is a live issue.

pkg/model/provider/vertexai/modelgarden_test.go
```diff
@@ -72,6 +72,119 @@ func TestIsModelGardenConfig(t *testing.T) {
...
+		{
+			name: "missing publisher leaves Provider untouched",
+			cfg: &latest.ModelConfig{
+				Provider: "google",
+				Model:    "gemini-2.5-flash",
+			},
+			wantProvider: "google",
+		},
+		{
+			name: "empty publisher leaves Provider untouched",
```
docker-agent, line 138: **[LOW] New tests use `t.Fatal`/`t.Errorf` instead of `testify/assert` (project convention)**

Per AGENTS.md: *"Use `github.com/stretchr/testify/assert` and `require` for assertions."* The new `TestWithPublisherProvider` and `TestWithPublisherProvider_NormalizesIDForCapabilityLookup` functions use raw `t.Fatal` / `t.Errorf` exclusively.

While functionally correct, this is inconsistent with the project convention and produces less readable failure messages compared to `require.NotNil`, `require.NotSame`, and `assert.Equal`.

gtardif (dismissed):

dgageot:
Addressed both review comments in 084d34ba6:

**[MEDIUM] Meta double-prefix lookup ID** — fixed. The remap now distinguishes Anthropic from other publishers:

- `publisher: anthropic` → `Provider = "anthropic"` (Vertex Anthropic uses bare model names; models.dev keys them under `anthropic/<model>`).
- Other publishers → `Provider = "google-vertex"` (Vertex's OpenAI-compat path uses `<publisher>/<model>` model names; models.dev keys them under `google-vertex/<publisher>/<model>`).

Renamed the helper to `withModelsDevProvider` and added a dedicated `modelsDevProvider` mapping function with a doc comment explaining the asymmetry. Added `TestWithModelsDevProvider_CapabilityLookupIDs` with cases for Anthropic, Meta and Mistral that assert the final composed ID exactly (so Meta now produces `google-vertex/meta/llama-...`, not `meta/meta/llama-...`).

**[LOW] testify usage** — fixed. The new tests now use `require.NotNil`, `require.NotSame`, and `assert.Equal` per AGENTS.md.

gtardif (approved):

PR #2768, opened 2026-05-12
fix(toolinstall): recover from panics during auto-install

### LINKED ISSUE #2765: toolinstall: EnsureCommand panics inside singleflight when aqua registry lookup fails
## Description

`toolinstall.EnsureCommand` can **panic** (instead of returning an error) when the aqua registry HTTP fetch fails in certain network-restricted environments. The panic originates inside `singleflight.Group.doCall` and is recovered+repanicked, crashing the process.

## Reproduction

1. Run docker-agent in an environment where outbound HTTPS to `raw.githubusercontent.com` and `api.github.com` is blocked or unreliable (e.g. SSRF-protected network, no GitHub token, restrictive firewall).
2. Use an agent config with an LSP toolset whose command is not on PATH and has no `version:` field (auto-install not disabled):

```yaml
agents:
  root:
    toolsets:
      - type: lsp
        command: typescript-language-server
        args: ["--stdio"]
        file_types: [".ts", ".tsx", ".js", ".jsx"]
```

3. `teamloader.Load` → `createLSPTool` → `toolinstall.EnsureCommand` → `resolve` → `singleflight.Group.Do` → `doInstall` panics.

## Stack trace

```
panic(...)
        runtime/panic.go:860 +0x13a
golang.org/x/sync/singleflight.(*Group).doCall.func1()
        golang.org/x/sync@v0.20.0/singleflight/singleflight.go:170 +0x29b
golang.org/x/sync/singleflight.(*Group).doCall(...)
        golang.org/x/sync@v0.20.0/singleflight/singleflight.go:205 +0x99
golang.org/x/sync/singleflight.(*Group).Do(...)
        golang.org/x/sync@v0.20.0/singleflight/singleflight.go:113 +0x155
github.com/docker/docker-agent/pkg/toolinstall.resolve(...)
        github.com/docker/docker-agent@v1.58.0/pkg/toolinstall/resolver.go:62 +0x139
github.com/docker/docker-agent/pkg/toolinstall.EnsureCommand(...)
        github.com/docker/docker-agent@v1.58.0/pkg/toolinstall/resolver.go:33 +0xf5
github.com/docker/docker-agent/pkg/teamloader.createLSPTool(...)
        github.com/docker/docker-agent@v1.58.0/pkg/teamloader/registry.go:457 +0x5a
```

## Expected behaviour

`EnsureCommand` should return an error when the registry lookup or install fails — never panic. The `createMCPTool` path already handles `EnsureCommand` errors gracefully (logs a warning, falls back to the original command name). `createLSPTool` propagates the error, which is fine, but the panic prevents that error path from being reached.

## Suggested fix

Ensure that every code path inside `doInstall` (and any function it calls — `lookupPackage`, `resolveVersion`, `Registry.Install`, etc.) returns an error on HTTP/network failure instead of panicking. The `singleflight` wrapper recovers panics and repanics them, so any nil-pointer dereference or unchecked error inside the `Do` callback becomes a process-fatal crash.

## Workaround

Set `DOCKER_AGENT_AUTO_INSTALL=false` in the environment to skip the auto-install path entirely.

## Version

docker-agent v1.58.0

dgageot:
Closes #2765.

`toolinstall.EnsureCommand` could crash the process when the aqua registry HTTP fetch failed in network-restricted environments. The panic (e.g. a nil-pointer deref deep in the HTTP/YAML stack) was wrapped by `singleflight` into a `*panicError` and re-raised via `go panic(...)`, which is unrecoverable by callers.

This change wraps `doInstall` in a `safeInstall` helper that uses `defer/recover` to convert any panic into a normal error before it ever reaches singleflight. The existing call sites (`createLSPTool`, `createMCPTool`) already handle errors from `EnsureCommand` gracefully.

Logs the panic value and stack trace at ERROR level so the underlying issue is still diagnosable.

### Tests

Three new tests in `resolver_panic_test.go`, all exercising the real `safeInstall`/`resolve` code paths through a small `doInstallFn` test seam:

- `TestSafeInstall_RecoversFromStringPanic` — explicit `panic("boom")`
- `TestSafeInstall_RecoversFromNilDeref` — runtime nil-pointer dereference (the failure mode reported in the issue)
- `TestResolve_ConcurrentPanic_DoesNotCrash` — 10 concurrent goroutines hit the same singleflight key while installation panics; all must observe an error rather than the process crashing

Verified the regression test fails without the fix (panic propagates) and passes with it.

### DIFF
--- a/pkg/toolinstall/resolver.go
+++ b/pkg/toolinstall/resolver.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime/debug"
 	"strings"
 
 	"golang.org/x/sync/singleflight"
@@ -60,7 +61,7 @@ func resolve(ctx context.Context, command, version string) (string, error) {
 
 	// Use singleflight to deduplicate concurrent installs of the same command.
 	result, err, _ := installGroup.Do(command, func() (any, error) {
-		return doInstall(ctx, command, version)
+		return safeInstall(ctx, command, version)
 	})
 	if err != nil {
 		return "", err
@@ -68,6 +69,27 @@ func resolve(ctx context.Context, command, version string) (string, error) {
 
 	return result.(string), nil
 }
+
+// safeInstall wraps doInstall with panic recovery. Without this,
+// singleflight wraps any panic in *panicError and re-raises it via
+// `go panic(...)` (see golang.org/x/sync/singleflight), which is
+// unrecoverable by callers and crashes the process. Issue #2765.
+func safeInstall(ctx context.Context, command, version string) (path string, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			slog.ErrorContext(ctx, "Panic during tool auto-install",
+				"command", command, "panic", r, "stack", string(debug.Stack()))
+			path = ""
+			err = fmt.Errorf("auto-install for %q panicked: %v", command, r)
+		}
+	}()
+	return doInstallFn(ctx, command, version)
+}
+
+// doInstallFn is the install function safeInstall delegates to. Indirected
+// through a var so tests can swap in a panicking implementation to verify
+// the recover path. Production code never reassigns this.
+var doInstallFn = doInstall
 
 // doInstall performs the actual package resolution and installation.
 func doInstall(ctx context.Context, command, versionRef string) (string, error) {
--- a/pkg/toolinstall/resolver_panic_test.go
+++ b/pkg/toolinstall/resolver_panic_test.go
@@ -1,0 +1,80 @@
+package toolinstall
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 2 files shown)


gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The panic-recovery fix is correctly implemented. The `safeInstall` wrapper using `defer/recover` properly intercepts panics before they reach `singleflight`'s re-panic mechanism, converting them to ordinary errors. Named return values are used correctly, `debug.Stack()` captures the full goroutine stack from within the deferred recover, and the `doInstallFn` indirection is a clean test seam.

The three new tests cover the fix well: string panic recovery, nil-pointer dereference recovery, and concurrent singleflight behavior under panic conditions. The `sync.WaitGroup.Go` call and per-iteration loop variable semantics are both correct for Go 1.26.3. The test global-variable mutation pattern is safe under Go's sequential test runner semantics (no `t.Parallel()` on the new tests, and all existing parallel tests in the package are suspended until the sequential tests complete and restore `doInstallFn` via `t.Cleanup`).

PR #2773, opened 2026-05-12
perf(tui): make streaming chunk rendering linear

dgageot:
Three independent perf wins for the TUI streaming path. Each commit is self-contained, ships its own benchmark, and passes `task test` + `task lint`.

The shared theme: when the LLM streams a long message in many small chunks, we used to do quadratic work per chunk. Each commit removes one of those quadratic factors.

## 1. `perf(tui/message): cache Render output to avoid redundant markdown parses`

`messageModel.Render(width)` was called by both `View()` and `Height()` (and again by hover/scroll updates). Each call re-parsed the entire accumulated assistant message. Memoize the result against `(content, msgType, width, selected, hovered, sameAgent)`.

| | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| uncached | 555,272 | 419,392 | 9,956 |
| cached | **4.5** | **0** | **0** |

Cache invalidates correctly on content/state changes, so streaming still re-renders new chunks; only redundant calls are skipped.

## 2. `perf(app): use strings.Builder when merging streamed events`

`mergeEvents` coalesces consecutive same-kind streaming events (`AgentChoiceEvent`, `AgentChoiceReasoningEvent`, `PartialToolCallEvent`) per throttle window. The original code did `merged.Content = merged.Content + next.Content` inside the loop — O(N²) in the run length. Refactored each merge into a helper that scans ahead, sums lengths, and builds with a single pre-sized `strings.Builder`.

256-chunk merge group:
| | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| before | 173,834 | 1,595,323 | 511 |
| after | **2,239** | **17,248** | **3** |

Includes correctness tests covering same-agent merging, agent-boundary preservation, and tool-call argument-delta concatenation.

## 3. `perf(tui/markdown): incremental renderer for streaming chunks`

New `markdown.IncrementalRenderer` caches the rendered output up to the last "safe" block boundary — a blank line that lies outside any fenced code block and is not flanked by list/blockquote lines (constructs that absorb blank lines into a single block). On each subsequent call only the trailing tail is re-rendered. Wired into `messageModel` so streamed assistant replies benefit transparently.

8.5 KB streaming-benchmark document, fed in 3–5 byte chunks:
| | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| `BenchmarkStreamingFastRenderer` | 519,361,000 | 917,394,140 | 3,434,638 |
| `BenchmarkStreamingIncrementalRenderer` | **293,493,521** | **312,866,876** | **2,280,702** |

That's ~1.75× faster and ~3× less memory per pass through the document. Correctness is pinned by a per-step equivalence test (`TestIncrementalRendererMatchesFullRenderForEachStep`) that asserts every intermediate output matches what a single full render would produce, plus targeted cases for tricky inputs (open code fences, lists, no trailing newline, width changes, content replacement).

## Validation

- [x] `task test` — green
- [x] `task lint` — 0 issues
- [x] All three commits compile and pass tests independently
- [x] Benchmarks numbers above were re-measured on the rebased branch

### DIFF
--- a/pkg/tui/components/markdown/incremental.go
+++ b/pkg/tui/components/markdown/incremental.go
@@ -1,0 +1,327 @@
+package markdown
+
+import "strings"
+
+// IncrementalRenderer is a markdown renderer specialized for streaming use:
+// it remembers the most recently rendered "stable prefix" of the input and,
+// when the next call extends that prefix, re-renders only the new trailing
+// region instead of the whole document.
+//
+// A "stable prefix" here is the portion of the input ending at the last block
+// boundary that the underlying FastRenderer is guaranteed to render the same
+// way regardless of what comes after it: a blank line that lies outside any
+// fenced code block and is not flanked by list or blockquote lines (those
+// constructs absorb blank lines into a single block, so the surrounding
+// content is not yet "frozen").
+//
+// The optimization is correct because the FastRenderer (and CommonMark in
+// general) treats such blank lines as hard breaks between top-level blocks:
+// rendering the document up to a stable boundary and rendering the rest
+// independently produces the same output as rendering the full document,
+// modulo the blank-line separator that joinPrefixAndTail re-inserts.
+//
+// IncrementalRenderer is not safe for concurrent use.
+type IncrementalRenderer struct {
+	width int
+
+	// inputPrefix is the longest stable prefix of the most recent input that
+	// ends at a block boundary. outputPrefix is its rendered counterpart.
+	// Both are empty until the first successful incremental render.
+	inputPrefix  string
+	outputPrefix string
+
+	// fallback is used for the actual rendering work; it is reused across calls
+	// so its parser pool (and chroma caches) stay warm.
+	fallback *FastRenderer
+}
+
+// NewIncrementalRenderer creates a new incremental renderer with the given
+// terminal width.
+func NewIncrementalRenderer(width int) *IncrementalRenderer {
+	return &IncrementalRenderer{
+		width:    width,
+		fallback: NewFastRenderer(width),
+	}
+}
+
+// Render produces styled terminal output for input. When successive calls pass
+// inputs that share a long common prefix (the streaming case), only the suffix
+// is parsed and rendered; the rest is served from the cached output.
+func (r *IncrementalRenderer) Render(input string) (string, error) {
+	if input == "" {
+		r.inputPrefix = ""
+		r.outputPrefix = ""
+		return "", nil
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/tui/components/markdown/incremental.go
```diff
@@ -0,0 +1,327 @@
...
+		trimmed := strings.TrimLeft(line, " \t")
+		switch {
+		case openFence != "":
+			lines = append(lines, true)
+			if strings.HasPrefix(strings.TrimSpace(trimmed), openFence) {
+				openFence = ""
+			}
+		case strings.HasPrefix(trimmed, "```"):
+			lines = append(lines, true)
+			openFence = "```"
```
docker-agent, line 317: **[MEDIUM] `classifyFenceLines` does not track fence length — 4+ backtick fences closed prematurely**

`openFence` is always set to `"``"`"` (exactly 3 backticks) regardless of how many backticks the opening fence actually uses:

```go
case strings.HasPrefix(trimmed, "```"):
    lines = append(lines, true)
    openFence = "```"   // ← always 3 backticks
```

Then the close-detection uses `strings.HasPrefix(strings.TrimSpace(trimmed), openFence)`, which matches any line that *starts with* 3 backticks — including lines that are themselves inside a 4-backtick block.

**CommonMark §4.5** requires the closing fence to be at least as long as the opening one. A content block like:

~~~markdown
````
```code here```
````
~~~

would be misclassified: the inner ```` ``` ```` line would be seen as closing the outer ```` ```` ```` fence, making `stableBoundary` believe the block ended at the wrong line. Subsequent blank lines that actually fall *inside* the unclosed fence region would then be returned as "safe" split boundaries, causing `IncrementalRenderer` to render fragments in isolation that belong together — producing output that diverges from a full render.

**Suggested fix**: record the actual fence opening length and require the closer to be at least that long:
```go
case strings.HasPrefix(trimmed, "```"):
    lines = append(lines, true)
    openFence = strings.TrimRight(trimmed[:countRuneRun(trimmed, '`')], "")
    // or simply: openFenceLen = countLeading(trimmed, '`')
```
Then compare `countLeading(trimmed, '`') >= openFenceLen` in the close check.

docker-agent (commented):

Three performance improvements reviewed: render caching in `messageModel`, `strings.Builder` refactor for `mergeEvents`, and the new `IncrementalRenderer`. The `mergeEvents` refactor and render-cache logic are clean. One bug was found in the `IncrementalRenderer`'s fence-detection helper.

gtardif (approved):

PR #2774, opened 2026-05-12
fix(tui): preserve user_prompt input across tab switches

### LINKED ISSUE #2770: user_prompt: switching tabs while prompt is open loses all entered input
## Summary

Follow-up to #2626. With tab switching during a `user_prompt` now allowed, any input the user has already typed into the prompt is discarded when they switch to another tab and return.

## Expected behavior

When the user switches tabs while a `user_prompt` is open and then returns, the prompt should still be open with any partially-entered input preserved.

## Current behavior

Switching tabs while a `user_prompt` is open clears the input. When the user comes back, they have to re-enter everything from scratch.

## Context

Reported from a TamPA session. The point of allowing tab switching during a `user_prompt` (#2626) is to let the user reference other conversations before responding — losing the in-progress input on return defeats that workflow.

dgageot:
Closes #2770.

When the user switched tabs while a background `user_prompt` (MCP elicitation) dialog was open, `handleSwitchTab` re-stashed only the originating runtime event in the supervisor and closed the dialog. On return, `replayPendingEvent` built a brand-new dialog from that event — discarding any text the user had typed.

This change stashes the live dialog instance (along with its event) per-tab when leaving, and re-opens the same instance on return so its in-progress input survives the round trip. If the agent supersedes the prompt while the user is away, the stale stash is discarded and a fresh dialog is built from the new event.

## Changes

- `pkg/tui/dialog/dialog.go` — add `Manager.TopDialog()` accessor.
- `pkg/tui/tui.go` — add `appModel.stashedDialogs` map + `stashedDialog` type; capture the dialog instance in `handleSwitchTab`; reuse it in `replayPendingEvent` when the event still matches; clear it in `handleCloseTab` and on stale/empty pending events.
- `pkg/tui/tui_stash_dialog_test.go` — three regression tests covering: stashed dialog reopened on match, stale stash discarded after supersession, orphaned stash cleared when no event is pending.
- `pkg/tui/dialog/manager_test.go` and `pkg/tui/tui_exit_test.go` — extend existing tests to cover the new accessor and map.

### DIFF
--- a/pkg/tui/dialog/dialog.go
+++ b/pkg/tui/dialog/dialog.go
@@ -47,6 +47,12 @@ type Manager interface {
 	// background dialog, or nil if the top dialog is not a background dialog
 	// or the dialog stack is empty.
 	TopBackgroundEvent() tea.Msg
+	// TopDialog returns the topmost dialog instance, or nil if the dialog
+	// stack is empty. Used by the app model to stash a background dialog's
+	// live state when the user navigates away from the tab that opened it,
+	// so the same instance (with any in-progress input) can be re-opened on
+	// return.
+	TopDialog() Dialog
 }
 
 // dialogEntry pairs a dialog with its drag offset so the two stay in sync.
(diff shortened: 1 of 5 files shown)


gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The stash-and-restore logic for preserving in-progress dialog input across tab switches is well-structured. The `stashedDialogs` map is properly initialized in `New()` and test helpers, entries are cleaned up in `handleCloseTab` and `replayPendingEvent`, and the `TopDialog()` accessor is correctly nil-guarded. The event-identity check (`stash.event == pendingEvent`) works correctly for the current pointer-typed runtime events.

No high- or medium-severity bugs were found in the changed code.

PR #2775, opened 2026-05-12
fix: two TUI control-plane bugs (SSE cancel, IPv6 listen)

dgageot:
While investigating the `run --listen` control plane that lets external
processes drive a running TUI, two real bugs surfaced. Both are small,
mechanical fixes with regression tests that fail on `main`.

## 1. `StreamSessionEvents` cancelled its own SSE stream synchronously

`pkg/runtime/client.go` was setting up the streaming context like this:

```go
streamCtx, cancel := context.WithTimeout(ctx, timeout)
defer cancel()                          // ← fires when StreamSessionEvents returns
req, _ := http.NewRequestWithContext(streamCtx, ...)
go func() { ... reads resp.Body ... }   // ← outlives the function
return eventChan, nil
```

`defer cancel()` fired as soon as the function returned, killing the
in-flight HTTP request and turning the SSE stream into a one-shot read
(the goroutine reading `resp.Body` would see EOF as soon as the kernel
buffer drained). `StreamSessionEventsWithRetry` masked the symptom by
constantly reconnecting and emitting `connection_lost` /
`connection_restored` events.

The fix moves the cancel into the goroutine that actually owns the
response body, so the stream lives until `ctx` is cancelled, the streaming
timeout fires, or the server closes the connection.

The new `TestClient_StreamSessionEvents_DeliversMultipleEvents` fails on
`main` (only the first event is observed) and passes after the fix; a
companion test confirms cancelling the caller's context still tears the
stream down.

## 2. `Listen` was hard-coded to IPv4

`pkg/server/listen.go` always used `tcp4`, so any IPv6 bind failed with
`"address ::1: no suitable address found"` even on dual-stack hosts.
Switched to `"tcp"` and added IPv4 + IPv6 regression tests (the IPv6 one
skips when the host has no IPv6 loopback).

## Validation

- `task lint` — 0 issues
- `task test` — all packages pass
- `task build` — clean
- Each regression test fails without its corresponding fix (verified via
  `git stash` of the fix and re-running the test)

### DIFF
--- a/pkg/runtime/client.go
+++ b/pkg/runtime/client.go
@@ -476,19 +476,24 @@ func (c *Client) GetAgentToolCount(ctx context.Context, agentFilename, agentName
 }
 
 // StreamSessionEvents streams events for a session as they occur via Server-Sent Events.
+// The returned channel is closed when ctx is cancelled, the stream's max
+// duration is reached, or the server closes the connection.
 func (c *Client) StreamSessionEvents(ctx context.Context, sessionID string) (<-chan Event, error) {
 	endpoint := fmt.Sprintf("/api/sessions/%s/events", sessionID)
 
 	u := *c.baseURL
 	u.Path = path.Join(u.Path, endpoint)
 
-	// Use long timeout for streaming
+	// Bound the maximum lifetime of a single SSE connection. The cancel
+	// must be tied to the goroutine consuming the stream, not to this
+	// function's return: cancelling streamCtx kills the in-flight HTTP
+	// request, which would turn the stream into a one-shot read.
 	timeout := c.timeoutFor("streaming")
 	streamCtx, cancel := context.WithTimeout(ctx, timeout)
-	defer cancel()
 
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Both fixes are correct and well-structured.

**1. SSE context-lifetime fix ():** The `defer cancel()` has been correctly relocated from the function scope into the goroutine that owns `resp.Body`. All early-return paths (request construction error, HTTP failure, status ≥ 400) now call `cancel()` explicitly before returning, satisfying the `context.WithTimeout` contract and avoiding any context leak. The fix precisely matches the described root cause.

**2. IPv6 listen fix (`pkg/server/listen.go`):** Changing `"tcp4"` to `"tcp"` is the correct one-line fix. Go's `net` package resolves the right IP family from the address literal, so IPv4 callers are unaffected.

**Regression tests:** Both test files are well-written — they use `t.Parallel()`, proper `httptest` servers, and cover both the happy path and the cancellation path for the SSE fix, plus IPv4/IPv6 probe-and-skip logic for the listen fix.

gtardif (approved):

PR #2777, opened 2026-05-13
set working dir properly

krissetto:
when opening a new tab in a new working dir in the TUI, i noticed my agents freaking out and doing weird stuff in the wrong directories, only to find out they we're getting the wrong working dir in their environment info

this must've gotten lost during some refactor, idk


rumpl (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is correct and well-targeted. `runtime.WithWorkingDir(runConfig.WorkingDir)` was missing from the `runtimeOpts` call, meaning all new TUI sessions inherited a stale or empty working directory. The one-line addition restores the intended behavior.

The new test (`TestRuntimeOptsPassesRunConfigWorkingDirToRuntime`) exercises the happy path end-to-end and gives good regression coverage for the fix.

PR #2778, opened 2026-05-13
feat(tui): add per-code-block copy affordance

### LINKED ISSUE #2776: Allow to copy code blocks
## Overview

When an agent displays a code block, a button should allow to easily copy its content into the clipboard 

## Motivation

Reduce the friction to copy / paste commands & code

## Use Cases

For example the agent proposes this

Image

If I copy/paste I get

```
docker run --rm --name jaeger \
-p 16686:16686 \   # Jaeger UI
-p 4318:4318 \     # OTLP/HTTP receiver
jaegertracing/all-in-one:latest
```

It's ok-ish for a CLI but for example if I have a code sample like

Image

I get

```
SELECT day,
COUNT(*)                                  AS hits,
COUNT(*) FILTER (WHERE status_code = 429) AS rate_limited
FROM traces
WHERE email = 'someone@example.com'
AND created_at > now() - interval '7 days'
GROUP BY day
ORDER BY day DESC;
```

which is less readable

## Proposed Solution

Today we have a copy feature but it's for the full answer.

Image

Could we have a local one at the code block level

## Alternatives

manually copy / paste

rumpl:
Each fenced code block now renders a copy glyph in its top-right corner. Clicking the glyph copies the block's raw content (preserving original whitespace and line breaks) instead of the whole assistant message.

The renderer returns the list of code blocks alongside the styled output, and the message click handler maps a click on the glyph back to the corresponding raw code.

Closes #2776

### DIFF
--- a/pkg/tui/components/markdown/codeblock.go
+++ b/pkg/tui/components/markdown/codeblock.go
@@ -1,0 +1,18 @@
+package markdown
+
+// CodeBlockCopyLabel is the unicode glyph rendered at the top-right corner of
+// every fenced code block. Clicking on it copies the block's raw content to
+// the clipboard. It matches the glyph used for the message-level copy
+// affordance so the visual language stays consistent; the two are
+// disambiguated by line index, not by glyph.
+const CodeBlockCopyLabel = "\u2398"
+
+// CodeBlock describes a fenced code block emitted by the renderer.
+//
+// Line is the 0-indexed line, within the renderer's output, where the copy
+// label is rendered on the code block's top padding row. Content holds the
+// raw code (without ANSI styling) so callers can place it on the clipboard.
+type CodeBlock struct {
+	Content string
+	Line    int
+}
--- a/pkg/tui/components/markdown/codeblock.go
+++ b/pkg/tui/components/markdown/codeblock.go
@@ -1,16 +1,16 @@
 package markdown
 
-// CodeBlockCopyLabel is the unicode glyph rendered at the top-right corner of
+// CodeBlockCopyIcon is the unicode glyph rendered at the top-right corner of
 // every fenced code block. Clicking on it copies the block's raw content to
 // the clipboard. It matches the glyph used for the message-level copy
 // affordance so the visual language stays consistent; the two are
 // disambiguated by line index, not by glyph.
-const CodeBlockCopyLabel = "\u2398"
+const CodeBlockCopyIcon = "\u2398"
 
 // CodeBlock describes a fenced code block emitted by the renderer.
 //
 // Line is the 0-indexed line, within the renderer's output, where the copy
-// label is rendered on the code block's top padding row. Content holds the
+// icon is rendered on the code block's top padding row. Content holds the
 // raw code (without ANSI styling) so callers can place it on the clipboard.
 type CodeBlock struct {
 	Content string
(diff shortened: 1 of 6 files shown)


pkg/tui/components/markdown/codeblock.go
```diff
@@ -0,0 +1,18 @@
+package markdown
+
+// CodeBlockCopyLabel is the unicode glyph rendered at the top-right corner of
+// every fenced code block. Clicking on it copies the block's raw content to
+// the clipboard. It matches the glyph used for the message-level copy
+// affordance so the visual language stays consistent; the two are
+// disambiguated by line index, not by glyph.
+const CodeBlockCopyLabel = "\u2398"
```
krissetto: nit: should we call this icon/glyph instead of "label"?

pkg/tui/components/markdown/fast_renderer.go
```diff
@@ -117,18 +117,19 @@ type cachedStyles struct {
...
+	ansiStrike      ansiStyle
+	ansiCode        ansiStyle
+	ansiLink        ansiStyle
+	ansiLinkText    ansiStyle
+	ansiText        ansiStyle    // base document text style
+	ansiHeadings    [6]ansiStyle // heading styles for inline restoration
+	ansiBlockquote  ansiStyle    // blockquote style for inline restoration
+	ansiFootnote    ansiStyle    // footnote reference style
+	ansiCodeBg      ansiStyle    // code block background (cached to avoid repeated buildAnsiStyle)
+	ansiCodeBgMuted ansiStyle    // muted foreground on code block background (for code-block chrome like copy label)
```
krissetto: should we add theme support for this?

pkg/tui/components/markdown/incremental.go
```diff
@@ -193,6 +217,41 @@ func (r *IncrementalRenderer) joinPrefixAndTail(prefix, tail string) string {
...
+	out = append(out, prefixBlocks...)
+	if len(tailBlocks) == 0 {
+		return out
+	}
+	offset := 0
+	if prefixOut != "" {
+		// Number of lines in prefix (FastRenderer trims its trailing newline so
+		// the count is newlines + 1) plus one for the blank separator inserted
+		// by joinPrefixAndTail.
+		offset = strings.Count(prefixOut, "\n") + 2
```
krissetto: suuper tiny nit, i think we should try not to use magic numbers and have this be a var with a good name, to simplify maintenance and be able to remove the comment

krissetto (dismissed):
some minor things, feel free to merge

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25791036773).

pkg/tui/components/markdown/codeblock.go
rumpl: Renamed to `CodeBlockCopyIcon` (and related locals) in 53d325f.

pkg/tui/components/markdown/fast_renderer.go
rumpl: It's already theme-aware: the cached style is built from `codeBg` (from `MarkdownStyle().CodeBlock`) and `styles.TextMutedGray` (theme-derived). `markdown.ResetStyles()` runs on theme change, rebuilding it. Renamed it to `ansiCodeBlockCopyIcon` in 53d325f to make the intent clearer.

pkg/tui/components/markdown/incremental.go
rumpl: Extracted as `joinSeparatorLines` constant in 53d325f, with a doc comment tying it back to `joinPrefixAndTail`'s contract.

dgageot (approved):

PR #2785, opened 2026-05-13
Use a non preview gemini model

dgageot:


examples/coder.yaml
```diff
@@ -10,7 +10,7 @@ models:
 
   gemini:
     provider: google
-    model: gemini-3.1-flash-lite-preview
+    model: gemini-3.1-flash-lite
```
docker-agent, line 13: **[MEDIUM] Model name `gemini-3.1-flash-lite` may not be a valid Google Gemini identifier**

The PR removes the `-preview` suffix from `gemini-3.1-flash-lite-preview` with the intent of switching to a GA model. However, public Google Gemini model catalogs document version series `1.0`, `1.5`, `2.0`, and `2.5` — a `3.1` series is not yet confirmed as a publicly available release.

If `gemini-3.1-flash-lite` is not a valid model name in the Google API, this example will fail at runtime when the agent tries to invoke the Google provider.

Consider double-checking the correct GA model identifier (e.g. `gemini-2.0-flash-lite` or `gemini-2.5-flash-lite`) against the [Google AI model documentation](https://ai.google.dev/gemini-api/docs/models/gemini) before merging.

examples/google_search_grounding.yaml
```diff
@@ -3,7 +3,7 @@
 models:
   gemini:
     provider: google
-    model: gemini-3.1-flash-lite-preview
+    model: gemini-3.1-flash-lite
```
docker-agent, line 6: **[MEDIUM] Model name `gemini-3.1-flash-lite` may not be a valid Google Gemini identifier**

Same issue as in `coder.yaml`: the `3.1` version series for Gemini is not confirmed in known public model catalogs as of May 2026. If this model identifier is not recognised by the Google API, the `google_search_grounding` example will fail at runtime.

Please verify the correct GA model name against the [Google AI model documentation](https://ai.google.dev/gemini-api/docs/models/gemini) (e.g. `gemini-2.0-flash-lite` may be the intended target).

docker-agent (commented):

The PR removes the `-preview` suffix from `gemini-3.1-flash-lite-preview` across two example files. The intent is correct (moving to a GA model), but the `3.1` version series is not confirmed in publicly available Google Gemini model catalogs — please verify the model identifier is valid before merging.

gtardif (approved):

PR #2786, opened 2026-05-13
Use sonnet 4.6 as default in anthropic

rumpl:
Same price, better model

### DIFF
--- a/pkg/config/auto.go
+++ b/pkg/config/auto.go
@@ -54,7 +54,7 @@ func (e *AutoModelFallbackError) Error() string {
 
 var DefaultModels = map[string]string{
 	"openai":         "gpt-5-mini",
-	"anthropic":      "claude-sonnet-4-5",
+	"anthropic":      "claude-sonnet-4-6",
 	"google":         "gemini-2.5-flash",
 	"dmr":            "ai/qwen3:latest",
 	"mistral":        "mistral-small-latest",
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The production-code change is correct and well-scoped — bumping `DefaultModels["anthropic"]` from `"claude-sonnet-4-5"` to `"claude-sonnet-4-6"`.

---

**ℹ️ Note: test file needs updating**

`pkg/config/auto_test.go` still hardcodes `"claude-sonnet-4-5"` in several places that test the default model value:

| Location | Test |
|---|---|
| line 171 | `TestAutoModelConfig` – "anthropic provider" subcase |
| line 213 | `TestAutoModelConfig` – "gateway defaults to anthropic" subcase |
| line 291 | `TestDefaultModels` – direct `assert.Equal(t, "claude-sonnet-4-5", DefaultModels["anthropic"])` |
| line 442 | `TestAutoModelConfig_UserDefaultModel` – "empty model falls back to auto detection" subcase |

These assertions depend on `DefaultModels["anthropic"]` returning the old value and will fail with `task test` as-is. (Lines 469/477 in the same file set `"claude-sonnet-4-5"` as an explicit user-supplied model and are unaffected — they test that user-provided models are preserved, not the default.)

dgageot:
@rumpl It's in the tests too

rumpl:
> @rumpl It's in the tests too

fixed

rumpl:
OMG HOW MANY TIMES ARE WE CHECKING FOR THIS

dgageot (approved):

PR #2787, opened 2026-05-13
route mouse-wheel events to background dialogs

dgageot:
When a background dialog (e.g. a pending elicitation) is open, mouse-wheel events were falling through to the chat area so the user could scroll the conversation behind the dialog. They now go to the dialog itself, like every other open dialog. Tab navigation still works as before, so the wheel naturally only ever reaches the dialog of the current tab.

### DIFF
--- a/pkg/tui/dialog/dialog.go
+++ b/pkg/tui/dialog/dialog.go
@@ -11,11 +11,12 @@ import (
 // OpenDialogMsg is sent to open a new dialog.
 //
 // OriginatingEvent is an optional runtime event whose presence marks the
-// dialog as a background dialog. Background dialogs do not block the rest of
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change correctly routes mouse-wheel events to background dialogs by removing the special-case pass-through that previously allowed wheel events to fall through to the chat area. The logic is sound: the background-dialog exclusion branch is cleanly removed from `handleWheelCoalesced`, the dialog manager's existing `forwardToTop()` path handles the wheel message correctly, nil-safety is maintained via `dialogMgr.Open()` guards, and tab navigation/click pass-through paths are unaffected. No bugs were found in the added lines.

rumpl (approved):

PR #2790, opened 2026-05-13
feat: add slash commands for agent switching

dgageot:
## Summary

Adds support for slash commands that switch the active agent. For example, declaring `/plan` in the agent config makes the planner sub-agent take over the conversation when the user types `/plan`.

## Usage

A new `agent:` field can be set on a command in `agent.yaml`:

```yaml
agents:
  root:
    sub_agents: [planner, reviewer]
    commands:
      plan:
        description: Hand off to the planner
        agent: planner
      review:
        description: Hand off to the reviewer
        agent: reviewer
```

When the user types `/plan`, the active agent is switched to `planner`. Anything typed after the command (e.g. `/plan add a logout button`) is forwarded to the target agent as the first user message.

`agent:` can be combined with `instruction:` to switch and send a fixed prompt; on its own it acts as a pure handoff.

A complete example lives in [`examples/agent_switching_commands.yaml`](../blob/feat/agent-switching-commands/examples/agent_switching_commands.yaml).

## Changes

- `pkg/config/types/commands.go` — new `Agent` field on `Command`; YAML parser accepts the `agent` key.
- `pkg/runtime/commands.go` — new `LookupCommand` helper; `ResolveCommand` forwards trailing args verbatim for agent-only commands.
- `pkg/app/app.go` — exposes `App.LookupCommand`.
- `pkg/tui/handlers.go` — `handleAgentCommand` switches the active agent before sending the resolved message.
- `pkg/cli/runner.go` — `PrepareUserMessage` does the same for the non-TUI flow.
- `agent-schema.json` — documents the new `agent` property.
- `examples/agent_switching_commands.yaml` — full example with `/plan`, `/review`, `/back`.
- `pkg/runtime/commands_test.go` — unit tests for the new behavior.

## Validation

- `task lint` clean.
- `task test` passes for all changed packages.
- `task build` succeeds.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -84,12 +84,16 @@
       "properties": {
         "profile": {
           "type": "string",
-          "description": "Shorthand that picks defaults for all the other fields. 'resilient' (default): auto-restart on failure with exponential backoff. 'strict': fail fast — the agent refuses to start when the toolset is unavailable. 'best-effort': single attempt, no auto-restart.",
-          "enum": ["resilient", "strict", "best-effort"]
+          "description": "Shorthand that picks defaults for all the other fields. 'resilient' (default): auto-restart on failure with exponential backoff. 'strict': fail fast \u2014 the agent refuses to start when the toolset is unavailable. 'best-effort': single attempt, no auto-restart.",
+          "enum": [
+            "resilient",
+            "strict",
+            "best-effort"
+          ]
         },
         "required": {
           "type": "boolean",
-          "description": "Marks the toolset as critical to the agent. NOTE: not yet enforced — a planned eager-startup phase will refuse to start the agent when a required toolset cannot reach Ready within startup_timeout. Defaults to true under the 'strict' profile, false otherwise."
+          "description": "Marks the toolset as critical to the agent. NOTE: not yet enforced \u2014 a planned eager-startup phase will refuse to start the agent when a required toolset cannot reach Ready within startup_timeout. Defaults to true under the 'strict' profile, false otherwise."
         },
         "startup_timeout": {
           "type": "string",
@@ -102,7 +106,11 @@
         "restart": {
           "type": "string",
           "description": "How the supervisor reacts to an unexpected disconnect.",
-          "enum": ["never", "on_failure", "always"]
+          "enum": [
+            "never",
+            "on_failure",
+            "always"
+          ]
         },
         "max_restarts": {
           "type": "integer",
@@ -244,7 +252,7 @@
           ]
         },
         "task_budget": {
-          "description": "Default total-token budget for an agentic task (forwarded to Anthropic as `output_config.task_budget`, with the required `task-budgets-2026-03-13` beta header attached automatically). Configurable on any Claude model — docker-agent does not gate by model name — but at the time of writing only Claude Opus 4.7 honors it. Accepts an integer token count or an object {type: tokens, total: N}.",
+          "description": "Default total-token budget for an agentic task (forwarded to Anthropic as `output_config.task_budget`, with the required `task-budgets-2026-03-13` beta header attached automatically). Configurable on any Claude model \u2014 docker-agent does not gate by model name \u2014 but at the time of writing only Claude Opus 4.7 honors it. Accepts an integer token count or an object {type: tokens, total: N}.",
           "oneOf": [
             {
               "type": "integer",
@@ -256,7 +264,9 @@
               "properties": {
                 "type": {
                   "type": "string",
-                  "enum": ["tokens"],
+                  "enum": [
+                    "tokens"
+                  ],
                   "description": "Budget kind. Only \"tokens\" is supported today."
                 },
                 "total": {
@@ -265,7 +275,9 @@
                   "description": "Total budget value."
                 }
               },
-              "required": ["total"],
+              "required": [
+                "total"
+              ],
               "additionalProperties": false
             }
           ]
@@ -283,7 +295,9 @@
       "properties": {
         "type": {
           "type": "string",
-          "enum": ["workload_identity_federation"],
+          "enum": [
+            "workload_identity_federation"
+          ],
           "description": "Authentication scheme discriminator."
         },
         "workload_identity_federation": {
@@ -291,12 +305,24 @@
           "description": "Parameters for the Anthropic OIDC federation flow. Required when type is workload_identity_federation."
         }
       },
-      "required": ["type"],
+      "required": [
+        "type"
+      ],
       "additionalProperties": false,
       "allOf": [
         {
-          "if": {"properties": {"type": {"const": "workload_identity_federation"}}},
-          "then": {"required": ["workload_identity_federation"]}
+          "if": {
+            "properties": {
+              "type": {
+                "const": "workload_identity_federation"
+              }
+            }
+          },
+          "then": {
+            "required": [
+              "workload_identity_federation"
+            ]
+          }
         }
       ]
     },
@@ -323,7 +349,11 @@
           "description": "How to obtain a fresh JWT identity token for each token exchange."
         }
       },
-      "required": ["federation_rule_id", "organization_id", "identity_token"],
+      "required": [
+        "federation_rule_id",
+        "organization_id",
+        "identity_token"
+      ],
       "additionalProperties": false
     },
     "IdentityTokenSourceConfig": {
@@ -340,7 +370,9 @@
         },
         "command": {
           "type": "array",
-          "items": {"type": "string"},
+          "items": {
+            "type": "string"
+          },
           "minItems": 1,
           "description": "Subprocess to execute; stdout is used as the JWT. Re-run on every token exchange."
         },
@@ -350,7 +382,9 @@
         },
         "headers": {
           "type": "object",
-          "additionalProperties": {"type": "string"},
+          "additionalProperties": {
+            "type": "string"
+          },
           "description": "HTTP headers sent with the URL request. Values support ${VAR} expansion. Only valid with 'url'."
         },
         "response_field": {
@@ -359,10 +393,102 @@
         }
       },
       "oneOf": [
-        {"required": ["file"], "not": {"anyOf": [{"required": ["env"]}, {"required": ["command"]}, {"required": ["url"]}]}},
-        {"required": ["env"], "not": {"anyOf": [{"required": ["file"]}, {"required": ["command"]}, {"required": ["url"]}]}},
-        {"required": ["command"], "not": {"anyOf": [{"required": ["file"]}, {"required": ["env"]}, {"required": ["url"]}]}},
-        {"required": ["url"], "not": {"anyOf": [{"required": ["file"]}, {"required": ["env"]}, {"required": ["command"]}]}}
+        {
+          "required": [
+            "file"
+          ],
+          "not": {
+            "anyOf": [
+              {
+                "required": [
+                  "env"
+                ]
+              },
+              {
+                "required": [
+                  "command"
+                ]
+              },
+              {
+                "required": [
+                  "url"
+                ]
+              }
+            ]
+          }
+        },
+        {
+          "required": [
+            "env"
+          ],
+          "not": {
+            "anyOf": [
+              {
+                "required": [
+                  "file"
+                ]
+              },
+              {
+                "required": [
+                  "command"
+                ]
+              },
+              {
+                "required": [
+                  "url"
+                ]
+              }
+            ]
+          }
+        },
+        {
+          "required": [
+            "command"
+          ],
+          "not": {
+            "anyOf": [
+              {
+                "required": [
+                  "file"
+                ]
+              },
+              {
+                "required": [
+                  "env"
+                ]
+              },
+              {
+                "required": [
+                  "url"
+                ]
+              }
+            ]
+          }
+        },
+        {
+          "required": [
+            "url"
+          ],
+          "not": {
+            "anyOf": [
+              {
+                "required": [
+                  "file"
+                ]
+              },
+              {
+                "required": [
+                  "env"
+                ]
+              },
+              {
+                "required": [
+                  "command"
+                ]
+              }
+            ]
+          }
+        }
       ],
       "additionalProperties": false
     },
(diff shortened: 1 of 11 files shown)


trungutt (dismissed):

pkg/cli/runner.go
```diff
@@ -326,6 +326,15 @@ func Run(ctx context.Context, out *Printer, cfg Config, rt runtime.Runtime, sess
 // attachment was used). Callers should pass that path to
 // session.Session.AddAttachedFile so sub-agents inherit the file context.
 func PrepareUserMessage(ctx context.Context, rt runtime.Runtime, userInput, globalAttachPath string) (*session.Message, string) {
+	// Switch the active agent first if the /command targets a sub-agent.
+	// This must happen before the message is added to the session so the
+	// next runtime turn runs on the right agent.
+	if cmd, _, ok := runtime.LookupCommand(ctx, rt, userInput); ok && cmd.Agent != "" {
+		if err := rt.SetCurrentAgent(cmd.Agent); err != nil {
```
docker-agent, line 349: **[HIGH] Agent switched before `ResolveCommand` — raw `/command` string sent to target agent**

In `PrepareUserMessage`, `rt.SetCurrentAgent(cmd.Agent)` is called on line 333 **before** `runtime.ResolveCommand` on line 339. Since `ResolveCommand` delegates to `LookupCommand`, which calls `rt.CurrentAgentInfo(ctx)` to look up the command by name, after `SetCurrentAgent` the lookup runs against the **new agent's** command table — not the original agent's.

If the target agent (e.g. `planner`) does not have the `/plan` command in its own `commands:` map (the common case — it's usually only defined on the root agent), `LookupCommand` returns `ok=false` and `ResolveCommand` falls back to returning the raw input verbatim, e.g. `/plan add a login button`, instead of the intended resolved form `add a login button`. That raw slash-command string is then sent to the target agent as the first user message.

**Trigger**: Any agent-only slash command (no `instruction:`) where the target agent doesn't re-declare that command.

**Fix**: Capture the resolved content **before** switching agents (mirroring the TUI flow in `handleAgentCommand`, which correctly calls `LookupCommand`/`ResolveCommand` first and switches after):

```go
resolvedContent := runtime.ResolveCommand(ctx, rt, userInput)

if cmd, _, ok := runtime.LookupCommand(ctx, rt, userInput); ok && cmd.Agent != "" {
    if err := rt.SetCurrentAgent(cmd.Agent); err != nil {
        slog.WarnContext(ctx, "Failed to switch agent for /command", "agent", cmd.Agent, "error", err)
    }
}
```

pkg/tui/handlers.go
```diff
@@ -624,8 +624,29 @@ func (m *appModel) handleOpenURL(url string) (tea.Model, tea.Cmd) {
...
+	var cmds []tea.Cmd
+	if ok && cmd.Agent != "" && cmd.Agent != m.sessionState.CurrentAgentName() {
+		switched, switchCmd := m.handleSwitchAgent(cmd.Agent)
+		m = switched.(*appModel)
+		if switchCmd != nil {
+			cmds = append(cmds, switchCmd)
+		}
+	}
+
+	if resolved != "" {
```
docker-agent: **[MEDIUM] Resolved message sent to old agent when `handleSwitchAgent` fails**

When `m.application.SwitchAgent(agentName)` returns an error, `handleSwitchAgent` returns early with a `notification.ErrorCmd` **without** updating `m.sessionState.CurrentAgentName`. Back in `handleAgentCommand`, the `if resolved != ""` branch on line 645 is then unconditionally reached, and `SendMsg{Content: resolved}` is appended to `cmds` and dispatched via `tea.Batch`.

The result: the user sees the "Failed to switch to agent" notification AND has their message silently dispatched to the **original** (still-active) agent — not the intended target. This can cause user confusion because the message appears to be accepted even though the intended agent hand-off failed.

**Suggested fix**: Skip the `SendMsg` when the agent switch was requested but failed, by tracking whether the switch succeeded:

```go
switchSucceeded := true
if ok && cmd.Agent != "" && cmd.Agent != m.sessionState.CurrentAgentName() {
    prevAgent := m.sessionState.CurrentAgentName()
    switched, switchCmd := m.handleSwitchAgent(cmd.Agent)
    m = switched.(*appModel)
    if switchCmd != nil {
        cmds = append(cmds, switchCmd)
    }
    switchSucceeded = m.sessionState.CurrentAgentName() != prevAgent
}

if resolved != "" && switchSucceeded {
    cmds = append(cmds, core.CmdHandler(messages.SendMsg{Content: resolved}))
}
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

pkg/cli/runner.go
```diff
@@ -326,9 +326,18 @@ func Run(ctx context.Context, out *Printer, cfg Config, rt runtime.Runtime, sess
...
+	// This ensures the command is looked up in the original agent's command table.
 	resolvedContent := runtime.ResolveCommand(ctx, rt, userInput)
 
+	// Switch the active agent if the /command targets a sub-agent.
+	// This must happen before the message is added to the session so the
+	// next runtime turn runs on the right agent.
+	if cmd, _, ok := runtime.LookupCommand(ctx, rt, userInput); ok && cmd.Agent != "" {
+		if err := rt.SetCurrentAgent(cmd.Agent); err != nil {
+			slog.WarnContext(ctx, "Failed to switch agent for /command", "agent", cmd.Agent, "error", err)
+		}
```
aheritier, line 356: **Blocking** — same bug `7df0e5a` just fixed in the TUI path is still live here.

If `SetCurrentAgent` fails (typo in `agent:`, agent name removed at runtime, or — for the remote runtime — a transport error fetching the team config), we log a warning and then carry on:

```go
resolvedContent := runtime.ResolveCommand(ctx, rt, userInput)
if cmd, _, ok := runtime.LookupCommand(ctx, rt, userInput); ok && cmd.Agent != "" {
    if err := rt.SetCurrentAgent(cmd.Agent); err != nil {
        slog.WarnContext(...) // swallowed
    }
}
```

For an *agent-only* command (no `instruction`), `ResolveCommand` returns the raw trailing args (e.g. `"add a logout button"` for `/plan add a logout button`). Those args then get wrapped into a `session.Message` and forwarded to the *original* agent, which has no `/plan` context and will treat the bare string as a normal user prompt. Worse, an empty string (from `/plan` with no args) still produces a message and gets sent.

Two options:
- Skip sending altogether on switch failure (mirror the TUI's `switchSucceeded` gate), and ideally surface the error to the user so they don't think the message went through silently.
- Or only call `SetCurrentAgent` when `cmd.Agent != currentAgent`, and propagate the error.

Either way, please add a regression test for the `agent: <missing>` path — `PrepareUserMessage` has zero coverage today.

agent-schema.json
```diff
@@ -604,6 +604,10 @@
         "instruction": {
           "type": "string",
           "description": "The prompt sent to the agent. Supports bang commands (!`command`) and positional arguments ($1, $2, etc.)"
+        },
+        "agent": {
```
aheritier, line 745: **Non-blocking** — schema vs. runtime mismatch.

The description promises *"must be reachable from the current agent's sub-agent graph"*, but `agentRouter.SetValidated` only checks `team.Agent(name)` — i.e. existence in the whole team, not reachability from the current agent's `sub_agents` list. So a config with `commands.go: { agent: foo }` where `foo` is a sibling tree-far-away will validate fine and switch successfully.

Pick one:
- Add a real reachability check in `SetCurrentAgent` (walk `sub_agents` transitively from the current agent), and validate at config load that every `command.agent` is reachable from its declaring agent.
- Or relax the schema text to *"must exist in the team"*.

Reachability would be the more useful guarantee — otherwise `/plan` from agent A can teleport you to an unrelated agent B.

pkg/tui/handlers.go
```diff
@@ -624,8 +624,32 @@ func (m *appModel) handleOpenURL(url string) (tea.Model, tea.Cmd) {
...
+		prevAgent := m.sessionState.CurrentAgentName()
+		switched, switchCmd := m.handleSwitchAgent(cmd.Agent)
+		m = switched.(*appModel)
+		if switchCmd != nil {
+			cmds = append(cmds, switchCmd)
+		}
+		switchSucceeded = m.sessionState.CurrentAgentName() != prevAgent
+	}
+
+	if resolved != "" && switchSucceeded {
```
aheritier, line 659: **Non-blocking** — `switchSucceeded` is inferred by comparing the agent name before/after, which couples this code to an internal side effect of `handleSwitchAgent`. If `handleSwitchAgent` ever pushes an async switch (e.g. via a `tea.Cmd`) the comparison will silently report success while the switch hasn't actually happened yet.

Cleaner contract: have `handleSwitchAgent` return `(tea.Model, tea.Cmd, bool)` (or split out a sync `trySwitchAgent(name) error` helper that this caller and the existing wrapper both use). Then this branch becomes:

```go
if err := m.trySwitchAgent(cmd.Agent); err != nil {
    return m, notification.ErrorCmd(...)
}
```

Same effect, no leaky abstraction.

pkg/cli/runner.go
```diff
@@ -326,9 +326,18 @@ func Run(ctx context.Context, out *Printer, cfg Config, rt runtime.Runtime, sess
...
+	}
```
aheritier, line 357: **Non-blocking** — both `PrepareUserMessage`'s new branch and `handleAgentCommand`'s switch-then-send wiring landed without unit tests. Only `LookupCommand` and the `ResolveCommand` agent-only branch are covered. A test that:

1. Configures a runtime with `/plan { agent: planner }`.
2. Calls `PrepareUserMessage(rt, "/plan design X", "")`.
3. Asserts (a) the runtime's current agent is now `planner`, and (b) the message content is `"design X"`.

…would have caught the failure mode in the Blocking comment above and pin behaviour for future refactors.

pkg/config/types/commands.go
```diff
@@ -113,23 +134,24 @@ func (c *Commands) UnmarshalYAML(unmarshal func(any) error) error {
...
 
-		if inst == "" && desc == "" {
-			return Command{}, errors.New("command must have at least 'instruction' or 'description'")
+		if inst == "" && desc == "" && agent == "" {
+			return Command{}, errors.New("command must have at least 'instruction', 'description' or 'agent'")
 		}
-		if inst == "" {
+		if inst == "" && agent == "" {
 			inst = desc
 		}
```
aheritier, line 156: **Nit** — the `inst == "" && agent == ""` → `inst = desc` fallback is preserved (good — keeps the legacy single-string-as-description shape working), but with three legal input shapes (instruction-only, agent-only, description-only) the implicit promotion now reads as magic. A one-line comment would help future readers, e.g.:

```go
// Backwards compatibility: when only `description` is set, promote it to
// `instruction` so the legacy "description doubles as prompt" shape keeps
// working. Skipped when `agent` is set, because agent-only commands are
// meant to have no instruction.
```

aheritier (changes requested):

pkg/tui/handlers.go
```diff
@@ -624,8 +624,32 @@ func (m *appModel) handleOpenURL(url string) (tea.Model, tea.Cmd) {
...
+	// the previous agent.
+	cmd, _, ok := m.application.LookupCommand(ctx, command)
+	resolved := m.application.ResolveCommand(ctx, command)
+
+	var cmds []tea.Cmd
+	switchSucceeded := true
+	if ok && cmd.Agent != "" && cmd.Agent != m.sessionState.CurrentAgentName() {
+		prevAgent := m.sessionState.CurrentAgentName()
+		switched, switchCmd := m.handleSwitchAgent(cmd.Agent)
+		m = switched.(*appModel)
```
docker-agent: **[LIKELY / medium] Unchecked type assertion `switched.(*appModel)` could panic on future refactors**

`handleAgentCommand` performs an unchecked type assertion after calling `handleSwitchAgent`:

```go
switched, switchCmd := m.handleSwitchAgent(cmd.Agent)
m = switched.(*appModel)  // panics if switched is not *appModel
```

Today this is safe because `handleSwitchAgent` is a method on `*appModel` and both return paths return the receiver `m`. However, if `handleSwitchAgent` is ever refactored to wrap or replace the model with a different concrete type, this will panic at runtime with no compiler warning.

**Suggestion:** Use the comma-ok form to handle the case gracefully:
```go
newModel, ok := switched.(*appModel)
if !ok {
    // fallback: skip send, surface an error
    return switched, tea.Batch(cmds...)
}
m = newModel
```

pkg/cli/runner.go
```diff
@@ -326,9 +326,18 @@ func Run(ctx context.Context, out *Printer, cfg Config, rt runtime.Runtime, sess
...
 func PrepareUserMessage(ctx context.Context, rt runtime.Runtime, userInput, globalAttachPath string) (*session.Message, string) {
-	// Resolve any /command to its prompt text
+	// Resolve any /command to its prompt text BEFORE switching agents.
+	// This ensures the command is looked up in the original agent's command table.
 	resolvedContent := runtime.ResolveCommand(ctx, rt, userInput)
 
+	// Switch the active agent if the /command targets a sub-agent.
+	// This must happen before the message is added to the session so the
+	// next runtime turn runs on the right agent.
+	if cmd, _, ok := runtime.LookupCommand(ctx, rt, userInput); ok && cmd.Agent != "" {
```
docker-agent, line 348: **[CONFIRMED / medium] Empty user message sent to LLM when agent-only `/command` has no trailing arguments (CLI mode)**

When a user types `/plan` (no trailing text) in CLI mode, `PrepareUserMessage` does the following:

1. `ResolveCommand` looks up the command, finds `instruction == ""` (agent-only), returns `rest = ""` (no trailing args after the slash command).
2. `resolvedContent = ""`  →  `messageText = ""`
3. `CreateUserMessageWithAttachment(ctx, "", "")` creates a valid `*Message` with empty content.
4. The caller unconditionally adds it: `sess.AddMessage(userMsg)` — the LLM receives an empty user turn.

The TUI path (`handleAgentCommand`) correctly guards against this:
```go
if resolved != "" && switchSucceeded {
    cmds = append(cmds, core.CmdHandler(messages.SendMsg{Content: resolved}))
}
```

The CLI path has no equivalent guard, so the LLM always receives a message — even an empty one — which can trigger unexpected or confusing responses.

**Suggestion:** Mirror the TUI guard in `PrepareUserMessage`:
```go
resolvedContent := runtime.ResolveCommand(ctx, rt, userInput)

if cmd, _, ok := runtime.LookupCommand(ctx, rt, userInput); ok && cmd.Agent != "" {
    if err := rt.SetCurrentAgent(cmd.Agent); err != nil {
        slog.WarnContext(ctx, "Failed to switch agent for /command", "agent", cmd.Agent, "error", err)
    }
    // Agent-only command with no trailing args: switch but send no message.
    if resolvedContent == "" {
        return nil, ""
    }
}
```
(Callers would need a nil-message guard, or return a sentinel.)

docker-agent (commented):

Two issues found in the agent-switching code paths — one confirmed behavioral difference between CLI and TUI modes, one fragility in the type assertion pattern.

pkg/cli/runner.go
```diff
@@ -326,9 +326,21 @@ func Run(ctx context.Context, out *Printer, cfg Config, rt runtime.Runtime, sess
...
+		// If the agent switch fails, we must not proceed with sending the message
+		// to the wrong agent. Return an empty message to signal the error.
+		if err := rt.SetCurrentAgent(cmd.Agent); err != nil {
+			slog.WarnContext(ctx, "Failed to switch agent for /command", "agent", cmd.Agent, "error", err)
+			return session.UserMessage(""), ""
```
docker-agent: **[HIGH] Empty message silently added to session and forwarded to LLM on agent-switch failure**

`PrepareUserMessage` returns `session.UserMessage("")` here to "signal" the error, but the caller in `Run` (line 107) unconditionally does:
```go
userMsg, attachedPath := PrepareUserMessage(ctx, rt, userInput, cfg.AttachmentPath)
sess.AddMessage(userMsg)   // no guard — empty message is added
rt.RunStream(ctx, sess)    // LLM receives empty user message in history
```

There is no nil/empty check at the call site. The LLM will receive a blank `user` turn in the conversation history, which can confuse the model and silently corrupt the session. The TUI path (`handleAgentCommand`) correctly suppresses the message on failure — the CLI path needs the same treatment.

**Suggested fix** — return `error` explicitly so callers are forced to handle it:
```go
func PrepareUserMessage(...) (*session.Message, string, error) {
    ...
    if err := rt.SetCurrentAgent(cmd.Agent); err != nil {
        slog.WarnContext(ctx, "Failed to switch agent for /command", ...)
        return nil, "", fmt.Errorf("switch agent %q: %w", cmd.Agent, err)
    }
    ...
    return msg, attachPath, nil
}
```
Then the caller can `continue` / surface the error instead of forwarding an empty message.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

Agent-switching slash commands are a well-structured addition. The runtime layer, schema, tests, and TUI path are all clean. One real bug found in the non-TUI (CLI) path: when agent switching fails, the error is signaled via an empty `session.UserMessage` that the caller unconditionally adds to the session and forwards to the LLM, silently corrupting conversation history.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26035133458).

aheritier (approved):

Issue #2781, opened 2026-05-13
API: allow switching models mid-session


dgageot:
In the TUI, you can swap the model for the current agent at any time through the model picker. Over `docker agent serve api` that's not possible — there's no endpoint for it.

The plumbing already exists on the runtime side (`runtime.SetAgentModel`, `SupportsModelSwitching`, `AvailableModels`), and `App.SetCurrentAgentModel` shows how to apply + persist the override. It would just need to be exposed, e.g.:

- `GET /api/sessions/:id/models` to list available models
- `PATCH /api/sessions/:id/model` to set/clear the override for the current agent

Pairs naturally with #2780.

PR #2791, opened 2026-05-13
feat(api): accept model overrides on session creation and add runtime model switching endpoints

### LINKED ISSUE #2780: API: allow choosing a model when creating a session
When using `docker agent serve api`, there's no way to tell the server which model to use for a new session.

`POST /api/sessions` accepts a session template, but the model-related fields (`agent_model_overrides`, `custom_models_used`) are silently dropped — only `MaxIterations`, `MaxConsecutiveToolCalls`, `MaxOldToolCallTokens`, `ToolsApproved`, `WorkingDir` and `Permissions` are copied onto the new session.

The TUI already supports this via `App.SetCurrentAgentModel` and persists overrides in the session, so the runtime knows how to honor them. It would be nice if the API did the same: pass `agent_model_overrides` in the create-session body and have `SessionManager` apply them when building the runtime.

dgageot:
Fixes #2780 and #2781.

### `POST /api/sessions` honours model selection (#2780)

The handler binds the request body to a `session.Session` so the existing
`agent_model_overrides` and `custom_models_used` JSON fields are now picked
up and copied onto the new session in `SessionManager.CreateSession`. The
runtime applies them the first time it is built for the session (see
`runtimeForSession`), so a freshly created session can be pinned to a
specific model without an extra round-trip.

### Runtime model switching over HTTP (#2781)

Two new endpoints:

- `GET  /api/sessions/:id/models` → list of `runtime.ModelChoice` for the
  session's current agent, with `IsCurrent` set on the active selection
  (override or default), and any inline `provider/model` ref from the
  session's history (or override) appended as a synthetic choice.
- `PATCH /api/sessions/:id/model` → applies a per-agent override and
  persists it. Empty `model` clears the override and reverts to the
  agent's configured default.

`POST /api/sessions/:id/model` is also accepted because the existing
`pkg/runtime` `Client.SetAgentModel` (used by `RemoteRuntime`) was
historically a POST — so no client upgrade is required.

Status code mapping:

| outcome | code |
| --- | --- |
| ok | 200 |
| invalid JSON body | 400 |
| no runtime attached for session | 404 |
| runtime doesn't support model switching | 422 |
| invalid model ref | 400 |

## Implementation notes

- Reuses `runtime.ModelChoice` directly as the wire type (added JSON tags).
  No duplicate `pkg/api.ModelChoice`.
- The TUI `App.AvailableModels` now delegates to a shared
  `runtime.DecorateModelChoices` helper, which is also used by
  `SessionManager.AvailableSessionModels`. HTTP and TUI clients see the
  same picker state.
- `SessionManager.SetSessionAgentModel` snapshots state before mutation
  and rolls back **both** the in-memory session and the runtime override
  if `sessionStore.UpdateSession` fails.
- New sentinel `ErrSessionNotRunning` distinguishes "session not found
  or not running" from generic errors so the HTTP layer can map it to
  404 deterministically.
- `runtimeForSession` re-applies persisted overrides via a small
  `applyStoredOverrides` helper. Failures are logged at WARN — a stored
  override that no longer resolves must not prevent session resumption.

## Tests

- `pkg/server/session_models_test.go` (new) — end-to-end coverage for
  GET/PATCH/POST, IsCurrent decoration, custom models appended, empty
  body clears override, store-write rollback, runtime-failure rollback,
  404 when no runtime attached, 422 when runtime doesn't support
  switching, POST verb compat.
- `pkg/runtime/model_switcher_test.go` — table test for
  `DecorateModelChoices` corner cases.
- `pkg/server/session_manager_test.go` — `SupportsModelSwitching` on
  `fakeRuntime`.

## Validation

- `task lint` — clean.
- `go test -race -count=3 ./pkg/server/... ./pkg/runtime/... ./pkg/api/...` — green.
- Full repo `go test` — green except pre-existing environmental failures
  in `pkg/teamloader` (require Docker Model Runner running locally,
  unrelated to this PR).

## Commits

- feat: support model overrides when creating sessions
- feat: add GET/PATCH/POST endpoints for runtime model switching
- fix: add backward compat note for POST /sessions/:id/model
- fix: thread-safe model reads and persistence rollback
- test: enhance model selection test coverage and fix goroutine leaks
- refactor(runtime): extract DecorateModelChoices and remove duplication
- fix: distinguish "session not running" from other errors

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -270,3 +270,15 @@ type SessionStatusResponse struct {
 	OutputTokens int64  `json:"output_tokens"`
 	NumMessages  int    `json:"num_messages"`
 }
+
+// SetSessionModelRequest is the body of PATCH /api/sessions/:id/model.
+// An empty Model clears the override and reverts to the agent's default.
+type SetSessionModelRequest struct {
+	Model string `json:"model"`
+}
+
+// SetSessionModelResponse is the response from PATCH /api/sessions/:id/model.
+type SetSessionModelResponse struct {
+	Agent string `json:"agent"`
+	Model string `json:"model,omitempty"`
+}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -777,76 +777,15 @@ func (a *App) AvailableModels(ctx context.Context) []runtime.ModelChoice {
 	if !a.runtime.SupportsModelSwitching() {
 		return nil
 	}
-	models := a.runtime.AvailableModels(ctx)
 
-	// Determine the currently active model for this agent
 	agentName := a.runtime.CurrentAgentName()
-	currentModelRef := ""
-	if a.session != nil && a.session.AgentModelOverrides != nil {
-		currentModelRef = a.session.AgentModelOverrides[agentName]
-	}
-
-	// Build a set of model refs already in the list
-	existingRefs := make(map[string]bool)
-	for _, m := range models {
-		existingRefs[m.Ref] = true
-	}
-
-	// Check if current model is in the list and mark it
-	currentFound := currentModelRef == ""
-	for i := range models {
-		if currentModelRef != "" {
-			// An override is set - mark the override as current
-			if models[i].Ref == currentModelRef {
-				models[i].IsCurrent = true
-				currentFound = true
-			}
-		} else {
-			// No override - the default model is current
-			models[i].IsCurrent = models[i].IsDefault
-		}
-	}
-
-	// Add custom models from the session that aren't already in the list
+	currentRef := ""
+	var customRefs []string
 	if a.session != nil {
-		for _, customRef := range a.session.CustomModelsUsed {
-			if existingRefs[customRef] {
-				continue // Already in the list
-			}
-			existingRefs[customRef] = true
-
-			providerName, modelName, _ := strings.Cut(customRef, "/")
-			isCurrent := customRef == currentModelRef
-			if isCurrent {
-				currentFound = true
-			}
-			models = append(models, runtime.ModelChoice{
-				Name:      customRef,
-				Ref:       customRef,
-				Provider:  providerName,
-				Model:     modelName,
-				IsDefault: false,
-				IsCurrent: isCurrent,
-				IsCustom:  true,
-			})
-		}
+		currentRef = a.session.AgentModelOverrides[agentName]
+		customRefs = a.session.CustomModelsUsed
 	}
-
-	// If current model is a custom model not in the list, add it
-	if !currentFound && strings.Contains(currentModelRef, "/") {
-		providerName, modelName, _ := strings.Cut(currentModelRef, "/")
-		models = append(models, runtime.ModelChoice{
-			Name:      currentModelRef,
-			Ref:       currentModelRef,
-			Provider:  providerName,
-			Model:     modelName,
-			IsDefault: false,
-			IsCurrent: true,
-			IsCustom:  true,
-		})
-	}
-
-	return models
+	return runtime.DecorateModelChoices(a.runtime.AvailableModels(ctx), currentRef, customRefs)
 }
 
 // trackCustomModel adds a custom model to the session's history if not already present.
--- a/pkg/runtime/model_switcher.go
+++ b/pkg/runtime/model_switcher.go
@@ -82,21 +82,30 @@ type SessionModelsResponse struct {
 //
 // currentRef is the model override active for the agent ("" when none),
 // and customRefs is the session's CustomModelsUsed history.
+//
+// The input slice is never mutated: callers can safely pass a slice that
+// is shared with or backed by an internal cache.
 func DecorateModelChoices(models []ModelChoice, currentRef string, customRefs []string) []ModelChoice {
-	existingRefs := make(map[string]bool, len(models))
-	for _, m := range models {
+	// Defensive copy: AvailableModels implementations may return a slice
+	// backed by an internal cache. Mutating its IsCurrent flag in place
+	// would leak picker state across sessions/agents.
+	result := make([]ModelChoice, len(models), len(models)+len(customRefs)+1)
+	copy(result, models)
+
+	existingRefs := make(map[string]bool, len(result))
+	for _, m := range result {
 		existingRefs[m.Ref] = true
 	}
 
 	currentFound := currentRef == ""
-	for i := range models {
+	for i := range result {
 		if currentRef != "" {
-			if models[i].Ref == currentRef {
-				models[i].IsCurrent = true
+			if result[i].Ref == currentRef {
+				result[i].IsCurrent = true
 				currentFound = true
 			}
 		} else {
-			models[i].IsCurrent = models[i].IsDefault
+			result[i].IsCurrent = result[i].IsDefault
 		}
 	}
 
@@ -111,7 +120,7 @@ func DecorateModelChoices(models []ModelChoice, currentRef string, customRefs []
 		if isCurrent {
 			currentFound = true
 		}
-		models = append(models, ModelChoice{
+		result = append(result, ModelChoice{
 			Name:      ref,
 			Ref:       ref,
 			Provider:  prov,
@@ -126,7 +135,7 @@ func DecorateModelChoices(models []ModelChoice, currentRef string, customRefs []
 	// choice so the picker can still highlight the active selection.
 	if !currentFound && strings.Contains(currentRef, "/") {
 		prov, name, _ := strings.Cut(currentRef, "/")
-		models = append(models, ModelChoice{
+		result = append(result, ModelChoice{
 			Name:      currentRef,
 			Ref:       currentRef,
 			Provider:  prov,
@@ -136,7 +145,7 @@ func DecorateModelChoices(models []ModelChoice, currentRef string, customRefs []
 		})
 	}
 
-	return models
+	return result
 }
 
 // ModelSwitcherConfig holds the configuration needed for model switching.
--- a/pkg/runtime/model_switcher.go
+++ b/pkg/runtime/model_switcher.go
@@ -73,6 +73,72 @@ type SessionModelsResponse struct {
 	Models          []ModelChoice `json:"models"`
 }
 
+// DecorateModelChoices marks the active selection with IsCurrent and
+// appends any custom (provider/model) refs from the session history that
+// the runtime does not already expose. It is used by every consumer that
+// wants to render a model picker (the TUI App, the HTTP /sessions/:id/models
+// endpoint, …) so they all agree on which entry is current and what the
+// final list looks like.
+//
+// currentRef is the model override active for the agent ("" when none),
+// and customRefs is the session's CustomModelsUsed history.
+func DecorateModelChoices(models []ModelChoice, currentRef string, customRefs []string) []ModelChoice {
+	existingRefs := make(map[string]bool, len(models))
+	for _, m := range models {
+		existingRefs[m.Ref] = true
+	}
+
+	currentFound := currentRef == ""
+	for i := range models {
+		if currentRef != "" {
+			if models[i].Ref == currentRef {
+				models[i].IsCurrent = true
+				currentFound = true
+			}
+		} else {
+			models[i].IsCurrent = models[i].IsDefault
+		}
+	}
+
+	for _, ref := range customRefs {
+		if existingRefs[ref] {
+			continue
+		}
+		existingRefs[ref] = true
+
+		prov, name, _ := strings.Cut(ref, "/")
+		isCurrent := ref == currentRef
+		if isCurrent {
+			currentFound = true
+		}
+		models = append(models, ModelChoice{
+			Name:      ref,
+			Ref:       ref,
+			Provider:  prov,
+			Model:     name,
+			IsCurrent: isCurrent,
+			IsCustom:  true,
+		})
+	}
+
+	// If the override points at an inline provider/model not in the
+	// runtime's list nor in the session's history, fabricate a synthetic
+	// choice so the picker can still highlight the active selection.
+	if !currentFound && strings.Contains(currentRef, "/") {
+		prov, name, _ := strings.Cut(currentRef, "/")
+		models = append(models, ModelChoice{
+			Name:      currentRef,
+			Ref:       currentRef,
+			Provider:  prov,
+			Model:     name,
+			IsCurrent: true,
+			IsCustom:  true,
+		})
+	}
+
+	return models
+}
+
 // ModelSwitcherConfig holds the configuration needed for model switching.
 // This is populated by the app layer when creating the runtime.
 type ModelSwitcherConfig struct {
--- a/pkg/runtime/model_switcher.go
+++ b/pkg/runtime/model_switcher.go
@@ -16,49 +16,61 @@ import (
 	"github.com/docker/docker-agent/pkg/modelsdev"
 )
 
-// ModelChoice represents a model available for selection in the TUI picker.
+// ModelChoice represents a model available for selection in the model picker.
+//
+// JSON tags are part of the public wire format used by
+// GET /api/sessions/:id/models; renaming a tag is a breaking change.
 type ModelChoice struct {
 	// Name is the display name (config key)
-	Name string
+	Name string `json:"name"`
 	// Ref is the model reference used internally (e.g., "my_model" or "openai/gpt-4o")
-	Ref string
+	Ref string `json:"ref"`
 	// Provider is the provider name (e.g., "openai", "anthropic")
-	Provider string
+	Provider string `json:"provider,omitempty"`
 	// Model is the specific model name (e.g., "gpt-4o", "claude-sonnet-4-0")
-	Model string
+	Model string `json:"model,omitempty"`
 	// IsDefault indicates this is the agent's configured default model
-	IsDefault bool
+	IsDefault bool `json:"is_default,omitempty"`
 	// IsCurrent indicates this is the currently active model for the agent
-	IsCurrent bool
+	IsCurrent bool `json:"is_current,omitempty"`
 	// IsCustom indicates this is a custom model from the session history (not from config)
-	IsCustom bool
+	IsCustom bool `json:"is_custom,omitempty"`
 	// IsCatalog indicates this is a model from the models.dev catalog
-	IsCatalog bool
+	IsCatalog bool `json:"is_catalog,omitempty"`
 
 	// The fields below are populated (best-effort) from the models.dev
 	// catalog. They are optional and may all be zero/empty when no
 	// catalog entry is found for the model.
 
 	// Family is the model family (e.g., "claude", "gpt").
-	Family string
+	Family string `json:"family,omitempty"`
 	// InputCost is the price (in USD) per 1M input tokens.
-	InputCost float64
+	InputCost float64 `json:"input_cost,omitempty"`
 	// OutputCost is the price (in USD) per 1M output tokens.
-	OutputCost float64
+	OutputCost float64 `json:"output_cost,omitempty"`
 	// CacheReadCost is the price (in USD) per 1M cached input tokens.
-	CacheReadCost float64
+	CacheReadCost float64 `json:"cache_read_cost,omitempty"`
 	// CacheWriteCost is the price (in USD) per 1M cache-write tokens.
-	CacheWriteCost float64
+	CacheWriteCost float64 `json:"cache_write_cost,omitempty"`
 	// ContextLimit is the maximum context window size in tokens.
-	ContextLimit int
+	ContextLimit int `json:"context_limit,omitempty"`
 	// OutputLimit is the maximum number of tokens the model can produce
 	// in a single response.
-	OutputLimit int64
+	OutputLimit int64 `json:"output_limit,omitempty"`
 	// InputModalities lists the input modalities supported by the model
 	// (e.g., "text", "image", "audio").
-	InputModalities []string
+	InputModalities []string `json:"input_modalities,omitempty"`
 	// OutputModalities lists the output modalities the model can produce.
-	OutputModalities []string
+	OutputModalities []string `json:"output_modalities,omitempty"`
+}
+
+// SessionModelsResponse is the response returned by
+// GET /api/sessions/:id/models. CurrentModelRef is the active override for
+// the named agent (empty when the agent is using its configured default).
+type SessionModelsResponse struct {
+	Agent           string        `json:"agent"`
+	CurrentModelRef string        `json:"current_model_ref,omitempty"`
+	Models          []ModelChoice `json:"models"`
 }
 
 // ModelSwitcherConfig holds the configuration needed for model switching.
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -574,10 +574,14 @@ func (s *Server) getSessionModels(c echo.Context) error {
 
 	agentName, current, choices, err := s.sm.AvailableSessionModels(c.Request().Context(), sessionID)
 	if err != nil {
-		if errors.Is(err, ErrModelSwitchingNotSupported) {
+		switch {
+		case errors.Is(err, ErrModelSwitchingNotSupported):
 			return echo.NewHTTPError(http.StatusUnprocessableEntity, err.Error())
+		case errors.Is(err, ErrSessionNotRunning):
+			return echo.NewHTTPError(http.StatusNotFound, err.Error())
+		default:
+			return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 		}
-		return echo.NewHTTPError(http.StatusNotFound, err.Error())
 	}
 
 	return c.JSON(http.StatusOK, runtime.SessionModelsResponse{
@@ -600,10 +604,14 @@ func (s *Server) setSessionModel(c echo.Context) error {
 
 	agentName, modelRef, err := s.sm.SetSessionAgentModel(c.Request().Context(), sessionID, req.Model)
 	if err != nil {
-		if errors.Is(err, ErrModelSwitchingNotSupported) {
+		switch {
+		case errors.Is(err, ErrModelSwitchingNotSupported):
 			return echo.NewHTTPError(http.StatusUnprocessableEntity, err.Error())
+		case errors.Is(err, ErrSessionNotRunning):
+			return echo.NewHTTPError(http.StatusNotFound, err.Error())
+		default:
+			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
 		}
-		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
 	}
 
 	return c.JSON(http.StatusOK, api.SetSessionModelResponse{
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -72,6 +72,10 @@ func (s *Server) registerRoutes() {
 	group.PATCH("/sessions/:id/tokens", s.updateSessionTokens)
 	group.PATCH("/sessions/:id/starred", s.setSessionStarred)
 	group.GET("/sessions/:id/models", s.getSessionModels)
+	// PATCH is the canonical verb for updating the agent's model. POST is
+	// also accepted because pkg/runtime Client.SetAgentModel (used by
+	// RemoteRuntime) was historically a POST; keep both so a client
+	// upgrade is not required.
 	group.PATCH("/sessions/:id/model", s.setSessionModel)
 	group.POST("/sessions/:id/model", s.setSessionModel)
 	group.POST("/sessions", s.createSession)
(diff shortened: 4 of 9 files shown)


trungutt (dismissed):

pkg/server/session_manager.go
```diff
@@ -729,6 +786,147 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+
+	agentName := rs.runtime.CurrentAgentName()
+	current := ""
+	var customRefs []string
+	if rs.session != nil {
+		current = rs.session.AgentModelOverrides[agentName]
+		customRefs = rs.session.CustomModelsUsed
+	}
+
+	choices := runtime.DecorateModelChoices(rs.runtime.AvailableModels(ctx), current, customRefs)
```
docker-agent, line 850: **[HIGH] Lock held across blocking runtime I/O in `AvailableSessionModels`**

`sm.mux.Lock()` / `defer sm.mux.Unlock()` wraps the entire function body, including the call to `rs.runtime.AvailableModels(ctx)` — a context-accepting call that can perform network I/O (e.g. models.dev catalog lookups, provider discovery). Any concurrent caller of `RunSession`, `DeleteSession`, `SetSessionAgentModel`, `UpdateSessionTitle`, or any other method that acquires `sm.mux` will block for the full duration of that network call.

The same pattern occurs in `SetSessionAgentModel`, which calls `rs.runtime.SetAgentModel(ctx, agentName, modelRef)` under `sm.mux`.

**Fix**: snapshot the `rs` pointer (and any session fields needed) under the lock, release the lock, then invoke the runtime method without holding `sm.mux`. The runtime is already concurrency-safe via its own internal locking.

pkg/server/server.go
```diff
@@ -557,6 +565,61 @@ func (s *Server) setSessionStarred(c echo.Context) error {
...
+
+	agentName, modelRef, err := s.sm.SetSessionAgentModel(c.Request().Context(), sessionID, req.Model)
+	if err != nil {
+		switch {
+		case errors.Is(err, ErrModelSwitchingNotSupported):
+			return echo.NewHTTPError(http.StatusUnprocessableEntity, err.Error())
+		case errors.Is(err, ErrSessionNotRunning):
+			return echo.NewHTTPError(http.StatusNotFound, err.Error())
+		default:
+			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
```
docker-agent: **[MEDIUM] Unknown errors from `SetSessionAgentModel` mapped to HTTP 400 instead of 500**

The `default` branch here returns `http.StatusBadRequest` (400) for any error that is not `ErrModelSwitchingNotSupported` or `ErrSessionNotRunning`. HTTP 400 means the client sent an invalid request — but unexpected errors from the session store or from the runtime (e.g. a network failure, an internal validation error wrapped in `fmt.Errorf`) are server-side faults, not client mistakes.

By contrast, the companion `getSessionModels` handler (a few lines above) correctly uses `http.StatusInternalServerError` (500) in its `default` branch.

**Fix**: Change the `default` case to return `http.StatusInternalServerError`.

```go
default:
    return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
```

pkg/runtime/model_switcher.go
```diff
@@ -16,49 +16,127 @@ import (
...
+	existingRefs := make(map[string]bool, len(models))
+	for _, m := range models {
+		existingRefs[m.Ref] = true
+	}
+
+	currentFound := currentRef == ""
+	for i := range models {
+		if currentRef != "" {
+			if models[i].Ref == currentRef {
+				models[i].IsCurrent = true
```
docker-agent: **[MEDIUM] `DecorateModelChoices` mutates the caller-supplied `models` slice in-place**

`models[i].IsCurrent = true` and `models[i].IsCurrent = models[i].IsDefault` modify elements of the slice that was passed in by the caller. In Go, a slice header is copied at the call boundary but the underlying array is shared. If `AvailableModels()` returns a slice backed by an internal cache (rather than a freshly allocated copy), these index-assignment mutations will silently update the cached structs.

A subsequent call to `AvailableModels()` could then return entries already marked `IsCurrent = true` (or `false`) from a previous decoration call, producing incorrect picker state for a different session or agent.

**Fix**: Make a defensive copy of the input slice at the top of `DecorateModelChoices`:

```go
result := make([]ModelChoice, len(models))
copy(result, models)
// ... operate on result, not models
```

and return `result` instead of the mutated input.

docker-agent (commented):

This PR introduces runtime model switching (GET/PATCH/POST endpoints) and session-creation model overrides. The implementation is well-structured with a rollback path and a new error sentinel. Three findings were identified in the changed code.

dgageot:
Thanks for the review! Addressed all three findings in 3 separate commits:

### [MEDIUM] `DecorateModelChoices` mutates caller-supplied slice — fixed in 96ea8bbe7
The function now allocates a `result` slice (sized `len(models)+len(customRefs)+1` to fit the synthetic-override case), copies the input, and operates on the copy. Doc-comment updated to spell out the non-mutation guarantee. Added `TestDecorateModelChoices/does_not_mutate_the_input_slice` that snapshots the input before calling and asserts deep-equality after.

### [MEDIUM] Unknown errors mapped to 400 — fixed in ad1366a9f
The handler's `default` branch now returns `500` (mirroring the companion `getSessionModels` handler). 400 is reserved for the upstream `Bind` failure on the request body. Added `TestAttachedServer_SetSessionModel_StoreFailureReturns500` driving a synthetic `UpdateSession` failure through the HTTP layer and asserting 500.

### [HIGH] Lock held across blocking runtime I/O — fixed in be9e8aeda
- Added a per-session `modelSwitch sync.Mutex` on `activeRuntimes` to serialise concurrent `SetSessionAgentModel` calls **on the same session** without blocking other sessions.
- `AvailableSessionModels` now snapshots `current` and `customRefs` under `sm.mux` briefly, releases the lock, then calls `runtime.AvailableModels(ctx)`.
- `SetSessionAgentModel` is restructured so `sm.mux` is taken in three short critical sections (snapshot prev → mutate session → rollback in-memory) but released around the slow runtime calls (`SetAgentModel` apply + rollback) and the store write. Atomicity wrt other model-switches on the same session is preserved by the new `modelSwitch` lock; the rollback contract is unchanged.
- Added two synchronisation-style tests that park a deliberately-slow runtime call open and verify that an unrelated `SetSessionStarred` (sm.mux-acquiring) completes within 2s while the runtime call is still in flight:
  - `TestSessionManager_AvailableSessionModels_DoesNotHoldMuxDuringRuntimeIO`
  - `TestSessionManager_SetSessionAgentModel_DoesNotHoldMuxDuringRuntimeIO`

### Validation
- `task lint` — clean.
- `go test -race -count=5 ./pkg/server/...` — green.
- `go test -race -count=3 ./pkg/server/... ./pkg/runtime/...` — green.

pkg/server/server.go
```diff
@@ -70,6 +71,13 @@ func (s *Server) registerRoutes() {
 	group.PATCH("/sessions/:id/title", s.updateSessionTitle)
 	group.PATCH("/sessions/:id/tokens", s.updateSessionTokens)
 	group.PATCH("/sessions/:id/starred", s.setSessionStarred)
+	group.GET("/sessions/:id/models", s.getSessionModels)
+	// PATCH is the canonical verb for updating the agent's model. POST is
+	// also accepted because pkg/runtime Client.SetAgentModel (used by
+	// RemoteRuntime) was historically a POST; keep both so a client
+	// upgrade is not required.
+	group.PATCH("/sessions/:id/model", s.setSessionModel)
```
aheritier, line 79: **Blocking — docs**: The three new endpoints (`GET /api/sessions/:id/models`, `PATCH /api/sessions/:id/model`, `POST /api/sessions/:id/model`) aren't reflected in `docs/features/api-server/index.md`'s endpoint table (around line 44). Worth adding rows so the API server doc stays in sync with the routes registered here. Also worth a short note that POST is accepted for backward compat with `RemoteRuntime`.

(Could be a follow-up doc PR if you'd rather — but since `CHANGELOG` is the only doc updated in this branch and that's auto-generated, the API page is the only place a user can actually discover these.)

pkg/server/server.go
```diff
@@ -557,6 +565,66 @@ func (s *Server) setSessionStarred(c echo.Context) error {
...
+// (it must have been started at least once or be attached out-of-band)
+// or 422 if the runtime does not support model switching.
+func (s *Server) getSessionModels(c echo.Context) error {
+	sessionID := c.Param("id")
+
+	agentName, current, choices, err := s.sm.AvailableSessionModels(c.Request().Context(), sessionID)
+	if err != nil {
+		switch {
+		case errors.Is(err, ErrModelSwitchingNotSupported):
+			return echo.NewHTTPError(http.StatusUnprocessableEntity, err.Error())
```
aheritier, line 579: **Question**: Mapping `ErrModelSwitchingNotSupported` to 422 is a defensible choice (the URL/method is fine, the runtime just can't honour it), but 501 Not Implemented or 409 Conflict are both more conventional for "this server doesn't support this capability" / "the resource is in a state that disallows this action". 422 traditionally signals "the body parsed but is semantically wrong", which doesn't really match here — there's no body for GET. Curious whether you considered 501 and rejected it. Not blocking, just worth a sentence in the doc-comment justifying the choice.

pkg/server/server.go
```diff
@@ -557,6 +565,66 @@ func (s *Server) setSessionStarred(c echo.Context) error {
 	return c.JSON(http.StatusOK, map[string]string{"status": "updated"})
 }
 
+// getSessionModels lists the models the user can pick from for the
```
aheritier, line 568: **Question** (about the API surface, not this handler specifically): the existing `createSession` binds straight into `session.Session` and `SessionManager.CreateSession` cherry-picks the fields it honours (now extended to `AgentModelOverrides` / `CustomModelsUsed`). A client that POSTs `{"input_tokens": 999, "starred": true, "id": "..."}` will silently have those ignored — correct behaviour, but the contract is implicit.

Worth either (a) introducing an explicit `api.CreateSessionRequest` DTO so the wire surface is type-level visible, or (b) at minimum a doc-comment on `CreateSession` listing exactly which template fields are honoured. Suggestion only; not a blocker for this PR.

pkg/server/session_manager.go
```diff
@@ -729,6 +793,174 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+//
+// On store-write failure the in-memory session state and the runtime
+// override are rolled back so the next call observes a consistent state.
+//
+// Concurrent SetSessionAgentModel calls on the same session are
+// serialised via the session-scoped modelSwitch lock so the runtime,
+// session and store never observe interleaved updates. The manager-wide
+// sm.mux is only held briefly while reading or mutating session fields,
+// never while calling into the runtime or the store.
+func (sm *SessionManager) SetSessionAgentModel(ctx context.Context, sessionID, modelRef string) (string, string, error) {
```
aheritier, line 867: **Suggestion**: `SetSessionAgentModel` returns `(string, string, error)` — readable inline but easy to misuse at call sites (which arg is agent, which is model?). The handler in `server.go` already destructures into `agentName, modelRef`. Consider returning a small named struct (e.g. `type ModelSwitchResult struct { Agent, Model string }`) to make the API self-documenting, or rename to `(agentName, modelRef, err)` in a doc-comment line. Same applies to `AvailableSessionModels` which returns four positional values. Not blocking.

pkg/server/session_manager.go
```diff
@@ -729,6 +793,174 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+		if appendedCustomUsed {
+			sess.CustomModelsUsed = sess.CustomModelsUsed[:prevCustomLen]
+		}
+		sm.mux.Unlock()
+
+		rollback := prevOverride
+		if !hadOverride {
+			rollback = ""
+		}
+		if rbErr := rs.runtime.SetAgentModel(ctx, agentName, rollback); rbErr != nil {
```
aheritier: **Suggestion**: the rollback block uses two passes (in-memory under `sm.mux`, then runtime without it) which is fine, but the rollback ref calculation is duplicated:

```go
rollback := prevOverride
if !hadOverride {
    rollback = ""
}
```

This same `(prevOverride, hadOverride) → ref` projection is what the in-memory rollback computes too. Worth extracting a tiny helper, e.g. `prevRef := prevOverrideRef(hadOverride, prevOverride)` and using it in both rollback branches and as the `SetAgentModel` arg, so a future reader doesn't have to re-derive that "missing key means clear" rule twice. Non-blocking.

pkg/server/session_manager.go
```diff
@@ -729,6 +793,174 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+			slog.ErrorContext(ctx, "Failed to roll back runtime model override", "session_id", sessionID, "agent", agentName, "error", rbErr)
+		}
```
aheritier, line 968: **Question**: rollback calls `rs.runtime.SetAgentModel` with the original `ctx`. If the request was cancelled mid-flight (which is presumably what made the store write fail), the rollback will also fail and we'll log an error but leave the runtime with the new override while the persisted state is the old one. That's the inverse of the invariant the rollback is trying to preserve.

Worth using `context.WithoutCancel(ctx)` (or a fresh background ctx with a short timeout) for the rollback `SetAgentModel` so cancellation doesn't poison the recovery path. Probably a real bug under request-cancellation, but borderline blocking — depends on whether cancelled-during-store-write is a realistic scenario you care about today.

pkg/server/session_manager.go
```diff
@@ -645,6 +682,33 @@ func (sm *SessionManager) loadTeam(ctx context.Context, agentFilename string, ru
...
+// ignored: a stored override that no longer resolves (e.g. because the
+// model was removed from the agent's config) must not prevent the
+// session from being resumed with the agent's default model.
+func applyStoredOverrides(ctx context.Context, sessionID string, run runtime.Runtime, overrides map[string]string) {
+	if len(overrides) == 0 || !run.SupportsModelSwitching() {
+		return
+	}
+	for agentName, modelRef := range overrides {
+		if err := run.SetAgentModel(ctx, agentName, modelRef); err != nil {
+			slog.WarnContext(ctx, "Failed to apply stored model override", "session_id", sessionID, "agent", agentName, "model", modelRef, "error", err)
```
aheritier, line 707: **Question**: `applyStoredOverrides` swallows `SetAgentModel` errors so a stale override doesn't prevent resume — sensible. But the runtime is then left in a state where the in-memory `sess.AgentModelOverrides[agentName]` says "use X", the runtime is actually using the agent default, and a subsequent `GET /sessions/:id/models` will report `current_model_ref = X` but the agent will be running on its default. The picker will mark X as IsCurrent (because `DecorateModelChoices` matches on `currentRef` from the session) — which is a lie.

Worth either: (a) clearing the bad override from the session map when SetAgentModel fails (and logging), or (b) attempting once more lazily on first turn. Either way the current behaviour silently misleads the picker. Not blocking but I'd flag it as a follow-up.

pkg/server/session_manager.go
```diff
@@ -645,6 +682,33 @@ func (sm *SessionManager) loadTeam(ctx context.Context, agentFilename string, ru
...
 }
 
+// loadTeamWithConfig is like loadTeam but also returns the loaded model and
+// provider configuration so the runtime can be wired for model switching.
+func (sm *SessionManager) loadTeamWithConfig(ctx context.Context, agentFilename string, runConfig *config.RuntimeConfig) (*teamloader.LoadResult, error) {
+	agentSource, found := sm.Sources[agentFilename]
+	if !found {
+		return nil, fmt.Errorf("agent not found: %s", agentFilename)
+	}
+
```
aheritier, line 692: **Nit**: `loadTeam` and `loadTeamWithConfig` differ only in which loader function they call (`teamloader.Load` vs `teamloader.LoadWithConfig`) and the return type. Now that every caller of `loadTeam` outside this file is gone (or could be), consider just inlining `loadTeamWithConfig` in `runtimeForSession` and dropping the helper, or replacing `loadTeam`'s body to delegate to `LoadWithConfig` and return only `.Team`. Two near-identical wrappers around the same source-lookup is a small smell. Non-blocking.

pkg/server/session_manager.go
```diff
@@ -729,6 +793,174 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+		delete(sess.AgentModelOverrides, agentName)
+	} else {
+		if sess.AgentModelOverrides == nil {
+			sess.AgentModelOverrides = make(map[string]string)
+		}
+		sess.AgentModelOverrides[agentName] = modelRef
+
+		// Track inline provider/model references so they remain easy to
+		// re-select via the model picker (mirrors App.SetCurrentAgentModel).
+		if strings.Contains(modelRef, "/") && !slices.Contains(sess.CustomModelsUsed, modelRef) {
```
aheritier: **Nit**: tiny copy-paste: this `slices.Contains(sess.CustomModelsUsed, modelRef)` track-on-set logic is already implemented in `App.trackCustomModel` (per the comment on the next line). Worth either making `trackCustomModel` a method on `*session.Session` so both code paths share it, or at least a `// keep in sync with App.trackCustomModel` reminder. Non-blocking.

pkg/runtime/model_switcher.go
```diff
@@ -16,49 +16,136 @@ import (
...
+	OutputLimit int64 `json:"output_limit,omitempty"`
 	// InputModalities lists the input modalities supported by the model
 	// (e.g., "text", "image", "audio").
-	InputModalities []string
+	InputModalities []string `json:"input_modalities,omitempty"`
 	// OutputModalities lists the output modalities the model can produce.
-	OutputModalities []string
+	OutputModalities []string `json:"output_modalities,omitempty"`
+}
+
```
aheritier, line 66: **Suggestion**: `ModelChoice` is now part of the wire format (per the new doc-comment) — every public field has a JSON tag. The catalog-only fields below (`Family`, `InputCost`, `OutputCost`, `CacheReadCost`, `CacheWriteCost`, `ContextLimit`, `OutputLimit`, `InputModalities`, `OutputModalities`) are now serialised on every response; that's a lot of optional surface to commit to. If any of those fields might rename or restructure (e.g. consolidating costs into a sub-object), now is the time to consider it before clients depend on them.

Also: `Ref` and `Name` carry no `omitempty` while everything else does — intentional? They'll always be present, so the asymmetry is fine, just calling it out.

pkg/runtime/model_switcher.go
```diff
@@ -16,49 +16,136 @@ import (
...
+	// choice so the picker can still highlight the active selection.
+	if !currentFound && strings.Contains(currentRef, "/") {
+		prov, name, _ := strings.Cut(currentRef, "/")
+		result = append(result, ModelChoice{
+			Name:      currentRef,
+			Ref:       currentRef,
+			Provider:  prov,
+			Model:     name,
+			IsCurrent: true,
+			IsCustom:  true,
```
aheritier, line 144: **Question**: the synthesized choice for an inline override only fires when the ref contains `/`. If the override is a config key (`my_model`) that has been removed from the agent config since the session was created, the picker silently won't highlight anything and `IsCurrent` is false everywhere. The unit test "non-provider override does not synthesize a fabricated choice" enshrines this as intentional — but is that really what callers want? An entry like `{Name: currentRef, Ref: currentRef, IsCurrent: true, IsCustom: true}` (with empty Provider/Model) would at least let the UI surface "you're on a model that no longer exists in the config". Non-blocking design call.

pkg/runtime/model_switcher.go
```diff
@@ -16,49 +16,136 @@ import (
...
+	OutputModalities []string `json:"output_modalities,omitempty"`
+}
+
+// SessionModelsResponse is the response returned by
+// GET /api/sessions/:id/models. CurrentModelRef is the active override for
+// the named agent (empty when the agent is using its configured default).
+type SessionModelsResponse struct {
+	Agent           string        `json:"agent"`
+	CurrentModelRef string        `json:"current_model_ref,omitempty"`
+	Models          []ModelChoice `json:"models"`
```
aheritier, line 73: **Question**: the response embeds `ref` for the agent default but no other notion of "default" beyond the per-choice `IsDefault` flag. Should `SessionModelsResponse` include a `DefaultModelRef string` too, so a client showing "Reset to default" doesn't have to scan `Models[].IsDefault`? Minor; ignore if you'd rather keep it lean.

pkg/api/types.go
```diff
@@ -270,3 +270,15 @@ type SessionStatusResponse struct {
 	OutputTokens int64  `json:"output_tokens"`
 	NumMessages  int    `json:"num_messages"`
 }
+
+// SetSessionModelRequest is the body of PATCH /api/sessions/:id/model.
+// An empty Model clears the override and reverts to the agent's default.
+type SetSessionModelRequest struct {
+	Model string `json:"model"`
```
aheritier, line 277: **Nit**: missing the `Assisted-By: docker-agent` trailer on a couple of commits in this branch (`feat: support model overrides when creating sessions` uses `Assisted-By: Docker Agent` — capitalised — and a couple of the later commits don't have a body at all). Project convention per `git-commit-conventions` is the lowercase `docker-agent` form. Cosmetic, but worth fixing if you squash-merge with a clean trailer; harmless if you merge as-is.

pkg/app/app.go
```diff
@@ -777,76 +777,15 @@ func (a *App) AvailableModels(ctx context.Context) []runtime.ModelChoice {
 	if !a.runtime.SupportsModelSwitching() {
 		return nil
 	}
-	models := a.runtime.AvailableModels(ctx)
 
-	// Determine the currently active model for this agent
 	agentName := a.runtime.CurrentAgentName()
```
aheritier, line 781: **Praise**: clean refactor — 66 lines deleted, 5 added, identical behaviour, and the extracted helper is now exhaustively unit-tested. Good catch on factoring this out.

dgageot:
Addressed the blocking docs feedback from @aheritier in commit da48ca80f:

Added all three model-switching endpoints to the API server documentation:
- `GET /api/sessions/:id/models` — List available models for the session's current agent
- `PATCH /api/sessions/:id/model` — Set or clear the agent's model override
- `POST /api/sessions/:id/model` — Set or clear the agent's model override (backward compat with RemoteRuntime)

The endpoints are now documented in `docs/features/api-server/index.md` in the Sessions table.

All commits pass `task lint`, `task test`, and `task build`. Branch has been rebased on `origin/main` and force-pushed.

pkg/server/session_manager.go
```diff
@@ -729,6 +793,174 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+			sess.CustomModelsUsed = append(sess.CustomModelsUsed, modelRef)
+			appendedCustomUsed = true
+		}
+	}
+	sm.mux.Unlock()
+
+	if err := sm.sessionStore.UpdateSession(ctx, sess); err != nil {
```
docker-agent: **[MEDIUM] Stale model override briefly visible to `AvailableSessionModels` during store-write window**

Between the `sm.mux.Unlock()` on line 931 (after mutating the in-memory session) and the `sm.mux.Lock()` inside the rollback block, `UpdateSession` is called without holding `sm.mux`. Because this is an I/O call, it can take arbitrarily long. During this window, a concurrent `GET /api/sessions/:id/models` request calls `AvailableSessionModels`, which acquires `sm.mux` and reads `rs.session.AgentModelOverrides[agentName]` — observing the not-yet-persisted (and about-to-be-rolled-back) override.

The code comment says *"Roll back in-memory under sm.mux first so concurrent readers never see the half-applied state"*, but that guarantee doesn't hold: `sm.mux` is released before the rollback lock is re-acquired, so there is a real window.

**Impact**: `GET /models` can transiently report a model override that was ultimately never persisted (because the store write failed). The in-memory state is corrected by the rollback, but any client that polled the endpoint during the window would see an incorrect `IsCurrent` marker.

**Suggested fix**: Keep the session mutation and the store write under a single critical section, or (if the I/O overhead is unacceptable) hold the `modelSwitch` lock long enough that the mutation + store write + any rollback are treated as an atomic unit visible only after the store confirms success. Alternatively, mutate the in-memory session *after* the store write succeeds, never before.

docker-agent (commented):

One concurrency consistency issue was found in the new model-switching logic. No data corruption or panics — the impact is limited to a transient stale read from  during a store-write failure + rollback scenario.

**Reviewed files** (risk-score-weighted):
- `pkg/server/session_manager.go` (risk: 6) — thread-safety of SetSessionAgentModel/AvailableSessionModels ✅ (one issue found, see inline)
- `pkg/server/server.go` (risk: 1) — HTTP status code mapping ✅ correct
- `pkg/runtime/model_switcher.go` — DecorateModelChoices logic ✅ correct
- `pkg/server/session_models_test.go` — test coverage ✅ comprehensive

pkg/server/session_manager.go
```diff
@@ -729,6 +793,203 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+	if err := sm.sessionStore.UpdateSession(ctx, updatedSess); err != nil {
+		rollback := prevOverride
+		if !hadOverride {
+			rollback = ""
+		}
+		if rbErr := rs.runtime.SetAgentModel(ctx, agentName, rollback); rbErr != nil {
+			slog.ErrorContext(ctx, "Failed to roll back runtime model override", "session_id", sessionID, "agent", agentName, "error", rbErr)
+		}
+		return "", "", fmt.Errorf("failed to persist model override: %w", err)
+	}
```
aheritier, line 970: **Non-blocking** — Nice fix on the clone-then-persist pattern — this kills the 'picker shows new override that gets rolled back' window. There's a smaller residual window between `runtime.SetAgentModel` succeeding and the post-`UpdateSession` `sm.mux.Lock()`: during that gap the runtime is on the new model but `AvailableSessionModels` reads the old override from `rs.session`, so `GET /models` reports `IsCurrent` on the previous model while requests already execute against the new one. Strictly less bad than what was fixed (briefly stale, never fictitious). Closing it would mean holding `modelSwitch` around the GET path — non-trivial. A short doc-comment acknowledging the residual window or a follow-up ticket would be enough. Not a merge blocker.

pkg/server/session_manager.go
```diff
@@ -729,6 +793,203 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+
+	agentName := rs.runtime.CurrentAgentName()
+	sess := rs.session
+
+	// Snapshot current state so we can roll back if persistence fails
+	// after we've already mutated the runtime.
+	var (
+		hadOverride     bool
+		prevOverride    string
+		hadOverridesMap bool
```
aheritier, line 888: **Nit** — `hadOverridesMap` is only used as the `if` condition immediately below — inline the check and drop the variable. golangci-lint misses it because the assignment makes it look 'used'.

pkg/runtime/model_switcher.go
```diff
@@ -16,49 +16,136 @@ import (
...
+}
```
aheritier, line 74: **Question** — `SessionModelsResponse` is a brand-new wire type but lives in `pkg/runtime` next to `ModelChoice`. The companion `SetSessionModelRequest`/`SetSessionModelResponse` sit in `pkg/api`. The PR description justifies keeping `ModelChoice` in `pkg/runtime` because it pre-existed there — but `SessionModelsResponse` is new code introduced specifically as the wire envelope. Either move it to `pkg/api` (consistent with the other request/response types) or move all three to `pkg/runtime` and explain in `pkg/api/types.go` why model types are an exception. Mixed placement makes `pkg/api` incomplete as a 'what's on the wire' reference.

pkg/server/session_manager.go
```diff
@@ -729,6 +793,203 @@ func (sm *SessionManager) SetSessionStarred(ctx context.Context, sessionID strin
...
+	}
+
+	// Persist the cloned session. If this fails, the live session is
+	// unchanged and we only need to roll back the runtime.
+	if err := sm.sessionStore.UpdateSession(ctx, updatedSess); err != nil {
+		rollback := prevOverride
+		if !hadOverride {
+			rollback = ""
+		}
+		if rbErr := rs.runtime.SetAgentModel(ctx, agentName, rollback); rbErr != nil {
```
aheritier, line 966: **Non-blocking** — Already raised by @aheritier in round 2 and not addressed in head. Rollback still uses the request `ctx` — if the original request was cancelled and that's what caused `UpdateSession` to fail, the rollback inherits the cancellation. `context.WithoutCancel(ctx)` (or a fresh background ctx with a short timeout) would protect the recovery path. One-line change with a clear correctness argument; worth fixing now.

aheritier (approved):
All Blocking items from prior review rounds are resolved and verified by new tests; CI is fully green. Approving. Items below are non-blocking polish — feel free to address in a follow-up or defer.

✅ Resolved and verified:
- DecorateModelChoices no longer mutates input; defensive-copy test added.
- setSessionModel default branch returns 500, matching getSessionModels; store-failure test added.
- sm.mux no longer held across runtime calls; per-session modelSwitch lock + contention tests added.
- Three new endpoints documented in docs/features/api-server/index.md.

Nice work on the app.go refactor — 67 lines → 5-line delegation with full helper coverage.

PR #2794, opened 2026-05-13
feat: add mcp_catalog toolset for on-demand MCP server discovery

dgageot:
## Summary

Adds a new built-in toolset, `mcp_catalog`, that lets agents discover and on-demand activate remote (streamable-http) MCP servers from the Docker MCP Catalog without declaring each one ahead of time in YAML.

The catalog (45 servers) is embedded at build time from `https://desktop.docker.com/mcp/catalog/v3/catalog.json` (filtered to `type=remote AND remote.transport_type=streamable-http`). The toolset surfaces five meta-tools to the model:

- `search_remote_mcp_servers` — case-insensitive fuzzy search over id / title / description / category / tags.
- `list_remote_mcp_servers` — show currently enabled servers.
- `enable_remote_mcp_server` — instantiate an `*mcp.Toolset` for the server (deferred — no network until tools are next enumerated).
- `disable_remote_mcp_server` — stop the toolset and remove its tools.
- `reset_remote_mcp_server_auth` — wipe persisted OAuth credentials so the next enable triggers a fresh authorization flow.

## How it works

- Each enabled `*mcp.Toolset` is wrapped in a `tools.NewStartable` so it gets the same lazy single-flight Start semantics as YAML-declared toolsets.
- `Tools()` lazily Start()s each enabled wrapped toolset on first enumeration. Treats `mcp.IsAuthorizationRequired` as expected deferral (silent on the non-interactive startup probe; user-facing OAuth dialog on the next interactive turn). Real Start failures get the StartableToolSet de-duplicated warning streak.
- Elicitation / OAuth-success / managed-OAuth / tools-changed handlers captured by the runtime are re-applied to each new mcp.Toolset on enable so the OAuth dance behaves identically to a YAML-declared `mcp.remote` toolset.
- `${ENV_VAR}` placeholders in catalog headers (e.g. Apify's `Authorization: Bearer ${APIFY_API_KEY}`) resolve at enable time via `js.Expander` — same code path the YAML toolsets use.
- `reset_remote_mcp_server_auth` stops the inner toolset (so the live MCP session drops its now-stale token) before notifying the runtime and clearing the keyring entry. The notify fires even when the keyring removal fails, so the runtime's tool list never desyncs from the actual `enabled` state.

## Configuration

```yaml
agents:
  root:
    model: anthropic/claude-sonnet-4-6
    toolsets:
      - type: mcp_catalog
```

See `examples/mcp_catalog.yaml` for a full example.

## What's in this branch

- `pkg/tools/builtin/mcpcatalog/{mcpcatalog.go, servers.go, servers.json}` — toolset, embedded catalog (45 servers).
- `pkg/teamloader/registry.go` — registers the new `mcp_catalog` toolset type.
- `agent-schema.json` — adds `mcp_catalog` to the toolset-type enums.
- `examples/mcp_catalog.yaml` — example agent using `anthropic/claude-sonnet-4-6`.
- `pkg/tools/builtin/mcpcatalog/mcpcatalog_test.go` — 19 tests (load, search, lifecycle, race-friendly concurrent enable/disable, ctx cancellation, lazy-start regression test against a fake MCP server, auth-required deferral, reset_auth happy path / unknown id / no-op for non-oauth / disable-on-reset / error surfacing / notify-on-keyring-failure).

## Known limitation

The runtime's MCP-prompt discovery uses `tools.As[*mcp.Toolset]` to find prompt-capable toolsets. Because `mcp_catalog` is a container, prompts exposed by servers activated through this catalog are **not** surfaced via `/prompts`. Tools (the primary interface) work fine. Plumbing prompt discovery through container toolsets would require a new interface and changes to `pkg/runtime/runtime.go` — out of scope for this PR.

## Validation

- `task build` ✅
- `go vet ./...` ✅
- `golangci-lint run ./...` → 0 issues ✅
- `go test ./pkg/tools/builtin/mcpcatalog/... -race -count=1` → all 19 tests pass ✅

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
+++ b/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
@@ -1,0 +1,505 @@
+// Package mcpcatalog exposes the Docker MCP Catalog's remote
+// streamable-http servers as a single agent-side toolset that supports
+// on-demand activation.
+//
+// The toolset surfaces three meta-tools to the model:
+//
+//   - search_remote_mcp_servers — case-insensitive fuzzy search over the
+//     curated catalog (id / title / description / category / tags).
+//   - enable_remote_mcp_server  — instantiate an *mcp.Toolset for a server
+//     (defers the actual TCP connect / OAuth handshake until the model
+//     calls one of the server's tools).
+//   - disable_remote_mcp_server — stop the toolset and remove its tools.
+//
+// Activated servers' tools are merged into Tools(); tool list changes are
+// reported via a tools.ChangeNotifier handler so the runtime refreshes
+// the LLM's tool catalogue as soon as a server is enabled or disabled.
+//
+// The expensive parts — DNS, TCP, MCP handshake, OAuth flow — happen on
+// the *first* tool call, courtesy of mcp.Toolset's lifecycle.Supervisor.
+// If the server requires OAuth, the existing elicitation pipeline picks
+// it up and surfaces an authorization URL to the user.
+package mcpcatalog
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/docker/docker-agent/pkg/environment"
+	"github.com/docker/docker-agent/pkg/js"
+	"github.com/docker/docker-agent/pkg/tools"
+	"github.com/docker/docker-agent/pkg/tools/mcp"
+)
+
+const (
+	ToolNameSearch  = "search_remote_mcp_servers"
+	ToolNameEnable  = "enable_remote_mcp_server"
+	ToolNameDisable = "disable_remote_mcp_server"
+	ToolNameList    = "list_remote_mcp_servers"
+)
+
+// Toolset implements on-demand activation of remote (streamable-http) MCP
+// servers from the Docker MCP Catalog.
+type Toolset struct {
+	catalog  *Catalog
+	byID     map[string]Server
+	expander *js.Expander
+	env      environment.Provider
+
+	mu      sync.RWMutex
+	enabled map[string]*mcp.Toolset
+
+	// elicitationHandler / oauthSuccessHandler / toolsChangedHandler are
+	// captured by the runtime via tools.As[...] *before* any server is
+	// enabled; we re-attach them to each new mcp.Toolset on enable so
+	// OAuth elicitation, OAuth-success refreshes and tool-list changes
+	// behave identically to a YAML-declared `mcp.remote` toolset.
+	elicitationHandler  tools.ElicitationHandler
+	oauthSuccessHandler func()
+	toolsChangedHandler func()
+}
+
+var (
+	_ tools.ToolSet        = (*Toolset)(nil)
+	_ tools.Startable      = (*Toolset)(nil)
+	_ tools.Instructable   = (*Toolset)(nil)
+	_ tools.Describer      = (*Toolset)(nil)
+	_ tools.ChangeNotifier = (*Toolset)(nil)
+	_ tools.Elicitable     = (*Toolset)(nil)
+	_ tools.OAuthCapable   = (*Toolset)(nil)
+)
+
+// New returns a Toolset backed by the embedded catalog. envProvider is used
+// to resolve ${ENV_VAR} placeholders in catalog headers (e.g. the Apify
+// `Authorization: Bearer ${APIFY_API_KEY}` header) at enable time, mirroring
+// how a YAML-declared `mcp.remote` toolset works.
+func New(envProvider environment.Provider) *Toolset {
+	cat := MustLoad()
+	byID := make(map[string]Server, len(cat.Servers))
+	for _, s := range cat.Servers {
+		byID[s.ID] = s
+	}
+	return &Toolset{
+		catalog:  cat,
+		byID:     byID,
+		expander: js.NewJsExpander(envProvider),
+		env:      envProvider,
+		enabled:  make(map[string]*mcp.Toolset),
+	}
+}
+
+// Describe returns a short, user-visible label for the /tools dialog.
+func (t *Toolset) Describe() string {
+	return fmt.Sprintf("mcp_catalog(remote streamable-http, %d servers)", t.catalog.Count)
+}
+
+// Instructions tell the model how to discover and activate servers.
+func (t *Toolset) Instructions() string {
+	return `## Remote MCP Catalog
+
+You have access to a curated catalog of remote MCP servers (Docker MCP
+Catalog, streamable-http only). They are NOT active by default.
+
+Workflow:
+  1. Call ` + ToolNameSearch + ` with a keyword to discover matching servers.
+     Use any term related to the user's intent ("notion", "stripe",
+     "docs", "search", "browser", …).
+  2. Call ` + ToolNameEnable + ` with the server's "id" to activate it.
+     This adds the server's tools to your set. Authentication (OAuth or
+     API key) is deferred — it's triggered on the first tool call.
+     For api_key servers, make sure the listed env var(s) are set in the
+     user's shell BEFORE enabling, otherwise the first call will fail.
+  3. Use the newly activated tools as you would any other.
+  4. Call ` + ToolNameDisable + ` to remove a server when no longer needed.
+
+Prefer enabling only the servers you actually need — every server adds
+tools to the prompt and contributes to context usage.`
+}
+
+// Start is a no-op: the catalog is embedded and no servers are auto-enabled.
+// Lifecycle for individual MCP toolsets is managed when Enable / Disable
+// are invoked.
+func (t *Toolset) Start(context.Context) error { return nil }
+
+// Stop tears down every enabled MCP toolset. Errors are logged but do not
+// abort the loop so a misbehaving server can't block agent shutdown.
+func (t *Toolset) Stop(ctx context.Context) error {
+	t.mu.Lock()
+	enabled := t.enabled
+	t.enabled = make(map[string]*mcp.Toolset)
+	t.mu.Unlock()
+
+	for id, ts := range enabled {
+		if err := ts.Stop(ctx); err != nil {
+			slog.WarnContext(ctx, "Failed to stop remote MCP toolset", "id", id, "error", err)
+		}
+	}
+	return nil
+}
+
+// SetElicitationHandler is captured here and re-attached to every freshly
+// activated MCP toolset so OAuth flows can prompt the user.
+func (t *Toolset) SetElicitationHandler(handler tools.ElicitationHandler) {
+	t.mu.Lock()
+	t.elicitationHandler = handler
+	enabled := t.snapshotEnabled()
+	t.mu.Unlock()
+	for _, ts := range enabled {
+		ts.SetElicitationHandler(handler)
+	}
+}
+
+// SetOAuthSuccessHandler is captured here and re-attached to every freshly
+// activated MCP toolset so the runtime refreshes its tool list once OAuth
+// completes.
+func (t *Toolset) SetOAuthSuccessHandler(handler func()) {
+	t.mu.Lock()
+	t.oauthSuccessHandler = handler
+	enabled := t.snapshotEnabled()
+	t.mu.Unlock()
+	for _, ts := range enabled {
+		ts.SetOAuthSuccessHandler(handler)
+	}
+}
+
+// SetManagedOAuth forwards the managed-OAuth flag to every enabled
+// toolset; new toolsets pick it up at enable time.
+func (t *Toolset) SetManagedOAuth(managed bool) {
+	t.mu.Lock()
+	enabled := t.snapshotEnabled()
+	t.mu.Unlock()
+	for _, ts := range enabled {
+		ts.SetManagedOAuth(managed)
+	}
+}
+
+// SetToolsChangedHandler is invoked by the runtime to be notified when
+// the set of available tools changes. We forward to the activated MCP
+// toolsets *and* call it ourselves on every Enable / Disable so the
+// runtime sees the meta-tool surface change too.
+func (t *Toolset) SetToolsChangedHandler(handler func()) {
+	t.mu.Lock()
+	t.toolsChangedHandler = handler
+	enabled := t.snapshotEnabled()
+	t.mu.Unlock()
+	for _, ts := range enabled {
+		ts.SetToolsChangedHandler(handler)
+	}
+}
+
+// snapshotEnabled returns the currently enabled toolsets as a fresh slice.
+// Caller MUST hold t.mu (read or write). Used to forward setter calls
+// outside the critical section.
+func (t *Toolset) snapshotEnabled() []*mcp.Toolset {
+	out := make([]*mcp.Toolset, 0, len(t.enabled))
+	for _, ts := range t.enabled {
+		out = append(out, ts)
+	}
+	return out
+}
+
+// Tools returns the meta-tools plus every tool exposed by an activated
+// remote MCP server. Tools from unactivated servers are intentionally
+// hidden so they don't bloat the prompt.
+func (t *Toolset) Tools(ctx context.Context) ([]tools.Tool, error) {
+	result := []tools.Tool{
+		{
+			Name:         ToolNameSearch,
+			Category:     "mcp_catalog",
+			Description:  "Search the Docker MCP Catalog for remote streamable-http MCP servers matching a keyword. Returns id, title, description, auth requirements and category for each hit.",
+			Parameters:   tools.MustSchemaFor[SearchArgs](),
+			OutputSchema: tools.MustSchemaFor[string](),
+			Handler:      tools.NewHandler(t.handleSearch),
+			Annotations: tools.ToolAnnotations{
+				Title:        "Search remote MCP servers",
+				ReadOnlyHint: true,
+			},
+		},
+		{
+			Name:         ToolNameList,
+			Category:     "mcp_catalog",
+			Description:  "List currently enabled remote MCP servers and their connection state.",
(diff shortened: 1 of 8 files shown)


pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+	t.mu.Lock()
+	if _, exists := t.enabled[id]; exists {
+		t.mu.Unlock()
+		return tools.ResultSuccess(fmt.Sprintf("server %q is already enabled", id)), nil
+	}
+
+	// Pre-flight: warn (don't block) if an api_key server is missing its env var.
+	// We do not block because the user may set the variable later, or rely on
+	// the model to surface the error from the first tool call.
+	missing := t.missingAPIKeyEnv(ctx, server)
```
docker-agent: **[MEDIUM] Blocking external calls held under write lock in `handleEnable`**

`missingAPIKeyEnv` (line 475) and `t.expander.ExpandMap` (line 477) are both called while `t.mu` is write-locked (acquired at line 466, released at line 499). `missingAPIKeyEnv` calls `t.env.Get(ctx, sec.Env)` which is an external provider that may hit the OS keychain or a network service. `ExpandMap` similarly invokes an external expander interface.

Holding a write lock during these calls blocks **all** concurrent readers — including `Tools()` (which every agent turn calls), `handleSearch`, and `handleList` — for the full duration of the external I/O. For servers with multiple `api_key` secrets (e.g. Apify with `APIFY_API_KEY`), each secret triggers a separate `env.Get` round-trip.

Both functions operate exclusively on `server`, which comes from `t.byID` — a map that is read-only after construction and requires no mutex protection. They should be moved before the `t.mu.Lock()` call:

```go
// Perform slow external calls BEFORE acquiring the lock —
// server data is immutable, no protection needed here.
missing := t.missingAPIKeyEnv(ctx, server)
headers := t.expander.ExpandMap(ctx, server.Headers)

t.mu.Lock()
if _, exists := t.enabled[id]; exists {
    t.mu.Unlock()
    return tools.ResultSuccess(...), nil
}
// ... rest of the enable logic
```

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+	}
+	t.mu.RUnlock()
+
+	for _, e := range enabled {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+
+		if !e.ts.IsStarted() {
+			if err := e.ts.Start(ctx); err != nil {
```
docker-agent, line 344: **[MEDIUM] Concurrent `Disable` during `Tools()` iteration may restart a just-disabled server**

`Tools()` snapshots `t.enabled` under `RLock` (lines 326–331), then iterates the snapshot **outside** the lock. If `handleDisable` runs concurrently it:
1. Removes the server from `t.enabled` and releases the lock
2. Calls `wrapped.Stop(ctx)` — marking the `StartableToolSet` as stopped

Meanwhile, `Tools()` still holds a reference to the same `*tools.StartableToolSet` and, on line 339, calls `e.ts.Start(ctx)` if `IsStarted()` is false — which it is, because `Stop` just cleared it.

The result is that a server the user explicitly disabled can be **silently reconnected** within the same `Tools()` invocation, restoring its tools to the active set until the next invocation. Whether `StartableToolSet.Start()` honours a post-`Stop()` call as a no-op is the key question; if it doesn't reject the call, this is a real reconnection.

A simple mitigation is to re-check membership in `t.enabled` before starting:

```go
for _, e := range enabled {
    if err := ctx.Err(); err != nil {
        return nil, err
    }

    // Skip servers that were disabled after we snapshotted.
    t.mu.RLock()
    _, stillEnabled := t.enabled[e.id]
    t.mu.RUnlock()
    if !stillEnabled {
        continue
    }

    if !e.ts.IsStarted() {
        // ... existing start logic
    }
    // ...
}
```

docker-agent (commented):

Two medium-severity findings in `pkg/tools/builtin/mcpcatalog/mcpcatalog.go` — see inline comments for details.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+
+	msg := strings.Builder{}
+	fmt.Fprintf(&msg, "enabled %q (%s) — connection deferred until the host next enumerates tools.\n", id, server.Title)
+	fmt.Fprintf(&msg, "endpoint: %s\n", server.URL)
+	switch server.Auth.Type {
+	case "oauth":
+		msg.WriteString("auth: OAuth — an authorization URL will be elicited on the next turn.\n")
+	case "api_key":
+		if len(missing) > 0 {
+			fmt.Fprintf(&msg, "auth: API key — WARNING: the following env vars are NOT set: %s. Set them and re-enable, otherwise tool calls will fail.\n", strings.Join(missing, ", "))
```
aheritier: **Suggestion** — the `WARNING ... Set them and re-enable` text contradicts what the next call actually does. Re-running `enable_remote_mcp_server` for the same id returns *"already enabled"* (line 467–470) without re-expanding the headers, because env vars are baked in at enable time via `t.expander.ExpandMap` on line 478. So the documented recovery path doesn't work — the user has to disable, then enable, to pick up newly-set env vars.

Two options:
1. Refuse the enable when required env vars are missing (return `tools.ResultError`, leave `t.enabled` untouched). Cleaner UX, and the model can react to the error result. Requires updating `TestEnableAPIKeyMissingEnv` which currently asserts `IsError == false`.
2. Keep the warn-and-enable behaviour but rewrite the message: *"Set them, then call `disable_remote_mcp_server` and `enable_remote_mcp_server` again"* — minimum-impact fix.

I'd take option 1; the failure mode is less surprising for the model and avoids leaving a half-broken toolset around.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+		{
+			Name:         ToolNameResetAuth,
+			Category:     "mcp_catalog",
+			Description:  "Clear persisted OAuth credentials (access token, refresh token, dynamic-client-registration data) for a catalog server. The next enable will trigger a fresh authorization flow. No-op for api_key/none servers.",
+			Parameters:   tools.MustSchemaFor[ResetAuthArgs](),
+			OutputSchema: tools.MustSchemaFor[string](),
+			Handler:      tools.NewHandler(t.handleResetAuth),
+			Annotations: tools.ToolAnnotations{
+				Title:           "Reset remote MCP server auth",
+				DestructiveHint: new(true),
```
aheritier, line 320: **Question** — only `reset_remote_mcp_server_auth` has `DestructiveHint: new(true)`. `disable_remote_mcp_server` also has user-visible side effects (drops live tools, kills an in-flight session) and `enable_remote_mcp_server` opens network connections that may persist credentials. Was the asymmetry deliberate (only credential deletion is "destructive enough"), or worth flagging the others too?

Side note: `new(true)` is a tidy Go 1.26-ism but unusual in this codebase — `mcp.go` uses `&[]bool{true}[0]` patterns elsewhere. Either pick one and stick to it, or add a tiny `ptr.Of` helper. Not blocking, just consistency.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+	_ tools.Elicitable     = (*Toolset)(nil)
+	_ tools.OAuthCapable   = (*Toolset)(nil)
+)
+
+// New returns a Toolset backed by the embedded catalog. envProvider is used
+// to resolve ${ENV_VAR} placeholders in catalog headers (e.g. the Apify
+// `Authorization: Bearer ${APIFY_API_KEY}` header) at enable time, mirroring
+// how a YAML-declared `mcp.remote` toolset works.
+func New(envProvider environment.Provider) *Toolset {
+	cat := MustLoad()
```
aheritier, line 112: **Nit** — `MustLoad()` re-decodes the embedded JSON (1129 lines) on every `New()` call. Today there's typically one `Toolset` per agent so this is invisible, but multi-agent configs and any future `mcp serve` plumbing will pay the tax repeatedly. A `sync.OnceValues` wrapper around `Load()` (or returning `*Catalog` from a package-level var initialized by `MustLoad()`) would amortize it for free, and `byID` could be cached the same way since servers are immutable.

Pre-existing concern: this also makes `New()` panic at startup time if the embedded JSON is ever corrupted in the build, which `embed` should prevent — but it's worth keeping the failure mode in mind.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+			Annotations: tools.ToolAnnotations{
+				Title:           "Reset remote MCP server auth",
+				DestructiveHint: new(true),
+			},
+		},
+	}
+
+	t.mu.RLock()
+	enabled := make([]enabledServer, 0, len(t.enabled))
+	for id, ts := range t.enabled {
```
aheritier, line 327: **Nit** — the iteration order of `t.enabled` is undefined, so `Tools()` returns the merged tool list in a different order on each call. The runtime probably doesn't care, but model-side prompt caches are sometimes order-sensitive, and stable order would also stabilize TUI rendering.

Cheap fix: sort `enabled` by `id` after the snapshot, before the `for _, e := range enabled` loop. The cost is one `sort.Slice` over what's typically a single-digit number of entries.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+		return tools.ResultSuccess(fmt.Sprintf("server %q is already enabled", id)), nil
+	}
+
+	// Pre-flight: warn (don't block) if an api_key server is missing its env var.
+	// We do not block because the user may set the variable later, or rely on
+	// the model to surface the error from the first tool call.
+	missing := t.missingAPIKeyEnv(ctx, server)
+
+	headers := t.expander.ExpandMap(ctx, server.Headers)
+	mcpToolset := mcp.NewRemoteToolset(id, server.URL, server.Transport, headers, nil)
```
aheritier, line 536: **Question** — `nil` is passed for the `*latest.RemoteOAuthConfig`. That means catalog-activated OAuth servers always go through the default flow (DCR, default callback port, no scope override). Some of the catalog entries (Atlassian, Notion, …) may need specific scopes, a fixed `client_id`, or a custom callback redirect URL when running headless.

Two questions:
1. Is the assumption that *all* current catalog servers work with pure DCR + defaults, or is there an implicit prerequisite that we should document on the toolset?
2. Should `Auth` in `Server` carry optional fields (`scopes`, `client_id`, `callback_redirect_url`) that get plumbed through here when present, so future catalog additions don't require code changes?

Not blocking — the current set apparently works — but worth a comment near the `nil` so the next reader knows it's intentional.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+	var missing []string
+	for _, sec := range s.Auth.Secrets {
+		if sec.Env == "" {
+			continue
+		}
+		if v, ok := t.env.Get(ctx, sec.Env); !ok || v == "" {
+			missing = append(missing, sec.Env)
+		}
+	}
+	return missing
```
aheritier, line 633: **Nit** — `missingAPIKeyEnv` only inspects `Auth.Secrets`, but the actual values that need to resolve at runtime are the `${...}` placeholders in `server.Headers`. They almost always agree (the Apify entry has `Authorization: Bearer ${APIFY_API_KEY}` and `Secrets: [{Env: APIFY_API_KEY}]`), but nothing enforces it — a future catalog entry that uses an env var in a header without listing it under Secrets would silently fail with no warning.

Cheap defence: scan the expanded headers for unresolved `${VAR}` patterns *after* `t.expander.ExpandMap` and surface those, instead of (or in addition to) the `Secrets` walk.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,664 @@
...
+			URL:         s.URL,
+			Enabled:     isEnabled,
+		})
+	}
+
+	if len(matches) == 0 {
+		return tools.ResultError(fmt.Sprintf("no remote MCP servers match %q (catalog has %d entries)", args.Query, t.catalog.Count)), nil
+	}
+
+	sort.Slice(matches, func(i, j int) bool { return matches[i].ID < matches[j].ID })
```
aheritier, line 458: **Nit** — empty query returns the full catalog (45 entries today, growing) as a JSON blob shoved into the LLM context window — that's roughly 8–12 KB of tokens with descriptions and tags. For a model on its first turn that just wants "is there anything here?", that's a lot.

Two small mitigations worth considering:
- Cap the result set when `q == ""` (e.g. 25 hits) and tell the model to refine.
- Truncate `Description` past N characters in `SearchResult` (the full text is in the catalog if the model needs more).

Non-blocking; can be a follow-up once we see real usage.

pkg/tools/builtin/mcpcatalog/mcpcatalog_test.go
```diff
@@ -0,0 +1,727 @@
...
+		removedURLs = append(removedURLs, url)
+		return nil
+	}
+
+	var oauthServer Server
+	for _, s := range ts.catalog.Servers {
+		if s.Auth.Type == "oauth" {
+			oauthServer = s
+			break
+		}
```
aheritier, line 535: **Praise** — `TestToolsAuthRequiredIsDeferred` is the right test to write for this feature: it verifies the on-demand contract end-to-end (probe with `WithoutInteractivePrompts`, expect meta-tools to come back and the auth-required server to be silently deferred) using a real HTTP server returning a 401 with `WWW-Authenticate`. Future regressions in the deferral path will show up here loud and clear. Same goes for `TestResetAuthNotifiesEvenWhenKeyringFails` — the state-vs-notification invariant is exactly what someone would otherwise break in six months.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,678 @@
...
+		// and potentially restart it on the next line.
+		t.mu.RLock()
+		_, stillEnabled := t.enabled[e.id]
+		t.mu.RUnlock()
+		if !stillEnabled {
+			continue
+		}
+
+		if !e.ts.IsStarted() {
+			if err := e.ts.Start(ctx); err != nil {
```
docker-agent, line 344: **[MEDIUM] TOCTOU race: disabled server may be restarted via `Start()` after `Stop()` in `Tools()`**

The `stillEnabled` guard in `Tools()` has a narrow but real race window:

1. `Tools()` acquires RLock, confirms the server is still in `t.enabled`, then releases the lock.
2. Before `Start()` is called on the next line, `handleDisable` acquires the write lock, removes the server from `t.enabled`, releases the lock, and calls `Stop()` on the same `*tools.StartableToolSet`.
3. `Tools()` then calls `Start()` on the already-stopped object.

The PR comments acknowledge this race ("a concurrent handleDisable could remove the server from t.enabled and call Stop(), but we'd still hold a reference and potentially restart it") but rely on the `stillEnabled` check to prevent it. That check does not eliminate the window — it only narrows it to after the check passes.

Whether this is harmful depends on `StartableToolSet`'s semantics: if `Start()` after `Stop()` reconnects (restarting a server the user just disabled) or panics, this is a real correctness bug. A user-visible symptom would be a server that was disabled continuing to surface its tools for one more `Tools()` cycle.

**Suggested fix**: Re-check `stillEnabled` immediately before calling `Start()` under a short-lived RLock, or have `handleDisable` set a "stopped" flag on the `StartableToolSet` that `Start()` respects.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,678 @@
...
+
+	msg := strings.Builder{}
+	fmt.Fprintf(&msg, "enabled %q (%s) — connection deferred until the host next enumerates tools.\n", id, server.Title)
+	fmt.Fprintf(&msg, "endpoint: %s\n", server.URL)
+	switch server.Auth.Type {
+	case "oauth":
+		msg.WriteString("auth: OAuth — an authorization URL will be elicited on the next turn.\n")
+	case "api_key":
+		if len(missing) > 0 {
+			fmt.Fprintf(&msg, "auth: API key — WARNING: the following env vars are NOT set: %s. Set them and re-enable, otherwise tool calls will fail.\n", strings.Join(missing, ", "))
```
docker-agent: **[MEDIUM] Misleading message: "Set them and re-enable" won't work without disabling first**

When `enable_remote_mcp_server` is called for an `api_key` server with missing environment variables, the server is added to `t.enabled` and then the user is told:

> "Set them and re-enable, otherwise tool calls will fail."

However, if the user sets the env vars and calls `enable_remote_mcp_server` again, they will hit the early-return guard at the top of `handleEnable`:

```go
if _, exists := t.enabled[id]; exists {
    t.mu.Unlock()
    return tools.ResultSuccess(fmt.Sprintf("server %q is already enabled", id)), nil
}
```

Because the server is already in `t.enabled`, the second call returns "already enabled" without creating a new toolset or re-expanding the headers. The env var snapshot from the first call is permanently baked into the existing toolset.

The user would need to call `disable_remote_mcp_server` first, then re-enable — but the message does not say this.

**Suggested fix**: Change the message to `"Set them, then disable and re-enable this server, otherwise tool calls will fail."` to give accurate instructions.

docker-agent (commented):

Two medium-severity findings in  — both in the newly added  toolset code. No issues found in the other changed files.

aheritier (approved):

PR #2796, opened 2026-05-14
Cross-harness orchestration: run Claude Code, Codex, OpenCode, Copilot, and OpenClaw as subagents

mcavage-docker:
## Summary

Adds cross-harness orchestration to docker-agent. Any agent in a team can now be backed by an external agent runtime (Claude Code, Codex, OpenCode, Copilot CLI, OpenClaw) instead of a model. The orchestrator delegates tasks to harness subagents via `transfer_task` and gets results back through docker-agent's normal event stream, TUI, and session model.

## What this does

Declare a harness-backed agent in your team YAML:

```yaml
agents:
  root:
    model: anthropic/claude-sonnet-4-5
    description: Orchestrator
    instruction: Route coding tasks to the appropriate specialist.
    sub_agents:
      - claude-coder
      - codex-coder

  claude-coder:
    description: Claude Code for complex refactors
    instruction: You are a senior software engineer.
    harness:
      type: claude-code

  codex-coder:
    description: Codex for code generation
    instruction: You are a software engineer.
    harness:
      type: codex
```

The orchestrator routes tasks to harness subagents. Each harness runs in its own subprocess, executes tools internally, and streams results back.

## Supported harnesses

| Type | Binary | Protocol |
|---|---|---|
| `claude-code` | `claude` | NDJSON stdout |
| `codex` | `codex` | JSONL stdout |
| `opencode` | `opencode` | NDJSON stdout |
| `copilot` | `copilot` | ACP (JSON-RPC stdio) |
| `openclaw` | `openclaw` | ACP (JSON-RPC stdio) |

## Key details

- **Aligned to [github.com/rumpl/harness](https://github.com/rumpl/harness)** -- uses Djordje's `Provider` interface as the adapter contract. PRs open upstream for `EventReasoning` and `--include-partial-messages` streaming.
- **Token streaming** -- Claude Code uses `--include-partial-messages` for live token streaming in the TUI. Text appears incrementally, not as a wall at the end.
- **Cost tracking** -- harness run cost attached to the assistant message and picked up by `TotalCost()`. Codex shows `--` (cost unknown) rather than `$0.00`.
- **Config version 10** -- adds `harness:` key to `AgentConfig`, mutually exclusive with `model:`. Version 9 configs upgrade automatically.
- **Security** -- subprocess env uses an allowlist (not full `os.Environ()`). ACP file operations sandboxed to session working directory. Harness command validated against injection characters. Subprocesses run in isolated process groups (`Setpgid: true`) to prevent terminal corruption.
- **Error messages** -- missing binary shows install hint. Bad type lists valid values. `model` + `harness` conflict caught at config parse.

## Testing

- 50+ unit tests across all adapters and the runtime integration
- UAT verified: file writes, tool calls, cost tracking, error cases, config validation
- Pre-existing test failures in `pkg/config TestCheckRequiredEnvVars` and `pkg/teamloader TestLoadExamples/dmr` are unrelated to this change

## Example

```
examples/harness-team.yaml
```

Closes #<issue> (if applicable)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25871619503).

pkg/harness/harness.go
```diff
@@ -0,0 +1,185 @@
...
+const (
+	ErrCodeContextExhausted   ErrorCode = "context_exhausted"
+	ErrCodeRateLimited        ErrorCode = "rate_limited"
+	ErrCodeAuthFailed         ErrorCode = "auth_failed"
+	ErrCodeHarnessCrashed     ErrorCode = "harness_crashed"
+	ErrCodeHarnessTimeout     ErrorCode = "harness_timeout"
+	ErrCodeUserCanceled       ErrorCode = "user_canceled"
+	ErrCodeCapabilityMismatch ErrorCode = "capability_mismatch"
+	ErrCodeUnknown            ErrorCode = "unknown"
+)
```
aheritier, line 59: **Blocking — half this PR doesn't compile.**

This file defines the canonical harness event vocabulary as exactly three types via the rumpl/harness alias: `EventText`, `EventResult`, `EventToolCall`. There is no `harness.EventSink`, no `RunStart`/`RunEnd`/`RunError`, no `TextDelta`/`ReasoningDelta`, no `ToolCallStart`/`ToolCallEnd`/`ToolCallResult`, no `harness.ACPAdapter`. And `RunResult` doesn't appear on the `RunACP` return path here either.

Yet `pkg/harness/acp/acp.go`, `pkg/harness/opencode/opencode.go`, `pkg/harness/copilot/copilot.go`, and `pkg/harness/openclaw/openclaw.go` all import these symbols and call `req.Events.Emit(...)`. That's why CI is red with ~20 `undefined:` errors across the ACP/opencode tree.

Either:
- finish the migration so all five adapters use the new 3-type vocabulary (drop the rich event types or move them upstream into rumpl/harness), or
- restore the rich `harness.*` types here so the ACP/opencode adapters resolve.

Per-PR-split note: this is exactly the kind of thing that becomes a non-issue if the foundation lands first and adapters come in one at a time.

pkg/harness/acp/acp.go
```diff
@@ -0,0 +1,401 @@
...
+	req.Events.Emit(harness.RunError{
+		RunID:   req.RunID,
+		Code:    harness.ErrCodeCapabilityMismatch,
+		Message: "ACP adapter requires ACPCallbacks; call RunACP instead of Run",
+		At:      time.Now(),
+	})
+}
+
+func (b *BaseAdapter) runACP(ctx context.Context, req harness.SubSessionRequest, callbacks harness.ACPCallbacks) error {
+	binary := b.BinaryName
```
aheritier, line 66: **Blocking — references undefined types.**

`req.Events`, `harness.RunError{...}`, `harness.RunStart{...}`, `harness.RunEnd{...}`, `harness.EventSink`, `harness.ACPAdapter` are all undefined against the `pkg/harness/harness.go` in this PR (which exposes only `EventText`/`EventResult`/`EventToolCall` from rumpl/harness, and a `Provider` interface without a `Capabilities()` method).

Also: `BaseAdapter.RunACP` here returns `void`, but `harness_delegation.go`'s `acpAdapter` type assertion expects `RunACP(...) harness.RunResult`, and so do `copilot.go` / `openclaw.go` (`return a.base.RunACP(ctx, req, callbacks)`). That mismatch alone breaks every consumer.

Pick one shape, propagate everywhere.

pkg/harness/acp/acp.go
```diff
@@ -0,0 +1,401 @@
...
+			}
+		}
+	}
+
+	return acpsdk.RequestPermissionResponse{
+		Outcome: acpsdk.RequestPermissionOutcome{
+			Cancelled: &acpsdk.RequestPermissionOutcomeCancelled{},
+		},
+	}, nil
+}
```
aheritier, line 347: **Suggestion — host-side terminal execution is silently disabled.**

Returning errors from `CreateTerminal` / `KillTerminal` / `TerminalOutput` / `ReleaseTerminal` / `WaitForTerminalExit` means any ACP-driven CLI that wants to run a shell command (which both Copilot CLI and OpenClaw do as their primary mechanism) will surface raw "terminal execution not supported in this host" errors back through the model. That's not a great first impression for the two adapters this base ships with.

If the v1 scope intentionally excludes terminals, the PR description / harness docs should call this out explicitly so users running a `copilot`-backed agent know to expect this. Better: thread the host's existing shell tool through `harness.ToolExecutor` and dispatch `terminal/*` to it the same way `fs/*` is sandboxed.

pkg/harness/acp/acp.go
```diff
@@ -0,0 +1,401 @@
...
+	// Emit RunEnd on success.
+	req.Events.Emit(harness.RunEnd{
+		RunID:        req.RunID,
+		HarnessRunID: string(sessResp.SessionId),
+		StopReason:   "success",
+		At:           time.Now(),
+	})
+
+	_ = cmd.Process.Kill()
+	return nil
```
aheritier, line 155: **Suggestion — `cmd.Process.Kill()` after a successful prompt is awkward.**

`runACP` ends a successful run by killing the subprocess, with no graceful shutdown attempt. For a CLI that may be flushing output, persisting state, or holding open a session for resume, SIGKILL is the wrong default. At minimum: close stdin first to signal EOF, give it a short grace window (≤ 1s), then escalate to `Kill` only if it's still running. Also worth thinking about: with this shape, you can never reuse the spawned process for a follow-up prompt — every turn pays the full subprocess startup cost. That undermines the "session resume" story the rest of the PR is built around.

pkg/harness/opencode/opencode.go
```diff
@@ -0,0 +1,523 @@
...
+	"os/exec"
+	"time"
+
+	"github.com/docker/docker-agent/pkg/harness"
+)
+
+const adapterName = "opencode"
+
+// Adapter implements harness.HarnessAdapter for the OpenCode CLI.
+type Adapter struct{}
```
aheritier, line 50: **Blocking — entire file is written against a non-existent API.**

This adapter calls `req.Events.Emit(harness.RunStart{...})`, `harness.TextDelta{...}`, `harness.ToolCallStart{...}`, `harness.UsageSummary{...}`, references `harness.EventSink`, `harness.AdapterCapabilities`, `harness.HostRequirements`, `harness.AdapterFeatures`, `harness.ProtocolStream`, etc. None of those types exist in `pkg/harness/harness.go` in this PR.

This file is from the pre-rebase event vocabulary (the one that was apparently dropped when the codebase was aligned to rumpl/harness's 3-type `Event`). It needs a full rewrite against the current vocabulary — same shape as `pkg/harness/claude/claude.go`, which already implements `Provider` correctly with `ParseStreamLine` returning `[]harness.Event`.

Until then opencode is fundamentally not part of the shippable surface and arguably shouldn't be in this PR at all. Pulling it out (as its own follow-up PR) would shrink this one substantially.

pkg/harness/copilot/copilot.go
```diff
@@ -0,0 +1,57 @@
...
+func (a *Adapter) InteractiveArgs(_ string) []string {
+	return []string{"copilot", "--acp", "--stdio"}
+}
+
+// ParseStreamLine implements [harness.Provider]. ACP adapters do not emit
+// NDJSON on stdout; events arrive via JSON-RPC and are surfaced by RunACP.
+func (a *Adapter) ParseStreamLine(_ string) []harness.Event { return nil }
+
+// RunACP is the real entry point for ACP-based execution.
+func (a *Adapter) RunACP(ctx context.Context, req harness.SubSessionRequest, callbacks harness.ACPCallbacks) harness.RunResult {
```
aheritier, line 53: **Blocking — return-type mismatch.**

`acp.BaseAdapter.RunACP` (this PR) doesn't return anything. This file's `RunACP` returns `harness.RunResult`. As written, this won't compile. Same issue in `pkg/harness/openclaw/openclaw.go`.

Whichever shape wins (probably `RunResult`-returning, to match `streamingAdapter.RunStreaming` and the `acpAdapter` type assertion in `pkg/runtime/harness_delegation.go`), `BaseAdapter.RunACP` needs to be updated and the result threaded back through `applyResult`.

pkg/config/latest/validate.go
```diff
@@ -53,6 +58,54 @@ func (t *Config) Validate() error {
...
+	validTypes := map[string]bool{
+		"claude-code": true,
+		"codex":       true,
+		"opencode":    true,
+		"copilot":     true,
+		"openclaw":    true,
+	}
+	if !validTypes[a.Harness.Type] {
+		return fmt.Errorf("agent %q: harness.type %q is not supported; valid values: claude-code, codex, opencode, copilot, openclaw", a.Name, a.Harness.Type)
+	}
```
aheritier, line 81: **Question — is the adapter list duplicated?**

The valid harness types are spelled out here. They're spelled again in `pkg/teamloader/harness.go`'s `defaultBinaryForType` and `installHintForType`. And again in the runtime delegation as the set of registered providers. Three sources of truth that have to stay in sync.

Consider moving validation through `harness.Lookup(t.Type)` (which already returns "no provider registered for type %q"), and exposing the install hint as a method on the provider. Drops the validate-side allowlist + keeps the canonical list at the registration site.

Not blocking — the current shape works — but with five adapters today and presumably more later, drift here is a question of when, not if.

.gm-agent-team/eng/cross-harness-orchestration/arch-spec-v2.md
```diff
@@ -0,0 +1,1030 @@
+# Architecture Spec v2: Cross-Harness Orchestration
```
aheritier, line 1: **Suggestion — these design docs probably don't belong in the repo.**

This PR adds ~7,000 lines under `.gm-agent-team/eng/cross-harness-orchestration/` (`prd.md`, `prd-v2.md`, `arch-spec.md`, `arch-spec-v2.md`, `impl-plan.md`, `impl-plan-v2.md`, `dx-review.md`, `dx-review-arch.md`, `arch-review.md`, `consistency-check.md`, `status.json`) plus a `.gm-agent-team/strategy/...` doc.

Two issues:

1. They're versioned (v1 *and* v2 of prd / arch-spec / impl-plan), suggesting these are work artifacts, not documentation users or future maintainers will reference.
2. They roughly double the diff size for reviewers. Even if the reviewer skips them, GitHub still pages through them.

Suggest: keep them in a separate planning repo / wiki / Notion, and land in `docs/` only the user-facing pieces (which already exist as `docs/configuration/agents/index.md` updates in this PR — those *are* keeper-quality). At minimum, drop them in a separate "design notes" PR so this PR is focused on shippable code.

pkg/runtime/harness_delegation.go
```diff
@@ -0,0 +1,584 @@
...
+}
+
+// --- translateSink ---
+
+// translateSink accumulates harness event state and translates the new
+// 3-type rumpl/harness event vocabulary (EventText / EventToolCall /
+// EventResult) into runtime events emitted to the underlying EventSink.
+//
+// State is written by the closure returned by translateFn (invoked from the
+// adapter goroutine) and by applyResult (invoked when the adapter returns).
```
aheritier, line 428: **Suggestion — host-side terminal execution shouldn't silently fail.**

`noopToolExecutor.Execute` returns an error for *every* method, including `terminal/*`. For Copilot CLI and OpenClaw, terminal execution is the primary mechanism the agent uses to actually do work. With this default, those harnesses look like they "run" but every shell tool call fails opaquely — and the failure shows up to the model, not to the user, which means the conversation continues but produces nothing useful.

Either:
- wire the host's existing shell/exec capability through here so terminals work end-to-end, or
- gate ACP harnesses (copilot/openclaw) behind a feature flag / opt-in until terminal support lands, with a startup message that explains the limitation.

Otherwise the v1 user experience for the two ACP harnesses ranges from "confusing" to "broken silently".

pkg/runtime/harness_delegation.go
```diff
@@ -0,0 +1,584 @@
...
+			Content:   content,
+			Cost:      sink.harnessRunCost,
+			CreatedAt: time.Now().Format(time.RFC3339),
+		})
+		sess.AddMessage(msg)
+		evts.Emit(MessageAdded(sess.ID, msg, a.Name()))
+	}
+
+	// Store resume token.
+	if sink.harnessRunID != "" {
```
aheritier, line 128: **Question — `runHarnessRoot` and `runHarnessForwarding` look like 80% the same code.**

Both:
1. Resolve `spec` from the agent.
2. Apply timeout.
3. Acquire/release the resume token.
4. Look up the adapter.
5. Build a `translateSink` and `runtimePermissionRequester`.
6. Run the adapter via `dispatchAdapter`.
7. Persist the final assistant message with cost.
8. Store the resume token.

The forwarding variant adds sub-session bookkeeping (newSubSession, AddSubSession, SubagentStop hooks, span, error propagation) but the core 1–8 dance is duplicated. With a third "collecting" variant for background agents, that's three near-copies.

Worth extracting `runHarnessCore(...)` that takes the sink + an "after-run" callback, so the three callers differ only in what they do after the adapter returns. Not a merge blocker, just feels like the kind of duplication that will drift the moment a new event needs to be plumbed.

pkg/teamloader/harness.go
```diff
@@ -0,0 +1,86 @@
...
+	case "opencode":
+		return "opencode"
+	case "copilot":
+		return "copilot"
+	case "openclaw":
+		return "openclaw"
+	default:
+		return harnessType
+	}
+}
```
aheritier, line 28: **Suggestion — duplication of the harness type list.**

This `switch` lives next to `installHintForType` (same `switch`, different return values) and `validateHarness` in `pkg/config/latest/validate.go` has yet another copy. Whenever a sixth harness shows up, three places need updating in lockstep.

Consider a single `type harnessKind struct { binary, installHint string }` map, or — better — make the install hint and default binary methods on the registered `Provider` so the registry is the only source of truth.

pkg/teamloader/harness.go
```diff
@@ -0,0 +1,86 @@
...
+		policy = &agent.PermissionPolicy{
+			Mode: agent.PermissionMode(cfg.PermissionPolicy.Mode),
+		}
+	}
+
+	timeout := cfg.Timeout.Duration
+	if timeout == 0 {
+		timeout = 10 * time.Minute
+	}
+
```
aheritier, line 75: **Question — why does this take a `*latest.HarnessConfig` (rather than a value), and why is the timeout default not in the schema layer?**

Two small things:

1. `*latest.HarnessConfig`: the function never mutates it, and callers handle nil-ness one level up (`if a.Harness == nil`). Take `latest.HarnessConfig` by value and the call sites get cleaner.
2. The 10-minute default timeout for harness runs is enforced here (in the loader) instead of in `pkg/config/latest`. A user inspecting a parsed `Config` won't see the default. Apply it during config defaulting so the same value is visible to every consumer (TUI, API, validators, debug dumps).

Both are ergonomics, not blockers.

pkg/harness/harness.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	registry[p.Name()] = p
+}
+
+// Lookup returns the provider registered for the given name.
+func Lookup(name string) (Provider, error) {
+	regMu.RLock()
+	defer regMu.RUnlock()
+	p, ok := registry[name]
+	if !ok {
```
aheritier, line 155: **Question — why a process-global mutex, and what's the eviction story?**

`tokenInUse` is a global `map[string]bool` guarded by `tokenMu`. Two questions:

1. **Globalness.** This couples the harness layer to a single-process runtime. If a future user runs two `LocalRuntime` instances in the same process (mocking, multi-tenant API server), they'll silently share token state. Consider attaching the registry to a `Manager` or letting the runtime own it.
2. **Eviction on crash / leak.** If the runtime panics between `AcquireToken` and `defer ReleaseToken` (e.g. a panic before `defer` registers), the token stays held until the process exits. Resume of that session then fails with `"token is already in use"`. Worth documenting that this is process-lifetime state and that the failure mode is "restart docker-agent" — or capturing it via `runtime.SetFinalizer` on the request struct.

Not a merge blocker; would be nice to know the intent.

pkg/harness/acp/acp.go
```diff
@@ -0,0 +1,401 @@
...
+// buildEnv constructs the environment for the ACP subprocess.
+// Only safeEnvKeys are inherited; additional vars come from SubSessionRequest.Env.
+func buildEnv(req harness.SubSessionRequest) []string {
+	safe := make(map[string]bool, len(safeEnvKeys))
+	for _, k := range safeEnvKeys {
+		safe[k] = true
+	}
+
+	var env []string
+	for _, kv := range os.Environ() {
```
aheritier, line 183: **Praise** — the env construction here is exactly right.

Allow-list of `safeEnvKeys` (system + AI provider tokens + node) instead of a blanket `os.Environ()` pass-through. Means a stray `AWS_SESSION_TOKEN` or `KUBE_CONFIG` doesn't accidentally leak into a Copilot/OpenClaw subprocess. Combined with the path containment in `sandbox.Resolve`, this is the right shape for "give the harness what it needs and nothing else." Worth keeping when you fix the compilation issues above.

pkg/harness/acp/acp.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("acp stdout pipe: %w", err)
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("acp stderr pipe: %w", err)
+	}
+
```
aheritier, line 85: **Question — `Setpgid: true` is platform-specific.**

The PR description mentions process-group isolation (`Setpgid: true`) for TUI safety, but `os/exec.Cmd.SysProcAttr` is a `*syscall.SysProcAttr` with platform-specific fields. `Setpgid` exists on Unix but not on Windows.

Where is this set, and is there a `_windows.go` companion to keep the build green there? `task cross` should fail Windows builds today if it isn't covered. Worth a `go:build` matrix or platform-split file with a no-op equivalent.

aheritier (commented):
interesting idea Marc

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26000733117).

dgageot:

PR #2805, opened 2026-05-15
ci: skip image push in forked repositories

areebahmeddd:
got notification about failing build job in my forked repo

image

image

this should be skipped while other jobs run as normal

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
           cache-to: type=gha,mode=min,scope=buildx
 
   build-and-push-image:
-    if: github.event_name != 'pull_request'
+    if: github.event_name != 'pull_request' && github.repository == 'docker/docker-agent'
     needs: [lint, build-and-test, license-check]
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
           cache-to: type=gha,mode=min,scope=buildx
 
   build-and-push-image:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -111,7 +111,7 @@ jobs:
           cache-to: type=gha,mode=min,scope=buildx
 
   build-and-push-image:
-    if: github.event_name != 'pull_request'
+    if: github.event_name != 'pull_request' && github.repository == 'docker/docker-agent'
```
aheritier: **Non-blocking (consider):** Hardcoding the repo slug `docker/docker-agent` works, but couples this guard to the repository name — a future rename or transfer would silently disable image pushes from `main`/tags.

A more intent-revealing alternative is to gate on whether the workflow is running in a fork:

```yaml
if: github.event_name != 'pull_request' && !github.event.repository.fork
```

`github.event.repository.fork` is `false` for the upstream and `true` for any fork, which is exactly the property we care about ("don't try to push from forks"). It's also rename/transfer-proof and matches the language of the PR title ("skip image push in forked repositories").

Either approach is fine — happy to merge as-is if you prefer the explicit allow-list form, but worth a thought.

areebahmeddd: Hm. didn’t know github events had a property like that. your suggestion is much cleaner 👍🏻

areebahmeddd:
gentle ping! 🙂 @aheritier @dgageot

dgageot:
@areebahmeddd Could you rebase, please?

areebahmeddd:
done! @dgageot

dgageot (approved):

PR #2806, opened 2026-05-16
fix(examples): use xai/grok-2-latest in grok.yaml

dgageot:
The example previously referenced `xai/grok-3-mini`, which is no longer present in the models.dev catalog (xAI now lists `grok-2*`, `grok-beta`, `grok-vision-beta`, `grok-4.*`, and `grok-imagine-*`). This caused `TestParseExamples` in `pkg/config` to fail.

Switch the example to `xai/grok-2-latest`, which is in the catalog.

### DIFF
--- a/examples/grok.yaml
+++ b/examples/grok.yaml
@@ -2,7 +2,7 @@
 
 agents:
   root:
-    model: xai/grok-3-mini
+    model: xai/grok-2-latest
     description: A helpful AI assistant powered by Grok
     instruction: |
       You are Grok, a helpful and maximally truthful AI built by xAI.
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change is a minimal, correct model name substitution in `examples/grok.yaml`. Both the `agents.root.model` field and the `models.grok-model.model` field are updated in lockstep from `xai/grok-3-mini` / `grok-3-mini` to `xai/grok-2-latest` / `grok-2-latest`. No schema violations, no logic errors, no inconsistencies detected. The fix correctly resolves the `TestParseExamples` failure caused by the deprecated model reference.

aheritier (approved):

PR #2807, opened 2026-05-17
Better tool registry

dgageot:


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/25983879160).

pkg/tools/builtin/lsp/lsp.go
```diff
@@ -339,13 +343,66 @@ type lspInlayHint struct {
...
+	info, err := os.Stat(dir)
+	if err != nil {
+		return fmt.Errorf("%s working_dir %q does not exist or is not accessible: %w", kind, dir, err)
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("%s working_dir %q is not a directory", kind, dir)
+	}
+	return nil
+}
+
```
aheritier: **Blocking:** this duplicates `workingdir.Resolve` / `workingdir.CheckDirExists` (just introduced in this PR for MCP) but with different behavior:

- No `~` / `${VAR}` expansion — set `working_dir: ~/projects` on an LSP toolset and you'll spawn the server in a literal `~/projects` directory, while the same value on an MCP toolset would expand. (Tests in `pkg/teamloader/registry_test.go` already cover the MCP side.)
- Different error wording (`"%s working_dir %q does not exist or is not accessible"` vs the workingdir package's `"working_dir %q for %s toolset does not exist"`).

Please drop these two helpers and call `workingdir.Resolve(toolset.WorkingDir, runConfig.WorkingDir)` + `workingdir.CheckDirExists(cwd, "lsp")` directly, mirroring what `mcp.CreateToolSet` does.

pkg/teamloader/registry.go
```diff
@@ -141,394 +135,5 @@ func checkDirExists(dir, toolsetType string) error {
 // a trusted, operator-authored value where cross-tree references (e.g. a sibling
 // module root in a monorepo) are intentional and must not be silently blocked.
 func resolveToolsetWorkingDir(toolsetWorkingDir, agentWorkingDir string) string {
```
aheritier: **Blocking:** this wrapper has no production callers — it's only referenced by `TestResolveToolsetWorkingDir` / `TestResolveToolsetWorkingDir_EnvVarExpansion` in the same package. Both tests duplicate the cases that already live in `pkg/tools/workingdir/workingdir_test.go`.

Suggest deleting `resolveToolsetWorkingDir` here together with both registry-level tests; keep the workingdir package as the single source of truth.

pkg/tools/builtin/memory/memory.go
```diff
@@ -27,35 +34,79 @@ type DB interface {
...
+	var basePath string
+	if filepath.IsAbs(toolsetPath) {
+		basePath = ""
+	} else if wd := runConfig.WorkingDir; wd != "" {
+		basePath = wd
+	} else {
+		basePath = parentDir
+	}
+
+	return path.ValidatePathInDirectory(toolsetPath, basePath)
```
aheritier: **Blocking:** identical body to `tasks.resolveToolsetPath` (the diff is empty). Two builtins now own the same private helper for the same problem (default the toolset path against `runConfig.WorkingDir` then `parentDir`, expanding `~`).

Two reasonable options:
- Hoist into a small package alongside `workingdir`, e.g. `pkg/tools/toolsetpath` with `Resolve(toolsetPath, parentDir, runConfig)`. Mirrors the working-dir refactor in this same PR.
- Or expose `tasks.resolveToolsetPath` (or a new helper in `pkg/path`) and have memory call it.

Either way, please don't ship two copies — the next time the resolution rules change you'll fix one and forget the other.

pkg/teamloader/registry.go
```diff
@@ -37,35 +24,81 @@ import (
...
+				return builtinrag.CreateToolSet(ctx, toolset, parentDir, runConfig)
+			},
+		},
+	}
+}
+
+// toolsetRegistry manages the registration of toolset creators by type.
+type toolsetRegistry struct {
 	creators map[string]ToolsetCreator
 }
```
aheritier, line 99: **Non-blocking:** the `ToolsetCreator` type is now a function with a fixed 5-argument shape, but more than half the registered creators only need 1–2 of those (e.g. `think`, `user_prompt`, `background_agents`, `todo`, `model_picker` ignore `ctx`, `parentDir`, `configName`, and `runConfig`). The wrapper closures in this map exist purely to discard the extras.

If you want to push the cleanup further, an interface — `type ToolsetCreator interface { Create(ctx, ToolsetCreationRequest) (tools.ToolSet, error) }` with a struct field for each input — lets each creator pick what it needs by destructuring the request, and lets you add new fields in the future without touching every entry. Not a blocker for this PR; flagging since the registry is the headline change.

pkg/tools/builtin/shell/shell.go
```diff
@@ -450,8 +452,19 @@ func reapSpawnedChild(cmd *exec.Cmd, pg *processGroup) {
...
+// CreateToolSet is used by the tools registry.
+func CreateToolSet(ctx context.Context, toolset latest.Toolset, runConfig *config.RuntimeConfig) (tools.ToolSet, error) {
+	env, err := environment.ExpandAll(ctx, environment.ToValues(toolset.Env), runConfig.EnvProvider())
+	if err != nil {
+		return nil, fmt.Errorf("failed to expand the tool's environment variables: %w", err)
+	}
+	env = append(env, os.Environ()...)
+
+	return New(env, runConfig), nil
+}
```
aheritier, line 468: **Question:** `os.Environ()` is appended into `env` here (and the same pattern repeats in `mcp.CreateToolSet`, `lsp.CreateToolSet`, `shell.CreateScriptToolSet`). Is that intentional now that creator signatures explicitly take an `EnvProvider` from `runConfig`? Walking 4× over `os.Environ()` per agent load isn't expensive, but it does mean tests that override `EnvProviderForTests` to control what reaches a subprocess can still leak the host env into spawned processes via this path.

If the goal is "expand `${...}` references against EnvProvider, but let the spawned process inherit the host env", that's fine — could you add a one-line comment explaining why we re-append `os.Environ()` after expansion? It surprised me on read.

pkg/tools/builtin/rag/rag.go
```diff
@@ -9,50 +9,76 @@ import (
...
+	"github.com/docker/docker-agent/pkg/config"
+	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/rag"
 	ragtypes "github.com/docker/docker-agent/pkg/rag/types"
 	"github.com/docker/docker-agent/pkg/tools"
 )
 
+// CreateToolSet is used by the tools registry.
+func CreateToolSet(ctx context.Context, toolset latest.Toolset, parentDir string, runConfig *config.RuntimeConfig) (tools.ToolSet, error) {
+	if toolset.RAGConfig == nil {
```
aheritier, line 21: **Nit:** the validation message says "should have been resolved from ref", which leaks an internal expectation to the YAML author. They might be hand-writing a config and just forgot the `rag_config:` block. Suggest: `"rag toolset requires either a rag_config block or a ref"`.

pkg/teamloader/registry.go
aheritier, line 99: **Nit:** about half of the new commits in this branch (e.g. `refactor: move shell toolset creation into package`, `refactor: move memory toolset creation into package`, `refactor: standardize builtin toolset naming`, `refactor: narrow toolset creator signatures`) are missing the `Assisted-By: docker-agent` trailer that the project conventions ask for on AI-authored commits — and missing `Signed-off-by:` too on the same set. Only ~3 commits in the branch carry either.

If they were authored by hand, ignore. Otherwise worth a one-shot `but rebase --signoff` (or amending in-place) before merge.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26023442456).

rumpl (approved):

pkg/teamloader/registry_test.go
```diff
@@ -73,90 +73,7 @@ func TestCreateMCPTool_BareCommandNotFound_CreatesToolsetAnyway(t *testing.T) {
...
-// because t.Setenv is incompatible with t.Parallel on the parent test.
-func TestResolveToolsetWorkingDir_EnvVarExpansion(t *testing.T) {
-	t.Setenv("TEST_REGISTRY_CWD_VAR", "/custom/path")
-
-	got := resolveToolsetWorkingDir("${TEST_REGISTRY_CWD_VAR}/app", "/workspace")
-	assert.Equal(t, "/custom/path/app", got)
-}
-
-// TestCreateMCPTool_WorkingDir_ReachesSubprocess verifies that working_dir is
+} // TestCreateMCPTool_WorkingDir_ReachesSubprocess verifies that working_dir is
```
aheritier, line 76: **Nit** — Stray missing newline: the closing `}` of `TestCreateMCPTool_GatewayRefMissingBinaryStillBuilds` is on the same line as the next function's leading comment (`} // TestCreateMCPTool_WorkingDir_…`). `gofmt` accepts it, but it's the only place in the file with that style.

pkg/tools/workingdir/workingdir.go
```diff
@@ -0,0 +1,56 @@
...
+		if err == nil {
+			return abs
+		}
+		return filepath.Join(agentWorkingDir, toolsetWorkingDir)
+	}
+	return toolsetWorkingDir
+}
+
+// Default returns the configured agent working directory or the process cwd.
+func Default(agentWorkingDir string) string {
```
aheritier, line 31: **Nit** — `Default` is exported but has zero callers. Meanwhile `pkg/tools/builtin/filesystem/filesystem.go` (~L130–137) still inlines the same `os.Getwd()` fallback. Either route filesystem through `workingdir.Default` or drop the exported helper.

pkg/teamloader/registry.go
```diff
@@ -43,29 +30,74 @@ import (
 // configName identifies the agent config file (e.g. "memory_agent" from "memory_agent.yaml").
 type ToolsetCreator func(ctx context.Context, toolset latest.Toolset, parentDir string, runConfig *config.RuntimeConfig, configName string) (tools.ToolSet, error)
```
aheritier, line 31: **Question** — `ToolsetCreator` is exported, but the new `ToolsetRegistry` interface only exposes `CreateTool` — there's no public `Register` method anymore. External implementations (e.g. ACP) wrap `NewDefaultToolsetRegistry()` and forward; they never construct a `ToolsetCreator`. Consider unexporting it if registration is closed.

pkg/tools/builtin/memory/memory.go
```diff
@@ -27,35 +34,64 @@ type DB interface {
...
+
+	db, err := sqlite.NewMemoryDatabase(validatedMemoryPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create memory database: %w", err)
+	}
+
+	return NewWithPath(db, validatedMemoryPath), nil
+}
+
+func New(manager DB) *ToolSet {
```
aheritier, line 78: **Nit** — After the rename, `New(manager)` (no path) is only used in tests; production code goes through `NewWithPath`. Worth folding the two into one constructor with an optional path, or marking `New` test-only / unexporting it.

pkg/tools/builtin/shell/shell.go
```diff
@@ -450,8 +452,23 @@ func reapSpawnedChild(cmd *exec.Cmd, pg *processGroup) {
...
+func CreateToolSet(ctx context.Context, toolset latest.Toolset, runConfig *config.RuntimeConfig) (tools.ToolSet, error) {
+	env, err := environment.ExpandAll(ctx, environment.ToValues(toolset.Env), runConfig.EnvProvider())
+	if err != nil {
+		return nil, fmt.Errorf("failed to expand the tool's environment variables: %w", err)
+	}
+	// Re-append os.Environ() after expansion so spawned processes inherit the
+	// host environment. EnvProvider is used only to expand ${...} references
+	// in toolset.Env; the subprocess still needs access to the full environment.
+
+	env = append(env, os.Environ()...)
```
aheritier, line 465: **Nit** — The new comment explaining the `os.Environ()` re-append is helpful — but the same logic is repeated verbatim (without the comment) in `script_shell.go`, `lsp.go`, `mcp.go`. Consider hoisting to a small `expandedEnv(ctx, toolset.Env, runConfig)` helper alongside `workingdir.Resolve` — same shape every time.

aheritier (approved):
Large but largely mechanical refactor — and unusually reviewable thanks to the commit-per-toolset cadence. Each step compiles and is well-scoped. CI is green; local build/test/vet/gofmt all pass. Approving.

Highlights:
- `pkg/tools/workingdir` + `pkg/tools/toolsetpath` are exactly the right size — small, focused, with their own tests, and `Resolve` semantics documented up-front.
- The end-to-end `working_dir` regression tests (`TestCreateMCPTool_WorkingDir_ReachesSubprocess`, `TestCreateLSPTool_WorkingDir_ReachesSubprocess`) are the kind that would have caught the original bug they defend against.
- Moving `lifecyclePolicyFromConfig` into `pkg/tools/lifecycle` removes a layering inversion — clean win.

Findings below are non-blocking polish.

PR #2808, opened 2026-05-17
Simpler code to cache value

dgageot:


pkg/userid/userid.go
```diff
@@ -23,26 +23,20 @@ import (
...
 // process so subsequent calls do not touch the filesystem.
-func Get() string {
-	mu.Lock()
-	defer mu.Unlock()
+var Get = sync.OnceValue(get)
 
-	if cached != "" {
-		return cached
-	}
+func ResetForTests() {
```
docker-agent, line 35: **[MEDIUM] Data race: unsynchronised write to `Get` in `ResetForTests()` vs concurrent reads**

`ResetForTests()` assigns a fresh `sync.OnceValue` to the package-level variable `Get` without any synchronisation. Meanwhile, parallel tests (e.g. `TestHeaders` uses `t.Parallel()`) or the HTTP client may be reading `Get` to call it. Writing to a shared variable from one goroutine while another goroutine reads it is a data race under the Go memory model — `go test -race` will flag it.

The old implementation avoided this by holding `mu` inside `ResetForTests()` before clearing `cached`. The new implementation has no equivalent protection on the variable itself.

**Suggested fix:** protect the variable with a `sync.RWMutex`, use `sync/atomic` (`atomic.Value` or a pointer), or — simplest for a test-only reset — guard the write and read with a mutex:
```go
var getMu sync.RWMutex
var getOnce = sync.OnceValue(get)

func Get() string {
    getMu.RLock()
    f := getOnce
    getMu.RUnlock()
    return f()
}

func ResetForTests() {
    getMu.Lock()
    getOnce = sync.OnceValue(get)
    getMu.Unlock()
}
```

docker-agent (commented):

One confirmed medium-severity data race introduced by the refactor.

PR #2810, opened 2026-05-18
Improve concurrent package

dgageot:
Improve concurrent package and use it in more places

### DIFF
--- a/cmd/root/run_event_hooks.go
+++ b/cmd/root/run_event_hooks.go
@@ -12,6 +12,7 @@ import (
 	tea "charm.land/bubbletea/v2"
 
 	"github.com/docker/docker-agent/pkg/app"
+	"github.com/docker/docker-agent/pkg/concurrent"
 	"github.com/docker/docker-agent/pkg/runtime"
 	"github.com/docker/docker-agent/pkg/shellpath"
 )
@@ -80,7 +81,7 @@ func runEventHook(command string, payload []byte) {
 	// it, and exits; the spawning goroutine ends with the subprocess.
 	cmd := exec.CommandContext(context.Background(), shell, append(argsPrefix, command)...)
 	cmd.Stdin = bytes.NewReader(payload)
-	var out boundedBuffer
+	var out boundedWriter
 	cmd.Stdout = &out
 	cmd.Stderr = &out
 	if err := cmd.Run(); err != nil {
@@ -88,25 +89,31 @@ func runEventHook(command string, payload []byte) {
 	}
 }
 
-// boundedBuffer captures up to maxHookOutput bytes from a hook subprocess
-// and silently discards the rest. It implements only io.Writer so it can be
-// assigned to exec.Cmd's Stdout/Stderr without forcing exec to buffer the
-// full output internally.
-type boundedBuffer struct {
-	buf bytes.Buffer
+// boundedWriter captures up to maxHookOutput bytes from a hook subprocess
+// and silently discards the rest. It satisfies io.Writer so it can be
+// assigned to exec.Cmd's Stdout/Stderr.
+//
+// exec.Cmd spawns separate copy goroutines for Stdout and Stderr, so the
+// underlying buffer must be safe for concurrent writes; that's what
+// [concurrent.Buffer] gives us. The cap is enforced softly: between
+// Len() and Write() another goroutine may slip in a chunk, so we may
+// over-shoot maxHookOutput by at most one Write per concurrent stream
+// (a few KB) — acceptable for diagnostic output.
+type boundedWriter struct {
+	buf concurrent.Buffer
 }
 
-func (b *boundedBuffer) Write(p []byte) (int, error) {
+func (b *boundedWriter) Write(p []byte) (int, error) {
 	if remaining := maxHookOutput - b.buf.Len(); remaining > 0 {
-		if len(p) > remaining {
-			b.buf.Write(p[:remaining])
-		} else {
-			b.buf.Write(p)
+		chunk := p
+		if len(chunk) > remaining {
+			chunk = chunk[:remaining]
 		}
+		_, _ = b.buf.Write(chunk)
 	}
 	return len(p), nil
 }
 
-func (b *boundedBuffer) String() string {
+func (b *boundedWriter) String() string {
 	return b.buf.String()
 }
--- a/pkg/concurrent/slice.go
+++ b/pkg/concurrent/slice.go
@@ -11,7 +11,9 @@ type Slice[V any] struct {
 }
 
 func NewSlice[V any]() *Slice[V] {
-	return &Slice[V]{}
+	return &Slice[V]{
+		values: []V{},
+	}
 }
 
 func (s *Slice[V]) Append(value V) {
@@ -57,6 +59,12 @@ func (s *Slice[V]) All() []V {
 	return slices.Clone(s.values)
 }
 
+// Range calls f for every element in the slice. Iteration stops early if f
+// returns false.
+//
+// f is invoked while a read lock is held on the slice. Callbacks must not
+// call methods that acquire the write lock (Append, Set, Update, Clear) on
+// the same Slice, or a deadlock will occur.
 func (s *Slice[V]) Range(f func(index int, value V) bool) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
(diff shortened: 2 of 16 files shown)


pkg/tools/builtin/shell/shell.go
```diff
@@ -324,21 +333,21 @@ func (h *shellHandler) RunShellBackground(_ context.Context, params RunShellBack
...
 	if job.status.Load() == statusStopped {
 		return
 	}
 
+	// Publish exitCode and err *before* the terminal status.Store so that
+	// readers observing a non-running status are guaranteed to see the
+	// final values via Go's atomic acquire/release ordering.
 	if err != nil {
 		if exitErr, ok := errors.AsType[*exec.ExitError](err); ok {
 			job.exitCode = exitErr.ExitCode()
```
docker-agent, line 336: **[HIGH] Data race: plain `int`/`error` fields not actually protected by atomic on a different variable**

`monitorJob` writes the plain (non-atomic) fields `job.exitCode` and `job.err` and then performs an `atomic.Int32.Store` on `job.status`. Readers in `ViewBackgroundJob` and `ListBackgroundJobs` perform an `atomic.Int32.Load` on `job.status` and then read `job.exitCode`/`job.err`.

The comment claims this is safe via "Go's atomic acquire/release ordering" — but **Go's memory model does not support this pattern**. Atomic operations in Go only synchronise accesses to the *same* atomic variable (see https://go.dev/ref/mem#atomic). Writing a non-atomic variable before an atomic store, and reading it after an atomic load on a *different* variable, does **not** create a happens-before relationship. This is different from C++11 acquire-release semantics. The Go race detector (`-race`) would likely flag reads of `exitCode` and `err` in the reader goroutines as racing with writes in `monitorJob`.

The previous code correctly used `outputMu sync.RWMutex` to protect both the buffer reads/writes *and* the `exitCode`/`err` fields. That mutex has been removed and replaced with this invalid synchronisation pattern.

**Fix**: Either restore a mutex around `exitCode`/`err`, or change them to `atomic.Int32`/`atomic.Value` respectively, or use a channel to communicate the exit result.

pkg/userid/userid.go
```diff
@@ -23,26 +23,20 @@ import (
...
-func Get() string {
-	mu.Lock()
-	defer mu.Unlock()
+var Get = sync.OnceValue(get)
 
-	if cached != "" {
-		return cached
-	}
+func ResetForTests() {
+	Get = sync.OnceValue(get)
```
docker-agent: **[MEDIUM] `ResetForTests()` writes package-level `Get` without synchronisation — data race**

`Get` is now a package-level variable (a `func() string`) rather than a plain function. `ResetForTests()` reassigns it without holding any lock:

```go
var Get = sync.OnceValue(get)

func ResetForTests() {
    Get = sync.OnceValue(get)  // unsynchronised write
}
```

Any goroutine that calls `userid.Get()` concurrently with `ResetForTests()` performs a concurrent read and write of the same package-level variable — a data race under Go's memory model. Running tests with `-race` would flag this.

Additionally, callers that captured a reference to `Get` before `ResetForTests()` was called (e.g. `f := userid.Get`) will still hold the old, already-executed `OnceValue` closure, so the reset has no effect for them.

**Fix**: Guard `Get` with a `sync.Mutex` or `sync.RWMutex`, or use `atomic.Pointer[func() string]`, so both reads and writes are synchronised. For example:
```go
var (
    getMu sync.Mutex
    getFn = sync.OnceValue(get)
)

func Get() string {
    getMu.Lock()
    f := getFn
    getMu.Unlock()
    return f()
}

func ResetForTests() {
    getMu.Lock()
    getFn = sync.OnceValue(get)
    getMu.Unlock()
}
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR improves the `concurrent` package and migrates callers to use it. The primitives themselves (Buffer, Map, Slice) look correct. However, two issues in the callers introduce real data races.

pkg/snapshot/snapshot.go
```diff
@@ -96,14 +96,8 @@ func (m *Manager) Cleanup(ctx context.Context, dir string) error {
...
 func (m *Manager) lock(key string) *sync.Mutex {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-	if l := m.locks[key]; l != nil {
-		return l
-	}
-	l := &sync.Mutex{}
-	m.locks[key] = l
-	return l
+	lock, _ := m.locks.LoadOrStore(key, &sync.Mutex{})
```
aheritier, line 99: Non-blocking, optional polish: this allocates a fresh `*sync.Mutex` on every call, even on the hot path where the lock for `key` already exists. The previous double-checked-locking version avoided the allocation. A `Load` fast-path keeps the new API but skips the alloc when the entry is cached:

```go
func (m *Manager) lock(key string) *sync.Mutex {
    if l, ok := m.locks.Load(key); ok {
        return l
    }
    l, _ := m.locks.LoadOrStore(key, &sync.Mutex{})
    return l
}
```

Cost is one mutex per `Open` so it's truly tiny — flag only because the cleanup pass is *almost* free here.

cmd/root/run_event_hooks.go
```diff
@@ -80,33 +81,39 @@ func runEventHook(command string, payload []byte) {
...
+// boundedWriter captures up to maxHookOutput bytes from a hook subprocess
+// and silently discards the rest. It satisfies io.Writer so it can be
+// assigned to exec.Cmd's Stdout/Stderr.
+//
+// exec.Cmd spawns separate copy goroutines for Stdout and Stderr, so the
+// underlying buffer must be safe for concurrent writes; that's what
+// [concurrent.Buffer] gives us. The cap is enforced softly: between
+// Len() and Write() another goroutine may slip in a chunk, so we may
+// over-shoot maxHookOutput by at most one Write per concurrent stream
+// (a few KB) — acceptable for diagnostic output.
```
aheritier, line 101: Nit on the bound stated here: "at most one Write per concurrent stream (a few KB)" undersells it. With two goroutines reading the cap close to its limit, both can take the `Len()` snapshot, both compute a positive `remaining`, and both append a *full* slice (`chunk = p` if `len(p) <= remaining`). `exec.Cmd`'s Stdout/Stderr copy goroutines hand us pipe-sized buffers (~32 KiB each on Unix), so the worst-case overshoot is closer to `maxHookOutput + N * 32KiB` — i.e. ~70 KiB for stdout+stderr, not "a few KB".

Still acceptable for diagnostic output. Either tighten the bound to "one full Write per concurrent stream (~32 KiB each)", or — if you want a hard cap — clamp `chunk` to `remaining` unconditionally and drop the wording about overshoot.

pkg/concurrent/slice.go
```diff
@@ -11,7 +11,9 @@ type Slice[V any] struct {
 }
 
 func NewSlice[V any]() *Slice[V] {
-	return &Slice[V]{}
+	return &Slice[V]{
+		values: []V{},
+	}
 }
```
aheritier, line 17: Behaviour change worth flagging in the PR body / changelog: previously `NewSlice[V]()` left `values` nil, so a fresh slice's `All()` returned `nil` (`slices.Clone(nil) == nil`). After this change, it returns a non-nil empty slice. The new `TestSlice_AllEmpty` codifies the contract, but any caller that JSON-marshals `s.All()` flips from `null` to `[]`.

I grep'd current call sites and only `MemoryTodoStorage.All` is affected, and its consumer (`listTodos`) already nil-guards before encoding, so there's no observable break today. Just worth a one-liner in the PR description so future API consumers don't get surprised.

aheritier (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed 1082 lines across 16 files. The concurrent package improvements look well-implemented:

- **`concurrent.Map`**: `LoadOrStore` correctly uses double-checked locking (read lock → re-check under write lock). `Range` clones the map under a read lock before iterating, so callbacks can safely mutate the map without deadlocking. `Clear` and `Delete` hold the write lock as expected.
- **`concurrent.Slice`**: `All()` returns a `slices.Clone` for snapshot isolation. `Range` and `Find` document the constraint that callbacks must not acquire the write lock.
- **`concurrent.Buffer`**: All operations (`Write`, `String`, `Bytes`, `Drain`) correctly hold the mutex. `Bytes()` returns a `slices.Clone` to prevent external mutation of the internal buffer.
- **Callers** (`fast_renderer.go`, `editfile/render.go`, `styles/composite.go`, `hooks/handler.go`, `hooks/model_handler.go`, `snapshot/snapshot.go`, `evaluation/progress.go`, `chatserver/conversation_lock.go`): All refactored usages correctly replace manual mutex+map patterns with the new primitives. No races or logic errors introduced.

PR #2813, opened 2026-05-18
fix(openai): explicitly send tool_choice=auto when tools are provided

### LINKED ISSUE #2804: tool_choice field not sent in API requests, causing 400 errors with strict OpenAI-compatible gateways
## Summary
`docker-agent` does not include a `tool_choice` field in its Chat Completions API requests, even when `tools` is present. The OpenAI spec treats omission of `tool_choice` as implying `"auto"`, so this is technically valid — but strict OpenAI-compatible gateway implementations (notably LiteLLM-based proxies) require the field to be explicitly present and reject requests with a `400 Bad Request` when it is absent.
## Error
HTTP 400: litellm.BadRequestError: BedrockException -
{"message":"The model returned the following errors: tool_choice.type: Field required"}
This error occurs on every request where tools are present, exhausting all fallback models since the validation fires before the model is reached.
## Root Cause
In `pkg/model/provider/openai/client.go`, `ChatCompletionNewParams` is built without setting `ToolChoice`. When the request reaches a LiteLLM gateway that enforces strict Pydantic validation, the missing field causes a `400` rejection.
Per the OpenAI spec, omitting `tool_choice` is equivalent to `"auto"` — but LiteLLM in certain configurations requires the object form explicitly:
\```json
"tool_choice": { "type": "auto" }
\```
## Proposed Fix
Explicitly send `"tool_choice": "auto"` whenever `tools` is non-empty in the request. This is:
- **Spec-compliant** — it matches the OpenAI default behavior exactly
- **Backwards-compatible** — no behavior change for gateways that already handle omission correctly
- **Safe for the agentic loop** — `"auto"` preserves the model's autonomy to decide when to call tools
Optionally, expose a `tool_choice:` field in the model config schema (alongside the existing `parallel_tool_calls`) for users who need to override the value.
## Workarounds
- **Gateway side:** Add `drop_params: true` to LiteLLM's `general_settings` to silently accept requests missing optional fields
- **Client side:** None currently available in `agent.yaml` — `provider_opts` does not forward `tool_choice`
## Environment
- docker-agent version: v1.59.0
- Gateway: LiteLLM proxy (OpenAI-compatible) in front of AWS Bedrock
- All models affected (validation fires before model routing)

dgageot:
## Summary

Fixes #2804.

The OpenAI spec treats an omitted `tool_choice` as equivalent to `"auto"`, but strict OpenAI-compatible gateways (notably LiteLLM-based proxies) reject requests with a 400 when the field is absent:

```
HTTP 400: litellm.BadRequestError: BedrockException -
{"message":"The model returned the following errors: tool_choice.type: Field required"}
```

This change sends `tool_choice: "auto"` explicitly in both the Chat Completions and Responses API paths whenever `tools` is non-empty. The behavior is:

- **Spec-compliant** — matches the OpenAI default exactly.
- **Backwards-compatible** — no behavior change for gateways that already handle omission correctly.
- **Safe for the agentic loop** — `auto` preserves the model's autonomy to decide when to call tools.

When no tools are provided, `tool_choice` is correctly omitted (sending it with no `tools` would be invalid).

## Changes

- `pkg/model/provider/openai/client.go` — set `ToolChoice` explicitly in both `CreateChatCompletionStream` and `CreateResponseStream` when tools are present.
- `pkg/model/provider/openai/tool_choice_test.go` — verifies the Chat Completions wire format.
- `pkg/model/provider/openai/tool_choice_responses_test.go` — verifies the Responses API wire format.

Tests assert the actual JSON request body (not just SDK types) to confirm the field is on the wire.

## Validation

- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./pkg/model/provider/openai/...` ✅
- `golangci-lint run ./pkg/model/provider/openai/...` ✅ (0 issues)

### DIFF
--- a/pkg/fake/proxy.go
+++ b/pkg/fake/proxy.go
@@ -220,6 +220,11 @@ func DefaultMatcher(onError func(err error)) recorder.MatcherFunc {
 	thinkingConfigRegex := regexp.MustCompile(`"thinkingConfig":\{[^}]*\},?`)
 	// Normalize OpenAI reasoning config (varies based on NoThinking flag and thinking budget).
 	reasoningRegex := regexp.MustCompile(`"reasoning":\{[^}]*\},?`)
+	// Normalize OpenAI tool_choice field. The string form ("auto", "none",
+	// "required") is now sent explicitly whenever tools are present so that
+	// strict gateways (LiteLLM) accept the request, but older cassettes were
+	// recorded without it.
+	toolChoiceRegex := regexp.MustCompile(`"tool_choice":"[^"]*",?`)
 
 	return func(r *http.Request, i cassette.Request) bool {
 		if r.Body == nil || r.Body == http.NoBody {
@@ -249,10 +254,12 @@ func DefaultMatcher(onError func(err error)) recorder.MatcherFunc {
 		normalizedReq = maxTokensRegex.ReplaceAllString(normalizedReq, "")
 		normalizedReq = thinkingConfigRegex.ReplaceAllString(normalizedReq, "")
 		normalizedReq = reasoningRegex.ReplaceAllString(normalizedReq, "")
+		normalizedReq = toolChoiceRegex.ReplaceAllString(normalizedReq, "")
 		normalizedCassette := callIDRegex.ReplaceAllString(i.Body, "call_ID")
 		normalizedCassette = maxTokensRegex.ReplaceAllString(normalizedCassette, "")
 		normalizedCassette = thinkingConfigRegex.ReplaceAllString(normalizedCassette, "")
 		normalizedCassette = reasoningRegex.ReplaceAllString(normalizedCassette, "")
+		normalizedCassette = toolChoiceRegex.ReplaceAllString(normalizedCassette, "")
 
 		return normalizedReq == normalizedCassette
 	}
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the LiteLLM/Bedrock gateway incompatibility by explicitly setting `tool_choice: "auto"` whenever tools are present — in both the Chat Completions and Responses API paths. The implementation is minimal, spec-compliant, and backwards-compatible.

**Summary of analysis:**
- Both code paths (`CreateChatCompletionStream` and `CreateResponseStream`) guard the `tool_choice` field behind `len(tools) > 0`, which is correct.
- The SDK union param usage (`ChatCompletionToolChoiceOptionUnionParam{OfAuto: param.NewOpt("auto")}`) is confirmed correct by the wire-format test that captures and deserializes the actual HTTP request body.
- The Responses API path uses `responses.ToolChoiceOptionsAuto` constant — the correct typed approach.
- Tests cover: tools present (verifies `tool_choice` is on the wire as bare string `"auto"`), and tools absent (verifies `tool_choice` is omitted).
- No high or medium severity issues found.

rumpl (dismissed):

rumpl (approved):

PR #2814, opened 2026-05-18
fix(runtime): use provider_opts.context_size for compaction

### LINKED ISSUE #2800: Failed to get model definition wit `/compact` + Docker Model Runner
When running docker-agent against a model served by Docker Model Runner (provider dmr), the manual /compact command always fails with the error: `Failed to get model definition`

Additionally, automatic compaction never runs.

**Environment**

- Provider: dmr (Docker Model Runner, local inference via llama-server)
- Example model: huggingface.co/unsloth/qwen3.5-4b-gguf:Q4_K_M
- provider_opts.context_size: 32768 is set in the agent YAML
- base_url: http://localhost:12434/engines/v1

**Steps to reproduce**

1. Configure an agent that uses a local DMR model (see agent.yaml snippet below).
2. Start a conversation and let it grow.
3. Run /compact at any point.

```yaml
  models:
    qwen-local:
      provider: dmr
      model: huggingface.co/unsloth/qwen3.5-4b-gguf:Q4_K_M
      base_url: http://localhost:12434/engines/v1
      max_tokens: 24576
      provider_opts:
        context_size: 32768
```

**Expected behavior**

- /compact should produce a summary of the conversation and free up context.
- When the conversation approaches the configured context_size (32768), proactive compaction should trigger automatically.

**Actual behavior**

- /compact emits Failed to get model definition and exits without compacting anything.
- The conversation continues to grow until the context window overflows; automatic compaction never fires.

Version of docker-agent: 1.50.0

dgageot:
## Summary

Fixes #2800.

When Docker Model Runner (DMR) is configured with a model that isn't catalogued in models.dev — typically a HuggingFace GGUF such as `huggingface.co/unsloth/qwen3.5-4b-gguf:Q4_K_M` — automatic compaction silently became a no-op:

- `compactionContextLimit` returned `0`, so the LLM strategy bailed.
- The proactive 90% trigger in `runStreamLoop` never fired.
- Post-overflow recovery surfaced as `Failed to get model definition` every time the assistant tried to respond, with no way out.

The user already supplies a `context_size` in `provider_opts` for DMR to size the inference context. The runtime now uses that same value as the authoritative context limit when set, falling back to the models.dev catalogue otherwise. This keeps planning aligned with what the engine actually enforces.

## Resolution order

1. `provider_opts.context_size` (when set and parseable as a positive integer)
2. models.dev catalogue limit
3. `0` — caller treats as \"can't compact\"

A single `LocalRuntime.resolveContextLimit` helper is the source of truth, used by:

- `compactionContextLimit` (LLM compaction strategy)
- `runStreamLoop` proactive 90% trigger
- `EmitStartupInfo` sidebar context-percent on session restore
- `compactWithReason` post-compaction `TokenUsageEvent`

So the sidebar, the proactive trigger, and the LLM compactor all plan against the same number.

## Tests

- 12-case helper matrix covering int / int64 / int32 / float64 / float32 / string-decimal / whitespace / non-numeric / negative / zero / bool / missing-key / nil-opts.
- Nil-provider safety.
- `provider_opts.context_size` takes precedence over the catalogue.
- Falls back to the catalogue when `context_size` is unset.
- Falls back to `provider_opts.context_size` when `modelsStore.GetModel` errors (the exact reported scenario).
- Returns `0` when neither source yields a usable limit.

\`task lint\` — 0 issues. \`task test\` — full suite passes.

Closes #2800

### DIFF
--- a/pkg/runtime/compaction_context_limit_test.go
+++ b/pkg/runtime/compaction_context_limit_test.go
@@ -1,0 +1,162 @@
+package runtime
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/pkg/agent"
+	"github.com/docker/docker-agent/pkg/chat"
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/model/provider/base"
+	"github.com/docker/docker-agent/pkg/modelsdev"
+	"github.com/docker/docker-agent/pkg/team"
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+// providerOptsProvider is a minimal provider used to test that
+// [providerContextLimit] reads the user-supplied context_size from
+// the resolved [latest.ModelConfig.ProviderOpts] map.
+type providerOptsProvider struct {
+	id   string
+	opts map[string]any
+}
+
+func (p *providerOptsProvider) ID() modelsdev.ID { return modelsdev.ParseIDOrZero(p.id) }
+
+func (p *providerOptsProvider) CreateChatCompletionStream(context.Context, []chat.Message, []tools.Tool) (chat.MessageStream, error) {
+	return &mockStream{}, nil
+}
+
+func (p *providerOptsProvider) BaseConfig() base.Config {
+	return base.Config{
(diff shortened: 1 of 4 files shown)


rumpl (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26026863720).

PR #2818, opened 2026-05-19
modelerrors: make overflow errors more specific

trungutt:
## Motivation

Context-overflow errors from LLM providers come in three distinct shapes, but `modelerrors` collapses them into one. The conflation has user-visible consequences:

- A provider rejecting a single oversized request (HTTP 413, e.g. paste over the wire-body cap) is reported as *"context window exceeded — try /compact"* — but `/compact` cannot help, because the offending turn still has to be sent on every subsequent call. Sessions get stuck failing in the same way until the user starts over.
- An image too large for the provider is reported as a generic context error, hiding the actual cause.
- Token-count overflow and wire-size overflow share one retry path even though they need different recovery.

Without distinguishing them at the classifier, every downstream consumer (runtime retry logic, error-code routing, UI hints) has to guess.

## Change

Introduce `OverflowKind` with three values, classified by a small two-tier function:

| Kind     | Meaning                                                | Recovery shape     |
|----------|--------------------------------------------------------|--------------------|
| `tokens` | Conversation exceeds the model's context window        | Compaction helps   |
| `wire`   | Request body exceeds the provider's wire-level limit   | Compaction cannot  |
| `media`  | Image, PDF, or attachment too large                    | Strip media        |

Classifier:

1. **Structured signals** (high confidence): HTTP 413, `body.error.type == "request_too_large"`, `body.error.code == "context_length_exceeded"`.
2. **Provider-prose substrings** as a fallback, covering the observed error wording across Anthropic, OpenAI, Bedrock, Gemini, Mistral, Groq, Vertex, OpenRouter, Ollama, Kimi, MiniMax, z.ai, and others. Best-effort — provider wording is not contractual, so this list is expected to drift and is easy to extend (one line per provider phrase).

`FormatError` now returns three distinct, provider-agnostic messages (no vendor names — the wire cap is a deployment detail of the provider, not something the user can act on by name). Two new `ErrorCode` constants (`request_too_large`, `media_too_large`) reach external consumers via the existing `ErrorEvent.Code` channel so they can render the right hint per kind.

## What is preserved

- `IsContextOverflowError` matches the same set of errors as before.
- Retry classification in `ClassifyModelError` is unchanged (overflow stays non-retryable).
- Auto-compaction still fires on overflow (this PR labels; it does not yet change control flow).
- `NewContextOverflowError` now auto-fills `Kind` from the underlying error, so existing wrap sites in `pkg/runtime/fallback.go` get the correct kind without code changes.

### DIFF
--- a/pkg/modelerrors/modelerrors.go
+++ b/pkg/modelerrors/modelerrors.go
@@ -76,19 +76,51 @@ const (
 	DefaultCooldown = 1 * time.Minute
 )
 
+// OverflowKind classifies the cause of a context overflow, so the runtime
+// and UI can react differently to each shape.
+//
+//   - [OverflowKindTokens]: the accumulated conversation exceeds the model's
+//     context window. Token-count rejection. Compaction can usually help.
+//   - [OverflowKindWire]: the request body exceeds the provider's wire-level
+//     limit (e.g. Anthropic's 32 MB cap, gateway 413s). Compaction usually
+//     CANNOT help because the offending single turn still has to be sent.
+//   - [OverflowKindMedia]: an image, PDF, or similar attachment in the
+//     conversation exceeds the provider's media constraints (size, page
+//     count, dimensions).
+type OverflowKind string
+
+const (
+	OverflowKindTokens OverflowKind = "tokens"
+	OverflowKindWire   OverflowKind = "wire"
+	OverflowKindMedia  OverflowKind = "media"
+)
+
 // ContextOverflowError wraps an underlying error to indicate that the failure
-// was caused by the conversation context exceeding the model's context window.
+// was caused by the conversation context exceeding some provider-side limit.
 // This is used to trigger auto-compaction in the runtime loop instead of
 // surfacing raw HTTP errors to the user.
+//
+// Kind classifies the specific shape of the overflow ([OverflowKindTokens] by
+// default for backwards compatibility). Use [NewContextOverflowError] to have
+// it set automatically by classification, or build the struct directly to
+// force a Kind.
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The overflow classification logic is well-structured and correct. The two-tier classifier (structured signals → prose fallback), the `OverflowKindOf` unwrapping semantics, `FormatError` switch, and `classifyErrorCode` routing all look sound. Backward compatibility is preserved: `IsContextOverflowError` still matches the same error set, and unclassifiable wraps default to `OverflowKindTokens`. Test coverage is thorough across 26 classifier cases, wrap semantics, and all three `FormatError` paths.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The classifier's two-tier approach (structured signals first, substring patterns as fallback) is sound. Tier 1 structured checks correctly map HTTP 413 → `OverflowKindWire` and `context_length_exceeded` → `OverflowKindTokens`. The `OverflowKindOf` helper correctly handles legacy zero-Kind wraps by falling back to classification on the underlying error. `NewContextOverflowError` auto-fills `Kind` correctly via `classifyOverflow`. The `classifyErrorCode` switch in `loop_steps.go` correctly routes all three kinds to distinct `ErrorCode` constants.

**Intentional design noted**: The PR description explicitly states "Auto-compaction still fires on overflow (this PR labels; it does not yet change control flow)" — the `handleStreamError` check is intentionally unchanged; control-flow differentiation for Wire/Media is deferred to a follow-up.

dgageot (approved):

PR #2819, opened 2026-05-19
runtime: skip auto-compaction for non-token overflow

trungutt:
Stacked on #2818. Review after that one merges.

When a request is rejected because the **whole request body** is too big to send (e.g. HTTP 413, large paste, oversized attachment), the runtime today reacts the same way it reacts to a token-count overflow: it tries to recover by compacting the conversation.

Compaction is itself a model call. It has to send the same conversation history that just got rejected, including the oversized turn. So that call is rejected too. Then the runtime gives up — but only after burning a second provider call and several seconds of wall-clock latency, and the user sees a *"context window exceeded — try /compact"* message that is actively wrong (compaction is exactly what just failed, and starting a new session is the only thing that helps).

The same trap applies to media-size rejections: there is no media-stripping path during compaction yet, so the offending attachment gets resent and the compaction call fails the same way.

```
Today                                       After this change
─────                                       ─────────────────
User sends an oversized message             User sends an oversized message
   │                                           │
   ▼                                           ▼
Provider → wire-level rejection             Provider → wire-level rejection
   │                                           │
   ▼                                           ▼
"context overflow" → start compaction       Recognise: compaction cannot help
   │                                           │
   ▼                                           ▼
Compaction model call                       Skip compaction entirely
   │   (with same oversized history)           │
   ▼                                           ▼
Rejected again, same reason                 Emit a specific, actionable error
   │                                           │
   ▼                                           ▼
Give up; show "context window exceeded"     User sees the right message

```

## What this change achieves

- A wire-level overflow (request body too large) is surfaced immediately with the structured error code introduced in #2818, instead of being routed through a compaction attempt that is guaranteed to fail.
- A media-size overflow is treated the same way, because the current compaction path cannot remove the offending attachment.
- Token-count overflow — the case compaction actually helps — is unchanged. Auto-compaction still fires there, with the same limits and the same retry semantics as before.

### DIFF
--- a/pkg/runtime/loop_steps.go
+++ b/pkg/runtime/loop_steps.go
@@ -161,22 +161,42 @@ func (r *LocalRuntime) handleStreamError(
 	// request instead of surfacing raw errors. We allow at most
 	// r.maxOverflowCompactions consecutive attempts to avoid an infinite
 	// loop when compaction cannot reduce the context enough.
+	//
+	// Compaction only helps for token-count overflow ([OverflowKindTokens]):
+	// summarising older turns reduces the input token count.
+	//
+	// For wire-level overflow ([OverflowKindWire]) the request body itself
+	// is over the provider's cap; the latest turn alone is too large and
+	// would still have to be sent during compaction. For media overflow
+	// ([OverflowKindMedia]) we have no media-stripping path today, so a
+	// retry would resend the same oversized attachment. In both cases the
+	// recovery attempt always fails, so we skip it and surface the error
+	// directly — the user can act on it (smaller paste, smaller file)
+	// faster, without burning a second provider call.
 	if _, ok := errors.AsType[*modelerrors.ContextOverflowError](err); ok && r.sessionCompaction && *overflowCompactions < r.maxOverflowCompactions {
-		*overflowCompactions++
-		slog.WarnContext(ctx, "Context window overflow detected, attempting auto-compaction",
+		kind := modelerrors.OverflowKindOf(err)
+		if kind == modelerrors.OverflowKindTokens {
+			*overflowCompactions++
+			slog.WarnContext(ctx, "Context window overflow detected, attempting auto-compaction",
+				"agent", a.Name(),
+				"session_id", sess.ID,
+				"input_tokens", sess.InputTokens,
+				"output_tokens", sess.OutputTokens,
+				"context_limit", contextLimit,
+				"attempt", *overflowCompactions,
+			)
+			events.Emit(Warning(
+				"The conversation has exceeded the model's context window. Automatically compacting the conversation history...",
+				a.Name(),
+			))
+			r.compactWithReason(ctx, sess, "", compactionReasonOverflow, events)
+			return streamErrorRetry
+		}
+		slog.InfoContext(ctx, "Skipping auto-compaction for non-token overflow",
 			"agent", a.Name(),
 			"session_id", sess.ID,
-			"input_tokens", sess.InputTokens,
-			"output_tokens", sess.OutputTokens,
-			"context_limit", contextLimit,
-			"attempt", *overflowCompactions,
+			"overflow_kind", string(kind),
 		)
-		events.Emit(Warning(
-			"The conversation has exceeded the model's context window. Automatically compacting the conversation history...",
-			a.Name(),
-		))
-		r.compactWithReason(ctx, sess, "", compactionReasonOverflow, events)
-		return streamErrorRetry
 	}
 
 	streamSpan.RecordError(err)
--- a/pkg/runtime/loop_steps_test.go
+++ b/pkg/runtime/loop_steps_test.go
@@ -292,3 +292,75 @@ func TestHandleStreamError_GenericError_FatalAndEmitsError(t *testing.T) {
 	}
 	assert.True(t, sawError, "generic error should emit ErrorEvent")
 }
+
+// TestHandleStreamError_WireOverflowSkipsCompaction verifies that wire-level
+// overflow does not trigger auto-compaction. Compaction would resend the same
+// oversized request that just got rejected, so it is guaranteed to fail; we
+// surface the error directly instead.
+func TestHandleStreamError_WireOverflowSkipsCompaction(t *testing.T) {
+	t.Parallel()
+
+	rt, a := newTestRuntime(t)
+	sess := session.New()
+	events := make(chan Event, 16)
+	_, sp := noop.NewTracerProvider().Tracer("t").Start(t.Context(), "x")
+
+	overflow := &modelerrors.ContextOverflowError{
+		Underlying: errors.New("HTTP 413: Payload Too Large"),
+		Kind:       modelerrors.OverflowKindWire,
+	}
+	overflowCount := 0
+
+	outcome := rt.handleStreamError(t.Context(), sess, a, overflow, 1000, &overflowCount, sp, NewChannelSink(events))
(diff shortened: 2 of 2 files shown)


pkg/modelerrors/modelerrors.go
```diff
@@ -437,16 +565,26 @@ func ClassifyModelError(err error) (retryable, rateLimited bool, retryAfter time
...
-	if _, ok := errors.AsType[*ContextOverflowError](err); ok {
+	switch OverflowKindOf(err) {
+	case OverflowKindWire:
+		return "Your message is too large for the AI provider. " +
+			"Try a smaller paste, attach the file separately, or split the content."
+	case OverflowKindMedia:
+		return "An image or file in this conversation is too large for the AI provider. " +
+			"Try a smaller file or remove it from context."
+	case OverflowKindTokens:
 		return "The conversation has exceeded the model's context window and automatic compaction is not enabled. " +
```
docker-agent, line 587: **[MEDIUM] `FormatError` returns misleading "automatic compaction is not enabled" when compaction is enabled but exhausted**

`FormatError` unconditionally returns this message for `OverflowKindTokens`:

```
"The conversation has exceeded the model's context window and automatic compaction is not enabled. Try running /compact…"
```

However, this path is reached in **two** distinct scenarios:
1. Compaction is genuinely disabled (`r.sessionCompaction == false`) — message is accurate.
2. Compaction is enabled but the retry limit is exhausted (`*overflowCompactions >= r.maxOverflowCompactions`) — message is **factually wrong**; compaction is enabled but was already attempted.

In case 2, users with auto-compaction turned on who hit `maxOverflowCompactions` will see "automatic compaction is not enabled", which is the opposite of the truth and could cause confusion ("I do have compaction enabled — why does it say it's not?").

**Suggested fix**: pass the compaction state/exhaustion into `FormatError` (or use a separate message variant), e.g.:
```go
case OverflowKindTokens:
    if compactionEnabled {
        return "The conversation has exceeded the model's context window and compaction could not reduce it further. Try starting a new session."
    }
    return "The conversation has exceeded the model's context window and automatic compaction is not enabled. Try running /compact…"
```

pkg/modelerrors/modelerrors.go
```diff
@@ -102,60 +134,156 @@ func (e *ContextOverflowError) Unwrap() error {
...
+	"max_tokens must be greater than",   // Anthropic edge case: thinking-budget cascade
 	"maximum number of tokens",
 	"content length exceeds",
-	"request too large",
-	"payload too large",
-	"input is too long",
 	"exceeds the model's max token",
 	"token limit",
 	"reduce your prompt",
 	"reduce the length",
```
docker-agent: **[MEDIUM] Overly broad `tokenOverflowPatterns` entry `"reduce the length"` risks misclassifying non-overflow validation errors**

The newly added pattern `"reduce the length"` is a very broad substring match. It will hit any provider error message containing that phrase — including parameter-validation errors that have nothing to do with context-window limits, e.g.:

- *"Please reduce the length of your system prompt"*
- *"Please reduce the length of the tools list"*

If such a validation error is misclassified as `OverflowKindTokens`:
- `handleStreamError` may attempt unnecessary auto-compaction (wasting a provider call).
- `FormatError` returns the wrong user message ("context window exceeded…" instead of the actual validation error).
- `classifyErrorCode` emits `ErrorCodeContextExceeded` instead of whatever code is appropriate for the real error.

Every other entry in this list is more specific (e.g. `"reduce the length of the messages"` for Groq). Consider tightening this to a more specific phrase or removing it if no concrete provider was observed to emit this exact string as a context-window signal.

docker-agent (commented):

pkg/modelerrors/modelerrors.go
trungutt: Good catch — this message text is factually wrong when compaction is enabled but exhausted, and that case is reachable.

For context though: the misleading wording predates this PR (the original `FormatError` returned the same string when compaction was disabled OR exhausted), so it isn't a regression introduced here. Fixing it cleanly needs threading the compaction state into `FormatError`, which is a signature change with reach beyond this stack.

Leaving as-is for this PR and tracking it as a follow-up — the broader plan is to revisit the tokens-overflow recovery path after this stack lands (specifically, to distinguish "compaction was attempted but couldn't reduce further" from "compaction was never tried" at the call site, and emit the right user message from there).

pkg/modelerrors/modelerrors.go
trungutt: Agreed and fixed. The broad `"reduce the length"` entry is gone — the Groq-specific `"reduce the length of the messages"` covers the actual provider we observed, and the broader pattern was inherited from the pre-existing list rather than added intentionally here.

Force-pushed onto the Step 1 branch (`overflow-kind-classification`) so this stack carries the fix. The existing test that exercised this case (`"please reduce the length of the messages"`) still passes via the more specific pattern.

docker-agent: Perfect, thanks for catching that inheritance issue and tightening it up! The Groq-specific pattern is exactly the right level of specificity — it'll match the real-world error while avoiding false positives on unrelated validation messages.

pkg/modelerrors/modelerrors.go
docker-agent: Thanks for the context! That makes complete sense — leaving the pre-existing wording as-is for this PR and tracking the fix separately is the right call. The signature change to thread compaction state into `FormatError` does touch multiple call sites, and bundling that refactor with the overflow classification work would muddy the PR scope.

Since the plan is to revisit the tokens-overflow recovery path post-merge and emit the right message at the call site (where the compaction state is already known), the fix should be straightforward once this stack lands. Marking this resolved — appreciate the explanation!

docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is correct and the logic is well-structured. Both overflow-kind classification and the compaction-skip decision are sound.

**What was checked:**
- `OverflowKind` type and constants (`tokens`, `wire`, `media`) — correct
- `classifyOverflow` two-tier classification (structured Tier 1 before substring Tier 2, media before wire before tokens) — correct priority ordering
- `OverflowKindOf` fallback chain for legacy `*ContextOverflowError` wraps — correct; defaults to `OverflowKindTokens` as historical behaviour
- `handleStreamError` skip-compaction branch: wire/media overflows return `streamErrorFatal` → `runTurn` returns `turnExit` → outer loop terminates immediately. No circuit-breaker bypass, no infinite retry.
- `FormatError` switch: non-overflow errors (`OverflowKindOf` returns `""`) correctly fall through to `err.Error()`. All three overflow kinds have actionable messages.
- `classifyErrorCode` switch covers all three new overflow kinds with distinct error codes.
- Test coverage: `TestHandleStreamError_WireOverflowSkipsCompaction` and `TestHandleStreamError_MediaOverflowSkipsCompaction` directly exercise the new skip paths. `TestClassifyOverflow` covers Tier 1 and Tier 2 patterns including the 413-wins-over-token-prose case.

**Hypotheses investigated and dismissed:**
1. *Counter not incremented for wire/media = defeated circuit-breaker*: Wire/media overflow always returns `streamErrorFatal`, which exits the loop via `turnExit`. The counter is irrelevant because there is no re-entry.
2. *`"content length exceeds"` misclassification*: That entry is pre-existing, unchanged code (context line in diff, not a `+` line) — outside the scope of this review.

dgageot (approved):

PR #2820, opened 2026-05-19
Misc Security fixes

dgageot:


pkg/tools/mcp/oauth_server.go
```diff
@@ -171,14 +178,13 @@ func (cs *CallbackServer) handleCallback(w http.ResponseWriter, r *http.Request)
...
-		cs.errCh <- errors.New("OAuth state mismatch (possible CSRF attempt or stale callback)")
+		cs.deliver(callbackResult{err: errors.New("OAuth state mismatch (possible CSRF attempt or stale callback)")})
 		w.WriteHeader(http.StatusBadRequest)
 		fmt.Fprint(w, "Invalid state parameter")
 		return
 	}
 
-	cs.codeCh <- code
-	cs.stateCh <- state
+	cs.deliver(callbackResult{code: code, state: state})
```
docker-agent: **[MEDIUM] Duplicate valid OAuth callbacks silently dropped but still return HTTP 200 "Authorization Successful"**

After a valid `code`+`state` pair passes CSRF verification, `handleCallback` calls `cs.deliver(...)` (non-blocking) and then **unconditionally** writes `HTTP 200` with an "Authorization Successful!" page — even when `deliver()` is a no-op because `resultCh` is already full (i.e., a first callback already won).

This means the 2nd, 3rd, … Nth duplicate callbacks (stale browser tabs, page refreshes, or any local process probing the loopback port) all receive a success page, even though their result was silently discarded. A user who refreshes the browser tab mid-flow sees "Authorization Successful!" for a request that actually did nothing, which is misleading. Worse, an attacker who can send crafted requests to the loopback port can use this to probe whether a valid authorization flow is in progress (200 = flow active and CSRF nonce already known; 400 = not active or wrong state).

**Suggested fix**: have `deliver` return `bool`, and return `HTTP 409 Conflict` (or `HTTP 200` with a "already processed" message) when the result was dropped:
```go
func (cs *CallbackServer) deliver(r callbackResult) bool {
    select {
    case cs.resultCh <- r:
        return true
    default:
        return false
    }
}

// In handleCallback, after the CSRF check:
if !cs.deliver(callbackResult{code: code, state: state}) {
    w.WriteHeader(http.StatusConflict)
    fmt.Fprint(w, "Authorization already processed")
    return
}
// Only reach here if this callback actually won the race
w.WriteHeader(http.StatusOK)
fmt.Fprint(w, `...Authorization Successful...`)
```

docker-agent (commented):

Security PR with mostly mechanical improvements (sentinel error comparisons, SSRF-safe HTTP client, Cache-Control parsing refactor). One finding in the OAuth callback handler.

aheritier (approved):

PR #2821, opened 2026-05-19
runtime: recover oversized user message after wire/media overflow

trungutt:
Stacked on #2819. Review after that one merges.

## The problem this step fixes

After #2818 + #2819, an oversized message produces the right error message and fails fast — but the **same chat session cannot continue**. The offending user message stays verbatim in `sess.Messages` for the rest of the process, and every retry resends the same oversized payload alongside the new (smaller) one. The user sees the same rejection every time they hit Send.

The original user-reported regression:

> *"I shortened my paste but it kept failing with the same character-limit error."*

## What this change does

After a wire- or media-overflow rejection, walk back to the most recent user message and **rewrite it in place in memory**:

- Each media part (image, file, document) becomes a short text placeholder that records what was attached (name, size, MIME when known).
- Plain-text content over a conservative threshold is replaced with a placeholder that records the original size.

The rewrite happens immediately after the failure event fires. The error itself is still emitted with the kind-specific code from #2818, and a Warning event explains to the user that the previous message was rewritten — so the hygiene action is observable rather than silent.

## Before / after

```
Today (after #2818 + #2819)          After this change
───────────────────────────          ─────────────────
User pastes huge content             User pastes huge content
       │                                    │
       ▼                                    ▼
Provider rejects                     Provider rejects
       │                                    │
       ▼                                    ▼
Error surfaced with the right code   Error surfaced with the right code
       │                                    │
       ▼                                    ▼
User shortens, retries               Offending message rewritten in
       │                              memory: media → placeholders,
       ▼                              oversized text → placeholder
sess.Messages still carries                 │
the oversized turn                          ▼
       │                             User shortens, retries
       ▼                                    │
Same limit tripped again                    ▼
       │                             sess.Messages now carries only
       ▼                             the slim placeholder + the new
Chat session stuck                   smaller message
                                            │
                                            ▼
                                     Provider accepts — chat session
                                     continues normally
```

## Scope — what this PR is and is not

| | In scope (this PR) | Out of scope (separate follow-up) |
|---|---|---|
| **Same-process recovery** | ✓ in-memory `sess.Messages` rewritten so the chat session continues immediately | — |
| **Persistence** | — | Mirroring the rewrite to the session store so it survives a docker-agent restart |

### Why the persistence side is a separate follow-up

The persistence side requires `Message.ID` to round-trip through `Store.AddMessage` (currently the returned ID is discarded by the `PersistenceObserver`) and through `loadSessionItems` (currently the `id` column is not selected on reload). That gap is **independent of overflow handling** — it would affect anything that needs `Store.UpdateMessage` against an in-memory message, including any future compaction-by-id work or message-editing features.

Folding that infrastructure fix into this PR doubled its size and conflated concerns. It now lands as its own focused change where it can be evaluated on its own merits (propagate the ID? position-based updates? new API?).

The scope of the persistence gap, for clarity:

```
Affected         Not affected
────────         ────────────
docker-agent     The chat session within a single docker-agent process
restart mid-     (closed by this PR)
session   →
oversized
message reloads
from disk
```

For the user-reported regression (paste oversized → 413 → shortening keeps failing in the same chat), the same-process fix in this PR is sufficient.

## What is preserved

- Token overflow handling is unchanged — that path goes through auto-compaction, which is the correct mechanism for that shape of failure.
- Hooks, telemetry, retry semantics: untouched.
- The fatal `ErrorEvent` for the original rejection is still emitted — scrubbing is **in addition to** surfacing the error, not instead of it.

## Files touched

```
pkg/runtime/loop_steps.go              Hook recoverFromOversizedTurn into handleStreamError
pkg/runtime/loop_steps_test.go         Tests for the new branch
pkg/runtime/overflow_recovery.go       scrubMessage + recoverFromOversizedTurn (new file)
pkg/runtime/overflow_recovery_test.go  Tests for scrub logic + recovery (new file)
pkg/session/session.go                 New Session.RewriteLatestUserMessage helper
```

5 files, ~850 lines (about half of that is tests).

### DIFF
--- a/pkg/runtime/loop_steps.go
+++ b/pkg/runtime/loop_steps.go
@@ -161,22 +161,42 @@ func (r *LocalRuntime) handleStreamError(
 	// request instead of surfacing raw errors. We allow at most
 	// r.maxOverflowCompactions consecutive attempts to avoid an infinite
 	// loop when compaction cannot reduce the context enough.
+	//
+	// Compaction only helps for token-count overflow ([OverflowKindTokens]):
+	// summarising older turns reduces the input token count.
+	//
+	// For wire-level overflow ([OverflowKindWire]) the request body itself
+	// is over the provider's cap; the latest turn alone is too large and
+	// would still have to be sent during compaction. For media overflow
+	// ([OverflowKindMedia]) we have no media-stripping path today, so a
+	// retry would resend the same oversized attachment. In both cases the
+	// recovery attempt always fails, so we skip it and surface the error
+	// directly — the user can act on it (smaller paste, smaller file)
+	// faster, without burning a second provider call.
 	if _, ok := errors.AsType[*modelerrors.ContextOverflowError](err); ok && r.sessionCompaction && *overflowCompactions < r.maxOverflowCompactions {
-		*overflowCompactions++
-		slog.WarnContext(ctx, "Context window overflow detected, attempting auto-compaction",
+		kind := modelerrors.OverflowKindOf(err)
+		if kind == modelerrors.OverflowKindTokens {
+			*overflowCompactions++
+			slog.WarnContext(ctx, "Context window overflow detected, attempting auto-compaction",
+				"agent", a.Name(),
+				"session_id", sess.ID,
+				"input_tokens", sess.InputTokens,
+				"output_tokens", sess.OutputTokens,
+				"context_limit", contextLimit,
+				"attempt", *overflowCompactions,
+			)
+			events.Emit(Warning(
+				"The conversation has exceeded the model's context window. Automatically compacting the conversation history...",
+				a.Name(),
+			))
+			r.compactWithReason(ctx, sess, "", compactionReasonOverflow, events)
+			return streamErrorRetry
+		}
+		slog.InfoContext(ctx, "Skipping auto-compaction for non-token overflow",
 			"agent", a.Name(),
 			"session_id", sess.ID,
-			"input_tokens", sess.InputTokens,
-			"output_tokens", sess.OutputTokens,
-			"context_limit", contextLimit,
-			"attempt", *overflowCompactions,
+			"overflow_kind", string(kind),
 		)
-		events.Emit(Warning(
-			"The conversation has exceeded the model's context window. Automatically compacting the conversation history...",
-			a.Name(),
-		))
-		r.compactWithReason(ctx, sess, "", compactionReasonOverflow, events)
-		return streamErrorRetry
 	}
 
 	streamSpan.RecordError(err)
--- a/pkg/runtime/loop_steps.go
+++ b/pkg/runtime/loop_steps.go
@@ -156,26 +156,28 @@ func (r *LocalRuntime) handleStreamError(
 		return streamErrorFatal
 	}
 
-	// Auto-recovery: if the error is a context overflow and session
-	// compaction is enabled, compact the conversation and retry the
-	// request instead of surfacing raw errors. We allow at most
-	// r.maxOverflowCompactions consecutive attempts to avoid an infinite
-	// loop when compaction cannot reduce the context enough.
+	// Overflow handling has two independent concerns:
 	//
-	// Compaction only helps for token-count overflow ([OverflowKindTokens]):
-	// summarising older turns reduces the input token count.
+	//   1. Auto-compaction (token overflow only): summarise older
+	//      turns to fit the context window, then retry. Gated by
+	//      r.sessionCompaction and the per-run attempt cap.
 	//
-	// For wire-level overflow ([OverflowKindWire]) the request body itself
-	// is over the provider's cap; the latest turn alone is too large and
-	// would still have to be sent during compaction. For media overflow
-	// ([OverflowKindMedia]) we have no media-stripping path today, so a
-	// retry would resend the same oversized attachment. In both cases the
-	// recovery attempt always fails, so we skip it and surface the error
-	// directly — the user can act on it (smaller paste, smaller file)
-	// faster, without burning a second provider call.
-	if _, ok := errors.AsType[*modelerrors.ContextOverflowError](err); ok && r.sessionCompaction && *overflowCompactions < r.maxOverflowCompactions {
+	//   2. Session hygiene (wire/media overflow): rewrite the
+	//      offending user message so the same oversized payload
+	//      cannot reload on the next call and re-poison the session.
+	//      Always runs when the kind warrants it, independent of
+	//      the compaction config — the hygiene step does not retry
+	//      and is correct even when compaction is disabled.
+	if _, ok := errors.AsType[*modelerrors.ContextOverflowError](err); ok {
 		kind := modelerrors.OverflowKindOf(err)
-		if kind == modelerrors.OverflowKindTokens {
+
+		// Token overflow: compaction is the right recovery — older
+		// turns can be summarised to free up context. Wire/media do
+		// not benefit from compaction (the latest turn alone is
+		// over the cap; resending it during compaction would just
+		// fail again), so we fall through to the hygiene step
+		// below for those.
+		if kind == modelerrors.OverflowKindTokens && r.sessionCompaction && *overflowCompactions < r.maxOverflowCompactions {
 			*overflowCompactions++
 			slog.WarnContext(ctx, "Context window overflow detected, attempting auto-compaction",
 				"agent", a.Name(),
@@ -192,11 +194,15 @@ func (r *LocalRuntime) handleStreamError(
 			r.compactWithReason(ctx, sess, "", compactionReasonOverflow, events)
 			return streamErrorRetry
 		}
-		slog.InfoContext(ctx, "Skipping auto-compaction for non-token overflow",
-			"agent", a.Name(),
-			"session_id", sess.ID,
-			"overflow_kind", string(kind),
-		)
+
+		// Hygiene scrub for wire/media overflow. Runs independently
+		// of r.sessionCompaction: this rewrites a single message in
+		// place, it does not retry, and the same-process
+		// session-poisoning bug it fixes occurs regardless of
+		// whether the user opted into auto-compaction.
+		if kind == modelerrors.OverflowKindWire || kind == modelerrors.OverflowKindMedia {
+			r.recoverFromOversizedTurn(ctx, sess, kind, events)
+		}
 	}
 
 	streamSpan.RecordError(err)
--- a/pkg/runtime/overflow_recovery.go
+++ b/pkg/runtime/overflow_recovery.go
@@ -1,0 +1,289 @@
+package runtime
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"path/filepath"
+	"strings"
+
+	"github.com/docker/docker-agent/pkg/chat"
+	"github.com/docker/docker-agent/pkg/modelerrors"
+	"github.com/docker/docker-agent/pkg/session"
+)
+
+// maxScrubbedTextBytes is the threshold at and below which a user
+// message's plain-text content is preserved verbatim during scrubbing.
+// Above the threshold the content is replaced with a placeholder that
+// records the original size — keeping it verbatim would re-poison the
+// session on the next turn by carrying the offending payload back into
+// the provider request.
+//
+// The value is intentionally well below any major provider's wire-size
+// cap. Smaller payloads pass through unchanged.
+const maxScrubbedTextBytes = 1 << 20 // 1 MiB
+
+// scrubReport summarises what scrubMessage rewrote, for observability.
+// All counters are zero on a no-op scrub.
+type scrubReport struct {
+	// textReplaced is true when [chat.Message.Content] was over
+	// [maxScrubbedTextBytes] and has been replaced.
+	textReplaced bool
+	// originalBytes is the size of the original plain text, set
+	// only when textReplaced is true.
+	originalBytes int64
+	// partsReplaced counts how many MultiContent parts were
+	// rewritten: media parts (image/file/document) are always
+	// counted; oversized text parts are also counted when they
+	// exceeded [maxScrubbedTextBytes].
+	partsReplaced int
+}
+
+func (r scrubReport) didScrub() bool {
+	return r.textReplaced || r.partsReplaced > 0
+}
+
+// scrubMessage returns a copy of msg in which media parts (image_url,
+// file, document) are replaced with text placeholders and oversized
+// plain-text content is replaced with a size-noting placeholder. The
+// returned scrubReport describes what changed; when it reports
+// [scrubReport.didScrub] false the message is byte-identical to msg.
+//
+// scrubMessage is pure: it does not consult the session, the model,
+// or any context. Callers decide *when* to apply it (post-failure
+// recovery, manual cleanup, etc.); this function only describes
+// *how*.
+func scrubMessage(msg chat.Message) (chat.Message, scrubReport) {
+	var report scrubReport
+
+	out := msg
+
+	// Plain text: replace only when oversized so we don't lose the
+	// user's intent for normal-sized messages.
+	if len(out.Content) > maxScrubbedTextBytes {
+		report.textReplaced = true
+		report.originalBytes = int64(len(out.Content))
+		out.Content = oversizedTextPlaceholder(int64(len(out.Content)))
+	}
+
+	// Multi-content parts: rewrite each media part in place.
+	if len(out.MultiContent) > 0 {
+		newParts := make([]chat.MessagePart, len(out.MultiContent))
+		for i, part := range out.MultiContent {
+			rewritten, replaced := scrubMessagePart(part)
+			if replaced {
+				report.partsReplaced++
+			}
+			newParts[i] = rewritten
+		}
+		out.MultiContent = newParts
+	}
+
+	return out, report
+}
+
+// scrubMessagePart replaces a single attachment part with a text
+// placeholder. Returns the rewritten part and whether anything
+// changed; small text parts and unrecognised types pass through
+// unchanged.
+//
+// Text parts over [maxScrubbedTextBytes] are themselves rewritten to
+// a size-noting placeholder — a single text part inside MultiContent
+// can be just as poisoning as oversized [chat.Message.Content].
+//
+// The placeholder describes what was attached (kind, name, size)
+// without preserving any of the content, so the rewritten message
+// can never re-trip the provider's media-size limits.
+func scrubMessagePart(part chat.MessagePart) (chat.MessagePart, bool) {
+	switch part.Type {
+	case chat.MessagePartTypeText:
+		if int64(len(part.Text)) > maxScrubbedTextBytes {
+			return chat.MessagePart{
+				Type: chat.MessagePartTypeText,
+				Text: oversizedTextPlaceholder(int64(len(part.Text))),
+			}, true
+		}
+		return part, false
+
+	case chat.MessagePartTypeImageURL:
+		return chat.MessagePart{
+			Type: chat.MessagePartTypeText,
+			Text: imagePlaceholder(part),
+		}, true
+
+	case chat.MessagePartTypeFile:
+		return chat.MessagePart{
+			Type: chat.MessagePartTypeText,
+			Text: filePlaceholder(part),
+		}, true
+
+	case chat.MessagePartTypeDocument:
+		return chat.MessagePart{
+			Type: chat.MessagePartTypeText,
+			Text: documentPlaceholder(part),
+		}, true
+	}
+	// Unknown part type — leave it alone rather than risk dropping
+	// data we don't recognise.
+	return part, false
+}
+
+func oversizedTextPlaceholder(originalBytes int64) string {
+	return fmt.Sprintf(
+		"[previous message was %s of text — too large for the AI provider; "+
+			"content was removed from the session so the conversation can continue]",
+		humanByteSize(originalBytes),
+	)
+}
+
+func imagePlaceholder(part chat.MessagePart) string {
+	if part.ImageURL == nil {
+		return "[image attachment removed: too large for the AI provider]"
+	}
+	if name := imageDisplayName(part.ImageURL.URL); name != "" {
+		return fmt.Sprintf("[image %q removed: too large for the AI provider]", name)
+	}
+	return "[image attachment removed: too large for the AI provider]"
+}
+
+func filePlaceholder(part chat.MessagePart) string {
+	if part.File == nil {
+		return "[file attachment removed: too large for the AI provider]"
+	}
+	name := part.File.Path
+	if name != "" {
+		name = filepath.Base(name)
+	}
+	if name == "" {
+		name = part.File.FileID
+	}
+	if name == "" {
+		return "[file attachment removed: too large for the AI provider]"
+	}
+	return fmt.Sprintf("[file %q removed: too large for the AI provider]", name)
+}
+
+func documentPlaceholder(part chat.MessagePart) string {
+	if part.Document == nil {
+		return "[document attachment removed: too large for the AI provider]"
+	}
+	doc := part.Document
+	if doc.Name != "" && doc.Size > 0 {
+		return fmt.Sprintf("[document %q (%s) removed: too large for the AI provider]",
+			doc.Name, humanByteSize(doc.Size))
+	}
+	if doc.Name != "" {
+		return fmt.Sprintf("[document %q removed: too large for the AI provider]", doc.Name)
+	}
+	if doc.MimeType != "" {
+		return fmt.Sprintf("[%s attachment removed: too large for the AI provider]", doc.MimeType)
+	}
+	return "[document attachment removed: too large for the AI provider]"
+}
+
+// imageDisplayName extracts a short display name from an image URL.
+// Returns "" for data: URIs (where the URL itself is the payload and
+// not user-meaningful) and falls back to the URL path's basename
+// otherwise.
+func imageDisplayName(url string) string {
+	if url == "" || strings.HasPrefix(url, "data:") {
+		return ""
+	}
+	// Strip query / fragment.
+	if i := strings.IndexAny(url, "?#"); i >= 0 {
+		url = url[:i]
+	}
+	base := filepath.Base(url)
+	if base == "/" || base == "." {
+		return ""
+	}
+	return base
+}
+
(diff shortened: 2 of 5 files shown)


pkg/runtime/loop_steps.go
```diff
@@ -161,22 +161,55 @@ func (r *LocalRuntime) handleStreamError(
...
+	//
+	// For wire-level overflow ([OverflowKindWire]) the request body itself
+	// is over the provider's cap; the latest turn alone is too large and
+	// would still have to be sent during compaction. For media overflow
+	// ([OverflowKindMedia]) we have no media-stripping path today, so a
+	// retry would resend the same oversized attachment. In both cases the
+	// recovery attempt always fails, so we skip it and surface the error
+	// directly — the user can act on it (smaller paste, smaller file)
+	// faster, without burning a second provider call.
 	if _, ok := errors.AsType[*modelerrors.ContextOverflowError](err); ok && r.sessionCompaction && *overflowCompactions < r.maxOverflowCompactions {
```
docker-agent: **[MEDIUM] Wire/Media overflow session scrub is skipped when `sessionCompaction` is disabled**

`recoverFromOversizedTurn` (the hygiene scrub that rewrites the offending user message to prevent session poisoning) is nested inside the `r.sessionCompaction` guard on line 176. When a user disables session compaction (`WithSessionCompaction(false)`) and then receives a Wire or Media overflow error, the entire block is skipped — including the scrub call — so the oversized message remains verbatim in the session store. This is exactly the session-poisoning problem the PR aims to fix, and it silently regresses for that configuration.

The in-code comment explains *why compaction+retry is skipped* for Wire/Media kinds, but says nothing about why the scrub should also be skipped. The scrub (`recoverFromOversizedTurn`) is a distinct concern from compaction: it only rewrites the bad message, it does not retry, and it has value even when compaction is off.

**Suggested fix**: Separate the scrub call from the compaction gate. The scrub should execute for Wire/Media overflows regardless of whether `r.sessionCompaction` is true, e.g.:

```go
if overflowErr, ok := errors.AsType[*modelerrors.ContextOverflowError](err); ok {
    kind := modelerrors.OverflowKindOf(overflowErr)
    if kind == modelerrors.OverflowKindTokens && r.sessionCompaction && *overflowCompactions < r.maxOverflowCompactions {
        // ... token compaction path ...
    } else if kind != modelerrors.OverflowKindTokens {
        r.recoverFromOversizedTurn(ctx, sess, kind, events)
    }
}
```

pkg/runtime/overflow_recovery.go
```diff
@@ -0,0 +1,329 @@
...
+// scrubMessagePart replaces a single attachment part with a text
+// placeholder. Returns the rewritten part and whether anything
+// changed; text parts and unrecognised types pass through unchanged.
+//
+// The placeholder describes what was attached (kind, name, size)
+// without preserving any of the content, so the rewritten message
+// can never re-trip the provider's media-size limits.
+func scrubMessagePart(part chat.MessagePart) (chat.MessagePart, bool) {
+	switch part.Type {
+	case chat.MessagePartTypeText:
```
docker-agent, line 99: **[MEDIUM] Oversized `MultiContent` text parts are not scrubbed — session can remain poisoned after Wire overflow**

`scrubMessagePart` unconditionally returns text parts unchanged (`return part, false`) with no size check. If a Wire overflow was caused by a very large text blob delivered as a `chat.MessagePartTypeText` part inside `MultiContent` (rather than the top-level `Content` string), the scrub will leave it intact, and the session store will continue to hold the oversized payload. The session-poisoning problem the PR aims to fix persists for this message shape.

The PR description's stated intent ("Each media part becomes a short text placeholder") suggests only *media* parts were in scope, but the feature claim is broader — "recover oversized user message after wire/media overflow" — and the `Content` size threshold does cover text in the `Content` field. A large text-in-MultiContent blob is a realistic input from API clients or tool responses.

**Suggested fix**: Apply the same size threshold to `MessagePartTypeText` parts inside `MultiContent`:

```go
case chat.MessagePartTypeText:
    if int64(len(part.Text)) > maxScrubbedTextBytes {
        return chat.MessagePart{
            Type: chat.MessagePartTypeText,
            Text: oversizedTextPlaceholder(int64(len(part.Text))),
        }, true
    }
    return part, false
```

docker-agent (commented):

2 medium-severity issues found in the overflow recovery implementation.

pkg/runtime/loop_steps.go
trungutt: Right — the scrub being inside the `r.sessionCompaction` gate is a real regression for users with auto-compaction disabled, and it's the exact session-poisoning bug this PR claims to fix.

Fixed in the latest force-push: the scrub now runs for `OverflowKindWire`/`OverflowKindMedia` independently of `r.sessionCompaction`. The token-compaction path is unchanged (still gated by the config and the attempt cap, since it requires both to make sense). Added a regression test (`TestHandleStreamError_WireOverflowScrubsEvenWithCompactionDisabled`) that pins this contract.

pkg/runtime/overflow_recovery.go
trungutt: Agreed — a giant text blob delivered as a `MessagePartTypeText` inside `MultiContent` is the same poisoning shape as one delivered as the top-level `Content`, and the scrub should catch both.

Fixed in the latest force-push: `scrubMessagePart` now applies the same `maxScrubbedTextBytes` threshold to text parts. Small text parts still pass through unchanged so we don't lose user intent on normal-sized messages. Added `TestScrubMessage_OversizedTextPart_Replaced` that exercises a mixed multi-content message (small text + oversized text in the same MultiContent) and verifies the small part survives while the oversized one becomes a placeholder.

Also renamed `report.mediaReplaced` → `report.partsReplaced` since the counter now covers both media parts and oversized text parts.

pkg/runtime/loop_steps.go
docker-agent: Perfect — thanks for the quick turnaround! I verified the fix: the hygiene scrub (lines 196–207) now runs for Wire/Media overflows independently of the `r.sessionCompaction` gate, so the session-poisoning regression is closed even when auto-compaction is disabled. The token-overflow path remains correctly gated by both the config and the attempt cap (lines 182–194).

The regression test pins the contract cleanly. This resolves the issue.

pkg/runtime/overflow_recovery.go
docker-agent: Perfect — the fix closes the gap completely. Applying the same threshold to text parts inside `MultiContent` ensures session poisoning can't sneak through via the multi-part message shape. Renaming `mediaReplaced` → `partsReplaced` makes the metric honest now that it counts both media and oversized text.

The new test case (mixed small + oversized text in the same `MultiContent`) is exactly the right scenario to lock in — it verifies selective scrubbing preserves user intent for normal-sized parts while still protecting the session store from the poison.

pkg/runtime/overflow_recovery.go
```diff
@@ -0,0 +1,337 @@
...
+		return errors.New("recover-from-oversize: rewritten message not found after scrub")
+	}
+	if err := r.sessionStore.UpdateMessage(ctx, messageID, persisted); err != nil {
+		// Logged but non-fatal — the in-memory session is still
+		// fixed for the current process, which is the main thing.
+		// On next restart the session will reload the old row and
+		// fail the same way, so this is worth alerting on.
+		slog.WarnContext(ctx, "Failed to persist scrubbed user message",
+			"session_id", sess.ID, "message_id", messageID, "error", err)
+		return err
```
docker-agent: **[MEDIUM] `emitScrubNotice` is skipped when `UpdateMessage` fails — user not notified of in-memory rewrite**

When `r.sessionStore.UpdateMessage` returns an error, the function logs a warning and returns `err` immediately (line 298). The `emitScrubNotice` call on line 300 is never reached.

**The problem**: `RewriteLatestUserMessage` already mutated the in-memory session *before* this point — the scrub happened. But the user receives no Warning event, so from their perspective the message was silently rewritten without any explanation. They see only the original "message too large" error.

The comment on this path even says *"the in-memory session is still fixed for the current process, which is the main thing"* — acknowledging the fix occurred — yet the notification is skipped.

**Suggested fix**: Call `emitScrubNotice(events, report)` before returning the error so the user is always informed when the in-memory rewrite succeeded, regardless of persistence outcome:

```go
if err := r.sessionStore.UpdateMessage(ctx, messageID, persisted); err != nil {
    slog.WarnContext(ctx, "Failed to persist scrubbed user message", ...)
    emitScrubNotice(events, report)  // rewrite happened; tell the user
    return err
}
```

pkg/session/session.go
```diff
@@ -431,6 +431,46 @@ func (s *Session) ApplyCompaction(inputTokens, outputTokens int64, item Item) {
...
+// callers to persist the rewrite via [Store.UpdateMessage]) together
+// with ok=true. Returns (0, false) when the session has no user
+// messages, when rewrite returns false, or when the latest user
+// message has not been persisted yet (ID is zero, e.g. a synthesised
+// implicit message).
+//
+// This is the runtime's hook for in-place message hygiene after a
+// failure that would otherwise poison the session — see the wire/
+// media overflow recovery in pkg/runtime.
+func (s *Session) RewriteLatestUserMessage(rewrite func(chat.Message) (chat.Message, bool)) (int64, bool) {
```
docker-agent: **[MEDIUM] `RewriteLatestUserMessage` silently skips sub-session items — overflow in a sub-session escapes scrubbing**

`RewriteLatestUserMessage` iterates `s.Messages` backward and skips any item where `!item.IsMessage()`. Sub-session items (`Item.SubSession`) are not messages, so they are skipped entirely — the function never recurses into them.

If an overflow-triggering user message happens to be stored inside a sub-session (e.g., an agent that routes user turns through a sub-session), `RewriteLatestUserMessage` returns `(0, false)` without touching it, the session store is not updated, and the session remains poisoned — exactly the bug this PR is meant to fix.

**Is this reachable?** The current runtime appears to store interactive user messages at the top level, so the common case is fine. But the function has no documentation of this assumption, and a caller that creates a sub-session before the user turn would silently regress. A guard or doc comment would prevent future surprises:

```go
// NOTE: this does not recurse into sub-session items. Callers must ensure
// the target user message is stored at the top level of s.Messages.
func (s *Session) RewriteLatestUserMessage(...) (int64, bool) {
```

docker-agent (commented):

Two medium-severity findings in the new overflow-recovery code. The core approach — scrubbing oversized messages in-place, persisting the rewrite, and emitting a Warning event — is sound and the two learned patterns (wire/media scrub independence from compaction config; top-level Content vs MultiContent text scrubbing) are both correctly implemented.

pkg/runtime/overflow_recovery.go
trungutt: Good catch — the in-memory rewrite has already happened by the time we reach `UpdateMessage`, so withholding the notification on a persistence failure leaves the user with no explanation for the silently rewritten conversation entry.

Fixed in the latest force-push: `emitScrubNotice` now fires regardless of the store outcome, with a comment explaining why (the visible session state for this process is fixed, even when the next-restart state isn't). The persistence error still propagates so the caller can log it. Added `TestRecoverFromOversizedTurn_NotifiesUserEvenWhenPersistFails` that wraps an `InMemorySessionStore` with a UpdateMessage-returns-error decorator and pins both contracts: in-memory message rewritten, Warning event emitted.

pkg/session/session.go
trungutt: Fair point — silent non-recursion is a footgun for future callers even if no current caller hits it.

Documented in the latest force-push: the doc comment now explicitly states that sub-session items are skipped, that callers routing user turns through a sub-session must rewrite the target message on that sub-session directly, and the rationale (sub-sessions own their own conversation transcript; the parent should not reach into them outside their store). Left the behaviour as-is for now since the reviewer confirmed it isn't reachable today; the contract is now explicit so a future caller can't be surprised by it.

pkg/runtime/overflow_recovery.go
docker-agent: Excellent fix — the notification now correctly reflects the actual runtime state. Moving `emitScrubNotice` outside the success path ensures users always see the Warning event when the in-memory session is rewritten, regardless of whether the persistence layer cooperates.

The new test name (`TestRecoverFromOversizedTurn_NotifiesUserEvenWhenPersistFails`) documents the exact contract you're pinning, and the decorator pattern for injecting the store failure keeps the test implementation clean. Great work on the comment explaining the "visible session state for this process is fixed" rationale — future readers will immediately understand why the notification fires unconditionally.

pkg/session/session.go
docker-agent: Perfect — the updated doc comment is clear and explicit about the skip behavior, caller responsibilities, and the ownership rationale. Future callers won't be surprised by the non-recursion, and the contract is now part of the function's interface rather than an implicit assumption. Thanks for documenting it!

docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed the overflow recovery changes across `pkg/runtime/overflow_recovery.go`, `pkg/runtime/loop_steps.go`, `pkg/session/session.go`, and the accompanying tests.

**What was checked:**
- Correctness of `findPersistedMessage` when sub-sessions are present — `GetAllMessages()` appends top-level messages before sub-session messages, so even in a hypothetical ID-collision scenario the first match is always the correctly-rewritten top-level message ✅
- Notification ordering (`emitScrubNotice` fires unconditionally after in-memory rewrite, regardless of persistence outcome) — correctly implemented per design intent ✅
- `UpdateMessage` call site — called with the post-mutation message content, `messageID == 0` guard handles un-persisted synthesised messages ✅
- Store-write failure path — non-fatal, logged with `slog.WarnContext`, and user is still notified via the scrub notice ✅
- Test coverage — happy path, persist-fails path, and no-user-message path all covered ✅

No confirmed or likely bugs found. The implementation correctly addresses the session-poisoning scenario described in the PR description.

pkg/runtime/overflow_recovery.go
```diff
@@ -0,0 +1,348 @@
...
+	return nil
+}
+
+// findPersistedMessage walks sess.Messages under no lock and returns
+// the *session.Message matching the supplied ID, or nil if not found.
+// Used by recoverFromOversizedTurn to hand the persistence layer the
+// same object that was just mutated in place — race-safe because the
+// caller mutates and reads under the same goroutine within
+// handleStreamError.
+func findPersistedMessage(sess *session.Session, id int64) *session.Message {
```
docker-agent: **[HIGH] `findPersistedMessage` may always return nil — SQLite-backed sessions never get the in-memory `Message.ID` populated**

`findPersistedMessage` searches `sess.GetAllMessages()` for a message whose `ID` matches the one returned by `RewriteLatestUserMessage`. This works when `Message.ID` is populated in the in-memory session, but it is not for SQLite-backed sessions:

1. `SQLiteSessionStore.AddMessage` inserts the row and returns `LastInsertId()`, but it does **not** call `session.AddMessage` itself — the runtime's `PersistenceObserver` calls `store.AddMessage` and discards the returned ID without writing it back into the in-memory `session.Message.ID`.
2. `loadSessionItems` (the path used on reload from disk) reconstructs `Message` structs without scanning the `id` column, so `Message.ID` stays `0` for every row loaded from SQLite.

Consequence: For any SQLite-backed session, `item.Message.ID` is `0` in memory. `RewriteLatestUserMessage` therefore returns `(0, true)`. The guard `if messageID == 0 || r.sessionStore == nil` fires, `emitScrubNotice` is still called (good), but `findPersistedMessage` / `UpdateMessage` are skipped — so the SQLite row is **never updated**. On the next session load the original oversized payload reloads from disk and the overflow trips again.

The existing test (`TestRecoverFromOversizedTurn_NotifiesUserEvenWhenPersistFails`) explicitly sets `stored.ID = persistedID` before `sess.AddMessage(stored)` to hit the `UpdateMessage` path — this pattern is not followed in production code paths (e.g. `PersistenceObserver.UserMessageEvent`).

**Fix options**:
- Have `SQLiteSessionStore.AddMessage` call `msg.ID = id` on the caller's pointer (mirroring `InMemorySessionStore`), **or**
- Have `loadSessionItems` scan and populate the `id` column into `Message.ID`, **or**
- Rewrite `findPersistedMessage` to use `RewriteLatestUserMessage`'s returned ID directly (skipping the search entirely), since `RewriteLatestUserMessage` already has the pointer and can expose the rewritten message object without a second scan.

pkg/session/session.go
```diff
@@ -431,6 +431,54 @@ func (s *Session) ApplyCompaction(inputTokens, outputTokens int64, item Item) {
...
+// it with the returned value. The slot is found and updated under
+// s.mu so concurrent readers (snapshotItems, persistence) cannot
+// observe a torn state.
+//
+// rewrite is called at most once. If it returns false the message is
+// left unchanged.
+//
+// Returns the database ID of the rewritten Item.Message (used by
+// callers to persist the rewrite via [Store.UpdateMessage]) together
+// with ok=true. Returns (0, false) when the session has no user
```
docker-agent: **[LOW] Docstring contradicts implementation: `RewriteLatestUserMessage` returns `(0, true)` for unpersisted messages, not `(0, false)`**

The doc comment says:

> Returns (0, false) when … the latest user message has not been persisted yet (ID is zero, e.g. a synthesised implicit message).

But the implementation returns `(item.Message.ID, true)` unconditionally after a successful rewrite — which is `(0, true)` when `ID == 0`.

The current caller (`recoverFromOversizedTurn`) handles this correctly via the `if messageID == 0` guard. However, a future caller who trusts the docstring might check only `ok == false` to detect "nothing happened" and miss the case where an unpersisted message was rewritten in memory.

The docstring should be corrected to match the actual contract: `(0, true)` when the message was rewritten but had no persisted ID.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

**1 high-severity** and **1 low-severity** finding in the new overflow recovery code.

pkg/runtime/overflow_recovery.go
trungutt: Verified the claim — the bug reproduces in production for both store types because the persistence ID is never written back to the in-memory `Message.ID` (the observer discards the return value from `Store.AddMessage`), and even on reload from disk `loadSessionItems` doesn't `SELECT id`, so `Message.ID` stays 0 forever.

Fixed in the latest force-push with two surgical changes. Both are needed; either alone leaves a gap (writes propagate via PersistenceObserver, reads via loadSessionItems):

1. **`pkg/session/store.go` — `loadSessionItems`**: the SELECT now includes `id` and populates `Message.ID` on the loaded row. Pinned by `TestSQLiteSessionStore_LoadSessionItems_PopulatesMessageID`.
2. **`pkg/runtime/persistence_observer.go` — `UserMessageEvent`**: after `Store.AddMessage` returns, the observer calls a new `Session.AssignLatestUserMessageID(id)` so the in-memory message picks up the persistence ID. Pinned by `TestPersistenceObserver_UserMessageEvent_PropagatesID` (SQLite-backed, mirrors the production runtime+store flow) and `TestPersistenceObserver_UserMessageEvent_AddMessageFailure_DoesNotAssignID` for the failure path.
3. **`pkg/session/session.go` — `AssignLatestUserMessageID`**: new helper used by the observer. Race-safe (under `s.mu`), refuses to overwrite a non-zero ID so two concurrent writers can't clobber each other. Pinned by `TestSession_AssignLatestUserMessageID` (covers happy path, zero-ID no-op, overwrite refusal, no-user-message, empty session).

End-to-end coverage: `TestRecoverFromOversizedTurn_EndToEndPersistsScrub` walks the full pipeline (PersistenceObserver writes via SQLite, scrub fires, reload from DB) and asserts the reloaded row reflects the scrubbed content with the same persistence ID — which is the exact regression the reviewer described.

Left `findPersistedMessage` and the existing scrub structure intact since the now-populated `Message.ID` makes it work correctly without further refactoring. The reviewer's option 3 (return the rewritten message pointer directly) is a code-simplification that can land as a follow-up if desired.

pkg/session/session.go
trungutt: Good catch — the docstring did claim `(0, false)` for unpersisted messages while the implementation returned `(0, true)`. Fixed in the latest force-push: the doc now says "`Returns the database ID of the rewritten Item.Message and ok=true on a successful rewrite. The returned ID is zero when the rewritten message has not been persisted yet ...; callers that intend to mirror the rewrite to the store via Store.UpdateMessage must check for a non-zero ID before doing so. Returns (0, false) when the session has no user messages or when rewrite returns false.`"

Note: with the related fix in this same force-push (PersistenceObserver now propagates the persistence ID back to the in-memory session, loadSessionItems now reads the `id` column), the `ID == 0` case shrinks to genuinely-unpersisted messages — the production runtime/store flow no longer hits it for real user messages, which was the regression the other reviewer flagged. The docstring contract is now both correct and unambiguous about the rare remaining case.

docker-agent: Excellent fix! The docstring now accurately describes the contract: `(0, true)` when the message was successfully rewritten but had no persisted ID, and `(0, false)` only when nothing happened (no user messages or rewrite declined).

Thanks for the context on the broader fix — propagating persistence IDs back from the observer to the in-memory session means the `ID == 0` edge case is now genuinely rare (only unpersisted synthesized messages), which makes the API much clearer for future callers.

pkg/runtime/overflow_recovery.go
docker-agent: Excellent fix — verified the two-pronged approach covers both write (observer propagates ID back to session) and read (loadSessionItems scans the id column on reload) paths, and the test coverage pins all entry points and failure modes as described.

The decision to preserve findPersistedMessage rather than refactor to option 3 is the right call for this PR: the now-populated Message.ID makes the search work correctly, and the "return the pointer directly" simplification can land as a follow-up without blocking the regression fix. Keeping the scope surgical reduces rebase risk and makes the history easier to bisect later.

One thing I'll note for the future: AssignLatestUserMessageID refuses to overwrite a non-zero ID (the "two concurrent writers can't clobber each other" guard), which is great for safety, but that means if two observers ever race on the same AddMessage return value, one silently loses. Not a problem in the current single-observer setup, but worth remembering if the observer wiring changes. The silent false return is appropriate for now — logging would just be noise.

Marking resolved. Thanks for the thorough fix and test coverage!

docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed the overflow recovery implementation across `pkg/runtime/overflow_recovery.go`, `pkg/runtime/loop_steps.go`, `pkg/session/session.go`, and `pkg/session/store.go`.

**Summary of analysis:**

The implementation is correct and well-structured. Three hypotheses were investigated and all were dismissed after verification:

1. **Token overflow gap** — The deliberate decision to only run `recoverFromOversizedTurn` for `OverflowKindWire`/`OverflowKindMedia` (not `OverflowKindTokens`) is correct by design. Token overflow is caused by accumulated conversation history, not a single oversized message; scrubbing only the latest user message would not recover it. Auto-compaction handles the token case differently and correctly.

2. **Sub-session ID collision in `findPersistedMessage`** — SQLite auto-increment guarantees globally unique message IDs across the table. Additionally, `GetAllMessages` returns top-level messages before sub-session messages, so `findPersistedMessage` would always find the correct top-level match first even in the theoretical event of an ID collision.

3. **Misleading log fields for MultiContent scrubbing** — `parts_replaced > 0` already communicates to operators that content was scrubbed when only MultiContent parts are oversized; the `text_replaced`/`original_text_bytes` fields are supplementary precision, not the primary scrub indicator.

**Notable design positives:**
- `RewriteLatestUserMessage` correctly holds `s.mu.Lock()` for the full rewrite, preventing torn state
- The `messageID == 0` guard correctly skips the store write for unpersisted messages
- `emitScrubNotice` is emitted unconditionally after a successful in-memory rewrite, even on persistence failure — this is the right user-observable behavior
- `scrubMessagePart` passes unknown part types through unchanged rather than silently dropping them

pkg/runtime/overflow_recovery.go
trungutt: After more thought, I'm reverting the persistence-side changes from this PR and keeping it scoped to its original goal: in-memory message hygiene after a wire/media overflow.

The ID propagation gap you identified is a real bug, but it's an **independent infrastructure issue**: `Message.ID` not round-tripping through `Store.AddMessage` / `loadSessionItems` predates overflow handling and affects anything that needs `Store.UpdateMessage` against a freshly-added or freshly-loaded message (future compaction-by-id work, message editing, etc.). Folding the fix into this PR conflated concerns and bloated the diff (a ~1200-line change for what is conceptually "rewrite a message after overflow").

This PR now scopes itself to what it can cleanly own:

- **In-memory scrub** of the latest user message after wire/media overflow.
- **Same-process recovery** — the conversation can continue immediately, which is the original user-reported regression (paste oversized → 413 → shortening keeps failing within the same chat session).

What is explicitly out of scope and tracked as a follow-up:

- Mirroring the scrub to the session store so it survives a docker-agent restart. That needs the ID propagation fix you described, which deserves its own focused PR (one that can be debated on its own merits — propagate the ID, use position-based updates, change the API, etc.).

The `recoverFromOversizedTurn` doc comment now documents this limitation explicitly so a future reader is not surprised.

Net diff after trim: ~1200 → ~850 lines, 9 files → 5 files. The PR now does exactly one thing.

docker-agent: I appreciate the thoughtful scoping decision — separating the in-memory recovery fix from the persistence round-trip issue is the right call for reviewability and future maintainability. The trimmed PR is much easier to reason about.

That said, **the original finding stands**: without the persistence-side fix, the overflow scrub does **not** survive a `docker-agent` restart, and users will hit the 413 loop again after reload. The in-memory recovery unblocks the same-process flow (which is valuable!), but the "overflow trips again on next session load" consequence I described is still live. The doc comment addition helps set expectations, but it's worth confirming that the follow-up PR is prioritized — otherwise this fix only papers over the issue for long-running agents, and anyone who restarts (deployments, crashes, etc.) re-encounters the original user-reported bug.

If the follow-up is tracked and queued, I'm fine with the scoped approach here. If it's not yet tracked, please file an issue linking back to this thread so the persistence gap doesn't get lost. The infrastructure fix you're deferring is load-bearing for the durability half of the user-facing regression.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26148572963).

PR #2824, opened 2026-05-19
Don't notify on click if the agent didn't change

rumpl:


pkg/tui/handlers.go
```diff
@@ -314,6 +314,10 @@ func copyToClipboard(text, successMsg string) tea.Cmd {
 // --- Agent management ---
 
 func (m *appModel) handleSwitchAgent(agentName string) (tea.Model, tea.Cmd) {
+	if agentName != "" && m.sessionState != nil && agentName == m.sessionState.CurrentAgentName() {
```
docker-agent: **[LOW] Subtle guard hole: empty `agentName` bypasses no-op check**

The new early-return guard reads:
```go
if agentName != "" && m.sessionState != nil && agentName == m.sessionState.CurrentAgentName() {
    return m, nil
}
```
When `agentName == ""` the guard short-circuits on the first condition and falls through to `m.application.SwitchAgent("")` → `m.sessionState.SetCurrentAgentName("")`, silently clearing the session's current agent name.

No current call site passes `""` (the `handleAgentCommand` caller already guards with `cmd.Agent != ""`), so this is not an active bug today. However, the `agentName != ""` condition creates an asymmetry: a no-op switch *to the empty string* is not treated as a no-op. If the condition were simply removed, the equality comparison `agentName == m.sessionState.CurrentAgentName()` would still fire correctly whenever the current agent name is also `""`.

Consider either:
- Removing the `agentName != ""` prefix (simplifies the guard), or
- Returning early explicitly for `agentName == ""` with an error or silent no-op

docker-agent (commented):
### Assessment: 🟢 APPROVE

The no-op guard correctly prevents a redundant agent-switch notification when the user clicks the already-active agent. The logic is simple and the new test covers the primary path well.

One minor guard asymmetry noted (see inline comment) — not a blocking issue.

dgageot (approved):

PR #2825, opened 2026-05-19
Treat wezterm as a terminal that knows how to handle shift+enter

rumpl:
Tested with wezterm 20240203-110809-5046fc22 (latest) on a Mac and shift+enter works

### DIFF
--- a/pkg/tui/components/editor/editor.go
+++ b/pkg/tui/components/editor/editor.go
@@ -27,6 +27,7 @@ import (
 	"github.com/docker/docker-agent/pkg/tui/components/editor/completions"
 	"github.com/docker/docker-agent/pkg/tui/core"
 	"github.com/docker/docker-agent/pkg/tui/core/layout"
+	"github.com/docker/docker-agent/pkg/tui/internal/termfeatures"
 	"github.com/docker/docker-agent/pkg/tui/messages"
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change is small, focused, and correct. WezTerm detection uses four complementary environment-variable heuristics (`TERM_PROGRAM`, `WEZTERM_PANE`, `WEZTERM_UNIX_SOCKET`, and `TERM` substring), all lowercased before comparison, making the check robust. The function is wired into both `tui.go` and `editor.go` at the right call sites and correctly OR-combined with the Kitty keyboard-enhancement flags. The nil-guard is a safe defensive pattern, and the test suite covers all four positive paths plus the nil and false-positive cases. No bugs found in the changed lines.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes correctly add WezTerm terminal detection for keyboard enhancement support, wiring it consistently into both the editor component and the top-level TUI model. The new `SupportsModifiedEnter` function is clean, well-tested, and the state synchronization between `tui.go` and `editor.go` is coherent.

No CONFIRMED or LIKELY bugs were found. All hypotheses raised during analysis were DISMISSED after verification.

dgageot (approved):

PR #2827, opened 2026-05-19
Add .cache to .gitignore

rumpl:
The taskfile sets it as the cache directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 bin
 dist
 .task
+/.cache/
 .DS_Store
 evals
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change adds `/.cache/` to `.gitignore`. The pattern is correct: the leading `/` anchors it to the repository root (avoiding unintended matches in subdirectories) and the trailing `/` restricts matching to directories only. No issues found.

dgageot (approved):

PR #2828, opened 2026-05-20
Fix mcp oauth

rumpl:


pkg/tools/mcp/oauth_helpers.go
```diff
@@ -33,6 +33,13 @@ import (
 // because httptest.NewServer binds to 127.0.0.1.
 var oauthHTTPClient = httpclient.NewSafeClient(30*time.Second, false)
 
+func oauthHTTPClientForAllowPrivateIPs(allowPrivateIPs bool) *http.Client {
+	if allowPrivateIPs {
+		return &http.Client{Timeout: 30 * time.Second}
```
docker-agent, line 38: **[HIGH] SSRF via attacker-controlled `resource=` URL when `allowPrivateIPs=true`**

The new `oauthHTTPClientForAllowPrivateIPs` function returns a plain `&http.Client{Timeout: 30*time.Second}` (no SSRF protection) when `allowPrivateIPs` is `true`. However, this client is used in the OAuth flow to fetch `resourceURL` extracted from the MCP server's `WWW-Authenticate: Bearer resource="..."` header — an entirely attacker-controlled value:

```go
// oauth.go
resourceURL := cmp.Or(resourceMetadataFromWWWAuth(wwwAuth), authServer+"/.well-known/oauth-protected-resource")
resourceReq, _ := http.NewRequestWithContext(ctx, http.MethodGet, resourceURL, http.NoBody)
resp, err := t.oauthClient().Do(resourceReq)  // plain client — no SSRF protection
```

A malicious MCP server can respond with `WWW-Authenticate: Bearer resource="http://169.254.169.254/latest/meta-data/"` and the code will issue a GET to the AWS IMDS (or any other internal endpoint) using the unprotected client.

**Intent mismatch**: The existing comment above `oauthHTTPClient` explicitly warns that hostile MCP servers could redirect OAuth metadata fetches to internal addresses — that protection is now bypassed when `allowPrivateIPs=true`. The flag was intended to allow the MCP *endpoint itself* to be on a private IP, not to remove SSRF protection for attacker-controlled OAuth metadata URLs.

**Suggested fix**: Keep the SSRF-safe `oauthHTTPClient` for all OAuth metadata fetches regardless of `allowPrivateIPs`. Only the connection to the MCP endpoint itself (the `base` transport in `createHTTPClient`) needs to honor the private-IP setting.

pkg/config/mcps.go
```diff
@@ -86,6 +86,9 @@ func applyMCPDefaults(ts, def *latest.Toolset) {
 	if ts.Defer.IsEmpty() {
 		ts.Defer = def.Defer
 	}
+	if !ts.AllowPrivateIPs {
```
docker-agent: **[LOW] `allow_private_ips: false` in a toolset cannot override `true` from its MCP definition**

Because `AllowPrivateIPs` is a plain `bool` and Go's zero value for `bool` is `false`, this check treats an explicit `allow_private_ips: false` in a toolset identically to "not set at all". A toolset that references a definition with `allow_private_ips: true` has no way to opt out:

```yaml
mcps:
  my-server:
    allow_private_ips: true  # set in definition

toolsets:
  - ref: my-server
    allow_private_ips: false  # silently ignored — inherits 'true' from definition
```

For a security-sensitive permission this is surprising. Consider using `*bool` (pointer) or a tri-state type to distinguish "explicitly false" from "unset", consistent with how other optional fields handle optionality.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR introduces OAuth support for remote MCP toolsets, PKCE flow, token refresh, and a new `allow_private_ips` config flag. One high-severity SSRF vulnerability was found in the new `allowPrivateIPs` path, plus one low-severity config-inheritance issue.

| Severity | Finding | File |
|----------|---------|------|
| 🔴 HIGH | SSRF via attacker-controlled `resource=` URL when `allowPrivateIPs=true` | `pkg/tools/mcp/oauth_helpers.go:38` |
| 🟡 LOW | `allow_private_ips: false` cannot override definition's `true` (zero-value ambiguity) | `pkg/config/mcps.go:89` |

pkg/tools/mcp/oauth_helpers.go
rumpl: Well yes, that's the feature

docker-agent: Thanks for clarifying the intent. I misunderstood the scope of `allowPrivateIPs` — I thought it was meant only for the MCP endpoint itself, not the OAuth metadata flow. If the design is that setting this flag disables SSRF protection across the board (MCP + OAuth), then the implementation is correct as-is.

For future readers: when `allow_private_ips: true` is set on a remote MCP toolset, **all** HTTP requests related to that server (the MCP endpoint, OAuth metadata discovery, token endpoints, etc.) are allowed to reach private IPs. Use this only in trusted environments where internal services are the legitimate target.

dgageot (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26224446060).

PR #2829, opened 2026-05-20
server: load initial source asynchronously so /api/ping is responsive immediately

trungutt:
### What this PR does

`newSourceLoader` calls `sl.load(ctx)` **synchronously** in the constructor. `NewSessionManager` waits on every source's constructor, and `cmd/root/api.go:runAPICommand` waits on `NewSessionManager` before calling `s.Serve(ctx, ln)`. So if `Source.Read` does network I/O — an OCI pull, an HTTPS fetch, a slow registry, etc. — nothing else in the agent runs until that finishes, including the `Accept()` loop that would answer `/api/ping` or `/health`.

That means an embedding client which probes `/api/ping` with a bounded timeout after spawning `docker-agent serve api …` may give up and kill the process even though the binary is perfectly healthy — it just hasn't reached `s.Serve` yet because the initial source load is still in flight.

```
   BEFORE                                       AFTER
   ──────                                       ─────
   newListener()       ◄── listener visible     newListener()
   NewSessionManager                            NewSessionManager
     newSourceLoader                              newSourceLoader
       sl.load(ctx)    ◄── 🛑 blocks here          go sl.load(ctx)   ◄── async
                          (network I/O,             return immediately
                           cold cache, etc.)
                                               s.Serve(ctx, ln)   ◄── Accept()
   s.Serve(ctx, ln)    ◄── Accept() starts                              starts NOW
                          here, possibly                                /api/ping is
                          after the client's                            answerable in ms
                          probe timeout

   /api/ping → timeout                          /api/ping → 200 OK
   /api/agents → ready immediately              /api/agents → blocks in Read
                                                  until first load completes
                                                  (or ctx is cancelled)
```

### Implementation

- `sourceLoader` gains a `ready chan struct{}` that is closed once the first load completes (whether it succeeded, failed, or was cancelled).
- The initial load runs in a goroutine inside `newSourceLoader`, so the constructor returns immediately.
- `Read(ctx)` now blocks on `<-sl.ready` or `<-ctx.Done()` before returning the cached snapshot. Callers that don't want to wait can pass a short-deadline context.

This preserves the previous Read contract — when `Read` returns, the data is either valid or there's an error — so existing consumers (`getAgents`, `getAgentConfig`, `createSession`, etc.) do not need any changes.

### DIFF
--- a/pkg/server/source_loader.go
+++ b/pkg/server/source_loader.go
@@ -13,12 +13,19 @@ type sourceLoader struct {
 	inner           config.Source
 	refreshInterval time.Duration
 
+	// ready is closed when the first load attempt has completed (whether it
+	// produced data, an error, or was cancelled). Read blocks on this so
+	// callers never observe the pre-load (nil, nil) state, preserving the
+	// contract of the previous synchronous-constructor implementation.
+	ready chan struct{}
+
 	mu   sync.RWMutex
 	data []byte
 	err  error
 }
 
 // NewSourceLoader creates a new source loader that caches and periodically refreshes a config source.
+// The initial load runs in a goroutine; callers of Read block until the first load completes.
 func NewSourceLoader(ctx context.Context, inner config.Source, refreshInterval time.Duration) *sourceLoader {
 	return newSourceLoader(ctx, inner, refreshInterval)
 }
@@ -27,9 +34,19 @@ func newSourceLoader(ctx context.Context, inner config.Source, refreshInterval t
 	sl := &sourceLoader{
 		inner:           inner,
 		refreshInterval: refreshInterval,
+		ready:           make(chan struct{}),
 	}
 
-	sl.load(ctx)
+	// Run the initial load asynchronously so the constructor returns
+	// immediately. This lets the HTTP server start serving non-source
+	// endpoints (e.g. /api/ping, /health) while a potentially slow OCI
+	// pull or HTTPS GET is still in flight. Endpoints that need the
+	// source content (getAgents, createSession, ...) block in Read
+	// until this load finishes.
+	go func() {
+		defer close(sl.ready)
+		sl.load(ctx)
+	}()
 
 	if refreshInterval > 0 {
 		go sl.refreshLoop(ctx)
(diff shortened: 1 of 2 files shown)


pkg/server/source_loader.go
```diff
@@ -46,7 +63,16 @@ func (sl *sourceLoader) ParentDir() string {
 	return sl.inner.ParentDir()
 }
 
-func (sl *sourceLoader) Read(_ context.Context) ([]byte, error) {
+// Read returns the latest cached data, blocking until the first load attempt
+// has completed or ctx is cancelled. Once the first load has finished, Read
+// returns the cached data/err snapshot without further blocking; the
+// background refreshLoop is responsible for keeping the snapshot fresh.
+func (sl *sourceLoader) Read(ctx context.Context) ([]byte, error) {
+	select {
```
docker-agent, line 71: **[MEDIUM] Non-deterministic `Read` result when context is already cancelled after initial load**

When the `ready` channel is already closed (initial load complete) **and** the caller's `ctx` is already done, both `select` cases are simultaneously ready and Go picks one at random (~50/50). This means a caller with a cancelled request context can receive `(nil, context.Canceled)` even though valid cached data sits in `sl.data`.

The previous implementation declared `Read(_ context.Context)` — the context was fully discarded, so cancelled contexts never caused data-access failures. This PR changes that contract: a request context cancelled just after the initial load completes (e.g. a slow client dropping its HTTP connection) will now non-deterministically return an error instead of the cached data.

**Minimal fix** — check `ready` first with a non-blocking select, and only then honour context cancellation:

```go
func (sl *sourceLoader) Read(ctx context.Context) ([]byte, error) {
    select {
    case <-sl.ready:
        // fast path: already loaded, fall through regardless of ctx state
    default:
        // initial load still in flight: wait for it or ctx cancellation
        select {
        case <-sl.ready:
        case <-ctx.Done():
            return nil, ctx.Err()
        }
    }
    sl.mu.RLock()
    defer sl.mu.RUnlock()
    return sl.data, sl.err
}
```

This preserves the blocking-until-loaded contract while never sacrificing already-cached data to a racing cancellation.

docker-agent (commented):

PR #2830, opened 2026-05-20
perf(tui): only render visible session rows in /sessions dialog

dgageot:
The `/sessions` dialog was visibly slow when scrolling or moving the
selection with arrow keys, especially with a large session history.

Every render of `sessionBrowserDialog.View()` was pre-rendering **all**
filtered session rows into styled lines (two `lipgloss.Style.Render`
calls plus string concatenation per row), even though the scrollview
only displays a small visible window. With hundreds of sessions, every
keystroke and scroll tick incurred hundreds of style renders before
the scrollview sliced the result down to the viewport.

The scrollview already exposes `ViewWithLines(visible)` for the
virtualised case (used by `picker.go` and `working_dir_picker.go`).
This change uses it to render only `[offset, offset+visibleHeight)`,
making render cost O(visible rows) (~10–20) instead of O(total
sessions).

`filterSessions` also now calls `SetContent(nil, len(filtered))` so
the scrollview's `totalHeight` is always in sync — this fixes a
latent bug where `EnsureLineVisible` (called on Up/Down keypress)
could clamp against a stale `totalHeight` between a filter change and
the next `View()`.

### DIFF
--- a/pkg/tui/dialog/session_browser.go
+++ b/pkg/tui/dialog/session_browser.go
@@ -253,6 +253,9 @@ func (d *sessionBrowserDialog) filterSessions() {
 	if d.selected >= len(d.filtered) {
 		d.selected = max(0, len(d.filtered)-1)
 	}
+	// Keep the scrollview's totalHeight in sync so EnsureLineVisible and the
+	// scrollbar clamp correctly even before View() runs.
+	d.scrollview.SetContent(nil, len(d.filtered))
 	d.scrollview.SetScrollOffset(0)
 }
 
@@ -285,13 +288,6 @@ func (d *sessionBrowserDialog) View() string {
 	dialogWidth, _, contentWidth := d.dialogSize()
 	d.textInput.SetWidth(contentWidth)
 
-	// Build all session lines
-	var allLines []string
-	for i, sess := range d.filtered {
-		allLines = append(allLines, d.renderSession(sess, i == d.selected, contentWidth))
-	}
-
-	// Configure scrollview and let it handle slicing + rendering
 	regionWidth := contentWidth + d.scrollview.ReservedCols()
 	visibleLines := d.scrollview.VisibleHeight()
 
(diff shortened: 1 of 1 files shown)


trungutt (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

**Virtualized session-row rendering** — the O(visible) rendering refactor is correct and well-structured.

**What was checked:**

- **Negative capacity in `make([]string, 0, end-offset)`**: Safe. After `SetContent(nil, total)` + `SetScrollOffset(ScrollOffset())`, the offset is clamped to `[0, max(0, total−height)]`. Since `total > 0` in the else-branch and `end = min(offset+visibleLines, total)`, we always have `end ≥ offset`. No panic path.
- **Selected-row highlight index**: `i == d.selected` uses the absolute index into `d.filtered` on both sides — correct for virtual windows.
- **`syncScrollbar()` desync**: `ViewWithLines` internally calls `syncScrollbar()` after the `offset` variable is captured. Both `SetScrollOffset` (called in `View()`) and `syncScrollbar` use the same clamping formula `max(0, total−height)` with the same already-updated `totalHeight`, so the read-back value is stable. No observable desync in the single-threaded TUI render path.
- **`nil` lines stored in scrollview**: `ViewWithLines` never reads `m.lines`; only the old `View()` path does. Since `ViewWithLines` is always called in the new path, the nil value is harmless.
- **`filterSessions()` double SetContent**: The call in `filterSessions()` primes `totalHeight` for `EnsureLineVisible`; the call in `View()` is idempotent given the same `len(d.filtered)`. The scroll-reset-to-0 in `filterSessions()` is consistent with pre-existing behaviour.
- **`regionWidth` variable**: Still used at line 358 (`NewContent(regionWidth)`). No dead-variable compile error.

The change cleanly delivers the stated O(visible) goal with no regressions found.

PR #2836, opened 2026-05-20
fix: use in-memory token store in tests to avoid OS keychain prompt

dgageot:


pkg/tools/mcp/tokenstore_keyring.go
```diff
@@ -52,7 +53,17 @@ func openKeyring() (keyring.Keyring, error) {
...
 // keyring lazily on first call. Multiple MCP toolsets share its in-memory
 // cache so they don't each trigger a credential prompt on construction.
+//
+// Under `go test` we always return an in-memory store: any test that
+// constructs a real *mcp.Toolset (directly or via the mcpcatalog
+// builtin) would otherwise reach into the real OS keychain on the first
+// outbound HTTP request, popping a macOS password prompt for the
+// `docker-agent-oauth` keychain item on developer machines that have a
+// token from a prior login.
 var defaultStore = sync.OnceValue(func() OAuthTokenStore {
```
docker-agent, line 63: **[NOTABLE] Shared singleton `InMemoryTokenStore` may leak token state between tests**

The `sync.OnceValue` factory runs exactly once per process lifetime. Under `go test`, all tests in the package share the **same** `InMemoryTokenStore` instance returned by `defaultStore()`. Any test that stores a token via `NewKeyringTokenStore()` (which delegates to `defaultStore()`) will leave that token visible to all subsequent tests in the same binary run.

This is an accepted trade-off for the production singleton design (multiple toolsets sharing one store), and most existing tests already use `NewInMemoryTokenStore()` directly for isolation. However, integration tests or future tests that call `NewKeyringTokenStore()` and expect an empty store at start-up will see stale tokens from earlier tests — a subtle source of test flakiness.

Consider documenting this shared-state behaviour at the `defaultStore` declaration, or providing a test-helper that resets the singleton between tests (e.g., via `t.Cleanup` and a package-internal reset function).

docker-agent (commented):

aheritier (approved):

gtardif (approved):

PR #2837, opened 2026-05-20
fix MCP tool name prefix stripping in callTool

dgageot:
Tool calls when using the mcp catalog builtin tool would fail because a prefix is added to each tool in order to avoid conflicts

### DIFF
--- a/pkg/tools/mcp/mcp.go
+++ b/pkg/tools/mcp/mcp.go
@@ -609,8 +609,18 @@ func (ts *Toolset) callTool(ctx context.Context, toolCall tools.ToolCall) (*tool
 		}
 	}
 
+	// Tools() prefixes every exposed tool with `<ts.name>_` when the toolset
+	// has a YAML name (or the catalog id, for mcpcatalog-activated servers).
+	// The remote MCP server doesn't know about that prefix, so we have to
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the root cause: `callTool` was forwarding the prefixed tool name (e.g. `github-official_get_issue`) to the remote MCP server, which only knows the unprefixed name (`get_issue`), causing every catalog tool call to fail with "tool not found".

**Fix analysis:**
- `strings.TrimPrefix(name, ts.name+"_")` is the right primitive — it strips exactly one leading occurrence of `<name>_` and is a no-op when `ts.name` is empty (unnamed toolsets) or the prefix is absent (misrouted calls)
- No off-by-one, no nil-dereference (Go strings are never nil), no double-stripping (TrimPrefix is not iterative)
- The `ts.name != ""` guard avoids any inadvertent behavior for unnamed toolsets

**Test coverage** is thorough — `TestToolsAndCallToolRoundTrip` and `TestCallToolStripsToolsetNamePrefix` together cover: named toolsets, unnamed toolsets, hyphenated catalog IDs, double-prefix tool names, and unprefixed calls forwarded unchanged.

No issues found in the changed lines.

gtardif (approved):

PR #2839, opened 2026-05-20
fix(openai): support Notion and Jira MCP tools with gpt-5 strict mode

dgageot:
## Problem

OpenAI's Responses API in strict mode (used by gpt-5/gpt-5.5 with `Strict: true`) requires `additionalProperties` to be literally `false` on every object schema. Some MCP servers ship tool schemas that violate this:

- **Atlassian remote MCP** (`editJiraIssue`): declares `fields.additionalProperties: {}` (or a richer schema). The request fails with:

  ```
  HTTP 400: Invalid schema for function 'atlassian-remote_editJiraIssue':
  In context=('properties', 'fields', 'additionalProperties'),
  'additionalProperties' is required to be supplied and to be false.
  ```

- **Notion MCP** (e.g. block-creation tools): declares `additionalProperties` as a *schema* (a map describing the dictionary value shape, like `bulleted_list_item`/`numbered_list_item`). Rewriting that to `false` would strip information the model relies on to call the tool correctly.

A previous fix overwrote schema-form `additionalProperties` with `false` everywhere — that fixed Jira but degraded Notion's tool calls.

## Fix

Decide strict mode per tool:

1. `ConvertParametersToSchema` now returns `(schema, strict, err)`.
2. A schema is **strict-incompatible** if any node has `additionalProperties` set to a map (schema form) or to `true`. We detect that with a read-only walk before any normalization.
3. **Strict-compatible** schemas go through the existing pipeline (`makeAllRequired` → forces `additionalProperties: false`, all properties required, nullables, etc.).
4. **Strict-incompatible** schemas are preserved (only cosmetic fixups: `ensureTypeFields`, `removeFormatFields`, `fixSchemaArrayItems`).
5. The Responses API tool definition uses `Strict: param.NewOpt(strict)` per tool.

Result:
- Jira's `editJiraIssue` works because we no longer claim strict mode on it.
- Notion's tools work because the inner dictionary schema survives the conversion.
- All other tools keep the existing strict-mode benefits.

## Tests

- `TestIsStrictCompatible` covers strict, `additionalProperties: true`, and schema-form cases.
- `TestConvertParametersToSchema_NotionStylePreservesShape` proves Notion's inner schema is preserved end-to-end.
- `TestMakeAllRequired_JiraEditIssueFields` is a regression for the original Atlassian failure.
- All existing schema/conversion tests updated to the new 3-return signature.

## Validation

- \`go build ./...\` ✅
- \`go vet ./...\` ✅
- \`go test -race -count=1 ./pkg/model/provider/...\` ✅
- \`task lint\` ✅ (0 issues)
- \`go mod tidy --diff\` clean ✅

### DIFF
--- a/pkg/model/provider/openai/client.go
+++ b/pkg/model/provider/openai/client.go
@@ -258,7 +258,7 @@ func (c *Client) CreateChatCompletionStream(
 		slog.DebugContext(ctx, "Adding tools to OpenAI request", "tool_count", len(requestTools))
 		toolsParam := make([]openai.ChatCompletionToolUnionParam, len(requestTools))
 		for i, tool := range requestTools {
-			parameters, err := ConvertParametersToSchema(tool.Parameters)
+			parameters, _, err := ConvertParametersToSchema(tool.Parameters)
 			if err != nil {
 				slog.DebugContext(ctx, "Failed to convert tool parameters to OpenAI schema", "tool_name", tool.Name, "error", err)
 				return nil, err
@@ -397,7 +397,7 @@ func (c *Client) CreateResponseStream(
 		slog.DebugContext(ctx, "Adding tools to OpenAI responses request", "tool_count", len(requestTools))
 		toolsParam := make([]responses.ToolUnionParam, len(requestTools))
 		for i, tool := range requestTools {
-			parameters, err := ConvertParametersToSchema(tool.Parameters)
+			parameters, strict, err := ConvertParametersToSchema(tool.Parameters)
 			if err != nil {
 				slog.DebugContext(ctx, "Failed to convert tool parameters to OpenAI schema", "tool_name", tool.Name, "error", err)
 				return nil, err
@@ -408,8 +408,12 @@ func (c *Client) CreateResponseStream(
 					Name:        tool.Name,
 					Description: param.NewOpt(tool.Description),
 					Parameters:  parameters,
-					Strict:      param.NewOpt(true),
+					Strict:      param.NewOpt(strict),
 				},
+			}
+
+			if !strict {
+				slog.DebugContext(ctx, "Tool not compatible with OpenAI strict mode, falling back", "tool_name", tool.Name)
 			}
 
 			slog.DebugContext(ctx, "Added tool to OpenAI responses request", "tool_name", tool.Name)
--- a/pkg/model/provider/openai/schema.go
+++ b/pkg/model/provider/openai/schema.go
@@ -9,12 +9,14 @@ import (
 	"github.com/docker/docker-agent/pkg/tools"
 )
 
-// ConvertParametersToSchema converts parameters to OpenAI Schema format.
-// It also returns whether the schema is compatible with strict mode.
-// Schemas that declare schema-form additionalProperties (e.g. Notion MCP)
-// are not strict-compatible: rewriting them to additionalProperties: false
-// would lose the dictionary value shape the model needs. The caller should
-// set Strict=false on the tool definition in that case.
+// ConvertParametersToSchema converts parameters to OpenAI Schema format and
+// reports whether the resulting schema is compatible with OpenAI strict mode.
+//
+// The same normalization pipeline runs in both cases — strict-incompatible
+// schemas (e.g. Notion MCP tools that declare schema-form additionalProperties)
+// still need fully-populated `required` arrays for the Chat Completions API,
+// which has no per-tool strict flag. The strict flag is only consumed by the
+// Responses API caller.
 func ConvertParametersToSchema(params any) (shared.FunctionParameters, bool, error) {
 	p, err := tools.SchemaToMap(params)
 	if err != nil {
@@ -22,37 +24,56 @@ func ConvertParametersToSchema(params any) (shared.FunctionParameters, bool, err
 	}
 
 	strict := isStrictCompatible(p)
-
-	if strict {
-		return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(makeAllRequired(p)))), true, nil
-	}
-	return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(p))), false, nil
+	return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(makeAllRequired(p)))), strict, nil
 }
 
 // isStrictCompatible reports whether the schema can use OpenAI strict mode.
 // Strict mode requires every object node to have additionalProperties: false.
 // Schema-form additionalProperties (a map) and additionalProperties: true are
-// both incompatible.
+// both incompatible. The walk stops at the first incompatible node.
 func isStrictCompatible(schema map[string]any) bool {
-	compatible := true
-	walkSchema(schema, func(node map[string]any) {
-		if !compatible {
-			return
-		}
-		v, ok := node["additionalProperties"]
-		if !ok {
-			return
-		}
+	return !hasIncompatibleNode(schema)
+}
+
+func hasIncompatibleNode(node map[string]any) bool {
+	if v, ok := node["additionalProperties"]; ok {
 		switch t := v.(type) {
 		case map[string]any:
-			compatible = false
+			return true
 		case bool:
 			if t {
-				compatible = false
+				return true
 			}
 		}
-	})
-	return compatible
+	}
+
+	if properties, ok := node["properties"].(map[string]any); ok {
+		for _, v := range properties {
+			if sub, ok := v.(map[string]any); ok && hasIncompatibleNode(sub) {
+				return true
+			}
+		}
+	}
+
+	for _, keyword := range []string{"anyOf", "oneOf", "allOf"} {
+		if variants, ok := node[keyword].([]any); ok {
+			for _, v := range variants {
+				if sub, ok := v.(map[string]any); ok && hasIncompatibleNode(sub) {
+					return true
+				}
+			}
+		}
+	}
+
+	if items, ok := node["items"].(map[string]any); ok && hasIncompatibleNode(items) {
+		return true
+	}
+
+	if additionalProps, ok := node["additionalProperties"].(map[string]any); ok && hasIncompatibleNode(additionalProps) {
+		return true
+	}
+
+	return false
 }
 
 // walkSchema calls fn on the given schema node, then recursively walks into
@@ -88,9 +109,11 @@ func walkSchema(schema map[string]any, fn func(map[string]any)) {
 	}
 }
 
-// makeAllRequired enforces OpenAI strict mode: every object property is
-// listed in `required` (newly-required ones are made nullable) and every
-// object node has `additionalProperties: false`.
+// makeAllRequired makes every object property `required` (newly-required ones
+// are made nullable) and ensures every object node has `additionalProperties`
+// set. It runs on every schema regardless of strict-mode compatibility, so
+// schema-form additionalProperties (e.g. Notion's dictionary value shape) is
+// preserved — only missing/true/nil values are forced to `false`.
 func makeAllRequired(schema shared.FunctionParameters) shared.FunctionParameters {
 	if schema == nil {
 		schema = map[string]any{"type": "object", "properties": map[string]any{}}
@@ -114,8 +137,13 @@ func makeAllRequired(schema shared.FunctionParameters) shared.FunctionParameters
 			}
 		}
 
+		// Only force additionalProperties: false when it isn't already a
+		// schema. Schema-form additionalProperties carries information the
+		// model needs (Notion-style dictionaries) and would be lost otherwise.
 		if isObject {
-			node["additionalProperties"] = false
+			if addProps, exists := node["additionalProperties"]; !exists || addProps == nil || addProps == true {
+				node["additionalProperties"] = false
+			}
 		}
 
 		properties, ok := node["properties"].(map[string]any)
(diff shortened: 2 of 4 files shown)


pkg/model/provider/openai/client.go
```diff
@@ -258,7 +258,7 @@ func (c *Client) CreateChatCompletionStream(
 		slog.DebugContext(ctx, "Adding tools to OpenAI request", "tool_count", len(requestTools))
 		toolsParam := make([]openai.ChatCompletionToolUnionParam, len(requestTools))
 		for i, tool := range requestTools {
-			parameters, err := ConvertParametersToSchema(tool.Parameters)
+			parameters, _, err := ConvertParametersToSchema(tool.Parameters)
```
docker-agent, line 261: **[MEDIUM] Non-strict schemas skip `makeAllRequired` normalization for Chat Completions API**

Before this PR, `ConvertParametersToSchema` always ran `makeAllRequired`, so *every* tool schema sent to the Chat Completions API (`CreateChatCompletionStream`) received full strict-mode normalization: `additionalProperties: false` on every object node and all properties listed in `required`.

After this PR, only strict-compatible schemas go through `makeAllRequired`. The `CreateChatCompletionStream` path discards the `strict` return value (`parameters, _, err := ...`) and applies no compensating normalization — so any tool with schema-form `additionalProperties` (e.g. Jira's `editJiraIssue`, Notion block tools) now reaches the Chat Completions API **without** `additionalProperties: false` and **without** a complete `required` array.

This matters whenever an older model (or a non-Responses-API-capable model) is used with these tools: the schema that previously worked may now fail validation or degrade tool-calling quality.

**Possible fix**: In `CreateChatCompletionStream`, apply `makeAllRequired` unconditionally before passing `parameters` to `ChatCompletionFunctionTool`, as was done before this refactor — the Chat Completions API does not take a per-tool `Strict` flag, so it should always receive fully-normalized schemas.

pkg/model/provider/openai/schema.go
```diff
@@ -9,14 +9,50 @@ import (
...
+		return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(makeAllRequired(p)))), true, nil
+	}
+	return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(p))), false, nil
+}
+
+// isStrictCompatible reports whether the schema can use OpenAI strict mode.
+// Strict mode requires every object node to have additionalProperties: false.
+// Schema-form additionalProperties (a map) and additionalProperties: true are
+// both incompatible.
+func isStrictCompatible(schema map[string]any) bool {
```
docker-agent, line 38: **[LOW] `isStrictCompatible` closure exits early but `walkSchema` continues full traversal**

Once `compatible` is set to `false`, the closure immediately returns on subsequent calls (`if !compatible { return }`), but `walkSchema` has no mechanism to stop recursing — it continues visiting every remaining node in the schema tree, paying the traversal cost for nothing.

For typical MCP tool schemas this is negligible, but for tools with very large or deeply-nested schemas the wasted work compounds for every tool on every request.

**Possible fix**: Use a sentinel (e.g. a named return via a separate function with early return, or a dedicated recursive walk that returns a bool) so the traversal stops as soon as the first incompatibility is found.

docker-agent (commented):

This PR correctly identifies the root cause (Notion's schema-form `additionalProperties` must survive strict-mode normalization) and the per-tool strict flag approach is sound. One behavioral regression was found in the Chat Completions path, and one minor performance issue in the compatibility check.

pkg/model/provider/openai/client.go
docker-agent, line 261: **[MEDIUM] Chat Completions path skips `makeAllRequired` for strict-incompatible schemas**

With this change, `ConvertParametersToSchema` returns a per-tool `strict` flag but `CreateChatCompletionStream` discards it:

```go
parameters, _, err := ConvertParametersToSchema(tool.Parameters)
```

Before this PR, every schema went through `makeAllRequired` (which fully populates `required` arrays and sets `additionalProperties: false` on every object node). After this PR, schemas that are strict-incompatible (Notion, Jira) now travel a lighter path — `ensureTypeFields` + `removeFormatFields` + `fixSchemaArrayItems` only — even when they are being sent to the Chat Completions endpoint, which doesn't enforce strict mode at all.

**Impact**: Tools like Notion's block-creation or Jira's `editJiraIssue` used via the Chat Completions API (e.g., gpt-4o) will now receive parameter schemas with incomplete `required` arrays. The OpenAI Chat API is lenient and won't reject these schemas, but models use `required` to distinguish mandatory from optional parameters. The absence of the array may reduce tool-calling reliability on non-gpt-5 models.

**Suggestion**: Apply `makeAllRequired` unconditionally (as before) for the Chat Completions path, and only branch on `strict` for the Responses API path. One way:

```go
func ConvertParametersToSchema(params any) (shared.FunctionParameters, bool, error) {
    p, err := tools.SchemaToMap(params)
    if err != nil {
        return nil, false, err
    }

    strict := isStrictCompatible(p)
    if strict {
        return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(makeAllRequired(p)))), true, nil
    }
    // For non-strict: still run makeAllRequired for Chat API correctness,
    // but the Responses API caller will pass Strict: false.
    return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(makeAllRequired(p)))), false, nil
}
```

docker-agent (commented):

One finding in changed code. The strict-incompatibility detection logic is sound, and the Responses API path is correctly fixed. One behavioral change on the Chat Completions path warrants attention.

pkg/model/provider/openai/schema.go
```diff
@@ -52,17 +109,17 @@ func walkSchema(schema map[string]any, fn func(map[string]any)) {
...
-// because that's what the OpenAI Response API demands.
-// Properties that were not originally required are made nullable.
-// Also ensures all object-type schemas have additionalProperties: false.
+// makeAllRequired makes every object property `required` (newly-required ones
+// are made nullable) and ensures every object node has `additionalProperties`
+// set. It runs on every schema regardless of strict-mode compatibility, so
+// schema-form additionalProperties (e.g. Notion's dictionary value shape) is
+// preserved — only missing/true/nil values are forced to `false`.
 func makeAllRequired(schema shared.FunctionParameters) shared.FunctionParameters {
 	if schema == nil {
```
aheritier, line 134: **Blocking:** This unconditional `node["additionalProperties"] = false` is what regressed Notion (and is the root cause flagged by the two existing bot reviews from different angles). Restore the lenient guard here so `makeAllRequired` only defaults `additionalProperties` to `false` when it is missing or explicitly `true`, and preserves an existing schema-form value or an explicit `false`:

```go
if isObject {
    switch node["additionalProperties"].(type) {
    case nil:
        node["additionalProperties"] = false
    case bool:
        if node["additionalProperties"].(bool) {
            node["additionalProperties"] = false
        }
    // map[string]any (schema form) → leave as-is
    }
}
```

With that guard in place, `ConvertParametersToSchema` can call `makeAllRequired` unconditionally on the Chat Completions path too — which fixes the other bot's finding that Jira/Notion tools currently skip `required`-array population — without re-breaking Notion's schema-form `additionalProperties`. One change unifies both bot findings.

pkg/model/provider/openai/schema.go
```diff
@@ -9,14 +9,71 @@ import (
...
+	return fixSchemaArrayItems(removeFormatFields(ensureTypeFields(makeAllRequired(p)))), strict, nil
+}
+
+// isStrictCompatible reports whether the schema can use OpenAI strict mode.
+// Strict mode requires every object node to have additionalProperties: false.
+// Schema-form additionalProperties (a map) and additionalProperties: true are
+// both incompatible. The walk stops at the first incompatible node.
+func isStrictCompatible(schema map[string]any) bool {
+	return !hasIncompatibleNode(schema)
+}
```
aheritier, line 40: **Question:** Worth a short doc comment here making the contract explicit: the strict decision is evaluated per tool (mirroring the Responses API), and a single non-compliant node anywhere in the schema disables strict mode for the entire tool. Is that all-or-nothing behavior intentional? It's the right call for the Responses API, but a one-liner above `isStrictCompatible` would save the next reader from re-deriving it.

pkg/model/provider/openai/schema_test.go
```diff
@@ -192,23 +192,53 @@ func TestMakeAllRequired_AdditionalProperties(t *testing.T) {
...
 	assert.Contains(t, additionalRequired, "numbered_list_item")
 
-	// numbered_list_item was not originally required, so its type should be nullable
 	numberedListItem := additionalProps["properties"].(map[string]any)["numbered_list_item"].(map[string]any)
 	assert.Equal(t, []string{"string", "null"}, numberedListItem["type"])
 
-	// bulleted_list_item was originally required, so its type should be unchanged
 	bulletedListItem := additionalProps["properties"].(map[string]any)["bulleted_list_item"].(map[string]any)
 	assert.Equal(t, "string", bulletedListItem["type"])
 }
```
aheritier, line 215: **Nit:** Rename to `TestConvertParametersToSchema_JiraEditIssueFields` — the test exercises the public `ConvertParametersToSchema` entry point, not `makeAllRequired` directly, and the name should track the SUT.

aheritier (dismissed):
Acknowledging the two prior bot reviews — both point at real issues (Chat Completions skipping `makeAllRequired`, and `isStrictCompatible` not short-circuiting on schema-form `additionalProperties`). The Blocking comment proposes a single unifying fix: restore the lenient guard in `makeAllRequired` so it preserves schema-form / explicit-`false` `additionalProperties`, then call it unconditionally on both paths. That resolves the Chat Completions regression for Jira/Notion tools without re-breaking Notion.

CI: no checks reported on the head commit yet — please confirm `task test` / `task lint` once they run.

Praise: the regression tests (`TestConvertParametersToSchema_NotionStylePreservesShape`, the Jira fields case) and the root-cause writeup in the PR body are excellent — they make the failure mode obvious to the next reader.

pkg/model/provider/openai/schema.go
```diff
@@ -9,14 +9,75 @@ import (
...
+		if variants, ok := node[keyword].([]any); ok {
+			for _, v := range variants {
+				if sub, ok := v.(map[string]any); ok && hasIncompatibleNode(sub) {
+					return true
+				}
+			}
+		}
+	}
+
+	if items, ok := node["items"].(map[string]any); ok && hasIncompatibleNode(items) {
```
docker-agent, line 72: **[MEDIUM] `hasIncompatibleNode` does not traverse `prefixItems`, risking false strict=true on tuple-typed tool schemas**

The incompatibility walk covers `properties`, `anyOf`/`oneOf`/`allOf`, `items`, and `additionalProperties`, but skips `prefixItems` (JSON Schema draft 2020-12 positional array items). If any MCP tool schema uses `prefixItems` with a sub-schema that declares schema-form `additionalProperties`, `isStrictCompatible` will return `true` and the tool will be sent with `Strict: true` to the Responses API — producing a 400 error at runtime identical to the Jira regression being fixed here.

```go
// prefixItems not checked:
if items, ok := node["items"].(map[string]any); ok && hasIncompatibleNode(items) {
    return true
}
// missing:
// if prefixItems, ok := node["prefixItems"].([]any); ok { ... }
```

**Suggested fix** — add a `prefixItems` traversal immediately after the `items` check:
```go
if prefixItems, ok := node["prefixItems"].([]any); ok {
    for _, v := range prefixItems {
        if sub, ok := v.(map[string]any); ok && hasIncompatibleNode(sub) {
            return true
        }
    }
}
```

Note: `walkSchema` already walks into `additionalProperties` sub-schemas recursively, but also lacks `prefixItems` traversal — consider adding it there too for consistency.

pkg/model/provider/openai/schema.go
```diff
@@ -9,14 +9,75 @@ import (
...
+		return true
+	}
+
+	if additionalProps, ok := node["additionalProperties"].(map[string]any); ok && hasIncompatibleNode(additionalProps) {
```
docker-agent: **[LOW] Unreachable second `additionalProperties` map-check — dead code**

The type-switch at the top of `hasIncompatibleNode` already returns `true` for any `map[string]any` `additionalProperties` value (line 45). The second check at line 76 can therefore never be reached: by the time execution could get there, a `map[string]any` value would have caused an early return.

```go
// Line 43-51: returns true for map[string]any — covers all maps
if v, ok := node["additionalProperties"]; ok {
    switch t := v.(type) {
    case map[string]any:
        return true   // ← any map exits here
    ...
    }
}

// Line 76: unreachable — map[string]any already returned true above
if additionalProps, ok := node["additionalProperties"].(map[string]any); ok && hasIncompatibleNode(additionalProps) {
    return true
}
```

The second check was presumably meant to recursively walk into the sub-schema for nested incompatibilities, but since the first check makes the function return `true` for any map-form value, the recursive walk is redundant. Consider removing lines 76-78 to avoid confusion about intent.

docker-agent (commented):

This PR correctly fixes the strict-mode per-tool decision for Notion and Jira MCP tools. The three-return-value `ConvertParametersToSchema` is properly threaded through all callers, and `makeAllRequired` correctly preserves schema-form `additionalProperties` (the Notion fix). Two issues found in the newly-added `hasIncompatibleNode` walk.

aheritier (approved):

PR #2840, opened 2026-05-20
feat(mcpcatalog): hide disable / reset_auth tools when no server is enabled

dgageot:
## What

Optimise the MCP catalog builtin tool's meta-surface: `disable_remote_mcp_server` and `reset_remote_mcp_server_auth` now only appear once at least one server is enabled. When the catalog has no enabled server, the LLM sees only the three useful actions (`search`, `list`, `enable`).

## Why

The two gated tools are no-ops when nothing is enabled — they just bloat the prompt and tempt the model to call them on an empty surface. Hiding them keeps the meta-surface minimal until the model has actually activated something.

## Notes on concurrency

`Tools()` snapshots ` + "`t.enabled`" + ` under ` + "`RLock`" + ` and reads ` + "`len(enabled)`" + ` on the local slice — no shared-state access after unlock. ` + "`handleEnable`" + ` and ` + "`handleDisable`" + ` already fire ` + "`toolsChangedHandler`" + ` whenever the count flips, so the runtime re-enumerates and any stale view self-corrects on the next turn.

## Tests

- ` + "`TestEnableDisableLifecycle`" + ` strengthened to assert disable/reset are hidden initially.
- ` + "`TestToolsExposesEnabledServerTools`" + ` extended to assert all five meta-tools (including ` + "`reset_auth`" + `) are visible once a server is enabled.
- New ` + "`TestDisableAndResetAuthGatedOnEnabledServers`" + ` covers the full hide → enable → show → disable → hide cycle, using a local httptest fake auth-required MCP server (no network, deterministic).

` + "`task lint`" + ` and ` + "`task test`" + ` are clean.

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
+++ b/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
@@ -2,7 +2,7 @@
 // streamable-http servers as a single agent-side toolset that supports
 // on-demand activation.
 //
-// The toolset surfaces five meta-tools to the model:
+// The toolset surfaces up to five meta-tools to the model:
 //
 //   - search_remote_mcp_servers — case-insensitive fuzzy search over the
 //     curated catalog (id / title / description / category / tags).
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The gating logic for `disable_remote_mcp_server` and `reset_remote_mcp_server_auth` is sound. The snapshot-based concurrency is the correct pattern here — the `toolsChangedHandler` notify mechanism ensures any transiently stale view self-corrects on the next turn, which is the intended design.

No actionable bugs found in the changed code.

rumpl (approved):

PR #2841, opened 2026-05-20
app: emit run_skill directive for forked skills

dgageot:
## Problem

When a user invokes a forked skill via a slash command (e.g. `/commit`), the agent was getting confused: it would call `read_skill`, then attempt `transfer_task`, then sometimes execute the work inline — instead of calling `run_skill` once.

Root cause: `ResolveSkillCommand` in `pkg/app/app.go` produced the same "Use the following skill" + inlined SKILL.md body for **every** skill, ignoring the `IsFork()` flag. The agent then had to *infer* from `<available_skills>` metadata that it should use `run_skill` instead of acting on the inlined body — which is exactly the inference that fails.

## Fix

Branch on `skill.IsFork()` in `ResolveSkillCommand`. For forked skills, emit a deterministic, body-free directive:

```
Call the run_skill tool with name="commit" and task="...". Do not call read_skill or transfer_task for this skill.
```

The SKILL.md body is no longer inlined for forked skills, removing the strongest in-context signal that was making the agent act on it directly. Non-forked skills are unchanged.

## Tests

- `TestApp_ResolveSkillCommand_NonForkInlinesSkillBody` — non-forked path still inlines the body
- `TestApp_ResolveSkillCommand_ForkSkillEmitsRunSkillDirective` — forked path:
  - emits the `run_skill` directive
  - suppresses the SKILL.md body
  - forwards trailing args as the task
  - whitespace-only args fall back to the no-argument task
  - quoted/backslash-containing args are escaped unambiguously via `%q`

## Validation

- `task lint` ✅
- `task test` ✅
- `go build ./...` ✅

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -226,7 +226,10 @@ func (a *App) CurrentAgentSkills() []skills.Skill {
 }
 
 // ResolveSkillCommand checks if the input matches a skill slash command (e.g. /skill-name args).
-// If matched, it reads the skill content and returns the resolved prompt. Otherwise returns "".
+// If matched, it returns the resolved prompt. Otherwise returns "".
+//
+// Forked skills emit a directive to invoke run_skill rather than inlining
+// SKILL.md, which would tempt the agent to act on the body directly.
 func (a *App) ResolveSkillCommand(ctx context.Context, input string) (string, error) {
 	if !strings.HasPrefix(input, "/") {
 		return "", nil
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed `pkg/app/app.go` and `pkg/app/app_test.go`.

**Summary**: The `IsFork()` branch in `ResolveSkillCommand` is correct. The logic properly:
- Uses `strings.Cut` + `TrimSpace` so whitespace-only args collapse to `""` and trigger the no-argument fallback
- Uses `%q` to safely escape both `skill.Name` and `task` in the directive string (validated by the `ForkSkillEscapesSpecialChars` test)
- Leaves the non-forked code path untouched

**Verification results**: 7 hypotheses checked, all dismissed or not applicable. One theoretical low-severity note: if a skill name somehow contained special characters (e.g. backslash), `%q` would produce Go escape sequences in the LLM directive. In practice skill names are slug-like identifiers from configuration, so this is not a realistic concern.

**Test coverage**: The three new test functions cover the main success paths, whitespace trimming, special-character quoting, and the body-suppression invariant. The mock correctly threads `skillsToolset` into `CurrentAgentSkillsToolset()` without breaking existing tests.

PR #2842, opened 2026-05-20
fix(tui): restore terminal on Ctrl-C when bubbletea shutdown stalls

dgageot:
## Problem

When the user hits Ctrl-C in the TUI and confirms the exit, the terminal is sometimes left in a broken state — raw mode still on, alt-screen still active, mouse tracking on, cursor hidden, etc.

## Root cause

The shutdown safety net in `appModel.cleanupAll` was implemented as:

```go
go func() {
    time.Sleep(shutdownTimeout)
    exitFunc(0)
}()
```

It exists because bubbletea's renderer can deadlock during shutdown when stdout is wedged (the final flush after `tea.Quit` re-acquires the renderer mutex still held by a blocked previous flush — `TestExitDeadlock_BlockedStdout` proves it). But:

- It always sleeps the full timeout, even on a clean shutdown — leaks a goroutine until the process exits.
- `os.Exit(0)` runs no defers, so the terminal stays in raw mode + alt-screen + mouse tracking + hidden cursor + kitty keyboard flags + …

## Fix

Race `program.Wait()` against the deadline so a clean shutdown skips the force-exit path entirely:

```go
select {
case <-done: // Wait() returned, bubbletea cleaned up normally
case <-time.After(timeout):
    go func() { _ = program.ReleaseTerminal() }()
    exit(0)
}
```

`Program.ReleaseTerminal()` is bubbletea's own terminal-restore primitive (out of raw mode, alt-screen exit, cursor on, mouse off, etc.) and is independent from the regular shutdown path. It can itself deadlock on the wedged renderer mutex, so it's run in a fire-and-forget goroutine — best-effort terminal restore, then exit no matter what.

Side fix: `shutdownTimeout` and `exitFunc` are package globals that tests mutate; they're now snapshotted synchronously inside `cleanupAll` so the safety-net goroutine can't race with `t.Cleanup` restoring the originals.

## Tests

New tests in `pkg/tui/tui_exit_test.go`, all passing under `-race`:

- `TestCleanupAll_SpawnsSafetyNet` — stuck program → `exitFunc` fires.
- `TestCleanupAll_GracefulShutdownSkipsExit` — clean shutdown → `exitFunc` does NOT fire.
- `TestCleanupAll_NilProgramIsSafe` — no program → no safety-net spawned.
- `TestCleanupAll_WedgedStdoutFiresExit` — realistic regression test: bubbletea is stuck on a wedged stdout, `Wait()` never returns, `ReleaseTerminal()` would itself deadlock — `exitFunc` must still fire. Verified to fail under the previous inline `ReleaseTerminal` call.

## Validation

- `task lint` ✅
- `task test` ✅
- `go test ./pkg/tui -race` ✅

### DIFF
--- a/pkg/tui/tui.go
+++ b/pkg/tui/tui.go
@@ -2376,15 +2376,32 @@ func (m *appModel) cleanupAll() {
 		ed.Cleanup()
 	}
 
-	// Safety net: force-exit if bubbletea's shutdown gets stuck.
-	// This can happen when the renderer's flush goroutine blocks on a
-	// stdout write (terminal buffer full) while holding the renderer
-	// mutex, preventing the event loop from completing the render call
-	// that follows tea.Quit.
+	// Safety net: bubbletea's renderer can deadlock during shutdown when
+	// stdout is wedged — the final flush after tea.Quit tries to
+	// re-acquire the renderer mutex that the still-blocked previous flush
+	// is holding. We race bubbletea's Wait() (which unblocks when Run()
+	// returns) against a deadline; if shutdown stalls past the deadline
+	// we restore the terminal and force-exit so the user isn't stuck with
+	// a wedged TUI.
+	program := m.program
+	if program == nil {
+		return
+	}
 	go func() {
-		time.Sleep(shutdownTimeout)
-		slog.Warn("Graceful shutdown timed out, forcing exit")
-		exitFunc(0)
+		done := make(chan struct{})
+		go func() {
+			program.Wait()
+			close(done)
+		}()
+
+		select {
+		case <-done:
+			// Graceful shutdown completed; nothing to do.
+		case <-time.After(shutdownTimeout):
+			slog.Warn("Graceful shutdown timed out, forcing exit")
+			_ = program.ReleaseTerminal()
+			exitFunc(0)
+		}
 	}()
 }
 
--- a/pkg/tui/tui.go
+++ b/pkg/tui/tui.go
@@ -2383,10 +2383,16 @@ func (m *appModel) cleanupAll() {
 	// returns) against a deadline; if shutdown stalls past the deadline
 	// we restore the terminal and force-exit so the user isn't stuck with
 	// a wedged TUI.
+	//
+	// shutdownTimeout and exitFunc are package globals that tests mutate;
+	// snapshot them synchronously here so the safety-net goroutine can't
+	// race with t.Cleanup restoring the originals.
 	program := m.program
 	if program == nil {
 		return
 	}
+	timeout := shutdownTimeout
+	exit := exitFunc
 	go func() {
 		done := make(chan struct{})
 		go func() {
@@ -2397,10 +2403,15 @@ func (m *appModel) cleanupAll() {
 		select {
 		case <-done:
 			// Graceful shutdown completed; nothing to do.
-		case <-time.After(shutdownTimeout):
+		case <-time.After(timeout):
 			slog.Warn("Graceful shutdown timed out, forcing exit")
-			_ = program.ReleaseTerminal()
-			exitFunc(0)
+			// ReleaseTerminal acquires the renderer mutex which may be
+			// the very thing that's stuck (a wedged stdout write inside
+			// the ticker's flush goroutine), so run it in its own
+			// goroutine. We don't wait for it: best-effort terminal
+			// restore, then force-exit no matter what.
+			go func() { _ = program.ReleaseTerminal() }()
+			exit(0)
 		}
 	}()
 }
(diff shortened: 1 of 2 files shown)


pkg/tui/tui_exit_test.go
```diff
@@ -348,38 +337,130 @@ func TestCleanupAll_SpawnsSafetyNet(t *testing.T) {
...
+	assert.NotPanics(t, func() { m.cleanupAll() })
+
+	time.Sleep(shutdownTimeout + 50*time.Millisecond)
+	assert.False(t, exitCalled.Load(), "exitFunc must not fire without a program")
+}
+
+// TestCleanupAll_WedgedStdoutFiresExit: the realistic case. The renderer is
+// stuck on a wedged stdout write, so Wait() never returns AND ReleaseTerminal
+// would itself deadlock on the same mutex. exitFunc must still fire.
+func TestCleanupAll_WedgedStdoutFiresExit(t *testing.T) {
```
docker-agent, line 418: **[MEDIUM] Test doesn't exercise the wedged-renderer scenario it claims to test**

`TestCleanupAll_WedgedStdoutFiresExit` never sends a `triggerQuitMsg` to the running program. Without a quit signal the program keeps running indefinitely, so `program.Wait()` blocks simply because the program is still alive — not because the renderer is stuck trying to acquire a mutex held by a blocked stdout flush.

The wedged-renderer deadlock path requires:
1. The program receives a quit signal.
2. bubbletea initiates its final render flush.
3. The renderer tries to re-acquire the mutex already held by the still-blocked write goroutine → deadlock.

None of these steps occur here, so the test is actually exercising: *"Wait() blocks when the program has no reason to quit → timeout fires → exitFunc is called"* — a trivially weaker property.

The companion test `TestExitDeadlock_BlockedStdout` does send `triggerQuitMsg` and correctly demonstrates the real deadlock. `TestCleanupAll_WedgedStdoutFiresExit` should do the same (send `triggerQuitMsg` after calling `m.cleanupAll()`, before waiting for `exitDone`) to actually validate the safety net against the renderer-mutex deadlock.

docker-agent (commented):

docker-agent (commented):

pkg/tui/tui_exit_test.go
docker-agent, line 418: **[MEDIUM] Test does not exercise the claimed quit-induced renderer deadlock**

`TestCleanupAll_WedgedStdoutFiresExit` never sends `triggerQuitMsg{}` to the program, so the bubbletea event loop never issues a `tea.Quit` and the renderer never attempts the final flush that triggers the real deadlock.

What the test actually exercises: `program.Wait()` blocks because the renderer's periodic write is stalled on the blocked writer — unrelated to the quit path. The program only exits when the test-supplied `t.Context()` is cancelled (at test cleanup), which may race with the 300 ms `shutdownTimeout` window.

The comment says _"the renderer is stuck on a wedged stdout write, so Wait() never returns AND ReleaseTerminal would itself deadlock on the same mutex"_ — but without `tea.Quit` being sent, the renderer is not in the quit-flush code path that actually holds the mutex. The scenario described in `TestExitDeadlock_BlockedStdout` (which *does* send `triggerQuitMsg{}`) is the actual regression; this test does not reproduce it.

Suggested fix: send `triggerQuitMsg{}` before calling `m.cleanupAll()`, so the program attempts the quit render against the blocked writer and truly reproduces the wedged-renderer deadlock.

pkg/tui/tui.go
```diff
@@ -2376,15 +2376,33 @@ func (m *appModel) cleanupAll() {
...
+	// blocked previous flush is holding. Race Wait() against a deadline
+	// and force-exit if shutdown stalls. Snapshot the package globals so
+	// they can't race with t.Cleanup.
+	program := m.program
+	if program == nil {
+		return
+	}
+	timeout := shutdownTimeout
+	exit := exitFunc
 	go func() {
```
docker-agent, line 2393: **[MEDIUM] No guard prevents multiple concurrent safety-net goroutines from each calling `exit(0)`**

`cleanupAll()` is called from at least three message-handler sites (lines ~697, ~768, ~772). Each call creates a fresh safety-net goroutine with its own `done` channel and snapshotted `exit`. If `cleanupAll()` is called N times before the process exits (e.g. `ExitSessionMsg` followed by `ExitConfirmedMsg` on the same model), N independent goroutines run concurrently; each can independently reach the `time.After(timeout)` branch and call `exit(0)`.

In production `exit` is `os.Exit` so the second call never happens (process is gone). In tests, though, `exit` is replaced with a function like `func(int) { close(exitDone) }` — a double call panics with "close of closed channel" and crashes the entire test binary.

Consider adding a `sync.Once` or atomically nil-ing `m.program` inside `cleanupAll` (before launching the goroutine) so only the first caller races Wait against the deadline:

```go
program := m.program
m.program = nil   // or use sync.Once at the appModel level
if program == nil {
    return
}
```

rumpl (approved):

PR #2843, opened 2026-05-20
fix(tui): user_prompt dialog opens scrolled to top and respects user scrolling

dgageot:
## Problem

When the `user_prompt` dialog showed a long question with many multiple-choice options, a scrollbar appeared (as expected) but the user could not scroll all the way up to read the beginning of the question — the viewport kept snapping back down to keep the focused option on screen.

## Cause

`ElicitationDialog.View()` was calling `d.ensureFocusVisible()` on every render. So whenever the user scrolled up (wheel, PgUp, scrollbar drag, …), the next render snapped the scroll position back down to keep the focused field/option visible.

## Fix

Auto-scroll is now triggered only on actual user interactions that change focus or selection:

- `moveFocus` / `focusField` (Tab, Shift-Tab, validation jump-to-error)
- `moveSelection` (Up/Down/Space inside an enum/boolean field)
- Typing or pasting into a text input (so a text field below the fold gets revealed once the user starts editing it)

Render-time auto-scroll has been removed from `View()`. As a result:

- The dialog now opens scrolled all the way to the top, so the question/message is visible from the start.
- The user can freely scroll up/down (wheel, PgUp/PgDn, Home/End, scrollbar drag) without the viewport snapping back.

## Tests

Added regression tests in `pkg/tui/dialog/elicitation_test.go`:

- `TestElicitationDialog_OpensScrolledToTop` — verifies the dialog opens with `ScrollOffset == 0`.
- `TestElicitationDialog_UserScrollUp_NotSnappedBack` — verifies that re-rendering after a manual scroll does not override the user's position.
- `TestElicitationDialog_TypingRevealsBelowFoldField` — verifies that typing into a text field below the fold scrolls it into view.

Existing scroll-related tests (`TestElicitationDialog_LongEnumScrolls`, `TestElicitationDialog_FieldsBelowFold_AreReachable`, `TestElicitationDialog_SmallContent_NoScrollbar`) still pass.

### DIFF
--- a/pkg/tui/dialog/elicitation.go
+++ b/pkg/tui/dialog/elicitation.go
@@ -164,6 +164,7 @@ func (d *ElicitationDialog) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 		if d.isTextInputField() {
 			var cmd tea.Cmd
 			d.inputs[d.currentField], cmd = d.inputs[d.currentField].Update(msg)
+			d.ensureFocusVisible()
 			return d, cmd
 		}
 		return d, nil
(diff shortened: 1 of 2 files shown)


melmennaoui (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly moves `ensureFocusVisible()` out of `View()` and into the four interaction handlers (`moveFocus`, `focusField`, `moveSelection`, `updateCurrentInput`). The render-time snap-back is eliminated and the three new regression tests cover the key contracts.

**No confirmed bugs found in the changed code.**

The only asymmetry identified (paste into free-form input does not call `ensureFocusVisible()`) was verified to be a non-issue: `fieldStarts` is only populated for schema fields, so the call would be a no-op anyway in the free-form path.

aheritier:
I confirm it's ok

PR #2844, opened 2026-05-20
feat(sandbox): docker-agent kit, gateway allowlist, and assorted --sandbox fixes

dgageot:
## Problem

Running an agent inside `--sandbox` is the right safety story, but in practice it broke as soon as you tried any non-trivial agent. The list of issues this PR fixes (in the order I hit them while debugging):

1. The agent inside the sandbox tried to read host files that aren't mounted: skills under `~/.codex/skills`, `~/.claude/skills`, `~/.agents/skills`, and `add_prompt_files` entries (`AGENTS.md` and friends) walked up from `cwd` and `$HOME`.
2. Just bind-mounting more host paths is hard:
   - `$HOME` inside the sandbox is unrelated to the host's `$HOME`.
   - Sandbox mounts target directories, not individual files.
   - Host files may contain secrets that must not leak.
   - Other host-only state (e.g. `.agents/skills` discovered above the workspace) is unreachable.
3. `docker agent run --sandbox <alias>` ignored the alias and launched without the alias's destination YAML mounted.
4. With a models gateway configured, `docker agent run --sandbox` failed at startup with `you first need to sign in Docker Desktop to use the Docker AI Gateway` even though the user was signed in.
5. With the gateway reachable, the same flow then 403'd because the sandbox's default-deny network policy blocks every `*.docker.com` host (the gateway is one).
6. Auto-install for tools like `gopls` produced opaque `403 Blocked by network policy` errors from `proxy.golang.org`.
7. The user couldn't see what was actually being shipped into the sandbox or which gateway the inner would reach.
8. Slightly changing the mount set between two runs accumulated suffixed sandbox names (`-1`, `-2`, …) until the daemon eventually refused to start a new one with `Error: opening filesystem /Users/.../.agents: no such file or directory`. Two compounding bugs: `sbx rm` requires `--force` when stdin isn't a TTY (we were calling it without), and `ForWorkspace` only inspected the first match when several existed for the same workspace.
9. A handful of latent regressions in `pkg/sandbox` made the symptom space confusing while debugging the above.

### 1. A docker-agent kit

Before launching the sandbox, the host stages a self-contained directory under `<cache>/sandbox-kits/<hash>/`:

```
skills/<skill-name>/      # local skills, recursively
prompt_files/<name>       # collected add_prompt_files inputs
manifest.json             # debug + cache key
```

It's bind-mounted read-only into the sandbox at `/agent-kit`, with `DOCKER_AGENT_KIT_DIR=/agent-kit` forwarded so the in-sandbox resolvers find it. Every text file is run through `portcullis.Redact` during staging, so secrets that survive in `~/.agents/skills/foo/SKILL.md` don't reach the sandbox.

The runtime resolvers consult `DOCKER_AGENT_KIT_DIR` and behave as no-ops when it isn't set:

- `pkg/skills/local.go`: when set, the search is rooted only at `<kit>/skills`. Host paths are skipped because they don't exist inside the sandbox.
- `pkg/hooks/builtins/add_prompt_files.go`: when set, prompt-file lookups prefer the kit over `$HOME` (the workspace cwd-walk still wins for files served live by the workspace mount).

The kit lifetime is now bound to the cache: it's keyed by content hash and rebuilt atomically (`mkdtemp` → `os.Rename` → reap), and we never delete it after a run because the docker sandbox we reuse holds a hard reference to its bind-mount path.

### 2. User-visible kit summary

Between the kit prep and the sandbox-create line, the CLI now prints what the agent will see:

```
Preparing docker-agent kit at /Users/.../sandbox-kits/<hash>
  skills:
    plain (from ~/.agents/skills/plain)
      SKILL.md
    with-secret (from ~/.agents/skills/with-secret)
      SKILL.md (redacted)
      helper.sh
  prompt files:
    AGENTS.md (from ~/AGENTS.md, redacted)
    AGENTS.md (from /Users/.../project/AGENTS.md, workspace mount)
  summary: 2 skills, 2 prompt files, 2 secrets redacted
Models gateway: https://ai-backend-service-stage.docker.com/proxy (allowlisting ai-backend-service-stage.docker.com in the sandbox proxy)
Tool install: agent has at least one MCP/LSP toolset, allowlisting 8 package hosts in the sandbox proxy
✓ Created sandbox '…'
```

`workspace mount` marks files reachable through the live workspace mount (no kit copy needed); `redacted` marks files where `portcullis.Redact` scrubbed at least one secret. The gateway URL is rendered through a credential-safe formatter (`https://user:pw@gw/...` → `https://***@gw/...`) so it never leaks userinfo to stdout / logs.

### 3. Network policy: allowlist the gateway and (when needed) tool-install hosts

The sandbox template's HTTP CONNECT proxy at `gateway.docker.internal:3128` enforces a default-deny policy that allows the major model providers (`api.anthropic.com`, `api.openai.com`, ...) but blocks every `*.docker.com` host **and** the package-registry / source hosts auto-install reaches at runtime. The kit pipeline now opens the minimum:

- the **gateway host** when `runConfig.ModelsGateway` is set;
- the **package-host set** (`github.com`, `api.github.com`, `*.githubusercontent.com`, `proxy.golang.org`, `sum.golang.org`, `storage.googleapis.com`, …) only when the agent has at least one MCP / LSP toolset that may auto-install.

Detection is done at kit build time by walking `cfg.MCPs` and `cfg.Agents[*].Toolsets` for entries with type `mcp`/`lsp`, a `Command` set, and `Version` not `"false"`/`"off"` (the per-toolset opt-out `toolinstall.EnsureCommand` already honours). When neither feature is in play the strict default-deny is preserved.

Per-backend command spelling is hidden in `Backend.allowHostsArgs`: `sbx policy allow network <name> <hosts>` for sbx, `docker sandbox network proxy <name> --allow-host <host>` for docker. `gatewayHostPort` handles every realistic shape: full URLs, scheme-relative `//host`, bare authorities with optional `:port` and trailing path/query/fragment, IPv6 hosts, scheme-without-host, opaque schemes (`mailto:`), and bogus `foo:bar://x`.

### 4. Token forwarding without leakage

For the inner agent's startup gateway check we forward the live Docker Desktop JWT. Two important details, both surfaced through review:

- We bypass the env-chain (which consults `OsEnvProvider` first and would shadow Docker Desktop with a stale exported `DOCKER_TOKEN`) and call `desktop.GetToken(ctx)` directly — same source as the file-based refresher.
- We pass `-e DOCKER_TOKEN` **name-only** in argv and inject the value through `cmd.Env`. The previous `-e DOCKER_TOKEN=<jwt>` form was leaking the live bearer token into the slog'd `Executing in sandbox` debug log every run.

The same name-only pattern is used for `DOCKER_AGENT_MODELS_GATEWAY`, both for argv-leak hygiene and consistency.

### 5. Sandbox lifecycle: stop accumulating suffixed sandboxes

Two compounding bugs were causing `Note: sandbox already exists, creating ...-2 instead` warnings to pile up across runs, eventually leading to `Error: opening filesystem /Users/.../.agents`:

- `sbx rm <name>` prompts for confirmation when stdin isn't a TTY (`ERROR: stdin is not a terminal; use --force to skip confirmation`). Our previous rm was running without `--force` and without an attached stdin, so every rm silently failed and the stale sandbox lived on. Then `sbx create` saw the name as taken and suffixed the new one with `-1`, `-2`, …
- `ForWorkspace` returned only the first match. Once suffixing had happened, the canonical name (`<workspace>` with no suffix) was the oldest entry; we'd inspect *its* mounts (stale), try to rm it (silent failure per above), and create yet another suffix on top.

`Backend` now carries an `rmExtraArgs` field — `["--force"]` for sbx, nil for docker (whose `rm` has no prompt and rejects the flag). New `Backend.allForWorkspace` returns every sandbox whose primary workspace matches; `Ensure` walks the whole list, reuses the first one whose mounts already cover the requested set (extra read-only mounts are harmless), and otherwise removes **every** match before creating fresh. After a run there is exactly one sandbox per workspace with the canonical name.

### 6. Latent regression fixes (drove half of the debugging)

- **`pkg/sandbox/sandbox.go`** — `docker sandbox ls --json` and `sbx ls --json` both return `{"sandboxes": [...]}` now; the code was still reading `vms`. Switched to the new key, kept a fallback for older CLIs with a warn-level log.
- **`cmd/root/flags.go`** — `addGatewayFlags` materialised `runConfig.EnvProvider()` *before* invoking the parent `PersistentPreRunE`. The env-provider chain caches its result, so any `paths.SetConfigDir` from `--config-dir` (set by the parent) was invisible — the in-sandbox `SandboxTokenProvider` then read from the wrong path on existing v1.59.x sandbox images. Fixed the ordering; regression-tested.
- **`pkg/sandbox/args.go`** — `ExtraWorkspace` did its own `filepath.Abs` + extension check, which produced `""` for alias names like `gopher`. Delegated to `config.Resolve` + `Source.ParentDir` so the alias's destination YAML actually gets mounted.

## Code organisation

| Package | Purpose |
|---|---|
| `pkg/promptfiles` | single source of truth for `add_prompt_files` lookup; consumed by both the runtime hook and the kit builder |
| `pkg/sandbox/kit` | stages skills + prompt files, redacts via `portcullis.Redact`, writes a manifest, atomic-promotes the staging dir, computes `NeedsToolInstall` |
| `pkg/sandbox` | `Backend.Ensure(extras []string)`, `Backend.AllowHosts(name, hosts)`, `Backend.allForWorkspace`, `Backend.rm` (with per-backend `rmExtraArgs`), `cleanExtras` (path-canonical dedup), the legacy-key fallback in `ForWorkspace` |
| `cmd/root/sandbox.go` | wiring: build the kit, print the summary + gateway + tool-install line, allowlist the right hosts, forward env vars by name only |

## Tests

| Area | Coverage |
|---|---|
| Kit staging | skills + prompt files, redaction of `ghp_…`-shaped tokens, workspace-scoped exclusion vs parent-dir staging, symlink-escape rejection, intra-root symlinks, executable-bit preservation, on-disk manifest leakage check, concurrent `Build` safety, `hashKey` canonicalisation across path forms |
| `NeedsToolInstall` | nil cfg, empty cfg, agent with installable lsp, top-level `mcps` entry, per-toolset disable (`"off"`/`"false"`, case-insensitive), no-command, shell-type toolsets |
| Resolvers | `pkg/skills` and `pkg/promptfiles`: kit-aware fallbacks, `~`/`$HOME` collapsing |
| Print summary | header / per-file redaction marker / workspace-mount tag / summary line / `~` collapsing / nil-receiver / empty-kit cases |
| Gateway | `gatewayHostPort` across every URL shape; `displayGatewayURL` userinfo masking; `printModelsGateway` cases including credentials-bearing URLs; `autoInstallHosts` spot-check |
| Sandbox backend | `cleanExtras` canonical-path collapse, workspace-self-match drop, order preservation; `AllowHosts` rejects comma/whitespace, drops empties; `ForWorkspace` legacy `vms` fallback |
| Pre-run | `addGatewayFlags` parent ordering regression test |
| Alias mount | `ExtraWorkspace` regression test for the `--sandbox <alias>` case |

`task build`, `task test`, `task lint` all green.

## Verified end-to-end

- `docker agent run --sandbox gopher` against the Docker AI Gateway: kit prepared, gateway allowlisted, token forwarded, agent answers. The original `403 Blocked by network policy` and `you first need to sign in Docker Desktop` errors are gone.
- `go install golang.org/x/tools/gopls@latest` inside the sandbox now succeeds and writes a working binary.
- Reproduced the suffixed-sandbox pile-up state (two stale sandboxes, one with the wrong mounts) — re-running cleans both up and creates a single canonical-named sandbox in their place.
- An agent with no MCP/LSP toolset gets neither the install hosts opened nor the printed `Tool install:` line; the strict default-deny stays.

## Commits (15)

```
feat: stage docker-agent kit in sandbox with skills and prompt files
fix: address kit feature code review findings
feat(sandbox): print docker-agent kit contents at startup
fix(sandbox): mount alias target into the sandbox
fix(sandbox): use the "sandboxes" JSON key for both backends
fix(cli): run parent PersistentPreRunE before materialising the env provider
fix(sandbox): forward DOCKER_TOKEN, keep the kit on disk between runs
fix(sandbox/kit): list workspace-resident prompt files in the manifest
test(sandbox/kit): verify parent-dir prompt files get staged
fix(sandbox): use the live Docker Desktop token, not the env-chain value
feat(sandbox): allowlist the models-gateway host in the sandbox proxy
feat(sandbox): print which models gateway will be used before launch
fix: address sandbox/networking review findings
feat(sandbox): allowlist tool-install hosts when the agent needs them
fix(sandbox): force rm and clear every stale match before create
```


pkg/sandbox/sandbox.go
```diff
@@ -87,22 +87,48 @@ func (b *Backend) ForWorkspace(ctx context.Context, wd string) *Existing {
...
+	absConfigDir, err := absClean(configDir)
+	if err != nil {
+		return "", fmt.Errorf("resolving config dir: %w", err)
+	}
+	configDir = absConfigDir
+
+	// Filter extras: drop empties, normalise (Abs+Clean), dedupe, and
+	// remove anything that resolves to wd — wd is already mounted
+	// read-write so a duplicate :ro mount would shadow it.
+	dedup := map[string]bool{wd: true}
```
docker-agent: **[MEDIUM] `configDir` not seeded into dedup map — duplicate `:ro` mount when an extra resolves to `configDir`**

The `dedup` map is seeded with `wd` to avoid shadowing the read-write working-directory mount, but `configDir` is never added to it. If a caller passes a path in `extras` that resolves to the same absolute path as `configDir` (e.g. the kit dir happens to equal the config dir, or a caller accidentally passes it twice through different routes), `absClean(e)` will not match any dedup key, the path will be appended to `cleaned`, and later the loop at line ~151 will append `e+":ro"` **and** the unconditional `configDir+":ro"` at line ~155 will also be appended — producing two identical `:ro` mount specs for the same directory in the `docker sandbox create` call. Docker may reject or silently misbehave with duplicate bind-mount arguments.

Fix: seed `configDir` into the dedup map alongside `wd`:
```go
dedup := map[string]bool{wd: true, configDir: true}
```

pkg/sandbox/sandbox.go
```diff
@@ -87,22 +87,48 @@ func (b *Backend) ForWorkspace(ctx context.Context, wd string) *Existing {
...
+		cleaned = append(cleaned, abs)
+	}
+	extras = cleaned
 
 	existing := b.ForWorkspace(ctx, wd)
 
 	// If the sandbox exists with the right mounts, reuse it.
 	if existing != nil &&
-		(extra == "" || existing.HasWorkspace(extra)) &&
+		hasAllWorkspaces(existing, extras) &&
```
docker-agent: **[MEDIUM] `hasAllWorkspaces` is a subset check — sandbox with stale extra mounts is incorrectly reused**

`hasAllWorkspaces` returns `true` if every entry in the *requested* `extras` is present in the existing sandbox, but it does **not** verify the inverse: that the existing sandbox has no *additional* mounts beyond those requested. As a result:

1. First call: `Ensure(wd, extras=[A, B], ...)` → creates sandbox with mounts `{wd, A, B, configDir}`.
2. Second call: `Ensure(wd, extras=[A], ...)` → `hasAllWorkspaces([A])` returns `true` (A is present), `HasWorkspace(configDir)` is `true` → the sandbox is reused, even though mount `B` (from the previous run) remains active.

The stale `B` mount is still visible inside the sandbox, which can be a security/isolation concern — e.g. a prior run's kit directory (potentially containing secrets from a different agent invocation) remains mounted and readable in the new run.

Consider also checking that the existing sandbox has no *extra* mounts beyond `{wd, extras..., configDir}`:
```go
// In addition to the subset check, verify no superfluous mounts exist.
func hasExactWorkspaces(s *Existing, wd string, extras []string, configDir string) bool {
    required := map[string]bool{wd: true, configDir: true}
    for _, e := range extras {
        required[e] = true
    }
    for _, ws := range s.Workspaces {
        if !required[strings.TrimSuffix(ws, ":ro")] {
            return false
        }
    }
    return hasAllWorkspaces(s, extras)
}
```

docker-agent (commented):

Reviewed PR #2844 — *feat(sandbox): stage a docker-agent kit with skills and prompt files*

The kit-staging logic, symlink containment, atomic rename, secret redaction, hash-key canonicalisation, and test coverage all look solid. Two medium-severity issues were found in the refactored `sandbox.Ensure` path introduced by this PR.

pkg/skills/local.go
```diff
@@ -22,12 +32,22 @@ type localSearchPath struct {
...
+// host has staged a kit) only the kit's skills directory is scanned, since
+// the usual host paths don't exist inside the sandbox.
+//
+// Otherwise the search paths are:
 //
 //  1. Global directories (under $HOME).
 //  2. .claude/skills under cwd (Claude project format, flat).
 //  3. .agents/skills from the git root down to cwd (closest wins).
 func localSearchPaths() []localSearchPath {
+	if kit := os.Getenv(KitDirEnv); kit != "" {
```
docker-agent, line 47: **[MEDIUM] Silent skills degradation when `DOCKER_AGENT_KIT_DIR` points to a non-existent directory**

When `DOCKER_AGENT_KIT_DIR` is set, `localSearchPaths()` returns **only** the kit-based path and skips all default `~/.codex/skills`, `~/.claude/skills`, and `~/.agents/skills` fallbacks. If the kit's `skills/` subdirectory doesn't exist on disk (e.g., due to a staging failure, wrong mount path, or any `kit.Build` error that still let the run proceed), `loadSkillsRecursive` calls `filepath.WalkDir` on the missing path, receives an error on the first callback, swallows it with `return nil`, and returns an empty slice — with no log line or warning.

The agent then silently starts with zero skills, which is especially hard to diagnose because the only hint is the `DOCKER_AGENT_KIT_DIR` env var being set.

**Suggested fix**: add a `filepath.Stat` (or `EvalSymlinks`) check in `loadSkillsFromDir` / `loadSkillsRecursive` and emit a `slog.WarnContext` when the directory is missing and `DOCKER_AGENT_KIT_DIR` was the source of the path. Alternatively, log a warning in `localSearchPaths()` when the resolved kit skills dir does not exist.

docker-agent (commented):

One LIKELY medium-severity finding in the new sandbox kit code.

trungutt (approved):

PR #2845, opened 2026-05-21
test(server): make TestAttachedServer_DeleteSessionStopsEventStream more robust

dgageot:
## Problem

`TestAttachedServer_DeleteSessionStopsEventStream` was occasionally failing on CI with:

```
event source ctx was not cancelled when session was deleted
```

(timing out at exactly 2.00s on the `select` at `server_attached_test.go:148`.)

## Root cause

The test was racing against the SSE handler:

1. The test issues `GET /api/sessions/:id/events`.
2. `http.Client.Do` returns as soon as the server flushes the SSE response headers in `Server.sessionEvents` — **before** that handler reaches `sm.StreamEvents` → `eventSources.Load(...)`.
3. The test then immediately calls `sm.DeleteSession`, which removes the entry from `eventSources`.
4. If `DeleteSession` won the race, `StreamEvents` returned `false` without ever invoking the registered source closure, so its `<-ctx.Done(); close(sourceCtxDone)` body never ran and the 2 s timeout fired.

## Fix

`pkg/server/server_attached_test.go`:

- Add a `sourceStarted` channel that is closed at the top of the registered `EventSource` closure.
- Wait on `<-sourceStarted` *before* calling `sm.DeleteSession`, guaranteeing the SSE handler has actually invoked the source.
- Bump the two cancellation/wait timeouts from 2 s to 10 s for slow CI runners.

## Validation

- `go test -race -run TestAttachedServer_DeleteSessionStopsEventStream -count=50 ./pkg/server/...` — pass
- `go test -race -run TestAttachedServer_DeleteSessionStopsEventStream -count=200 ./pkg/server/...` — pass
- `go test -race -count=1 ./pkg/server/...` — pass
- `task lint` — 0 issues

### DIFF
--- a/pkg/server/server_attached_test.go
+++ b/pkg/server/server_attached_test.go
@@ -122,8 +122,10 @@ func TestAttachedServer_DeleteSessionStopsEventStream(t *testing.T) {
 	sm := NewSessionManager(ctx, config.Sources{}, store, 0, &config.RuntimeConfig{})
 	sm.AttachRuntime(sess.ID, &fakeRuntime{}, sess)
 
+	sourceStarted := make(chan struct{})
 	sourceCtxDone := make(chan struct{})
 	sm.RegisterEventSource(sess.ID, func(ctx context.Context, _ func(any)) {
+		close(sourceStarted)
 		<-ctx.Done()
 		close(sourceCtxDone)
 	})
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the race condition in `TestAttachedServer_DeleteSessionStopsEventStream`. By introducing a `sourceStarted` synchronization channel, the test now ensures the SSE handler has actually invoked the event source before `DeleteSession` is called — eliminating the window where deletion could win the race and leave the source context never cancelled.

The 2 s → 10 s timeout bumps are also appropriate for slow CI runners.

No bugs were found in the changed code. The drafter noted two low-severity observations (neither constitutes a real bug):

- `close(sourceStarted)` would panic on a double-invocation of the source closure, but the single-invocation contract of `StreamEvents` makes this a non-issue in practice.
- `time.After` in the `select` creates a timer that is not explicitly stopped, which is a minor resource concern only relevant in very long-running test suites.

Both are informational only and do not warrant changes.

trungutt (approved):

PR #2850, opened 2026-05-21
feat(skills): run `context: fork` slash commands as sub-sessions

dgageot:
## Problem

Typing `/<skill-name>` in the TUI for a skill declared with `context: fork` used to inline the SKILL.md body into the parent conversation as a regular user message and let the parent agent answer it directly. The fork flag was silently ignored on this path: no sub-session was created, the parent transcript was polluted with a giant `<skill name="…">…</skill>` block, and the skill's own `allowed-tools` / `description` had no effect.

## Solution

Route fork-mode slash commands through the same machinery the `run_skill` tool uses, so the skill runs in an isolated child session and control returns to the parent when it ends:

- `Runtime.RunSkillFork` is now a public entry point shared by the `run_skill` tool and the new App-level slash dispatcher; it handles lookup, fork-mode validation, content + placeholder expansion, optional model override, and `runForwarding`.
- `App.SkillCommandFork` classifies user input as a fork-skill invocation; `App.RunSkillFork` drives the runtime and forwards events through the App's bus with the same context/cancellation discipline as `App.Run`.
- `App.ResolveSkillCommand` short-circuits for fork-mode skills so `ResolveInput` never returns the inlined body for them.
- The chat page calls `SkillCommandFork` first and falls back to the existing inline path for non-fork skills, agent commands and plain messages.

## Prompt shape

The skill body is now framed as the **first user message** of the sub-session (matching the inline `<skill name="…">` envelope), not as an extra system message — otherwise the agent's invariant system prompt would drown it out and the model would ignore the instructions. The YAML frontmatter is stripped because the runtime has already consumed it; passing it through just adds noise. `` !`shell` `` placeholders continue to be expanded by the existing `ReadSkillContent` path, while plain backtick markdown such as `` `git commit -q -m "<subject>"` `` passes through verbatim.

## Tests

- `pkg/app/skills_fork_test.go` — slash-command detection, fork dispatch contract (parent session not mutated, args forwarded), preserved inline path for non-fork skills, and a full-flow non-fork e2e test.
- `pkg/tools/builtin/skills/build_user_message_test.go` — user-message envelope, frontmatter stripping, system-message contract (no `transfer_task` boilerplate), and the markdown pass-through.

`task lint` clean, `task test` green, `-race` clean across the touched packages.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -28,6 +28,7 @@ import (
 	"github.com/docker/docker-agent/pkg/shellpath"
 	"github.com/docker/docker-agent/pkg/skills"
 	"github.com/docker/docker-agent/pkg/tools"
+	skillstool "github.com/docker/docker-agent/pkg/tools/builtin/skills"
 	mcptools "github.com/docker/docker-agent/pkg/tools/mcp"
 	"github.com/docker/docker-agent/pkg/tui/messages"
 )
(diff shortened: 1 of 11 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26215595375).

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26215595662).

trungutt (approved):

PR #2852, opened 2026-05-21
refactor(api): fold model override into runAgent request body

dgageot:
## Summary

Folds the per-agent model override into the SSE turn-runner endpoint
`POST /api/sessions/:id/agent/:agent[/:agent_name]` and removes the
dedicated `PATCH/POST /api/sessions/:id/model` routes.

The dedicated routes were confusing: clients had to PATCH the model and
then POST the messages in two separate calls, with a race window in
between. This change makes model selection an atomic part of starting
the next turn.

## Wire-level changes

- **`POST /api/sessions/:id/agent/:agent[/:agent_name]`** body changes
  from a bare ` + "`[]Message`" + ` to ` + "`{ \"messages\": [...], \"model\": \"optional/ref\" }`" + `.
  A non-empty ` + "`model`" + ` is applied as a **persistent** per-agent override
  on the session before the user message is appended; empty leaves the
  current override untouched. Validation happens before the message is
  recorded so a bad ref does not leave an orphaned user message in
  history.
- **` + "`PATCH /api/sessions/:id/model`" + `** and **` + "`POST /api/sessions/:id/model`" + `**
  removed (along with ` + "`api.SetSessionModelRequest`" + `/` + "`Response`" + `).
- **` + "`GET /api/sessions/:id/models`" + `** unchanged.

This is a **breaking** change for HTTP clients of ` + "`docker agent serve api`" + `:
- the bare-array body is no longer accepted, and
- the dedicated model endpoint is gone.

The TUI and other in-process callers are unaffected.

## Internal changes

- ` + "`SessionManager.RunSession`" + ` gained a ` + "`modelOverride string`" + ` parameter.
  When set, it is applied via the new ` + "`applyRunModelOverride`" + ` helper
  (validates with the runtime, mutates in-memory session fields, returns
  an ` + "`undo`" + ` closure that rolls the runtime override back if the subsequent
  ` + "`UpdateSession`" + ` store-write fails).
- ` + "`SessionManager.SetSessionAgentModel`" + ` is kept for in-process callers.
  The unused ` + "`activeRuntimes.modelSwitch`" + ` mutex was removed.
- ` + "`runtime.Client.RunAgent`" + `/` + "`RunAgentWithAgentName`" + ` gained a ` + "`model string`" + `
  parameter; ` + "`Client.SetAgentModel`" + ` was removed; the ` + "`RemoteClient`" + `
  interface was updated to match.
- ` + "`RemoteRuntime.SetAgentModel`" + ` now buffers the ref locally
  (` + "`pendingModelOverride`" + ` guarded by ` + "`pendingMu`" + `) and forwards it on the
  next ` + "`RunStream`" + ` rather than calling a separate endpoint. The buffer
  is only cleared **after** the server accepts the request, so a
  transient HTTP failure no longer silently drops the user's switch
  (with a CAS-style guard so a concurrent ` + "`SetAgentModel`" + ` queued mid-dispatch
  is not overwritten).

## Docs

` + "`docs/features/api-server/index.md`" + ` updated with the new request shape and
new ` + "`curl`" + ` examples.

## Validation

- ` + "`task build`" + ` ✅
- ` + "`task test`" + ` ✅
- ` + "`task lint`" + ` ✅ (golangci-lint, custom lint checker, ` + "`go mod tidy`" + `)

## Commits

1. ` + "`refactor: fold model override into runAgent request body`" + ` — main refactor.
2. ` + "`fix(runtime): retain pending model override when RunStream dispatch fails`" + ` —
   addresses a race uncovered during self-review where the ` + "`RemoteRuntime`" + `
   buffer was cleared before the HTTP dispatch could fail, silently
   swallowing the user's model switch on a transient error.

### DIFF
--- a/docs/features/api-server/index.md
+++ b/docs/features/api-server/index.md
@@ -53,8 +53,6 @@ All endpoints are under the `/api` prefix.
 | `POST`   | `/api/sessions/:id/steer`           | Inject messages into a running turn (pre-empts current) |
 | `POST`   | `/api/sessions/:id/followup`        | Enqueue messages to run after the current turn finishes |
 | `GET`    | `/api/sessions/:id/models`          | List available models for the session's current agent   |
-| `PATCH`  | `/api/sessions/:id/model`           | Set or clear the agent's model override                 |
-| `POST`   | `/api/sessions/:id/model`           | Set or clear the agent's model override (backward compat with RemoteRuntime) |
 
 ### Agent Execution
 
@@ -77,19 +75,19 @@ All endpoints are under the `/api` prefix.
 # Run the root agent:
 curl -N -X POST http://localhost:8080/api/sessions/$SID/agent/my-assistant \
   -H "Content-Type: application/json" \
-  -d '[{"role": "user", "content": "Hello!"}]'
+  -d '{"messages":[{"role": "user", "content": "Hello!"}]}'
 
 # Multi-agent config: team.yaml (defines agents: root, coder, reviewer)
 # Start: docker agent serve api team.yaml
 # Run the root agent:
 curl -N -X POST http://localhost:8080/api/sessions/$SID/agent/team \
   -H "Content-Type: application/json" \
-  -d '[{"role": "user", "content": "Review this PR"}]'
+  -d '{"messages":[{"role": "user", "content": "Review this PR"}]}'
 
 # Run a specific sub-agent (reviewer):
 curl -N -X POST http://localhost:8080/api/sessions/$SID/agent/team/reviewer \
   -H "Content-Type: application/json" \
-  -d '[{"role": "user", "content": "Review this PR"}]'
+  -d '{"messages":[{"role": "user", "content": "Review this PR"}]}'
 ```
 
 ### Health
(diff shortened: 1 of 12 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor cleanly consolidates model-override handling into the turn-runner request body and addresses the race window that existed with the separate `PATCH /model` + `POST /messages` two-step. No bugs were identified in the changed code.

**Reviewed areas:**
- `applyRunModelOverride` undo closure logic in `session_manager.go` — correctly rolls back the in-memory override when the subsequent `UpdateSession` store-write fails; no double-apply or missing unlock paths found.
- CAS-style guard in `remote_runtime.go` (`pendingModelOverride` + `pendingMu`) — the buffer is cleared **after** server acceptance, and a concurrent `SetAgentModel` queued mid-dispatch is correctly preserved.
- `RemoteClient` interface and all implementing types (`RemoteRuntime`, stub client in contract tests) — updated consistently; `SetAgentModel` removal propagated everywhere.
- New request body parsing (`RunAgentRequest`) — error propagation and validation order (validate before appending user message) looks correct.
- HTTP error handling path — on transient dispatch failure the pending override is retained for the next call.

trungutt (approved):

PR #2854, opened 2026-05-21
fix: reduce retained tool output memory

dgageot:
- wait for lifecycle supervisor watcher goroutines during shutdown
- bound retained tool output in session history with a regression test
- spool large MCP media payloads to disk instead of retaining inline base64
- slim retained TUI tool results and avoid duplicating file contents in metadata

### DIFF
--- a/pkg/runtime/toolexec/dispatcher.go
+++ b/pkg/runtime/toolexec/dispatcher.go
@@ -724,13 +724,21 @@ func buildMultiContent(text string, images []tools.MediaContent) []chat.MessageP
 	parts := make([]chat.MessagePart, 0, 1+len(images))
 	parts = append(parts, chat.MessagePart{Type: chat.MessagePartTypeText, Text: text})
 	for _, img := range images {
-		parts = append(parts, chat.MessagePart{
-			Type: chat.MessagePartTypeImageURL,
-			ImageURL: &chat.MessageImageURL{
-				URL:    "data:" + img.MimeType + ";base64," + img.Data,
-				Detail: chat.ImageURLDetailAuto,
-			},
-		})
+		switch {
+		case img.FilePath != "":
+			parts = append(parts, chat.MessagePart{
+				Type: chat.MessagePartTypeText,
+				Text: fmt.Sprintf("[image saved to %s (%s)]", img.FilePath, img.MimeType),
+			})
+		case img.Data != "":
+			parts = append(parts, chat.MessagePart{
+				Type: chat.MessagePartTypeImageURL,
+				ImageURL: &chat.MessageImageURL{
+					URL:    "data:" + img.MimeType + ";base64," + img.Data,
+					Detail: chat.ImageURLDetailAuto,
+				},
+			})
+		}
 	}
 	return parts
 }
--- a/pkg/tools/lifecycle/supervisor_test.go
+++ b/pkg/tools/lifecycle/supervisor_test.go
@@ -17,18 +17,39 @@ import (
 // fakeSession is a controllable session: its Wait blocks until either
 // Close is called or fail is invoked.
 type fakeSession struct {
-	mu     sync.Mutex
-	closed bool
-	failCh chan error
+	mu       sync.Mutex
+	closed   bool
+	waitDone atomic.Bool   // set true after Wait returns
+	waiting  chan struct{} // closed once Wait has parked on failCh
+	waitOnce sync.Once
+	failCh   chan error
 }
 
 func newFakeSession() *fakeSession {
-	return &fakeSession{failCh: make(chan error, 1)}
+	return &fakeSession{
+		waiting: make(chan struct{}),
+		failCh:  make(chan error, 1),
+	}
 }
 
 func (f *fakeSession) Wait() error {
+	f.waitOnce.Do(func() { close(f.waiting) })
 	err := <-f.failCh
+	f.waitDone.Store(true)
 	return err
+}
+
+// waitParked blocks until the watcher goroutine has entered sess.Wait().
+// Used by tests that need to exercise Stop against an actively-blocking
+// watcher rather than the racy connect-then-stop path where the watcher
+// could exit before parking.
+func (f *fakeSession) waitParked(t *testing.T) {
+	t.Helper()
+	select {
+	case <-f.waiting:
+	case <-time.After(time.Second):
+		t.Fatal("watcher did not enter Wait()")
+	}
 }
 
 func (f *fakeSession) Close(context.Context) error {
@@ -457,4 +478,55 @@ func TestBackoff_Jitter(t *testing.T) {
 	// random ≈ 0 → -50% offset
 	d = lifecycle.ExportedBackoffDelay(b, 0, func() float64 { return 0 })
 	assert.Check(t, d == 50*time.Millisecond)
+}
+
+func TestSupervisor_StopWaitsForWatcher(t *testing.T) {
+	t.Parallel()
+
+	sess := newFakeSession()
+	c := newScriptedConnector(scriptStep{session: sess})
+	s := lifecycle.New("test", c, lifecycle.Policy{})
+
+	assert.NilError(t, s.Start(t.Context()))
+	sess.waitParked(t)
+
+	assert.NilError(t, s.Stop(t.Context()))
+	assert.Check(t, is.Equal(s.State().State, lifecycle.StateStopped))
+
+	// Stop must not return until the watcher has observed Wait() unblock.
+	assert.Check(t, sess.waitDone.Load(), "Stop returned before watcher's Wait() completed")
+}
+
+// TestSupervisor_StopConcurrent exercises the s.stopping guard: several
+// goroutines call Stop concurrently while the watcher is live in
+// sess.Wait(). All calls must return without error and observe a
+// fully-shut-down supervisor.
+func TestSupervisor_StopConcurrent(t *testing.T) {
+	t.Parallel()
+
+	sess := newFakeSession()
+	c := newScriptedConnector(scriptStep{session: sess})
+	s := lifecycle.New("test", c, lifecycle.Policy{})
+
+	assert.NilError(t, s.Start(t.Context()))
+	sess.waitParked(t)
+
+	const n = 4
+	errs := make(chan error, n)
+	var wg sync.WaitGroup
+	wg.Add(n)
+	for range n {
+		go func() {
+			defer wg.Done()
+			errs <- s.Stop(t.Context())
+		}()
+	}
+	wg.Wait()
+	close(errs)
+
+	for err := range errs {
+		assert.NilError(t, err)
+	}
+	assert.Check(t, is.Equal(s.State().State, lifecycle.StateStopped))
+	assert.Check(t, sess.waitDone.Load(), "a Stop returned before watcher's Wait() completed")
 }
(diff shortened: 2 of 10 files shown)


pkg/runtime/toolexec/dispatcher.go
```diff
@@ -720,6 +720,20 @@ func (c *call) recordToolResponse(res *tools.ToolCallResult) {
 // buildMultiContent attaches inline images to a tool response as a
```
rumpl, line 720: The comment is for the wrong function now

pkg/runtime/toolexec/dispatcher.go
```diff
@@ -720,6 +720,20 @@ func (c *call) recordToolResponse(res *tools.ToolCallResult) {
...
 // MultiContent payload, following the data-URL convention expected by
 // chat clients.
+func retainedToolOutput(output string, maxOldToolCallTokens int) string {
```
rumpl: Can you explain what this is please?

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26225206596).

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26230588872).

pkg/tools/mcp/mcp.go
```diff
@@ -760,12 +763,54 @@ func processMCPContent(toolResult *mcp.CallToolResult) *tools.ToolCallResult {
...
+		media.Data = base64.StdEncoding.EncodeToString(data)
+		return media
+	}
+	media.FilePath = path
+	return media
+}
+
+func writeMediaFile(data []byte, mimeType string) (string, error) {
+	dir, err := os.MkdirTemp("", "docker-agent-mcp-media-*")
+	if err != nil {
```
aheritier, line 808: **[BLOCKING] Disk leak — these temp directories are never removed.**

`os.MkdirTemp` creates a directory in `os.TempDir()` on every call to `writeMediaFile`, but nothing cleans it up when the `ToolCallResult` is discarded or the session ends. Over a long session with many large media tool responses this will exhaust disk space.

Suggested fix: introduce a session-scoped temp directory (created once, removed on `Stop`) instead of a per-payload directory. Alternatively, register a cleanup closure on the `Toolset` lifecycle the same way `GatewayToolset` does today.

pkg/tools/mcp/mcp_test.go
```diff
@@ -536,3 +539,19 @@ func TestCallToolRecoversFromErrSessionMissing(t *testing.T) {
...
+	large := bytes.Repeat([]byte("x"), maxInlineMediaBytes+1)
+	result := processMCPContent(callToolResult(&mcp.ImageContent{Data: large, MIMEType: "image/png"}))
+
+	require.Len(t, result.Images, 1)
+	img := result.Images[0]
+	assert.Empty(t, img.Data)
+	assert.Equal(t, "image/png", img.MimeType)
+	require.NotEmpty(t, img.FilePath)
+	defer os.RemoveAll(filepath.Dir(img.FilePath))
+
```
aheritier, line 599: The test correctly cleans up with `defer os.RemoveAll(...)`, but this highlights that the production code does not. This `defer` will be unnecessary once a lifecycle cleanup hook is added on the production side.

pkg/tui/components/reasoningblock/reasoningblock.go
```diff
@@ -240,7 +240,7 @@ func (m *Model) UpdateToolResult(toolCallID, content string, status types.ToolSt
 
 		entry.msg.Content = strings.ReplaceAll(content, "\t", "    ")
 		entry.msg.ToolStatus = status
-		entry.msg.ToolResult = result
+		entry.msg.ToolResult = result.WithoutPayload()
```
aheritier, line 244: **[Non-blocking] Double `WithoutPayload()` call.**

The caller in `messages.go` already passes `msg.Result.WithoutPayload()` here, so `result` is already a stripped copy. Calling `result.WithoutPayload()` a second time is idempotent but redundant and may confuse future readers. Either remove `WithoutPayload()` from this line (since it's already stripped at the call site), or remove it from the call site in `messages.go` and let `UpdateToolResult` own the stripping — but not both.

pkg/tools/lifecycle/supervisor_test.go
```diff
@@ -458,3 +458,18 @@ func TestBackoff_Jitter(t *testing.T) {
...
+	sess := newFakeSession()
+	c := newScriptedConnector(scriptStep{session: sess})
+	s := lifecycle.New("test", c, lifecycle.Policy{})
+
+	assert.NilError(t, s.Start(t.Context()))
+
+	start := time.Now()
+	assert.NilError(t, s.Stop(t.Context()))
+	assert.Check(t, time.Since(start) < time.Second)
+	assert.Check(t, is.Equal(s.State().State, lifecycle.StateStopped))
```
aheritier, line 494: **[Non-blocking] Brittle timing assertion and weak concurrency coverage.**

This `< time.Second` bound will be flaky under heavy CI load. More importantly, the test only covers the sequential Start→Stop path, not the concurrent-Stop path that the `s.stopping` guard was added to fix. A test with two goroutines calling `Stop` concurrently while the watcher is alive (e.g. blocked in `sess.Wait()`) would be a stronger regression guard. Consider using `goleak` to assert no goroutines are left behind.

aheritier (changes requested):
The PR attacks four distinct memory sources and the implementation is mostly clean. The `WithoutPayload` slimming, the `watchDone` shutdown guarantee, and the `ReadFileMeta.Content` removal are all sound. One blocking issue must be resolved before merge.

## Blocking

**Disk leak: spooled MCP media files are never cleaned up.**
`writeMediaFile` creates temp directories under `os.TempDir()` with prefix `docker-agent-mcp-media-*` but nothing removes them when the session ends or the `ToolCallResult` is GC'd. The PR itself calls `defer os.RemoveAll(filepath.Dir(img.FilePath))` in the test, which highlights the production gap. Over a long session with many large media responses this trades the memory leak for a disk leak.

Fix options:
- Register a cleanup func on the `GatewayToolset` lifecycle (cf. `cleanUp` pattern already used there), called on `Stop`.
- Use a single session-scoped temp directory created once at `Toolset` construction and removed atomically on teardown — simpler than one dir per payload and easier to reason about.

## Non-blocking

1. **Double `WithoutPayload()` in `reasoningblock.go`** — `messages.go` already calls `msg.Result.WithoutPayload()` before passing the result to `UpdateToolResult`; `reasoningblock.go` then calls `result.WithoutPayload()` again internally. Harmless (idempotent on a stripped result) but confusing. Choose one call site and remove the other.

2. **Missing test: disk-write fallback in `encodeMedia`** — the `slog.Warn` + inline fallback branch is untested. Worth covering with an injected error (e.g. a package-level `writeMediaFile` var that tests can replace).

3. **Missing test: threshold boundary** — `TestProcessMCPContentSpoolsLargeMedia` only checks `maxInlineMediaBytes+1`. Add a case for exactly `maxInlineMediaBytes` bytes → inline, and `maxInlineMediaBytes+1` → spooled.

4. **`TestSupervisor_StopWaitsForWatcher` is weak coverage for the concurrency fix** — the test exercises the sequential path (Start → Stop), not the concurrent-Stop path that the `s.stopping` guard is specifically designed to handle. A test calling `Stop` twice concurrently from two goroutines while the watcher is live would be a stronger regression guard. The `< time.Second` timing assertion is also brittle under heavy CI load; `goleak` would be more idiomatic.

5. **`readImageFile` in `filesystem.go` still inlines large local images** — up to `chat.MaxImageBytes` (~4.5 MB) as base64. This is inconsistent with the new MCP spooling strategy. Out of scope for this PR, but worth a follow-up issue.

docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed `fix: reduce retained tool output memory` (PR #2854).

**Scope:** 10 files changed — watcher shutdown sequencing (`supervisor.go`), metadata deduplication (`filesystem.go`), MCP media spooling to disk (`mcp.go`), TUI payload slimming (`messages.go`), and supporting tests.

**Verification summary:** The drafter raised 5 hypotheses (3 medium, 2 low). After code-level verification, all 5 were **dismissed**:

| # | File | Hypothesis | Verdict | Reason |
|---|------|-----------|---------|--------|
| 1 | `supervisor.go` | `RestartAndWait` blocks indefinitely on ctx cancel during backoff | ✅ DISMISSED | `RestartAndWait` is not changed by this PR; the new `watchDone` mechanism applies to `Stop()` only. No regression introduced. |
| 2 | `filesystem.go` | TOCTOU between `resolveAndCheckPath` and `readFile` in WalkDir | ✅ DISMISSED | The diff only removes a duplicate metadata field and stale assignment. No I/O path changed; TOCTOU was already addressed in prior commit `eb7bb600`. |
| 3 | `messages.go` | Data race on `renderedItems` LRU cache between `View()`/`Update()` | ✅ DISMISSED | bubbletea runs `View()` and `Update()` on the same single-threaded loop goroutine — no concurrency. The diff only swaps `msg.Result` → `msg.Result.WithoutPayload()` with no new cache access. |
| 4 | `supervisor.go` | Panic in `OnFailed`/`OnRestart` callbacks uncaught in watcher | ✅ DISMISSED | Callback invocations are unchanged; lack of panic recovery is pre-existing, not introduced by this PR. |
| 5 | `filesystem.go` | Repair loop capped at 3 passes may silently fail | ✅ DISMISSED | No such loop exists anywhere in the changed code. |

**Positive observations:**
- The `watchDone` channel approach in `supervisor.go` is clean and correct: created before the watcher goroutine spawns, closed in the goroutine's `defer`, and waited on by both the first and concurrent `Stop()` callers. The `stopping` guard path now correctly blocks instead of silently returning, which is the right fix.
- Spooling MCP media to disk via `ensureMediaDir` / `cleanupMediaDir` is well-structured, with proper fallback to inline base64 on spool failure and cleanup tied to `Stop()`.
- `WithoutPayload()` in `messages.go` correctly avoids retaining large payloads in the TUI state.
- The new tests (`TestSupervisor_StopWaitsForWatcher`, `TestSupervisor_StopConcurrent`) are solid — `waitParked` eliminates the racy connect-then-stop path and the concurrent variant exercises the `stopping` guard properly.

aheritier (approved):

pkg/tools/mcp/mcp.go
```diff
@@ -760,12 +772,94 @@ func processMCPContent(toolResult *mcp.CallToolResult) *tools.ToolCallResult {
...
 }
 
-// encodeMedia re-encodes raw bytes (as decoded by the MCP SDK) back to base64
-// for our internal MediaContent representation.
-func encodeMedia(data []byte, mimeType string) tools.MediaContent {
-	return tools.MediaContent{
-		Data:     base64.StdEncoding.EncodeToString(data),
-		MimeType: mimeType,
+// encodeMedia keeps small payloads inline and spools larger ones to disk so the
+// session and TUI do not retain duplicate base64 copies. Spooled files live
```
rumpl, line 776: If this is a TUI issue then the code should live in the TUI and not deep down in our tools code. Especiually since a filesystem can go away while you are not looking

PR #2855, opened 2026-05-21
refactor(remote): faster, safer, simpler OCI pull path

dgageot:
Reviewing how we pull agent YAMLs from OCI artifacts surfaced a few wins. Each commit is independently reviewable.

### `4242b42b4` refactor(remote): detect proxy socket errors via type, not strings

`isProxySocketError` decides whether the Docker Desktop proxy is unavailable so the transport can fall back to a direct connection. It used substring matching on `err.Error()` (`"no such file or directory"`, `"connection refused"`, `"dial unix"`, …), which breaks across Go versions and locales and gives false positives whenever any error text happens to contain those phrases.

Use `errors.Is` against `syscall.ENOENT` / `syscall.ECONNREFUSED` and `errors.As` against `*net.OpError` (`Op == "proxyconnect"` or unix-socket dial). Tests now construct realistic typed error chains, including the `*url.Error` wrapper `http.Client` adds in production.

### `169d6dff8` refactor(remote): collapse OCI agent read path into `PullAgent`

The flow to load an agent YAML from an OCI reference was split across three places: `cmd/root/pull.go` (Pull + open store + GetArtifact), `ociSource.Read` (Pull + load + corruption-aware retry + cache fallback), and `remote.Pull` itself (its own cache check). Each duplicated parts of the others.

Introduce `remote.PullAgent(ctx, ref, force) ([]byte, digest, error)` as the single entry point: it handles the digest-ref shortcut, the network-error fallback to cache, and the one-shot recovery on store corruption. `ociSource.Read` shrinks to a single call.

Also: skip the upfront `crane.Digest` HEAD when the cache is empty. With no local copy to compare against we'd have to `crane.Pull` anyway, so the HEAD is pure latency. Cold pulls drop from two manifest round-trips to one.

### `0ffbe5895` fix(remote): keep `share pull` strict on registry errors

Routing the `share pull` command through `PullAgent` meant a registry 404 or auth failure would silently overwrite the on-disk YAML with a stale cached copy. `PullAgent`'s cache-fallback policy is the right behaviour for `ociSource.Read` (running an agent shouldn't break when the registry blips) but the wrong behaviour for an explicit user-driven refresh. `share pull` now calls `remote.Pull` directly so registry failures surface as errors.

### `385816cb5` fix(remote): treat unannotated cache as cache-miss, not as terminal error

Two related defense-in-depth fixes:

1. `Pull` bailed out with a permanent error if local metadata lacked the cagent annotation, before contacting the registry. A user with a stale or bad cache entry on a tag was stuck until they manually wiped the store, even after the remote tag was updated to a valid agent artifact. Treat an unannotated cache entry as a cache miss so `Pull` falls through to `crane.Pull`, which re-validates the freshly fetched manifest and overwrites the bad entry.
2. `PullAgent`'s cache fast-paths (digest-ref shortcut and the network-error fallback) returned cached bytes without re-checking the annotation, while `Pull` validates it on the registry path. `readCached` now refuses entries missing the annotation, so the cache cannot be used to bypass that validation.

Tests:
- `TestReadCached` — missing / unannotated / annotated cases.
- `TestPull_RecoversFromUnannotatedCache` — uses an in-memory OCI registry to verify a poisoned cache entry no longer blocks a valid re-pull.
- `TestPullAgent_DigestRefDoesNotServeUnannotatedCache` — verifies the digest-ref fast path no longer trusts an unannotated cache entry.

## Validation

- `task lint` — clean
- `task test` — clean (130 packages)

### DIFF
--- a/pkg/remote/pull.go
+++ b/pkg/remote/pull.go
@@ -52,13 +52,10 @@ func Pull(ctx context.Context, registryRef string, force bool, opts ...crane.Opt
 	localRef := ref.Context().RepositoryStr() + separator(ref) + ref.Identifier()
 
 	// Cache check: only worth a HEAD round-trip when we actually have a
-	// local copy to compare against. When the cache is empty we'd have to
-	// pull anyway, so skip the digest probe entirely.
+	// usable local copy. When the cache is empty or contains a non-agent
+	// artifact we'd have to pull anyway, so skip the digest probe.
 	if !force {
-		if meta, metaErr := store.GetArtifactMetadata(localRef); metaErr == nil {
-			if !hasCagentAnnotation(meta.Annotations) {
-				return "", fmt.Errorf("artifact %s found in store wasn't created by `docker agent share push`\nTry to push again with `docker agent share push`", localRef)
-			}
+		if meta, metaErr := store.GetArtifactMetadata(localRef); metaErr == nil && hasCagentAnnotation(meta.Annotations) {
 			remoteDigest, err := crane.Digest(ref.String(), opts...)
 			if err != nil {
 				return "", fmt.Errorf("resolving remote digest for %s: %w", registryRef, err)
@@ -154,11 +151,16 @@ func PullAgent(ctx context.Context, registryRef string, force bool) ([]byte, str
 	return []byte(data), digest, nil
 }
 
-// readCached returns the cached YAML bytes and digest for storeKey, or
-// ok=false if no usable cached copy is present.
+// readCached returns the cached YAML bytes and digest for storeKey. It
+// returns ok=false when no usable copy exists, including when the cached
+// metadata lacks the cagent annotation — mirroring the validation Pull
+// applies on the registry path so the cache cannot be used to bypass it.
 func readCached(store *content.Store, storeKey string) ([]byte, string, bool) {
 	meta, err := store.GetArtifactMetadata(storeKey)
 	if err != nil {
+		return nil, "", false
+	}
+	if !hasCagentAnnotation(meta.Annotations) {
 		return nil, "", false
 	}
 	data, err := store.GetArtifact(storeKey)
--- a/pkg/remote/pull.go
+++ b/pkg/remote/pull.go
@@ -2,7 +2,9 @@ package remote
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"log/slog"
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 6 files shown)


docker-agent (commented):

Reviewed the OCI pull path refactor. Found 2 issues in the changed code.

- `isProxySocketError`: typed error detection via `errors.Is`/`errors.As` (replaces string matching)
- `PullAgent`: new unified entry point with cache fallback on network errors + corruption recovery
- `share pull`: reverted to use `Pull` directly for strict registry error propagation
- `readCached`: annotation guard added to prevent bypassing validation via cache

pkg/remote/pull.go
```diff
@@ -85,6 +87,89 @@ func Pull(ctx context.Context, registryRef string, force bool, opts ...crane.Opt
...
+		return nil, "", fmt.Errorf("failed to pull OCI image %s: %w", registryRef, pullErr)
+	}
+
+	data, err := store.GetArtifact(storeKey)
+	if err == nil {
+		return []byte(data), digest, nil
+	}
+
+	// Store went missing right after a successful pull. Re-pull once.
+	if !errors.Is(err, content.ErrStoreCorrupted) {
```
docker-agent, line 139: **[MEDIUM] Corruption recovery re-pull is skipped for non-`ErrStoreCorrupted` store errors**

After a successful `Pull`, `PullAgent` calls `store.GetArtifact(storeKey)` and only triggers the forced re-pull when the error satisfies `errors.Is(err, content.ErrStoreCorrupted)`. However, `GetArtifact` passes through errors from `GetArtifactImage` without wrapping them — specifically, `resolveIdentifier` can return `ErrInvalidDigest` (or another non-`ErrStoreCorrupted` error) if the refs file is corrupt. In that case:

```go
if !errors.Is(err, content.ErrStoreCorrupted) {
    return nil, "", fmt.Errorf("loading artifact from store: %w", err)  // fatal — no recovery re-pull
}
```

A user whose refs file is corrupted would be permanently stuck even though a forced re-pull would regenerate the file and fix the situation. The recovery guard should be broader — either use `errors.Is(err, content.ErrStoreCorrupted) || errors.As(err, &content.ErrInvalidDigest)`, or simply attempt the re-pull for **any** post-pull `GetArtifact` error (since `Pull` just succeeded, the only plausible cause is store corruption).

pkg/remote/transport.go
```diff
@@ -121,29 +122,33 @@ func (f *fallbackTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
-// This includes:
-// - "no such file or directory" - socket file was deleted
-// - "connection refused" - socket exists but nothing is listening
-// - "dial unix" errors - general Unix socket connection failures
+// isProxySocketError reports whether err indicates the Docker Desktop proxy
+// socket is unavailable: socket file gone, nothing listening on it, or any
+// failure dialing a Unix socket. Detection is type-based (net.OpError /
+// syscall errno) rather than string matching, so it stays correct across
+// Go versions and locales.
 func isProxySocketError(err error) bool {
```
docker-agent, line 130: **[LOW] `isProxySocketError` refactor may drop coverage for two legacy error patterns**

The old string-based implementation matched two patterns that the new typed implementation no longer covers:

1. **`"unix socket"`** — any error whose message contained this substring was caught. This is not covered by `net.OpError{Op: "dial", Net: "unix"}` (which only matches when the error arrives as a typed `*net.OpError`).
2. **`"proxyconnect tcp"`** as a plain string — if a custom `http.RoundTripper` or proxy middleware returns a plain `errors.New("proxyconnect tcp: ...")`, it will not be unwrapped to a `*net.OpError` and will be missed.

For standard Go `net/http` proxy errors these paths are likely unreachable (the stdlib always wraps in `*net.OpError`). However, if any code in the transport stack (e.g., WSL path, custom DialContext) ever produces a bare error string, the proxy fallback will silently fail to trigger, causing all subsequent requests to fail instead of falling back to direct.

Suggest adding a test that generates a `net.OpError{Op: "dial", Net: "unix"}` (without a `proxyconnect` wrapper) to confirm the typed path is exercised, or adding a comment explaining why `"unix socket"` string errors cannot occur in practice.

PR #2857, opened 2026-05-21
feat(eval): add input_id passthrough field to preserve caller-supplied correlation ID in results output

### LINKED ISSUE #2853: eval: preserve input eval JSON "id" as session ID in results output
## Summary

`docker agent eval` does not preserve the `"id"` field from input eval JSON files in the results output. Instead, it generates a new UUID for each session, discarding any `"id"` value supplied by the caller. The `"title"` field **is** preserved correctly — only `"id"` is dropped.

---

## Steps to Reproduce

```bash
mkdir -p /tmp/probe-eval /tmp/probe-out

cat > /tmp/probe-eval/test.json <<'EOF'
{
  "id": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
  "title": "Probe eval",
  "evals": { "relevance": ["Answers the question"] },
  "messages": [{ "message": { "agentName": "", "message": { "role": "user", "content": "What is Docker?" } } }]
}
EOF

docker agent eval ~/Workspace/gordon/assistant/gordon_dev.yaml   /tmp/probe-eval --output /tmp/probe-out --concurrency 1

jq '.sessions[0] | {id, title}' /tmp/probe-out/*.json
```

---

## Current Behaviour

The output session receives a freshly generated UUID, ignoring the `"id"` present in the input file:

```json
{
  "id": "91907fe1-cd72-4e88-b1a5-1b439675f7c5",
  "title": "Probe eval"
}
```

---

## Expected Behaviour

When an eval JSON file contains an `"id"` field, `docker agent eval` should carry that value through to the corresponding session entry in the results output:

```json
{
  "id": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
  "title": "Probe eval"
}
```

If no `"id"` is present in the input file, the current behaviour (auto-generating a UUID) is acceptable.

---

## Motivation / Impact

Any downstream system that writes eval JSON files with a known `"id"` — for example to correlate results back to a record in a source database — is broken by this behaviour. Because the output session ID is an unrelated UUID, there is no reliable way to map a result back to the originating eval record without resorting to fragile heuristics (e.g. matching on `"title"`, which is not guaranteed to be unique).

Preserving the caller-supplied `"id"` is a minimal, non-breaking change: it only affects sessions whose input file already carries an `"id"`, and leaves auto-generation in place for all other cases.

hamza-jeddad:
# Plan: Preserve caller-supplied `"input_id"` from input eval JSON in results output

## 1. Goal Restatement

Each eval input is **one `.json` file** in a directory. When a file contains a top-level `"input_id"` field, that value must be carried through untouched to the corresponding session entry in the results output (the `*.json` file and the SQLite `.db` file). The session's own `"id"` (a random UUID) is **never touched** — it is always freshly generated as today. `"input_id"` is simply a new passthrough field. If the input file has no `"input_id"`, the field is absent from the output — no change to existing behaviour.

---

## 2. User-side: Before vs After

| | Before (broken) | After (fixed) |
|---|---|---|
| User writes eval file with `"input_id": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"` | ✅ file is accepted | ✅ file is accepted |
| User runs `docker agent eval` | ✅ runs fine | ✅ runs fine |
| Output session `"id"` | `91907fe1-cd72-4e88-b1a5-1b439675f7c5` (random UUID) | `91907fe1-cd72-4e88-b1a5-1b439675f7c5` (same random UUID) |
| Output session `"input_id"` | ❌ missing — field is dropped entirely | ✅ `"aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"` — carried through from the input file |
| User matches output back to their database record | ❌ impossible — `input_id` is gone | ✅ IDs match, correlation works |
| User writes eval file **without** `"input_id"` | No `input_id` in output | Same — no `input_id` in output ✅ |

---

## 3. Root Cause

`loadEvalSessions` deserializes each `.json` file into `session.Session` via `json.Unmarshal`. `session.Session` has no `InputID` field, so `"input_id"` from the file is silently discarded right at load time and never reaches the output.

---

## 4. Affected Files

| File | What changes |
|---|---|
| `pkg/session/session.go` | Add `InputID string` field (`` `json:"input_id,omitempty"` ``) to the `Session` struct. |
| `pkg/evaluation/eval.go` | After `SessionFromEvents`, copy `evalSess.Session.InputID` → `result.Session.InputID`. |
| `pkg/evaluation/save_test.go` | Add a new test `TestSessionFromEventsPreservesInputID` that verifies `input_id` is carried through and `id` is still a fresh UUID. |

No changes needed to `SessionFromEvents`, `session.New`, or any other file.

---

### Step 1 — Add `InputID` field to `session.Session` in `session.go`

In the `Session` struct, add after the `ID` field:

```go
// ID is the unique identifier for the session
ID string `json:"id"`

// InputID is an optional caller-supplied correlation ID read from the eval
// input file's "input_id" field. It is carried through to the output as-is
// and never used internally. The session's own "id" is always a fresh UUID.
InputID string `json:"input_id,omitempty"`
```

`omitempty` ensures the field is absent from output JSON when empty — no change for existing eval files that don't include it.

**Verify:** `go build ./...` compiles clean.

---

### Step 2 — Copy `InputID` through in `eval.go`

In `runSingleEval`, after the existing line:

```go
result.Session = SessionFromEvents(events, title, userMessages)
```

add:

```go
result.Session.InputID = evalSess.Session.InputID
```

`evalSess.Session.InputID` is already populated by `json.Unmarshal` in `loadEvalSessions` (from Step 1). When the input file had no `"input_id"`, it is an empty string and `omitempty` keeps it out of the output — correct behaviour.

**Verify:** `go build ./pkg/evaluation/...` compiles clean.

---

### Step 3 — Add test `TestInputIDPassthrough` to `save_test.go`

Add a test that:
1. Creates an `InputSession` whose `Session.InputID` is set to a known value.
2. Runs `SessionFromEvents` and copies `InputID` as in Step 2.
3. Asserts `result.Session.InputID` equals the known value.
4. Asserts `result.Session.ID` is a non-empty UUID **different** from `InputID` (i.e. the random UUID was not disturbed).

```go
func TestInputIDPassthrough(t *testing.T) {
    t.Parallel()

    const knownInputID = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"

    sess := SessionFromEvents(nil, "title", []string{"q"})
    sess.InputID = knownInputID

    assert.Equal(t, knownInputID, sess.InputID)
    assert.NotEmpty(t, sess.ID)
    assert.NotEqual(t, knownInputID, sess.ID)
}
```

**Verify:** `go test ./pkg/evaluation/... -run TestInputIDPassthrough` passes.

---

### Step 4 — Run the full test suite

```bash
go test ./...
```

All pre-existing tests must continue to pass with no regressions.

---

## 6. Risks and Unknowns

| Risk | Likelihood | Notes |
|---|---|---|
| Two input eval files share the same `"input_id"` | Low | Fine — `input_id` is purely a passthrough label, not used as a key anywhere. No uniqueness constraint applies. |
| Adding `InputID` to `session.Session` affects unrelated session serialization | Low | `omitempty` means the field is invisible in all existing sessions that don't set it. |
| SQLite store serializes `session.Session` as JSON blob — `input_id` will be included automatically | Confirmed | No extra work needed; the store round-trips the full struct. |

---

## 7. Open Questions

None.

### DIFF
--- a/pkg/evaluation/save_test.go
+++ b/pkg/evaluation/save_test.go
@@ -680,3 +680,16 @@ func TestSessionFromEventsWithSessionTitle(t *testing.T) {
 	// Title should be updated from the event
 	assert.Equal(t, "Auto-generated title", sess.Title)
 }
+
+func TestInputIDPassthrough(t *testing.T) {
+	t.Parallel()
+
+	const knownInputID = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+
+	sess := SessionFromEvents(nil, "title", []string{"q"})
+	sess.InputID = knownInputID
+
+	assert.Equal(t, knownInputID, sess.InputID)
+	assert.NotEmpty(t, sess.ID)
+	assert.NotEqual(t, knownInputID, sess.ID)
+}
--- a/pkg/evaluation/eval.go
+++ b/pkg/evaluation/eval.go
@@ -366,6 +366,7 @@ func (r *Runner) runSingleEval(ctx context.Context, evalSess *InputSession) (Res
 
 	// Build session from events for database storage
 	result.Session = SessionFromEvents(events, title, userMessages)
+	result.Session.InputID = evalSess.Session.InputID
 	result.Session.Evals = evals
 
 	// Re-apply the display title in case a session_title event overrode it.
--- a/pkg/evaluation/eval.go
+++ b/pkg/evaluation/eval.go
@@ -366,7 +366,7 @@ func (r *Runner) runSingleEval(ctx context.Context, evalSess *InputSession) (Res
 
 	// Build session from events for database storage
 	result.Session = SessionFromEvents(events, title, userMessages)
-	result.Session.InputID = evalSess.Session.InputID
+	result.Session.InputID = evalSess.InputID
 	result.Session.Evals = evals
 
 	// Re-apply the display title in case a session_title event overrode it.
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -78,6 +78,11 @@ type Session struct {
 	// ID is the unique identifier for the session
 	ID string `json:"id"`
 
+	// InputID is an optional caller-supplied correlation ID read from the eval
+	// input file's "input_id" field. It is carried through to the output as-is
+	// and never used internally. The session's own "id" is always a fresh UUID.
+	InputID string `json:"input_id,omitempty"`
+
 	// Title is the title of the session, set by the runtime
 	Title string `json:"title"`
 


pkg/evaluation/save_test.go
```diff
@@ -680,3 +680,16 @@ func TestSessionFromEventsWithSessionTitle(t *testing.T) {
 	// Title should be updated from the event
 	assert.Equal(t, "Auto-generated title", sess.Title)
 }
+
+func TestInputIDPassthrough(t *testing.T) {
```
docker-agent, line 684: **[MEDIUM] TestInputIDPassthrough is tautological — it does not exercise the runSingleEval copy path**

The test manually assigns `sess.InputID = knownInputID` and then immediately asserts `sess.InputID == knownInputID`. That assertion is trivially true regardless of what `runSingleEval` does — removing the copy line `result.Session.InputID = evalSess.Session.InputID` in `eval.go` would not cause this test to fail.

The actual behaviour being fixed — that `InputID` flows from the deserialized `evalSess.Session` through `runSingleEval` into `result.Session` — is left uncovered. A regression in `eval.go` would be silent.

**Suggested fix:** test the end-to-end path. At minimum, construct an `InputSession` with `Session.InputID` pre-populated, call `runSingleEval` (or a lightweight helper that mirrors its copy logic), and assert the output `result.Session.InputID` equals the input value. Alternatively, add a round-trip integration test that writes a `.json` eval file containing `"input_id"`, runs the evaluator, and checks that the field appears in the result.

docker-agent (commented):

dgageot (approved):

PR #2858, opened 2026-05-21
feat(api): expose agent_model_overrides on GET /sessions/:id

trungutt:
## Summary

Surfaces the session's persisted per-agent model overrides on
`GET /api/sessions/:id` so HTTP clients can rehydrate a model picker
after a restart from a single round trip.

## Background

#2852 (released in v1.62.0) made `model` part of the
`POST /api/sessions/:id/agent/:agent` body and removed the dedicated
`PATCH /api/sessions/:id/model` endpoint. A non-empty `model` is
persisted on the session as a per-agent override
(`session.AgentModelOverrides[agentName]`), survives restarts, and is
reused on subsequent turns.

Today the only HTTP-accessible way for a client to read that
persisted state back is `GET /api/sessions/:id/models`, which
requires an active runtime (see
`SessionManager.AvailableSessionModels`). A session loaded purely
from the store after a restart has no attached runtime until the
user sends the next message, so a client UI that needs to show "what
model is this session using?" can't answer the question without
forcing a runtime attach.

## Wire-level change

- `SessionResponse` gains an `agent_model_overrides` field
  (`map[string]string`, `omitempty`), populated from the existing
  `session.AgentModelOverrides` field. The map mirrors the storage
  shape: `agent_name -> provider/model_ref`.
- Omitted entirely when no overrides are active so clients can
  distinguish "no override picked" from "override is the empty map"
  on the wire.
- Pure addition — no existing fields change shape and no endpoints
  are removed.

## Concrete client use case (Docker Desktop / Gordon)

Gordon's chat UI in Docker Desktop has a model picker per session
(behind a feature flag). With #2852 the user's selection is
persisted server-side, but after a Desktop restart the picker had
no way to know what was selected without speculatively starting the
session's runtime. This change lets the picker hydrate from
`getSession` like it already does for `permissions`, `working_dir`
and `tools_approved`.

## Validation

- `task build` ✅
- `go test ./pkg/server/... ./pkg/api/... ./pkg/session/... -count=1` ✅
- New tests in `pkg/server/server_test.go`:
  - `TestGetSession_ModelOverrides` — exposes overrides
  - `TestGetSession_NoOverrides` — `omitempty` is respected on the wire
- `task lint` blocked locally on a golangci-lint/Go version skew, but
  `go vet ./pkg/server/... ./pkg/api/...` and `gofmt -l` are clean.

### DIFF
--- a/docs/features/api-server/index.md
+++ b/docs/features/api-server/index.md
@@ -43,7 +43,7 @@ All endpoints are under the `/api` prefix.
 | -------- | ----------------------------------- | ------------------------------------------------------- |
 | `GET`    | `/api/sessions`                     | List all sessions                                       |
 | `POST`   | `/api/sessions`                     | Create a new session                                    |
-| `GET`    | `/api/sessions/:id`                 | Get a session by ID (messages, tokens, permissions)     |
+| `GET`    | `/api/sessions/:id`                 | Get a session by ID (messages, tokens, permissions, per-agent model overrides) |
 | `DELETE` | `/api/sessions/:id`                 | Delete a session                                        |
 | `PATCH`  | `/api/sessions/:id/title`           | Update session title                                    |
 | `PATCH`  | `/api/sessions/:id/permissions`     | Update session permissions                              |
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -135,6 +135,16 @@ type SessionResponse struct {
 	OutputTokens  int64                      `json:"output_tokens"`
 	WorkingDir    string                     `json:"working_dir,omitempty"`
 	Permissions   *session.PermissionsConfig `json:"permissions,omitempty"`
+	// AgentModelOverrides maps an agent name to the model ref currently
+	// applied as a per-agent override for that agent in this session
+	// (set on prior turns via the `model` field of POST /sessions/:id/agent/:agent).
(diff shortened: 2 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, additive change that correctly surfaces `agent_model_overrides` on the session response.

**What was reviewed:**
- `pkg/api/types.go`: New `AgentModelOverrides` field on `SessionResponse` with proper `omitempty` JSON tag
- `pkg/server/server.go`: Population of the field from `session.AgentModelOverrides` in the session handler
- `pkg/server/server_test.go`: Tests for both the populated and empty (omit) cases
- `docs/features/api-server/index.md`: API documentation updated to match

No bugs were found. The change is minimal, backward-compatible, and well-tested.

trungutt:
Closing — turns out we don't need this. `chat.Message.Model` is already persisted on every assistant message in the canonical `provider/model` form (via `modelsdev.ID.String()` in `pkg/runtime/harness.go`), so a client can rehydrate its model picker after a restart by walking the session's message history back to the most recent assistant turn. The pinata side has been updated to take that path, which sidesteps the agent-name lookup entirely and works against existing docker-agent releases (including v1.62.0) with no server change.

PR #2864, opened 2026-05-21
fix: allow configured HTTP proxy on private IPs in SSRF transport

dgageot:
Inside the docker-agent sandbox, all egress traffic must route through a network-policy proxy on a private IP (e.g. `http://172.17.0.0:3128`). The `pkg/httpclient.NewSSRFSafeTransport` function installs a dial-time SSRF check that rejects every non-public address. When the HTTP client follows the `HTTP_PROXY` environment variable, it dials the proxy itself first—and our check saw the private IP proxy address, rejected it, and every outbound request failed. This broke the aqua-registry HTTP fetch that auto-installs `gopls`, so the binary never landed on PATH and the MCP toolset couldn't start.

`NewSSRFSafeTransport` now snapshots `HTTP_PROXY`, `HTTPS_PROXY`, and `ALL_PROXY` (and lowercase variants) at construction time, parses each proxy spec into one or more `host:port` entries, and bypasses the SSRF dial control for dials whose post-resolution address matches. The underlying `SSRFDialControl` function is unchanged, so DNS-rebinding defenses continue to apply to every other dial. Allowlisting an operator-configured proxy doesn't widen the SSRF threat model: refusing to dial it provides zero protection (the proxy itself enforces destination policy), and breaking every request inside the sandbox is not acceptable.

Three new tests pin the behavior: `TestProxyHostPorts` covers URL/bare-host parsing and default port assignment for `http`, `https`, and `socks5` schemes; `TestNewSSRFSafeTransport_AllowsConfiguredProxy` is the regression test verifying private-IP proxies bypass SSRF rejection; `TestNewSSRFSafeTransport_AllowlistFrozenAtConstruction` documents that the allowlist is captured at construction time (which is correct since proxy env vars are set at process start).

### DIFF
--- a/pkg/httpclient/ssrf.go
+++ b/pkg/httpclient/ssrf.go
@@ -1,9 +1,12 @@
 package httpclient
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
+	"os"
 	"syscall"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


melmennaoui (dismissed):

pkg/httpclient/ssrf.go
```diff
@@ -53,16 +56,90 @@ func SSRFDialControl(_, address string, _ syscall.RawConn) error {
...
+		case "socks5", "socks5h":
+			port = "1080"
+		default:
+			port = "80"
+		}
+	}
+	out := map[string]struct{}{}
+	host := u.Hostname()
+	if ip := net.ParseIP(host); ip != nil {
+		out[net.JoinHostPort(ip.String(), port)] = struct{}{}
```
docker-agent, line 152: **[MEDIUM] IPv4-mapped IPv6 addresses may not match allowlist entries on dual-stack systems**

When a proxy is specified as an IPv4 literal (e.g. `http://10.0.0.1:3128`), `proxyHostPorts` stores the allowlist key as `"10.0.0.1:3128"` via `net.JoinHostPort(ip.String(), port)`.

On dual-stack Linux hosts where the OS dialer uses an IPv6 socket to connect to an IPv4 endpoint, the `Control` callback receives `"[::ffff:10.0.0.1]:3128"` as `address`. That string is not present in the allowlist, so the code falls through to `SSRFDialControl`, which rejects the private address — and the proxy dial fails.

**Trigger**: dual-stack host + IPv4-literal proxy address. The consequence is an availability failure (all proxied requests refused), not a security bypass.

**Suggested fix**: normalise both sides to the IPv4 canonical form before comparison, or add both the IPv4 and IPv4-mapped IPv6 representations to the allowlist when the input is an IPv4 literal:

```go
if ip4 := ip.To4(); ip4 != nil {
    out[net.JoinHostPort(ip4.String(), port)] = struct{}{}
    // also cover IPv4-mapped IPv6 dial path
    out[net.JoinHostPort(ip.To16().String(), port)] = struct{}{}
} else {
    out[net.JoinHostPort(ip.String(), port)] = struct{}{}
}
```

docker-agent (commented):

One medium-severity correctness finding verified in the new SSRF proxy allowlist code.

**Summary**: The PR correctly solves the sandbox proxy problem — the construction-time allowlist snapshot and post-resolution dial-control bypass are sound. DNS-rebinding attacks are not introduced (a rebind would cause the new IP to miss the allowlist and fall through to SSRFDialControl, blocking it). The security model is preserved for all non-proxy dials.

One edge case was found: IPv4-literal proxy addresses may not match on dual-stack systems where the OS connects over an IPv6 socket (receiving `[::ffff:10.0.0.1]:3128` in the Control callback instead of `10.0.0.1:3128`), causing proxy connections to be incorrectly rejected by SSRFDialControl.

pkg/httpclient/ssrf.go
dgageot: Addressed in 8ce99b3a5: added a `canonicalHostPort` helper that folds IPv4-mapped IPv6 dial addresses (`[::ffff:a.b.c.d]:port`) back to dotted-quad form before the allowlist lookup. So an IPv4-literal proxy on a dual-stack host will now match regardless of which socket family the OS dialer picks. Covered by `TestCanonicalHostPort`.

docker-agent: Perfect — canonicalHostPort is exactly the right approach. Folding IPv4-mapped addresses at lookup time is cleaner than pre-populating both forms in the allowlist, and TestCanonicalHostPort nails the test coverage (including the hex-form case). Thanks for the quick fix!

rumpl (approved):

PR #2866, opened 2026-05-21
fix(#2861): release per-message render caches when streaming completes

### LINKED ISSUE #2861: tui: per-message render caches leak across long sessions, triggering macOS jetsam kills
### Summary

On long-running TUI sessions, `docker-agent` retains roughly **161 KB of live heap per assistant message**, growing linearly without plateau. Multiple concurrent sessions exacerbate the problem. On macOS, sessions are eventually terminated by the kernel's `memorystatus` (jetsam) subsystem.

### Reproduction

Synthetic harness in `pkg/tui/components/messages/leak_test.go` streams 200 assistant messages of 8 KiB each (alternating prose and ```go code blocks) through `AddAssistantMessage` / `AppendToLastMessage`, calling `View()` between chunks.

| messages | HeapInuse (MB) | HeapAlloc (MB) | objects |
|---:|---:|---:|---:|
| 0   | 6.97  | 4.34  | 34,807 |
| 100 | 23.81 | 20.50 | 42,839 |
| 200 | 33.56 | 30.10 | 47,174 |
| 400 | 39.96 | 36.51 | 50,068 |

Slope: **+161 KB live / message**, perfectly linear.

### Symptom on macOS

```
kernel: [com.apple.xnu:memorystatus] memorystatus: killing largest compressed process docker-agent [<pid>] 390361 MB
```

Reported by users running multiple long sessions on a 48 GB Mac. The "390361 MB" figure is the kernel's compressed-pages footprint — ANSI output (escape sequences + line padding to terminal width) compresses extremely well, which explains the dramatic gap between live heap and jetsam's reported size.

### Root cause

Each `*messageModel` (`pkg/tui/components/message/message.go`) retains, for the lifetime of the session:

1. `renderCache.result` — full styled ANSI output of `View()`, populated on first render and only invalidated when an input changes (never released after the stream completes).
2. `mdRenderer *markdown.IncrementalRenderer` — holds `inputPrefix` (raw markdown), `outputPrefix` (rendered ANSI), and `codeBlocksPrefix` for the entire message.

400 `messageModel` instances × ~80–120 KB of retained strings each matches the observed slope. `m.renderedItems` LRU(500) and global syntax-highlight caches are bounded and not contributors.

### Regression window

The retention paths were introduced by these perf commits:

- `54c4020d6` — `perf(tui/message): cache Render output to avoid redundant markdown parses` (May 12)
- `eddc8fb20` — `perf(tui/markdown): incremental renderer for streaming chunks` (May 12)

Both correctly invalidate on input change but never *release* the cached strings once a message is no longer the active streaming target.

### Proposed fix

Add a `Finalize()` method on `*messageModel` that drops `renderCache` and resets+nils `mdRenderer`, called on the previous assistant view when a new message is appended (and on all loaded views in `LoadFromSession`). Re-renders triggered later by scroll/hover lazily re-populate, paying one full markdown parse per re-visit instead of holding the result forever. The `renderedItems` LRU absorbs most of the re-render cost.

### Repro environment

- macOS, Apple silicon, 48 GB RAM
- `docker-agent` HEAD as of May 20, 2026
- Multiple concurrent TUI sessions over many hours

aheritier:
## Summary

`messageModel` held its rendered ANSI output (`renderCache.result`) and a per-message `IncrementalRenderer` (`inputPrefix` + `outputPrefix` + `codeBlocksPrefix`) for the full lifetime of the session. With ~80–120 KB of retained strings per assistant turn and no eviction, a long TUI session grew its heap linearly until macOS's `memorystatus` (jetsam) subsystem killed `docker-agent` as the largest compressed process.

Reported in #2861 via:

```
kernel: [com.apple.xnu:memorystatus] memorystatus: killing largest compressed process docker-agent [<pid>] 390361 MB
```

The compressed-pages number is dramatic because ANSI output (escape sequences + line padding to terminal width) compresses extremely well, but the underlying retained heap is real and grows at +161 KB / assistant message in a synthetic harness.

## Fix

- New `Finalize()` and `HasLiveRenderState()` on the `message.Model` interface and `*messageModel`. `Finalize()` drops `renderCache` and `Reset()`s + nils `mdRenderer`. Sets a `finalized` flag so `Render()` skips repopulating `renderCache`, and `renderAssistantMarkdown()` falls back to a transient renderer rather than persisting one on the struct.
- `Finalize()` is gated on `MessageTypeAssistant`. Only assistant views allocate an `IncrementalRenderer` and accumulate large rendered ANSI strings during streaming. User messages, tool calls and other types have nothing to release; setting `finalized = true` on them would only have the side effect of permanently disabling their `renderCache` for selected/hovered states.
- `SetMessage()` un-finalizes the view if streaming ever resumes into it (Finalize is purely a cache hint).
- `messages.model.finalizePreviousMessageView()` runs whenever a new top-level entry is appended (`addMessage`, `AddCancelledMessage`, `AddOrUpdateToolCall`).
- `addReasoningBlock()` deliberately does **not** finalize. Reasoning blocks interleave with an actively streaming assistant turn, and finalizing mid-stream would force every subsequent chunk through the slow transient-renderer path. The next non-reasoning entry finalizes via `addMessage()` when the streaming turn actually ends.
- `LoadFromSession` finalizes every loaded view except the last.
- `removeSpinner()` no longer calls `invalidateAllItems()`. The spinner is always at the tail, so wiping the entire 500-entry rendered-item LRU on every assistant turn was both wasteful and immediately re-armed the leak by forcing a full re-render of every previous message into a fresh `renderCache`. It now only resets the joined `renderedLines` slab; spinner-driven messages are never put in the LRU (`shouldCacheMessage` returns `false`), so no LRU `Delete` is needed.

The chat-list's bounded LRU continues to memoize finalized messages' rendered output, so subsequent re-renders are still O(1) when a viewed message is in cache.

## Validation

`pkg/tui/components/messages/leak_test.go` streams 400 8-KiB assistant messages (200 warmup + 200 measured) and asserts both a per-message heap budget and a structural invariant. The structural assertion uses the new exported `HasLiveRenderState()` method rather than cross-package reflection on unexported fields.

| | per-message slope | live render-state holders |
|---|---:|---:|
| HEAD before fix | **161 KB / msg** (linear, no plateau) | 400 / 400 |
| HEAD after fix | **18 KB / msg** | 1 / 800 |

```
$ task build && task lint
Built ./bin/docker-agent
0 issues.
$ go test ./pkg/tui/components/message ./pkg/tui/components/messages -count=1
ok  	github.com/docker/docker-agent/pkg/tui/components/message
ok  	github.com/docker/docker-agent/pkg/tui/components/messages
```

`task test` passes everywhere except the two `pkg/teamloader/TestLoadExamples/.../dmr.yaml` cases that require Docker Model Runner to be running locally — environmental, unrelated to this change.

Fixes #2861

---

Assisted-By: docker-agent

### DIFF
--- a/pkg/tui/components/message/message.go
+++ b/pkg/tui/components/message/message.go
@@ -27,6 +27,16 @@ type Model interface {
 	SetSelected(selected bool)
 	SetHovered(hovered bool)
 	CodeBlocks() []markdown.CodeBlock
+	// Finalize releases per-message render state that is only needed while the
+	// message is actively streaming. The message content and code-block metadata
+	// are preserved; calling View() afterwards still produces correct output
+	// without retaining a per-view render cache or IncrementalRenderer.
+	Finalize()
+	// HasLiveRenderState reports whether this view currently retains a
+	// populated renderCache or an IncrementalRenderer instance. Used by tests
+	// to assert that finalized views have actually released their per-message
+	// render state without reaching into unexported fields via reflection.
+	HasLiveRenderState() bool
 }
 
 // messageModel implements Model
@@ -59,6 +69,14 @@ type messageModel struct {
 	// streamed-in chunks only re-render the trailing block instead of the whole
 	// accumulated markdown each time.
 	mdRenderer *markdown.IncrementalRenderer
+
+	// finalized is set by Finalize() once the message is no longer the active
+	// streaming view. After it is set, Render() still produces correct output,
+	// but does not store anything in renderCache and does not retain an
+	// IncrementalRenderer between calls — both are pure caches whose memory
+	// dominates a long session, and they are not worth keeping for messages
+	// that are unlikely to be re-rendered hot.
+	finalized bool
 }
 
 // renderCache stores the most recent Render result keyed by the inputs that
@@ -99,6 +117,11 @@ func (mv *messageModel) Init() tea.Cmd {
 }
 
 func (mv *messageModel) SetMessage(msg *types.Message) {
+	// Un-finalize when the underlying message is changed (e.g. streaming
+	// resumes into this view). Finalize is meant for views that have
+	// permanently lost their actively-streaming status; mutating the message
+	// re-arms the per-message caches so subsequent renders are fast again.
+	mv.finalized = false
 	// If the new content is not an extension of the previous one (different
 	// message, or the message was edited), drop the IncrementalRenderer's
 	// cached prefix so its memory is released immediately rather than on the
@@ -177,7 +200,11 @@ func (mv *messageModel) Render(width int) string {
 	// MessageTypeAssistant placeholder) animate on every tick, so the result is
 	// not cacheable. Everything else is a pure function of the inputs tracked in
 	// renderCache below.
-	cacheable := !mv.isSpinnerDriven()
+	// Spinner-driven messages animate every tick and are not cacheable.
+	// Finalized messages skip writing into renderCache so the per-view
+	// retained ANSI string does not pile up across long sessions; the chat
+	// list's bounded LRU still memoizes their rendered output.
+	cacheable := !mv.isSpinnerDriven() && !mv.finalized
 	if cacheable {
 		c := &mv.renderCache
 		if c.valid &&
@@ -376,10 +403,22 @@ func (mv *messageModel) render(width int) string {
 // so each new chunk only re-parses the trailing region. The first render at a
 // given width is equivalent to a fresh full render.
 //
+// For finalized messages we use a transient renderer that is discarded after
+// each call. Finalized messages are no longer streamed, so the prefix-cache
+// inside an IncrementalRenderer is not earning its keep — keeping it resident
+// across the lifetime of every historical message in a session is the
+// dominant source of retained memory in long sessions. The parent message
+// list's bounded rendered-item LRU can still memoize finalized message output
+// without storing an additional per-view copy.
+//
 // It also returns the list of fenced code blocks emitted by the renderer so
 // that callers can map clicks on the per-block copy affordance back to the
 // underlying raw code.
 func (mv *messageModel) renderAssistantMarkdown(content string, width int) (string, []markdown.CodeBlock, error) {
+	if mv.finalized {
+		r := markdown.NewIncrementalRenderer(width)
+		return r.RenderWithCodeBlocks(content)
+	}
 	if mv.mdRenderer == nil {
 		mv.mdRenderer = markdown.NewIncrementalRenderer(width)
 	} else {
@@ -440,6 +479,54 @@ func (mv *messageModel) StopAnimation() {
 	if mv.message.Type == types.MessageTypeSpinner || mv.message.Type == types.MessageTypeLoading {
 		mv.spinner.Stop()
 	}
+}
+
+// Finalize releases per-message render state that no longer needs to be kept
+// resident once the message is no longer the actively streaming view. This is
+// called by the parent message list when a new top-level message arrives, and
+// for every historical view loaded from a session.
+//
+// Finalize is a no-op for non-assistant message types: only assistant views
+// allocate an IncrementalRenderer and accumulate large rendered ANSI strings
+// during streaming, so user messages, tool calls, error/welcome banners and
+// the like have nothing to release. Setting `finalized = true` on those views
+// would only have the side-effect of permanently disabling renderCache for
+// selected/hovered states (which bypass the parent's bounded LRU), forcing a
+// fresh re-render on every animation tick. Restricting the disable to
+// assistant views keeps the leak fix scoped to the type that actually leaks.
+//
+// The retained payload of an assistant view is dominated by the renderCache
+// (a copy of the rendered ANSI string) and the IncrementalRenderer's internal
+// caches (last rendered prefix, glamour AST state). Both are pure render
+// state — they can be regenerated from mv.message on demand. We deliberately
+// leave mv.message, mv.codeBlocks and the spinner untouched so that View()
+// keeps returning correct output, click-targeting on code blocks still works,
+// and the spinner-driven types continue to animate.
+//
+// Finalize is idempotent and durable: subsequent renders do not re-populate
+// renderCache or store an IncrementalRenderer on the struct. This is
+// important because the parent message list invalidates its own LRU on
+// several events (spinner removal, theme change, window resize) and would
+// otherwise re-render every previously finalized view, putting the per-
+// message render state right back where it was.
+func (mv *messageModel) Finalize() {
+	if mv.message == nil || mv.message.Type != types.MessageTypeAssistant {
+		return
+	}
+	mv.renderCache = renderCache{}
+	if mv.mdRenderer != nil {
+		mv.mdRenderer.Reset()
+		mv.mdRenderer = nil
+	}
+	mv.finalized = true
+}
+
+// HasLiveRenderState reports whether this view still retains per-message
+// render state — either a populated renderCache or an IncrementalRenderer
+// instance. Used as a structural assertion in regression tests that verify
+// Finalize() actually released what it was supposed to release.
+func (mv *messageModel) HasLiveRenderState() bool {
+	return mv.renderCache.result != "" || mv.mdRenderer != nil
 }
 
 // SetSize sets the dimensions of the message view
--- a/pkg/tui/components/messages/leak_test.go
+++ b/pkg/tui/components/messages/leak_test.go
@@ -1,0 +1,169 @@
+package messages
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/docker/docker-agent/pkg/tui/components/message"
+	"github.com/docker/docker-agent/pkg/tui/core/layout"
+	"github.com/docker/docker-agent/pkg/tui/service"
+)
+
+// TestLongSessionDoesNotRetainPerMessageRenderState streams a long sequence
+// of large assistant messages through the same public API the TUI uses
+// (AddUserMessage / AddAssistantMessage / AppendToLastMessage), rendering the
+// active assistant view between chunks so messageModel.renderCache and
+// IncrementalRenderer state are exercised. It asserts two things:
+//
+//  1. Heap growth stays far below the pre-fix slope. Before the fix every
+//     assistant view kept its own renderCache (~70 KB rendered ANSI) and an
+//     IncrementalRenderer (cached prefix + rendered output) for a measured
+//     ~161 KB of retained state per message. After the fix, per-message render
+//     state is released as soon as the message stops being the active
+//     streaming view, so the remaining slope is mostly the raw session content.
+//
+//  2. The structural invariant: after every non-active assistant view has
+//     been finalized, only the single actively-streaming view (at most one)
+//     should be holding a non-empty renderCache or a live IncrementalRenderer.
+//
+// Run with -short to skip; the test is deterministic but takes a few seconds.
+func TestLongSessionDoesNotRetainPerMessageRenderState(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping leak test in -short mode")
+	}
+
+	// `go test` may reset runtime.MemProfileRate when its own -memprofile flag
+	// is unset, which would override the value set in init().
+	runtime.MemProfileRate = 1
+
+	const (
+		warmupMessages  = 200
+		measureMessages = 200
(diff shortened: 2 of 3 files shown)


docker-agent (commented):

This PR correctly addresses the memory leak in long TUI sessions by introducing `Finalize()` on `messageModel` to drop `renderCache` and `mdRenderer` once streaming completes. The `removeSpinner()` targeted-invalidation improvement is architecturally sound. Two issues were found in the new code — one performance edge case and one dead code placement.

pkg/tui/components/messages/messages.go
```diff
@@ -1550,6 +1579,8 @@ func (m *model) addReasoningBlock(agentName, content string) tea.Cmd {
 	m.clearSelection()
 	shouldAutoScroll := !m.userHasScrolled
 
+	m.finalizePreviousMessageView()
```
docker-agent: **[MEDIUM] Premature finalization of streaming assistant message when reasoning block is added mid-stream**

`addReasoningBlock()` unconditionally calls `finalizePreviousMessageView()` before appending the new reasoning block. In a typical LLM streaming session the sequence is:

1. LLM streams assistant content → view is actively rendered via its cached `mdRenderer`
2. LLM emits a reasoning block → `AppendReasoning` → `addReasoningBlock` → **`Finalize()` is called on the still-streaming assistant view**
3. `Finalize()` drops `renderCache` and nils `mdRenderer`, sets `finalized = true`
4. If the same assistant view later receives `SetMessage` (streaming resumes), `finalized` is reset to `false` — but until then, any `View()` call triggers the slow transient-renderer path (new `IncrementalRenderer` per render)

The recovery path exists (`SetMessage` re-arms the view), so there is no data loss or incorrect state. However, the `finalized = true` window between step 3 and step 4 can cause unexpectedly expensive renders for a view the user is actively watching.

**Suggestion**: Consider only finalizing if the previous view is a different kind of message (e.g., `types.MessageTypeAssistantReasoningBlock`), or document explicitly that premature finalization of a streaming assistant view is intentional and the cost is acceptable.

pkg/tui/components/messages/messages.go
```diff
@@ -1238,6 +1254,7 @@ func (m *model) AddWelcomeMessage(content string) tea.Cmd {
 	if content == "" || len(m.views) > 0 {
 		return nil
 	}
+	m.finalizePreviousMessageView()
```
docker-agent: **[LOW] `finalizePreviousMessageView()` is unreachable dead code in `AddWelcomeMessage`**

The early-return guard `if content == "" || len(m.views) > 0 { return nil }` ensures the function only proceeds when `m.views` is empty. This means the `finalizePreviousMessageView()` call immediately after is always a no-op — `finalizePreviousMessageView()` itself has its own guard `if len(m.views) == 0 { return }` that will immediately return.

The intent was likely to finalize any existing message before adding the welcome message, but the outer guard prevents that scenario from ever being reached. In practice this is harmless (a welcome message is only ever added when views is empty), but the call is misleading to future readers.

**Suggestion**: Remove the `finalizePreviousMessageView()` call from `AddWelcomeMessage` since it can never have any effect, keeping the code consistent with its actual invariant.

pkg/tui/components/messages/messages.go
aheritier: Fixed in 306af6ab2 by removing the `finalizePreviousMessageView()` call from `addReasoningBlock()` entirely. Reasoning blocks routinely interleave with an actively streaming assistant turn, so finalizing here would force every subsequent chunk through the slow transient-renderer path. The next non-reasoning entry (next user message, tool call, error) finalizes via `addMessage()` when the streaming turn actually ends. Added a code comment explaining the invariant.

pkg/tui/components/messages/messages.go
aheritier: Fixed in 306af6ab2 by removing the unreachable call. The outer `len(m.views) > 0` guard makes it dead code by construction.

docker-agent: Thanks for the fix! The removal cleans up the misleading code path nicely — now the function clearly reflects its invariant that it only operates on empty view lists.

pkg/tui/components/messages/messages.go
```diff
@@ -1670,7 +1706,19 @@ func (m *model) removeSpinner() {
...
 		m.messages = m.messages[:lastIdx]
-		m.invalidateAllItems()
+		// The spinner is always at the tail, so other indices are unchanged
+		// and their cached entries remain valid. Drop just the spinner's own
+		// LRU entry and the joined renderedLines (which still references the
+		// removed spinner). Avoiding invalidateAllItems here is essential for
+		// long sessions: it would otherwise wipe up to 500 cached renderings
+		// once per assistant turn, forcing every previous message to be
+		// re-parsed from markdown on the next render.
+		m.renderedItems.Delete(lastIdx)
```
docker-agent: **[LOW] `removeSpinner` LRU delete is always a no-op**

`m.renderedItems.Delete(lastIdx)` operates on a key that was never stored — `shouldCacheMessage` returns `false` for `MessageTypeSpinner` (via the `default` branch), so `renderItem` never calls `m.renderedItems.Put` for spinner entries. The delete is safe (no panic) and the surrounding cache invalidation (`m.renderedLines = nil`, etc.) still works correctly. But the accompanying comment — *"Drop just the spinner's own LRU entry"* — is misleading and may confuse future maintainers who assume the spinner had a cached entry to remove.

Consider removing the `Delete(lastIdx)` call and updating the comment to reflect that the spinner was never in the LRU, only in `renderedLines`.

pkg/tui/components/message/message.go
```diff
@@ -177,7 +195,11 @@ func (mv *messageModel) Render(width int) string {
 	// MessageTypeAssistant placeholder) animate on every tick, so the result is
 	// not cacheable. Everything else is a pure function of the inputs tracked in
 	// renderCache below.
-	cacheable := !mv.isSpinnerDriven()
+	// Spinner-driven messages animate every tick and are not cacheable.
+	// Finalized messages skip writing into renderCache so the per-view
+	// retained ANSI string does not pile up across long sessions; the chat
+	// list's bounded LRU still memoizes their rendered output.
+	cacheable := !mv.isSpinnerDriven() && !mv.finalized
```
docker-agent, line 207: **[LOW] `Finalize()` permanently disables `renderCache` on non-assistant messages (user, tool calls) when they are selected or hovered**

`finalizePreviousMessageView()` is called before every new message append (`addMessage`, `AddCancelledMessage`, `AddOrUpdateToolCall`). The last view at that point is often a **user message or tool call**, not an assistant message. `Finalize()` sets `finalized = true` on these views, which makes `cacheable = false` in `Render()`.

For the common case (non-selected, non-hovered), the outer LRU in `m.renderedItems` still caches the rendered string, so no extra work is done. However, for a **selected or hovered** finalized message, `renderItem` sets `shouldCache = !isSelected && !isHovered = false`, bypassing the LRU entirely and calling `view.View()` → `mv.Render()` every frame. With `mv.finalized = true`, `mv.renderCache` is also disabled — so the message is re-rendered from scratch on every animation tick or hover-tracking update.

User message rendering is cheap (plain text + lipgloss styling), but the original PR motivation was specifically to shed per-*assistant* message renderer state. Propagating `finalized` to user messages is an unintended side-effect.

A targeted fix: check `mv.message.Type == types.MessageTypeAssistant` inside `Finalize()` before setting `mv.finalized = true`, or restrict `finalizePreviousMessageView()` to only finalize assistant-typed views. The `mdRenderer` and `renderCache.result` growth that the PR addresses is exclusive to assistant messages.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The memory-leak fix is well-reasoned and correctly targets the dominant sources of retained heap in long TUI sessions: the per-message `renderCache.result` ANSI string and the `IncrementalRenderer`'s prefix cache. The `Finalize()`/`finalizePreviousMessageView()` lifecycle is sound, the `SetMessage()` un-finalize path is correct, and the regression test methodology (heap-inuse slope + structural invariant via reflection) is solid.

Two minor issues found in the added code (see inline comments).

pkg/tui/components/messages/leak_test.go
```diff
@@ -0,0 +1,188 @@
...
+		cacheField := elem.FieldByName("renderCache")
+		rendererField := elem.FieldByName("mdRenderer")
+		if !cacheField.IsValid() || !rendererField.IsValid() {
+			continue
+		}
+		hasCache := false
+		if result := cacheField.FieldByName("result"); result.IsValid() && result.Kind() == reflect.String {
+			hasCache = result.Len() > 0
+		}
+		hasRenderer := !rendererField.IsNil()
```
docker-agent: **[HIGH] Reflection on unexported fields from a different package will panic at test runtime**

`countLiveRenderState` accesses `renderCache` and `mdRenderer` — both unexported fields of `messageModel` defined in `package message` — via reflection, from `package messages` (a different package).

In Go, `reflect.Value.FieldByName("unexportedField")` on a struct from a *different* package returns a valid `reflect.Value`, so the `!cacheField.IsValid()` guard does **not** protect you. However, calling `.IsNil()` (line 155) or `.Len()` on that value panics at runtime with:

```
reflect: reflect.Value.IsNil using value obtained using unexported field
```

The test will therefore panic rather than pass or fail gracefully whenever a `message.Model` view with a live `mdRenderer` is encountered. To fix, either:
1. Move `leak_test.go` into `package message` (rename to `package message_test` external, or `package message` internal), or
2. Add an exported helper/method to `message.Model` (e.g., `HasLiveRenderState() bool`) that the test calls instead of using reflection across package boundaries.

The comment "colocated with the package under test" is misleading — the file is in `package messages`, not `package message`.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

pkg/tui/components/messages/messages.go
aheritier: Fixed in 243c31f64. Removed the `Delete(lastIdx)` no-op and updated the comment to reflect that the spinner was never put in the LRU in the first place (`shouldCacheMessage` returns false for spinner-driven types).

pkg/tui/components/message/message.go
aheritier: Fixed in 243c31f64. `Finalize()` now early-returns for any non-`MessageTypeAssistant` view. Only assistant views allocate an `IncrementalRenderer` and accumulate large `renderCache` strings during streaming, so user messages and tool calls have nothing to release — and crucially, `finalized = true` no longer leaks onto types whose `renderCache` would then be permanently disabled for selected/hovered states. Doc comment on `Finalize()` updated to spell this out.

pkg/tui/components/messages/leak_test.go
aheritier: Fixed in 243c31f64. Added an exported `HasLiveRenderState() bool` method on the `message.Model` interface, and rewrote `countLiveRenderState` in the test to call it instead of reaching into unexported fields via reflection. The `reflect` import is gone, and the test no longer depends on the cross-package access semantics.

docker-agent: Perfect, thank you! Adding the exported method was the right call — the test now has a clean, package-boundary-safe API to query the state without relying on reflection internals. The implementation in `messageModel.HasLiveRenderState()` is straightforward and covers both the `renderCache` and `mdRenderer` fields as needed.

This is exactly the kind of fix that makes the test both more robust and easier to maintain going forward.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR introduces a well-designed memory leak fix for the TUI's `messageModel`. The `Finalize()` / `HasLiveRenderState()` / `finalizePreviousMessageView()` mechanics are coherent, the gating on `MessageTypeAssistant` is correct, and the reasoning about why `addReasoningBlock` deliberately skips finalization is sound.

**What was reviewed:**
- `pkg/tui/components/message/message.go` — new `Finalize()`, `HasLiveRenderState()`, `renderAssistantMarkdown` transient-renderer path, `SetMessage` un-finalize logic, `shouldCacheMessage` guard
- `pkg/tui/components/messages/messages.go` — `finalizePreviousMessageView()` call sites, `removeSpinner` narrowing from `invalidateAllItems` to targeted reset, `LoadFromSession` finalization loop
- `pkg/tui/components/messages/leak_test.go` — heap-slope and live-render-state assertions

No confirmed or likely bugs were found in the added code. The implementation correctly addresses the linear heap growth described in #2861.

dgageot (approved):

PR #2869, opened 2026-05-21
Show the path from where the skill is loaded

rumpl:
Screenshot 2026-05-22 at 01 53 44

### DIFF
--- a/pkg/path/display.go
+++ b/pkg/path/display.go
@@ -1,0 +1,67 @@
+package path
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// RelativeTo returns p relative to baseDir when both paths are absolute and
+// filepath.Rel can compute a relative path. Otherwise it returns p cleaned.
+func RelativeTo(p, baseDir string) string {
+	if p == "" {
+		return p
+	}
+
+	p = filepath.Clean(p)
+	if baseDir == "" || !filepath.IsAbs(p) || !filepath.IsAbs(baseDir) {
+		return p
+	}
+
+	rel, err := filepath.Rel(filepath.Clean(baseDir), p)
+	if err != nil {
+		return p
+	}
+	return rel
+}
+
+// ShortenHome replaces the current user's home directory prefix with "~".
+func ShortenHome(p string) string {
+	homeDir, err := os.UserHomeDir()
+	if err != nil || homeDir == "" {
+		return p
+	}
+	return ShortenHomeDir(p, homeDir)
+}
+
+// ShortenHomeDir replaces a leading homeDir prefix with "~".
+func ShortenHomeDir(p, homeDir string) string {
+	if p == "" || homeDir == "" {
+		return p
+	}
+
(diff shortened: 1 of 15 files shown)


pkg/path/display.go
```diff
@@ -0,0 +1,67 @@
...
+	rel, err := filepath.Rel(filepath.Clean(baseDir), p)
+	if err != nil {
+		return p
+	}
+	return rel
+}
+
+// ShortenHome replaces the current user's home directory prefix with "~".
+func ShortenHome(p string) string {
+	homeDir, err := os.UserHomeDir()
```
docker-agent, line 30: **[LOW] ShortenHome calls os.UserHomeDir() on every invocation**

`ShortenHome` calls `os.UserHomeDir()` directly on each call, which involves an environment variable lookup or syscall each time. The previous `ShortenPath` helper used the cached result from `paths.GetHomeDir()` (an atomic pointer). Since this function is called from multiple TUI rendering paths (directorytree, editfile, listdirectory, readfile, readmultiplefiles, searchfilescontent) which may be invoked on every render frame, this is a minor performance regression. Consider caching the home dir at package init or accepting it as a parameter (as `ShortenHomeDir` already does).

pkg/tui/dialog/skills.go
```diff
@@ -66,10 +68,40 @@ func formatSkill(s *skills.Skill, contentWidth int) []string {
...
+	if path == "" {
+		return ""
+	}
+
+	if s.Local {
+		if skills.IsHomeSkillPath(path) {
+			return pathx.ShortenHome(path)
+		}
+		cwd, _ := os.Getwd()
+		return pathx.RelativeTo(path, cwd)
```
docker-agent, line 99: **[LOW] skillLoadedFrom: RelativeTo may return a verbatim non-absolute path**

When `s.Local == true`, `skills.IsHomeSkillPath(path)` is false (project-local skill), and `path` is not absolute, `pathx.RelativeTo(path, cwd)` returns the path cleaned but verbatim (since `RelativeTo` only relativizes when both paths are absolute). In production, `loadSkillFile` always produces an absolute `BaseDir` via `filepath.Dir(path)`, so this edge case shouldn't arise. However, there's no guard/assertion enforcing this invariant. A defensive check (`if !filepath.IsAbs(path)`) could make the assumption explicit and prevent a confusing display string if the invariant is ever violated.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR introduces a clean `pkg/path` package (`RelativeTo`, `ShortenHome`, `ShortenHomeDir`, `IsWithin`) to replace ad-hoc path-shortening logic, and plumbs the skill load path through the skills dialog. The logic is correct: `IsWithin` properly guards against prefix-spoofing (e.g. `/home/user2` wrongly matching `/home/user`), and `ShortenHomeDir` handles edge cases like the exact home-directory match. No high or medium severity issues found. Two minor observations are left as inline comments.

dgageot (approved):

PR #2878, opened 2026-05-22
Polish --sandbox auto-kit output and tool auto-install logging

dgageot:
The `--sandbox` auto-kit output now reads better and carries more context when tools auto-install at runtime. These changes improve both the user-visible summary and the underlying tool-staging logic.

**Auto-kit summary readability.** Subtle ANSI styling is now applied to the kit summary printed by `--sandbox` (kit path in cyan, section headers and names in bold, faint notes for `(from …)` / `workspace mount`, yellow for `(redacted)` markers and the secret count). The styling gracefully degrades when stdout isn't a TTY — tests and piped output stay plain thanks to fatih/color's auto-detection.

**Path display.** Every host path in the kit summary now goes through `pathx.ShortenHome`, including the kit directory itself, so `$HOME` is consistently collapsed to `~` for readability.

**Skill staging respects the agent config.** The `stageSkills` logic now consults each agent's `SkillsConfig` instead of unconditionally shipping every local skill found. It ships nothing if no agent enables local skills, every local skill if any agent enables them without an `include` filter (the wide case wins), and otherwise only the union of agents' `include` filters. Five new tests cover these scenarios.

**Auto-install logging mentions the package.** When an MCP/LSP server auto-installs at runtime, both the debug log and a new user-visible stderr line now name the package as `command (owner/repo@version)` before the download starts, providing context for the `go install` chatter that follows. The stderr banner respects `isatty` on stderr directly and honours `NO_COLOR` / `TERM=dumb`.

All changes maintain backward compatibility. Linting and all 132 package tests pass.

### DIFF
--- a/pkg/sandbox/kit/kit.go
+++ b/pkg/sandbox/kit/kit.go
@@ -242,7 +242,7 @@ func Build(ctx context.Context, opts Options) (*Result, error) {
 		cfg = &latestcfg.Config{}
 	}
 
-	skillsEntries, redactions, err := stageSkills(stagingDir)
+	skillsEntries, redactions, err := stageSkills(stagingDir, cfg)
 	if err != nil {
 		return nil, err
 	}
@@ -371,12 +371,24 @@ func loadConfig(ctx context.Context, opts Options) (*latestcfg.Config, error) {
 	return config.Load(ctx, source)
 }
 
-// stageSkills copies every local skill discovered on the host into
+// stageSkills copies every local skill the agent config enables into
 // <kit>/skills/<skill-name>/, redacting text files in place.
-func stageSkills(kitDir string) ([]Entry, []Redaction, error) {
+//
+// Only skills the agent will actually load are staged: if no agent in
+// cfg enables local skills the kit ships nothing, and if every agent
+// that enables local skills also restricts them via an `include`
+// filter, only the union of those filters is staged. An agent that
+// enables local skills without a filter is the wide case — every local
+// skill is staged.
+func stageSkills(kitDir string, cfg *latestcfg.Config) ([]Entry, []Redaction, error) {
 	target := filepath.Join(kitDir, skills.KitSkillsSubdir)
 	if err := os.MkdirAll(target, 0o750); err != nil {
 		return nil, nil, fmt.Errorf("creating kit skills dir: %w", err)
+	}
+
+	include, ok := localSkillFilter(cfg)
+	if !ok {
+		return nil, nil, nil
 	}
 
 	var (
@@ -385,6 +397,9 @@ func stageSkills(kitDir string) ([]Entry, []Redaction, error) {
 	)
 	for _, skill := range skills.Load([]string{"local"}) {
 		if skill.BaseDir == "" {
+			continue
+		}
+		if include != nil && !include[skill.Name] {
 			continue
 		}
 		dst := filepath.Join(target, sanitise(skill.Name))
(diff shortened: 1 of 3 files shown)


pkg/toolinstall/resolver.go
```diff
@@ -130,6 +132,40 @@ func doInstall(ctx context.Context, command, versionRef string) (string, error)
...
+// intentionally avoid stdout so this never gets piped into the
+// agent's prompt or programmatic output.
+//
+// fatih/color's global NoColor is computed from stdout, so we cannot
+// rely on it to suppress styling when only stderr is redirected
+// (e.g. `agent run ... 2>log` keeps stdout on the TTY). Decide based
+// on stderr explicitly and honour NO_COLOR / TERM=dumb.
+func announceInstall(command, pkgName, version string) {
+	if stderrSupportsColor() {
+		bold := color.New(color.Bold).SprintFunc()
```
docker-agent: **[LOW] Cosmetic: `announceInstall` enters the styled branch but `fatih/color` may suppress ANSI codes**

`stderrSupportsColor()` correctly gates on stderr's TTY state, so the `if` branch is entered whenever stderr is a terminal. However, `color.New(color.Bold).SprintFunc()` and `color.New(color.Faint).SprintFunc()` internally consult `color.NoColor`, which `fatih/color` sets at init time based on **stdout** (not stderr). In the common case where stdout is piped (e.g. `agent run ... | tee log`) but stderr is still a TTY, `color.NoColor == true` and both `SprintFunc()` calls become plain-text pass-throughs — so the colored branch is taken yet the output has no styling.

The comment above the function (lines 141–144) acknowledges the stdout/stderr mismatch but describes only the inverse scenario (`2>log`). The current scenario is the more common one in practice.

**Impact**: cosmetic only — the install message still appears on stderr, just without bold/faint formatting. No data is lost.

**Suggested fix**: force styling by calling `.EnableColor()` on the local `*color.Color` instances, or use `color.New(color.Bold).FprintfFunc()` and pass `os.Stderr` so `fatih/color` checks the right fd:

```go
bold := color.New(color.Bold)
bold.EnableColor()
faint := color.New(color.Faint)
faint.EnableColor()
fmt.Fprintf(os.Stderr, "Installing %s %s\n",
    bold.Sprint(command), faint.Sprintf("(%s@%s)", pkgName, version))
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are well-structured and the logic is sound. The `stageSkills`/`localSkillFilter` union logic is correct. ANSI styling in `PrintSummary` correctly delegates TTY detection to `fatih/color`'s stdout-based `NoColor`. One cosmetic nit on `announceInstall` (see inline comment).

pkg/toolinstall/resolver.go
```diff
@@ -10,6 +10,8 @@ import (
 	"runtime/debug"
 	"strings"
 
+	"github.com/fatih/color"
+	"github.com/mattn/go-isatty"
 	"golang.org/x/sync/singleflight"
```
aheritier, line 15: **Nit — minor diagnostic regression.** The pre-lookup `slog.InfoContext` was removed, so if `lookupPackage` fails (unknown command, network error) no structured log records the *intent* to install. The error still carries `command` via `EnsureCommand`'s wrapping, but a debug-level breadcrumb before the lookup makes log-tailing easier:

```go
slog.DebugContext(ctx, "looking up package for auto-install", "command", command)
```

Non-blocking — acceptable trade-off given the user-visible `announceInstall` line covers the interactive case.

pkg/sandbox/kit/kit.go
```diff
@@ -36,11 +36,13 @@ import (
 	"path/filepath"
 	"slices"
 	"sort"
+	"strconv"
 	"strings"
```
aheritier, line 40: **Nit — unnecessary `strconv` import.** `fmt` is already imported; `fmt.Sprintf("%d", n)` is simpler and avoids pulling in `strconv` for a single call site:

```go
return fmt.Sprintf("%s %ss", styleCount(fmt.Sprintf("%d", n)), what)
```

pkg/sandbox/kit/kit.go
```diff
@@ -358,14 +371,26 @@ func loadConfig(ctx context.Context, opts Options) (*latestcfg.Config, error) {
...
+// enables local skills without a filter is the wide case — every local
+// skill is staged.
+func stageSkills(kitDir string, cfg *latestcfg.Config) ([]Entry, []Redaction, error) {
 	target := filepath.Join(kitDir, skills.KitSkillsSubdir)
 	if err := os.MkdirAll(target, 0o750); err != nil {
 		return nil, nil, fmt.Errorf("creating kit skills dir: %w", err)
 	}
 
+	include, ok := localSkillFilter(cfg)
+	if !ok {
```
aheritier, line 390: **Nit — `os.MkdirAll` runs before the early-return check.** If `localSkillFilter` returns `(nil, false)` (no agent enables local skills), an empty `<kit>/skills/` directory is created and then immediately abandoned. Harmless, but moving `os.MkdirAll` to after the `!ok` guard would be tidier:

```go
include, ok := localSkillFilter(cfg)
if !ok {
    return nil, nil, nil
}
if err := os.MkdirAll(target, 0o750); err != nil {
    return nil, nil, fmt.Errorf("creating kit skills dir: %w", err)
}
```

aheritier (dismissed):
Approve with nits — core logic is sound, UX improvements are well-executed, and test coverage is thorough. CI is green.

**Behavioral note:** agents without `skills: true` (or an include list) will no longer have local skills staged into the kit. The builtin `default` agent already has `skills: true` so the common path is unaffected, but user-written configs that relied on the previous "always-stage" behaviour will silently lose local skills. A migration note in the release/changelog would reduce surprise.

aheritier (approved):

PR #2879, opened 2026-05-22
fix: don't close shared session store in runtime.Close

### LINKED ISSUE #2872: failed to load sessions: sql: database is closed
## Description

I wanted to open an existing session with `/sessions` and got the error `failed to load sessions: sql: database is closed`

## Expected Behavior

No error

## Actual Behavior

Error `failed to load sessions: sql: database is closed`

## Steps to Reproduce

1. call `/sessions`

## Environment

- **Docker Agent Version**: main branch on May 22nd
- **OS & Terminal**: macOS + cmux

## Error Output

If applicable, paste the full error message or stack trace:

```
failed to load sessions: sql: database is closed
```

## Screenshots

Image

dgageot:
The TUI was failing to open `/sessions` after switching to a session with a different working directory, with the error "failed to load sessions: sql: database is closed". This happened because when the TUI swapped in a new runtime for the different working directory, it canceled the old runtime's context and ran cleanup, which called `rt.Close()` → `r.sessionStore.Close()`. However, the spawned new runtime was given the same SQLite store (via `createSessionSpawner` reusing `rt.SessionStore()`), so closing it from the old runtime broke every subsequent read against that store.

The fix moves session store ownership to the backend layer, which now owns its lifecycle independently. The `Runtime.Close()` contract is updated to make explicit that the session store is owned by the embedder and may be shared. `localBackend` now lazily opens the SQLite store on first `CreateSession` call (via `sync.Once`) and closes it once in a new `backend.Close()` method. The runtime no longer closes the store it was given, and the TUI defers `b.Close()` to clean up resources when the entire session ends.

Fixes #2872

### DIFF
--- a/cmd/root/backend.go
+++ b/cmd/root/backend.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/docker/docker-agent/pkg/config"
+	pathx "github.com/docker/docker-agent/pkg/path"
 	"github.com/docker/docker-agent/pkg/runtime"
 	"github.com/docker/docker-agent/pkg/session"
 	"github.com/docker/docker-agent/pkg/teamloader"
@@ -28,6 +29,11 @@ type backend interface {
 	CreateSession(ctx context.Context, loaded *teamloader.LoadResult, req runtime.CreateSessionRequest) (runtime.Runtime, *session.Session, func(), error)
 
 	Spawner(rt runtime.Runtime) tui.SessionSpawner
+
+	// Close releases backend-owned resources (e.g., the shared session
+	// store). It is called once when the embedder is shutting down,
+	// after every per-session cleanup has run.
+	Close() error
 }
 
 // selectBackend picks the backend implied by the current flags.
@@ -43,9 +49,20 @@ func (f *runExecFlags) selectBackend(agentFileName string) (backend, error) {
 }
 
 // localBackend builds the in-process runtime and session.
+//
+// The session store is owned by the backend (not by individual
+// runtimes) because the TUI's session spawner reuses the same store
+// across spawned sessions. Closing it inside a per-session cleanup
+// would break later session lookups (issue #2872). The store is
+// lazily opened on the first CreateSession call and closed once when
+// Close is invoked.
 type localBackend struct {
 	flags       *runExecFlags
 	agentSource config.Source
+
+	storeOnce sync.Once
+	storeErr  error
+	store     session.Store
 }
 
 func (b *localBackend) LoadTeamRequest() runtime.LoadTeamRequest {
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the shared session store lifecycle issue. The  pattern for lazy initialization in  is a sound approach,  is properly deferred in , and the  contract update is clearly documented. No resource leaks, double-close, or use-after-close issues were identified in the changed code.

cmd/root/backend.go
```diff
@@ -83,6 +126,13 @@ func (b *localBackend) Spawner(rt runtime.Runtime) tui.SessionSpawner {
 	return b.flags.createSessionSpawner(b.agentSource, rt.SessionStore())
 }
 
+func (b *localBackend) Close() error {
+	if b.store == nil {
```
docker-agent, line 134: **[MEDIUM] Potential data race: `Close()` reads `b.store` outside `sync.Once`**

`localBackend.Close()` reads `b.store` directly (line 130) without any synchronization, while `sessionStore()` writes `b.store` inside `storeOnce.Do(...)`. The Go memory model only guarantees that the write to `b.store` inside the `Do` closure is visible *after* `Do` returns — it does not provide a happens-before guarantee for concurrent reads of `b.store` that happen outside of `Do`.

If `Close()` races with an in-progress `storeOnce.Do()` call (e.g., a session is still being initialized when shutdown begins), the Go race detector would flag this.

**Suggested fix** — either use `storeOnce.Do` to query the store in `Close()` as well, or protect `b.store` with a mutex:

```go
func (b *localBackend) Close() error {
    b.storeOnce.Do(func() {}) // ensure the write, if any, has completed
    if b.store == nil {
        return nil
    }
    return b.store.Close()
}
```

Or, more robustly, add a `sync.Mutex` to guard access to `b.store` in both `sessionStore` and `Close`.

docker-agent (commented):

The fix correctly moves session-store ownership to the backend layer, resolving the use-after-close bug (#2872). One potential data race was found in the new `localBackend.Close()` method.

cmd/root/backend.go
```diff
@@ -60,8 +77,34 @@ func (b *localBackend) CreateSessionRequest(workingDir string) runtime.CreateSes
 	return b.flags.createSessionRequest(workingDir)
 }
 
+// sessionStore returns the backend-owned session store, opening it on
+// first use. The store is shared by the initial runtime and by every
+// runtime spawned by [localBackend.Spawner].
+func (b *localBackend) sessionStore(req runtime.CreateSessionRequest) (session.Store, error) {
+	b.storeOnce.Do(func() {
```
aheritier, line 84: **Sticky `sync.Once` failure (low severity):** if `ExpandHomeDir` or `NewSQLiteSessionStore` errors here, `storeOnce` is permanently consumed and every subsequent call to `sessionStore()` will return `(nil, storeErr)` — even if the underlying problem has been corrected (e.g., a retry with a different `req`). In the current call graph `CreateSession` is called exactly once, so this is not a practical issue. Flagging in case future callers add retry logic.

cmd/root/backend.go
aheritier, line 134: **Nil guard is correct and defensive.** `b.store` can only be `nil` here if `CreateSession` was never called (e.g., an early-exit code path) or if `storeOnce` recorded an error without setting `b.store`. The check is not racy: `b.Close()` is invoked from the `defer` registered after `selectBackend()`, by which point any concurrent `Do()` invocations have completed. Technically redundant on the happy path, but harmless and good defensive practice.

aheritier (dismissed):
Clean fix for the shared-store double-close bug (#2872). Ownership transfer to `localBackend` is the right architectural move, defer ordering in `runOrExec` is correct (LIFO ensures `cleanup()` runs before `b.Close()`), and the regression test in `close_test.go` directly exercises the fixed invariant. CI is green.

Two minor non-blocking observations (inline):

1. **Sticky `sync.Once` failure** — if `sessionStore()` errors on first call the `Once` is consumed; retries permanently get the cached error. Not a real problem given the current single-call graph, but worth noting for future callers.
2. **`b.store == nil` guard in `Close()`** is technically unreachable after a successful `Do()`, but it is correct and defensively handles the no-`CreateSession`-called path.

Pre-existing gap (out of scope): spawned runtimes' cleanup closure does not call `localRt.Close()`, so `bgAgents.StopAll()` is not invoked for spawned runtimes. Not introduced here; worth a follow-up issue.

aheritier (approved):

PR #2882, opened 2026-05-22
a2a: honour `timeout` and `allow_private_ips` config (with SSRF protection)

dgageot:
Aligns the `a2a` builtin toolset with the other HTTP-based builtin toolsets (`fetch`, `api`, `openapi`, remote `mcp`) so it honours the shared `timeout` and `allow_private_ips` configuration fields.

Before this change, `a2a` was the odd one out. It silently ignored `timeout:` and `allow_private_ips:` from the config, and it used `httpclient.NewHTTPClient` instead of `httpclient.NewSafeClient`, meaning an a2a `url:` could reach loopback, RFC1918, and link-local addresses (notably the cloud-metadata endpoint at `169.254.169.254`) — bypassing the SSRF protection applied to every other HTTP-based builtin toolset. There was also no operator knob to override the default timeout for a slow a2a peer.

This change adds `WithTimeout` and `WithAllowPrivateIPs` options on the a2a toolset, updates `Start` to use `httpclient.NewSafeClient(timeout, allowPrivateIPs)` for both the agent-card resolver and the JSON-RPC streaming HTTP client, hoists the duplicated `30 * time.Second` default into `httpclient.DefaultToolHTTPTimeout` for a single source of truth, and validates `allow_private_ips` on `a2a` while passing `a2aclient.WithDefaultsDisabled()` to avoid SSRF-safe HTTP client bypass via gRPC transport fallback.

### DIFF
--- a/pkg/config/latest/validate.go
+++ b/pkg/config/latest/validate.go
@@ -145,8 +145,8 @@ func (t *Toolset) validate() error {
 	if len(t.BlockedDomains) > 0 && t.Type != "fetch" {
 		return errors.New("blocked_domains can only be used with type 'fetch'")
 	}
-	if t.AllowPrivateIPsEnabled() && t.Type != "fetch" && t.Type != "mcp" && t.Type != "api" && t.Type != "openapi" {
-		return errors.New("allow_private_ips can only be used with type 'fetch', 'api', 'openapi' or remote MCP toolsets")
+	if t.AllowPrivateIPsEnabled() && t.Type != "fetch" && t.Type != "mcp" && t.Type != "api" && t.Type != "openapi" && t.Type != "a2a" {
+		return errors.New("allow_private_ips can only be used with type 'fetch', 'api', 'openapi', 'a2a' or remote MCP toolsets")
 	}
 	if len(t.AllowedDomains) > 0 && len(t.BlockedDomains) > 0 {
 		return errors.New("allowed_domains and blocked_domains are mutually exclusive")
@@ -235,7 +235,7 @@ func (t *Toolset) validate() error {
 			return errors.New("either command, remote or ref must be set, but only one of those")
 		}
 		if t.AllowPrivateIPsEnabled() && t.Remote.URL == "" && t.Ref == "" {
-			return errors.New("allow_private_ips can only be used with type 'fetch', 'api', 'openapi' or remote MCP toolsets")
+			return errors.New("allow_private_ips can only be used with type 'fetch', 'api', 'openapi', 'a2a' or remote MCP toolsets")
 		}
 		if t.Remote.OAuth != nil {
 			if t.Remote.URL == "" {
--- a/pkg/tools/a2a/a2a_test.go
+++ b/pkg/tools/a2a/a2a_test.go
@@ -1,0 +1,86 @@
+package a2a
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	goa2a "github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/a2asrv"
+	"github.com/a2aproject/a2a-go/a2asrv/eventqueue"
+
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+func TestToolSetRejectsPrivateIPForAgentCard(t *testing.T) {
+	t.Parallel()
+
+	toolSet := NewToolset("test", "http://127.0.0.1/.well-known/agent-card.json", nil)
+
+	if err := toolSet.Start(t.Context()); err == nil {
+		t.Fatal("Start() expected error")
+	}
+}
+
+func TestToolSetStreamingWithAllowPrivateIPs(t *testing.T) {
+	t.Parallel()
+
+	server := httptest.NewServer(a2asrv.NewJSONRPCHandler(a2asrv.NewHandler(testA2AHandler{})))
+	t.Cleanup(server.Close)
+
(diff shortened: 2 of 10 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes correctly align the `a2a` toolset with the other HTTP-based builtins:

- **SSRF protection**: Both the agent-card resolver and the JSON-RPC streaming client now use `httpclient.NewSafeClient`, closing the bypass that the old `httpclient.NewHTTPClient` left open.
- **`WithDefaultsDisabled()`**: Properly prevents a2a-go's own default transport from silently re-opening a gRPC/plaintext path that would circumvent the safe client.
- **`nil` transport guard**: The `base == nil → http.DefaultTransport` fallback before wrapping with `HeaderTransport` is sound — `NewSafeClient` always sets a transport, so this path is a belt-and-suspenders guard that does no harm.
- **Timeout propagation**: Default is set in `NewToolset`, options are applied after, `CreateToolSet` wires config values correctly — no ordering bug.
- **Validation**: `allow_private_ips` whitelist extended to include `a2a` in both the outer type check and the MCP-local-command guard; error messages updated consistently.
- **Constant consolidation**: `DefaultToolHTTPTimeout` cleanly replaces the three independent `30 * time.Second` literals with no change in behaviour.
- **Tests**: `TestToolSetRejectsPrivateIPForAgentCard` exercises the SSRF rejection path; `TestToolSetStreamingWithAllowPrivateIPs` exercises the full happy path with a real in-process a2a server. Coverage is proportional to the risk.

No bugs or security issues found in the changed code.

pkg/config/latest/validate.go
```diff
@@ -235,7 +235,7 @@ func (t *Toolset) validate() error {
 			return errors.New("either command, remote or ref must be set, but only one of those")
 		}
 		if t.AllowPrivateIPsEnabled() && t.Remote.URL == "" && t.Ref == "" {
-			return errors.New("allow_private_ips can only be used with type 'fetch', 'api', 'openapi' or remote MCP toolsets")
+			return errors.New("allow_private_ips can only be used with type 'fetch', 'api', 'openapi', 'a2a' or remote MCP toolsets")
```
aheritier, line 238: This error is only reachable for `type: mcp` with a local command (the `case "mcp"` block, inside `t.AllowPrivateIPsEnabled() && t.Remote.URL == "" && t.Ref == ""`). Listing `'a2a'` in the message can mislead an operator into thinking they should switch to `a2a`, rather than understanding the real issue: their MCP toolset must use a remote URL or ref.

A more precise message, e.g.:

```
allow_private_ips requires a remote MCP server (remote.url or ref); local MCP commands do not make outbound HTTP calls
```

would make the corrective action obvious. Nit — not a blocker.

pkg/tools/a2a/a2a_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+type testA2AHandler struct{}
+
+func (testA2AHandler) Execute(ctx context.Context, reqCtx *a2asrv.RequestContext, queue eventqueue.Queue) error {
+	return queue.Write(ctx, goa2a.NewMessageForTask(goa2a.MessageRoleAgent, reqCtx, goa2a.TextPart{Text: "ok"}))
+}
+
+func (testA2AHandler) Cancel(context.Context, *a2asrv.RequestContext, eventqueue.Queue) error {
+	return nil
+}
```
aheritier, line 86: No test for `WithTimeout` behaviour. Consider adding one: spin up a `httptest` server that sleeps before responding, call `NewToolset` with `WithTimeout(1*time.Millisecond)`, and assert that `Start()` returns a deadline-exceeded error. The current tests cover SSRF rejection and the `allow_private_ips` happy path, but the timeout wiring is unobserved in CI. Non-blocking given the simplicity of the delegation.

aheritier (approved):
SSRF fix is complete and correct. Both HTTP call sites — agent-card resolution and streaming JSON-RPC — are now routed through `httpclient.NewSafeClient`, the nil-transport guard is sound, `WithDefaultsDisabled()` closes the gRPC-fallback bypass, and the `DefaultToolHTTPTimeout` consolidation has no behavioural change. All CI checks pass.

Three non-blocking items below.

**1. Slightly misleading error message in the local-command MCP branch (`validate.go` ~line 238)**

The message `"allow_private_ips can only be used with type 'fetch', 'api', 'openapi', 'a2a' or remote MCP toolsets"` is only reachable when `type: mcp` with a local command (no `remote.url` or `ref`). Listing `'a2a'` as a valid alternative may confuse a user into switching to `a2a` rather than understanding that their MCP toolset must be remote. A message like `"allow_private_ips requires a remote MCP server (remote.url or ref); local MCP commands do not make outbound HTTP calls"` would be unambiguous. Nit — not a blocker.

**2. No test for `WithTimeout` behaviour**

`a2a_test.go` covers SSRF rejection and the `allow_private_ips` happy path, but there is no test that a slow server is cut off when a short timeout is configured (e.g. `httptest` server with an artificial delay + `WithTimeout(1*time.Millisecond)`). Given the simplicity of the delegation to `NewSafeClient`, this is acceptable, but would make the wiring observable in CI.

**3. Missing example YAML for new config features**

`AGENTS.md` says: *"When adding new features to the config: Create an example YAML that demonstrates the new feature."* The PR adds `timeout` and `allow_private_ips` support to `a2a` toolsets but does not include an example YAML. A snippet like:

```yaml
toolsets:
  - type: a2a
    url: https://my-internal-agent.example.com
    timeout: 60
    allow_private_ips: true
```

would satisfy the convention. Non-blocking, but worth a follow-up.

PR #2888, opened 2026-05-23
feat(sandbox): alias/runtime sandbox defaults and persistent network allowlist

dgageot:
This PR makes sandbox enablement configurable from three new sources, and lets users curate a persistent network allowlist for hosts the kit's tool resolver can't infer (custom MCP servers, third-party APIs, registries not covered by aqua, ...).

## Sandbox-by-default sources

Three new ways to opt an agent into the sandbox without typing `--sandbox`:

- **Aliases** declared in `~/.config/cagent/config.yaml`:
  ```yaml
  aliases:
    my-agent:
      ref: ...
      sandbox: true
  ```
  Also exposed via `docker agent alias add --sandbox`.

- **Agent runtime config** in the YAML itself:
  ```yaml
  runtime:
    sandbox: true
  ```
  Per-agent counterpart of the alias flag, useful for catalog agents that always need filesystem or network isolation.

- An explicit `--sandbox=false` on the CLI still wins over both, matching the precedence the README and JSON schema already promise. The dispatch logic was previously falling through to the alias / runtime defaults when the user passed `--sandbox=false` because it only checked `!f.sandbox`; it now consults `cmd.Flags().Changed("sandbox")` and the precedence is extracted into a testable `resolveSandboxDefault` helper.

## Network allowlist

Two coordinated additions to widen the default-deny proxy policy without sliding back to the conservative host union the kit falls back to on resolver failure:

- **Author-declared, per agent**:
  ```yaml
  runtime:
    sandbox: true
    network_allowlist:
      - api.example.com
      - registry.npmjs.org
  ```

- **User-managed, persistent** for hosts the agent author can't anticipate:
  ```
  docker agent sandbox allow <host> [<host>...]
  docker agent sandbox deny <host>
  docker agent sandbox list
  ```
  Entries land in `~/.config/cagent/config.yaml` under `sandbox_allowlist` and are unioned on every subsequent `--sandbox` run.

Both lists are unioned with the gateway and the kit-resolved tool install hosts. Commas and embedded whitespace are rejected at write time so a single entry can't smuggle several rules into the policy engine. The launch summary now prints each source on its own line so it's clear which layer punched which hole, and the tool-install summary surfaces a hint pointing at `sandbox allow` when the per-toolset resolver had to fall back.

## Files

```
agent-schema.json                   |  22 +++
cmd/root/alias.go                   |  15 +-
cmd/root/root.go                    |   1 +
cmd/root/run.go                     |  13 +-
cmd/root/sandbox.go                 | 132 ++++++++++++++++-
cmd/root/sandbox_cmd.go             | 157 +++++++++++++++++++++
cmd/root/sandbox_cmd_test.go        |  52 +++++++
cmd/root/sandbox_test.go            | 127 +++++++++++++++-
docs/configuration/sandbox/index.md |  90 ++++++++++++
docs/features/cli/index.md          |  22 +++
examples/sandbox_agent.yaml         |  18 ++-
pkg/config/latest/types.go          |  26 ++++
pkg/config/resolve_test.go          |  18 +++
pkg/userconfig/userconfig.go        |  69 +++++++++-
pkg/userconfig/userconfig_test.go   |  60 +++++++-
```

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -170,6 +170,15 @@ func (f *runExecFlags) runRunCommand(cmd *cobra.Command, args []string) (command
 		}()
 	}
 
+	// Resolve alias-driven sandbox opt-in before dispatch so aliases
+	// like `docker-agent run scary-agent` (where the alias declares
+	// sandbox: true) take the sandbox path without an explicit flag.
+	if !f.sandbox && len(args) > 0 {
+		if alias := config.ResolveAlias(args[0]); alias != nil && alias.Sandbox {
+			f.sandbox = true
+		}
+	}
+
 	if f.sandbox {
 		return runInSandbox(ctx, cmd, args, &f.runConfig, f.sandboxTemplate, f.sbx, f.noKit)
 	}
@@ -218,7 +227,7 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 	// Apply alias options if this is an alias reference
 	// Alias options only apply if the flag wasn't explicitly set by the user
 	if alias := config.ResolveAlias(agentFileName); alias != nil {
-		slog.DebugContext(ctx, "Applying alias options", "yolo", alias.Yolo, "model", alias.Model, "hide_tool_results", alias.HideToolResults)
+		slog.DebugContext(ctx, "Applying alias options", "yolo", alias.Yolo, "model", alias.Model, "hide_tool_results", alias.HideToolResults, "sandbox", alias.Sandbox)
 		if alias.Yolo && !f.autoApprove {
 			f.autoApprove = true
 		}
@@ -227,6 +236,9 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		}
 		if alias.HideToolResults && !f.hideToolResults {
 			f.hideToolResults = true
+		}
+		if alias.Sandbox && !f.sandbox {
+			f.sandbox = true
 		}
 	}
 
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -18,6 +18,7 @@ import (
 	"github.com/docker/docker-agent/pkg/app"
 	"github.com/docker/docker-agent/pkg/cli"
 	"github.com/docker/docker-agent/pkg/config"
+	latestcfg "github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/hooks"
 	"github.com/docker/docker-agent/pkg/hooks/builtins"
 	pathx "github.com/docker/docker-agent/pkg/path"
@@ -173,12 +174,17 @@ func (f *runExecFlags) runRunCommand(cmd *cobra.Command, args []string) (command
 	// Resolve alias / runtime-declared sandbox opt-in before dispatch.
 	// An explicit --sandbox=<bool> on the CLI always wins, so we only
 	// consult the lower-priority sources when the flag wasn't set.
+	var agentCfg *latestcfg.Config
 	if !cmd.Flags().Changed("sandbox") {
-		f.sandbox = resolveSandboxDefault(ctx, args, f.sandbox)
+		var agentRef string
+		if len(args) > 0 {
+			agentRef = args[0]
+		}
+		f.sandbox, agentCfg = resolveSandboxDefault(ctx, agentRef, f.sandbox)
 	}
 
 	if f.sandbox {
-		return runInSandbox(ctx, cmd, args, &f.runConfig, f.sandboxTemplate, f.sbx, f.noKit)
+		return runInSandbox(ctx, cmd, args, &f.runConfig, f.sandboxTemplate, f.sbx, f.noKit, agentCfg)
 	}
 
 	out := cli.NewPrinter(cmd.OutOrStdout())
@@ -236,9 +242,9 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 			f.hideToolResults = true
 		}
 		// alias.Sandbox is consumed earlier in runRunCommand before
-		// dispatch; by the time we reach runOrExec the sandbox path
-		// has already been taken (or the user opted out via
-		// --sandbox=false), so flipping it here would be a no-op.
+		// dispatch; reaching runOrExec means the sandbox decision
+		// resolved to false (or the user opted out via --sandbox=false),
+		// so flipping it here would be a no-op.
 	}
 
 	// Build global permissions checker from user config settings.
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -179,6 +179,15 @@ func (f *runExecFlags) runRunCommand(cmd *cobra.Command, args []string) (command
 		}
 	}
 
+	// Honour `runtime.sandbox: true` declared by the agent author.
+	// Best-effort: a config we can't peek at falls through to the
+	// normal path which will surface the load error properly.
+	if !f.sandbox && len(args) > 0 {
+		if peekAgentSandbox(ctx, args[0]) {
+			f.sandbox = true
+		}
+	}
+
 	if f.sandbox {
 		return runInSandbox(ctx, cmd, args, &f.runConfig, f.sandboxTemplate, f.sbx, f.noKit)
 	}
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -170,22 +170,11 @@ func (f *runExecFlags) runRunCommand(cmd *cobra.Command, args []string) (command
 		}()
 	}
 
-	// Resolve alias-driven sandbox opt-in before dispatch so aliases
-	// like `docker-agent run scary-agent` (where the alias declares
-	// sandbox: true) take the sandbox path without an explicit flag.
-	if !f.sandbox && len(args) > 0 {
-		if alias := config.ResolveAlias(args[0]); alias != nil && alias.Sandbox {
-			f.sandbox = true
-		}
-	}
-
-	// Honour `runtime.sandbox: true` declared by the agent author.
-	// Best-effort: a config we can't peek at falls through to the
-	// normal path which will surface the load error properly.
-	if !f.sandbox && len(args) > 0 {
-		if peekAgentSandbox(ctx, args[0]) {
-			f.sandbox = true
-		}
+	// Resolve alias / runtime-declared sandbox opt-in before dispatch.
+	// An explicit --sandbox=<bool> on the CLI always wins, so we only
+	// consult the lower-priority sources when the flag wasn't set.
+	if !cmd.Flags().Changed("sandbox") {
+		f.sandbox = resolveSandboxDefault(ctx, args, f.sandbox)
 	}
 
 	if f.sandbox {
@@ -246,9 +235,10 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		if alias.HideToolResults && !f.hideToolResults {
 			f.hideToolResults = true
 		}
-		if alias.Sandbox && !f.sandbox {
-			f.sandbox = true
-		}
+		// alias.Sandbox is consumed earlier in runRunCommand before
+		// dispatch; by the time we reach runOrExec the sandbox path
+		// has already been taken (or the user opted out via
+		// --sandbox=false), so flipping it here would be a no-op.
 	}
 
 	// Build global permissions checker from user config settings.
--- a/cmd/root/sandbox.go
+++ b/cmd/root/sandbox.go
@@ -24,6 +24,7 @@ import (
 	"github.com/docker/docker-agent/pkg/sandbox"
 	"github.com/docker/docker-agent/pkg/sandbox/kit"
 	"github.com/docker/docker-agent/pkg/skills"
+	"github.com/docker/docker-agent/pkg/userconfig"
 )
 
 // peekAgentSandbox returns true when the agent referenced by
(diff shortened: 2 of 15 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are well-structured and implement the sandbox-by-default and persistent network allowlist features cleanly. No bugs were identified in the added code.

- `resolveSandboxDefault` helper correctly consults `cmd.Flags().Changed("sandbox")` so explicit `--sandbox=false` wins over alias/runtime defaults
- Host validation for the allowlist rejects commas and embedded whitespace at write time
- Persistent allowlist entries are stored atomically under `~/.config/cagent/config.yaml`
- Union of user allowlist, agent-declared allowlist, gateway, and tool-install hosts is correctly composed
- New `sandbox allow/deny/list` subcommands are properly wired and tested
- Schema additions for `runtime.sandbox` and `runtime.network_allowlist` are consistent with the types in `pkg/config/latest/types.go`

aheritier (changes requested):
Two blocking issues must be addressed before merge. Eight non-blocking suggestions below.

---

## 🔴 Blocking

### 1. Agent config loaded 2–3× per sandboxed run

`resolveSandboxDefault` calls `loadAgentConfig` once to decide whether to sandbox. Then `agentNetworkAllowlist` (inside `runInSandbox`) calls it a second time on the same ref. `kit.Build` resolves the same config a third time. For OCI-registry refs each `config.Load` is a network round-trip, so every sandboxed run of an OCI-hosted agent pays 3× that cost before work begins.

**Fix:** cache the `*latestcfg.Config` returned from `resolveSandboxDefault` and thread it through to `runInSandbox`, eliminating the second `loadAgentConfig` call in `agentNetworkAllowlist`. At minimum, document the duplication as a known inefficiency.

### 2. `network_allowlist` entries not validated at YAML load time

The comment on `NetworkAllowlist` says _"commas and whitespace are rejected"_, but that validation only lives in `AddSandboxHosts` (the CLI write path). An agent YAML with `"api.example.com, registry.npmjs.org"` as a single entry passes `config.Load` without error, bypasses `AddSandboxHosts` entirely, and feeds the malformed host straight into `allowSandboxHosts` and the proxy policy engine.

**Fix:** validate entries in `agentNetworkAllowlist` before returning them, e.g.:

```go
var valid []string
for _, h := range cfg.Runtime.NetworkAllowlist {
    if strings.ContainsAny(h, ", \t") {
        slog.WarnContext(ctx, "Ignoring invalid network_allowlist entry", "host", h)
        continue
    }
    valid = append(valid, h)
}
return valid
```

---

## 🟡 Non-blocking suggestions

### 3. `resolveSandboxDefault` signature takes `[]string` but only uses `[0]`

Narrowing to `agentRef string` makes the function self-contained and removes the implicit `args[0]` contract from the call-site.

```go
func resolveSandboxDefault(ctx context.Context, agentRef string, current bool) bool
```

### 4. `sandbox deny` is non-idempotent — errors for an already-absent host

Remove operations are almost universally idempotent in CLIs (`docker rm --force`, `git branch -D`, etc.). Returning an error with a non-zero exit code when the host is already absent surprises scripts. Consider `slog.WarnContext` + `return nil`.

### 5. `AddSandboxHosts` leaves in-memory state partially mutated on error

If a batch contains a valid host followed by a malformed one, `c.SandboxAllowlist` is mutated for the valid entries before the error is returned. The caller checks `err` before `Save()`, so disk is never corrupted — but the in-memory `*Config` is left inconsistent. A future caller that reuses the object after a failed `AddSandboxHosts` would observe partial state.

**Fix:** validate all entries in a first pass, then mutate in a second pass only if all are valid.

### 6. Sandbox hint shown unconditionally from `printToolInstallAllowance`

The `"hint: persist a missing host..."` line is correctly scoped today (only called from `runInSandbox`), but `kit.Build` is exported and `printToolInstallAllowance` could be called from a non-sandbox path in the future. Minor forward-compat note.

### 7. No test for malformed YAML `network_allowlist` entries

`TestAgentNetworkAllowlist` covers valid entries. There is no test asserting what happens when an agent YAML contains `"api.example.com, other.com"` as a single entry. Without validation at parse time (issue #2), this silently passes through — a test documenting that behaviour makes a future fix verifiable.

### 8. Misleading comment in `runOrExec` on `alias.Sandbox`

The comment says _"the sandbox path has already been taken"_. Since we are in `runOrExec` (the non-sandbox branch), the sandbox path was **not** taken. The comment should read: _"the sandbox decision was already made in runRunCommand; since we are here, it resolved to false."_

### 9. Redundant `args` in `list` telemetry

`cobra.NoArgs` is set on `list`, so `args` is always empty. `append([]string{"list"}, args...)` is harmless but needlessly confusing — `args` can be dropped from the telemetry call.

### 10. Raw hostnames in telemetry

`allow` and `deny` forward user-supplied hostnames into `TrackCommand`. Sandbox hosts are more likely to be sensitive internal names (corp proxies, internal registries) than agent refs. Worth confirming this is intentional before the PR lands.

---

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Blocking | `cmd/root/sandbox.go` | Config loaded 2–3× per sandboxed run |
| 2 | 🔴 Blocking | `pkg/config/latest/types.go` | `network_allowlist` not validated at YAML load |
| 3 | 🟡 | `cmd/root/sandbox.go` | `resolveSandboxDefault` takes `[]string`, uses `[0]` |
| 4 | 🟡 | `cmd/root/sandbox_cmd.go` | `deny` non-idempotent |
| 5 | 🟡 | `pkg/userconfig/userconfig.go` | Partial mutation on validation failure |
| 6 | 🟡 | `cmd/root/sandbox.go` | Sandbox hint unconditional |
| 7 | 🟡 | `cmd/root/sandbox_test.go` | No test for malformed YAML allowlist entries |
| 8 | 🟡 | `cmd/root/run.go` | Misleading `alias.Sandbox` comment |
| 9 | 🟡 | `cmd/root/sandbox_cmd.go` | Redundant `args` in `list` telemetry |
| 10 | 🟡 | `cmd/root/sandbox_cmd.go` | Raw hostnames in telemetry |

cmd/root/sandbox.go
```diff
@@ -18,13 +18,87 @@ import (
...
+	return peekAgentSandbox(ctx, args[0])
+}
+
+// agentNetworkAllowlist returns the hostnames the agent declared in
+// runtime.network_allowlist. Same best-effort contract as
+// peekAgentSandbox: any failure to load the config returns nil so
+// the run continues with the inferred host set only.
+func agentNetworkAllowlist(ctx context.Context, agentRef string) []string {
+	cfg := loadAgentConfig(ctx, agentRef)
+	if cfg == nil || cfg.Runtime == nil {
```
aheritier, line 63: **[Blocking #1]** Second `loadAgentConfig` call for the same `agentRef`. `resolveSandboxDefault` already called it once (via `peekAgentSandbox`) to decide *whether* to sandbox. For OCI refs that is two network round-trips before work begins, and `kit.Build` adds a third. Cache the resolved `*latestcfg.Config` in `resolveSandboxDefault` and thread it into `runInSandbox` so `agentNetworkAllowlist` can consume it without another load.

cmd/root/sandbox.go
```diff
@@ -85,8 +159,13 @@ func runInSandbox(ctx context.Context, cmd *cobra.Command, args []string, runCon
 		}
 	}
 
+	agentHosts := agentNetworkAllowlist(ctx, agentRef)
+	userHosts := userSandboxAllowlist(ctx)
```
aheritier, line 175: **[Blocking #1 — call site]** `agentNetworkAllowlist` triggers the redundant `loadAgentConfig` call here. If the resolved config were threaded from `resolveSandboxDefault`, this call and its extra network round-trip for OCI refs would be unnecessary.

pkg/config/latest/types.go
```diff
@@ -28,6 +28,32 @@ type Config struct {
...
+	// runner already opens automatically; commas and whitespace are
+	// rejected to keep a single entry from smuggling several rules
+	// into the policy engine.
+	//
+	// Use this when an agent's tools call hosts the auto-installer
+	// can't infer (custom MCP endpoints, third-party APIs, registries
+	// not covered by the aqua resolver, etc.) instead of relying on
+	// the wider fallback host set the kit uses on resolution failures.
+	NetworkAllowlist []string `json:"network_allowlist,omitempty" yaml:"network_allowlist,omitempty"`
 }
```
aheritier, line 57: **[Blocking #2]** The comment says *"commas and whitespace are rejected"* but that validation only exists in `AddSandboxHosts` (the CLI write path). An agent YAML with `"api.example.com, registry.npmjs.org"` as a single entry passes `config.Load` without error, bypasses `AddSandboxHosts` entirely, and feeds the malformed host directly into `allowSandboxHosts` and the proxy policy engine. Validate entries in `agentNetworkAllowlist` (or add a `Validate()` method on `RuntimeDefaults`) to enforce the invariant on the read path.

cmd/root/sandbox.go
```diff
@@ -18,13 +18,87 @@ import (
...
+// taken when the user did not pass --sandbox on the CLI. The first
+// source that declares sandbox: true wins; in priority order:
+//
+//  1. an alias entry (`docker agent alias add ... --sandbox`);
+//  2. the agent's own `runtime.sandbox: true`.
+//
+// Callers must only invoke this when the CLI flag was not set; an
+// explicit --sandbox=<bool> always wins and bypasses this logic.
+func resolveSandboxDefault(ctx context.Context, args []string, current bool) bool {
+	if current || len(args) == 0 {
```
aheritier: **[Suggestion #3]** The function only reads `args[0]`; accepting the full slice is misleading. Consider narrowing to `agentRef string` so the function is self-contained and callers pass `args[0]` explicitly:\n```go\nfunc resolveSandboxDefault(ctx context.Context, agentRef string, current bool) bool\n```

cmd/root/sandbox_cmd.go
```diff
@@ -0,0 +1,157 @@
...
+	host := strings.TrimSpace(args[0])
+
+	cfg, err := userconfig.Load()
+	if err != nil {
+		return fmt.Errorf("failed to load config: %w", err)
+	}
+
+	if !cfg.RemoveSandboxHost(host) {
+		return fmt.Errorf("host %q is not on the allowlist", host)
+	}
```
aheritier, line 129: **[Suggestion #4]** Remove operations are conventionally idempotent in CLIs (`docker rm --force`, `git branch -D`). Returning a non-zero exit code when the host is already absent surprises scripts that run `deny` as a cleanup step. Consider `slog.WarnContext(cmd.Context(), "host not on allowlist", "host", host)` + `return nil`.

pkg/userconfig/userconfig.go
```diff
@@ -154,6 +157,14 @@ type Config struct {
 	Settings *Settings `yaml:"settings,omitempty"`
 	// CredentialHelper configures an external command to retrieve Docker credentials
 	CredentialHelper *CredentialHelper `yaml:"credential_helper,omitempty"`
+	// SandboxAllowlist is the persistent list of hosts the user has
+	// taught docker-agent to open in the sandbox proxy on every run
+	// (in addition to the gateway, the kit-resolved tool install
+	// hosts, and the agent-declared runtime.network_allowlist).
+	// Managed via `docker agent sandbox allow/deny/list`. Each entry
+	// is a hostname with an optional ":port" suffix; commas and
```
aheritier, line 165: **[Suggestion #5]** The error is returned here, but `c.SandboxAllowlist` is already mutated for any valid hosts that appeared before this entry in the batch. The caller checks `err` before `cfg.Save()` so disk is never corrupted, but the in-memory `*Config` is left partially updated. A future caller reusing the object after a failed `AddSandboxHosts` would observe inconsistent state. Fix: validate all entries in a first pass before mutating anything.

cmd/root/sandbox_cmd.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("failed to save config: %w", err)
+	}
+	out.Printf("Removed %s from the persistent sandbox allowlist.\n", host)
+	return nil
+}
+
+func runSandboxListCommand(cmd *cobra.Command, args []string) (commandErr error) {
+	telemetry.TrackCommand(cmd.Context(), "sandbox", append([]string{"list"}, args...))
+	defer func() {
+		telemetry.TrackCommandError(cmd.Context(), "sandbox", append([]string{"list"}, args...), commandErr)
```
aheritier: **[Suggestion #9]** `cobra.NoArgs` is set on `list`, so `args` is always empty here. The `append([]string{"list"}, args...)` pattern is harmless but confusing — `args` can be dropped from this telemetry call.

cmd/root/sandbox_cmd.go
```diff
@@ -0,0 +1,157 @@
...
+		Aliases: []string{"ls"},
+		Short:   "List the persistent sandbox network allowlist",
+		Args:    cobra.NoArgs,
+		RunE:    runSandboxListCommand,
+	}
+}
+
+func runSandboxAllowCommand(cmd *cobra.Command, args []string) (commandErr error) {
+	telemetry.TrackCommand(cmd.Context(), "sandbox", append([]string{"allow"}, args...))
+	defer func() {
```
aheritier, line 77: **[Suggestion #10]** User-supplied hostnames (e.g. `corp-proxy.internal.acme.com`) are forwarded into telemetry here. Sandbox hosts are more likely to be sensitive internal infrastructure names than agent refs. Please confirm this is intentional — if so, a brief comment would document the conscious decision.

cmd/root/run.go
```diff
@@ -228,6 +235,10 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		if alias.HideToolResults && !f.hideToolResults {
 			f.hideToolResults = true
 		}
+		// alias.Sandbox is consumed earlier in runRunCommand before
+		// dispatch; by the time we reach runOrExec the sandbox path
+		// has already been taken (or the user opted out via
```
aheritier: **[Suggestion #8]** Slightly misleading: the comment says *"the sandbox path has already been taken"* but since we are in `runOrExec` (the non-sandbox branch), the sandbox path was **not** taken — it resolved to false. Suggested wording: *"alias.Sandbox is consumed in runRunCommand; since we are here, the sandbox decision resolved to false (explicit --sandbox=false, or neither the alias nor the agent config requested it)."*

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26339342905).

aheritier (approved):

aheritier (approved):

PR #2889, opened 2026-05-24
feat(config): enable redact_secrets by default

rumpl:
Make the agent-level `redact_secrets` flag opt-out instead of opt-in.

- `pkg/config/latest`: field is now `*bool`; nil means enabled. New `RedactSecretsEnabled()` helper centralises the default.
- `pkg/teamloader`: uses the helper when wiring `agent.WithRedactSecrets`.
- `agent-schema.json`: `default: true` and updated description.
- Examples updated to call out the new default; `redact_secrets: false` opts out.

Existing configs that omit the field gain redaction automatically. Configs that explicitly set `true`/`false` are unchanged. Auto-injection still dedupes against manually-written hook entries, so the `redact_secrets_hooks.yaml` example keeps working.

Tests: new unit tests cover default/explicit-true/explicit-false and YAML round-trip; existing redact_secrets builtin + teamloader tests pass.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -564,7 +564,8 @@
         },
         "redact_secrets": {
           "type": "boolean",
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed 6 changed files (183 diff lines). The `bool → *bool` tri-state pattern for `RedactSecrets` is correctly implemented:

- `RedactSecretsEnabled()` correctly handles nil receiver and nil pointer (default-on).
- The only production call site (`teamloader.go:167`) already uses `RedactSecretsEnabled()` — no unupdated callers.
- `pkg/runtime/hooks.go` calls `a.RedactSecrets()` on `*Agent` (not `*AgentConfig`), which returns the already-resolved bool set by `teamloader` via `WithRedactSecrets(agentConfig.RedactSecretsEnabled())` — correct downstream path.
- `omitempty` on `*bool` correctly omits only nil (unset), not `false` (explicitly disabled).
- Test coverage in `redact_secrets_test.go` covers nil cfg, nil field, explicit true, and explicit false cases.

aheritier (approved):

PR #2892, opened 2026-05-26
feat(pkg/history): redact secrets in command history

dgageot:
Command history should never contain secrets. This change routes all history entries through `portcullis.Redact` before storing them, ensuring that API tokens and other credentials pasted at the prompt never reach the in-memory slice or the `~/.cagent/history` file.

The history package now redacts entries when adding them and when migrating from the legacy history file. A critical fix ensures that pre-existing secrets in the old `~/.cagent/history.json` format are redacted on migration, closing a gap where secrets could be copied verbatim into the new format on first launch.

The remaining commits consolidate test setup code and clean up redundant parsing logic, improving maintainability with no behavioral change.

### DIFF
--- a/pkg/history/history.go
+++ b/pkg/history/history.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+
+	"github.com/docker/portcullis"
 )
 
 // History is the in-memory view of a persistent message history. The cursor
@@ -43,8 +45,12 @@ func New(baseDir string) (*History, error) {
 }
 
 // Add records a new message. Any prior occurrence of the same message is
-// removed and the new one becomes the most recent entry.
+// removed and the new one becomes the most recent entry. The message is
+// scrubbed of secret material via [portcullis.Redact] before being stored
+// in memory or written to disk so secrets pasted into the prompt never
+// linger in the persistent history.
 func (h *History) Add(message string) error {
+	message = portcullis.Redact(message)
 	h.addInMemory(message)
 	h.current = len(h.Messages)
 	return h.append(message)
@@ -162,7 +168,10 @@ func (h *History) load() error {
 		if err := json.Unmarshal(line, &msg); err != nil {
 			continue
 		}
-		h.addInMemory(msg)
+		// Redact on read so a history file that predates redaction
+		// (or was tampered with externally) never exposes secrets to
+		// the in-memory navigation cursor or downstream callers.
+		h.addInMemory(portcullis.Redact(msg))
 	}
 	return nil
 }
@@ -188,7 +197,7 @@ func (h *History) migrateOldHistory(baseDir string) error {
 	}
 
 	for _, msg := range old.Messages {
-		if err := h.append(msg); err != nil {
+		if err := h.append(portcullis.Redact(msg)); err != nil {
 			return err
 		}
 	}
--- a/pkg/history/history_test.go
+++ b/pkg/history/history_test.go
@@ -5,8 +5,11 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/docker/portcullis"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/pkg/internal/portcullistest"
 )
 
 func TestNew(t *testing.T) {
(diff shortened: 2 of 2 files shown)


rumpl:
> The history package now redacts entries when adding them and when migrating from the legacy history file.

This is true, but people that are using the new history file don't get their history redacted.

pkg/history/history.go
```diff
@@ -154,12 +160,8 @@ func (h *History) load() error {
 	}
 
 	for line := range bytes.Lines(data) {
-		line = bytes.TrimSuffix(line, []byte("\n"))
-		if len(line) == 0 {
-			continue
-		}
 		var msg string
-		if err := json.Unmarshal(line, &msg); err != nil {
+		if json.Unmarshal(line, &msg) != nil {
```
docker-agent: **[MEDIUM] `load()` does not redact pre-existing entries from the new-format history file**

Entries are loaded from `~/.cagent/history` and added to memory without calling `portcullis.Redact`. Any secrets written to that file by an older version of the code (before this PR introduced `Add`-time redaction) will be loaded into memory unredacted on the next session.

The legacy `history.json` migration path correctly redacts, but the new-format file is never retroactively cleaned at load time. Consider applying redaction here too:

```go
h.addInMemory(portcullis.Redact(msg))
```

Without this, users upgrading from a pre-redaction version will continue to have their secrets in memory (and potentially re-persisted) for the lifetime of sessions that start with an old history file.

pkg/history/history.go
```diff
@@ -188,7 +190,7 @@ func (h *History) migrateOldHistory(baseDir string) error {
 	}
 
 	for _, msg := range old.Messages {
-		if err := h.append(msg); err != nil {
+		if err := h.append(portcullis.Redact(msg)); err != nil {
```
docker-agent, line 200: **[MEDIUM] Legacy `history.json` deleted before confirming new history file is readable**

`migrateOldHistory` calls `os.Remove(oldPath)` only after all `append` calls succeed. However, the deletion happens before `New()` calls `h.load()` to verify the new file can actually be read back. If the write file-descriptor succeeds but a subsequent `ReadFile` fails (e.g., a filesystem edge case, permissions race, or partial flush), the old file is already gone and the user's history is permanently lost.

A safer approach verifies readability before committing to deletion:

```go
// verify the new file is readable before deleting the legacy one
if _, err := os.ReadFile(h.path); err != nil {
    return fmt.Errorf("migration wrote entries but new file is unreadable: %w", err)
}
return os.Remove(oldPath)
```

Alternatively, rename to `.bak` first and delete only after a successful `load()`.

docker-agent (commented):

Two medium-severity issues found in the new redaction logic in `pkg/history/history.go`. Both are in code added or changed by this PR.

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | 🟡 MEDIUM | `history.go:164` | `load()` does not redact pre-existing entries from the new-format file |
| 2 | 🟡 MEDIUM | `history.go:193` | Legacy `history.json` deleted before confirming new history file is readable |

dgageot:
Thanks for the review! Rebased on `origin/main` and addressed both points:

- `load()` now passes every entry through `portcullis.Redact` so users upgrading from a pre-redaction build get their existing on-disk history scrubbed too. When any entry changes, the file is atomically rewritten (via temp file + rename) so the secrets are removed from disk and not just hidden in memory. Covered by `TestHistory_RedactsSecretsOnLoad`.
- `migrateOldHistory` now verifies the new file is readable before deleting `history.json`, so a partial-flush / permissions edge case can no longer destroy the user's history during the upgrade. Covered by `TestHistory_MigrationKeepsLegacyOnFailure`.

@rumpl that should also cover your point about users already on the new format — their secrets get cleaned up the next time they launch.

`task lint`, `task test`, and `task build` all green locally.

rumpl:
Nice but seems kinda off that we are redacting the history on each load, why not make a new migration

dgageot:
Good call — replaced the per-load redaction with a one-time migration:

- Dropped `fix(pkg/history): redact pre-existing secrets when loading`.
- Added `fix(pkg/history): add one-time migration to redact pre-existing history`. The migration scrubs `~/.cagent/history` once on next launch (atomic temp+rename), drops a sentinel file (`~/.cagent/history.redacted`) next to it, and short-circuits on every subsequent load — so the steady-state path stays a plain `ReadFile` with no extra cost.
- Tests added: `TestHistory_RedactExistingHistoryRunsOnce` proves the file is rewritten on first launch, the marker is created, and a tampered file written *after* the migration is left untouched (file is trusted once flagged). `TestHistory_RedactExistingHistoryNoFile` covers the cold-start path.

Rebased on `origin/main`. `task lint`, `task test`, `task build` all green locally.

rumpl (dismissed):

rumpl (approved):

PR #2893, opened 2026-05-26
Revert "fix: spool large mcp media to disk"

dgageot:
This reverts commit 6dd026249c0dc0788ec53cb56972ff46127986d5.

### DIFF
--- a/pkg/runtime/toolexec/dispatcher.go
+++ b/pkg/runtime/toolexec/dispatcher.go
@@ -724,21 +724,13 @@ func buildMultiContent(text string, images []tools.MediaContent) []chat.MessageP
 	parts := make([]chat.MessagePart, 0, 1+len(images))
 	parts = append(parts, chat.MessagePart{Type: chat.MessagePartTypeText, Text: text})
 	for _, img := range images {
-		switch {
-		case img.FilePath != "":
-			parts = append(parts, chat.MessagePart{
-				Type: chat.MessagePartTypeText,
-				Text: fmt.Sprintf("[image saved to %s (%s)]", img.FilePath, img.MimeType),
-			})
-		case img.Data != "":
-			parts = append(parts, chat.MessagePart{
-				Type: chat.MessagePartTypeImageURL,
-				ImageURL: &chat.MessageImageURL{
-					URL:    "data:" + img.MimeType + ";base64," + img.Data,
-					Detail: chat.ImageURLDetailAuto,
-				},
-			})
-		}
+		parts = append(parts, chat.MessagePart{
+			Type: chat.MessagePartTypeImageURL,
+			ImageURL: &chat.MessageImageURL{
+				URL:    "data:" + img.MimeType + ";base64," + img.Data,
+				Detail: chat.ImageURLDetailAuto,
+			},
+		})
 	}
 	return parts
 }
--- a/pkg/tools/mcp/mcp.go
+++ b/pkg/tools/mcp/mcp.go
@@ -160,11 +160,6 @@ type Toolset struct {
 
 	supervisor *lifecycle.Supervisor
 
-	// mediaDir is the toolset-scoped temp dir holding spooled media
-	// payloads. Created lazily on first spool, removed by Stop.
-	mediaMu  sync.Mutex
-	mediaDir string
-
 	mu sync.Mutex
 
 	// Cached tools and prompts, invalidated via MCP notifications and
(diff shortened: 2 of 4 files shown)


rumpl (approved):

docker-agent (commented):

This PR reverts the MCP media spool-to-disk feature. The revert is clean and intentional, but re-introduces two behavioural regressions worth noting.

pkg/tools/mcp/mcp.go
```diff
@@ -772,94 +760,12 @@ func (ts *Toolset) processMCPContent(toolResult *mcp.CallToolResult) *tools.Tool
...
-		return ".webp"
-	case "audio/wav", "audio/wave", "audio/x-wav":
-		return ".wav"
-	case "audio/mpeg", "audio/mp3":
-		return ".mp3"
-	default:
-		return ".bin"
+// encodeMedia re-encodes raw bytes (as decoded by the MCP SDK) back to base64
+// for our internal MediaContent representation.
+func encodeMedia(data []byte, mimeType string) tools.MediaContent {
```
docker-agent, line 765: **[MEDIUM] Memory pressure regression: large MCP media always held in-memory as base64**

The reverted `encodeMedia` unconditionally base64-encodes all MCP media payloads — no matter how large — directly into a `string` field held in memory for the lifetime of the session. The spool-to-disk fix (commit `6dd0262`) was introduced specifically to avoid this: it kept payloads ≤ 256 KB inline and wrote larger ones to a temp file.

With this revert, a single MCP tool response containing, e.g., a 10 MB screenshot will allocate ~13 MB of base64 in RAM, duplicated across the session history and the TUI. Multiple such responses compound quickly.

If the spool approach is being dropped intentionally, consider at minimum documenting the decision or adding a size cap/warning log for unexpectedly large payloads.

pkg/runtime/toolexec/dispatcher.go
```diff
@@ -724,21 +724,13 @@ func buildMultiContent(text string, images []tools.MediaContent) []chat.MessageP
...
-		case img.Data != "":
-			parts = append(parts, chat.MessagePart{
-				Type: chat.MessagePartTypeImageURL,
-				ImageURL: &chat.MessageImageURL{
-					URL:    "data:" + img.MimeType + ";base64," + img.Data,
-					Detail: chat.ImageURLDetailAuto,
-				},
-			})
-		}
+		parts = append(parts, chat.MessagePart{
```
docker-agent, line 727: **[MEDIUM] Missing empty-data guard may produce malformed data URIs**

The previous `buildMultiContent` had a `case img.Data != "":` guard that silently skipped any `MediaContent` with an empty `Data` field. The restored code unconditionally builds a `MessagePart` for every image:

```go
URL: "data:" + img.MimeType + ";base64," + img.Data,
```

If `img.Data` is empty (e.g., an MCP server returns an `ImageContent` with a zero-length `Data` slice — `base64.StdEncoding.EncodeToString(nil)` returns `""`), the resulting URL is `"data:image/png;base64,"` — an invalid data URI that will be sent verbatim to the LLM API. Most LLM clients will return an error or silently fail on such a URL.

Since `encodeMedia` always calls `base64.StdEncoding.EncodeToString`, this can only trigger when an MCP server sends an empty `ImageContent.Data`, but the old guard handled that defensively. Consider adding back a check:
```go
if img.Data != "" {
    parts = append(parts, ...)
}
```

pkg/tools/tools.go
```diff
@@ -73,11 +73,8 @@ type FunctionCall struct {
 // MediaContent represents base64-encoded binary data (image, audio, etc.)
 // returned by a tool.
 type MediaContent struct {
-	// Data is the base64-encoded payload. It is kept only for small media; large
-	// MCP payloads are spooled to FilePath to avoid retaining duplicate base64.
-	Data string `json:"data,omitempty"`
-	// FilePath is an optional local file containing the decoded media payload.
-	FilePath string `json:"filePath,omitempty"`
+	// Data is the base64-encoded payload.
+	Data string `json:"data"`
```
docker-agent, line 77: **[LOW] Removing `omitempty` from `MediaContent.Data` changes JSON serialisation**

The field tag changed from `json:"data,omitempty"` to `json:"data"`. When `Data` is an empty string, the old tag would omit the field entirely from JSON output; the new tag always emits `"data":""`.

In the current code path `encodeMedia` always populates `Data`, so this is unlikely to matter in practice. However, any code that constructs a `MediaContent` with an empty `Data` (e.g. tests, future zero-value defaults) will now emit the extra field, which may surprise consumers or break strict-equality JSON assertions.

PR #2899, opened 2026-05-26
refactor(tools): use github.com/docker/aijson for tool-arg shape repair

trungutt:
## What

Replace the in-tree tool-argument shape-repair layer (`pkg/tools/repair.go`, ~270 LOC) with the upstream [`github.com/docker/aijson`](https://github.com/docker/aijson) package, which is itself an extraction of that same code.

## Why

The four shape repairs (`unwrap_string_array`, `wrap_in_array`, `wrap_object_in_array`, `drop_null`) were lifted into a standalone, reusable package per its README. Adopting the upstream copy means:

- **One source of truth.** Future fixes / new repairs land in one place.
- **Other call sites can adopt it directly.** Several handlers in this repo currently bypass `tools.NewHandler` and roll their own `json.Unmarshal` (e.g. `pkg/runtime/skill_runner.go`, `pkg/runtime/model_picker.go`, `pkg/acp/filesystem.go`, background-agent handlers in `pkg/tools/builtin/agent/agent.go`). They'll be migrated in follow-up PRs without re-importing tools-internal symbols.
- **`edit_file`'s double-serialized `edits` branch is subsumed by `unwrap_string_array`.** A separate follow-up PR will simplify `ParseEditFileArgs` accordingly (the byte-level syntax-repair layer for malformed JSON stays — aijson explicitly doesn't cover that).


rumpl (approved):

pkg/tools/repair_test.go
```diff
@@ -2,13 +2,18 @@ package tools
```
rumpl, line 2: Why do we still need this?

pkg/tools/repair_test.go
```diff
@@ -113,71 +102,27 @@ func TestRepair_VisitsPromotedFieldsFromEmbeddedStruct(t *testing.T) {
...
-	assert.Equal(t, "ok", result.Output)
-	assert.Equal(t, []string{"a.txt", "b.txt"}, got.Paths)
+func TestNewHandler_RefusesMultiKeyObjectAsArray(t *testing.T) {
+	// Two keys in the placeholder object — too ambiguous to safely wrap.
+	// Surface the original schema error rather than guessing.
+	_, err := runHandler[argsWithStrings](t, `{"paths":{"path":"a.txt","extra":"ignore"}}`)
+	require.Error(t, err)
 }
 
 func TestNewHandler_UnrepairableInputReturnsOriginalError(t *testing.T) {
```
docker-agent, line 128: **[MINOR] Missing regression tests for non-object top-level input and unknown-field passthrough**

Two defensive tests from the old `repair_test.go` were not ported as integration-level tests:

- `TestRepair_LeavesUnknownFieldsAlone` — verifies that unknown fields in the JSON payload are not mutated by the repair pass.
- `TestRepair_ReturnsFalseOnNonObjectInput` — verifies that a top-level non-object payload (e.g. `"just a string"` or `42`) does not panic or silently corrupt the result.

The non-object case is particularly relevant because LLMs occasionally send bare JSON strings as tool arguments. Without a regression test, if `aijson.Unmarshal` behaves differently on these inputs (panics, silently drops the payload, etc.), it won't be caught.

Consider adding integration-style tests via `NewHandler`/`runHandler` that cover both cases.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean extraction refactor — the in-tree `repair.go` (~270 LOC) is deleted and replaced with a single call to `aijson.Unmarshal` from `github.com/docker/aijson`. The `tools.go` change is minimal and correct: the import swap is clean, the `OnRepair` callback preserves telemetry, and error propagation is maintained.

The test suite is sensibly migrated from white-box unit tests on `tryRepairToolArgs` to black-box integration tests through `NewHandler`/`runHandler`, covering all four repair kinds plus the multi-field, embedded-struct, and field-ordering cases.

One minor observation about test coverage is noted inline.

PR #2900, opened 2026-05-26
feat(tui): word-level highlighting in edit_file diff view

rumpl:
Inside a modified line, highlight only the specific tokens that changed instead of painting the whole line. Paired delete/insert lines are run through a word-aware LCS diff; differing segments are rendered with a stronger, more saturated background tint derived from the active theme's DiffAddBg / DiffRemoveBg colors.

Screenshot 2026-05-26 at 22 51 13

### DIFF
--- a/pkg/tui/components/tool/editfile/render.go
+++ b/pkg/tui/components/tool/editfile/render.go
@@ -71,6 +71,10 @@ func InvalidateCaches() {
 type chromaToken struct {
 	Text  string
 	Style lipgloss.Style
+	// Emphasized marks tokens whose underlying text was identified as part
+	// of a word-level diff. Renderers paint these with a stronger background
+	// so users can locate the precise edit within a long line.
+	Emphasized bool
 }
 
 type linePair struct {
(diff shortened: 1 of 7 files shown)


pkg/tui/components/tool/editfile/wordiff.go
```diff
@@ -0,0 +1,155 @@
...
+	diffs := lcs.DiffLines(oldTokens, newTokens)
+
+	// Walk diffs in order to interleave equal regions with changed regions.
+	oldPos, newPos := 0, 0
+	for _, d := range diffs {
+		if d.Start > oldPos {
+			eq := strings.Join(oldTokens[oldPos:d.Start], "")
+			if eq != "" {
+				oldSegs = append(oldSegs, wordSegment{Text: eq, Changed: false})
+				newSegs = append(newSegs, wordSegment{Text: eq, Changed: false})
```
docker-agent, line 115: **[MEDIUM] `diffWords` equal-region uses old-side tokens for both sides — new-side equal content may be lost**

In `diffWords`, when the LCS diff has a gap between consecutive entries (an equal region), the code uses `oldTokens[oldPos:d.Start]` as the equal text and appends it to **both** `oldSegs` and `newSegs`:

```go
if d.Start > oldPos {
    eq := strings.Join(oldTokens[oldPos:d.Start], "")
    if eq != "" {
        oldSegs = append(oldSegs, wordSegment{Text: eq, Changed: false})
        newSegs = append(newSegs, wordSegment{Text: eq, Changed: false})  // old-side text used for new side!
    }
}
```

This is only safe because the LCS invariant guarantees that equal regions have identical tokens on both sides (`oldTokens[oldPos:d.Start]` == `newTokens[newPos:d.ReplStart]`). However, the guard `d.Start > oldPos` does **not** have a corresponding `d.ReplStart > newPos` check. If the `lcs.DiffLines` library returns any edge case where the old-side gap is empty but the new-side gap is non-empty (e.g., degenerate or optimized output), the new-side equal tokens would be silently dropped. The consequence is that `concat(newSegs) != newLine`, causing misaligned byte offsets in `applyWordEmphasis` and corrupted emphasis rendering.

**Recommendation**: Use `newTokens[newPos:d.ReplStart]` for `newSegs` and guard with `d.ReplStart > newPos`, making the intent explicit and robust against any library behaviour:
```go
if d.Start > oldPos {
    eq := strings.Join(oldTokens[oldPos:d.Start], "")
    if eq != "" {
        oldSegs = append(oldSegs, wordSegment{Text: eq, Changed: false})
    }
}
if d.ReplStart > newPos {
    eq := strings.Join(newTokens[newPos:d.ReplStart], "")
    if eq != "" {
        newSegs = append(newSegs, wordSegment{Text: eq, Changed: false})
    }
}
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR adds word-level diff highlighting using an LCS-based tokenizer, a new `applyWordEmphasis` function, and a `deriveEmphasisBg` colour transform. The approach is solid, but a **high-severity byte-offset mismatch** was found in the `prepareContent` → `diffWords` pipeline that affects all tab-indented files (i.e., essentially all Go code in this project).

**Findings summary:**
- 🔴 **HIGH** — Tab expansion in `prepareContent` creates byte-offset divergence between word-diff segments and chroma tokens; wrong emphasis and potential panic on tab-indented lines (`render.go:626`)
- 🟡 **MEDIUM** — `diffWords` equal-region guard only checks old-side position; new-side equal tokens may be silently dropped if LCS output is asymmetric (`wordiff.go:107`)
- 🟢 **LOW** — `deriveEmphasisBg` lightness cap at `0.40` can produce a darker emphasis for dark-theme backgrounds with `l ∈ [0.40, 0.50)` (`colorutil.go:450`)

pkg/tui/components/tool/editfile/render.go
```diff
@@ -473,14 +609,20 @@ func wrapTokens(tokens []chromaToken, maxWidth int) [][]chromaToken {
...
 	if line == nil {
 		return []string{renderEmptySplitSide(width)}
 	}
 
 	content := prepareContent(line.Content)
 	tokens := syntaxHighlight(content, filePath)
+	if len(wordSegs) > 0 {
+		tokens = applyWordEmphasis(tokens, wordSegs)
+	}
 	lineStyle := getLineStyle(line.Kind)
```
docker-agent, line 631: **[HIGH] Tab expansion in `prepareContent` causes byte-offset mismatch with word-diff segments — wrong emphasis and potential panic on tab-indented lines**

`prepareContent()` expands each `\t` to 4 spaces before passing content to `syntaxHighlight`. The chroma tokens produced therefore have byte offsets relative to the **tab-expanded** string. However, `diffWords` is called with the **raw** (non-tab-expanded) content:

```go
// renderSplitSide — uses tab-expanded content for tokens:
content := prepareContent(line.Content)    // \t → "    " (4 bytes)
tokens := syntaxHighlight(content, filePath)

if len(wordSegs) > 0 {
    tokens = applyWordEmphasis(tokens, wordSegs)  // segments have unexpanded byte offsets!
}

// Word segs built with unexpanded content:
oldSegs, newSegs = diffWords(
    strings.TrimRight(pair.old.Content, "\r\n"),  // \t is still 1 byte here
    strings.TrimRight(pair.new.Content, "\r\n"),
)
```

`applyWordEmphasis` maps segment byte ranges onto chroma token byte ranges. For any line containing a `\t` before a changed word, the segment byte positions are shifted left by `(tabWidth-1) * numTabsBefore` relative to the chroma token byte positions. This causes:

1. **Wrong emphasis**: Changed tokens are highlighted at the wrong positions (or not at all).
2. **Potential panic**: The slice `text[cursor-tokStart : end-tokStart]` can produce a negative or out-of-bounds index if `end` (derived from the segment range) exceeds the actual chroma token's byte length.

Since Go source files — including the test fixtures in this PR — are almost always tab-indented, this affects every real-world usage of the word-level emphasis.

**Fix**: Call `applyWordEmphasis` using segments derived from `prepareContent`-processed input (i.e., pass `prepareContent(...)` to `diffWords` as well), or expand tabs in the content passed to `diffWords` before computing segments.

pkg/tui/styles/colorutil.go
```diff
@@ -420,3 +420,38 @@ func labF(t float64) float64 {
...
+	// typically very desaturated (~0.1-0.25); we want emphasis to read as
+	// a distinct color block.
+	s = max(s, 0.55)
+
+	// Shift lightness away from the line background so the emphasis block
+	// is visible. For dark backgrounds (l < 0.5) we lift; for light ones we
+	// drop. The magnitude is intentionally modest so changed tokens remain
+	// readable against the surrounding row color.
+	if l < 0.5 {
+		l = min(l+0.18, 0.40)
```
docker-agent: **[LOW] `deriveEmphasisBg` lightness cap can make emphasis darker than the row background for certain dark themes**

For dark-theme diff backgrounds where lightness is in the range `[0.40, 0.50)`, the formula `l = min(l+0.18, 0.40)` clamps the output down to `0.40`, which is **less than** the input. For example:

- Input `l = 0.45` → `min(0.63, 0.40) = 0.40` (emphasis is darker than background)
- Input `l = 0.48` → `min(0.66, 0.40) = 0.40` (emphasis is darker than background)

An emphasis background darker than the surrounding row background defeats the purpose: changed tokens become harder to see rather than easier.

```go
if l < 0.5 {
    l = min(l+0.18, 0.40)  // cap at 0.40 can REDUCE lightness for inputs in [0.40, 0.50)
} else {
    l = max(l-0.18, 0.65)
}
```

Note: Typical dark-theme diff backgrounds tend to have very low lightness (`l ≈ 0.10–0.25`), so this range may not be hit with default themes. But any custom theme or third-party theme with a medium-dark diff background would trigger this.

**Recommendation**: Change the cap to `0.50` (or remove it entirely and just clamp to `[0, 1]`) so the emphasis colour is always at least as light as its source:
```go
if l < 0.5 {
    l = min(l+0.18, 0.50)
}
```

rumpl:
Addressed all three review findings in 39dc76a:

- **HIGH (tab offsets):** `buildLineWordDiffs` and the split-view inline call now feed `prepareContent`-processed text into `diffWords`, so segment byte offsets align with the chroma tokens. Added `TestRenderEditFile_TabIndentedLineDoesNotPanic` as a regression guard.
- **MEDIUM (asymmetric LCS gaps):** Old- and new-side equal regions are now handled with independent guards and each side uses its own tokens. Added `TestDiffWords_SegmentsReconstructInputs` to ensure `concat(segs) == input` for several edge cases.
- **LOW (lightness cap):** Cap raised from 0.40 to 0.50 so the emphasis is never darker than its source.

dgageot (approved):

PR #2901, opened 2026-05-26
Smarter search

rumpl:
In the command palette typing "the" will now put "Theme" first on the list, and typing "ses" will put "Sessions" first on the list

### DIFF
--- a/pkg/tui/dialog/command_palette.go
+++ b/pkg/tui/dialog/command_palette.go
@@ -1,6 +1,7 @@
 package dialog
 
 import (
+	"slices"
 	"strings"
 
 	"charm.land/bubbles/v2/key"
@@ -124,10 +125,15 @@ func (d *commandPaletteDialog) filterCommands() {
 	d.filtered = d.filtered[:0]
 	for _, cat := range d.categories {
 		for _, cmd := range cat.Commands {
-			if query == "" || matchesCommandQuery(cmd, query) {
+			if query == "" || commandQueryScore(cmd, query) < commandQueryNoMatch {
 				d.filtered = append(d.filtered, cmd)
 			}
 		}
+	}
+	if query != "" {
+		slices.SortStableFunc(d.filtered, func(a, b commands.Item) int {
+			return commandQueryScore(a, query) - commandQueryScore(b, query)
+		})
 	}
 
 	// Clearing the search returns the cursor to the top, matching the file
@@ -138,6 +144,8 @@ func (d *commandPaletteDialog) filterCommands() {
 	d.scrollview.SetScrollOffset(0)
 }
 
+const commandQueryNoMatch = 1 << 30
+
 // matchesCommandQuery reports whether the given command matches the lowercase
 // query string by searching label, description, or slash command. The
 // category is intentionally excluded: category names act as section headers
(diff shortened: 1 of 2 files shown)


pkg/tui/dialog/command_palette_test.go
```diff
@@ -86,6 +86,36 @@ func TestCommandPaletteFilteringIgnoresCategory(t *testing.T) {
...
+	d := dialog.(*commandPaletteDialog)
+
+	d.textInput.SetValue("the")
+	d.filterCommands()
+
+	var ids []string
+	for _, c := range d.filtered {
+		ids = append(ids, c.ID)
+	}
+	require.Equal(t, []string{"settings.theme", "session.attach"}, ids,
```
docker-agent, line 115: **[HIGH] Test assertion incorrectly expects `session.attach` to appear for query `"the"`**

The test `TestCommandPaletteFilteringRanksLabelPrefixFirst` queries for `"the"` and asserts `[]string{"settings.theme", "session.attach"}`, but `session.attach` has:
- Label: `"Attach"` → lowercase `"attach"` — does **not** contain `"the"`
- SlashCommand: `"/attach"` — does **not** contain `"the"`
- Description: `"Attach a file to your message"` — does **not** contain `"the"`

All four `commandFieldQueryScore` calls return `commandQueryNoMatch`, so `session.attach` is filtered out entirely. The actual filtered result would be `["settings.theme"]` only, causing `require.Equal` to fail.

The intent of the test appears to be verifying that a **label prefix** match ranks ahead of a **substring** match. A suitable second command would be one whose description or label contains `"the"` as a mid-word substring (e.g., a command with description containing "the") to demonstrate that `settings.theme` (label prefix) ranks first.

pkg/tui/dialog/command_palette.go
```diff
@@ -138,16 +144,57 @@ func (d *commandPaletteDialog) filterCommands() {
...
+		return base + 100 + index
+	}
+	return base + 200 + index
+}
+
+func isCommandQueryWordStart(value string, index int) bool {
+	if index == 0 {
+		return true
+	}
+	previous := value[index-1]
```
docker-agent: **[LOW] `isCommandQueryWordStart` uses byte indexing; non-ASCII characters before the match position are not recognized as word boundaries**

`value[index-1]` reads one byte before the match (where `index` is a byte offset from `strings.Index`). If the preceding character is a multi-byte UTF-8 sequence (e.g., `é`, `ñ`, or CJK), `index-1` lands inside that sequence and reads a continuation byte (0x80–0xBF), which never equals any ASCII word-separator. This means a non-ASCII character immediately before the query match is never classified as a word boundary, slightly degrading ranking for non-ASCII content.

For the current use case (ASCII command names and descriptions) this has no practical impact. Consider using `utf8.DecodeLastRuneInString(value[:index])` if internationalized command labels are ever needed.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

The new test `TestCommandPaletteFilteringRanksLabelPrefixFirst` has a broken assertion — it expects `session.attach` to appear in filtered results for query `"the"`, but none of that command's fields (label, slash command, or description) contain the substring `"the"`. The test will fail as written.

dgageot (approved):

PR #2902, opened 2026-05-26
persist cookies in remote MCP client for sticky sessions

maxcleme:
## Persist session cookies for remote MCP clients

This PR fixes a connectivity issue that surfaces when cagent talks to a remote MCP server fronted by a replicated.

### Symptom

After a successful `initialize`, the **next** request (`notifications/initialized` / first tool call) intermittently failed
because it landed on a different ingress replica that had no session state.

### Root cause

The ingress provides session stickiness by setting a cookie on the first response. The `*http.Client` built in
`pkg/tools/mcp/remote.go::createHTTPClient` had no `http.CookieJar`, so the cookie was dropped and subsequent requests were free
to be routed to any replica.

### Change

- **`pkg/tools/mcp/remote.go`** — `createHTTPClient` now creates a per-client `cookiejar.New(nil)` and assigns it to
`http.Client.Jar`. The jar is naturally scoped per remote MCP server (one client per `remoteMCPClient` per toolset), so cookies
cannot leak across toolsets. Servers that do not use cookies are unaffected.

### Tests

- **`pkg/tools/mcp/remote_test.go`** — new `TestCreateHTTPClient_PersistsCookies` using `httptest.Server`: the first request
must arrive without `mcp_session` and gets a `Set-Cookie: mcp_session=abc123`; subsequent requests must echo the cookie back.

### DIFF
--- a/pkg/tools/mcp/remote.go
+++ b/pkg/tools/mcp/remote.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"net/http/cookiejar"
 
 	gomcp "github.com/modelcontextprotocol/go-sdk/mcp"
 
@@ -57,7 +58,10 @@ func (c *remoteMCPClient) Initialize(ctx context.Context, _ *gomcp.InitializeReq
 	// oauthTransport so we can enrich Connect errors with the server's own
 	// explanation — without this, a plain `Bad Request` bubbles up and the
 	// user has no idea that, say, the Slack app hasn't been enabled for MCP.
-	httpClient, oauthT := c.createHTTPClient()
+	httpClient, oauthT, err := c.createHTTPClient()
+	if err != nil {
+		return nil, fmt.Errorf("creating HTTP client: %w", err)
+	}
 
 	var transport gomcp.Transport
 
@@ -146,7 +150,7 @@ func (c *remoteMCPClient) SetManagedOAuth(managed bool) {
 // The oauthTransport is returned alongside the client so callers can inspect
 // the most recent server-side failure (via lastServerError) when Connect()
 // returns a bare HTTP-status error and we need to surface the actual cause.
-func (c *remoteMCPClient) createHTTPClient() (*http.Client, *oauthTransport) {
+func (c *remoteMCPClient) createHTTPClient() (*http.Client, *oauthTransport, error) {
 	base := c.headerTransport()
 
 	// Then wrap with OAuth support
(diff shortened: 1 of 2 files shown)


pkg/tools/mcp/remote.go
```diff
@@ -160,7 +161,13 @@ func (c *remoteMCPClient) createHTTPClient() (*http.Client, *oauthTransport) {
 		oauthHTTPClient: oauthHTTPClientForAllowPrivateIPs(c.allowPrivateIPs),
 	}
 
-	return &http.Client{Transport: oauthT}, oauthT
+	// Persist cookies across requests
+	// So sticky sessions work if implemented by the server (e.g. in a multiple replica setup)
+	jar, err := cookiejar.New(nil)
```
docker-agent, line 169: **[LOW] Silent cookie-persistence failure when `cookiejar.New` returns an error**

`cookiejar.New(nil)` always returns a nil error in current stdlib (the error is reserved for future use), so this path is unreachable today. However, if a future Go version were to return an error, the code logs a warning but continues by assigning `jar = nil` into `http.Client.Jar`. A nil `Jar` is legal (`http.Client` simply skips persistence), so there's no crash — but the caller silently gets a client without cookie persistence, contrary to the function's stated purpose.

Consider returning the error to the caller so sticky-session failures are surfaced explicitly:
```go
jar, err := cookiejar.New(nil)
if err != nil {
    return nil, nil, fmt.Errorf("failed to create cookie jar: %w", err)
}
```
Or, at minimum, the existing warning is an acceptable defensive fallback given the current stdlib guarantee.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly implements cookie jar persistence for the remote MCP HTTP client, enabling sticky sessions. The implementation is clean and the new test coverage is solid.

**Summary of findings:**
- 1 low-severity / informational note (see inline comment)

pkg/tools/mcp/remote.go
```diff
@@ -160,7 +164,13 @@ func (c *remoteMCPClient) createHTTPClient() (*http.Client, *oauthTransport) {
...
```
docker-agent, line 169: **[MEDIUM] Sticky-session cookie jar is ephemeral: cookies are lost on reconnection**

`createHTTPClient()` creates a fresh `cookiejar` on every call, and `Initialize()` calls `createHTTPClient()` each time it runs. This means cookies set by the server during one session are discarded when `Initialize` is called again (e.g. on a disconnect/reconnect).

Within a single session (first request → subsequent requests on the same `http.Client`) cookies are correctly forwarded, which is the scenario covered by the new test. However if the MCP client reconnects to the same replicated backend, the new client presents an empty jar and the ingress may route it to a different replica — the exact failure mode this PR aims to fix.

To persist cookies across reconnections, the jar should be allocated once (e.g. lazily on the `remoteMCPClient` struct, or in `newRemoteClient`) and reused across `Initialize` calls rather than recreated each time.

docker-agent (commented):

One **medium** finding in the changed code: the new cookie jar is created inside `createHTTPClient()`, which is called on every `Initialize()` invocation. Cookies are correctly persisted within a single session (the PR's stated fix), but they are discarded on reconnection, so the sticky-session guarantee does not survive a reconnect cycle.

rumpl (approved):

dgageot (approved):

PR #2905, opened 2026-05-27
test(tools): trim aijson re-tests, keep docker-agent integration

trungutt:
Follow-up to #2899 addressing @rumpl's review comment on `pkg/tools/repair_test.go` ("Why do we still need this?").

## Why

After #2899 swapped the in-tree shape-repair layer for [`github.com/docker/aijson`](https://github.com/docker/aijson), most of `pkg/tools/repair_test.go` became a thin re-test of aijson's own test suite through the `NewHandler` facade. Every `TestNewHandler_*` case in that file had a direct `TestUnmarshal_*` counterpart in aijson.
Keeping both copies means every aijson behavior change has to be mirrored in two places, and the docker-agent suite slowly turns into a brittle snapshot of an external library.

Those three concerns live in `pkg/tools/tools_test.go` now alongside the other `NewHandler` tests; the standalone `repair_test.go` is removed.


pkg/tools/tools_test.go
```diff
@@ -79,6 +81,90 @@ func TestNewHandler_InvalidArguments(t *testing.T) {
...
+			Name:      "read_multiple_files",
+			Arguments: `{"paths":"only.txt"}`,
+		},
+	})
+	require.NoError(t, err)
+
+	out := buf.String()
+	assert.Contains(t, out, "tool_input_repaired")
+	assert.Contains(t, out, "tool=read_multiple_files")
+	assert.Contains(t, out, "wrap_in_array")
```
docker-agent: **[LOW] Assertion couples test to undocumented `aijson.Kind` string serialization**

`assert.Contains(t, out, "wrap_in_array")` relies on `aijson.Kind` producing exactly the string `"wrap_in_array"` when rendered by slog's text handler. That serialization format (snake_case, PascalCase, integer…) is an internal detail of the aijson package and is not part of its documented public API contract.

If a future aijson release changes `Kind.String()` — e.g., to `"WrapInArray"` or `"wrap-in-array"` — this test will fail even though `NewHandler` is working correctly, turning it into a false negative rather than a meaningful regression gate.

Consider either:
- Importing the constant directly (`aijson.WrapInArray.String()`) so the assertion tracks the library's own representation, or
- Asserting only on the presence of `"repairs"` in the log output (the field name, which is set by production code and therefore stable), and verifying the repair _occurred_ via the returned value rather than the log text.

pkg/tools/tools_test.go
```diff
@@ -79,6 +81,90 @@ func TestNewHandler_InvalidArguments(t *testing.T) {
...
+		},
+	})
+	require.NoError(t, err)
+	assert.Equal(t, []string{"only.txt"}, got.Paths)
+}
+
+func TestNewHandler_EmitsRepairTelemetry(t *testing.T) {
+	var buf bytes.Buffer
+	prev := slog.Default()
+	slog.SetDefault(slog.New(slog.NewTextHandler(&buf, &slog.HandlerOptions{Level: slog.LevelInfo})))
```
docker-agent, line 117: **[LOW] `slog.SetDefault` mutates global logger — latent data race with future parallel tests**

Both `TestNewHandler_EmitsRepairTelemetry` (line 116) and `TestNewHandler_NoTelemetryOnValidInput` (line 147) replace the process-wide default logger via `slog.SetDefault`. Neither test calls `t.Parallel()`, so there is no current race. The `t.Cleanup` restore is correct.

However, any future test in this package that does call `t.Parallel()` will race against these two tests on the global logger pointer, producing either a data race detected by the race detector or silently wrong buffer contents.

The idiomatic fix is to thread a `*slog.Logger` through `NewHandler` (or use `slog.New` with a test-local handler and pass it via `context.WithValue`/`slog.NewLogLogger`) instead of mutating the global default. Alternatively, the tests could be guarded with a package-level mutex, though that is more fragile.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The three stated contracts (aijson wiring, `tool_input_repaired` telemetry, hot-path quiet) are all covered by the new tests and the removal of the duplicate aijson re-tests is well-motivated. Two low-severity observations are left as inline comments.

dgageot (approved):

PR #2911, opened 2026-05-28
fix(runtime): populate ModelID in after_llm_call hook payload

kimizuka:
## Problem

Hook handlers wired to `after_llm_call` read `model_id` and always observed
the empty string, even though `pkg/hooks/types.go:177-186` documents `ModelID`
as populated for this event:

```go
// ModelID identifies the model the runtime is about to call (for
// before_llm_call) or just called (for after_llm_call), in the
// canonical "<provider>/<model>" form...
ModelID string `json:"model_id,omitempty"`
```

Anyone writing a cost/audit/telemetry sidecar that trusts the doc and keys off
`model_id` ends up with no way to distinguish responses across models in the
same session.

## Root cause

`executeAfterLLMCallHooks` (`pkg/runtime/hooks.go:446-453`) never set the field
on the dispatched `hooks.Input`:

```go
func (r *LocalRuntime) executeAfterLLMCallHooks(
    ctx context.Context, sess *session.Session,
    a *agent.Agent, responseContent string,
) {
    r.dispatchHook(ctx, a, hooks.EventAfterLLMCall, &hooks.Input{
        SessionID:       sess.ID,
        AgentName:       a.Name(),
        StopResponse:    responseContent,
        LastUserMessage: sess.GetLastUserMessageContent(),
        // ModelID was missing — `json:"model_id,omitempty"` then elided it
    }, nil)
}
```

`json:"model_id,omitempty"` on the zero-string then elides the key from the
wire payload entirely, so handlers can't even branch on "is the field present".

The sibling `executeBeforeLLMCallHooks` already populates `ModelID` from a
parameter (`pkg/runtime/loop.go:530`), so this is an after-only impl gap, not a
runtime-wide one.

## Fix

Thread the model identifier into the dispatch so the payload matches the
documented contract. Both callsites already have the identifier in scope:

- `pkg/runtime/hooks.go`: add `modelID` parameter to `executeAfterLLMCallHooks`
  and set `Input.ModelID` on the dispatch.
- `pkg/runtime/harness.go:192`: pass the existing local `modelID`.
- `pkg/runtime/loop.go:572`: pass `modelID.String()` from the resolved
  `modelsdev.ID`.

No public API or wire-format addition — the field has been on `hooks.Input` and
documented all along; this PR just stops eliding it.

For harness agents, `modelID` is the harness label (e.g. `claude-code`) rather
than the `<provider>/<model>` canonical form the types.go doc describes. That's
pre-existing: the sibling `executeBeforeLLMCallHooks` in the harness path is
dispatched with the same `modelID` (`pkg/runtime/harness.go:50`). This PR
preserves the before/after symmetry and introduces no new asymmetry.

## Tests

`pkg/runtime/after_llm_call_test.go` adds `TestAfterLLMCallHook_PopulatesModelID`,
which registers a builtin `after_llm_call` hook against a fixed-id `mockProvider`,
runs one turn, and asserts the captured `Input.ModelID` equals the provider's
canonical id. Verified the test fails on the unfixed tree
(`expected "test/mock-model" / actual ""`) and passes with the fix.

## Verification

- `go vet ./pkg/runtime/` — clean
- `golangci-lint run ./pkg/runtime/...` — 0 issues
- `go test ./pkg/runtime/ ./pkg/hooks/...` — all green

## Out of scope

`docs/configuration/hooks/index.md:258-259` lists per-event payload fields but
doesn't mention `model_id` for either `before_llm_call` (already populated in
code) or `after_llm_call` (this PR). That's a pre-existing docs gap covering
both events; a follow-up docs PR (matching the spirit of #2569) is a better
venue for catching them up together than fixing only the after-side here.

### DIFF
--- a/pkg/runtime/after_llm_call_test.go
+++ b/pkg/runtime/after_llm_call_test.go
@@ -1,0 +1,76 @@
+package runtime
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/pkg/agent"
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/hooks"
+	"github.com/docker/docker-agent/pkg/session"
+	"github.com/docker/docker-agent/pkg/team"
+)
+
+// TestAfterLLMCallHook_PopulatesModelID is a regression test for the
+// doc/impl mismatch where [hooks.Input.ModelID] is documented as
+// populated for after_llm_call but executeAfterLLMCallHooks never
+// actually set it — handlers reading model_id always saw an empty
+// string. A single successful turn must dispatch after_llm_call with
+// ModelID equal to the provider's canonical "<provider>/<model>" id.
+func TestAfterLLMCallHook_PopulatesModelID(t *testing.T) {
+	t.Parallel()
+
+	const (
+		hookName = "test-after-llm-model-id"
+		modelID  = "test/mock-model"
+	)
+
+	var captured atomic.Pointer[hooks.Input]
+
+	stream := newStreamBuilder().
+		AddContent("ok").
+		AddStopWithUsage(1, 1).
+		Build()
+	prov := &mockProvider{id: modelID, stream: stream}
+
+	root := agent.New("root", "test agent",
+		agent.WithModel(prov),
+		agent.WithHooks(&latest.HooksConfig{
+			AfterLLMCall: []latest.HookDefinition{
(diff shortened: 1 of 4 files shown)


dgageot (approved):

dgageot:
Hi @kimizuka, thank you for the contribution! I'm sorry, we do require all contributors to sign their commits now. Is it possible for you to sign them?

kimizuka:
> Hi @kimizuka, thank you for the contribution! I'm sorry, we do require all contributors to sign their commits now. Is it possible for you to sign them?

Thanks for the heads-up, @dgageot! I've signed the commit (SSH signing) and force-pushed — it now shows as Verified. The DCO Signed-off-by was already in place. Let me know if there's anything else needed.

PR #2913, opened 2026-05-28
feat: add --disable-commands flag to hide and disable slash commands in TUI

dgageot:
Users sometimes want to limit which slash commands are available in the TUI—for example, to hide commands that shouldn't be used in a particular environment or workflow. This change adds a `--disable-commands` flag to `docker agent run` and `docker agent exec` that lets you specify which commands to hide.

The flag accepts a comma-separated list of slash command names, like `--disable-commands="/cost,/eval,/model"`. Leading slashes are optional and matching is case-insensitive (`/Cost`, `cost`, `/cost` all disable `/cost`). When set, the specified commands are completely hidden and disabled: they're stripped from the command palette (Ctrl+K), the slash-command parser, and the editor completion popup. The filtering happens in a single `commandCategories()` chokepoint, so all three surfaces inherit the behavior consistently.

### DIFF
--- a/pkg/tui/tui.go
+++ b/pkg/tui/tui.go
@@ -188,6 +188,10 @@ type appModel struct {
 
 	appName    string
 	appVersion string
+
+	// disabledCommands holds slash commands to hide and disable.
+	// Normalized to start with "/".
+	disabledCommands map[string]bool
 }
 
 // Transcriber is the speech-to-text interface used by the TUI. It is an
@@ -227,6 +231,32 @@ func WithAppName(name string) Option {
 func WithVersion(v string) Option {
 	return func(m *appModel) {
 		m.appVersion = v
+	}
+}
+
+// WithDisabledCommands hides and disables the given slash commands so they
+// are stripped from the command palette, the slash-command parser, and
+// completion. Each entry is normalized to start with "/" (so "cost" and
+// "/cost" are equivalent) and lower-cased to match the registered slash
+// command names (so "/Cost" and "/cost" are equivalent).
+func WithDisabledCommands(slashCommands []string) Option {
+	return func(m *appModel) {
+		if len(slashCommands) == 0 {
+			return
+		}
+		if m.disabledCommands == nil {
+			m.disabledCommands = make(map[string]bool, len(slashCommands))
+		}
+		for _, c := range slashCommands {
+			c = strings.ToLower(strings.TrimSpace(c))
+			if c == "" {
+				continue
+			}
+			if !strings.HasPrefix(c, "/") {
+				c = "/" + c
+			}
+			m.disabledCommands[c] = true
+		}
 	}
 }
 
@@ -392,7 +422,26 @@ func (m *appModel) reapplyKeyboardEnhancements() {
 }
 
 func (m *appModel) commandCategories() []commands.Category {
-	return m.buildCommandCategories(context.Background(), m)
+	categories := m.buildCommandCategories(context.Background(), m)
+	if len(m.disabledCommands) == 0 {
+		return categories
+	}
+	filtered := make([]commands.Category, 0, len(categories))
+	for _, cat := range categories {
+		items := make([]commands.Item, 0, len(cat.Commands))
+		for _, item := range cat.Commands {
+			if m.disabledCommands[item.SlashCommand] {
+				continue
+			}
+			items = append(items, item)
+		}
+		if len(items) == 0 {
+			continue
+		}
+		cat.Commands = items
+		filtered = append(filtered, cat)
+	}
+	return filtered
 }
 
 // chatPageOpts returns the chat.PageOption slice derived from the current
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -62,11 +62,12 @@ type runExecFlags struct {
 	outputJSON    bool
 
 	// Run only
-	hideToolResults bool
-	lean            bool
-	appName         string
-	listenAddr      string
-	onEventSpecs    []string
+	hideToolResults  bool
+	lean             bool
+	appName          string
+	listenAddr       string
+	onEventSpecs     []string
+	disabledCommands []string
 
 	// globalPermissions holds the user-level global permission checker built
 	// from user config settings. Nil when no global permissions are configured.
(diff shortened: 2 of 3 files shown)


gtardif (dismissed):

pkg/tui/tui.go
```diff
@@ -392,7 +422,17 @@ func (m *appModel) reapplyKeyboardEnhancements() {
...
 
 func (m *appModel) commandCategories() []commands.Category {
-	return m.buildCommandCategories(context.Background(), m)
+	categories := m.buildCommandCategories(context.Background(), m)
+	if len(m.disabledCommands) == 0 {
+		return categories
+	}
+	isDisabled := func(item commands.Item) bool { return m.disabledCommands[item.SlashCommand] }
+	for i := range categories {
+		categories[i].Commands = slices.DeleteFunc(categories[i].Commands, isDisabled)
```
docker-agent: **[MEDIUM] `slices.DeleteFunc` mutates the `Commands` slice in-place — could corrupt shared/cached slices from custom builders**

`commandCategories()` calls `slices.DeleteFunc(categories[i].Commands, isDisabled)` in-place on the slice returned by `buildCommandCategories`. In Go, `slices.DeleteFunc` modifies the underlying array and zeroes out the removed tail — it does **not** copy. If a caller provides a custom builder via `WithCommandBuilder` that returns a cached or shared `[]commands.Category`, the first call to `commandCategories()` with any `disabledCommands` will permanently zero out the tail entries of that shared slice. Every subsequent call will see corrupted (nil/zero) items.

The default builder (`commands.BuildCommandCategories`) allocates a fresh slice each call and is safe. However, the `WithCommandBuilder` doc comment only warns about not accessing `tea.Model` during the build — it does **not** warn callers against returning shared slices, leaving a silent corruption trap for custom builders.

**Fix**: copy `categories[i].Commands` before filtering, e.g.:
```go
cmds := make([]commands.Item, len(categories[i].Commands))
copy(cmds, categories[i].Commands)
categories[i].Commands = slices.DeleteFunc(cmds, isDisabled)
```

pkg/tui/tui.go
```diff
@@ -230,6 +235,31 @@ func WithVersion(v string) Option {
...
+		}
+		if m.disabledCommands == nil {
+			m.disabledCommands = make(map[string]bool, len(slashCommands))
+		}
+		for _, c := range slashCommands {
+			c = strings.TrimSpace(c)
+			if c == "" {
+				continue
+			}
+			if !strings.HasPrefix(c, "/") {
```
docker-agent, line 255: **[LOW] `WithDisabledCommands` is case-sensitive — `/Cost` silently fails to disable `/cost`**

The function normalizes leading slashes (e.g. `"cost"` → `"/cost"`) and trims whitespace, but does **not** fold to lowercase. If a programmatic caller passes `"/Cost"` or `"/COST"`, the map lookup `m.disabledCommands[item.SlashCommand]` will not match the actual slash command (which is stored in lowercase). The command will silently remain enabled.

The doc comment currently only mentions the slash-prefix normalization. In practice, all defined slash commands use lowercase, so interactive users of the CLI flag are unlikely to encounter this. But the `WithDisabledCommands` public API gives no indication that casing matters.

**Suggestion**: either add a note to the doc comment (`// Case-sensitive: entries must match the exact casing of the slash command.`) or normalize with `strings.ToLower`.

docker-agent (commented):

dgageot:
Addressed review feedback:

- **MEDIUM (in-place mutation in `commandCategories`)**: dropped the `slices.DeleteFunc` simplification commit. The original implementation already allocated fresh slices and is restored as the only behavior, so custom builders returning shared/cached slices are no longer at risk.
- **LOW (case-sensitive matching in `WithDisabledCommands`)**: entries are now lower-cased after trimming, so `/Cost`, `cost`, and `/cost` all disable `/cost`. Doc comment updated; new test case `matching is case-insensitive` covers the behavior.

Rebased on top of `origin/main` (the snapshot test fix already landed there as `ac2ff3add`, so the duplicate commit was dropped). Branch force-pushed.

trungutt (approved):

pkg/tui/tui.go
```diff
@@ -392,7 +422,26 @@ func (m *appModel) reapplyKeyboardEnhancements() {
...
-	return m.buildCommandCategories(context.Background(), m)
+	categories := m.buildCommandCategories(context.Background(), m)
+	if len(m.disabledCommands) == 0 {
+		return categories
+	}
+	filtered := make([]commands.Category, 0, len(categories))
+	for _, cat := range categories {
+		items := make([]commands.Item, 0, len(cat.Commands))
+		for _, item := range cat.Commands {
+			if m.disabledCommands[item.SlashCommand] {
```
docker-agent, line 433: **[MEDIUM] Case-insensitive filtering silently fails for dynamic commands with uppercase names**

`WithDisabledCommands` normalises user input to lowercase before storing in the `disabledCommands` map (e.g. `"MyTool"` → `"/mytool"`), but the lookup in `commandCategories()` uses `item.SlashCommand` verbatim — with no `strings.ToLower` applied to the registered value.

For built-in commands this is harmless (they're hardcoded as lowercase strings like `"/cost"`, `"/eval"`). However, dynamically-registered entries — agent commands and skill commands — are built as `"/" + name` where `name` comes directly from the runtime registry:

```go
// pkg/tui/commands/commands.go (existing, not in diff)
SlashCommand: "/" + name,      // agent command — name not lowercased
SlashCommand: "/" + skillName, // skill command — skillName not lowercased
```

If any such name contains uppercase letters (e.g. `"MySkill"` → `"/MySkill"`), passing `--disable-commands=/myskill` adds `"/myskill"` to the map, but the lookup `m.disabledCommands["/MySkill"]` misses it. The PR's claim of case-insensitive matching only holds for the user-input side, not for the registered value side.

**Fix**: lowercase `item.SlashCommand` at the point of lookup:
```go
if m.disabledCommands[strings.ToLower(item.SlashCommand)] {
```

docker-agent (commented):

One medium-severity finding in the newly added  filter. The guard condition for empty categories and the redundant outer guard in  are clean. Flag wiring through , normalization logic, and test coverage are all correct.

PR #2914, opened 2026-05-28
feat: add --app-name flag and fix macOS test symlink issue

dgageot:
The `docker agent run` command now accepts an `--app-name` flag to override the default "docker agent" label shown in the TUI. This string is passed to the existing `tui.WithAppName` option, affecting the status bar, terminal window title, and "/exit" notifications.

We also fixed an unrelated test issue in `pkg/snapshot/snapshot_test.go`. The `bootstrapRepo` helper now resolves symlinks via `filepath.EvalSymlinks` on its temporary directory. This ensures test assertions pass on macOS, where `t.TempDir()` returns `/var/folders/...` (a symlink to `/private/var/folders/...`), while `git rev-parse --show-toplevel` returns the canonical resolved path. This unblocks `TestTrackPatchFromSubfolder` and `TestOpenCanonicalizesSymlinkedDirectory`.

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -64,6 +64,7 @@ type runExecFlags struct {
 	// Run only
 	hideToolResults bool
 	lean            bool
+	appName         string
 	listenAddr      string
 	onEventSpecs    []string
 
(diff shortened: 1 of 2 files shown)


gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, focused PR. The `--app-name` flag is correctly added to the `runExecFlags` struct and wired through `tuiOpts()` with an appropriate empty-string guard. The `filepath.EvalSymlinks` fix in the snapshot test is correct — there is no variable shadowing (the `if`-init `err` is scoped to that clause only), and `t.TempDir()` guarantees the directory exists for the test lifetime.

No confirmed bugs found in the changed code.

PR #2916, opened 2026-05-28
Propagate user-initiated cancellation across the WithoutCancel boundary

trungutt:
## Depends on

This PR is stacked on top of #2896 (`extend-unmanaged-oauth-flow`). The two predecessor commits (`c97124d0f`, `a93ae4d3c`) shown in the diff belong to #2896; review only the last commit in this PR. Will be rebased onto a clean main after #2896 (or its replacement) merges.

## Problem

`clientConnector.Connect` detaches the caller's `ctx` with `context.WithoutCancel` before sending the MCP `initialize` handshake. The rationale is correct: an MCP toolset's session must outlive any single request — otherwise the underlying SSE/stdio connection tears down when the request that triggered `toolset.Start` returns, and a successful OAuth flow gets killed the moment its triggering request completes.

But that detachment also severs **user-initiated cancellation**, which the OAuth flow inside `Connect` DOES need to observe. Concretely:

`handleUnmanagedOAuthFlow` blocks on a select waiting for an elicitation reply or an out-of-band callback. If the embedder aborts the in-progress agent run while the user is in the OAuth dialog, the `streamCtx` the `SessionManager` derived from the request `ctx` gets cancelled — but every ctx in the chain BELOW `Connect`'s `WithoutCancel` boundary stays alive. The OAuth select then blocks forever, the per-session streaming lock is never released, and the next user message returns 409 Conflict from `RunSession.ErrSessionBusy`.

## Fix

Preserve the connection-longevity invariant **and** restore user-initiated cancellation by stashing the caller's original ctx as a value on the detached ctx. The detached ctx remains the primary driver of the connection's lifetime; operations that want to observe user-initiated cancellation can opt in by reading the parent back out and selecting on its `Done` channel alongside the local ctx.

Carrying ctx-as-value is unusual but appropriate here: the parent is **metadata about how the detached ctx was constructed**, not data the request is operating on. The helpers (`withCancellableParent` / `cancellableParentFromContext`) live next to the connector that creates the relationship so the rationale stays close to the code.

`handleUnmanagedOAuthFlow` now extends its select with a `userCancelCh` case keyed on the parent's `Done` channel. When the parent is nil (e.g. unit tests, CLI invocations that don't go through `clientConnector.Connect`), `userCancelCh` stays nil and the select silently never picks it — preserving back-compat.

## Plumbing

- `pkg/tools/mcp/cancellable_parent.go` — new helper file with `withCancellableParent` + `cancellableParentFromContext`, extensively commented because ctx-as-value is unusual.
- `pkg/tools/mcp/mcp.go` — `clientConnector.Connect` captures the caller's ctx before `WithoutCancel` and attaches it via `withCancellableParent`.
- `pkg/tools/mcp/oauth.go` — `handleUnmanagedOAuthFlow` extracts the parent and adds a fourth select case watching its `Done` channel; returns `parent.Err()` so the agent loop sees a cancellable error and the `SessionManager` goroutine's deferred `Unlock` fires.

## Tests

- `cancellable_parent_test.go` — round-trip, absent-returns-nil, nil-parent-no-op, and the key invariant that the local ctx stays independent of the parent's cancellation (only opt-in callers see it).
- `oauth_test.go` — `TestUnmanagedOAuthFlow_DriveFlow_AbortsOnParentCtxCancellation` asserts the new branch fires under the exact boundary `clientConnector.Connect` sets up, returns the parent's ctx error, and skips the token exchange.

### DIFF
--- a/cmd/root/flags.go
+++ b/cmd/root/flags.go
@@ -36,6 +36,12 @@ func addRuntimeConfigFlags(cmd *cobra.Command, runConfig *config.RuntimeConfig) 
 	cmd.PersistentFlags().StringArrayVar(&runConfig.HookSessionEnd, "hook-session-end", nil, "Add a session-end hook command (repeatable)")
 	cmd.PersistentFlags().StringArrayVar(&runConfig.HookOnUserInput, "hook-on-user-input", nil, "Add an on-user-input hook command (repeatable)")
 	cmd.PersistentFlags().StringArrayVar(&runConfig.HookStop, "hook-stop", nil, "Add a stop hook command, fired when the model finishes responding (repeatable)")
+	cmd.PersistentFlags().StringVar(&runConfig.MCPOAuthRedirectURI, "mcp-oauth-redirect-uri", "",
+		"Public HTTPS URL to advertise as the OAuth `redirect_uri` for MCP servers "+
+			"running in unmanaged OAuth mode. When set, docker-agent drives the OAuth flow "+
+			"itself (PKCE + DCR + token exchange) and expects clients to return `{code, state}` "+
+			"via ResumeElicitation. When empty, the client is expected to perform the OAuth "+
+			"flow and return an access token (legacy behavior).")
 }
 
 func setupWorkingDirectory(workingDir string) error {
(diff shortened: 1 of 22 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26577239985).

PR #2917, opened 2026-05-28
feat: add --sidebar flag to control sidebar visibility

### LINKED ISSUE #2912: Flag to hide the sidebar
Add a `--sidebar` bool flag to `docker agent run` that allows hiding the sidebar. When hidden, the sidebar can't be shown back from the TUI.

I hides the sidebar no matter whether it's using the horizontal or vertical layout.

dgageot:
Closes #2912

Adds a `--sidebar` flag to control sidebar visibility in the TUI. When set to `--sidebar=false`, the sidebar is hidden and the user cannot re-enable it via keyboard shortcut or mouse interaction. The flag defaults to `true`, preserving existing behavior.

This complements the existing `--lean` flag, which hides the entire UI chrome. The `--sidebar=false` option is lighter—it keeps the tab bar, status bar, and dialogs intact while hiding only the sidebar. The implementation gates the Ctrl+B keybinding and sidebar toggle behavior when the sidebar is disabled, and adjusts the chat layout to fill the full width when the sidebar is absent.

Build, lint, and tests all pass. (Unrelated pre-existing failures in `pkg/snapshot` on macOS stem from `/private/var` symlink handling and are out of scope.)

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -65,6 +65,7 @@ type runExecFlags struct {
 	hideToolResults  bool
 	lean             bool
 	appName          string
+	sidebar          bool
 	listenAddr       string
 	onEventSpecs     []string
 	disabledCommands []string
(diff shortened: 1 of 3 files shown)


trungutt (dismissed):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is clean and correct. All hypotheses were investigated and dismissed:

**`AllBindings()` guard change** — The comment `"leanMode already returned above"` is accurate: there is an explicit early-return `if m.leanMode { return []key.Binding{quitBinding} }` at the top of the function. The prior `if !m.leanMode` guard on the Ctrl+b binding was therefore always `true` and redundant. Replacing it with `if !m.hideSidebar` is the correct change. (The previous commit message even says "simplify AllBindings: remove redundant leanMode guard", confirming intent.)

**Mouse interaction when `hideSidebar=true`** — `computeSidebarLayout()` returns a layout with `sidebarHeight=0` (zero value) and `showToggle()==false` when hidden. The hit-test `y < sl.sidebarHeight` (i.e. `y < 0`) is never true, so no mouse click can reach the sidebar target. The sidebar cannot be re-enabled by clicking.

**`--sidebar` flag default `true`** — Cobra's `BoolVar` writes the default value immediately at registration time, not at parse time. `f.sidebar` is `true` before any command runs, so `!f.sidebar` is always `false` at call time unless the user explicitly passes `--sidebar=false`. No issue.

**`hideSidebar` vs `leanMode` height difference in `View()`** — Intentional by design: lean mode hides all chrome (no fixed height constraint needed), while `--sidebar=false` keeps the tab bar and status bar (requires `chatHeight` constraint). The `hideSidebar` branch correctly mirrors the `default` collapsed-sidebar branch.

The `sidebarHidden()` helper cleanly centralises the guard, `WithHideSidebar()` correctly disables the keymap binding, and the `ToggleSidebarMsg` gate in `tui.go` provides defence-in-depth. Looks good to merge.

aheritier (approved):

PR #2920, opened 2026-05-29
fix: rune-safe truncation and dead-code cleanup

dgageot:
## Summary

A focused pass of small, high-confidence bug fixes found while reviewing the codebase. Two categories:

### Logic / dead code
- **`evaluation`: progress bar width was always `10`** — `min(max(termWidth-80, 10), 10)` made the `max` dead, so the bar never scaled with the terminal. Restored a sensible upper clamp.
- **`wasm`: removed an `if/else` with identical branches** in `runAgentLoop`'s default-agent selection (the `len(Agents)==1` check was meaningless).

### UTF-8 truncation (byte-index slicing could split a multi-byte rune → `�` / invalid bytes)
The project already standardizes on rune-aware truncation (`[]rune`, `ansi.Truncate`, `truncateRunes`), but several spots still sliced by byte. Fixed in:
- `evaluation` running-task name
- TUI file picker, session browser, working-dir picker, tab titles, search-query preview
- OpenAPI operation descriptions
- `wasm` tool-output preview
- `filesystem` grep search-result preview window (snap to rune boundaries)
- `rag` semantic-embedding input — could feed invalid UTF-8 (encoded as U+FFFD) to the embedding API; now clamped to valid UTF-8 while preserving the byte budget

## Validation
- `go build ./...` and `GOOS=js GOARCH=wasm go build ./cmd/wasm/`
- `go vet` + `golangci-lint`: clean on all touched packages
- Tests pass for every touched package

All changes are behavior-preserving for ASCII input and consistent with existing conventions. Each fix is an individual commit.

### DIFF
--- a/pkg/tools/builtin/filesystem/filesystem.go
+++ b/pkg/tools/builtin/filesystem/filesystem.go
@@ -14,6 +14,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/config"
@@ -1195,8 +1196,15 @@ func (t *ToolSet) handleSearchFilesContent(_ context.Context, args SearchFilesCo
 				preview := line
 				if len(preview) > 100 {
 					start := max(matchStart-20, 0)
-					end := matchEnd + 20
-					end = min(end, len(preview))
+					end := min(matchEnd+20, len(preview))
+					// Snap to rune boundaries so we never split a multi-byte
+					// UTF-8 sequence in the preview.
+					for start > 0 && !utf8.RuneStart(preview[start]) {
+						start--
+					}
+					for end < len(preview) && !utf8.RuneStart(preview[end]) {
+						end++
+					}
 					preview = preview[start:end]
 				}
 
--- a/cmd/wasm/runtime_wasm.go
+++ b/cmd/wasm/runtime_wasm.go
@@ -788,8 +788,8 @@ func parseToolArgs(args string) map[string]any {
 
 // truncateOutput truncates a string to maxLen characters, appending "…" if truncated.
 func truncateOutput(s string, maxLen int) string {
-	if len(s) <= maxLen {
-		return s
+	if r := []rune(s); len(r) > maxLen {
+		return string(r[:maxLen]) + "…"
 	}
-	return s[:maxLen] + "…"
+	return s
 }
--- a/cmd/wasm/runtime_wasm.go
+++ b/cmd/wasm/runtime_wasm.go
@@ -250,11 +250,7 @@ func (rt *wasmRuntime) agentInstruction(name string) string {
 // maxIterations.
 func (rt *wasmRuntime) runAgentLoop(ctx context.Context, agentName string, messages []chat.Message) (map[string]any, error) {
 	if agentName == "" {
-		if len(rt.cfg.Agents) == 1 {
-			agentName = rt.cfg.Agents[0].Name
-		} else {
-			agentName = rt.cfg.Agents[0].Name
-		}
+		agentName = rt.cfg.Agents[0].Name
 	}
 	rt.currentAgent = agentName
 
--- a/pkg/evaluation/progress.go
+++ b/pkg/evaluation/progress.go
@@ -216,8 +216,8 @@ func (p *progressBar) render(final bool) {
 		// Calculate available space for running task name
 		availableForName := max(termWidth-len(status)-10, 5)
 		name := firstName
-		if len(name) > availableForName {
-			name = name[:availableForName-1] + "…"
+		if r := []rune(name); len(r) > availableForName {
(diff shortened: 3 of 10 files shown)


pkg/tools/builtin/filesystem/filesystem.go
```diff
@@ -1195,8 +1196,15 @@ func (t *ToolSet) handleSearchFilesContent(_ context.Context, args SearchFilesCo
...
-					end = min(end, len(preview))
+					end := min(matchEnd+20, len(preview))
+					// Snap to rune boundaries so we never split a multi-byte
+					// UTF-8 sequence in the preview.
+					for start > 0 && !utf8.RuneStart(preview[start]) {
+						start--
+					}
+					for end < len(preview) && !utf8.RuneStart(preview[end]) {
+						end++
+					}
```
aheritier, line 1207: **[LOW] End rune-boundary snap may leave an incomplete UTF-8 sequence at the tail**

The `end`-snapping loop advances forward while the byte is a UTF-8 continuation byte, stopping when `end == len(preview)`. If `matchEnd+20` lands exactly mid-rune and the rest of the sequence is beyond the slice boundary, the loop exits at `len(preview)` and `preview[start:end]` can still end with an incomplete multi-byte sequence.

```go
for end < len(preview) && !utf8.RuneStart(preview[end]) {
    end++
}
```

The `start` side correctly walks *backward* to find a rune boundary, but the `end` side can only walk forward up to `len(preview)` — it has no way to detect that it stopped mid-rune. Snapping `end` backward (like `start`) would be safer:

```go
for end > start && !utf8.RuneStart(preview[end-1]) {
    // actually: snap end backward to last rune boundary
}
```

Or simply use `string([]rune(preview[start:end]))` to drop any trailing incomplete sequence.

Impact is low — this only occurs if the match region falls at the very end of a multi-byte character in a long line.

aheritier (commented):
### Assessment: 🟢 APPROVE

The rune-safe truncation fixes are correct and consistent with the project's existing conventions. The dead-code removal in `runAgentLoop` is safe — both branches of the old if/else were identical. All high/medium findings were pre-existing and not introduced or worsened by this PR.

One low-severity edge case noted inline.

docker-agent (commented):
### Assessment: 🟢 APPROVE

**PR:** fix: rune-safe truncation and dead-code cleanup

This PR correctly fixes byte-level string truncation across multiple files by converting to `[]rune` slices before slicing, preventing splits of multi-byte UTF-8 sequences. The dead-code removal in `runAgentLoop` (collapsing an if/else with identical branches) is also a clean, correct simplification.

**Verification summary:**
- `strings.ToValidUTF8(s[:maxLen], "")` in `semantic_embeddings.go`: ✅ Correct — dropping the partial rune at a hard byte-boundary is the only safe option; the resulting string is valid UTF-8 and within budget.
- `[]rune` truncation in `working_dir_picker.go`, `file_picker.go`, `session_browser.go`, `tabbar/tab.go`, `searchfilescontent.go`, `openapi.go`, `runtime_wasm.go`: ✅ All correctly fixed.
- `barWidth` cap raised from 10→30 in `progress.go`: ✅ Correct — the old cap of 10 meant the bar was always clamped to 10 chars regardless of terminal width.
- UTF-8 boundary snapping in `filesystem.go` preview: ✅ Correct use of `utf8.RuneStart` to expand/contract to valid boundaries.
- Dead-code removal (`agentName` if/else with identical branches): ✅ No behavior change.

No bugs found in the changed code. All changes are improvements.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR makes three well-scoped categories of fixes:

1. **Dead-code removal** (`runtime_wasm.go`): The redundant `if/else` with identical branches is correctly collapsed. The existing nil/empty-slice panic potential on `rt.cfg.Agents[0]` is pre-existing and not introduced here.

2. **Progress-bar clamp fix** (`evaluation/progress.go`): Changing the upper bound from `10` to `30` in `min(max(termWidth-80, 10), 30)` correctly restores dynamic scaling. The formula is sound — minimum bar width remains 10 (small terminals), maximum is 30 (wide terminals).

3. **Rune-safe truncation** (9 sites): All conversions from byte-index slicing to `[]rune` slicing are correct. The `strings.ToValidUTF8(s[:n], "")` pattern in the RAG embedding path correctly strips incomplete trailing sequences. The `utf8.RuneStart` boundary-snapping in `filesystem.go` is safe — no out-of-bounds access is possible.

**Minor observation** (`pkg/tools/builtin/filesystem/filesystem.go`, line 1205): The `end`-adjustment loop (`for end < len(preview) && !utf8.RuneStart(preview[end]) { end++ }`) can expand the preview window slightly past `matchEnd+20` when continuation bytes happen to fall exactly at the cut-point. This is harmless (no panic, no data corruption), and the alternative would be to shrink instead of expand — which would also work but could leave a dangling continuation byte in the other direction. Worth knowing but not a blocking issue.

All changes are consistent with existing project conventions and the PR description is accurate.

aheritier (approved):

PR #2925, opened 2026-05-29
fix(mcpcatalog): tell the model to proceed after enabling an OAuth server

trungutt:
## Problem

When a user asks something that requires an OAuth-protected catalog
server (e.g. "List all pages by David Gageot in Notion"), the model:

1. calls `search_remote_mcp_servers`,
2. calls `enable_remote_mcp_server` for `notion`,
3. receives a tool result ending with **"auth: OAuth — an authorization URL will be elicited on the next turn."**,
4. then narrates to the user "OAuth will happen on your next request, please ask me again" and stops.

The user must repeat their original prompt to actually get an answer:

> The Notion server is now enabled and ready to use. When you make your next request, it will initiate OAuth authorization. […] Would you like me to proceed with searching your Notion workspace for pages and David Gageot's comments?

## Why this happens

The runtime actually drives the OAuth flow inside the **same** loop iteration:

- `processToolCalls` runs `handleEnable`, which only registers the toolset.
- `reprobe` (`pkg/runtime/loop.go:608-610`) calls `getTools` → `Tools()` on the catalog → first-call lazy `Start()` on the freshly enabled toolset (`mcpcatalog.go:371-372`).
- `Start()` runs the MCP handshake, the 401 triggers the OAuth elicitation, the user grants, `Start()` returns success and the server's tools become available.
- The next loop iteration calls the model with the new tools already in the prompt.

So by the time the next LLM turn begins, OAuth is already complete. The model just doesn't know that — it only sees the previous tool result, which literally said "elicited on the next turn".

## Fix

This PR rewrites the two strings the model uses to reason about the post-enable state:

1. **`handleEnable`'s OAuth-branch success message** now tells the model the host triggers any required authorization transparently the first time the server's tools are used, and that the model should **continue with the user's original request now**.

2. **`Instructions()` step 3** is hardened from a soft "Use the newly activated tools as you would any other" into an explicit "Immediately proceed with the user's ORIGINAL request […]; do NOT stop and ask the user to repeat themselves; do NOT narrate the OAuth flow".

No runtime behaviour changes; this is a prompt fix only

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
+++ b/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
@@ -143,27 +143,22 @@ Workflow:
      Use any term related to the user's intent ("notion", "stripe",
      "docs", "search", "browser", …).
   2. Call ` + ToolNameEnable + ` with the server's "id" to activate it.
-     This adds the server's tools to your set on the *next* turn.
-     Authentication (OAuth or API key) is deferred — it is triggered when
-     the host enumerates the server's tools, which happens once enabling
-     completes. For api_key servers, make sure the listed env var(s) are
-     set in the user's shell BEFORE enabling, otherwise the server will
-     refuse the connection.
+     The server's tools appear in your set on the *next* turn; the host
+     handles any required connection setup. For api_key servers, make
+     sure the listed env var(s) are set in the user's shell BEFORE
+     enabling, otherwise the server will refuse the connection.
   3. On your NEXT turn (once the activated server's tools appear in
      your set), go straight to the user's ORIGINAL request using those
      tools. Do NOT stop and ask the user to repeat themselves; do NOT
-     narrate the OAuth flow or say things like "OAuth will happen on
-     the next turn" — if OAuth is needed the host surfaces the
-     authorization dialog and the tool call resumes once the user
-     grants access. Enabling a server is a means to an end, not a
-     stopping point.
+     narrate connection setup. Enabling a server is a means to an end,
+     not a stopping point.
   4. Call ` + ToolNameDisable + ` to remove a server when no longer needed.
      This tool only appears once at least one server is enabled.
-  5. If a previously authorized OAuth server starts rejecting requests
-     (token revoked, scopes changed, signed in to the wrong account),
-     call ` + ToolNameResetAuth + ` to wipe the persisted credentials.
-     The next enable will trigger a fresh authorization URL. This tool
-     also only appears once at least one server is enabled.
+  5. If a previously enabled server starts rejecting requests
+     (credentials revoked, scopes changed, signed in to the wrong
+     account), call ` + ToolNameResetAuth + ` to clear any persisted
+     credentials so the next enable starts fresh. This tool also only
+     appears once at least one server is enabled.
 
 Prefer enabling only the servers you actually need — every server adds
 tools to the prompt and contributes to context usage.`
@@ -315,7 +310,7 @@ func (t *Toolset) Tools(ctx context.Context) ([]tools.Tool, error) {
 		{
 			Name:         ToolNameEnable,
 			Category:     "mcp_catalog",
-			Description:  "Activate a remote MCP server from the catalog by id. Connection (and any required OAuth flow or API-key check) is deferred until the host next lists the agent's tools.",
+			Description:  "Activate a remote MCP server from the catalog by id. Connection setup is deferred until the host next enumerates tools; the server's tools become available on the next turn.",
 			Parameters:   tools.MustSchemaFor[EnableArgs](),
 			OutputSchema: tools.MustSchemaFor[string](),
 			Handler:      tools.NewHandler(t.handleEnable),
@@ -352,7 +347,7 @@ func (t *Toolset) Tools(ctx context.Context) ([]tools.Tool, error) {
 			tools.Tool{
 				Name:         ToolNameResetAuth,
 				Category:     "mcp_catalog",
-				Description:  "Clear persisted OAuth credentials (access token, refresh token, dynamic-client-registration data) for a catalog server. The next enable will trigger a fresh authorization flow. No-op for api_key/none servers.",
+				Description:  "Clear any persisted credentials for a catalog server so the next enable starts a fresh connection. Use when a previously enabled server starts rejecting requests. No-op for servers with no persisted credentials.",
 				Parameters:   tools.MustSchemaFor[ResetAuthArgs](),
 				OutputSchema: tools.MustSchemaFor[string](),
 				Handler:      tools.NewHandler(t.handleResetAuth),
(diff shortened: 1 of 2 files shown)


pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -149,7 +149,13 @@ Workflow:
      completes. For api_key servers, make sure the listed env var(s) are
      set in the user's shell BEFORE enabling, otherwise the server will
      refuse the connection.
-  3. Use the newly activated tools as you would any other.
+  3. Immediately proceed with the user's ORIGINAL request by calling the
```
docker-agent: **[LOW] Mild ambiguity between "next turn" mechanics (step 2) and "immediately proceed" directive (step 3)**

Step 2 (unchanged) tells the model: *"This adds the server's tools to your set on the **next** turn."* The new step 3 says: *"Immediately proceed with the user's ORIGINAL request by calling the newly activated server's tools."*

A model reading these sequentially could interpret "immediately proceed … by calling the newly activated server's tools" as meaning "call them right now, in this same response" — before those tools are actually available. In practice, most LLMs will read "proceed" as directional guidance (don't stop, make your next turn about those tools), which is the intended meaning. But the word **"Immediately"** combined with **"by calling the newly activated server's tools"** creates a surface-level tension with the step 2 text that could trip up a literal model.

Consider softening to something like: *"On your next turn, go straight to the user's ORIGINAL request using the newly activated server's tools — do NOT stop to ask the user to repeat themselves…"* This makes the timing explicit and removes the apparent contradiction.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -597,7 +603,7 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
 	fmt.Fprintf(&msg, "endpoint: %s\n", server.URL)
 	switch server.Auth.Type {
 	case "oauth":
-		msg.WriteString("auth: OAuth — an authorization URL will be elicited on the next turn.\n")
+		msg.WriteString("auth: OAuth — the host will trigger any required authorization dialog transparently the first time the server's tools are used. Continue with the user's original request now; do not pause to announce the OAuth flow.\n")
```
docker-agent: **[MEDIUM] "transparently" overstates seamlessness for unmanaged-OAuth path**

The new message:
> *"the host will trigger any required authorization dialog **transparently** the first time the server's tools are used"*

The word **transparently** implies the OAuth handshake is invisible/in-process — which is accurate for the managed-OAuth path (`managedOAuth=true`, where `Start()` drives the 401→redirect→grant cycle in-process). However, when `managedOAuth=false` (unmanaged OAuth), the flow requires the user to follow a browser redirect and grant access externally; it is not transparent.

Because `handleEnable` uses a single `case "oauth":` branch with no check on `managedOAuth`, both flows receive this same message. A model reading this during an unmanaged-OAuth flow may expect the authorization to happen silently and be surprised when an elicitation URL appears — or, worse, it may proceed to call tools immediately and receive auth failures.

Consider either:
1. Branching the message on `managedOAuth` to say "transparently" only when it's true, or
2. Replacing "transparently" with wording that acknowledges possible user interaction: *"the host will surface any required authorization and resume once access is granted"*.

docker-agent (commented):

This is a prompt-only fix that rewrites two LLM-facing instruction strings in `mcpcatalog.go`. No runtime logic is changed. Two findings surfaced during review — both are edge-case wording concerns rather than code bugs.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
trungutt: Good catch — applied your suggested phrasing in dddf7faff. Step 3 now starts with "On your NEXT turn (once the activated server's tools appear in your set), go straight to the user's ORIGINAL request using those tools", which removes the apparent contradiction with step 2 and makes the timing explicit.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
trungutt: Fair point — "transparently" did overstate things for the unmanaged-OAuth path where a browser redirect is required. Picked option 2 in dddf7faff: the message now reads "the host will surface any required authorization (browser redirect or in-app dialog) the first time the server's tools are used, and resume the tool call once access is granted". I went with a single-branch wording rather than splitting on `managedOAuth` because the unmanaged-aware phrasing is also accurate for the managed path.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
docker-agent: Thanks for the quick fix! The new phrasing completely resolves the timing ambiguity — "On your NEXT turn (once the activated server's tools appear in your set)" makes the sequence crystal clear and removes any possible misinterpretation.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
docker-agent: Thanks for the quick fix in dddf7fa! The new wording — *"the host will surface any required authorization (browser redirect or in-app dialog) ... and resume the tool call once access is granted"* — is much clearer and accurately represents both OAuth paths without overpromising seamlessness.

Good call keeping a single message rather than branching on `managedOAuth`: the unmanaged-aware phrasing works equally well for the managed flow, and it avoids duplicating logic.

rumpl:
Hold on, why does the LLM need to know that there is an oauth thing?

trungutt:
@rumpl Good question — and you're right, it shouldn't. The original wording (`auth: OAuth — elicited on the next turn`) caused this bug precisely because it leaked OAuth as a concept the model then tried to reason about. My first patch only rewrote that sentence to be less misleading; that left the leaky abstraction in place.

c3c85b3 removes OAuth as an LLM-visible concept from the catalog entirely:

- `handleEnable` no longer branches on `server.Auth.Type`. The success message is a single line — "the server's tools will appear on your next turn. Proceed with the user's original request; the host handles any required connection setup." The only conditional emission left is the `WARNING: env var(s) NOT set` line, which is the one case the model has actual agency (it can ask the user to set the var and retry).
- `Instructions()` drops "Authentication (OAuth or API key) is deferred" from step 2, drops "do not narrate the OAuth flow" from step 3 (the model no longer learns OAuth exists, so the instruction can just say "do not narrate connection setup"), and reframes step 5 (reset_auth) as "if a previously enabled server starts rejecting requests, clear any persisted credentials" — no mention of OAuth tokens or authorization URLs.
- `enable_remote_mcp_server` and `reset_remote_mcp_server_auth` tool descriptions, `ResetAuthArgs` jsonschema, and `handleResetAuth` result strings all lose every OAuth/authorization reference.

Internal field names (`oauthSuccessHandler`, `managedOAuth`, …), Go package-doc comments, and slog debug messages still say OAuth — those are developer-facing and removing them would obscure intent for anyone reading the package.

A new test assertion (`NotContains("OAuth")`, `NotContains("authorization")` in `TestEnableAddsServerAndFiresChangeNotification`) guards against re-adding the leak.

PTAL.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a prompt/message-only fix that removes OAuth-specific wording from LLM-facing tool results and instructions so the model proceeds with the user's original request instead of stopping to ask them to repeat themselves.

**What was reviewed:**
- `pkg/tools/builtin/mcpcatalog/mcpcatalog.go` — `handleEnable` message consolidation, `handleResetAuth` message updates, `Instructions()` step 3 hardening, tool description rewrites
- `pkg/tools/builtin/mcpcatalog/mcpcatalog_test.go` — updated string assertions to match new production strings

**Findings summary:**
- ✅ No high or medium severity bugs found
- ✅ Auth-type consolidation in `handleEnable` correctly preserves env-var warnings for all relevant auth types (api_key, custom headers); the removed OAuth-specific branch was informational-only and its removal is intentional
- ✅ All updated test assertions (`"enabled"`, `"cleared credentials"`, `"no persisted credentials"`, `assert.NotContains` for `"OAuth"` / `"authorization"`) correctly match the new production string literals
- ℹ️ Minor (not flagged): `handleResetAuth`'s early-exit message `"has no persisted credentials"` is slightly imprecise for api_key servers (they *have* credentials, just not ones stored by this tool) — worth a follow-up comment, but not a bug and consistent with the PR's intent of not leaking auth-type details to the model

The fix correctly addresses the root cause: the previous `"auth: OAuth — an authorization URL will be elicited on the next turn"` string was the signal causing models to pause and narrate instead of proceeding. The new wording and explicit step-3 instruction (`"do NOT stop and ask the user to repeat themselves"`) should resolve the reported behavior.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -593,28 +595,17 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
 	}
 
 	msg := strings.Builder{}
-	fmt.Fprintf(&msg, "enabled %q (%s) — connection deferred until the host next enumerates tools.\n", id, server.Title)
+	fmt.Fprintf(&msg, "enabled %q (%s) — the server's tools will appear on your next turn. Proceed with the user's original request; the host handles any required connection setup.\n", id, server.Title)
```
docker-agent, line 598: **[MEDIUM] Contradictory model instructions when API-key env vars are missing**

When `missing` is non-empty, the message assembled in `handleEnable` simultaneously tells the model two conflicting things:

1. **"Proceed with the user's original request; the host handles any required connection setup."**
2. **"WARNING: the following env var(s) are NOT set: … Ask the user to set them, then call `disable_remote_mcp_server` and `enable_remote_mcp_server` for this id again, otherwise tool calls will fail."**

An LLM receiving both instructions in the same tool-result has no explicit prioritisation signal — the `Proceed` clause comes first and is unconditional, while the `WARNING` appears later but doesn't say "ignore the previous instruction". Depending on which instruction the model weighs more heavily it may either:
- dutifully proceed and immediately fail every downstream tool call (env vars not set), or
- correctly pause and ask the user to set the vars but now via a path the new wording was designed to remove.

**Suggested fix**: move the "Proceed with the user's original request" sentence into the `else` branch (i.e., only emit it when `len(missing) == 0`), so it is never paired with a WARNING that contradicts it:

```go
msg := strings.Builder{}
fmt.Fprintf(&msg, "enabled %q (%s) — the server's tools will appear on your next turn.\n", id, server.Title)
fmt.Fprintf(&msg, "endpoint: %s\n", server.URL)
if len(missing) > 0 {
    fmt.Fprintf(&msg, "WARNING: the following env var(s) are NOT set: %s. Ask the user to set them, then call %s and %s for this id again, otherwise tool calls will fail.\n",
        strings.Join(missing, ", "), ToolNameDisable, ToolNameEnable)
} else {
    fmt.Fprintf(&msg, "Proceed with the user's original request; the host handles any required connection setup.\n")
}
```

docker-agent (commented):

One medium-severity issue found in the changed code. The rest of the prompt-text changes look correct and well-motivated.

dgageot (approved):

PR #2928, opened 2026-05-29
feat: add allow/block-list of servers to the mcp_catalog tool

dgageot:
The `mcp_catalog` builtin tool now supports allow-listing and block-listing of MCP servers. This lets agents enforce a curated set of available integrations and prevent access to specific servers—useful for compliance, security policies, or ensuring a focused set of tools in multi-agent deployments.

Configuration is straightforward: set `allowed_servers` (optional) to whitelist specific catalog server IDs, and `blocked_servers` (optional) to exclude them. Block-list takes precedence if a server appears in both. Filtering applies to all catalog operations: search, enable, and `reset_auth`. Empty or whitespace-only entries are rejected at validation time, and a warning is logged if a configuration references unknown server IDs, helping catch typos early.

An example configuration can be found in `examples/mcp_catalog_filtered.yaml`. All changes are backward compatible—existing configurations without these fields work unchanged.

### DIFF
--- a/examples/mcp_catalog_filtered.yaml
+++ b/examples/mcp_catalog_filtered.yaml
@@ -1,0 +1,31 @@
+# An agent that exposes only a curated subset of the Docker MCP Catalog.
+#
+# The mcp_catalog toolset can be narrowed with two optional lists:
+#   - allowed_servers: when non-empty, ONLY these catalog server ids are
+#     searchable and enableable. Every other entry is hidden.
+#   - blocked_servers: removes individual ids from the offered set. It is
+#     applied after allowed_servers, so a server present in both is blocked.
+#
+# Use these to give an agent a focused, predictable menu of servers instead
+# of the full catalog (which contains hundreds of entries).
+
+agents:
+  root:
+    model: anthropic/claude-sonnet-4-6
+    description: Agent restricted to a curated set of remote MCP servers.
+    instruction: |
+      You can discover and activate remote MCP servers on demand, but only
+      a curated subset of the catalog is available to you. Use
+      search_remote_mcp_servers to find one, then enable_remote_mcp_server
+      to activate it.
+    toolsets:
+      - type: mcp_catalog
+        # Only these servers are offered by default.
+        allowed_servers:
+          - docker-docs
+          - microsoft-learn
+          - hugging-face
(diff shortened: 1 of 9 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The allow/block-list filtering implementation for `mcp_catalog` is correct and well-structured. Core logic is sound: block-list takes precedence over allow-list, filtering runs at construction time (no shared-state mutation), unknown server IDs emit a warning at startup, and config validation rejects empty/whitespace entries. Wiring through `registry.go` and the schema update are both correct. All changes are backward compatible.

**Low-severity observations** (no action required, noted for awareness):

- **`toIDSet` nil-vs-empty-map contract** (`mcpcatalog.go:237`): `toIDSet(nil)` returns `nil` while `toIDSet([]string{"  "})` returns a non-nil empty map. Callers gate on `len(set) > 0` so observable behavior is identical today. Validation already rejects whitespace entries upstream, making this path unreachable in production—safe but subtly inconsistent.
- **`TestSearchRespectsAllowList` substring risk** (`mcpcatalog_test.go:324`): The test verifies that searching for the dropped server's ID returns an error. If that ID becomes a substring of an allowed server's title or description as the embedded catalog evolves, the test could produce a false pass. Low risk given the test's controlled server IDs, but worth watching.
- **Noisy warnings for future-proofed block-lists** (`mcpcatalog.go:175`): IDs in `blocked_servers` that don't exist in today's catalog generate a `slog.Warn`—intentional for catching typos, but could be noisy for operators who proactively block anticipated-future server IDs.

aheritier (approved):

examples/mcp_catalog_filtered.yaml
```diff
@@ -0,0 +1,31 @@
...
+    toolsets:
+      - type: mcp_catalog
+        # Only these servers are offered by default.
+        allowed_servers:
+          - docker-docs
+          - microsoft-learn
+          - hugging-face
+          - gitmcp
+        # ...minus any explicitly blocked one (block wins over allow).
+        blocked_servers:
```
rumpl, line 30: This is a bad example because blocked_servers doesn't do anything here since only the allowed ones are allowed

PR #2931, opened 2026-05-29
TUI - Improve notifications

krissetto:
Misc improvements to the TUI notifications:

- don't auto remove notifications if the mouse is hovering them
- clicking the notification will copy its contents
- use a glyph for the `x` to close, and make it red when hovered
- show action hint text in the top left of the notification border

Screenshots:

Screenshot From 2026-05-29 14-45-53

---

Screenshot From 2026-05-29 14-46-25

---

Screenshot From 2026-05-29 14-46-13

---

Screenshot From 2026-05-29 14-46-18

### DIFF
--- a/pkg/tui/components/notification/notification.go
+++ b/pkg/tui/components/notification/notification.go
@@ -5,22 +5,29 @@ import (
 	"strings"
 	"sync/atomic"
 	"time"
+	"unicode/utf8"
 
 	tea "charm.land/bubbletea/v2"
 	"charm.land/lipgloss/v2"
+	"github.com/mattn/go-runewidth"
 
-	"github.com/docker/docker-agent/pkg/tui/core"
 	"github.com/docker/docker-agent/pkg/tui/styles"
 )
 
 const (
-	closeButton          = " [x]"
-	defaultDuration      = 3 * time.Second
+	defaultDuration      = 10 * time.Second
 	notificationPadding  = 2
 	maxNotificationWidth = 80 // Maximum width to prevent covering too much screen
+	closeGlyph           = "✕"
+	closeInset           = 1
+	hoverLabel           = "click to copy"
+	copiedLabel          = "copied!"
+	closeLabel           = "dismiss"
 )
 
 var nextID atomic.Uint64
+
+var timerDuration = defaultDuration
 
 // Type represents the type of notification
 type Type int
@@ -32,9 +39,8 @@ const (
 	TypeError
 )
 
-// persistent returns true for notification types that stay until manually dismissed.
-func (t Type) persistent() bool {
-	return t == TypeWarning || t == TypeError
+func (t Type) autoHideDuration() time.Duration {
+	return timerDuration
 }
 
 // style returns the lipgloss style for this notification type.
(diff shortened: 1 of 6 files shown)


rumpl (dismissed):

docker-agent (commented):

Two medium-severity issues found in the new notification interaction code.

**Summary of changes reviewed:** Mouse-hover-aware auto-hide suppression, click-to-copy clipboard integration, close-button glyph overlay, and border hint text injection — all in the TUI notification component.

pkg/tui/components/notification/notification.go
```diff
@@ -197,47 +297,245 @@ func (n *Manager) GetLayer() *lipgloss.Layer {
...
+	lbl := styles.NoStyle.Foreground(styles.TextMutedGray)
+	lines[0] = bdr.Render("╭─") +
+		lbl.Render(paddedLabel) +
+		bdr.Render(strings.Repeat("─", totalWidth-3-labelWidth)) +
+		bdr.Render("╮")
+
+	return strings.Join(lines, "\n")
+}
+
+func visibleColumnByteIndex(s string, targetCol int) int {
```
docker-agent, line 508: **[MEDIUM] `visibleColumnByteIndex` only handles SGR ANSI sequences — non-SGR sequences corrupt the byte index**

The ANSI escape-sequence skipper in `visibleColumnByteIndex` scans forward until it finds `'m'`, which correctly handles SGR (Select Graphic Rendition) sequences like `\x1b[31m`. However, other ANSI/VT sequence families are **not** handled:

- **Cursor-movement sequences** end in `A`, `B`, `C`, `D`, `H`, etc. — the scanner would scan past `m` characters inside the sequence parameters and land at the wrong offset.
- **OSC sequences** (`\x1b]…\x07` or `\x1b]…\x1b\\`) are used by newer charmbracelet libraries for hyperlinks. The inner scan for `'m'` could find a literal `m` in the URL, stopping the skipper prematurely and leaving the cursor inside the escape sequence.

If lipgloss ever emits such sequences (charmbracelet/x/ansi in lipgloss v1.x does add hyperlink support), `overlayCloseButton` would compute a wrong byte index and either corrupt the rendered line or silently skip the overlay (`idx < 0` guard).

**Suggested fix**: Use `github.com/charmbracelet/x/ansi`'s `StripString`/sequence iterator (already a transitive dependency) rather than a hand-rolled SGR-only skipper, or at minimum handle the full set of escape sequence terminators (`[\x40-\x7E]` for CSI final bytes).

pkg/tui/tui.go
```diff
@@ -2024,37 +2031,58 @@ func (m *appModel) handleMouseClick(msg tea.MouseClickMsg) (tea.Model, tea.Cmd)
...
 func (m *appModel) handleMouseMotion(msg tea.MouseMotionMsg) (tea.Model, tea.Cmd) {
+	var cmds []tea.Cmd
+	batchWith := func(cmd tea.Cmd) tea.Cmd {
+		if cmd != nil {
+			cmds = append(cmds, cmd)
+		}
+		return tea.Batch(cmds...)
+	}
+
+	if !m.leanMode {
```
docker-agent, line 2042: **[MEDIUM] Click-to-copy silently broken in `leanMode` — `hoveredID` is never populated**

`handleMouseMotion` guards the call to `HandleMouseMotion` behind `!m.leanMode`, so `n.hoveredID` is **never set** in lean mode:

```go
if !m.leanMode {
    updated, cmd := m.notification.HandleMouseMotion(msg.X, msg.Y)
    m.notification = updated
    ...
}
```

Meanwhile `handleMouseClick` calls `CopyHit` **unconditionally** (no `leanMode` guard):

```go
if id, text, ok := m.notification.CopyHit(msg.X, msg.Y); ok {
    return m, copyNotificationToClipboard(id, text)
}
```

`CopyHit` requires `id == n.hoveredID`, but since `hoveredID` is always `0` in lean mode, the condition never passes and click-to-copy silently fails. Notifications are still rendered and mouse clicks are still dispatched in lean mode, so this is a user-visible gap: the user clicks a notification and nothing happens.

**Suggested fix**: Either add a `!m.leanMode` guard around the `CopyHit` call (and document that the feature is disabled), or call `HandleMouseMotion` in lean mode as well so hover state is tracked.

dgageot (approved):

PR #2937, opened 2026-05-30
feat: add --agent-picker flag for agent selection UI

dgageot:
When running `docker agent run` with multiple agents available, users currently have to know the agent name or use trial-and-error to pick the right one. This change adds a `--agent-picker` flag that opens a full-screen pre-TUI dialog for browsing and selecting an agent interactively before launching the main agent TUI.

The new `--agent-picker` flag accepts an optional comma-separated list of agent references (defaulting to `default,coder`) and displays them in a scrollable dialog with the selected agent's raw YAML configuration visible in a collapsible panel. Keyboard shortcuts let users view or hide the YAML details (`?` to toggle), navigate with arrow keys, and confirm with Enter. The agent is resolved before sandbox or alias processing, so the chosen agent's defaults are honored. The feature is rejected cleanly in `--exec` or non-TTY modes where interactive selection doesn't make sense.

The implementation includes comprehensive test coverage, proper sanitization of untrusted YAML content (stripping terminal escape sequences), soft-wrap support for long lines, and syntax highlighting using the active TUI theme.

### DIFF
--- a/cmd/root/agent_picker.go
+++ b/cmd/root/agent_picker.go
@@ -152,6 +152,7 @@ type agentPickerModel struct {
 func newAgentPickerModel(choices []agentChoice) *agentPickerModel {
 	vp := viewport.New()
 	vp.FillHeight = true
+	vp.SoftWrap = true
 	return &agentPickerModel{
 		choices:    choices,
 		details:    vp,
@@ -278,15 +279,16 @@ func (m *agentPickerModel) detailsContent(choice agentChoice) string {
 	case choice.yaml != "":
 		return highlightYAML(strings.TrimRight(choice.yaml, "\n"))
 	case choice.err != nil:
-		return "Failed to load agent:\n\n" + choice.err.Error()
+		return "Failed to load agent:\n\n" + sanitizeYAML(choice.err.Error())
 	default:
 		return "No configuration available."
 	}
 }
 
 // highlightYAML syntax-colorizes YAML using chroma with the active TUI theme.
-// On any tokenisation error it returns the source unchanged.
+// On any tokenisation error it returns the (sanitized) source unchanged.
 func highlightYAML(src string) string {
+	src = sanitizeYAML(src)
 	lexer := lexers.Get("yaml")
 	if lexer == nil {
 		return src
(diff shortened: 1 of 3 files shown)


cmd/root/agent_picker.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// stripControl removes control characters (including ESC) that could inject
+// terminal escape sequences or corrupt the layout. Newlines are preserved.
+func stripControl(s string) string {
+	return strings.Map(func(r rune) rune {
+		if r == '\n' {
+			return r
+		}
+		if r < 0x20 || r == 0x7f {
```
docker-agent, line 326: **[MEDIUM] `stripControl` does not filter C1 control characters (U+0080–U+009F)**

The `stripControl` function only removes C0 control characters (`< 0x20`) and DEL (`0x7f`), but entirely skips the C1 control range (`U+0080`–`U+009F`). Several characters in this range are meaningful on terminals: `U+009B` (CSI) is the 8-bit equivalent of `ESC [` and can introduce ANSI escape sequences on terminals that support 8-bit control codes.

Since the PR description explicitly states that untrusted remote YAML content is sanitized to prevent terminal escape injection, this is a gap in the defence. A malicious agent YAML could embed `\x9b` (CSI) followed by ANSI sequence bytes and bypass the current filter entirely.

**Suggested fix:**
```go
func stripControl(s string) string {
    return strings.Map(func(r rune) rune {
        if r == '\n' {
            return r
        }
        if r < 0x20 || (r >= 0x7f && r <= 0x9f) {
            return -1
        }
        return r
    }, s)
}
```

This removes DEL (`0x7f`) and the entire C1 range (`0x80`–`0x9f`) in one range check, which also correctly handles DEL without a separate condition.

docker-agent (commented):

This PR adds a `--agent-picker` flag with a full-screen interactive TUI for selecting an agent before launching. The implementation is generally well-structured: flag wiring is correct, navigation guards are in place, the viewport/scrollbar sizing formula in `renderDetails()` is intentional and correct (title+help are sized to the full body width including the scrollbar column), and `prependAgentRef` is safe. One medium finding in the YAML sanitization logic.

trungutt (approved):

PR #2938, opened 2026-05-30
feat: run agents in isolated git worktrees

dgageot:
## What

Adds first-class support for running an agent inside a git worktree, so its changes stay isolated from your checkout on a dedicated branch. Built up over four focused commits:

1. **`--worktree[=name]` / `-w`** — Run in a fresh worktree of the current repo. An explicit name is optional; otherwise a friendly name is generated (Docker-style, e.g. `focused_turing`). Stored at `<data-dir>/worktrees/<name>` on a `worktree-<name>` branch. Validated against path traversal; mutually exclusive with `--remote` and `--sandbox`.

2. **`worktree_create` hook** — A new lifecycle event fired just after the worktree is created and before the session starts. Hooks run *inside* the new worktree, receiving `worktree_path`, `worktree_branch`, and `worktree_source_dir`, so they can prepare the checkout (copy `.env`, install deps, warm caches). A blocking verdict aborts the run.

3. **Cleanup on exit** — When the interactive session ends:
   - **Clean** (no uncommitted changes, untracked files, or new commits) → worktree and branch removed automatically.
   - **Has work** → prompt to keep or remove (defaults to *keep* on any non-yes answer, so work is never lost silently).
   - **`--exec`** → never cleaned up; left for inspection.
   - **TUI error** → left in place.
   - Pre-existing worktrees are never touched.

4. **`--worktree-pr <number|url>`** — Check out an existing GitHub pull request to continue it. The PR's head branch is checked out tracking its remote (via `gh pr checkout`), so commits push back to the PR. Handles fork remotes and upstream tracking. Requires the GitHub CLI.

## Notes

- Worktree name generation vendors Moby's `namesgenerator` (Apache-2.0) verbatim to avoid pulling in the full engine module.
- `--worktree-pr` requires `gh` to be installed and authenticated; missing/malformed inputs produce clear errors.

## Testing

- Unit tests for worktree create/status/remove, name validation, PR-ref parsing, and the cleanup decision paths (auto-remove, keep-on-decline, confirm-remove).
- `worktree_create` hook covered end-to-end; fixed `hookWorkingDir` to default to the executor's working dir (surfaced by the CLI-dispatched event).
- Docs (CLI reference + hooks) and an example config included.
- `task build` / tests / `golangci-lint` / custom lint cops all pass.

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -22,6 +22,7 @@ import (
 	latestcfg "github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/hooks"
 	"github.com/docker/docker-agent/pkg/hooks/builtins"
+	"github.com/docker/docker-agent/pkg/input"
 	"github.com/docker/docker-agent/pkg/paths"
 	"github.com/docker/docker-agent/pkg/permissions"
 	"github.com/docker/docker-agent/pkg/profiling"
@@ -372,6 +373,7 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 	}
 
 	wd, _ := os.Getwd()
+	var createdWorktree *worktree.Worktree
 	if f.worktree {
 		name := f.worktreeName
 		if name == worktreeAutoName {
@@ -388,6 +390,7 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 				return err
 			}
 		}
+		createdWorktree = wt
 		wd = wt.Dir
 		f.runConfig.WorkingDir = wt.Dir
 		out.Println("Using git worktree: " + wt.Dir + " (branch " + wt.Branch + ")")
@@ -408,6 +411,9 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 	defer cleanup()
 
 	if !useTUI {
+		// Non-interactive (--exec) runs never clean up the worktree: there
+		// is no safe moment to prompt, and silently discarding work would
+		// be surprising. The worktree is left in place for later inspection.
 		return f.handleExecMode(ctx, out, rt, sess, args)
 	}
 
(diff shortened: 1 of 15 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26683816441).

pkg/worktree/worktree.go
```diff
@@ -0,0 +1,322 @@
...
+// removal is appropriate (e.g. only for worktrees they created, never
+// pre-existing ones).
+func (wt *Worktree) Remove(ctx context.Context) error {
+	if err := git(ctx, wt.SourceDir, "worktree", "remove", "--force", wt.Dir); err != nil {
+		return fmt.Errorf("removing git worktree: %w", err)
+	}
+	// The branch can only be deleted once the worktree no longer occupies it;
+	// -D discards unmerged commits, which is the intended "remove and forget".
+	if err := git(ctx, wt.SourceDir, "branch", "-D", wt.Branch); err != nil {
+		return fmt.Errorf("deleting worktree branch: %w", err)
```
aheritier, line 241: `Remove()` unconditionally deletes `wt.Branch` with `git branch -D`. For `--worktree` runs that branch is `worktree-<name>`, owned by this tool — fine to delete. For `--worktree-pr` runs, `wt.Branch` is the PR's actual head branch (set by `gh pr checkout`, e.g. `fix/auth`). If the user made new commits during the session and did not push them, confirming removal at the cleanup prompt silently destroys those commits: `-D` bypasses merge checks and the remote never received the work.

Suggested fix: add an `OwnsBranch bool` field to `Worktree` (true when `Create` minted the branch, false when `CreatePR` checked out an existing one) and skip the `git branch -D` step when `OwnsBranch` is false. `worktree remove --force` already removes the directory and the remote-tracking ref; only the local branch pointer needs to survive.

cmd/root/run.go
```diff
@@ -319,13 +349,37 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
...
+		wd = createdWorktree.Dir
+		f.runConfig.WorkingDir = createdWorktree.Dir
+		out.Println("Using git worktree: " + createdWorktree.Dir + " (branch " + createdWorktree.Branch + ")")
+		// loadResult is nil for the remote backend; worktrees are mutually
+		// exclusive with --remote so this is belt-and-suspenders, matching
+		// the nil-guard used for cleanup throughout this function.
+		if loadResult != nil {
+			if err := f.dispatchWorktreeCreate(ctx, out, loadResult.Team, createdWorktree); err != nil {
+				stopToolSets(loadResult.Team)
+				return err
```
aheritier, line 398: When `dispatchWorktreeCreate` returns an error (hook blocked the run), the function returns early but the worktree is already on disk at `<data-dir>/worktrees/<name>`. It is never cleaned up.

Two consequences:
1. The user gets an error message with no hint for manual cleanup.
2. The next invocation with the same name hits the `os.Stat` guard in `Create()` and fails immediately with `ErrInvalidName: worktree "X" already exists` — the retry is broken.

Fix: call `_ = createdWorktree.Remove(ctx)` before returning in this error path. No agent work has been done yet, so there is nothing to preserve.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26743926320).

gtardif (approved):

Issue #2303, opened 2026-04-01
copilot: unsupported_api_for_model


Adphi:
**Describe the bug**

> all models failed: error receiving from stream: HTTP 400: POST "https://api.githubcopilot.com/chat/completions": 400 Bad Request {"message":"model \"gpt-5.3-codex\" is not accessible via the /chat/completions                endpoint","code":"unsupported_api_for_model"}

**Version affected**

```
$ docker-agent version
docker-agent version v1.40.0
Commit: Homebrew
```

**How To Reproduce**

1. Export the GITHUB_TOKEN env variable:
    ```
    $ export GITHUB_TOKEN=$(gh auth token)
    ```
2. Select the `github-copilot/gpt-5.3-codex` model
3. Ask anything to the agent

joshbarrington:
Are you using your own defined agent `.yaml` file?

I tested this and it works if you specify the `api_type` as `openapi_response`

```
models:
  gpt-5-codex:
    provider: github-copilot
    model: gpt-5.3-codex
    provider_opts:
      api_type: openai_responses

agents:
  root:
    model: gpt-5-codex
```

liketosweep:
Hi @Adphi and @joshbarrington! I would love to pick this up. 
Building on Josh's finding, it looks like the `api_type: openai_responses` configuration is the correct path. Should the permanent fix be to update the provider's default configuration in the Go code so that `gpt-5.3-codex` automatically defaults to this API type without requiring users to write a custom `.yaml` override? 
If that sounds like the right approach, I can draft a PR to implement that logic. Could I get this assigned?

/assign

joshbarrington:
@liketosweep i actually already have a default fix for this. I'll open the MR.

liketosweep:
Awesome, thanks for taking care of it @joshbarrington Feel free to tag or link me when the PR/MR is up I already have the environment pulled down locally, so I'd be more than happy to help test the fix or review the code.

PR #2942, opened 2026-06-01
fix: github-copilot Responses API auto-selection and error preservation

### LINKED ISSUE #2471: Ability to set the `Copilot-Integration-Id` header for github-copilot provider
**What you'd like to see**

If you use a Personal Access Token (PAT) as your `GITHUB_TOKEN` and then try to use a github-copilot model, you will get a `Bad Request` response with the following error message:
`Personal Access Tokens are not supported for this endpoint`

This does not happen when the header `Copilot-Integration-Id` is set to the value `copilot-developer-cli`.

It would be nice if one could set the value of this header in the `provider_opts` object for github-copilot.

**Why you'd like to see it**

So that I can use my PAT when working with github-copilot models.

**Workarounds?**

Use a local models gateway.

### LINKED ISSUE #2885: github-copilot provder is not working
## Description

When configuring github copilot as provider and correct access token, I get the following error:
```bash
model failed: error receiving from stream: HTTP 400: POST "https://api.githubcopilot.com/chat/completions":
400 Bad Request
```

Configuration file used
```yaml
agents:
  root:
    model: github-copilot/gemini-3.1-pro-preview
    skills: true
    toolsets:
      - type: filesystem
```

dgageot:
GitHub Copilot returns HTTP 400 Bad Request when newer models like gpt-5 and Codex are sent to the legacy `/chat/completions` endpoint, since Copilot proxies the same OpenAI models but has a different set of supported endpoints. This change extends the Responses API auto-selection logic to the github-copilot provider so that models are routed to the correct endpoint based on their capabilities, not a hardcoded allow-list.

The fix extends the OpenAI client's endpoint auto-selection to the github-copilot provider, driven by `modelinfo.SupportsResponsesAPI` so new models are picked up by naming convention. It also wires `oaistream.ErrorBodyMiddleware` into the direct OpenAI client path to preserve provider error bodies—like Copilot's bare "400 Bad Request"—instead of discarding them, so users see meaningful error messages.

Comprehensive tests verify endpoint selection for different model families, explicit `api_type` override behavior, and error-body preservation. This relates to the earlier Copilot header fix #2471.

Fixes #2885

### DIFF
--- a/docs/providers/github-copilot/index.md
+++ b/docs/providers/github-copilot/index.md
@@ -97,6 +97,25 @@ models:
 Header names are matched case-insensitively, so `copilot-integration-id`
 works too.
 
+## Chat Completions vs. Responses API
+
+GitHub Copilot proxies OpenAI models behind two endpoints: the legacy
+`/chat/completions` and the newer `/responses`. Newer models (the `gpt-5`
+family, Codex variants, etc.) are only served via `/responses` and reject
+`/chat/completions` with a `400 Bad Request`. docker-agent auto-selects the
+right endpoint per model, so no configuration is needed in the common case.
+
+If you ever need to force one or the other, set `api_type` explicitly:
+
+```yaml
+models:
+  copilot:
+    provider: github-copilot
+    model: gpt-5
+    provider_opts:
+      api_type: openai_responses # or openai_chatcompletions
+```
+
 ## Custom HTTP Headers
 
 `provider_opts.http_headers` is a generic escape hatch that works for any
--- a/examples/github-copilot.yaml
+++ b/examples/github-copilot.yaml
@@ -4,7 +4,12 @@
 # but you can override it (or add any other custom header) via
 # `provider_opts.http_headers`.
 #
+# Newer Copilot models (gpt-5, Codex, ...) are only served via the Responses
+# API; docker-agent auto-selects the right endpoint per model, so no extra
+# config is needed. You can still force it with `provider_opts.api_type`.
+#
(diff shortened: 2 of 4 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26749164808).

joshbarrington:
@dgageot does this also close: https://github.com/docker/docker-agent/issues/2303 ?

aheritier (approved):
Looks good.

I reviewed the Copilot Responses API auto-selection, explicit `api_type` override behavior, direct-client error-body preservation, docs, and tests. The implementation keeps explicit `provider_opts.api_type` precedence before auto-selection, extends automatic Responses routing only to `openai` and `github-copilot`, and relies on the existing `modelinfo.SupportsResponsesAPI` capability predicate rather than adding a Copilot-specific model allow-list.

The new tests cover:
- Copilot `gpt-5` and Codex variants routing to `/responses`.
- Older/chat-only Copilot models staying on `/chat/completions`.
- Explicit `openai_chatcompletions` overriding auto-selection.
- Plain-text provider error bodies being surfaced to callers.

Security-wise, token resolution and BaseURL handling are unchanged, and Copilot header handling remains centralized through the existing sanitized header path.

CI is green for head SHA `ffa07cd5aa7eb3f7c85ea85241eec78718dcb0dd`:
- `build-and-test`: success
- `lint`: success
- `license-check`: success
- `build-image`: success
- `PR Review`: success
- `save-context`: success

No blocking findings.

PR #2943, opened 2026-06-01
fix: dedupe persistent toolset-failure notifications

### LINKED ISSUE #2884: tui: persistent toolset-failure notifications stack without dedup, masking the screen on long MCP failure streaks
## Summary

When a remote MCP toolset enters a persistent failure state (e.g. an SSE endpoint returning `toolset not started`), the TUI accumulates a new persistent warning notification on the right side of the screen on every conversation turn. After a long session the cards cover most of the viewport — each must be dismissed individually with `[x]`.

This is **distinct** from #2861 / #2866: #2861 is the heap-leak root cause of the jetsam kill, this one is a UX bug that surfaces every time MCP fails repeatedly. They can both bite the same session (mine did).

## Reproduction

1. Start a session with at least one remote MCP whose endpoint will keep failing on `Tools()` — e.g. an `sse` server that has accepted Initialize but later returns `lifecycle.ErrNotStarted` on every list, or a transport that flaps.
2. Have a few normal conversation turns.
3. Each turn appends a new persistent `Some toolsets failed to initialize for agent '<name>'.\n\nDetails:\n\n- mcp(remote host=… transport=sse) list failed: toolset not started [x]` card.

Observed across three of my real sessions (screenshots in the linked DM): 5–7 identical stacked cards, several with two repeated lines inside the same card, after a few hours of activity.

## Root cause

Two cooperating gaps:

**1. No once-per-streak guard on the "list failed" path** — `pkg/agent/agent.go:321`:

```go
ta, err := toolSet.Tools(ctx)
if err != nil {
    desc := tools.DescribeToolSet(toolSet)
    slog.WarnContext(ctx, "Toolset listing failed; skipping", ...)
    a.AddToolWarning(fmt.Sprintf("%s list failed: %v", desc, err))
    continue
}
```

The "start failed" path at `agent.go:361` and `runtime.go:1152` correctly gates emission through `StartableToolSet.ShouldReportFailure()` (which returns `true` exactly once per failure streak). The "list failed" path skips this guard, so every iteration of `collectTools` re-emits a fresh warning for the same underlying problem.

**2. No dedup in the notification manager** — `pkg/tui/components/notification/notification.go:131`:

```go
case ShowMsg:
    id := nextID.Add(1)
    ...
    item := notificationItem{ID: id, Text: msg.Text, Type: notifType}
    n.items = append([]notificationItem{item}, n.items...)
```

`ShowMsg` is appended unconditionally; persistent items (`TypeWarning`/`TypeError`) never auto-expire (`persistent()` returns `true`, no `tea.Tick` scheduled), so identical cards stack until the user clicks each `[x]`.

The combination produces N copies of the same warning per failing toolset per session.

## Proposed fix

Either side fixes the symptom, both together fix it cleanly:

- **Agent side:** route `collectTools` errors through the same once-per-streak guard the start path uses. `StartableToolSet` already tracks `pendingWarning` for `Start()` failures; extend it (or add a sibling `pendingListWarning`) so a repeated `Tools()` failure for an already-started-but-now-broken toolset only surfaces once until it recovers.
- **Notification side:** in `Manager.Update`'s `ShowMsg` case, if a persistent notification with identical `Text` is already present, drop the new one (or bump a counter "× N" appended to the existing card). Cheap, defensive, and useful for any future caller that emits duplicate warnings.

I'd recommend doing both: the agent fix is the right primary, the notification fix is a safety net.

## Repro environment

- macOS, Apple silicon, 64 GB
- `docker-agent` HEAD as of 2026-05-22
- Multi-agent config with several remote MCPs (Notion SSE, an internal streamable MCP)
- Sessions of several hours; warnings observed across multiple agents (`mark_iv`, `root`)

## Related

- #2861 — heap leak per assistant message (memory growth → jetsam kill). Independent root cause; this issue is purely the TUI surface area for repeated MCP failures.

dgageot:
Toolset initialization and Tools() listing failures were stacking persistent notifications in the TUI when a toolset remained down across multiple conversation turns. The root cause was that the agent side had no mechanism to suppress duplicate warnings, so the TUI received the same error on every turn and piled them on-screen.

This fix adds a once-per-streak guard to the StartableToolSet so a failure warning is emitted only once until the streak clears (either via a successful Start/Tools or an explicit Stop). The failureStreak type encapsulates the streak-tracking logic and is reused for both Start and Tools listing failures, keeping the warning deduplication uniform and reliable across all callers—including non-TUI consumers like ACP, which persist warnings into the transcript.

The TUI-layer dedup logic is removed as it becomes redundant; the agent-side guard is the single source of truth for this behavior.

Closes #2884.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -317,8 +317,16 @@ func (a *Agent) collectTools(ctx context.Context) ([]tools.Tool, error) {
 		ta, err := toolSet.Tools(ctx)
 		if err != nil {
 			desc := tools.DescribeToolSet(toolSet)
-			slog.WarnContext(ctx, "Toolset listing failed; skipping", "agent", a.Name(), "toolset", desc, "error", err)
-			a.AddToolWarning(fmt.Sprintf("%s list failed: %v", desc, err))
+			// Route through the once-per-streak guard so a toolset stuck
+			// returning an error (e.g. a remote MCP server replying
+			// "toolset not started") surfaces a single warning per streak
+			// instead of one on every conversation turn.
+			if toolSet.ShouldReportListFailure() {
+				slog.WarnContext(ctx, "Toolset listing failed; skipping", "agent", a.Name(), "toolset", desc, "error", err)
+				a.AddToolWarning(fmt.Sprintf("%s list failed: %v", desc, err))
+			} else {
+				slog.DebugContext(ctx, "Toolset listing still failing; retrying next turn", "agent", a.Name(), "toolset", desc, "error", err)
+			}
 			continue
 		}
 		agentTools = append(agentTools, ta...)
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -126,6 +126,34 @@ func TestAgentTools(t *testing.T) {
 	}
 }
 
+// TestAgentNoDuplicateListWarnings verifies that a toolset that starts
+// successfully but keeps failing to list its tools (e.g. a remote MCP server
+// stuck returning "toolset not started") surfaces only one warning per
+// failure streak, not one on every turn.
+func TestAgentNoDuplicateListWarnings(t *testing.T) {
+	t.Parallel()
+
+	stub := newStubToolSet(nil, nil, errors.New("toolset not started"))
+	a := New("root", "test", WithToolSets(stub))
+
+	// Turn 1: first list failure → warning.
+	_, err := a.Tools(t.Context())
+	require.NoError(t, err)
+	warnings := a.DrainWarnings()
+	require.Len(t, warnings, 1, "turn 1: exactly one warning on first list failure")
+	assert.Contains(t, warnings[0], "list failed")
+
+	// Turn 2: repeated list failure → no new warning.
+	_, err = a.Tools(t.Context())
+	require.NoError(t, err)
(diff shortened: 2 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed `pkg/tools/startable.go`, `pkg/tools/startable_test.go`, `pkg/agent/agent.go`, and `pkg/agent/agent_test.go`.

**Summary**: The `failureStreak` helper is correctly implemented — `fail()` only sets `active`/`pending` on the first call of a streak, `reset()` clears both fields, and `shouldReport()` returns `true` exactly once per streak. Mutex discipline is sound across `Start`, `Stop`, `ShouldReportFailure`, and `ShouldReportListFailure`. The `Tools()` method calls the inner toolset outside the lock (correct — avoids holding the lock during a potentially slow RPC) and updates the streak under the lock before any caller can observe it in the sequential `collectTools` loop. The `listFlappyToolSet` test helper correctly increments `callIdx` unconditionally before the nil check, so error sequences are consumed faithfully. Test coverage addresses the once-per-streak guarantee, recovery/reset behavior, and the `Stop` path. No bugs found in the changed code.

aheritier (approved):
Approved. CI green across all checks; changes are sound with appropriate test coverage for the de-duplicated list warnings and failure-streak reset behavior. No blocking concerns.

PR #2946, opened 2026-06-01
feat: support inline skills in agent YAML config

dgageot:
Adds support for defining "inline skills" directly in the agent YAML `skills:` field, enabling users to express skills without maintaining separate files or requiring external sources. Inline skills are defined as YAML mappings mixed with existing string sources and name filters, and support the YAML-expressible subset of the skill definition format.

An inline skill specifies a name, description, and instructions (all required), plus optional fields for context (only "fork" allowed), model, and allowed_tools. They cannot bundle files or use command expansion. Inline skills are always exposed to the agent, work in sandbox mode since they travel inside the agent YAML, and enable skills independently without needing to pull in local sources.

The implementation includes schema updates, validation (required fields, duplicate-name detection, context restrictions), a new InlineSkill type with mixed YAML/JSON parsing, and updates to the skills toolset and TUI to handle inline skill retrieval and display.

Also includes a fix for test hermeticity: `pkg/worktree/worktree_test.go` now disables commit signing in the test bootstrap repo to prevent leakage from the developer's global 1Password SSH commit-signing config, which was flaky in tests.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -706,7 +706,7 @@
           "description": "Optional response cache: when the same user question is asked again, replay the previous answer instead of calling the model."
         },
         "skills": {
-          "description": "Enable skills discovery for this agent. Set to true to load all discovered skills from local filesystem sources; false disables skills. A list can mix sources (\"local\" or an HTTP/HTTPS URL) and/or skill names to include. If only names are given, local sources are loaded and filtered to just those skills.",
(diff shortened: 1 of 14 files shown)


rumpl (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26752480522).

PR #2947, opened 2026-06-01
fix: keep failed chat continuations out of conversation cache

### LINKED ISSUE #2890: chatserver: failed X-Conversation-Id turns mutate cached session and affect retries
### Summary

When `docker agent serve chat` is used with `X-Conversation-Id`, continuation requests mutate the cached `*session.Session` before execution. If the run later fails (timeout, 500, stream error, etc.), the mutated session remains the canonical conversation state, so a retry can see duplicated or half-failed history.

### Why this looks wrong

On current `main` (`0aa4436d`), the continuation path is:

- `pkg/chatserver/server.go`: `handleChatCompletions` calls `resolveSession(...)`
- `pkg/chatserver/conversations.go`: `Get(...)` returns the live cached `*session.Session`
- `pkg/chatserver/server.go`: `resolveSession(...)` reuses that pointer and calls `appendLatestUser(existing, msgs)`
- `pkg/chatserver/agent.go`: `appendLatestUser(...)` mutates the session in place before execution
- `pkg/chatserver/server.go`: after the run returns, `maybeStoreConversation(...)` is still called

That means the conversation has already advanced before the request outcome is known, and there is no rollback path if the run fails.

### History context

This appears to date back to the feature introduction:

- `d0c9985f` introduced `X-Conversation-Id` and the in-memory conversation cache
- `2ee80aff` later changed the logic to always store the conversation after request completion so evicted conversations are restored
- `047167a4` then simplified away the old `isNew` flag

The eviction fix makes sense on its own, but it also means failed turns are reinserted after eviction as well. More importantly, even without `Put(...)`, the cached session pointer has already been mutated in place.

### Repro shape

1. Start `serve chat` with `--conversations-max` enabled.
2. Send a request with `X-Conversation-Id: conv-1` and a valid user message.
3. Send a second request with the same conversation id and a new user message, but force the run to fail after the message has been appended to the session (for example via provider timeout or an injected runtime error).
4. Retry that second request with the same conversation id.
5. The retry now runs against a session that already contains the failed turn, so history is no longer equivalent to "retry the same request against the last successful conversation state".

Depending on where the failure happens, this can show up as duplicated user turns or other half-failed conversation state surviving into the next attempt.

### Expected behavior

A failed continuation request should not advance the cached conversation state.

### Possible fixes

A couple of directions that seem reasonable:

- Clone/snapshot the cached session before `appendLatestUser(...)`, and only commit it back to the cache on success.
- Or mutate a working copy during the request and leave the cached session untouched until the run succeeds.
- If streaming failures are intended to preserve some state, that likely needs an explicit policy rather than the current in-place mutation behavior.

### Tests that seem worth adding

I could not find coverage for either of these cases:

- failed continuation requests do not advance cached history
- continuation requests with no new user message are rejected instead of reusing the prior session state

There is coverage for eviction restore, `appendLatestUser`, and same-conversation locking, but not for transactional behavior on failure.

dgageot:
## Summary

Fixes #2890.

- Make `X-Conversation-Id` continuations transactional: work on a cloned cached session and commit it back only when the run succeeds.
- Reject continuation requests that do not include a usable new user message, instead of replaying the previous turn.
- Add `Session.Clone()` as a faithful deep-copy primitive and harden clone helpers so cloned sessions do not share mutable message/config/eval/tool-definition state.
- Add package tests and an e2e test covering failed-turn cache behavior.

## Validation

- `task lint`
- `task test`
- `task build`

### DIFF
--- a/e2e/chatserver_test.go
+++ b/e2e/chatserver_test.go
@@ -1,0 +1,306 @@
+package e2e_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/cmd/root"
+)
+
+func TestServeChatConversationFailedTurnDoesNotAdvanceCache(t *testing.T) {
+	modelServer := newRecordingChatCompletionsServer(t)
+
+	agentFile := filepath.Join(t.TempDir(), "agent.yaml")
+	agentYAML := fmt.Appendf(nil, `version: "9"
+
+providers:
+  e2e:
+    api_type: openai_chatcompletions
+    base_url: %s/v1
+
+models:
+  e2e-model:
+    provider: e2e
+    model: e2e-model
+    max_tokens: 64
+
+agents:
+  root:
+    model: e2e-model
+    description: E2E chat server agent
+    instruction: Reply concisely.
+`, modelServer.URL())
+	require.NoError(t, os.WriteFile(agentFile, agentYAML, 0o644))
+
+	addr := freeTCPAddr(t)
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	os.Unsetenv("DOCKER_CLI_PLUGIN_ORIGINAL_CLI_COMMAND")
+	t.Setenv("DOCKER_AGENT_MODELS_GATEWAY", "")
+	t.Setenv("CAGENT_MODELS_GATEWAY", "")
+	t.Setenv("OPENAI_API_KEY", "DUMMY")
+
+	var stdout, stderr bytes.Buffer
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- root.Execute(ctx, nil, &stdout, &stderr,
+			"--cache-dir", filepath.Join(t.TempDir(), "cache"),
+			"--config-dir", filepath.Join(t.TempDir(), "config"),
+			"--data-dir", filepath.Join(t.TempDir(), "data"),
+			"serve", "chat",
+			"--listen", addr,
+			"--conversations-max", "10",
+			"--request-timeout", "2s",
+			agentFile,
+		)
+	}()
+	baseURL := "http://" + addr
+	waitForChatServer(t, baseURL)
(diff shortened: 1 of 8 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly implements the transactional conversation-cache pattern. The changes were reviewed across all 8 modified files.

**What was reviewed:**
- Transactional clone logic in `pkg/chatserver/server.go` (`resolveSession` / `commitConversation`): the clone→run→commit-on-success-only flow is correct and race-free.
- Deep-copy completeness in `pkg/session/session.go` and `pkg/session/branch.go`: all mutable slice/map/pointer fields (messages, tool definitions, eval criteria, eval results, configs) are properly deep-copied; cloned sessions do not share mutable state with the original.
- Error handling for the "no usable new user message" rejection path: correctly returns 400 for empty/missing user messages rather than silently replaying the prior turn.
- Test coverage in `conversations_test.go`, `conversations_transaction_test.go`, `clone_test.go`, and `e2e/chatserver_test.go`: tests are well-structured and assert the intended behavior.
- The `return nil` in the refactored non-streaming handler path is intentional and correct — `chatCompletion` already writes the error response directly, so returning the error to Echo would cause a double-response. The tradeoff is acknowledged in the inline comment.

No confirmed bugs found in the changed code.

aheritier (approved):
LGTM. CI is green for head SHA `72b361c2a50e47925fa43a4124aefee83edd892c` (`build-and-test`, `build-image`, `license-check`, `lint`, `PR Review`, and `save-context` all succeeded; `build-and-push-image` was skipped).

I reviewed the cache/transaction semantics and concurrency path in detail:
- Continuations now work on a clone of the cached session before appending the latest user message (`pkg/chatserver/server.go:423-430`), so failed turns cannot mutate the canonical cached pointer.
- The cache is advanced only on successful runs (`pkg/chatserver/server.go:449-453`), while both streaming and non-streaming handlers pass the actual agent run error into that commit decision (`pkg/chatserver/server.go:398-407`).
- Per-conversation locking still prevents two concurrent continuations for the same `X-Conversation-Id` from racing (`pkg/chatserver/server.go:373-377`, `pkg/chatserver/conversation_lock.go:25-39`).
- The clone path deep-copies mutable session/message state rather than sharing slices/maps/pointers (`pkg/session/session.go:339-404`), with regression coverage for config, eval fields, tool definitions, sub-sessions, and item metadata.
- Test coverage includes focused transaction tests (`pkg/chatserver/conversations_transaction_test.go`) and an e2e regression proving a failed continuation is not replayed on the next successful turn (`e2e/chatserver_test.go:88-104`).

No blocking findings.

PR #2949, opened 2026-06-01
fix(mcp): stop the OAuth Authentication Request loop after the user clicks Cancel

trungutt:
## The bug

The first time the agent enables a remote MCP server that requires OAuth (e.g. `mcp.notion.com`), the host shows an `Authentication Request` dialog. **Cancel** used to be effectively useless.

There are two distinct loops that conspire to produce this. Each is harmless on its own — together they make the cancel feel broken.

### Loop A — clicking Cancel does nothing

**What the user sees.** They click **Cancel** on the Authentication Request. The dialog flips to `DENIED`. Half a second later a fresh, identical Authentication Request appears. They click **Cancel** again. Another one appears. Trying to dismiss the prompt feels broken.

**What is actually happening underneath.** A single connection attempt to the remote MCP server fans out into several parallel HTTP requests; every one of them comes back asking for OAuth. They queue up on a single-flight mutex (`oauthFlowMu`, `8bc1fa90e`) so only one OAuth dance runs at a time. The mutex only protects against *simultaneous* OAuth — it does not remember that the user just said "no". So the moment the first goroutine returns with the cancel and releases the mutex, the next queued goroutine wakes up, sees no token, decides "I'd better start an OAuth flow", and pops the same dialog again.

```mermaid
sequenceDiagram
    participant SDK as MCP SDK
    participant RT1 as RoundTrip #1
    participant RT2 as RoundTrip #2
    participant Mu as oauthFlowMu
    participant User

    SDK->>RT1: initialize-stage RPC
    SDK->>RT2: initialize-stage RPC (parallel)
    RT1->>Mu: Lock (wins)
    RT2->>Mu: Lock (queued)
    RT1->>User: elicit OAuth
    User-->>RT1: Cancel
    RT1-->>SDK: OAuthDeclinedError
    RT1->>Mu: Unlock
    Mu->>RT2: acquired
    RT2->>RT2: getValidToken() == nil
    RT2->>User: elicit OAuth (NEW dialog)
    Note over RT2,User: same dialog the user just dismissed
```

### Loop B — the model lies after the user cancels

**What the user sees.** Eventually they manage to dismiss the prompt. The agent calmly replies *"The Notion server is now enabled. On my next turn I'll have access to Notion tools. Let me proceed with your request"* — and then either makes up an answer, or asks for workspace details so it can call tools that do not exist. The user knows they cancelled; the agent is pretending nothing happened.

**What is actually happening underneath.** The catalog tells the model *"enabled X — proceed with the user's original request"* the moment the model calls `enable_remote_mcp_server`, before the OAuth dialog has even been shown. Nothing later in the conversation tells the model "by the way, the user just cancelled". On top of that, the catalog's `Tools()` only treats one specific error as a soft failure (`IsAuthorizationRequired` = "defer this"); a user cancel falls through to the generic "transient failure" branch, so the server stays enabled in the catalog state and the next loop iteration would try `Start()` again.

```mermaid
flowchart LR
    A[loop iter N] --> B[Tools]
    B --> C[Start fails<br/>decline]
    C --> D{IsAuthorizationRequired?}
    D -->|no| E[server stays in t.enabled<br/>started=false]
    E --> F[loop iter N+1]
    F --> B
```

---

### Layer 1 — `pkg/tools/mcp` — latch the decline under the mutex

**What the user sees.** Click **Cancel** once. The dialog goes away and stays away.

**What changes underneath.** Give the OAuth transport one new bit of memory: "did the user just cancel?". Set that bit *while still holding the mutex*, before any queued goroutine can wake up. Check the bit at the top of every authorization attempt; if it's set, return the cancel sentinel immediately instead of running the OAuth flow again.

The MCP SDK wraps transport errors in a way that destroys the unwrap chain (`fmt.Errorf("%w: %v", …)` — the original sits in `%v` as text, not in the chain), so `errors.As` can't find our sentinel after it bubbles up. That's the exact reason an existing flag `lastAuthRequired` already exists for the deferred-OAuth case; the new bit is a parallel of that, and `enrichConnectError` in `remote.go` reconstitutes a clean `*OAuthDeclinedError` for callers.

```
oauthTransport {
    + lastOAuthDeclined bool       // NEW
}

authorizeOnce(ctx) error {
    Lock(oauthFlowMu); defer Unlock

    if validToken { return nil }

    if lastOAuthDeclined {                     // NEW short-circuit
        return &OAuthDeclinedError{URL}
    }

    err = handleOAuthFlow(ctx)
    if IsOAuthDeclined(err) {                  // NEW latch
        lastOAuthDeclined = true               // BEFORE deferred Unlock
    }
    return err
}
```

The queued goroutine now sees the latched cancel and never opens a fresh dialog:

```mermaid
sequenceDiagram
    participant RT1 as RoundTrip #1
    participant RT2 as RoundTrip #2
    participant Mu as oauthFlowMu
    participant User

    RT1->>Mu: Lock
    RT2->>Mu: Lock (queued)
    RT1->>User: elicit OAuth
    User-->>RT1: Cancel
    RT1->>RT1: lastOAuthDeclined = true
    RT1->>Mu: Unlock
    Mu->>RT2: acquired
    RT2->>RT2: lastOAuthDeclined? YES
    RT2-->>RT2: return OAuthDeclinedError
    Note over RT2,User: NO new dialog
```

### Layer 2 — `pkg/tools/builtin/mcpcatalog` — drop the server and stop lying

**What the user sees.** After the cancel, the agent replies honestly — *"the request needs you to authorize the connection — want to retry?"* — instead of claiming Notion is connected. If the user re-asks the same thing (or says "yes" / "retry"), a fresh authorization dialog appears.

**What changes underneath.** Two small things:

1. *Recognise the cancel as a distinct outcome.* The catalog's `Tools()` already special-cases `IsAuthorizationRequired`; we add a sibling case for `IsOAuthDeclined`. When a server's `Start()` fails because the user dismissed the dialog, we delete the server from `t.enabled`, stop the inner toolset (so any half-initialised session is cleaned up), and fire `toolsChangedHandler` exactly once. Concurrent `handleDisable` / `handleResetAuth` / re-enable races are handled by only doing the eviction when *we* still own the wrapper. After this, the next agent loop iteration enumerates a smaller tool list and does NOT retry `Start()` — the server is simply gone until somebody calls `enable_remote_mcp_server` again.

2. *Rewrite the `handleEnable` tool result so the model can't lie.* The old message was a flat promise. The new message is a conditional self-check the model can ground its next turn against: *"if `<id>_` tools appear, proceed; if they don't, the user dismissed the dialog — be honest, and on user re-ask call `enable_remote_mcp_server` again to surface a fresh dialog"*.

```mermaid
flowchart TD
    A[Tools] --> B[Start fails]
    B --> C{IsAuthorizationRequired?}
    C -->|yes| D[continue silently]
    C -->|no| E{IsOAuthDeclined?}
    E -->|yes - NEW| F[disableAfterDecline:<br/>delete from t.enabled<br/>Stop inner toolset<br/>fire toolsChangedHandler once]
    E -->|no| G[log + continue<br/>retry next iter]
```

```
old: "enabled <id> — the server's tools will appear on your next
      turn. Proceed with the user's original request; the host
      handles any required connection setup."

new: "enable requested for <id>. The host will surface any required
      authorization dialog. On your next turn:
        - if tools whose names start with <id>_ appear, proceed
          with the user's original request using them.
        - if NO such tools appear, the user dismissed the dialog.
          Tell the user honestly; on user re-ask (or yes/retry),
          call enable_remote_mcp_server again for a fresh dialog."
```

No registries, no TTLs, no visibility tweaks. The transport latch handles intra-`Initialize` concurrency; every cross-turn enable starts with a fresh `mcp.Toolset`, so nothing carries over.

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
+++ b/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
@@ -480,6 +480,14 @@ func (t *Toolset) Tools(ctx context.Context) ([]tools.Tool, error) {
 
 		if !e.ts.IsStarted() {
 			if err := e.ts.Start(ctx); err != nil {
+				// Diagnostic breadcrumb so the deferred-vs-declined
+				// classification is visible in --debug logs without
+				// having to instrument the OAuth transport itself.
+				slog.DebugContext(ctx, "Enabled remote MCP server failed to start",
+					"id", e.id,
+					"auth_required", mcp.IsAuthorizationRequired(err),
+					"oauth_declined", mcp.IsOAuthDeclined(err),
+					"error", err)
 				// Auth-required is an *expected* deferral when probing
 				// with a non-interactive context (startup tool count) or
 				// when the elicitation bridge is not yet ready. Silent
@@ -488,6 +496,18 @@ func (t *Toolset) Tools(ctx context.Context) ([]tools.Tool, error) {
 				if mcp.IsAuthorizationRequired(err) {
 					slog.DebugContext(ctx, "Remote MCP server requires authorization; deferred to next turn",
 						"id", e.id)
+					continue
+				}
+				// User explicitly dismissed the host's authorization
+				// dialog (clicked "Cancel" / "Decline"). Treat that as
+				// "deactivate this server" so the very next Tools()
+				// call does not re-fire the OAuth flow and re-pop the
+				// dialog the user just closed. The model can re-enable
+				// the server explicitly if the user changes their mind.
+				if mcp.IsOAuthDeclined(err) {
+					slog.InfoContext(ctx, "Remote MCP server OAuth declined by user; removing from enabled set",
+						"id", e.id)
+					t.disableAfterDecline(ctx, e.id, e.ts)
 					continue
 				}
 				// Real failure: log once per streak (StartableToolSet
@@ -703,7 +723,9 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
 	}
 
 	msg := strings.Builder{}
-	fmt.Fprintf(&msg, "enabled %q (%s) — the server's tools will appear on your next turn. Proceed with the user's original request; the host handles any required connection setup.\n", id, server.Title)
+	fmt.Fprintf(&msg, "enable requested for %q (%s). The host will surface any required authorization dialog. On your next turn:\n", id, server.Title)
+	fmt.Fprintf(&msg, "  - if tools whose names start with %q appear in your available tools, proceed with the user's original request using them.\n", id+"_")
+	fmt.Fprintf(&msg, "  - if NO such tools appear, the user dismissed the authorization dialog. Do NOT claim the server is connected and do NOT ask the user for workspace details to use \"the available tools\" (there are none). Tell the user the request needs them to authorize the connection. If the user then re-asks for the same thing (or says \"yes\", \"retry\", \"try again\", etc.), call %s for this id again to surface a fresh authorization dialog.\n", ToolNameEnable)
 	fmt.Fprintf(&msg, "endpoint: %s\n", server.URL)
 	// Only surface unresolved env vars — the model can act on those by
 	// asking the user to set the variable and retrying. Every other auth
@@ -794,6 +816,43 @@ func unresolvedHeaderEnvs(headers map[string]string) []string {
 // DisableArgs is the input schema for disable_remote_mcp_server.
 type DisableArgs struct {
 	ID string `json:"id" jsonschema:"Catalog id of the server to disable."`
+}
+
+// disableAfterDecline removes a server from the enabled set after the user
+// declined its OAuth flow, mirroring handleDisable's lifecycle invariants
+// (delete from t.enabled, stop the inner toolset, notify the runtime that
+// the tool surface changed) but without producing a model-facing tool
+// result — the decline happens asynchronously inside Tools(), not in
+// response to a model tool call.
+//
+// The guard (current == wrapped) protects against a concurrent
+// handleDisable / handleResetAuth / handleEnable that already swapped or
+// removed the entry between the failing Start() and this cleanup.
+func (t *Toolset) disableAfterDecline(ctx context.Context, id string, wrapped *tools.StartableToolSet) {
+	t.mu.Lock()
+	current, stillEnabled := t.enabled[id]
+	if stillEnabled && current == wrapped {
+		delete(t.enabled, id)
+	} else {
+		// Already gone (or replaced by a fresh enable) — let whoever
+		// owns the live wrapper handle its lifecycle. We still want
+		// to stop OUR reference because Start() may have left the
+		// session in a partially-initialised state.
+		stillEnabled = false
+	}
+	notify := t.toolsChangedHandler
+	t.mu.Unlock()
+
+	if err := wrapped.Stop(ctx); err != nil && !errors.Is(err, context.Canceled) {
+		slog.DebugContext(ctx, "Failed to stop declined-OAuth remote MCP toolset",
+			"id", id, "error", err)
+	}
+
+	// Only notify when WE removed the entry; if a concurrent caller
+	// already mutated t.enabled, they will have notified themselves.
+	if stillEnabled && notify != nil {
+		notify()
+	}
 }
 
 func (t *Toolset) handleDisable(ctx context.Context, args DisableArgs) (*tools.ToolCallResult, error) {
--- a/pkg/tools/mcp/oauth.go
+++ b/pkg/tools/mcp/oauth.go
@@ -302,6 +302,15 @@ type oauthTransport struct {
 	// instead of unwrapping to know that OAuth was deferred rather than
 	// failed for some other reason.
 	lastAuthRequired bool
+	// lastOAuthDeclined records when the user explicitly declined or
+	// cancelled an interactive OAuth flow (clicked "Cancel" on the host's
+	// Authentication Request dialog). Same rationale as lastAuthRequired:
+	// the MCP SDK wraps transport errors with %v before they reach our
+	// callers, so errors.As cannot reliably recover the underlying
+	// *OAuthDeclinedError. enrichConnectError reads this flag to
+	// reconstitute a clean sentinel for the catalog's retry-loop short
+	// circuit (see mcpcatalog.Toolset.Tools / disableAfterDecline).
+	lastOAuthDeclined bool
 }
 
 func (t *oauthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
@@ -323,7 +332,41 @@ func (t *oauthTransport) authorizeOnce(ctx context.Context, authServer, wwwAuth 
 		return nil
 	}
 
-	return t.handleOAuthFlow(ctx, authServer, wwwAuth)
+	// Sticky decline: the MCP SDK's Connect() runs several
+	// initialize-stage RPCs concurrently. Each one that gets a 401
+	// queues here on oauthFlowMu. Without this short-circuit, the
+	// goroutine that wins the mutex runs the full OAuth flow; when
+	// the user clicks Cancel, this goroutine returns OAuthDeclinedError
+	// and releases the mutex — at which point the NEXT queued
+	// goroutine sees no valid token and fires a fresh OAuth flow,
+	// re-popping the dialog the user just dismissed. Latching the
+	// decline state under the same mutex makes the queued callers
+	// see the prior decline before they can start a new flow.
+	t.mu.Lock()
+	declined := t.lastOAuthDeclined
+	t.mu.Unlock()
+	if declined {
+		slog.DebugContext(ctx, "OAuth flow short-circuited: user already declined on this transport", "url", t.baseURL)
+		return &OAuthDeclinedError{URL: t.baseURL}
+	}
+
+	err := t.handleOAuthFlow(ctx, authServer, wwwAuth)
+	if err != nil {
+		// Latch the decline state BEFORE the deferred Unlock fires so
+		// any goroutine queued on oauthFlowMu observes it on its next
+		// iteration of the getValidToken / declined / handleOAuthFlow
+		// dance. Setting this in roundTrip (after we return) would
+		// race: the queued goroutine would acquire the mutex first
+		// and start a fresh flow while we are still bubbling the
+		// error up the stack.
+		var declinedErr *OAuthDeclinedError
+		if errors.As(err, &declinedErr) {
+			t.mu.Lock()
+			t.lastOAuthDeclined = true
+			t.mu.Unlock()
+		}
+	}
+	return err
 }
 
 func (t *oauthTransport) roundTrip(req *http.Request, isRetry bool) (*http.Response, error) {
@@ -391,6 +434,24 @@ func (t *oauthTransport) roundTrip(req *http.Request, isRetry bool) (*http.Respo
 					t.lastAuthRequired = true
 					t.mu.Unlock()
 					return nil, authErr
+				}
+				// User-driven decline of the host's Authentication
+				// Request dialog. Flag it explicitly so callers can
+				// recover the sentinel through enrichConnectError —
+				// the SDK's transport-error wrapping (fmt.Errorf
+				// "%w: %v") otherwise destroys the unwrap chain. See
+				// remote.go enrichConnectError + the lastAuthRequired
+				// pattern this mirrors.
+				var declinedErr *OAuthDeclinedError
+				if errors.As(err, &declinedErr) {
+					slog.Debug("OAuth flow declined by user", "url", t.baseURL)
+					if declinedErr.URL == "" {
+						declinedErr.URL = t.baseURL
+					}
+					t.mu.Lock()
+					t.lastOAuthDeclined = true
+					t.mu.Unlock()
+					return nil, declinedErr
 				}
 				return nil, fmt.Errorf("OAuth flow failed: %w", err)
 			}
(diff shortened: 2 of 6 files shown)


pkg/tools/mcp/oauth.go
```diff
@@ -323,7 +332,41 @@ func (t *oauthTransport) authorizeOnce(ctx context.Context, authServer, wwwAuth
...
+	if err != nil {
+		// Latch the decline state BEFORE the deferred Unlock fires so
+		// any goroutine queued on oauthFlowMu observes it on its next
+		// iteration of the getValidToken / declined / handleOAuthFlow
+		// dance. Setting this in roundTrip (after we return) would
+		// race: the queued goroutine would acquire the mutex first
+		// and start a fresh flow while we are still bubbling the
+		// error up the stack.
+		var declinedErr *OAuthDeclinedError
+		if errors.As(err, &declinedErr) {
```
docker-agent, line 363: **[HIGH] `handleManagedOAuthFlow` returns a plain error on user decline — the new `OAuthDeclinedError` latch is bypassed for the primary production code path**

The PR fixes `handleUnmanagedOAuthFlow` to return `&OAuthDeclinedError{URL: t.baseURL}` but leaves `handleManagedOAuthFlow` returning a plain `errors.New("user declined OAuth authorization")` (line 852).

`LocalRuntime` is constructed with `managedOAuth: true` (`pkg/runtime/runtime.go:499`), so the managed path is the default for the Docker Desktop agent. When the managed flow's elicitation is declined:

1. `errors.As(err, &declinedErr)` in `authorizeOnce` **does not match** → `lastOAuthDeclined` is never latched → subsequent queued goroutines see no valid token and each fires a fresh OAuth flow, re-popping the dialog the user just dismissed (the exact intra-connect loop the PR is supposed to fix)
2. `enrichConnectError`'s `t.oauthDeclined()` returns `false` → no sentinel reconstitution → generic error path
3. `mcp.IsOAuthDeclined(err)` in `mcpcatalog.Tools()` returns `false` → `disableAfterDecline` is never called → the cross-turn retry loop also persists

The fix is the same pattern already applied to the unmanaged path:

```go
// in handleManagedOAuthFlow, around line 851
if result.Action != tools.ElicitationActionAccept {
-   return errors.New("user declined OAuth authorization")
+   return &OAuthDeclinedError{URL: t.baseURL}
}
```

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -796,6 +818,43 @@ type DisableArgs struct {
...
+		// Already gone (or replaced by a fresh enable) — let whoever
+		// owns the live wrapper handle its lifecycle. We still want
+		// to stop OUR reference because Start() may have left the
+		// session in a partially-initialised state.
+		stillEnabled = false
+	}
+	notify := t.toolsChangedHandler
+	t.mu.Unlock()
+
+	if err := wrapped.Stop(ctx); err != nil && !errors.Is(err, context.Canceled) {
```
docker-agent, line 846: **[LOW] `disableAfterDecline` suppresses `context.Canceled` but not `context.DeadlineExceeded` from `Stop`**

The error filter in `disableAfterDecline`:
```go
if err := wrapped.Stop(ctx); err != nil && !errors.Is(err, context.Canceled) {
    slog.DebugContext(ctx, "Failed to stop declined-OAuth remote MCP toolset", ...)
}
```
only swallows `context.Canceled`. If the context has a deadline that expires while `Stop` is running, `context.DeadlineExceeded` produces a spurious debug log. The same pattern appears in `handleDisable` (line 871) and `handleResetAuth` (line 949) — for consistency consider using `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)` (or a shared helper).

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR correctly fixes the unmanaged OAuth cancel loop (Notion and similar servers that use the browser-redirect flow). However, there is a high-severity gap: **the primary production code path (managed OAuth, the default in ) still re-pops the dialog after the user clicks Cancel** because  was not updated to return . Both the intra-connect latch and the cross-turn  eviction depend on recognising that sentinel type — and they cannot recognise a plain .

The unmanaged-path tests added in this PR all pass because the unmanaged path was fixed; there are no tests for the managed-path decline, so the gap goes undetected by CI.

One additional low-severity observation about 's error filtering is noted inline.

pkg/tools/mcp/oauth.go
docker-agent, line 363: **[HIGH] `handleManagedOAuthFlow` returns a plain error on user decline — the new `OAuthDeclinedError` latch is bypassed for the primary production code path**

The PR fixes `handleUnmanagedOAuthFlow` to return `&OAuthDeclinedError{URL: t.baseURL}` but leaves `handleManagedOAuthFlow` returning a plain `errors.New("user declined OAuth authorization")` (line 852).

`LocalRuntime` is constructed with `managedOAuth: true` (`pkg/runtime/runtime.go:499`), so the managed path is the default for the Docker Desktop agent. When the managed flow's elicitation is declined:

1. `errors.As(err, &declinedErr)` in `authorizeOnce` **does not match** → `lastOAuthDeclined` is never latched → subsequent queued goroutines see no valid token and each fires a fresh OAuth flow, re-popping the dialog the user just dismissed (the exact intra-connect loop the PR is supposed to fix)
2. `enrichConnectError`'s `t.oauthDeclined()` returns `false` → no sentinel reconstitution → generic error path
3. `mcp.IsOAuthDeclined(err)` in `mcpcatalog.Tools()` returns `false` → `disableAfterDecline` is never called → the cross-turn retry loop also persists

The fix is the same pattern already applied to the unmanaged path:

```go
// in handleManagedOAuthFlow, around line 851
if result.Action != tools.ElicitationActionAccept {
-   return errors.New("user declined OAuth authorization")
+   return &OAuthDeclinedError{URL: t.baseURL}
}
```

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
docker-agent, line 846: **[LOW] `disableAfterDecline` suppresses `context.Canceled` but not `context.DeadlineExceeded` from `Stop`**

The error filter in `disableAfterDecline`:
```go
if err := wrapped.Stop(ctx); err != nil && !errors.Is(err, context.Canceled) {
    slog.DebugContext(ctx, "Failed to stop declined-OAuth remote MCP toolset", ...)
}
```
only swallows `context.Canceled`. If the context has a deadline that expires while `Stop` is running, `context.DeadlineExceeded` produces a spurious debug log. The same pattern appears in `handleDisable` (line 871) and `handleResetAuth` (line 949) — for consistency consider using `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)` (or a shared helper).

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR correctly fixes the unmanaged OAuth cancel loop (Notion and similar servers that use the browser-redirect flow). However, there is a high-severity gap: **the primary production code path (managed OAuth, the default in `LocalRuntime`) still re-pops the dialog after the user clicks Cancel** because `handleManagedOAuthFlow` was not updated to return `*OAuthDeclinedError`. Both the intra-connect latch and the cross-turn `disableAfterDecline` eviction depend on recognising that sentinel type — and they cannot recognise a plain `errors.New(...)`.

The unmanaged-path tests added in this PR all pass because the unmanaged path was fixed; there are no tests for the managed-path decline, so the gap goes undetected by CI.

One additional low-severity observation about `disableAfterDecline`'s error filtering is noted inline.

pkg/tools/mcp/oauth.go
trungutt: Intentionally out of scope.

The deployment this PR targets uses the **unmanaged** OAuth path exclusively. The managed (elicitation-driven) flow is configurable per-toolset via `SetManagedOAuth`; on this deployment it is never enabled, so `handleManagedOAuthFlow` is dead code for our use case and dragging it into a behavioural change here would mean shipping (and testing) code we do not actually exercise.

If a managed-OAuth deployment hits the same bug in the future, the same sentinel pattern applies trivially — one line in `handleManagedOAuthFlow`, mirroring the unmanaged change — and can land as a focused follow-up.

trungutt: See [previous reply](https://github.com/docker/docker-agent/pull/2949#discussion_r3334964656) — duplicate of the same observation; intentionally out of scope (this PR targets the unmanaged path only).

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
trungutt: Intentionally left alone for consistency.

The error filter mirrors the existing pattern already used by `handleDisable` (around line 871) and `handleResetAuth` (around line 949). Changing `disableAfterDecline` in isolation to also swallow `context.DeadlineExceeded` would introduce a behavioural asymmetry between the three callers that all share the same "Stop on the way out, do not yell on caller-induced cancellation" intent.

If the broader pattern should accept `DeadlineExceeded` too, that is a focused follow-up touching all three callsites (and arguably a shared helper), not an asymmetric tweak in this PR.

trungutt: See [previous reply](https://github.com/docker/docker-agent/pull/2949#discussion_r3334923603) — duplicate of the same observation.

trungutt:
Thanks for the review. Summary of dispositions:

| Finding | Disposition |
|---|---|
| 🔴 [HIGH] `handleManagedOAuthFlow` missing sentinel | Intentionally out of scope — see [inline reply](https://github.com/docker/docker-agent/pull/2949#discussion_r3334964656). This PR targets the unmanaged path exclusively (the deployment we ship does not use managed OAuth). |
| 🟡 [LOW] `disableAfterDecline` does not swallow `DeadlineExceeded` | Intentionally left alone for consistency — see [inline reply](https://github.com/docker/docker-agent/pull/2949#discussion_r3334966471). The pattern matches the two sibling callers (`handleDisable` / `handleResetAuth`); changing one in isolation would be worse than leaving them aligned. |

No code changes from this review pass. Happy to revisit either point in a focused follow-up if needed.

pkg/tools/mcp/oauth.go
docker-agent: I appreciate the clarification — if your deployment never enables managed OAuth, there's no runtime risk for this PR. That's a fair scoping decision.

One suggestion: could you add a short comment in `handleManagedOAuthFlow` near line 851 referencing the pattern in `handleUnmanagedOAuthFlow`? Something like:

```go
// TODO: if managed OAuth is enabled in production, return &OAuthDeclinedError{URL: t.baseURL}
// here (see handleUnmanagedOAuthFlow) to prevent retry loops.
```

That way the next engineer who touches the managed path (or enables it) sees the pattern immediately. But if that feels like clutter for genuinely dead code, no strong objection — the unmanaged fix is the one that matters for your use case.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
docker-agent: You're absolutely right — I should have framed this as a consistency issue across all three sites rather than singling out `disableAfterDecline`. Making the change in isolation would indeed create an asymmetry, and your reasoning about "caller-induced cancellation" applying equally to both `Canceled` and `DeadlineExceeded` makes sense.

A follow-up PR that introduces a small helper (e.g., `isContextDone(err) bool`) and updates all three callsites would be cleaner. That said, if the current pattern hasn't caused issues in practice, it's a low-priority polish item rather than a bug. Thanks for the context!

docker-agent: Thanks for the pointer! I see you've already addressed this in the other thread — agreed that changing all three callsites together (ideally with a shared helper) makes more sense than creating asymmetry here. Closing this duplicate.

pkg/tools/mcp/oauth.go
docker-agent: You're right — my apologies for the noise. I missed that the HTTP server overrides the runtime's default with WithManagedOAuth(false) in session_manager.go, making handleManagedOAuthFlow dead code for this deployment. The one-line fix I suggested does apply if a managed-OAuth deployment hits the same bug in the future, but as you said, shipping and testing code that isn't actually exercised here would dilute the PR's focus. Thanks for the clarification.

dgageot (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26762094905).

aheritier (approved):
LGTM. CI is green for head `95a26196279cd8496dd2c6aa4dca84f0558093c2`.

I reviewed the OAuth cancel loop fix across the transport and catalog layers. The transport now latches a user decline while still under `oauthFlowMu`, so queued concurrent initialize requests short-circuit instead of opening a fresh auth dialog (`pkg/tools/mcp/oauth.go:327-369`). It also reconstitutes a clean `OAuthDeclinedError` after the MCP SDK wrapping path via `enrichConnectError` (`pkg/tools/mcp/remote.go:129-138`).

The catalog side breaks the cross-turn retry loop by treating `mcp.IsOAuthDeclined(err)` as a deactivate outcome, removing the server from `enabled`, stopping the wrapped toolset, and notifying only when that wrapper is still the live entry (`pkg/tools/builtin/mcpcatalog/mcpcatalog.go:481-512`, `pkg/tools/builtin/mcpcatalog/mcpcatalog.go:831-855`). The model-facing enable message is also safer: it tells the model to verify `<id>_` tools are actually present before proceeding, and to be honest if the dialog was dismissed (`pkg/tools/builtin/mcpcatalog/mcpcatalog.go:725-729`).

Tests cover the important regression paths: declined unmanaged elicitation returns the sentinel without hitting the token endpoint, concurrent RoundTrips emit exactly one elicitation, catalog cleanup removes/stops/notifies once, and stale-wrapper cleanup does not evict a replacement entry.

Non-blocking note: the managed OAuth path still appears to return a generic "user declined OAuth authorization" on decline, while this PR changes the unmanaged path. Since the PR/test scenario targets the catalog/API-host unmanaged flow and runtime API server mode opts into unmanaged OAuth, I don't think this should block, but it may be worth a follow-up if managed clients can hit the same loop.

No blocking findings.

PR #2959, opened 2026-06-02
fix(gemini): coalesce parallel tool responses into a single Content

### LINKED ISSUE #2958: Vertex Gemini: parallel function-call responses are split across separate Contents, causing deterministic "function response parts" INVALID_ARGUMENT
## Description

When a Gemini model (via Vertex AI) returns parallel function calls (multiple `functionCall` parts in one assistant turn), docker-agent returns each tool result as its own `Content` (one `functionResponse` part per `Content`). Vertex AI requires the response turn to contain the same number of `functionResponse` parts as the preceding turn's `functionCall` parts, so the request is rejected. The failure is deterministic for a given turn, so retries reproduce it and the orchestrator eventually aborts.

## Expected Behavior

Parallel function-call results should be returned together as multiple `functionResponse` parts within a single `Content` (role `user`), with the part count matching the number of function calls — as described in the Vertex AI function-calling docs:

> collect API responses in parts and send them back to the model ... combined into one `Content(role="user", parts=function_response_parts)`

- Introduction to function calling: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling
- Parallel function calling: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/samples/generativeaionvertexai-function-calling-generate-parallel-calls
- Gemini API (function calling): https://ai.google.dev/gemini-api/docs/function-calling

## Actual Behavior

The request fails with HTTP 400 INVALID_ARGUMENT. Because the conversation history is fixed, every retry re-sends the same (mis-grouped) responses and fails identically; the orchestrator hits its consecutive-identical-call guard and aborts.

## Steps to Reproduce

1. Configure an orchestrator agent that delegates to a sub-agent via `transfer_task`, using a Gemini model on Vertex AI (`provider: google`).
2. Have the sub-agent perform two tool calls in a single turn (e.g. a `filesystem` agent told to write two files), so the assistant turn emits two `functionCall` parts.
3. The two tool results are sent back as two separate `Content` objects → the next model call returns 400.

Note: Vertex Gemini exposes no parameter to disable parallel function calling (`go-genai` `FunctionCallingConfig` only has `Mode`), so there is no opt-out — the responses must be grouped correctly.

## Environment

- **Docker Agent Version**: v1.70.2
- **OS & Terminal**: Linux
- **Model Used**: gemini-2.5-flash-lite (Vertex AI, location `global`)

## Error Output

```
model failed: error receiving from stream: Error 400, Message: Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn., Status: INVALID_ARGUMENT, Details: []
```

## Screenshots

N/A

## Additional Context

**Root cause** — `pkg/model/provider/gemini/client.go`, `convertMessagesToGemini`: an assistant message with N `ToolCalls` becomes one `genai.Content` with N `functionCall` parts, but each tool-response message becomes its own `genai.Content` with a single `functionResponse` part (the `role == Tool` branch emits one `Content` per message). The response turn therefore has 1 part per `Content` instead of N parts in one `Content`.

**Proposed fix** — coalesce consecutive tool-response messages into a single `genai.Content` with N `functionResponse` parts (flush on the next non-tool message and at end of loop). A fix plus a regression test are prepared (the test fails on `main` and passes with the fix; `go build ./...`, `go vet`, package tests, and `golangci-lint` are green). PR to follow.

- Not a regression — present since the per-message tool-response conversion in the gemini provider.

t-mizumoto1203:
Fixes #2958.

## Problem

When a Gemini model (via Vertex AI) returns parallel function calls (multiple `functionCall` parts in one assistant turn), `convertMessagesToGemini` returned each tool result as its **own** `genai.Content` (one `functionResponse` part per `Content`). Vertex AI requires the response turn to carry the same number of `functionResponse` parts as the preceding turn's `functionCall` parts, so the request fails with:

```
Error 400 ... Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn. INVALID_ARGUMENT
```

This is deterministic for a given turn, so retries reproduce it and the run eventually aborts. There is no opt-out: Vertex Gemini exposes no parameter to disable parallel function calling (`go-genai` `FunctionCallingConfig` only has `Mode`).

## Change

`convertMessagesToGemini` now **coalesces consecutive tool-response messages into a single `Content` (role `user`) with N `functionResponse` parts**, flushing on the next non-tool message and at the end of the loop — instead of emitting one `Content` per response. Single (non-parallel) tool calls are unaffected (one response → one Content).

Added regression test `TestConvertMessagesToGemini_ParallelToolResponsesCoalesced` (fails on `main`, passes with this change).

## Reference

Per the Vertex AI function-calling docs, parallel results are returned together as one `Content(role="user", parts=function_response_parts)` with the part count matching the calls:
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/samples/generativeaionvertexai-function-calling-generate-parallel-calls

## Verification

- `go build ./...` — ok
- `go vet ./pkg/model/provider/gemini/...` — ok
- `go test ./pkg/model/provider/...` — ok
- `golangci-lint run` (repo `.golangci.yml`, v2.12.2) — 0 issues

### DIFF
--- a/pkg/model/provider/gemini/client.go
+++ b/pkg/model/provider/gemini/client.go
@@ -196,6 +196,19 @@ func thoughtSignatureOrDefault(sig []byte) []byte {
 // convertMessagesToGemini converts chat.Messages into Gemini Contents
 func convertMessagesToGemini(ctx context.Context, messages []chat.Message, id modelsdev.ID, store *modelsdev.Store) []*genai.Content {
 	contents := make([]*genai.Content, 0, len(messages))
+
+	// Vertex Gemini rejects a request unless the turn answering an N-function-call turn
+	// carries exactly N function-response parts, so consecutive tool responses are
+	// coalesced into a single Content instead of one Content per response.
+	var pendingToolParts []*genai.Part
+	var pendingToolRole genai.Role
+	flushToolParts := func() {
+		if len(pendingToolParts) > 0 {
+			contents = append(contents, genai.NewContentFromParts(pendingToolParts, pendingToolRole))
+			pendingToolParts = nil
+		}
+	}
+
 	for i := range messages {
 		msg := &messages[i]
 
@@ -206,7 +219,7 @@ func convertMessagesToGemini(ctx context.Context, messages []chat.Message, id mo
 
 		role := messageRoleToGemini(msg.Role)
 
-		// Handle tool responses
+		// Handle tool responses (accumulated, then flushed as one Content)
 		if msg.Role == chat.MessageRoleTool && msg.ToolCallID != "" {
 			response := map[string]any{"result": msg.Content}
 
@@ -218,9 +231,12 @@ func convertMessagesToGemini(ctx context.Context, messages []chat.Message, id mo
 			} else {
 				part = genai.NewPartFromFunctionResponse(msg.ToolCallID, response)
 			}
-			contents = append(contents, genai.NewContentFromParts([]*genai.Part{part}, role))
+			pendingToolParts = append(pendingToolParts, part)
+			pendingToolRole = role
 			continue
 		}
+
+		flushToolParts()
 
 		// Handle assistant messages with tool calls
 		if msg.Role == chat.MessageRoleAssistant && len(msg.ToolCalls) > 0 {
@@ -255,6 +271,9 @@ func convertMessagesToGemini(ctx context.Context, messages []chat.Message, id mo
 			contents = append(contents, genai.NewContentFromParts([]*genai.Part{part}, role))
 		}
 	}
+
+	flushToolParts()
+
 	return contents
 }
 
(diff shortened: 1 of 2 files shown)


dgageot (approved):

t-mizumoto1203:
Rebased onto the latest `main`. The earlier red `build-and-test` was an unrelated pre-existing failure in `pkg/config` (`TestParseExamples` on `examples/github-copilot.yaml`: `model "gpt-4o" not found in provider "github-copilot"`), not caused by this change, which only touches `pkg/model/provider/gemini/`. That failure is fixed on `main` by #2961 (commit 36af646, gpt-4o → gpt-4.1), which this branch now includes — so the build-and-test check should pass on the re-run.

dgageot:
Sorry @t-mizumoto1203 we require all the contributors to sign and verify their commits. Could you sign your commit?

t-mizumoto1203:
@dgageot  Apologies for the late update. I've signed and pushed the commit now

PR #2963, opened 2026-06-02
feat(examples): add Atlassian expert agent example

maxcleme:


docker-agent (commented):
### Assessment: 🟢 APPROVE

The new `examples/atlassian-expert.yaml` is a clean and correct addition.

**What was checked:**
- OAuth configuration: The absence of an explicit `oauth:` block is correct and intentional. The runtime automatically handles OAuth via RFC 7591 Dynamic Client Registration when connecting to SaaS MCP endpoints like Atlassian. This matches the same pattern used by `notion-expert.yaml`.
- Transport type: `streamable` is a valid alias for the Streamable HTTP transport (equivalent to `streamable-http` per the source code in `pkg/tools/mcp/remote.go`).
- Model config: Using a named `models:` block with the `sonnet` alias is a valid and consistent pattern, matching other examples like `alloy.yaml` and `coder.yaml`.

examples/atlassian-expert.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+agents:
+  root:
+    model: sonnet
+    description: Atlassian Expert
+    instruction: You are an AI assistant with expertise in the Atlassian suite of products (Jira, Confluence).
+    toolsets:
+      - type: mcp
+        remote:
+          url: https://mcp.atlassian.com/v1/mcp/authv2
```
docker-agent, line 14: **[MEDIUM] Atlassian MCP endpoint may require OAuth credentials that are not configured**

The URL `https://mcp.atlassian.com/v1/mcp/authv2` contains `/authv2` in the path, which strongly suggests this endpoint requires OAuth 2.0 authentication. No `oauth` block is configured under `remote`, which may cause the agent to fail at runtime when the Atlassian MCP server rejects the unauthenticated connection.

If the docker-agent runtime supports Dynamic Client Registration (RFC 7591) or browser-based OAuth flows without a static `oauth` block, this may work as-is. However, if Atlassian's MCP server does **not** support DCR, explicit credentials are required. The `examples/remote_mcp_oauth.yaml` file shows the pattern for servers that require static OAuth credentials:

```yaml
remote:
  url: "https://mcp.atlassian.com/v1/mcp/authv2"
  transport_type: streamable
  oauth:
    clientId: "your-atlassian-client-id"
    clientSecret: "your-atlassian-client-secret"
    callbackPort: 8080
    scopes:
      - "read:jira-work"
      - "read:confluence-space.summary"
```

Please verify whether the Atlassian MCP server supports Dynamic Client Registration or whether explicit OAuth credentials should be documented (even as placeholder values) in this example.

docker-agent (commented):

dgageot (approved):

PR #2969, opened 2026-06-02
fix: detect Shift+Enter on macOS terminals without Kitty protocol support

gtardif:
macOS Terminal.app (and other legacy terminals) don't support the Kitty keyboard protocol, so Shift+Enter sends the same byte as plain Enter. This makes it impossible to distinguish the two at the terminal protocol level.

Use the macOS CoreGraphics API (CGEventSourceFlagsState) to query the system-level keyboard modifier state when a plain Enter key event is received. If Shift is physically held, treat it as a newline insertion instead of submitting the message.

This approach works universally across all macOS terminal emulators without requiring any terminal configuration. For terminals that do support the Kitty protocol (VSCode, iTerm2 3.5+, Ghostty), the existing shift+enter detection continues to work via the protocol, and the CoreGraphics fallback is never reached.

### DIFF
--- a/pkg/tui/components/editor/editor.go
+++ b/pkg/tui/components/editor/editor.go
@@ -771,6 +771,16 @@ func (e *editor) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 				return e, nil
 			}
 
+			// On macOS, terminals like Terminal.app don't support the Kitty keyboard
+			// protocol, so Shift+Enter sends the same byte as Enter. Use the macOS
+			// CoreGraphics API to check if Shift is physically held, and treat it
+			// as a newline insertion.
+			if msg.String() == "enter" && termfeatures.IsShiftPressed() {
+				e.textarea.InsertString("\n")
+				e.refreshSuggestion()
+				return e, nil
+			}
+
 			// Let textarea process the key - it handles newlines via InsertNewline binding
 			prev := e.textarea.Value()
 			e.textarea, _ = e.textarea.Update(msg)
--- a/pkg/tui/internal/termfeatures/keyboard.go
+++ b/pkg/tui/internal/termfeatures/keyboard.go
@@ -1,10 +1,19 @@
 package termfeatures
 
-import "strings"
+import (
+	"runtime"
+	"strings"
+)
 
 // SupportsModifiedEnter returns true for terminals that can distinguish
 // Shift+Enter from Enter even when they do not report Kitty keyboard flags.
+// On macOS, we use the CoreGraphics API to detect modifier key state at the
+// system level, so all terminals effectively support this.
 func SupportsModifiedEnter(getenv func(string) string) bool {
+	if runtime.GOOS == "darwin" {
+		return true
+	}
+
 	if getenv == nil {
 		return false
 	}
--- a/pkg/tui/internal/termfeatures/modifiers_darwin.go
+++ b/pkg/tui/internal/termfeatures/modifiers_darwin.go
@@ -1,0 +1,23 @@
+//go:build darwin
+
+package termfeatures
+
+/*
+#cgo LDFLAGS: -framework CoreGraphics
+#include <CoreGraphics/CoreGraphics.h>
+
+static int isShiftPressed() {
+    CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateCombinedSessionState);
+    return (flags & kCGEventFlagMaskShift) != 0;
+}
+*/
+import "C"
+
+// IsShiftPressed queries the macOS CoreGraphics event system to determine
+// whether the Shift key is currently held down. This allows us to distinguish
+// Shift+Enter from Enter in terminals that don't support the Kitty keyboard
+// protocol (like macOS Terminal.app), since those terminals send the same byte
+// for both key combinations.
+func IsShiftPressed() bool {
+	return C.isShiftPressed() != 0
+}
--- a/pkg/tui/internal/termfeatures/keyboard_test.go
+++ b/pkg/tui/internal/termfeatures/keyboard_test.go
@@ -1,6 +1,9 @@
 package termfeatures
 
-import "testing"
+import (
+	"runtime"
+	"testing"
+)
 
 func TestSupportsModifiedEnter(t *testing.T) {
 	t.Parallel()
@@ -14,8 +17,8 @@ func TestSupportsModifiedEnter(t *testing.T) {
 		{name: "wezterm pane", env: map[string]string{"WEZTERM_PANE": "1"}, want: true},
 		{name: "wezterm socket", env: map[string]string{"WEZTERM_UNIX_SOCKET": "/tmp/wezterm.sock"}, want: true},
 		{name: "wezterm term", env: map[string]string{"TERM": "wezterm"}, want: true},
-		{name: "other terminal", env: map[string]string{"TERM_PROGRAM": "Apple_Terminal", "TERM": "xterm-256color"}, want: false},
-		{name: "nil getenv", env: nil, want: false},
+		{name: "other terminal", env: map[string]string{"TERM_PROGRAM": "Apple_Terminal", "TERM": "xterm-256color"}, want: runtime.GOOS == "darwin"},
+		{name: "nil getenv", env: nil, want: runtime.GOOS == "darwin"},
 	}
 
 	for _, tt := range tests {
--- a/pkg/tui/internal/termfeatures/modifiers_other.go
+++ b/pkg/tui/internal/termfeatures/modifiers_other.go
@@ -1,0 +1,9 @@
+//go:build !darwin
+
+package termfeatures
+
+// IsShiftPressed returns false on non-macOS platforms. The CoreGraphics-based
+// modifier detection is only available on macOS.
+func IsShiftPressed() bool {
+	return false
+}


pkg/tui/components/editor/editor.go
```diff
@@ -771,6 +771,16 @@ func (e *editor) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 				return e, nil
 			}
 
+			// On macOS, terminals like Terminal.app don't support the Kitty keyboard
+			// protocol, so Shift+Enter sends the same byte as Enter. Use the macOS
+			// CoreGraphics API to check if Shift is physically held, and treat it
+			// as a newline insertion.
+			if msg.String() == "enter" && termfeatures.IsShiftPressed() {
```
docker-agent, line 778: **[MEDIUM] Shift+Enter polling race: Shift release between keypress and handler poll misclassifies input**

`IsShiftPressed()` polls `CGEventSourceFlagsState` *after* the terminal has already sent the Enter byte and the Go runtime has delivered it to the bubbletea message loop. If the user releases the Shift key in the brief interval between pressing Enter and this handler executing, `IsShiftPressed()` returns `false` and the input is treated as a form submission rather than a newline insertion.

This is an inherent design-level timing issue with post-hoc key-state polling. The physical key state at the time of the event cannot be reconstructed from `kCGEventSourceStateCombinedSessionState` — it only reflects the *current* state of the key at the moment of the call.

While the window is small (a few milliseconds) and rare in practice, the failure mode is subtle: the user sees their multi-line draft unexpectedly submitted mid-sentence. Consider documenting this limitation, or explore capturing modifier state closer to the raw terminal read loop.

pkg/tui/internal/termfeatures/modifiers_darwin.go
```diff
@@ -0,0 +1,23 @@
+//go:build darwin
+
+package termfeatures
+
+/*
+#cgo LDFLAGS: -framework CoreGraphics
+#include <CoreGraphics/CoreGraphics.h>
+
+static int isShiftPressed() {
+    CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateCombinedSessionState);
```
docker-agent, line 10: **[MEDIUM] `CGEventSourceFlagsState` silently returns 0 in sandboxed macOS processes**

`CGEventSourceFlagsState(kCGEventSourceStateCombinedSessionState)` is a privileged CoreGraphics API. On macOS 10.15+ (Catalina and later), processes running under the App Sandbox (or with certain Hardened Runtime entitlements) receive a silent zero return value — no modifiers, no error. This means `IsShiftPressed()` will **always return `false`** in a sandboxed distribution, causing Shift+Enter to be indistinguishable from Enter — the exact failure mode this PR is trying to fix.

If docker-agent is ever distributed as a notarized `.app` bundle with sandboxing enabled, this feature will silently regress on the target platform (macOS Terminal.app users) without any log message or user-visible warning.

Recommendation: add a code comment noting this sandbox limitation so future distributors are aware. Alternatively, consider detecting the sandboxed state at startup (via `APP_SANDBOX_CONTAINER_ID` env var) and logging a warning when the feature is unavailable.

docker-agent (commented):

pkg/tui/components/editor/editor.go
```diff
@@ -771,6 +771,16 @@ func (e *editor) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
...
+				e.textarea.InsertString("\n")
```
docker-agent, line 779: **[MEDIUM] Timing race: Shift released between key-down delivery and CGo poll may cause accidental submit**

`IsShiftPressed()` polls `CGEventSourceFlagsState` asynchronously — after BubbleTea has already dispatched the terminal key event. There is an inherent time gap (potentially 5–50 ms) between when the PTY receives the Enter byte and when the Go event loop executes the `if msg.String() == "enter" && termfeatures.IsShiftPressed()` check.

A user who types Shift+Enter and releases Shift before `CGEventSourceFlagsState` is polled will see the keypress treated as plain Enter (submitting the message) instead of inserting a newline. This is especially likely during fast typing.

The code has no mitigation: no timestamp comparison, no event-synchronized modifier state, and the comment above the block only explains the protocol limitation, not the timing limitation.

**Trigger path**: User holds Shift, taps Enter, releases Shift quickly → PTY byte arrives → Go event loop delayed by scheduling → `IsShiftPressed()` called after Shift is already released → returns `false` → message is submitted instead of newline inserted.

This is an inherent limitation of polling-based modifier detection vs. an event-synchronized approach, and may be acceptable given the alternatives, but it should be acknowledged in comments or documentation.

pkg/tui/internal/termfeatures/keyboard.go
```diff
@@ -1,10 +1,19 @@
...
 
-import "strings"
+import (
+	"runtime"
+	"strings"
+)
 
 // SupportsModifiedEnter returns true for terminals that can distinguish
 // Shift+Enter from Enter even when they do not report Kitty keyboard flags.
+// On macOS, we use the CoreGraphics API to detect modifier key state at the
```
docker-agent, line 10: **[MEDIUM] `SupportsModifiedEnter` semantics changed: returns `true` on Darwin even for Kitty-unaware terminals**

`SupportsModifiedEnter` now returns `true` unconditionally on Darwin, but its original contract is "returns true for terminals that can distinguish Shift+Enter from Enter at the terminal protocol level." Terminal.app *cannot* — it always sends plain `enter` for Shift+Enter.

**Practical effect**: `configureNewlineKeybinding()` will register `"shift+enter"` as an `InsertNewline` keybinding on Terminal.app. But since Terminal.app never sends `"shift+enter"`, this binding is dead and can never fire. The actual newline insertion on Terminal.app relies entirely on the `IsShiftPressed()` path in `editor.go`.

**Risk**: Any downstream code that reads `keyboardEnhancementsSupported` and assumes `true` means "the terminal sends `shift+enter` bytes" could behave incorrectly on Terminal.app. If this flag is used elsewhere — e.g., for UI hints, capability negotiation, or input processing — those paths may receive incorrect signals.

The dead `"shift+enter"` keybinding itself is harmless (it can never match). But the semantic change from "terminal protocol capability" to "platform can detect shift via some mechanism" should be explicit — consider renaming the function or adding a comment clarifying the new meaning.

docker-agent (commented):

This PR introduces macOS CoreGraphics-based Shift key detection to handle Shift+Enter in terminals that don't support the Kitty keyboard protocol. The build tag split (`darwin` / `!darwin`) is correct and exhaustive. The CGo constants (`kCGEventSourceStateCombinedSessionState`, `kCGEventFlagMaskShift`) are appropriate. Kitty-capable terminals (iTerm2, Ghostty, VSCode) are not affected since they send `shift+enter` as a distinct key event.

Two LIKELY issues found in the added code — see inline comments.

pkg/tui/internal/termfeatures/modifiers_darwin.go
```diff
@@ -0,0 +1,23 @@
+//go:build darwin
```
docker-agent, line 1: **[HIGH] CGo dependency breaks darwin cross-compilation from Linux CI**

`modifiers_darwin.go` introduces a CGo dependency (`#cgo LDFLAGS: -framework CoreGraphics` + `import "C"`), which requires a native darwin C toolchain at build time. The project's `task cross` command builds darwin/amd64 and darwin/arm64 targets — if that task runs on a Linux CI host (typical for cross-platform builds), the darwin CGo target will fail to compile because:

1. A darwin cross-compiler (e.g. osxcross) is not a standard Go CI setup, and
2. Forcing `CGO_ENABLED=0` for darwin would also break the build since the `import "C"` cannot be excluded.

**Impact**: CI pipelines doing cross-compilation for darwin will fail. Docker's own multi-platform build (`task cross`) is likely affected.

**Suggestion**: Consider adding a build tag guard or a `//go:build ignore` variant so that the CGo implementation is only compiled natively on a darwin host, and document that darwin binaries must be built natively. Alternatively, restructure to allow `CGO_ENABLED=0` fallback (where `IsShiftPressed()` returns false, falling back to the ctrl+j path).

pkg/tui/components/editor/editor.go
docker-agent, line 778: **[MEDIUM] Timing race: Shift key may be released before IsShiftPressed() is polled**

`CGEventSourceFlagsState` is called at the moment bubbletea dispatches the key event to the `Update` function — not at the exact moment the physical key was pressed. If the user releases Shift before the Go event loop processes the message (possible under goroutine scheduling delays or system load), `IsShiftPressed()` returns `false`, and the Enter is treated as a plain Enter — submitting the message instead of inserting a newline.

**Impact**: Intermittent false negatives: the user presses Shift+Enter intending to insert a newline, but the message is submitted unexpectedly. This is a regression compared to terminals that use the Kitty protocol (where the key event carries shift modifier flags, immune to timing).

**Suggestion**: The PR description acknowledges this is a best-effort approach, but this limitation should be documented (e.g., a code comment or a note in the PR). Consider whether the timing window is acceptable in practice (bubbletea dispatch is typically sub-millisecond), and if so, document it explicitly to set expectations.

pkg/tui/components/editor/editor.go
docker-agent, line 779: **[LOW] InsertString("\n") vs InsertNewline may behave differently**

The Kitty-protocol path inserts a newline by passing the key event through `e.textarea.Update(msg)`, which dispatches the `InsertNewline` key binding internally. The new macOS path directly calls `e.textarea.InsertString("\n")`.

For the current textarea implementation these are likely equivalent, but if the bubbletea `textarea` widget ever adds smart-indentation, auto-pairing, or other `InsertNewline`-specific logic, the two paths will diverge silently.

**Suggestion**: If a direct `InsertNewline` API is available on the textarea, prefer calling it for consistency with the Kitty-protocol path.

docker-agent (commented):

3 finding(s): 1 HIGH (LIKELY), 1 MEDIUM (LIKELY), 1 LOW (LIKELY). See inline comments for details.

dgageot:
Let's close this one for now

PR #2971, opened 2026-06-02
feat(config): support top-level shared skills and commands

### LINKED ISSUE #2967: Support top level skills and commands
In order to share skills or commands between multiple agents in the same yaml, it would be nice to have top level skills and commands

Sayt-0:
## What

Implements top-level, shareable `skills` and `commands`, so the same command set or skill configuration can be reused across multiple agents in a single YAML instead of being duplicated.

## How

- New top-level `commands` and `skills` maps of named, reusable groups.
- Agents opt in by name via new `use_commands` / `use_skills` fields.
- Groups are merged into each agent during config resolution, so `teamloader` and the runtime are unchanged. Agent inline entries take precedence on name conflicts; skill sources/includes are unioned and inline skills deduped by name.
- Config schema bumped to v11 (previous schema frozen as `pkg/config/v10`).
- Updated `agent-schema.json`, the HCL block rules, added `examples/shared-commands-skills.yaml`, and documented the sections in `docs/configuration/overview`.

## Testing

- New tests in `pkg/config` cover merge semantics, inline override, missing-group errors, and top-level group validation.
- Verified end-to-end with `cagent debug config examples/shared-commands-skills.yaml`.

Closes #2967


pkg/config/v10/auth.go
```diff
@@ -0,0 +1,253 @@
+package v10
```
dgageot: All those changes should go in the `latest` package, no need to create a new version

dgageot (changes requested):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26831087402).

dgageot (approved):

pkg/config/skills.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/docker/docker-agent/pkg/config/latest"
+)
+
+// resolveSkillDefinitions merges reusable skill groups referenced by agents
+// (via use_skills) from the top-level skills section into each agent's
+// SkillsConfig. Every top-level group is validated up front (parity with
+// resolveMCPDefinitions), then each referenced group is merged into the agent.
+func resolveSkillDefinitions(cfg *latest.Config) error {
+	for name := range cfg.Skills {
```
docker-agent, line 15: **[LOW] Non-deterministic error message when multiple skill groups are invalid**

`for name := range cfg.Skills` iterates the map in random order. If two or more top-level skill groups are simultaneously invalid, the error returned on each run may name a different group, producing unstable output that can make error messages hard to reproduce and could cause flaky tests if any future test ever checks which group's error surfaces.

Suggested fix: collect all group names into a sorted slice before iterating, so validation always reports errors in a deterministic order (consistent with how other validation loops in the codebase use `slices.Sorted` / `sort.Strings`).

pkg/config/commands.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/config/types"
+)
+
+// resolveCommandDefinitions merges reusable command groups referenced by agents
+// (via use_commands) from the top-level commands section into each agent's
+// Commands map. Commands defined inline on the agent take precedence over a
+// group's command of the same name, mirroring the override semantics used for
+// MCP definitions (see applyMCPDefaults).
+func resolveCommandDefinitions(cfg *latest.Config) error {
```
docker-agent, line 15: **[LOW] Top-level `commands` groups are not validated proactively**

`resolveSkillDefinitions` validates every top-level skill group (even unreferenced ones) before merging. `resolveCommandDefinitions` has no equivalent loop — top-level command groups are only inspected when an agent explicitly references them via `use_commands`. A misconfigured but unreferenced command group will never be caught.

While current `CommandConfig` fields are validated at YAML parse time, the asymmetry will become a silent correctness gap if richer semantic validation is added to commands in the future.

Suggested fix: add an up-front validation loop over `cfg.Commands` (mirroring the `cfg.Skills` loop) or add a comment explaining the intentional difference.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is well-structured. Merge precedence (inline wins over shared), nil-map guards, skill deduplication, and validation ordering are all correct. No high or medium severity bugs were found.

Two minor observations are left as inline comments.

PR #2973, opened 2026-06-02
feat(mcp_catalog): block enable_remote_mcp_server until the server is connected

trungutt:
### 1. The user has to re-ask their original question after enabling a server

`enable_remote_mcp_server`'s handler defers the MCP connect (and any OAuth handshake) to the next `Tools()` enumeration. The tool's result text and the catalog Instructions block both tell the model "tools appear on your **next turn**". Every model reading that ends its turn after enable with a "please try again" message; the user then has to **re-send their original question** for the activated server's tools to actually be invoked.

```
User : List all Notion posts by David
Model: <calls search → calls enable("notion")>
       Notion is being set up. Please try again on your next message.
User : List all Notion posts by David          ← forced re-ask
Model: <now actually uses notion_* tools>
```

This was never a runtime constraint — `getTools` rebuilds the tool list at the top of every iteration of the run loop, and `reprobe` (with its regression test `TestReprobe_NewToolsAvailableAfterToolCall`) explicitly supports new tools appearing mid-stream. The deferred-Start design was an artifact: registering the toolset was cheap, but the prompt then had to admit the runtime couldn't guarantee tools would actually be available, since OAuth might be declined or the server might refuse the connection.

After this PR `handleEnable` drives Start synchronously inside the tool call. OAuth elicitation surfaces a dialog and blocks until the user responds; the success / failure result reaches the model at the point of the call, in the same RunStream. The Instructions block and the tool description are rewritten to match (no more "appears on your next turn" wording on the happy path).

### 2. OAuth is still required on the next question even if the previous OAuth failed

If the user stops a turn while the OAuth dialog is parked (the browser-side flow failed externally, the user gave up, etc.), Start returns `context.Canceled`. On main this falls into a generic warning-log branch and the entry stays in `t.enabled` in an unstarted state. The next user message — typically unrelated to the abandoned server — calls `Tools()`, which silently re-Starts the wrapper, hits 401, and re-pops the same dialog the user just abandoned.

The root cause is a lifetime mismatch: the catalog `Toolset` is owned by the `RuntimeSession`, not the `RunStream`, so `Toolset.Stop` does NOT run between a stopped turn and the user's next message.

```
                  RuntimeSession  (lives across many user messages)
┌─────────────────────────────────────────────────────────────┐
│  RunStream #1   │   RunStream #2   │   RunStream #3         │
│   Stop here─┘                                               │
│                                                             │
│  mcpcatalog.Tools()        runs at the start of every       │
│                            RunStream — iterates t.enabled   │
│                            and Start()s any unstarted entry │
│                                                             │
│  mcpcatalog.Toolset.Stop() runs ONCE, at session teardown   │
└─────────────────────────────────────────────────────────────┘
```

`context.Canceled` is now routed through `disableAfterDecline` (the same path that already handles an explicit `OAuthDeclined`) in both code paths that can Start a catalog server — the synchronous handleEnable and the per-turn Tools() iteration. The model-facing result tells the model the user stopped the turn while authorization was pending and to only retry if the user asks.

Public docs at `docs/tools/mcp-catalog/index.md` are updated for the new contract.

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
+++ b/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
@@ -8,8 +8,9 @@
 //     curated catalog (id / title / description / category / tags).
 //   - list_remote_mcp_servers   — show currently enabled servers.
 //   - enable_remote_mcp_server  — instantiate an *mcp.Toolset for a server
-//     (defers the actual TCP connect / OAuth handshake until Tools() is
-//     next enumerated).
+//     and synchronously drive its connect (including any required OAuth
+//     handshake) so the model gets a deterministic success / declined /
+//     transport-error result in the same turn.
 //   - disable_remote_mcp_server — stop the toolset and remove its tools.
 //     Only exposed once at least one server is enabled.
 //   - reset_remote_mcp_server_auth — drop persisted OAuth credentials so
@@ -26,11 +27,20 @@
 // (the primary interface) work fine; the prompt feature would need a
 // separate plumb-through interface to walk into container toolsets.
 //
-// On-demand semantics: the expensive parts — DNS, TCP, MCP handshake,
-// OAuth flow — happen the first time Tools() is called for a freshly
-// enabled server. The handshake runs through the same lifecycle.Supervisor
-// the YAML-declared `mcp.remote` toolset uses, so OAuth elicitation and
-// tool-list-change notifications behave identically.
+// On-demand semantics: DNS, TCP, MCP handshake and any OAuth flow happen
+// synchronously inside enable_remote_mcp_server's handler. Tool calls run
+// with an interactive context, so OAuth elicitation surfaces a dialog and
+// blocks until the user responds; the handshake runs through the same
+// lifecycle.Supervisor the YAML-declared `mcp.remote` toolset uses, so
+// elicitation and tool-list-change notifications behave identically. The
+// handler returns a deterministic success / declined / auth-required /
+// transport-error result so the model can continue with the user's
+// original request in the *same* turn (success) or recover appropriately
+// (failure) — no second user message required. Tools() additionally
+// retries any toolset left in the "deferred — auth required" state, which
+// keeps the startup non-interactive probe (mcp.WithoutInteractivePrompts)
+// from hanging while still surfacing OAuth dialogs on the first
+// interactive turn.
 package mcpcatalog
 
 import (
@@ -95,6 +105,14 @@ type Toolset struct {
 	// Defaults to mcp.RemoveOAuthToken; tests inject a stub to avoid
 	// touching the OS keyring.
 	removeOAuthToken func(resourceURL string) error
+
+	// startToolset drives the inner mcp.Toolset connect (including any
+	// OAuth handshake) synchronously from handleEnable so the model
+	// gets a deterministic success/decline/error result in the same
+	// turn. Defaults to (*tools.StartableToolSet).Start; tests inject a
+	// stub to exercise the success / declined / auth-required / errored
+	// branches without standing up a real MCP server.
+	startToolset func(ctx context.Context, ts *tools.StartableToolSet) error
 }
 
 var (
@@ -150,6 +168,9 @@ func New(envProvider environment.Provider, opts ...Option) *Toolset {
 		env:              envProvider,
 		enabled:          make(map[string]*tools.StartableToolSet),
 		removeOAuthToken: mcp.RemoveOAuthToken,
+		startToolset: func(ctx context.Context, ts *tools.StartableToolSet) error {
+			return ts.Start(ctx)
+		},
 	}
 	t.filterCatalog(cfg.allowedServers, cfg.blockedServers)
 	return t
@@ -251,18 +272,26 @@ Workflow:
      Use any term related to the user's intent ("notion", "stripe",
      "docs", "search", "browser", …).
   2. Call ` + ToolNameEnable + ` with the server's "id" to activate it.
-     The server's tools appear in your set on the *next* turn; the host
-     handles any required connection setup. For api_key servers, make
-     sure the listed env var(s) are set in the user's shell BEFORE
-     enabling, otherwise the server will refuse the connection.
-  3. On your NEXT turn (once the activated server's tools appear in
-     your set), go straight to the user's ORIGINAL request using those
-     tools. Do NOT stop and ask the user to repeat themselves; do NOT
-     narrate connection setup. Enabling a server is a means to an end,
-     not a stopping point.
-  4. Call ` + ToolNameDisable + ` to remove a server when no longer needed.
+     This call BLOCKS until the connection (and any required OAuth
+     handshake) completes, so by the time it returns you have a
+     deterministic answer:
+       - SUCCESS — the server's tools (names prefixed with the server's
+         id and an underscore) are live. Continue with the user's
+         ORIGINAL request using them right away. Do NOT stop, do NOT
+         ask the user to repeat themselves, do NOT narrate connection
+         setup. Enabling a server is a means to an end, not a stopping
+         point.
+       - ERROR — the result text tells you exactly why (user declined
+         the authorization dialog, missing env var, server refused).
+         Act on that specific reason; do NOT pretend the server is
+         connected and do NOT call any "<id>_..." tools (there are
+         none).
+     For api_key servers, make sure the listed env var(s) are set in
+     the user's shell BEFORE enabling, otherwise the enable will fail
+     with a missing-env warning.
+  3. Call ` + ToolNameDisable + ` to remove a server when no longer needed.
      This tool only appears once at least one server is enabled.
-  5. If a previously enabled server starts rejecting requests
+  4. If a previously enabled server starts rejecting requests
      (credentials revoked, scopes changed, signed in to the wrong
      account), call ` + ToolNameResetAuth + ` to clear any persisted
      credentials so the next enable starts fresh. This tool also only
@@ -418,7 +447,7 @@ func (t *Toolset) Tools(ctx context.Context) ([]tools.Tool, error) {
 		{
 			Name:         ToolNameEnable,
 			Category:     "mcp_catalog",
-			Description:  "Activate a remote MCP server from the catalog by id. Connection setup is deferred until the host next enumerates tools; the server's tools become available on the next turn.",
+			Description:  "Activate a remote MCP server from the catalog by id. Blocks until the connection (and any required OAuth handshake) completes; on success the server's tools are immediately available and you should continue with the user's original request using them.",
 			Parameters:   tools.MustSchemaFor[EnableArgs](),
 			OutputSchema: tools.MustSchemaFor[string](),
 			Handler:      tools.NewHandler(t.handleEnable),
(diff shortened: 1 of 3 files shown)


docker-agent (commented):

Two medium-severity issues found in the new synchronous-enable state machine. Both relate to cases where `t.enabled[id]` can be left with `IsStarted() == false` and a subsequent `enable` call hits the idempotent guard instead of re-attempting `Start()`.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -717,27 +766,73 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
...
+//     drop the entry and surface the underlying message so the model can
+//     decide what to tell the user.
+func (t *Toolset) handleEnableStartError(ctx context.Context, id string, server Server, wrapped *tools.StartableToolSet, err error) *tools.ToolCallResult {
+	switch {
+	case mcp.IsOAuthDeclined(err):
+		t.disableAfterDecline(ctx, id, wrapped)
+		return tools.ResultError(fmt.Sprintf(
+			"user declined the authorization dialog for %q (%s). No tools were activated — do NOT claim the server is connected and do NOT call any %q tools. Tell the user the request needs them to authorize the connection. If the user then says \"yes\", \"retry\", or re-asks for the same thing, call %s for %q again to surface a fresh authorization dialog.",
+			id, server.Title, id+"_", ToolNameEnable, id))
+	case mcp.IsAuthorizationRequired(err):
```
trungutt, line 863: Good catch — fixed in 33d73f569.

The root cause was the same for both your comments: the top-of-`handleEnable` guard short-circuited any entry already in `t.enabled` regardless of `IsStarted()`. Now the guard detects existing-but-unstarted entries and falls through to a fresh `Start` attempt on the same wrapper, so the AuthorizationRequired branch's "call enable again to surface a fresh authorization dialog" instruction actually does something.

`StartableToolSet.Start` is already idempotent and single-flight: when the previous attempt left the wrapper not-started, the retry re-invokes the inner `Start` without re-creating the toolset or re-firing `tools_changed`. The "already enabled but not yet connected" dead-end message is gone.

Regression test: `TestEnableRetriesStartOnExistingUnstartedEntry`.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -717,27 +766,73 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
...
+		slog.DebugContext(ctx, "Remote MCP server enable deferred: authorization required, leaving in enabled set for next interactive Tools() to retry",
+			"id", id, "error", err)
+		return tools.ResultSuccess(fmt.Sprintf(
+			"enable requested for %q (%s); authorization is required and the host will surface the dialog. On your next turn, if tools whose names start with %q appear in your available tools, proceed with the user's original request using them. If NO such tools appear, the user dismissed the dialog — tell them the request needs them to authorize, and call %s for %q again if they want to retry.",
+			id, server.Title, id+"_", ToolNameEnable, id))
+	case errors.Is(err, context.Canceled):
```
trungutt, line 869: Fixed in 33d73f569 — same root cause as the AuthorizationRequired comment above.

The `context.Canceled` branch still intentionally leaves the entry in `t.enabled` (rolling back would race a Ctrl+C-driven `Toolset.Stop` on the same map), but the dead-end you flagged is gone: the next `enable` for the same id now falls through the guard and re-attempts `Start` on the existing wrapper. The "call `enable_remote_mcp_server` again to retry" hint in the cancel result text has somewhere to land.

I also added a sentence to the branch comment so the invariant the design relies on (guard retries unstarted entries) is documented at the failure site, not just at the guard.

Regression test: `TestEnableRetriesStartAfterCancellation`.

docker-agent (commented):

3 medium-severity findings in the new synchronous  flow.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -677,67 +708,152 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
...
 
+	// Two paths land here: a fresh enable (no entry yet) and a re-enable
+	// of an existing entry whose previous Start did not complete (deferred
+	// auth-required fallback, cancellation, or any other transient
+	// failure). Both must converge on a Start attempt, otherwise the
+	// model-facing retry instructions in the failure branches dead-end at
+	// the guard.
 	t.mu.Lock()
-	if _, exists := t.enabled[id]; exists {
+	wrapped, alreadyEnabled := t.enabled[id]
```
trungutt, line 744: I don't think this race exists — the lookup *and* the insert both happen inside the same `t.mu` critical section, so two goroutines for the same id can't both observe `alreadyEnabled=false`:

```go
t.mu.Lock()                                  // ← serializes
wrapped, alreadyEnabled := t.enabled[id]     // read
...
if !alreadyEnabled {
    mcpToolset := mcp.NewRemoteToolset(...)
    ...
    wrapped = tools.NewStartable(mcpToolset)
    t.enabled[id] = wrapped                   // write — still under lock
    notify = t.toolsChangedHandler
}
t.mu.Unlock()
```

Whichever goroutine acquires the lock first does the read, creates the toolset, inserts it, and releases. The second goroutine then acquires, finds the just-inserted entry, observes `alreadyEnabled=true`, and either returns the "already enabled and connected" fast path or falls through to the Start retry — which delegates to `StartableToolSet.Start`, itself documented as single-flight (it holds an internal mutex and short-circuits if `s.started` is already true). So we get at most one `NewRemoteToolset`, one `notify()`, and one in-flight `Start` per id.

I don't want to plumb a "pending" sentinel through this path just to defend against a race the lock already prevents — happy to revisit if I'm missing a scenario where the lock is released between the read and the write.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -660,11 +689,13 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
...
-	// is immutable (from t.byID), no mutex protection needed here.
+	// Pre-flight: block if an api_key server is missing its env var. The
+	// catalog promises enable returns a deterministic answer in the same
+	// turn, so we cannot register a toolset we already know will be
+	// rejected on the first request. The model is told to ask the user to
+	// set the variable and retry.
+	// Perform these slow external calls BEFORE acquiring the lock — server
+	// data is immutable (from t.byID), no mutex protection needed here.
 	missing := t.missingAPIKeyEnv(ctx, server)
 	headers := t.expandHeaders(ctx, server.Headers)
```
trungutt, line 718: Correct observation but I'd like to defer this — it's tangential to what this PR set out to do (block enable so the user does not have to re-ask), and the realistic risk is low:

- Catalog headers come from the static, embedded `servers.go`. They reference env vars by name (e.g. `Authorization: Bearer ${APIFY_API_KEY}`), and those env vars are read at process start; they don't normally change between turns.
- The retry path only triggers on `AuthorizationRequired` (defensive, basically never fires interactively) or `context.Canceled` (RunStream is shutting down — there is no "next turn" for the retry to matter to).
- Re-creating the toolset on retry would double the path's complexity for a behaviour change no current catalog entry needs.

Filing as a known limitation in my head; happy to do it as a follow-up if a future catalog entry actually needs short-lived headers (at which point we'd probably also want a documented "headers must be long-lived OR opt into refresh-on-retry" contract on `Server`).

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
```diff
@@ -677,67 +708,152 @@ func (t *Toolset) handleEnable(ctx context.Context, args EnableArgs) (*tools.Too
...
+//     and surface the underlying message so the model can decide what to
+//     tell the user. A subsequent enable lands on a fresh entry.
+func (t *Toolset) handleEnableStartError(ctx context.Context, id string, server Server, wrapped *tools.StartableToolSet, err error) *tools.ToolCallResult {
+	switch {
+	case mcp.IsOAuthDeclined(err):
+		t.disableAfterDecline(ctx, id, wrapped)
+		return tools.ResultError(fmt.Sprintf(
+			"user declined the authorization dialog for %q (%s). No tools were activated — do NOT claim the server is connected and do NOT call any %q tools. Tell the user the request needs them to authorize the connection. If the user then says \"yes\", \"retry\", or re-asks for the same thing, call %s for %q again to surface a fresh authorization dialog.",
+			id, server.Title, id+"_", ToolNameEnable, id))
+	case mcp.IsAuthorizationRequired(err):
```
trungutt, line 863: The runtime already has a generic safeguard for this: the loop detector at `pkg/runtime/loop.go` (`NewLoopDetector(loopThreshold, ...)`, default `5`) terminates the agent with `ErrorCodeLoopDetected` after N consecutive identical tool calls. A model that keeps re-issuing `enable_remote_mcp_server` with the same `id` would trip it long before this turned into a real infinite loop, and the failure surface is already user-visible.

I don't think adding a per-id retry counter inside the catalog earns its keep when the runtime already handles "model is stuck calling the same tool" generically — it'd be a second, narrower copy of the same policy.

For context, this branch is only reachable when `Start` returns `AuthorizationRequired`, which happens when the request context disallows interactive prompts (`mcp.WithoutInteractivePrompts`). Tool handlers run with the interactive context, so in normal operation `Start` blocks on the OAuth elicitation instead and we never enter this branch. It is, as the comment says, a defensive fallback for hosts that haven't wired up the elicitation bridge.

pkg/tools/builtin/mcpcatalog/mcpcatalog.go
trungutt: Good question — and the honest answer is "I can't name one in current code". `ctx` here is the runtime's RunStream context, which only cancels when the whole stream shuts down (Ctrl+C, parent cancellation). I was hedging against a hypothetical future per-tool timeout, which was speculative.

Tightened the comments in 27b3f5034 to state what actually happens today: cancellation means RunStream is going away, `Toolset.Stop` cleans up, and the top-of-function guard would also re-Start an unstarted wrapper if a follow-up enable somehow ran on the same id — so the retry instruction in the cancel result has somewhere to land even though the realistic path is "session ended".

docker-agent (commented):
### Assessment: 🟢 APPROVE

The synchronous-connect refactor is well-structured. The mutex discipline is correct, the injectable `startToolset` seam keeps tests deterministic, and all four Start-failure branches (OAuthDeclined, AuthorizationRequired, Canceled, transport error) are individually tested with clear rollback semantics. The pre-flight env-var guard is a welcome improvement — blocking instead of warning prevents the model from claiming a connection that will immediately fail.

Three minor observations (none block merge):

1. **`stubStartOK`/`stubStartErr` bypass `wrapped.Start`** so `wrapped.IsStarted()` is never `true` in stub-based tests. The "already enabled and connected" fast-path guard (`alreadyEnabled && IsStarted()`) is consequently never exercised. Production behaviour is correct because the real `startToolset` calls `wrapped.Start(ctx)` which sets the flag. A future test could use a real `StartableToolSet` with a no-op inner toolset to close this gap — not urgent.

2. **`disableAfterDecline` called with a potentially-cancelled ctx** in the `default` error branch. A wrapped error containing `context.Canceled` falls through to `default` (only unwrapped `context.Canceled` matches the earlier branch), so `wrapped.Stop(ctx)` receives a cancelled context. `Stop` silently swallows its error, so no observable failure — low risk.

3. **Concurrent re-enable of an unstarted entry**: two goroutines can both pass `alreadyEnabled && !IsStarted()` and both call `startToolset`. Because `StartableToolSet.Start` is single-flight, this is benign — the second call returns immediately. Mentioned for completeness.

All high-risk paths (rollback correctness, error-branch coverage, idempotency) look solid.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly implements synchronous blocking in `enable_remote_mcp_server` with proper error classification (OAuthDeclined, AuthRequired, Canceled, transport error) and appropriate state rollback on each failure path. The `disableAfterDecline` guard (`current == wrapped`) correctly handles concurrent failure races. Test coverage for the new failure branches is thorough.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly implements synchronous blocking `Start()` in `handleEnable` and properly routes `context.Canceled` through `disableAfterDecline`. The logic is sound and the approach resolves both stated issues (forced re-ask and OAuth re-pop on the next turn).

**What was checked:**
- Concurrency safety of `handleEnable` for concurrent calls on the same unstarted entry → `StartableToolSet.Start` is documented and implemented as single-flight/idempotent; no race
- Handler propagation to existing unstarted entries on retry → `Set*Handler` methods on `Toolset` already iterate over `t.enabled` (including unstarted entries) and push updates through, so the retry path has current handlers
- `context.Canceled` handling in both `handleEnable` and the `Tools()` loop → correctly routes to `disableAfterDecline` in all affected code paths
- `disableAfterDecline` concurrent-access guard (`current == wrapped` pointer check) → correct
- Test coverage for the new synchronous Start path → comprehensive (success, OAuthDeclined, AuthorizationRequired, context.Canceled, retry branches all covered)
- Documentation update in `docs/tools/mcp-catalog/index.md` → consistent with new contract

dgageot (approved):

PR #2975, opened 2026-06-02
refactor: reuse registry session for OCI pulls

dgageot:
Anonymous-first pull with credential fallback on 401/403/429. This change reuses a single `remote.Puller` across digest, manifest, and layer downloads to avoid duplicate authentication attempts and improve performance.

The refactored implementation preserves platform digest behavior and retries credentialed storage on lazy layer authentication and rate-limit failures. By maintaining one session state, we eliminate redundant auth flows while keeping the same error recovery semantics.

Validation: `task lint`, `go test -race ./pkg/remote/...`, `task build`, `task test`.

### DIFF
--- a/pkg/remote/pull.go
+++ b/pkg/remote/pull.go
@@ -2,10 +2,17 @@ package remote
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
+	"slices"
 
+	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 
 	"github.com/docker/docker-agent/pkg/content"
 )
@@ -33,16 +40,27 @@ func IsDigestReference(registryRef string) bool {
 	return ok
 }
 
-// Pull pulls an artifact from a registry and stores it in the content store
+// Pull pulls an artifact from a registry and stores it in the content store.
+//
+// The digest check, manifest read, and layer downloads all reuse a single
+// authenticated registry session (a remote.Puller) so the token exchange and
+// underlying connection are established only once, rather than re-doing
+// authentication for each step.
 func Pull(ctx context.Context, registryRef string, force bool, opts ...crane.Option) (string, error) {
 	opts = append(opts, crane.WithContext(ctx), crane.WithTransport(NewTransport(ctx)))
+	o := crane.GetOptions(opts...)
 
-	ref, err := name.ParseReference(registryRef)
+	ref, err := name.ParseReference(registryRef, o.Name...)
 	if err != nil {
 		return "", fmt.Errorf("parsing registry reference %s: %w", registryRef, err)
 	}
 
-	remoteDigest, err := crane.Digest(ref.String(), opts...)
+	s, err := newSession(o)
+	if err != nil {
+		return "", fmt.Errorf("creating registry session: %w", err)
+	}
+
+	remoteDigest, err := s.digest(ctx, ref)
 	if err != nil {
 		return "", fmt.Errorf("resolving remote digest for %s: %w", registryRef, err)
 	}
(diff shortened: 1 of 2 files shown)


docker-agent (commented):

The session-based anonymous-first pull refactor is well-structured overall. Two issues in the newly added code were confirmed or likely:

1. **`storeArtifact` error swallow (line 115)** — When `retryWithCredentials()` fails, the original store error is returned instead of the auth error, hiding the real failure cause from callers.
2. **Fragile `opts.Remote[0]` assumption (line 250)** — The anonymous/credentialed auth swap relies on an undocumented crane internal ordering that could silently break on a dependency upgrade.

The concurrency concern (session mutation from concurrent layer fetches) was investigated and **dismissed** — the call path is sequential and layer fetches use the descriptor's captured fetcher, not `s.puller` directly.

pkg/remote/pull.go
```diff
@@ -77,14 +95,180 @@ func Pull(ctx context.Context, registryRef string, force bool, opts ...crane.Opt
...
+func storeArtifact(ctx context.Context, store *content.Store, s *session, ref name.Reference, localRef string, img v1.Image) (string, error) {
+	digest, err := store.StoreArtifact(img, localRef)
+	if err == nil {
+		return digest, nil
+	}
+	if s.fellBack || !shouldRetryWithCredentials(err) {
+		return "", err
+	}
+	if perr := s.retryWithCredentials(); perr != nil {
+		return "", err
```
docker-agent: **[MEDIUM] `storeArtifact` silently discards the auth-retry error (`perr`) and returns the original store error (`err`)**

When `s.retryWithCredentials()` fails (e.g., because no credentials are configured or the keychain returns an error), the function returns `err` — the original transport/store error that triggered the retry — instead of `perr`, the error explaining *why* the credential acquisition failed.

```go
if perr := s.retryWithCredentials(); perr != nil {
    return "", err   // ← should be `perr`
}
```

**Impact**: A caller debugging a failed pull would see something like `"401 Unauthorized"` (the store error) rather than `"no credentials configured for registry"` (the actual reason retrying is impossible). The real root cause is hidden, making this class of failure very hard to diagnose.

**Fix**: Return `perr` when `retryWithCredentials` fails:
```go
if perr := s.retryWithCredentials(); perr != nil {
    return "", perr
}
```

pkg/remote/pull.go
```diff
@@ -77,14 +95,180 @@ func Pull(ctx context.Context, registryRef string, force bool, opts ...crane.Opt
...
+
+// remoteOptions builds the options for a puller, reusing the exact remote.Option
+// slice crane assembled (transport, context, platform, user-agent, jobs, ...).
+// crane keeps the authenticator at index 0, so we only override that entry:
+// authn.Anonymous for the first attempt, or the keychain (crane's original
+// option) for the credentialed fallback.
+func (s *session) remoteOptions(auth authn.Authenticator) []remote.Option {
+	opts := slices.Clone(s.opts.Remote)
+	if auth != nil && len(opts) > 0 {
+		opts[0] = remote.WithAuth(auth)
```
docker-agent, line 250: **[MEDIUM] Fragile undocumented assumption that `crane` places the authenticator at `opts.Remote[0]`**

`remoteOptions` overrides `opts[0]` to swap in `authn.Anonymous` for the first attempt, relying on an **internal crane convention** that the authenticator is always placed at index 0 of the `Remote` slice:

```go
func (s *session) remoteOptions(auth authn.Authenticator) []remote.Option {
    opts := slices.Clone(s.opts.Remote)
    if auth != nil && len(opts) > 0 {
        opts[0] = remote.WithAuth(auth)   // ← relies on index 0 = auth
    }
    return opts
}
```

`remote.Option` is an opaque `func(*remote.Options)` — there is no type-safe way to identify which entry is the authenticator. If crane ever reorders its options (or an intermediate `crane.WithXxx` call inserts an option before the authenticator), this code silently overrides the *wrong* option: the anonymous override is lost and the original keychain credentials are used from the very first request, defeating the anonymous-first design. Conversely, when falling back (`auth == nil`, leaving `opts[0]` untouched), the assumption is the same.

**Suggestion**: Consider building the `remote.Option` slice from scratch rather than mutating crane's pre-assembled slice, or add a fast assertion / integration test that verifies the index-0 assumption against the actual crane version in use. At minimum, a comment linking to the specific crane source location would make the coupling explicit and easier to audit on dependency upgrades.

aheritier (approved):

PR #2981, opened 2026-06-03
fix: use cmd.Context() instead of context.Background()

dgageot:
During an audit of `context.Background()` usage in the codebase, this was identified as the one genuinely suspicious instance. The shell-completion handler for `completeMessage` was passing `context.Background()` to `config.Load()`, which meant completion operations ignored the cobra command's own context and could not be cancelled or respect deadline constraints.

This change passes `cmd.Context()` instead, ensuring that shell completion respects cancellation and any deadlines set by the caller. The now-unused `context` import was removed.

All other `context.Background()` occurrences in the codebase were reviewed and deemed intentional (detached long-lived resources, `sync.Once` caches, fire-and-forget calls), so no further changes are needed.

### DIFF
--- a/cmd/root/completion.go
+++ b/cmd/root/completion.go
@@ -1,7 +1,6 @@
 package root
 
 import (
-	"context"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, well-scoped fix. The change correctly replaces `context.Background()` with `cmd.Context()` in the `completeMessage` shell-completion handler.

**Analysis:**
- `cmd.Context()` in Cobra always returns a non-nil context (falls back to `context.Background()` if none is explicitly set), so this is strictly safer than a bare `context.Background()`.
- The existing error handling (`return nil, cobra.ShellCompDirectiveNoFileComp` on `config.Load` error) already handles any context-cancellation-driven errors gracefully.
- The `context` import removal is correct — no other usages of the `context` package remain in the file.
- The change is minimal, targeted, and correct. No bugs were found.

rumpl (approved):

PR #2982, opened 2026-06-03
feat: add ContextLimit helper to modelinfo

dgageot:
Model context-window handling was scattered across providers. Each provider either hardcoded a limit (Anthropic had `anthropicContextLimit` embedded in its package) or lacked a fallback when the models.dev store was unavailable, risking silent failures or degraded behavior in production environments.

This change centralizes context-window logic into a new `ContextLimit(ctx, store, id, fallback)` helper in `pkg/modelinfo`. It first attempts to source the window from the models.dev store with a bounded timeout, then falls back to a provided default. A new `DefaultAnthropicContextLimit` constant (200000) replaces Anthropic's hardcoded function, and all five provider packages now use `modelinfo.ContextLimit` consistently.

Additionally, `LoadCaps` now accepts a `context.Context` as its first parameter so callers can implement proper timeout behavior. This is wrapped internally with `loadCapsTimeout` to bound the underlying models.dev load operation, though callers should understand that the timeout does not cover contention from the shared store lock.

### DIFF
--- a/pkg/model/provider/anthropic/attachments.go
+++ b/pkg/model/provider/anthropic/attachments.go
@@ -24,7 +24,7 @@ import (
 //   - text with InlineText → TextBlockParam with TXTEnvelope
 //   - unsupported / no content → nil (logged as warning)
 func convertDocument(ctx context.Context, doc chat.Document, id modelsdev.ID, store *modelsdev.Store) ([]anthropic.ContentBlockParamUnion, error) {
-	mc := modelinfo.LoadCaps(store, id)
+	mc := modelinfo.LoadCaps(ctx, store, id)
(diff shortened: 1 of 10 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `ContextLimit` helper is implemented correctly. The nil-store early-return, bounded `context.WithTimeout`, and `<= 0` fallback guard all work as intended. The `DefaultAnthropicContextLimit` constant migration is type-safe (untyped integer constant implicitly converts to `int64`). All five provider `convertDocument` call-sites are updated consistently. The `LoadCaps` context threading is correct. Tests cover the happy path, unknown model, and nil-store cases.

trungutt (approved):

PR #2983, opened 2026-06-03
fix: prevent memory explosion in search_files_content

dgageot:
The `search_files_content` filesystem tool can produce massive results when searching across large directory trees, causing unbounded memory growth as the output cascades through multiple layers: tool output buffer → in-memory message list → session store/WAL → re-serialization into every subsequent model request for the rest of the session.

This change caps the total output at 1 MiB with a truncation notice and skips files larger than 10 MiB. Output is now built incrementally with `strings.Builder` instead of accumulating a slice and joining at the end. The file-size check now stats the target (following symlinks, not the symlink itself) and only scans regular files, which also prevents reading forever from device files or FIFOs like `/dev/zero`.

Tests cover output-cap truncation, large-file skipping, and the symlink-to-large-file case.

### DIFF
--- a/pkg/tools/builtin/filesystem/filesystem.go
+++ b/pkg/tools/builtin/filesystem/filesystem.go
@@ -1126,12 +1126,18 @@ func (t *ToolSet) handleSearchFilesContent(_ context.Context, args SearchFilesCo
 		}
 	}
 
-	var results []string
+	var out strings.Builder
 	filesWithMatches := make(map[string]struct{})
+	matchCount := 0
+	truncated := false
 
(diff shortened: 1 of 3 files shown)


gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are correct and well-implemented. The output cap (1 MiB), large-file skip (10 MiB), symlink-aware `stat` guard, and incremental `strings.Builder` pattern all work as intended. Tests cover truncation, large-file skipping, and the symlink-to-large-file regression case.

PR #2984, opened 2026-06-03
refactor: remove empty query truncation from MCP server search

dgageot:
The `search_remote_mcp_servers` tool truncated results for empty queries to 25 servers, ostensibly to avoid bloating the LLM's context window. In practice, this hides relevant servers and requires users to already know keywords to discover them.

This change removes the truncation logic entirely. An empty query now returns all matching servers from the catalog. Models can still make intelligent choices about how many results to present to a user, and they're not artificially blinded by the tool's limits.

The test expectations were updated accordingly.

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
+++ b/pkg/tools/builtin/mcpcatalog/mcpcatalog.go
@@ -626,12 +626,6 @@ type SearchResult struct {
 	Enabled     bool     `json:"enabled"`
 }
 
-// emptyQuerySearchLimit caps the result set for an empty query so the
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor cleanly removes the `emptyQuerySearchLimit = 25` truncation constant and all associated branching logic. The updated result path is correct — `len(matches)` is used after the sort, which is the right value now that no slicing occurs. The test update is well-structured: it asserts the returned count equals `ts.catalog.Count` (the full catalog), which is a stronger invariant than the old fixed-length check.

No bugs, regressions, or logic errors found in the changed lines.

trungutt (approved):

PR #2986, opened 2026-06-03
refactor: use strings.Builder for message merging in oaistream

dgageot:
Consecutive same-role messages in the OpenAI stream handler were being merged by concatenating strings with `+=` in a loop. This causes repeated allocations and copying as the accumulated content grows.

The `MergeConsecutiveMessages` function now uses `strings.Builder` instead, which buffers the concatenation and reduces allocations to a single final operation. The change is strictly behavior-preserving: the condition `mergedContent.Len() > 0` replaces `mergedContent != ""`, and `mergedContent.String()` yields the same accumulated value.

Testing confirmed the change passes `go build ./...`, unit tests in `pkg/model/provider/oaistream/`, and golangci-lint with zero issues.

### DIFF
--- a/pkg/model/provider/oaistream/messages.go
+++ b/pkg/model/provider/oaistream/messages.go
@@ -270,7 +270,7 @@ func MergeConsecutiveMessages(openaiMessages []openai.ChatCompletionMessageParam
 
 		// Only merge system or user messages
 		if currentRole == "system" || currentRole == "user" {
-			var mergedContent string
+			var mergedContent strings.Builder
 			var mergedMultiContent []openai.ChatCompletionContentPartUnionParam
 			j := i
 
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `strings.Builder` refactor in `MergeConsecutiveMessages` is **behaviorally equivalent** to the original `+=` concatenation in all code paths:

- `var mergedContent strings.Builder` is declared **inside** the per-group `if` block, so it is freshly zero-valued on every outer-loop iteration — no cross-group state leakage.
- `mergedContent.Len() > 0` is semantically identical to `mergedContent != ""` for the newline-separator guard.
- `mergedContent.String()` on a never-written builder returns `""`, matching the old zero-value string — so calls to `openai.SystemMessage("")` / `openai.UserMessage("")` when only multi-content parts exist are unaffected.
- `WriteString("")` (for an empty `str` returned by `getStringContent`) is a no-op, matching the old `mergedContent += ""` behaviour.

No bugs or regressions identified. The change correctly reduces allocations without altering observable behaviour.

rumpl (approved):

PR #2987, opened 2026-06-03
fix(mcp): harden managed OAuth flow (decline sentinel + bounded wait)

dgageot:
## What

Fixes two correctness bugs in the **managed** MCP OAuth flow (`pkg/tools/mcp/oauth.go`), which had not received the hardening the unmanaged path already has.

## Why

While scanning the MCP code for bugs, leaks, and context-handling issues, the managed OAuth path stood out as missing two protections the unmanaged path implements:

1. **Unbounded interactive wait → pinned session lock.** `handleManagedOAuthFlow` waited on the detached `clientConnector.Connect` context (`context.WithoutCancel`), whose `Done()` never fires. An abandoned browser tab or a silent MCP-client disconnect would block the elicitation/callback wait forever, holding the per-session streaming lock and turning every subsequent user message into a `409 / ErrSessionBusy` until a process restart.

2. **Decline returned a generic error.** On user decline the flow returned `errors.New("user declined OAuth authorization")`, so the sticky-decline latch (`authorizeOnce`), `enrichConnectError`, and mcpcatalog's `IsOAuthDeclined` short-circuit could not recognise it — re-popping the dialog the user just dismissed and re-entering the `Tools() → Start() → elicitation` retry loop the sentinel exists to prevent.

## What changed

- **`fix(mcp): return OAuthDeclinedError when managed OAuth is declined`** — return the recognisable `*OAuthDeclinedError{URL: ...}` sentinel on decline, matching the unmanaged path.
- **`fix(mcp): bound managed OAuth interactive wait and observe cancellation`** — derive a `waitCtx` with `oauthInteractiveWaitTimeout` and tie it to the caller's original ctx via `context.AfterFunc`, so the timeout and user-initiated "Stop" both unblock the flow. Mirrors `handleUnmanagedOAuthFlow`.
- **`test(mcp): cover managed OAuth decline/timeout; rename shared wait timeout`** — renamed `unmanagedOAuthWaitTimeout` → `oauthInteractiveWaitTimeout` (it is now shared by both flows) and added `TestManagedOAuthFlow_ElicitationDeclineReturnsSentinel` and `TestManagedOAuthFlow_TimesOutWhenNoReplyArrives` (the managed flow previously had no test coverage).

## Testing

- `task test` — all packages pass.
- `task lint` — 0 issues (golangci-lint + custom cops + `go mod tidy`).

### DIFF
--- a/pkg/tools/mcp/oauth.go
+++ b/pkg/tools/mcp/oauth.go
@@ -849,7 +849,12 @@ func (t *oauthTransport) handleManagedOAuthFlow(ctx context.Context, authServer,
 	slog.DebugContext(ctx, "Elicitation response received", "result", result)
 
 	if result.Action != tools.ElicitationActionAccept {
-		return errors.New("user declined OAuth authorization")
+		// Surface the recognisable sentinel (mirroring the unmanaged
+		// path) so authorizeOnce can latch the decline and queued
+		// initialize-stage goroutines don't re-pop the dialog the user
+		// just dismissed, and so enrichConnectError / mcpcatalog's
+		// IsOAuthDeclined short-circuit can break the retry loop.
+		return &OAuthDeclinedError{URL: t.baseURL}
 	}
 
 	slog.DebugContext(ctx, "Requesting authorization code", "url", authURL)
--- a/pkg/tools/mcp/oauth.go
+++ b/pkg/tools/mcp/oauth.go
@@ -27,13 +27,13 @@ import (
 // resourceMetadataFromWWWAuth extracts resource metadata URL from WWW-Authenticate header
 var re = regexp.MustCompile(`resource="([^"]+)"`)
 
-// unmanagedOAuthWaitTimeout is the upper bound on how long the unmanaged
-// OAuth flow blocks waiting for a reply (elicitation result or
-// out-of-band callback). Generous enough to accommodate a user clicking
-// through an IdP consent screen and any IdP-side prompts; small enough
-// that a silently-disconnected MCP client can't hold the per-session
-// streaming lock indefinitely.
-var unmanagedOAuthWaitTimeout = 10 * time.Minute
+// oauthInteractiveWaitTimeout is the upper bound on how long an interactive
+// OAuth flow (managed or unmanaged) blocks waiting for a reply (elicitation
+// result, browser callback, or out-of-band callback). Generous enough to
+// accommodate a user clicking through an IdP consent screen and any IdP-side
+// prompts; small enough that a silently-disconnected MCP client can't hold
+// the per-session streaming lock indefinitely.
+var oauthInteractiveWaitTimeout = 10 * time.Minute
 
 // oauth is a simple struct for compatibility with existing code
 type oauth struct {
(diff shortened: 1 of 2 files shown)


pkg/tools/mcp/oauth.go
```diff
@@ -849,12 +864,17 @@ func (t *oauthTransport) handleManagedOAuthFlow(ctx context.Context, authServer,
...
+		// initialize-stage goroutines don't re-pop the dialog the user
+		// just dismissed, and so enrichConnectError / mcpcatalog's
+		// IsOAuthDeclined short-circuit can break the retry loop.
+		return &OAuthDeclinedError{URL: t.baseURL}
 	}
 
 	slog.DebugContext(ctx, "Requesting authorization code", "url", authURL)
 
-	code, receivedState, err := RequestAuthorizationCode(ctx, authURL, callbackServer, state)
+	code, receivedState, err := RequestAuthorizationCode(waitCtx, authURL, callbackServer, state)
```
docker-agent, line 877: **[LOW] Shared `waitCtx` leaves little time for the authorization-code exchange if elicitation is accepted late**

`waitCtx` is a single 10-minute deadline covering both the `requestElicitation` call and the subsequent `RequestAuthorizationCode` call. If a user takes close to the full timeout to approve the elicitation dialog (e.g. 9m 55s), only a few seconds remain for the browser redirect and OAuth callback — causing the code-exchange to fail with `context deadline exceeded` despite the user having correctly approved the request.

The unmanaged flow avoids this by running the elicitation in a goroutine with its own context (`elicCtx`) and separately selecting on `callbackCh`, so the callback step effectively gets the remaining wall-clock time independently. Consider resetting or extending the deadline after elicitation completes, or use separate timeouts for each phase.

In practice this only manifests when a user delays acceptance until very close to the 10-minute boundary, so the risk is low.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The two correctness fixes are sound:

- **Decline sentinel** — returning `&OAuthDeclinedError{URL: t.baseURL}` correctly mirrors the unmanaged path. The `authorizeOnce` latch uses `errors.As` to detect this sentinel, `enrichConnectError` uses a boolean flag (`lastOAuthDeclined`) rather than string matching, and `IsOAuthDeclined` / `mcpcatalog` use the same `errors.As` pattern — all of which work correctly with the new return value.
- **Bounded wait** — `context.WithTimeout(ctx, oauthInteractiveWaitTimeout)` works correctly even when `ctx` is a `WithoutCancel` context (the deadline mechanism is independent of parent cancellation). The `defer context.AfterFunc(parentCtx, cancelWait)()` pattern is idiomatic: the stop-function is deferred so it runs on exit, preventing a goroutine leak when `parentCtx` hasn't fired. Double-calling `cancelWait` is safe in Go.
- **Tests** — both new tests (`TestManagedOAuthFlow_ElicitationDeclineReturnsSentinel` and `TestManagedOAuthFlow_TimesOutWhenNoReplyArrives`) correctly exercise the new code paths.

One minor design note is left as an inline comment.

PR #2989, opened 2026-06-03
test(mcp): fix staticcheck SA5011 nil-pointer errors in oauth_test

dgageot:
The main branch CI was failing with 6 golangci-lint staticcheck SA5011 ("possible nil pointer dereference") errors in oauth_test. These were triggered by three `if got == nil { t.Fatal(...) }` guards that staticcheck did not recognize as terminating the execution path, so it flagged subsequent `got.AccessToken` accesses as potentially accessing a nil pointer.

Replacing these three guards with `require.NotNil(t, got, ...)` from testify removes the explicit nil comparison that trips staticcheck, while preserving the same fail-fast behavior and readability. The require helper is already in use throughout the test file.

Tests and linter pass locally.

### DIFF
--- a/pkg/tools/mcp/oauth_test.go
+++ b/pkg/tools/mcp/oauth_test.go
@@ -159,9 +159,7 @@ func TestGetValidToken_UsesStoredCredentialsForRefresh(t *testing.T) {
 	}
 
 	got := transport.getValidToken(t.Context())
-	if got == nil {
-		t.Fatal("getValidToken returned nil, expected refreshed token")
-	}
+	require.NotNil(t, got, "getValidToken returned nil, expected refreshed token")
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR is a clean, targeted lint fix. The three `if got == nil { t.Fatal(...) }` guards are replaced with semantically equivalent `require.NotNil(t, got, ...)` calls from testify. Since `require.NotNil` calls `t.FailNow()` internally, the fail-fast behavior is preserved. No logic changes, no new control-flow paths, no resource management issues, and no security implications. The fix correctly resolves the staticcheck SA5011 false positives.

**No bugs found in the changed code.**

rumpl (approved):

PR #2991, opened 2026-06-03
feat: add first_available model selection

dgageot:
This adds a `first_available` model selector to the config schema. It enables automatic fallback across multiple model candidates by resolving the first reachable one with configured credentials, including support for local dmr and ollama fallbacks.

The implementation automatically tries each candidate in order until one succeeds based on available credentials. If no candidates can be resolved, it reports grouped missing-credential errors indicating that at least one candidate group is needed. During alias resolution, unresolved selectors are skipped, allowing other aliases to proceed.

The change includes an example configuration, comprehensive tests, and all necessary schema updates.

## Validation

```bash
go test ./pkg/config/... ./pkg/teamloader/... ./pkg/runtime/...
task build
task lint
```

### DIFF
--- a/pkg/config/first_available.go
+++ b/pkg/config/first_available.go
@@ -1,0 +1,282 @@
+package config
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strings"
+
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/environment"
+	modelprovider "github.com/docker/docker-agent/pkg/model/provider"
+)
+
+// ResolveFirstAvailableModels resolves reachable `first_available` models in the
+// config into concrete provider/model definitions by selecting the first
+// candidate whose credentials are configured. The config is mutated in place so
+// the rest of the pipeline (env-var gathering, model instantiation, runtime
+// switching) sees regular model definitions for models that can be reached from
+// agents, fallbacks, tool model overrides, skills, RAG reranking, or routing.
+//
+// Candidates are tried in order. A candidate is considered available when all
+// the environment variables it requires are set; local providers (dmr, ollama)
+// require none and therefore make reliable final fallbacks. When a models
+// gateway is configured, the first candidate is selected since the gateway
+// supplies the credentials.
+func ResolveFirstAvailableModels(ctx context.Context, cfg *latest.Config, modelsGateway string, env environment.Provider) error {
+	// Snapshot which models are selectors before mutating the map, so the
+	// nested-selector check is independent of map iteration order.
+	selectors := map[string]bool{}
+	for name, m := range cfg.Models {
+		if m.IsFirstAvailable() {
+			selectors[name] = true
+		}
+	}
+	if len(selectors) == 0 {
+		return nil
+	}
+
+	reachable := reachableFirstAvailableModels(cfg, selectors)
+	for _, name := range sortedKeys(reachable) {
+		if err := resolveFirstAvailableModel(ctx, cfg, name, selectors, modelsGateway, env); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func resolveFirstAvailableModel(ctx context.Context, cfg *latest.Config, name string, selectors map[string]bool, modelsGateway string, env environment.Provider) error {
+	m := cfg.Models[name]
+	chosen, ref, err := selectFirstAvailable(ctx, cfg, m.FirstAvailable, selectors, modelsGateway, env)
+	if err != nil {
+		var missingErr *firstAvailableMissingEnvError
+		if errors.As(err, &missingErr) {
+			return missingErr
+		}
+		return fmt.Errorf("model '%s': %w", name, err)
+	}
+
+	slog.DebugContext(ctx, "Resolved first_available model",
+		"model_name", name, "selected", ref, "provider", chosen.Provider, "model", chosen.Model)
+	cfg.Models[name] = chosen
+	return nil
+}
+
+func reachableFirstAvailableModels(cfg *latest.Config, selectors map[string]bool) map[string]bool {
+	reachable := map[string]bool{}
+	seen := map[string]bool{}
+
+	hasRootReference := false
+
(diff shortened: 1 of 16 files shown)


gtardif (dismissed):

pkg/config/first_available.go
```diff
@@ -0,0 +1,282 @@
...
+
+	hasRootReference := false
+
+	var visit func(string)
+	visit = func(ref string) {
+		ref = strings.TrimSpace(ref)
+		if ref == "" || seen[ref] {
+			return
+		}
+		hasRootReference = true
```
docker-agent, line 79: **[MEDIUM] `hasRootReference` set for any non-empty ref, including refs to non-existent models — `first_available` selectors silently go unresolved**

In `reachableFirstAvailableModels`, `hasRootReference` is set to `true` for every non-empty model reference encountered in `visit`, unconditionally — before the `model, exists := cfg.Models[ref]` check. If an agent's toolset, skill, or fallback references a model name that does **not** exist in `cfg.Models`, `hasRootReference` becomes `true` but `reachable` stays empty (the early `return` on `!exists` prevents adding it to `reachable`).

The function then returns the empty `reachable` map (line 196) rather than the full `selectors` map, causing **all `first_available` selectors to silently go unresolved** — they remain in `cfg.Models` as `{FirstAvailable: [...]}` entries that no model instantiation code knows how to handle.

**Trigger**: Any config where an agent references a non-existent model name in a toolset/skill alongside a `first_available` selector (e.g. a stale or typo'd model ref).

**Fix**: Only set `hasRootReference = true` after confirming `cfg.Models[ref]` exists, or move the flag-set after the `!exists` guard:
```go
model, exists := cfg.Models[ref]
if !exists {
    return
}
hasRootReference = true  // only count refs to known models
```

pkg/config/gather.go
```diff
@@ -105,43 +105,52 @@ func addEnvVarsForModelConfig(model *latest.ModelConfig, customProviders map[str
...
 	case "anthropic":
 		requiredEnv["ANTHROPIC_API_KEY"] = true
 	case "google":
 		if model.ProviderOpts["project"] == nil && model.ProviderOpts["location"] == nil {
-			if os.Getenv("GOOGLE_GENAI_USE_VERTEXAI") != "" {
+			if value, _ := env.Get(ctx, "GOOGLE_GENAI_USE_VERTEXAI"); value != "" {
 				requiredEnv["GOOGLE_CLOUD_PROJECT"] = true
 				requiredEnv["GOOGLE_CLOUD_LOCATION"] = true
-			} else if _, exist := os.LookupEnv("GEMINI_API_KEY"); !exist {
+			} else if value, _ := env.Get(ctx, "GEMINI_API_KEY"); value == "" {
```
docker-agent, line 153: **[LOW] Empty `GEMINI_API_KEY` now treated as absent — potential spurious `GOOGLE_API_KEY` requirement**

The old code used `os.LookupEnv("GEMINI_API_KEY")` which distinguishes between *unset* and *set-but-empty*: `exist == true` for an empty-string value, so `GOOGLE_API_KEY` was **not** added to required env vars.

The new code uses `env.Get(ctx, "GEMINI_API_KEY"); value == ""`, which treats an empty-string `GEMINI_API_KEY` the same as an unset one. This could cause a spurious "missing `GOOGLE_API_KEY`" error for users who have `GEMINI_API_KEY=` (explicitly set to empty) as a way to signal "no key" without triggering the fallback.

In the context of `first_available`, treating an empty key as absent is arguably correct (an empty API key is not a usable credential). However, this is a silent behavioral change from the previous semantics that could surface in existing configurations.

pkg/teamloader/teamloader.go
```diff
@@ -121,6 +119,18 @@ func LoadWithConfig(ctx context.Context, agentSource config.Source, runConfig *c
 
 	// Early check for required env vars before loading models and tools.
 	env := runConfig.EnvProvider()
+
+	// Resolve `first_available` model selectors into concrete provider/model
+	// definitions now that the environment is available, so the rest of the
+	// pipeline sees regular model definitions.
+	if err := config.ResolveFirstAvailableModels(ctx, cfg, runConfig.ModelsGateway, env); err != nil {
```
docker-agent, line 126: **[MEDIUM] `ApplyModelOverrides` runs before `ResolveFirstAvailableModels` — CLI model override for a `first_available` selector name bypasses resolution**

`ApplyModelOverrides` (line 116) rewrites `agent.Model` string references based on CLI flags, then `ResolveFirstAvailableModels` (line 126) builds the `reachable` set by walking agent model references.

If a user's `first_available` selector is named `"smart"` and they pass `--model agent1=openai/gpt-5`:
- `ApplyModelOverrides` changes `agent1.Model` from `"smart"` → `"openai/gpt-5"` ✓ (intended)
- `reachableFirstAvailableModels` visits `"openai/gpt-5"` — not `"smart"` — so `"smart"` is never added to `reachable`
- `ResolveFirstAvailableModels` skips `"smart"` entirely

This is acceptable when the override fully specifies the target model. However, if **any other agent** still references `"smart"` and the config relies on `"smart"` being resolved (e.g. for routing or env-var gathering), the entry in `cfg.Models["smart"]` remains as a raw `{FirstAvailable: [...]}` struct — which downstream code (model instantiation, `CheckRequiredEnvVars`) may not handle correctly.

Additionally, if a user tries to test `first_available` fallback behavior by passing a partial override, the interaction is non-obvious. The ordering should be documented or `ResolveFirstAvailableModels` should be moved before `ApplyModelOverrides`.

docker-agent (commented):

The `first_available` model selector implementation is well-structured with good test coverage. Three issues were found in the changed code:

| Severity | Finding |
|----------|---------|
| 🟡 Medium | `hasRootReference` set for non-existent model refs — selectors silently skipped |
| 🟡 Medium | `ApplyModelOverrides` ordering may leave unresolved `first_available` entries visible to downstream code |
| ⚪ Low | Empty `GEMINI_API_KEY` now treated as absent (behavior change from `os.LookupEnv`) |

gtardif (approved):

PR #2992, opened 2026-06-03
fix: don't drop tool call when finish_reason shares the chunk

### LINKED ISSUE #1544: Subagent tool_calls not handled when using LiteLLM proxy
## Summary
When using cagent with a LiteLLM proxy (OpenAI-compatible API), a subagent can return a valid `tool_calls` response, but the host agent reports “no response from subagent” and does not execute the tool call.

## Environment
- cagent: v1.19.7
- OS: Linux (WSL2) x86_64
- LLM: gemini-pro via LiteLLM proxy (OpenAI-compatible)

## Expected
Host agent receives the subagent response and executes the tool call(s), then continues the conversation with the tool result.

## Actual
Host agent reports there is “no response from subagent” even though the subagent response contains a valid `tool_calls` entry. The requested tool call is not executed.

## Minimal reproduction
1. Configure cagent to use LiteLLM as the model provider (OpenAI-compatible API).
2. Run a workflow where the subagent is asked to use tools.
3. The subagent returns a tool call (e.g., `company_search` with JSON arguments) and finishes with `finish_reason: "stop"`.
4. The host agent logs “no response from subagent” and does not invoke the tool.

## Notes
- The subagent response includes a valid tool call object (`type: function`, `name: company_search`, `arguments: {"datasources": ["confluence", "github", "gdrive"], "query": "LiteLLM usage teams integrations standards"}`).
- I can reproduce consistently with LiteLLM as the proxy in front of an OpenAI-compatible endpoint.

Please let me know if you want a sanitized example payload or a minimal agent config.

Sayt-0:
Fixes #1544

## Problem
With OpenAI-compatible proxies like LiteLLM (e.g. in front of Gemini), a subagent's valid `tool_calls` response is dropped and the host agent reports "no response from subagent" without executing the tool.

## Root cause
`handleStream` (pkg/runtime/streaming.go) checked the terminal `finish_reason` **before** accumulating the current chunk's tool call deltas. The OpenAI-native protocol sends them in separate chunks, but LiteLLM/Gemini packs the tool call and the terminal `finish_reason` into a single chunk. The early return then saw `len(toolCalls) == 0` and dropped the call, leaving an empty assistant message.

## Fix
Accumulate tool call deltas before evaluating the finish reason; only short-circuit the iteration when the chunk has no terminal finish reason.

## Tests
- `TestHandleStream_ToolCallAndStopInSameChunk` — regression test (red before, green after).
- `TestHandleStream_ToolCallThenSeparateStop` — guards the OpenAI-native path.

### DIFF
--- a/pkg/runtime/streaming.go
+++ b/pkg/runtime/streaming.go
@@ -129,33 +129,12 @@ func handleStream(ctx context.Context, stream chat.MessageStream, a *agent.Agent
 			thoughtSignature = choice.Delta.ThoughtSignature
 		}
 
-		if choice.FinishReason == chat.FinishReasonStop || choice.FinishReason == chat.FinishReasonLength {
-			recordUsage()
-			applyXMLFallback()
-			finishReason := choice.FinishReason
-			if finishReason == chat.FinishReasonStop && len(toolCalls) > 0 {
-				finishReason = chat.FinishReasonToolCalls
-			}
-			return streamResult{
-				Calls:             toolCalls,
-				Content:           fullContent.String(),
-				ReasoningContent:  fullReasoningContent.String(),
-				ThinkingSignature: thinkingSignature,
-				ThoughtSignature:  thoughtSignature,
-				Stopped:           len(toolCalls) == 0, // stop only when there are no tool calls to execute
-				FinishReason:      finishReason,
-				Usage:             messageUsage,
-			}, nil
-		}
-
-		// Track the provider's explicit finish reason (e.g. tool_calls) so we
-		// can prefer it over inference after the loop.  stop/length are already
-		// handled by the early return above.
-		if choice.FinishReason != "" {
-			providerFinishReason = choice.FinishReason
-		}
-
-		// Handle tool calls
+		// Accumulate tool call deltas from this chunk *before* evaluating the
+		// finish reason below. Some OpenAI-compatible providers (e.g. LiteLLM
+		// in front of Gemini) pack a complete tool call and a terminal
+		// finish_reason ("stop") into the same chunk; handling the finish
+		// reason first would drop the call and the turn would end with an
+		// empty assistant message ("No response from agent").
 		if len(choice.Delta.ToolCalls) > 0 {
 			// Process each tool call delta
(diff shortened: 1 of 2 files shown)


pkg/runtime/streaming.go
```diff
@@ -208,7 +187,38 @@ func handleStream(ctx context.Context, stream chat.MessageStream, a *agent.Agent
 					}
 				}
 			}
-			continue
+			// Short-circuit only when the chunk carried nothing but tool call
+			// deltas. If it also carries a terminal finish reason, fall through
+			// so the early return below sees the freshly accumulated call.
+			if choice.FinishReason == "" {
```
docker-agent, line 193: **[LOW] Unrecognized `finish_reason` with tool-call deltas silently falls through**

When a provider sends a non-empty `finish_reason` that is neither `stop`, `length`, nor `tool_calls` (e.g. `"content_filter"`, `"max_tokens"`) **in the same chunk as tool-call deltas**, the new control flow neither `continue`s (because `FinishReason != ""`) nor hits the early `return` (because it's not `stop`/`length`). The chunk falls through to the `providerFinishReason` tracking block and then continues into the content-append branches, after which the loop goes on to the next iteration. At end-of-stream the post-loop logic picks up `providerFinishReason = "content_filter"`, which the switch doesn't recognise, likely yielding an incorrect or empty finish reason. The accumulated tool calls would also be lost because the post-loop path for `content_filter` doesn't inject `FinishReasonToolCalls`.

This is a low-probability edge case (requires an exotic finish reason and a same-chunk tool call), but the safe pattern would be an `else if` or a `return`/`continue` after the `FinishReasonStop`/`FinishReasonLength` block to prevent fall-through for any other non-empty finish reason when tool calls are present.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is logically correct: accumulating tool-call deltas before evaluating `finish_reason` directly addresses the LiteLLM/Gemini same-chunk bug. The two new tests cover both the combined-chunk (regression) and the OpenAI-native (guard) paths. One low-severity edge case is noted inline.

dgageot (approved):

PR #2993, opened 2026-06-03
feat: add opt-in self-update

dgageot:
docker-agent can now update itself from GitHub releases via the opt-in `DOCKER_AGENT_AUTO_UPDATE` environment variable. The implementation performs digest and checksum verification to ensure authenticity, and includes fail-safe fallback and restore mechanisms on re-exec failure. Metadata, version, help, and shell completion invocations are skipped to avoid unnecessary updates.

When a newer release is available, the user is now asked to confirm the upgrade:

```
An update is available (vX.Y.Z). Do you want to install it or continue with version vA.B.C? [Y/n]
```

- Interactive sessions (stdin is a terminal) get the prompt; the default answer is yes.
- Non-interactive sessions (CI, piped input, redirected stdin) auto-confirm so automation keeps working.
- Declining keeps the current binary untouched.

The feature is fully tested and documented, with implementation details covering Unix and Windows platforms separately. It integrates seamlessly with the existing agent lifecycle and respects configuration preferences.

All changes are backward compatible. Existing users will not be affected unless they explicitly enable the auto-update feature.

### DIFF
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/docker-agent/pkg/feedback"
 	"github.com/docker/docker-agent/pkg/logging"
 	"github.com/docker/docker-agent/pkg/paths"
+	"github.com/docker/docker-agent/pkg/selfupdate"
 	"github.com/docker/docker-agent/pkg/telemetry"
 	"github.com/docker/docker-agent/pkg/version"
 )
@@ -168,6 +169,16 @@ We collect anonymous usage data to help improve docker agent. To disable:
 }
 
 func Execute(ctx context.Context, stdin io.Reader, stdout, stderr io.Writer, args ...string) error {
+	selfupdate.Cleanup(ctx)
+
+	// Opt-in self-update: when enabled, replace this binary with the latest
+	// release and re-exec before doing any real work. Skipped for invocations
+	// where restarting would be wrong (plugin metadata handshake, shell
+	// completion) and always falls back to the current binary on any failure.
+	if selfupdate.Enabled() && !isManagementInvocation(args) {
+		selfupdate.Run(ctx, stderr)
+	}
+
 	rootCmd := NewRootCmd()
 	rootCmd.SetIn(stdin)
 	rootCmd.SetOut(stdout)
@@ -218,6 +229,23 @@ func visitAll(cmd *cobra.Command, fn func(*cobra.Command)) {
 	fn(cmd)
 	for _, cmd := range cmd.Commands() {
 		visitAll(cmd, fn)
+	}
+}
+
+// isManagementInvocation reports whether args correspond to an invocation that
+// must not trigger a self-update + restart: the docker CLI plugin metadata
+// handshake, shell-completion script generation, and the version/help queries.
+// Updating mid-handshake would corrupt the plugin protocol, and restarting a
+// completion call would be surprising.
+func isManagementInvocation(args []string) bool {
+	if len(args) == 0 {
+		return false
+	}
+	switch args[0] {
+	case metadata.MetadataSubcommandName, cobra.ShellCompRequestCmd, cobra.ShellCompNoDescRequestCmd, "completion", "version", "help", "--help", "-h", "--version":
+		return true
+	default:
+		return false
 	}
 }
 
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -176,7 +176,7 @@ func Execute(ctx context.Context, stdin io.Reader, stdout, stderr io.Writer, arg
 	// where restarting would be wrong (plugin metadata handshake, shell
 	// completion) and always falls back to the current binary on any failure.
 	if selfupdate.Enabled() && !isManagementInvocation(args) {
-		selfupdate.Run(ctx, stderr)
+		selfupdate.Run(ctx, stdin, stderr)
 	}
 
 	rootCmd := NewRootCmd()
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -237,16 +237,25 @@ func visitAll(cmd *cobra.Command, fn func(*cobra.Command)) {
 // handshake, shell-completion script generation, and the version/help queries.
 // Updating mid-handshake would corrupt the plugin protocol, and restarting a
 // completion call would be surprising.
+//
+// Help and version are detected anywhere in args, not just at args[0], so that
+// per-subcommand help (e.g. "run --help") is also skipped.
 func isManagementInvocation(args []string) bool {
 	if len(args) == 0 {
 		return false
 	}
 	switch args[0] {
-	case metadata.MetadataSubcommandName, cobra.ShellCompRequestCmd, cobra.ShellCompNoDescRequestCmd, "completion", "version", "help", "--help", "-h", "--version":
+	case metadata.MetadataSubcommandName, cobra.ShellCompRequestCmd, cobra.ShellCompNoDescRequestCmd, "completion", "version", "help", "--version":
 		return true
-	default:
-		return false
 	}
+	// A help request can appear after a subcommand ("run --help"); never update
+	// just to print help text.
+	for _, arg := range args {
+		if arg == "--help" || arg == "-h" {
+			return true
+		}
+	}
+	return false
 }
 
 // setupLogging configures slog logging behavior.
--- a/pkg/selfupdate/selfupdate.go
+++ b/pkg/selfupdate/selfupdate.go
@@ -1,0 +1,567 @@
+// Package selfupdate implements an opt-in, fail-safe self-update mechanism.
+//
+// When enabled via the [EnvAutoUpdate] environment variable, the running
+// binary checks GitHub for a newer release, downloads the asset matching the
+// current OS/arch, verifies it, atomically replaces itself on disk, and
+// re-executes the new binary with the original arguments.
+//
+// The whole mechanism is best-effort: any failure (network, disk, permissions,
+// version parsing, a corrupt download, ...) is logged and swallowed so the
+// caller always falls back to running the current binary. The only observable
+// effect of a failed update is a short delay and a log line.
+package selfupdate
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/Masterminds/semver/v3"
+
+	"github.com/docker/docker-agent/pkg/atomicfile"
+)
+
+const (
+	// EnvAutoUpdate enables the self-update mechanism when set to a truthy
+	// value ("1", "true", "yes", "on", case-insensitive).
+	EnvAutoUpdate = "DOCKER_AGENT_AUTO_UPDATE"
+
+	// envReExecMarker is set on the re-executed child process to prevent an
+	// infinite update→re-exec loop. Its presence means "an update has already
+	// been attempted in this process tree, do not try again".
+	envReExecMarker = "DOCKER_AGENT_SELF_UPDATE_REEXEC"
+
+	// envBackupMarker points the re-executed child at the previous binary backup
+	// so it can clean up after a successful restart.
+	envBackupMarker = "DOCKER_AGENT_SELF_UPDATE_BACKUP"
+
+	defaultRepoOwner = "docker"
+	defaultRepoName  = "docker-agent"
+
+	defaultAPIBaseURL      = "https://api.github.com"
+	defaultDownloadBaseURL = "https://github.com"
+
+	// devVersion marks a local/unreleased build that must never be updated.
+	devVersion = "dev"
+
+	// httpTimeout bounds every network operation. The full update budget is
+	// roughly downloadTimeout for the asset plus this for metadata.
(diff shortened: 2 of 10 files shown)


pkg/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+	if resolved, err := filepath.EvalSymlinks(exe); err == nil {
+		exe = resolved
+	}
+	return filepath.Abs(exe)
+}
+
+// Cleanup removes the previous-binary backup after a successful re-exec. It is
+// deliberately best-effort: cleanup failure must never block normal execution.
+func Cleanup(ctx context.Context) {
```
docker-agent, line 509: **[MEDIUM] Arbitrary file deletion via `DOCKER_AGENT_SELF_UPDATE_BACKUP` env var**

`Cleanup()` is called **unconditionally** at the top of `Execute()`, before any check on whether auto-update is enabled. It reads `DOCKER_AGENT_SELF_UPDATE_BACKUP` from the environment and immediately calls `os.Remove()` on that path:

```go
func Cleanup(ctx context.Context) {
    backup := os.Getenv(envBackupMarker)
    if backup == "" {
        return
    }
    if err := os.Remove(backup); err != nil && !errors.Is(err, os.ErrNotExist) {
        ...
    }
}
```

Any local user — or any process that can influence the environment before `docker-agent` is invoked (e.g. a Makefile, a shell wrapper, a test harness) — can set `DOCKER_AGENT_SELF_UPDATE_BACKUP=/path/to/important/file` and cause `docker-agent` to silently delete it on startup, regardless of whether `DOCKER_AGENT_AUTO_UPDATE` is set.

The intent is to clean up the temp backup left by the previous run's updater. A simple mitigation is to validate that the path looks like a temp file produced by the updater (e.g. starts with the expected `.docker-agent-backup-` prefix, lives in the same directory as the executable, or was created within a recent time window):

```go
func Cleanup(ctx context.Context) {
    backup := os.Getenv(envBackupMarker)
    if backup == "" || !isOwnedBackupPath(backup) {
        return
    }
    _ = os.Remove(backup)
}

func isOwnedBackupPath(p string) bool {
    return strings.HasPrefix(filepath.Base(p), ".docker-agent-backup-")
}
```

cmd/root/root.go
```diff
@@ -221,6 +232,23 @@ func visitAll(cmd *cobra.Command, fn func(*cobra.Command)) {
...
+// isManagementInvocation reports whether args correspond to an invocation that
+// must not trigger a self-update + restart: the docker CLI plugin metadata
+// handshake, shell-completion script generation, and the version/help queries.
+// Updating mid-handshake would corrupt the plugin protocol, and restarting a
+// completion call would be surprising.
+func isManagementInvocation(args []string) bool {
+	if len(args) == 0 {
+		return false
+	}
+	switch args[0] {
```
docker-agent, line 247: **[LOW] `isManagementInvocation` skips update for `--help`/`-h` at argv[0] but not global flags mixed with subcommands**

`isManagementInvocation` only inspects `args[0]` (the first subcommand token). If the user runs:

```
docker-agent run --help
```

then `args[0]` is `"run"`, which does **not** match the guard list, so an update + re-exec will be triggered before the help text is shown. The user might not expect an update when asking for help on a specific subcommand. This is probably acceptable by design (the PR description says version/help queries are skipped), but the comment says "version/help queries" yet the guard only catches top-level `--help`/`help`, not per-subcommand help.

This is a minor UX issue, not a correctness bug, but worth documenting in the code comment or extending the guard:

```go
// Check both first-position flags and trailing --help on any subcommand
for _, arg := range args {
    if arg == "--help" || arg == "-h" {
        return true
    }
}
```

docker-agent (commented):

Reviewed the self-update implementation. The overall design is solid — digest/checksum verification, fail-safe fallback, loop guard, backup/restore — all look well-thought-out. One medium security finding and one low UX/documentation finding.

pkg/selfupdate/selfupdate.go
dgageot: Fixed in 31b7de867. `Cleanup()` now validates the `DOCKER_AGENT_SELF_UPDATE_BACKUP` path via `isOwnedBackupPath` (basename must carry the `.docker-agent-backup-` prefix used by `backupExecutable`) before removing it, so a hostile or accidental env value can no longer delete arbitrary files on startup. Added `TestCleanupIgnoresForeignBackupPath` and `TestIsOwnedBackupPath`.

cmd/root/root.go
dgageot: Fixed in 8a43cf91a. `isManagementInvocation` now scans all args for `--help`/`-h` (in addition to the `args[0]` switch), so per-subcommand help such as `run --help` is skipped and no longer triggers an update + re-exec. Updated the doc comment and added `run --help` / `run agent.yaml -h` / `share push --help` cases to the test.

pkg/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,644 @@
...
+
+	for _, asset := range release.Assets {
+		if asset.Name == assetName {
+			if asset.BrowserDownloadURL == "" {
+				return releaseInfo{}, fmt.Errorf("release asset %s has no download URL", assetName)
+			}
+			return releaseInfo{
+				Tag:         release.TagName,
+				Asset:       asset.Name,
+				DownloadURL: asset.BrowserDownloadURL,
```
docker-agent, line 292: **[MEDIUM] Binary download URL from GitHub API is followed without hostname validation**

`latestRelease` stores `asset.BrowserDownloadURL` directly from the GitHub API response, and `downloadAndStage` issues an HTTP request to that URL without validating its origin. In a scenario where the GitHub API response is tampered with (e.g., via a compromised API token or a server-side compromise), the URL could redirect the binary download to an attacker-controlled host.

Importantly, the **checksum verification path is partially protected**: `verifyChecksum` fetches `checksums.txt` from the hardcoded `u.DownloadBaseURL` (not from the API-supplied URL), so the checksum is verified from a trusted source. However, if the asset also supplies a `Digest` field in the API response (line 289), that digest *is* API-supplied — so a compromised API response can supply a matching digest for a malicious binary, bypassing checksum verification entirely.

Consider validating the `BrowserDownloadURL` hostname is `github.com` or `objects.githubusercontent.com` before following it, and/or ignoring the `Digest` field from the API response in favour of always using the `checksums.txt` fallback from the hardcoded base URL.

pkg/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,644 @@
...
+	if err != nil {
+		_ = os.Remove(newPath)
+		return fmt.Errorf("replacing executable: %w", err)
+	}
+
+	fmt.Fprintf(stderr, "docker-agent: updated to %s, restarting...\n", release.Tag)
+
+	// Re-execute the freshly installed binary with the original arguments,
+	// marking the child so it will not attempt to update again.
+	env := append(os.Environ(), envReExecMarker+"=1", envBackupMarker+"="+backup)
```
docker-agent: **[LOW] `append(os.Environ(), ...)` may leave duplicate env vars; child reads stale backup path**

If `DOCKER_AGENT_SELF_UPDATE_BACKUP` is already present in the parent environment (from a prior update cycle that didn't clean up), `append(os.Environ(), ..., envBackupMarker+"="+backup)` produces two entries for the same key. On Linux, `os.Getenv` (and the underlying `getenv(3)`) returns the **first** match — so the re-executed child's `Cleanup()` reads the old, stale backup path rather than the newly created backup, and the new backup temp file is never removed.

While the re-exec loop guard (`DOCKER_AGENT_SELF_UPDATE_REEXEC`) prevents a second update attempt, the stale env var is still passed down. The fix is to deduplicate/filter existing `DOCKER_AGENT_SELF_UPDATE_*` entries from `os.Environ()` before appending the new values.

pkg/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,644 @@
...
+
+	line, err := bufio.NewReader(stdin).ReadString('\n')
+	if err != nil && line == "" {
+		// Could not read an answer (EOF on an empty line): be conservative and
+		// keep running the current version rather than upgrading unprompted.
+		return false
+	}
+
+	return answerIsYes(line)
+}
```
docker-agent, line 617: **[LOW] `isInteractive` type-asserts to `*os.File`; any wrapped stdin silently auto-confirms**

`isInteractive` returns `false` for any `stdin` that is not a `*os.File`, which causes `confirmUpdate` to auto-confirm the update without user interaction. In the current call site `Execute(ctx, stdin, ...)` passes stdin as-is, so if a caller wraps `os.Stdin` (e.g., in a test harness, integration runner, or a future refactor that buffers stdin), the update is silently auto-confirmed.

The comment documents this behaviour for `*os.File` vs non-`*os.File`, but a wrapped `os.Stdin` from a real terminal is a realistic scenario. Consider also accepting an `*os.File` unwrapped via `io.Reader` interface probing, or document the contract explicitly so callers know they must pass `os.Stdin` directly rather than a wrapper.

docker-agent (commented):

This PR adds a well-structured opt-in self-update mechanism with checksum verification, backup/restore, loop guard, and platform-specific exec. A few issues were found in the added code worth addressing.

pkg/selfupdate/selfupdate.go
docker-agent: Perfect, thank you! The fix correctly validates the backup path via `isOwnedBackupPath` before deletion, ensuring only files with the `.docker-agent-backup-` basename prefix can be removed. The test coverage looks solid — `TestCleanupIgnoresForeignBackupPath` demonstrates that arbitrary paths like `important.txt` are now protected, even when pointed to by the environment variable.

This closes the arbitrary file deletion vulnerability while preserving the cleanup functionality for legitimate backup files.

cmd/root/root.go
docker-agent: Perfect! The fix correctly addresses the issue:

✅ isManagementInvocation now scans all args for --help/-h (lines 253-257), not just args[0]
✅ Updated doc comment accurately describes the behavior: Help and version are detected anywhere in args, not just at args[0]
✅ Test cases confirm per-subcommand help is now recognized: run --help, run agent.yaml -h, share push --help

This ensures users won't see unexpected self-update behavior when asking for help on any subcommand. Thanks for the quick fix!

pkg/selfupdate/selfupdate.go
dgageot: Fixed in 001a22a16. The asset `browser_download_url` is now validated against the hardcoded `DownloadBaseURL` host and scheme (`validateDownloadURL`) before it is followed, so a tampered API response can no longer redirect the download to an attacker host. The API-supplied `digest` field has been dropped entirely: `verifyChecksum` now always fetches `checksums.txt` from the trusted `DownloadBaseURL`, so a compromised API response cannot supply a matching digest for a malicious binary. Added `TestLatestReleaseRejectsUntrustedDownloadHost` and `TestValidateDownloadURL`.

pkg/selfupdate/selfupdate.go
dgageot: Fixed in 001a22a16. Before re-exec, `selfUpdateEnv` strips any inherited `DOCKER_AGENT_SELF_UPDATE_*` entries from `os.Environ()` so the fresh markers are the only ones present. This prevents a stale backup path from a prior cycle shadowing the new one (getenv returns the first match) and leaking the temp backup. Added `TestSelfUpdateEnvStripsMarkers`.

pkg/selfupdate/selfupdate.go
dgageot: Fixed in 001a22a16. `isInteractive` now probes for an `interface{ Fd() uintptr }` accessor instead of asserting the concrete `*os.File` type, so a wrapped real terminal (test harness, buffered stdin, future refactor) is still detected and prompted rather than silently auto-confirmed. The doc comment was updated accordingly.

docker-agent: Perfect! The interface probe for Fd() method is exactly the right approach — it handles wrapped terminals while still correctly returning false for test buffers or non-terminal stdin. The updated doc comment makes the contract clear for future callers.

Thanks for the quick fix!

pkg/selfupdate/selfupdate.go
docker-agent: Perfect, thank you for the thorough fix! I've verified the implementation:

✅ **`validateDownloadURL`** now enforces that the asset URL matches the hardcoded `DownloadBaseURL` scheme and hostname before following it, blocking tampered API responses from redirecting downloads.

✅ **API-supplied digest dropped**: The `Digest` field is no longer extracted from the API response. `verifyChecksum` now unconditionally fetches `checksums.txt` from the trusted `DownloadBaseURL`, eliminating the path where a compromised API could supply a matching digest for a malicious binary.

✅ **Test coverage**: The new tests properly exercise the validation (including hostname and scheme checks) and the untrusted-host rejection.

The security issue is fully resolved.

pkg/selfupdate/selfupdate.go
docker-agent: Perfect! The `selfUpdateEnv` filter cleanly solves the duplicate env var issue. Stripping inherited `DOCKER_AGENT_SELF_UPDATE_*` entries before appending fresh markers ensures `getenv` always reads the correct backup path, and the test coverage confirms both the filtering logic and the single-entry guarantee.

Thanks for the thorough fix and test — this closes the temp file leak scenario cleanly.

aheritier (approved):

PR #2994, opened 2026-06-03
feat(runtime): expose per-turn usage and cost in the after_llm_call hook payload

### LINKED ISSUE #2948: Expose per-turn token usage and cost in the after_llm_call hook payload
## Overview

Add per-turn token usage and cost to the `after_llm_call` hook payload. Today the payload carries only:

```
&hooks.Input{
    SessionID:       sess.ID,
    AgentName:       a.Name(),
    StopResponse:    responseContent,
    LastUserMessage: sess.GetLastUserMessageContent(),
}
```

This proposal adds, additively:
- `usage` — the per-call `chat.Usage` (input / cached input / cache write / output tokens), nested to avoid colliding with existing flat token fields
- `cost` — the per-call cost in USD
- ~~`model_id` — the model actually used for this call~~ *(edit: already landed in #2911 — remaining scope is `usage` + `cost` only)*

## Motivation

There is currently no supported way to observe per-LLM-call usage or cost from outside the core. The only place spend is materialized is the session store, and the session store only records a sub-session **when it completes successfully**.

When a sub-session (delegated agent / skill / background agent) **fails**, its turns are dropped from persistence:
- `pkg/runtime/agent_delegation.go` `runForwarding` returns early on the first `ErrorEvent` (~line 289), so `parent.AddSubSession(s)` and `SubSessionCompletedEvent` are never reached.
- `runCollecting` has the same shape (returns on `errMsg != ""` before `AddSubSession`).
- The persistence observer only writes a sub-session in response to `SubSessionCompletedEvent`, so a failed sub-session's turns never reach the DB.

This is reasonable for the **session DB** (a session is resumable conversation state, and a failed sub-session isn't part of it). But the API calls were **already billed**. The cost the session reports can be far below the actual provider invoice.

Concretely: in a real run against Vertex AI with several failing sub-sessions, the actual invoice was **~9× the cost shown by the session**. Each failed sub-session made multiple billed model calls that never appeared in any total. This makes spend impossible to reconcile and is a real safety concern when running against a metered provider.

The root cause is that cost recording is coupled to session lifecycle (`SubSessionCompletedEvent`). The cleanest place to expose spend is the layer where it actually happens — once per LLM call. `executeAfterLLMCallHooks` already fires inside the shared turn loop (`pkg/runtime/loop.go:572`) on every successful model call, for sub-sessions too, and **before** the sub-session failure handling in `agent_delegation.go`. A per-turn payload with usage and cost therefore captures failed sub-session spend automatically, with no coupling to session completion.

## Use Cases

1. **Spend reconciliation / cost ledger** — a sidecar (agent YAML + a shell / Node / Python script) appends every billed turn to its own store, so the total matches the provider invoice even when sub-sessions fail.
2. **Budget guard** — a handler that warns or stops a run once cumulative cost crosses a threshold.
3. **Alerting / telemetry** — forward per-turn usage and cost to an external monitoring system, independent of session persistence.

## Proposed Solution

Widen the `after_llm_call` payload at `executeAfterLLMCallHooks` to include `usage` and `cost`. The data already exists locally at the call site in the turn loop; this is plumbing, not new computation. With this primitive, a cost ledger lives entirely outside the core — no new core subsystem and no new storage owned by the core.

Because this widens a **public, irreversible** payload contract, the schema shape should be settled first. Open questions:

1. **Unpriced vs free.** A flat `cost: 0` can't distinguish a free call from a model with no pricing. Prefer a structural signal (e.g. nullable cost = unpriced) over a documented "check `usage != nil`" convention.
2. **Token field naming.** Nesting under `usage` avoids collision with the existing compaction-related token fields — is this the preferred convention?
3. **Coverage.** Main runtime and harness paths are straightforward; compaction sub-runtimes and the chatserver / a2a / acp paths likely need follow-up. Should the initial change scope to the main loop?

I have a working implementation of the payload widening plus an example sidecar ledger, and am happy to open a PR once the schema is agreed.

## Alternatives

- **Internal Observer + built-in cost ledger** (a new `pkg/costlog` package with its own SQLite store). This works and needs no public API change, but it makes the core own a new subsystem, storage format, and config surface, and is Go-only. The hook approach keeps the core to a primitive and lets the ledger be any language. (I have prototyped both.)
- **Persist failed sub-sessions into the session DB.** This conflicts with the intended meaning of the session store as resumable state, so it's a poorer fit.

## Related Issues

- [#504](https://github.com/docker/docker-agent/issues/504) Token tracking when using subagents (closed)
- [#90](https://github.com/docker/docker-agent/issues/90) Cost tracking for multi-agent runs sometimes resets to zero (closed)
- [#1345](https://github.com/docker/docker-agent/issues/1345) Stats on usage + interaction summary (open)

## Additional Context

Line references are against `main` at the time of writing and may drift. ~~The proposal also reconciles the existing doc that says `after_llm_call` populates the model id, which today it does not.~~ *(edit: that was fixed in #2911; `model_id` is already populated. The remaining scope of this issue is `usage` + `cost`.)*

kimizuka:
## Summary

Adds per-turn token `usage` and computed USD `cost` to the `after_llm_call` hook payload (`hooks.Input`), so a sidecar cost ledger can record per-call spend from the hook payload alone — without subscribing to the runtime event channel. This is the primitive-first half of #2948.

`model_id` was already populated by #2911, so the remaining scope here is just `usage` and `cost`.

This implements what was discussed in #2948. The one design decision worth a second look is the `cost` JSON encoding, covered under Wire contract below.

Closes #2948

## Wire contract

`hooks.Input` is the struct shared by every hook event and serialized to JSON for handlers, so the additions are deliberately conservative:

```go
Usage *chat.Usage `json:"usage,omitempty"`
Cost  *float64    `json:"cost,omitempty"`
```

For a **native model call**, `cost` has three meaningful states:

| Go value | JSON | meaning |
|----------|------|---------|
| `nil`    | key absent | unpriced — no pricing table, or no usage reported |
| `&0`     | `"cost": 0` | a priced call that was genuinely free |
| `&N`     | `"cost": N` | the priced USD cost of the response |

`omitempty` on a pointer drops only `nil`, never a pointer to `0`, so a free call still emits an explicit `"cost": 0`. A present `cost` is therefore authoritative and an absent one means "unpriced", with no need to cross-check `usage`. Both fields are populated only for `after_llm_call`; they are `nil` (and thus absent) on every other event, so no other event's payload changes.

### Note on the schema (deviation from my #2948 comment)

In #2948 I suggested `json:"cost"` (no `omitempty`, explicit `cost: null` for unpriced). I switched to `omitempty` here because `hooks.Input` is shared by all events: without `omitempty`, every non-`after_llm_call` event (`before_llm_call`, `session_end`, …) would start emitting `"cost": null`, which both pollutes unrelated payloads and breaks the struct's all-`omitempty` convention. The pointer + `omitempty` form keeps the same three-way distinction within `after_llm_call` (absent / `0` / `N`) while leaving other events untouched. An explicit `null` instead of `omitempty` is a one-line change if the team prefers it.

## Cost is computed once and equals the session's recorded cost

The per-turn cost is computed once in `runTurn` via a new `computeMessageCost(usage, m) *float64` helper and threaded into **both** the hook payload and `recordAssistantMessage`. The previous inline arithmetic in `recordAssistantMessage` is replaced by this single source (the `m *modelsdev.Model` param becomes the precomputed `cost *float64`), so the cost a handler sees is exactly the cost the session bills for the turn. The persisted message cost is unchanged (`nil` records as `0`, matching prior behavior).

## Harness agents

For harness agents, `cost` is the harness's own reported total rather than a computed price. The harness library defaults the total to `0` when the harness output omits a cost (e.g. the `codex` harness reports token counts but no cost), which is indistinguishable from a genuinely free call — so to avoid telling a ledger that a billed turn was free, `cost` is surfaced only when the harness reported a non-zero value (otherwise it is `nil`/unpriced).

## Sub-sessions

`after_llm_call` fires for every model call, including those inside sub-sessions (transferred tasks, background agents, skills), each with the sub-session's own `session_id`. Summing `cost` across `after_llm_call` events therefore captures all spend — including sub-sessions whose cost may never reach the session store — which is the motivating case in #2948. The change does not touch sub-session persistence in any way.

## Example: a cost-ledger sidecar

A `command` hook can append one CSV row per model call straight from the payload — no event-channel subscription needed. `has("cost")` distinguishes an **unpriced** call (key absent) from a **priced free** one (`cost: 0`):

```yaml
after_llm_call:
  - type: command
    command: |
      cat | jq -r '[
        (now | todateiso8601), .session_id, .model_id,
        (.usage.input_tokens  // 0), (.usage.output_tokens // 0),
        (if has("cost") then (.cost | tostring) else "unpriced" end)
      ] | @csv' >> /tmp/cost-ledger.csv
```

A runnable version is wired into the canonical `examples/hooks.yaml`. Because `after_llm_call` fires for sub-session turns too (each with its own `session_id`), summing the `cost` column is the full spend for the run.

## Out of scope (follow-ups)

- **Fallback-model pricing**: on a turn that fell back to a secondary model, `cost`/`model_id` reflect the primary model. This is pre-existing behavior in `recordAssistantMessage` and is preserved here (hook cost == recorded cost); attributing cost to the model that actually ran is a separate change.
- Coverage of compaction sub-runtimes and the chatserver / a2a / acp paths.

## Testing

`pkg/runtime/after_llm_call_test.go`:

- priced call → `usage` + non-nil `cost`, and `*cost == sess.OwnCost()`
- unpriced model → `usage` present, `cost` nil
- JSON wire contract (absent / explicit `0` / `N`)
- harness path (codex) → `usage` present, `cost` nil
- `computeMessageCost` unit tests (every nil branch + all token classes)

`docs/configuration/hooks/index.md` updated for the new fields and the sub-session / harness caveats; `examples/hooks.yaml` demonstrates a cost-ledger sidecar consuming the payload.

Signed off per DCO.


kimizuka:
Heads-up that I'm planning to flip #2994 from draft to **Ready for review** tomorrow.

On the one open call (`cost` schema): I'll default to `json:"cost,omitempty"` — full rationale is in my comment above and pinned by a test. If you'd rather keep "unpriced" as a visible explicit `null`, just say so and I'll switch it (one-line change).

Scope unchanged: main run loop + harness path, `usage` + `cost` only; compaction sub-runtimes and the chatserver/a2a/acp paths remain a documented follow-up.

aheritier:
This PR has a merge conflict with the base branch. Please rebase onto `main` and resolve the conflicts so it can be reviewed. I've marked it as a draft in the meantime.

kimizuka:
Thanks for the heads-up, @aheritier! I've rebased onto `main`, resolved the conflict, and force-pushed — it now shows as mergeable again.

The only collision was in `pkg/runtime/loop.go`, where main's new `FinishReasonRefusal` handling landed right next to this PR's `recordAssistantMessage` signature change; I kept both.
Build, pkg/runtime tests, and `golangci-lint` all pass, and the commits are still Verified.

Let me know if there's anything else needed.

Sayt-0 (approved):

Sayt-0:
looks good to me !

PR #2996, opened 2026-06-03
refactor(coder): use first_available model selection with multi-provider fallbacks

dgageot:
The builtin coder agent previously hardcoded Anthropic models, which meant users without Anthropic credentials had to manually configure alternative providers. This change leverages the new `first_available` model selection feature to work seamlessly with any configured provider.

The coder agent's `default` and `fast` models now each list candidate provider/model combinations across Anthropic, OpenAI, and Google, with a fallback to the local `dmr` model. The agent automatically selects the first available option based on the user's configured credentials, eliminating the need for manual configuration in most cases.

The `fast` model was renamed from `haiku` for consistency, and the librarian sub-agent reference was updated accordingly.

### DIFF
--- a/pkg/config/builtin-agents/coder.yaml
+++ b/pkg/config/builtin-agents/coder.yaml
@@ -1,10 +1,16 @@
 models:
   default:
-    provider: anthropic
-    model: claude-opus-4-6
-  haiku:
-    provider: anthropic
-    model: claude-haiku-4-5
+    first_available:
+      - anthropic/claude-opus-4-8
+      - openai/gpt-5.5
+      - google/gemini-3.5-flash
+      - dmr/ai/qwen3
+  fast:
+    first_available:
+      - anthropic/claude-haiku-4-5
+      - openai/gpt-5.4-mini
+      - google/gemini-3.1-flash-lite
+      - dmr/ai/qwen3
 
 agents:
   root:
@@ -185,7 +191,7 @@ agents:
         agent: root
 
   librarian:
-    model: haiku
+    model: fast
     description: Web Researcher
     instruction: |
       You search the web for documentation, articles, and technical references.


docker-agent (commented):

pkg/config/builtin-agents/coder.yaml
```diff
@@ -1,10 +1,16 @@
 models:
   default:
-    provider: anthropic
-    model: claude-opus-4-6
-  haiku:
-    provider: anthropic
-    model: claude-haiku-4-5
+    first_available:
+      - anthropic/claude-opus-4-8
```
docker-agent, line 4: **[MEDIUM] Speculative model IDs in `default` first_available list may cause runtime API failures**

The new `default` model candidates include `anthropic/claude-opus-4-8`, `openai/gpt-5.5`, and `google/gemini-3.5-flash`. None of these are known public model identifiers. The `first_available` selector picks the first candidate whose **credentials are configured** — not whether the model ID is valid — so a user with Anthropic credentials will always resolve to `anthropic/claude-opus-4-8`, and the subsequent API call will fail if Anthropic's API does not recognize that model ID.

The old config used `claude-opus-4-6` (also non-public), so this project likely uses pre-release/internal model names. If these are intentional internal identifiers, this is expected. If they are typos or speculative future names, users with cloud credentials will get runtime failures instead of a working fallback — the local `dmr/ai/qwen3` fallback is only reached when no cloud credentials are configured at all.

Consider verifying these model IDs against the provider's current API catalog (e.g., Anthropic's model list, OpenAI's `/v1/models`, Google's Gemini API).

pkg/config/builtin-agents/coder.yaml
```diff
@@ -1,10 +1,16 @@
...
+      - openai/gpt-5.5
+      - google/gemini-3.5-flash
+      - dmr/ai/qwen3
+  fast:
+    first_available:
+      - anthropic/claude-haiku-4-5
+      - openai/gpt-5.4-mini
```
docker-agent, line 11: **[MEDIUM] Unrecognized model IDs in `fast` first_available list may cause runtime API failures**

`openai/gpt-5.4-mini` and `google/gemini-3.1-flash-lite` are not recognized public model identifiers. The current OpenAI fast/mini model is typically `gpt-4o-mini`; `gpt-5.4-mini` is not a known offering. Similarly `gemini-3.1-flash-lite` does not match any known Gemini model name.

As with the `default` list, the `first_available` runtime resolves on credential availability, not model validity. A user with OpenAI credentials will select `openai/gpt-5.4-mini` and receive an API error at inference time if the model ID is wrong.

`anthropic/claude-haiku-4-5` (unchanged from the old config) follows the same versioning pattern, so this may be an intentional pre-release naming scheme — but worth confirming that all three provider IDs are correct before shipping.

trungutt (approved):

PR #2997, opened 2026-06-03
feat: update default agent to use first_available model selection

dgageot:
The default agent now uses the new `first_available` model selection mechanism introduced in the previous commit. Instead of relying on environment detection at startup, the agent tries a prioritized list of models in order: Claude Sonnet 4.6, GPT-5, Gemini 3.5 Flash, Mistral Small, and Qwen3 (local fallback).

This change also bumps the default model versions for cloud providers. OpenAI moves from `gpt-5-mini` to `gpt-5`, and Google moves from `gemini-2.5-flash` to `gemini-3.5-flash`. The test expectations are updated to reflect these newer versions.

No breaking changes; the agent continues to work seamlessly across different environments without requiring users to reconfigure anything.

### DIFF
--- a/pkg/config/auto.go
+++ b/pkg/config/auto.go
@@ -53,9 +53,9 @@ func (e *AutoModelFallbackError) Error() string {
 }
 
 var DefaultModels = map[string]string{
-	"openai":         "gpt-5-mini",
+	"openai":         "gpt-5",
 	"anthropic":      "claude-sonnet-4-6",
-	"google":         "gemini-2.5-flash",
+	"google":         "gemini-3.5-flash",
 	"dmr":            "ai/qwen3:latest",
 	"mistral":        "mistral-small-latest",
 	"amazon-bedrock": "global.anthropic.claude-sonnet-4-5-20250929-v1:0",
--- a/pkg/config/builtin-agents/default.yaml
+++ b/pkg/config/builtin-agents/default.yaml
@@ -1,6 +1,15 @@
+models:
+  smart:
+    first_available:
+      - anthropic/claude-sonnet-4-6
+      - openai/gpt-5
+      - google/gemini-3.5-flash
+      - mistral/mistral-small-latest
+      - dmr/ai/qwen3
+
 agents:
   root:
-    model: auto
+    model: smart
     description: A helpful AI assistant
     welcome_message: |
       Hello! I'm your AI assistant. How can I help you today?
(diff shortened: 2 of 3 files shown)


pkg/config/auto.go
```diff
@@ -53,9 +53,9 @@ func (e *AutoModelFallbackError) Error() string {
 }
 
 var DefaultModels = map[string]string{
-	"openai":         "gpt-5-mini",
+	"openai":         "gpt-5",
 	"anthropic":      "claude-sonnet-4-6",
-	"google":         "gemini-2.5-flash",
+	"google":         "gemini-3.5-flash",
```
docker-agent, line 58: **[LIKELY HIGH] Potentially non-existent Google model name `gemini-3.5-flash`**

The `DefaultModels` map was updated from `"gemini-2.5-flash"` to `"gemini-3.5-flash"`, and the same name appears in `default.yaml` (`google/gemini-3.5-flash`).

Google's Gemini versioning has followed the series `1.0`, `1.5`, `2.0`, `2.5` — a jump to `3.5` is a significant version skip. As of 2026-06-03, `gemini-3.5-flash` does not appear in Google's Gemini API catalog, while `gemini-2.5-flash` is a confirmed, released model.

**Impact**: Any user whose sole configured key is `GOOGLE_API_KEY` (or `GOOGLE_GENAI_API_KEY`) will receive an API error ("unknown model" or similar) when the `first_available` resolver or `DefaultModels` auto-detection selects the Google provider. The two affected code paths are:
1. `default.yaml` line 6: `- google/gemini-3.5-flash` (new `first_available` path)
2. `auto.go` line 58: `"google": "gemini-3.5-flash"` (legacy `DefaultModels` path)

If the intent is to bump to a newly available model, please verify the exact model identifier in the [Google Generative AI model docs](https://ai.google.dev/gemini-api/docs/models) before merging. If it should remain the previous version, revert to `"gemini-2.5-flash"`.

pkg/config/builtin-agents/default.yaml
```diff
@@ -1,6 +1,15 @@
+models:
+  smart:
+    first_available:
+      - anthropic/claude-sonnet-4-6
+      - openai/gpt-5
+      - google/gemini-3.5-flash
```
docker-agent, line 6: **[LIKELY HIGH] Potentially non-existent Google model name `gemini-3.5-flash`**

Same issue as flagged in `auto.go` — this YAML entry uses the same model name `google/gemini-3.5-flash`. If `gemini-3.5-flash` is not a valid Google API model identifier, users with only Google credentials configured will fall through to `dmr/ai/qwen3` (local fallback) unexpectedly, or fail with an API error if `google` is the resolved provider.

docker-agent (commented):

trungutt (approved):

PR #2815, opened 2026-05-18
feat(mcp): add sampling/createMessage support

dgageot:
Fixes #2809.

Implements MCP sampling (`sampling/createMessage`) so MCP servers attached to a docker-agent can ask the host to drive its own LLM, mirroring the existing elicitation plumbing.

## What

- New `tools.SamplingHandler` capability (`pkg/tools/sampling.go`, `pkg/tools/capabilities.go`).
- MCP toolset advertises `Sampling` in `ClientCapabilities` and wires `CreateMessageHandler` for both stdio and remote transports (`pkg/tools/mcp/{mcp.go,session_client.go,stdio.go,remote.go}`).
- `pkg/runtime/sampling.go`: converts inbound `CreateMessageParams` to `[]chat.Message` (system prompt, text/image/audio, role mapping), clones the current agent's model with the requested `MaxTokens`, drains the resulting stream and returns an `mcp.CreateMessageResult` with the assistant text, model id and a translated `stopReason`.
- `pkg/runtime/loop.go` passes the sampling handler through `tools.ConfigureHandlers`.

## Design notes

- **Host stays in control.** Server-supplied `ModelPreferences` are advisory only; we use the agent's currently-selected model. Tool use is intentionally disabled (`tools=nil`) so sampling cannot recurse into the agent loop.
- **Structured output is cleared per call.** Without this, the server would inherit the agent's JSON-schema response format and silently get a reshaped reply.
- **Hardened against misbehaving servers.** Per-request limits are enforced before we touch the model: at most 256 messages, 1 MiB per text block (and per system prompt), 8 MiB per image/audio block. Nil messages and unknown roles are rejected.
- **Visibility.** Inbound requests log at `Info`; completion logs model + finish reason + byte count at `Debug`.

## Tests

`pkg/runtime/sampling_test.go` covers the message conversion (text, system prompt, image data URL, audio fallback, default role, unsupported role), every rejection path (nil request, system-prompt-only, nil entry, oversize text/image/system-prompt, too-many-messages), the finish-reason → stopReason mapping, and the data-URL helper. Existing MCP client mocks (`pkg/tools/mcp/{mcp_test.go,reconnect_test.go}`) were updated for the new interface method.

## Validation

- `task lint` — 0 issues (golangci-lint + custom cops + `go mod tidy --diff`)
- `task test` — all packages pass

Issue #2809, opened 2026-05-18
feat: MCP sampling (sampling/createMessage)


EronWright:
## Overview

Support the [MCP sampling specification](https://modelcontextprotocol.io/specification/2025-11-25/client/sampling) (`sampling/createMessage`), which allows MCP servers to request LLM completions from the client. This enables servers to incorporate inference into their behavior without holding API keys — the client retains full control over model access, selection, and user approval.

<img width="650" height="567" alt="Image" src="https://github.com/user-attachments/assets/00a69266-7769-41c2-bc61-353a86d76d3b" />

credit: modelcontextprotocol.io

## Motivation

MCP servers increasingly need LLM inference as part of their operation (e.g. synthesizing retrieved content, generating structured output from raw data). Without sampling, servers must either embed their own API keys or return raw results and leave all synthesis to the outer agent loop. The spec provides a standard protocol for this that keeps the client in control of model choice and cost, and keeps users in the loop.

## Use Cases

1. **Natural language tool responses** — a tool server uses sampling to convert raw structured data (e.g. a database query result) into a natural language summary before returning it to the agent, without the server holding its own model credentials.

2. **Sampling with tools** — a server issues a `sampling/createMessage` request with a `tools` array; the client's LLM calls one or more tools (e.g. fetching live data), the server executes them and returns results, and the client completes the conversation — all under user supervision.

3. **Agent-to-agent** — a subordinate agent (acting as an MCP server) needs clarification mid-task and sends a sampling request to the orchestrating agent's client: *"Please provide more context about the purpose of this query."*

## Requirements

**Capability Declaration.** A client that supports sampling MUST declare the `sampling` capability at initialization. If the client also supports tool use within sampling, it MUST declare `sampling.tools`; servers MUST NOT send tool-enabled requests to clients that have not declared this.

**Human-in-the-Loop.** There SHOULD always be a human in the loop with the ability to deny sampling requests. The client SHOULD present the prompt for user review before forwarding it to the LLM, allow edits, and similarly present the response for approval before returning it to the server.

**Tool Use.** When a server includes a `tools` array, the LLM may respond with `ToolUseContent` blocks (`stopReason: "toolUse"`). Every such assistant message MUST be followed by a user message consisting entirely of `ToolResultContent` blocks matched by `toolUseId` — mixing tool results with other content types is not permitted. The server MUST enforce these invariants in follow-up requests. Both parties SHOULD impose iteration limits to bound tool loops.

[...]

PR #2998, opened 2026-06-04
Support tool callbacks in MCP sampling

EronWright:
## Summary

Closes the **tool callbacks** functional gap in MCP sampling support — a follow-up to #2815, addressing one of the remaining items from #2809.

When an MCP server includes a `tools` array in a `sampling/createMessage` request, the host now drives its model with those tools and returns any `tool_use` blocks back to the server as `ToolUseContent`. The server remains responsible for executing the tool and continuing the loop in a follow-up sampling request.

```mermaid
sequenceDiagram
     participant H as cagent
      participant S as MCP Server
      participant L as LLM

      activate H
      H->>+S: tools/call {name, arguments}

      note over S: needs LLM inference

      S->>+H: sampling/createMessage<br/>{messages, tools: [...]}
      H->>+L: chat completion
      L-->>-H: ToolUseContent<br/>stopReason: "toolUse"
      H-->>-S: CreateMessageResult<br/>{tool_use, stopReason: "toolUse"}

      note over S: executes tool locally

      S->>+H: sampling/createMessage<br/>{messages + tool_use + tool_result, tools: [...]}
      H->>+L: chat completion
      L-->>-H: TextContent<br/>stopReason: "endTurn"
      H-->>-S: CreateMessageResult<br/>{text, stopReason: "endTurn"}

      S-->>-H: tool result
      deactivate H
```

### What's new

- **New `SamplingWithToolsHandler` type and `SampleableWithTools` interface** — additive, parallel to the existing `SamplingHandler` / `Sampleable`. No breaking changes to the basic sampling path merged in #2815.
- **MCP toolset wires both handler types.** At `Initialize`, exactly one of the SDK's mutually exclusive `ClientOptions.CreateMessage*` fields is populated — prefer with-tools when registered, fall back to basic.
- **Capability handshake advertises `sampling.tools`** so servers know the host can receive tool-enabled requests.
- **Runtime handler** (`pkg/runtime/sampling.go`):
  - Converts V2 multi-block messages: `text`, `image`/`audio`, `tool_use` → assistant `ToolCalls`, `tool_result` → `MessageRoleTool` rows (parallel tool_results expand to multiple chat.Message rows).
  - Converts `[]*mcp.Tool` → `[]tools.Tool` with a no-op handler (the server, not the host, executes).
  - Drives `model.CreateChatCompletionStream`, aggregates streamed tool calls.
  - Builds result `Content` with `TextContent` + `ToolUseContent` blocks; `stopReason: "toolUse"` when tool calls are present.
- **New limits**: `maxSamplingTools=64`, `maxSamplingToolCalls=32`.
- **End-to-end test** (`e2e/sampling_test.go`): mounts an in-process `gomcp.NewServer` on an `httptest` server via `StreamableHTTPHandler`. The server exposes one tool (`ask_with_calculator`) whose handler drives a real sampling-with-tools loop against the connecting cagent. The Gemini side is recorded once and replayed on subsequent runs, so the test runs offline in CI.

### Out of scope (separate gaps from #2809)

- Human-in-the-loop approval UI
- Model-preference hints

## Test plan

- [x] Unit tests pass: \`go test ./pkg/runtime/... ./pkg/tools/...\`
- [x] \`go build ./...\` clean
- [x] \`go vet ./...\` clean
- [x] \`task lint\` (0 offenses)
- [x] \`gofmt\` clean on all changed files
- [x] End-to-end via \`TestExec_Gemini_SamplingWithTools\` (cassette replays offline; verified once against live Gemini):
  - [x] Handshake includes \`sampling.tools\` capability (implicit — \`ServerSession.CreateMessageWithTools\` would refuse otherwise)
  - [x] LLM receives the server-supplied tools
  - [x] Response contains \`ToolUseContent\` with \`stopReason: "toolUse"\` when the model emits a tool_use
  - [x] Follow-up sampling request with \`tool_result\` blocks is converted correctly and the loop terminates with \`endTurn\`


pkg/runtime/sampling.go
```diff
@@ -265,3 +283,361 @@ func stopReason(fr chat.FinishReason) string {
...
+					text.WriteString("\n")
+				}
+				text.WriteString(t)
+			}
+			parts = append(parts, p...)
+		}
+	}
+
+	var out []chat.Message
+	if text.Len() > 0 || len(parts) > 0 || (len(toolCalls) > 0 && role == chat.MessageRoleAssistant) {
```
docker-agent: **[MEDIUM] Silent message drop when `tool_use` blocks appear in a non-assistant role**

`samplingV2BlocksToMessages` appends `tool_use` blocks to `toolCalls`, but only emits them into an output message when `role == chat.MessageRoleAssistant`:

```go
if text.Len() > 0 || len(parts) > 0 || (len(toolCalls) > 0 && role == chat.MessageRoleAssistant) {
```

If a V2 message arrives with a non-assistant role (e.g., `user`) containing **only** `tool_use` blocks — `text` is empty, `parts` is empty, `toolResults` is empty — none of the output conditions are satisfied, so `out` stays empty and the function returns `([], nil)`. The entire message is silently dropped with no error, making the resulting chat history shorter than expected without any indication of why.

While the MCP spec places `tool_use` on assistant turns, malformed MCP servers (or future spec revisions) could trigger this path. Returning an error instead of silently discarding the content would surface the problem immediately.

**Suggested fix**: add a guard after the loop — if `toolCalls` is non-empty and `role != MessageRoleAssistant`, return an explicit error:
```go
if len(toolCalls) > 0 && role != chat.MessageRoleAssistant {
    return nil, fmt.Errorf("tool_use blocks in non-assistant message (role=%s)", role)
}
```

pkg/tools/mcp/remote.go
```diff
@@ -90,12 +90,19 @@ func (c *remoteMCPClient) Initialize(ctx context.Context, _ *gomcp.InitializeReq
...
+	// Sampling: prefer the with-tools handler when registered. The SDK's two
+	// CreateMessage* handlers are mutually exclusive, so populate exactly one.
 	opts := &gomcp.ClientOptions{
 		ElicitationHandler:       c.handleElicitationRequest,
-		CreateMessageHandler:     c.handleSamplingRequest,
 		ToolListChangedHandler:   toolChanged,
 		PromptListChangedHandler: promptChanged,
 	}
+	switch {
+	case c.samplingWithToolsHandler != nil:
```
docker-agent: **[MEDIUM] Handler selection baked at `Initialize` time — reconnect may silently lose sampling-with-tools capability**

The SDK `CreateMessageWithToolsHandler` (vs `CreateMessageHandler`) is chosen once when `Initialize` runs by reading `c.samplingWithToolsHandler` at that moment:

```go
switch {
case c.samplingWithToolsHandler != nil:
    opts.CreateMessageWithToolsHandler = c.handleSamplingWithToolsRequest
case c.samplingHandler != nil:
    opts.CreateMessageHandler = c.handleSamplingRequest
}
```

In the **normal startup path** this is fine: `configureToolsetHandlers` (line 221 of `loop.go`) calls `SetSamplingWithToolsHandler` before `getTools` (line 223) triggers `Initialize`, so the handler is present when the switch executes.

The concern is the **reconnect path**: if `Initialize` is invoked again (e.g., after a dropped connection) without `configureToolsetHandlers` being re-called first, the stored handler fields still hold the right value, so this should be fine in practice. However, if a future refactor re-orders the reconnect sequence or resets the handler fields, the MCP session would silently fall back to no handler — the server's `sampling/createMessage` requests with tools would receive no response and time out.

Consider using an accessor closure (captured at startup time) instead of reading the field at `Initialize`, or adding a comment to `Initialize` noting the ordering dependency so future maintainers know not to call it before the handlers are set.

docker-agent (commented):

Two medium-severity findings in the newly-added sampling-with-tools code. The core stream-aggregation logic, capability handshake, content building, and limits enforcement are all well-structured and correctly tested.

pkg/runtime/sampling.go
EronWright: Fixed in 938896850 — added the suggested guard and a regression test case (`tool_use in non-assistant message is rejected`) under `TestSamplingMessagesV2ToChat`.

pkg/tools/mcp/remote.go
EronWright: Addressed in 938896850 with the lightweight option — extended the existing `// Sampling:` comment in both `stdio.go` and `remote.go` to call out the ordering dependency (`SetSampling*Handler` must run before `Initialize`, including on reconnect). Skipped the closure refactor since the handler fields are sticky on `sessionClient` and reconnect reuses the same client instance — happy to revisit if you'd prefer the stronger guarantee.

dgageot (dismissed):

dgageot:
Sorry @EronWright, our project requires all the commit to be signed and verified. Could you sign them? Thanks!

aheritier:
I moved it back to draft @EronWright the time you rewrite the history with signed commits. Sorry for the issue

EronWright:
@aheritier sorry for delay, all commits are now signed/verified.

pkg/runtime/sampling.go
```diff
@@ -265,3 +283,369 @@ func stopReason(fr chat.FinishReason) string {
...
+			}
+			if delta.Function.Arguments != "" {
+				tc.Function.Arguments += delta.Function.Arguments
+			}
+		}
+
+		if choice.FinishReason != "" {
+			providerFinishReason = choice.FinishReason
+		}
+		if choice.FinishReason == chat.FinishReasonStop || choice.FinishReason == chat.FinishReasonLength {
```
docker-agent, line 611: **[MEDIUM] Early `break` on `FinishReasonStop` may drop tool-call chunks from later SSE events**

`drainSamplingStreamWithTools` breaks out of the stream-read loop as soon as it sees `FinishReasonStop` or `FinishReasonLength`:

```go
if choice.FinishReason == chat.FinishReasonStop || choice.FinishReason == chat.FinishReasonLength {
    break
}
```

Tool-call deltas that appear in the **same** chunk as the finish-reason are captured correctly (they're processed in the `for _, delta := range choice.Delta.ToolCalls` loop a few lines earlier). However, if a provider splits its response so that tool-call argument chunks arrive in **subsequent** SSE frames after the finish-reason frame, those deltas are silently discarded.

The analogous `handleStream` in `streaming.go` guards against this with a `continue` that skips the early-exit when the current chunk contains only tool-call deltas (no finish reason yet). That pattern is missing here.

**Impact:** Incomplete or missing tool calls are returned to the MCP server as a `ToolUseContent` block with truncated arguments, causing the server's tool-execution loop to fail or produce wrong results. The failure is silent — no error is returned.

**Suggested fix:** Mirror the `streaming.go` guard: after accumulating tool-call deltas, `continue` to the next chunk if `choice.FinishReason == ""`, and only break on `FinishReasonStop`/`FinishReasonLength` once all pending deltas have been flushed.

pkg/tools/mcp/stdio.go
```diff
@@ -38,13 +38,23 @@ func (c *stdioMCPClient) Initialize(ctx context.Context, _ *gomcp.InitializeRequ
...
+	// SetSamplingHandler / SetSamplingWithToolsHandler before Initialize (and
+	// before any reconnect) or the chosen branch falls back to no handler.
 	opts := &gomcp.ClientOptions{
 		ElicitationHandler:       c.handleElicitationRequest,
-		CreateMessageHandler:     c.handleSamplingRequest,
 		ToolListChangedHandler:   toolChanged,
 		PromptListChangedHandler: promptChanged,
 	}
+	switch {
+	case c.samplingWithToolsHandler != nil:
```
docker-agent: **[MEDIUM] Reconnect race: sampling handler may not be registered with the SDK if `configureToolsetHandlers` hasn't run yet**

`Initialize` (called on both the initial connect and every supervisor-driven reconnect) reads `c.samplingWithToolsHandler` and `c.samplingHandler` at call time to decide which SDK-level callback to register:

```go
switch {
case c.samplingWithToolsHandler != nil:
    opts.CreateMessageWithToolsHandler = c.handleSamplingWithToolsRequest
case c.samplingHandler != nil:
    opts.CreateMessageHandler = c.handleSamplingRequest
}
// if both are nil → no sampling callback registered with the SDK client
```

The SDK client is constructed with `opts` and never updated afterwards. If a reconnect happens before the runtime's `configureToolsetHandlers` has called `SetSamplingWithToolsHandler` / `SetSamplingHandler`, both fields are `nil` at `Initialize` time. The resulting SDK session has **no** `CreateMessage*` handler at all, so any `sampling/createMessage` request from the server is rejected with an error for the lifetime of that session.

This is a regression from the previous unconditional registration pattern, where the basic sampling callback was always wired. The comment "callers must invoke … before any reconnect" documents the contract, but it is not enforced by any ordering guarantee in the reconnect path today.

**Impact:** An MCP server that triggers sampling during or shortly after a reconnect (e.g., a long-running agentic loop that reconnects due to a network hiccup) will receive an error from the sampling handler instead of a model response, breaking its tool-execution loop.

**Suggested mitigation:** Either (a) guarantee in the supervisor/reconnect path that `configureToolsetHandlers` is always called before `Initialize` completes, and add a test that verifies this ordering; or (b) fall back to registering a best-effort handler at `Initialize` time that reads the handler fields lazily at request time (similar to how `handleSamplingWithToolsRequest` already reads `c.samplingWithToolsHandler` under the mutex), so a late `SetSamplingWithToolsHandler` call is still effective even after reconnect.

docker-agent (commented):

Two medium-severity issues found in the new sampling-with-tools path. Both relate to edge cases in the streaming and reconnect logic.

pkg/runtime/sampling.go
EronWright: Fixed in `e4ff6d496` by mirroring the `streaming.go` continue-on-tool-call-only-chunk guard — any chunk with tool-call deltas and no finish reason now short-circuits to the next frame instead of falling into the FinishReason capture/break path. Functionally a no-op for the current code, but pre-empts the regression you described if anyone inserts processing between the tool-call loop and the break. New subtest `TestDrainSamplingStreamWithTools/tool-call-only chunks continue without falling into finish logic` exercises a 4-chunk argument-fragment stream.

pkg/tools/mcp/stdio.go
EronWright: Fixed in `e4ff6d496` by extracting the registration logic into `sessionClient.applySamplingHandlerOpts` (used by both `stdio.Initialize` and `remote.Initialize`). The helper now wires `CreateMessageWithToolsHandler` eagerly even when both handler fields are still nil at `Initialize` time; `CreateMessageHandler` is only registered when the caller has explicitly opted into basic-only sampling. The callback itself already reads `samplingWithToolsHandler` under `c.mu` at request time (`session_client.go:handleSamplingWithToolsRequest`), so a late `SetSamplingWithToolsHandler` — including on a supervisor-driven reconnect that races `configureToolsetHandlers` — takes effect without re-init.

Coverage: `TestApplySamplingHandlerOpts_RegistrationMatrix` pins the choice across all four handler-field combinations (and asserts we never register both, which would panic the SDK), and `TestHandleSamplingWithToolsRequest_LateSetterTakesEffect` walks the helper → nil-error → late `SetSamplingWithToolsHandler` → invoked sequence end-to-end.

EronWright:
The build issue seems for unrelated reasons, re-run?

aheritier:
I think you rebased it on main which was red at that time. Can you rebase again @EronWright ?

EronWright:
Rebased on `fbe505064` (current main; green). The previous `build-and-test` failure was on a red `main` revision as you suspected. One merge-fix folded into the base commit: `main` added a second `tools.ConfigureHandlers` call site in `skillSubSessionTools` (loop.go) that now also passes `r.samplingWithToolsHandler`. Local `go build ./...`, `go test ./pkg/runtime/... ./pkg/tools/mcp/...`, the sampling cassette e2e, and `golangci-lint` are all green. Ready for another CI approval.

Sayt-0:
`build-and-test` fails: the `e2e` package panics with `SetDefaultTokenStoreFactory called after the default token store was already created`, because the new sampling test builds the global token store and the next `Execute` re-runs `keyringstore.Register()`. Making `Register()` idempotent (e.g. a `sync.Once`) should fix it.

dgageot (approved):

PR #3000, opened 2026-06-04
feat: add user_steering_messages_submit and user_followup_submit hooks

simonferquel-clanker:
Adds two hooks that close the coverage gaps around queued user messages, both modelled on `user_prompt_submit`.

**`user_steering_messages_submit`**
- Fires each time the runtime drains the steering queue: mid-turn, after the model stops, and while idle before the first model call.
- Drained messages passed via `steering_messages`.
- Wired into `drainAndEmitSteered` (the single drain chokepoint) so it covers all three cases — unlike `turn_end`+`reason=steered`, which misses the idle drain.

**`user_followup_submit`**
- Fires each time a queued follow-up (FollowUp API/queue) is dequeued at end-of-turn and starts a fresh turn.
- Follow-up text passed via `prompt`.
- Previously uncovered: `user_prompt_submit` only fires for the first interactive prompt.

Both can block the run (`decision: block` / `continue: false` / exit 2) and contribute **transient** `additional_context` (spliced into the steered/follow-up turn only, never persisted — same as `user_prompt_submit`). The user messages themselves are persisted as normal.

**Changes**
- Config fields + validation in `pkg/config/latest`; events + input fields in `pkg/hooks`; dispatch in `pkg/runtime`.
- `agent-schema.json`, `examples/hooks.yaml`, and `docs/configuration/hooks/index.md` updated (event tables, payload fields, context lists, dedicated sections).

**Testing**
- Contract tests (block + input visibility) and runtime tests (each fires once with the right payload) for both hooks.
- `go build`, package tests (hooks/runtime/config/teamloader), `golangci-lint` (0 issues), and the custom `./lint` all pass.


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/26938773748).

dgageot (approved):

PR #3001, opened 2026-06-04
fix: drop orphaned tool results on session resume

### LINKED ISSUE #1676: Session resume via /sessions immediately fails with Bedrock ValidationException (toolResult/toolUse mismatch)
## Summary

Session resume via `cagent run /sessions` fails immediately with a Bedrock `ValidationException` indicating a `toolResult`/`toolUse` block count mismatch. This occurs regardless of conversation length or complexity — even a minimal "hello" conversation fails on resume. The error occurs when the restored conversation history is sent to the Bedrock API, suggesting the issue is related to how session state is serialized and deserialized.

## Environment

- **LLM Provider**: AWS Bedrock via litellm proxy (`localhost:4000`)
- **Model**: Claude Haiku (`haiku` model group)
- **Proxy**: litellm (OpenAI-compatible endpoint)
- **API**: Bedrock `ConverseStream` operation
- **Fallback Configuration**: None (`Available Model Group Fallbacks=None`)

## Steps to Reproduce

1. Configure cagent to use a litellm proxy pointing to AWS Bedrock (Claude Haiku)
2. Start a new cagent session
3. Have a **minimal interaction** — even just saying "hello" is sufficient
4. Exit the session
5. Run `cagent run /sessions` and select the previous session to resume
6. **Immediately fails** with `ValidationException` — no further interaction is possible

> **Note**: This reproduces on any session, regardless of length or complexity.

## Expected Behavior

cagent should be able to resume any previously saved session via `cagent run /sessions`, correctly restoring the conversation history in a format compliant with the Bedrock ConverseStream API contract (i.e., every `toolResult` block must correspond to a `toolUse` block in the preceding assistant message).

## Actual Behavior

Session resume **immediately fails** on every attempt, regardless of how short or simple the original conversation was. The following error is returned (sanitized):

```
all models failed: error receiving from stream: POST "http://localhost:4000/chat/completions": 503 Service Unavailable

{"message":"litellm.ServiceUnavailableError: litellm.MidStreamFallbackError: litellm.APIConnectionError:
APIConnectionError: OpenAIException - Stream generation failed: ProviderException: (400, \"AWS client error
encountered: <class 'botocore.errorfactory.ValidationException'>: An error occurred (ValidationException) when
calling the ConverseStream operation (reached max retries: 0): The number of toolResult blocks at
messages.128.content exceeds the number of toolUse blocks of previous turn.\", {'Error': {'Message': 'The number
of toolResult blocks at messages.128.content exceeds the number of toolUse blocks of previous turn.', 'Code':
'ValidationException'}, 'ResponseMetadata': {'RequestId': '<redacted>', 'HTTPStatusCode': 400, 'HTTPHeaders':
{'date': 'Tue, 10 Feb 2026 08:58:41 GMT', 'content-type': 'application/json', 'content-length': '124',
'connection': 'keep-alive', 'x-amzn-requestid': '<redacted>', 'x-amzn-errortype':
'ValidationException:http://internal.amazon.com/coral/com.amazon.bedrock/'}, 'MaxAttemptsReached': True,
'RetryAttempts': 0}, 'message': 'The number of toolResult blocks at messages.128.content exceeds the number of
toolUse blocks of previous turn.'}). Received Model Group=haiku\nAvailable Model Group Fallbacks=None",
"type":null,"param":null,"code":"503"}
```

### Observations

The following observations may help narrow down the cause:

1. **The original session worked fine** — the conversation history was valid during the live session
2. **The failure happens immediately on resume** — the very first API call with the restored history is rejected
3. **It reproduces on any session** — even trivially short ones, ruling out conversation length as a factor
4. **The error is about message structure** — the restored history has `toolResult` blocks without matching `toolUse` blocks in the preceding assistant message

### How the Error Occurs

When a session is resumed via `cagent run /sessions`, the conversation history is reconstructed and sent to the LLM provider. The reconstructed history contains `toolResult` blocks in user messages that do not have corresponding `toolUse` blocks in the preceding assistant messages — violating Bedrock's API contract.

### Possible Contributing Factors

1. **Session serialization may drop or reorder tool-related message blocks**: When saving the session state, the exact structure of assistant messages containing `toolUse` blocks may not be preserved, while the corresponding user messages with `toolResult` blocks are preserved — breaking the pairing.

2. **Deserialization may reconstruct messages incorrectly**: When loading the session state back, the conversation may be reconstructed in a way that orphans `toolResult` blocks (e.g., merging multiple assistant messages, dropping `tool_calls` from assistant messages, or reordering content blocks).

3. **Internal tool calls may not round-trip correctly**: cagent's internal tool operations (reading files, searching, directory listing, etc.) generate `toolUse`/`toolResult` pairs in the conversation history. These may not survive the save → load cycle correctly.

4. **Format translation on restore**: If session state is stored in one format (e.g., OpenAI's `tool_calls`/`tool` message roles) and reconstructed differently when sending to Bedrock (which uses `toolUse`/`toolResult` content blocks), the translation may introduce mismatches.

### Error Chain

1. User runs `cagent run /sessions` and selects a previous session to resume
2. cagent loads saved session state and reconstructs the conversation history
3. cagent sends the reconstructed conversation to litellm proxy → AWS Bedrock
4. AWS Bedrock rejects the request with HTTP 400 `ValidationException` (toolResult/toolUse block count mismatch)
5. litellm wraps the error as `MidStreamFallbackError` → `ServiceUnavailableError` (503)
6. cagent reports `all models failed` — session resume fails completely

## Impact

- **Severity**: High — session resumption via `/sessions` does not work
- **Scope**: Affects all sessions when using Bedrock as the LLM provider. Even minimal conversations cannot be resumed.
- **Workaround**: None for session resumption. Users must start new sessions, losing prior context.

## Suggested Investigation Areas

- Inspect how cagent serializes conversation history to disk (session save)
- Inspect how cagent deserializes and reconstructs conversation history on resume (session load)
- Verify that `toolUse`/`toolResult` block pairing is preserved through the save → load cycle
- Check if the session format assumes a specific LLM provider's message schema and whether Bedrock's requirements are handled correctly
- Add validation before sending restored history to ensure `toolResult`/`toolUse` pairing invariant holds

## Related Issues

- #1593 — AWS Bedrock `ValidationException`: Protocol violation when assistant message contains both `tool_calls` and content (related Bedrock validation issue with different root cause)

Sayt-0:
## Problem

Resuming a session via `--session` / `/sessions` immediately fails on AWS Bedrock with `ValidationException: The number of toolResult blocks ... exceeds the number of toolUse blocks of previous turn`.

## Root cause

When a session is compacted, the kept-tail boundary (`FirstKeptEntry`) can land between an assistant `tool_use` and its `tool_result`. On resume, `buildSessionSummaryMessages` emits the conversation from that boundary, so the reconstructed history starts with an orphaned `tool_result` (no matching `tool_use`). `sanitizeToolCalls` — the last guard before messages reach the provider — only injected synthetic results for dangling tool calls and always forwarded `tool` messages, so orphans passed straight through. OpenAI-compatible providers (litellm) forward the invalid request; Bedrock rejects it.

## Fix

`sanitizeToolCalls` now also drops tool results whose `ToolCallID` has no matching `tool_use` in the preceding assistant message, keeping the `toolUse`/`toolResult` invariant balanced in both directions.

## Tests

- `TestSanitizeToolCalls_DropsOrphanedToolResult` — unit guard.
- `TestGetMessages_ResumeAfterCompaction_NoOrphanedToolResult` — full `GetMessages` resume path with a summary boundary splitting a tool pair.
- `TestSanitizeToolCalls_KeepsMissingResultBalanced` — pins existing dangling-tool-call behaviour.

Both new tests fail without the fix and pass with it.

Fixes #1676

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -1250,17 +1250,29 @@ func normalizeMessageContent(messages []chat.Message) []chat.Message {
 	return out
 }
 
-// sanitizeToolCalls ensures every tool call in assistant messages has a
-// corresponding tool-result message. It walks the message list tracking
-// pending tool calls; when a tool-result message arrives its ID is marked
-// fulfilled. When the next assistant or user message is encountered (or the
-// end of the list is reached), any still-pending tool calls receive synthetic
-// error results injected just before that boundary. This guarantees the
-// provider always sees a valid request/response pair for every tool call.
+// sanitizeToolCalls ensures the tool_use/tool_result blocks sent to the
+// provider are always balanced, in both directions:
+//
+//   - Every tool call in an assistant message gets a corresponding tool-result
+//     message. It walks the message list tracking pending tool calls; when a
+//     tool-result message arrives its ID is marked fulfilled. When the next
+//     assistant or user message is encountered (or the end of the list is
+//     reached), any still-pending tool calls receive synthetic error results
+//     injected just before that boundary.
+//   - Every tool-result message has a matching tool_use in the preceding
+//     assistant message. Orphaned tool results — a tool-result whose
+//     ToolCallID was never issued by the preceding assistant message — are
+//     dropped. This happens when compaction's kept-tail boundary lands between
+//     an assistant tool_use and its result, leaving the result at the head of
+//     the resumed history with no matching tool_use. Providers such as AWS
+//     Bedrock reject the request outright in that case ("The number of
+//     toolResult blocks ... exceeds the number of toolUse blocks of previous
+//     turn"), so we must strip these before the request goes out.
 func sanitizeToolCalls(messages []chat.Message) []chat.Message {
 	var (
 		out              []chat.Message
 		pendingToolCalls []tools.ToolCall
+		pendingIDs       = make(map[string]bool)
 		resultIDs        = make(map[string]bool)
 	)
 
@@ -1276,20 +1288,30 @@ func sanitizeToolCalls(messages []chat.Message) []chat.Message {
 			}
 		}
 		pendingToolCalls = nil
+		pendingIDs = make(map[string]bool)
 		resultIDs = make(map[string]bool)
 	}
 
 	for _, msg := range messages {
 		switch {
 		case msg.Role == chat.MessageRoleTool:
-			if msg.ToolCallID != "" {
-				resultIDs[msg.ToolCallID] = true
+			// Drop orphaned tool results: a tool_result with no matching
+			// tool_use in the preceding assistant message violates the
+			// provider contract.
+			if msg.ToolCallID == "" || !pendingIDs[msg.ToolCallID] {
+				continue
 			}
+			resultIDs[msg.ToolCallID] = true
 
 		case msg.Role == chat.MessageRoleAssistant && len(msg.ToolCalls) > 0:
 			flushPending()
 			out = append(out, msg)
 			pendingToolCalls = msg.ToolCalls
+			for _, tc := range msg.ToolCalls {
+				if tc.ID != "" {
+					pendingIDs[tc.ID] = true
+				}
+			}
 			continue
 
 		case msg.Role == chat.MessageRoleUser || msg.Role == chat.MessageRoleAssistant:
--- a/pkg/session/sanitize_orphan_test.go
+++ b/pkg/session/sanitize_orphan_test.go
@@ -1,0 +1,116 @@
+package session
+
+import (
+	"testing"
+
+	"github.com/docker/docker-agent/pkg/agent"
+	"github.com/docker/docker-agent/pkg/chat"
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+// assertToolPairingInvariant fails if any tool-result message lacks a matching
+// tool_call ID in a preceding assistant message. This is exactly the invariant
+// AWS Bedrock's ConverseStream enforces: "The number of toolResult blocks ...
+// exceeds the number of toolUse blocks of previous turn." (issue #1676)
+func assertToolPairingInvariant(t *testing.T, messages []chat.Message) {
+	t.Helper()
+	seen := map[string]bool{}
+	for i, m := range messages {
+		if m.Role == chat.MessageRoleAssistant {
+			for _, tc := range m.ToolCalls {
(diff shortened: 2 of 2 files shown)


pkg/session/session.go
```diff
@@ -1276,20 +1288,30 @@ func sanitizeToolCalls(messages []chat.Message) []chat.Message {
...
 
 	for _, msg := range messages {
 		switch {
 		case msg.Role == chat.MessageRoleTool:
-			if msg.ToolCallID != "" {
-				resultIDs[msg.ToolCallID] = true
+			// Drop orphaned tool results: a tool_result with no matching
+			// tool_use in the preceding assistant message violates the
+			// provider contract.
+			if msg.ToolCallID == "" || !pendingIDs[msg.ToolCallID] {
```
docker-agent, line 1301: **[LOW] Behavioral change: tool results with empty `ToolCallID` are now silently dropped**

Under the old logic the condition was `if msg.ToolCallID != ""` which only gated the `resultIDs` book-keeping — a tool-result with an empty ID still fell through to `out = append(out, msg)` and was forwarded to the provider. Under the new condition:

```go
if msg.ToolCallID == "" || !pendingIDs[msg.ToolCallID] {
    continue   // drops the message
}
```

an empty-ID tool result is now silently dropped. In practice this is likely fine — a tool result with no ID cannot be matched to any `tool_use` and most providers mandate non-empty IDs — but it is a silent behavioral regression from the previous behavior. If any in-flight session persisted with empty-ID tool results they would be discarded on the next resume.

Consider documenting the intentional drop (e.g., `// empty ToolCallID is malformed; drop it`) so the behavior is explicit rather than a side-effect of the orphan check.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the Bedrock `ValidationException` by dropping orphaned tool results in `sanitizeToolCalls`. The logic is sound: `pendingIDs` is populated when an assistant message with tool calls is processed and reset on `flushPending()`, so each turn's scope is correctly isolated. The three new tests provide good coverage of the primary fix path and the existing dangling-call direction.

One minor behavioral change is noted inline.

dgageot (approved):

PR #3006, opened 2026-06-05
fix: skip binary files in content search

dgageot:
The `search_files_content` tool now sniffs file headers to detect and skip binary files before attempting to read them. This prevents slow reads, unnecessary memory usage, and potential encoding errors when encountering binary data during search operations.

The implementation detects ELF binaries, PNG images, PDF files, and other common binary formats by checking magic bytes at the start of files. Tests cover all major detection patterns and verify end-to-end behavior with mixed text and binary files in a directory.

### DIFF
--- a/pkg/tools/builtin/filesystem/filesystem.go
+++ b/pkg/tools/builtin/filesystem/filesystem.go
@@ -1,6 +1,7 @@
 package filesystem
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
@@ -9,6 +10,7 @@ import (
 	"io"
 	"io/fs"
 	"log/slog"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly implements binary file detection for `search_files_content` by sniffing the first 512 bytes of each file. The implementation is clean, consistent with existing patterns, and well-tested.

**Approach**: Checks for null bytes first (fast path for ELF, ZIP, etc.), then falls back to `http.DetectContentType` for magic-byte-based detection (PDF, PNG, and other formats). Silent skip on detection errors is consistent with how the rest of the walk handles I/O failures.

**Test coverage**: Unit tests cover empty input, plain text, UTF-8 text, ELF, PNG, PDF, and an end-to-end integration test with mixed text/binary files in a directory — adequate for the scope of change.

aheritier (approved):

PR #3007, opened 2026-06-05
fix: preserve agent field during command expansion

dgageot:
When invoking a slash command that targets a sub-agent via the `agent:` field, the routing was incorrectly going to the root agent instead of the named sub-agent. The root cause was in `pkg/js/expand.go`'s `ExpandCommands` function, which rebuilt the `types.Command` struct field-by-field at config-load time and silently dropped the `Agent` field. When `LookupCommand` later tried to route the command, it saw an empty `Agent` and never switched agents.

This change fixes the bug by preserving the `Agent` field during command expansion. A follow-up refactor makes `ExpandCommands` copy the struct and overwrite only the expanded text fields (`Description`, `Instruction`), ensuring that future fields are preserved automatically and preventing this class of bug from recurring.

The fix includes unit tests in `pkg/js/expand_agent_test.go` and an end-to-end test in `e2e/exec_command_agent_test.go` with a recorded cassette and config fixture, verifying that the instruction reaches the specialist sub-agent directly. All validation checks pass: build, go vet, race tests, and the full test suites for js, cli, runtime, config, and e2e.

### DIFF
--- a/e2e/exec_command_agent_test.go
+++ b/e2e/exec_command_agent_test.go
@@ -1,0 +1,18 @@
+package e2e_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestExec_CommandTargetsAgent verifies that invoking a /command which targets
+// a sub-agent sends the instructions directly to that agent, bypassing the root
+// agent (no transfer_task round-trip). The recorded cassette only contains a
+// single request carrying the specialist agent's system prompt, which proves
+// the message reached the specialist directly.
+func TestExec_CommandTargetsAgent(t *testing.T) {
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix in `pkg/js/expand.go` is correct and well-tested. Copying the `types.Command` struct and overwriting only the `Description` and `Instruction` text fields reliably preserves the `Agent` field and prevents this class of bug from recurring with future fields.

The new unit test (`TestExpandCommandsPreservesAgent`) directly covers the regression, and the e2e test (`TestExec_CommandTargetsAgent`) validates the full routing path end-to-end. No bugs found in the added code.

melmennaoui (approved):

PR #3008, opened 2026-06-05
fix: filter completion candidates to aliases and agent YAMLs only

hamza-jeddad:
Previously, `docker agent run <TAB>` included all directories from cwd in completions.

**What this PR fixes:**
- Directories are excluded from plain-prefix (no path typed) completion
- Dotfile YAMLs (e.g. `.golangci.yml`) are excluded

**Known limitation / follow-up:** Non-dotfile, non-agent YAMLs like `Taskfile.yml` and `docker-compose.yaml` are regular `.yml` files and still appear — filtering those is out of scope for this PR.

**Implementation:** `completeAlias` now calls a new `completeAgentYAMLInCwd` helper instead of `completeAgentFilename` for plain-prefix input. `completeAgentFilename` is unchanged — `./` and `/` path drill-down still works.

Also sorts alias names alphabetically.

### DIFF
--- a/cmd/root/completion.go
+++ b/cmd/root/completion.go
@@ -1,8 +1,10 @@
 package root
 
 import (
+	"maps"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


cmd/root/completion.go
```diff
@@ -113,6 +111,35 @@ func completeTheme(_ *cobra.Command, _ []string, toComplete string) ([]string, c
...
+// this is the "no path typed yet" case where suggesting the full filesystem
+// tree would be noise.
+func completeAgentYAMLInCwd(prefix string) []string {
+	entries, err := os.ReadDir(".")
+	if err != nil {
+		return nil
+	}
+	var out []string
+	for _, e := range entries {
+		if e.IsDir() {
```
docker-agent: **[LOW] Symlinks-to-directories with `.yaml`/`.yml` names pass the `IsDir()` guard**

`os.ReadDir` on Linux returns `DirEntry` entries where `IsDir()` returns `false` for symlinks — even when the symlink target is a directory. This means a symlink like `ln -s /some/dir my-agent.yaml` in the working directory would pass all guards in `completeAgentYAMLInCwd` and appear as a valid YAML completion candidate, even though it is not a file.

A more robust check would be:
```go
if !e.Type().IsRegular() {
    continue
}
```
This correctly excludes directories, symlinks, pipes, devices, and other non-regular entries.

The practical impact is low (unusual edge case for a completion helper), but using `IsRegular()` would make the intent explicit and correct.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly narrows tab-completion candidates to aliases and agent YAML files, excluding directories and dotfiles. The sorting via `slices.Sorted(maps.Keys(...))` is a clean improvement. Test coverage is solid — four new tests cover the key scenarios. One minor edge case noted inline.

hamza-jeddad:
Addressed in the latest commit: replaced `e.IsDir()` with `!e.Type().IsRegular()` so symlinks-to-directories and other non-regular entries are also excluded.

PR #3009, opened 2026-06-05
fix(anthropic): route input_json deltas by content-block index

mcavage-docker:
## Summary

Parallel `tool_use` blocks streamed from Anthropic share a single `streamAdapter.toolID` field that gets overwritten on every `ContentBlockStartEvent`. Subsequent `InputJSONDelta` events for the first block carry the **second** block's tool ID, and the runtime accumulator (keyed by `ToolCall.ID` in `pkg/runtime/streaming.go`) concatenates both calls' argument fragments into one buffer. The first tool's args end up empty; the second tool's args contain interleaved fragments that fail to parse, surfacing as:

```
invalid character 's' looking for beginning of value
invalid character '-' after object key:value pair
```

(the exact character depends on where the chunk boundaries fall.)

This fix tracks tool IDs per content-block index via `toolIDByBlock map[int64]string` and routes `InputJSONDelta` events using `eventVariant.Index`. The same fix is applied to `betaStreamAdapter`, which shared the bug.

## How it manifests in production

Observed against a local GM agent team (cagent v1.57.0) running Claude Opus with ~18 parallel MCP tool calls in a single turn. Five calls failed with the JSON errors above before reaching the MCP gateway; the other thirteen succeeded. Retrying the same calls in isolation always succeeded — which made it look like a transient parser corruption issue, but the actual condition is deterministic: it only fires when the provider interleaves `input_json_delta` events across blocks, which is more likely under heavier fan-out.

The bug is present on current `main` (verified at HEAD `03386b31`). Bug was never reported via gh search for \"parallel tool\", \"invalid character\", or \"tool call arguments\" before this PR.

## Test plan

`pkg/model/provider/anthropic/parallel_toolcall_test.go` (added in this PR) feeds a fake SSE decoder a sequence of two parallel `tool_use` blocks with interleaved `input_json_delta` events, mirroring what Anthropic emits for parallel tool calls. Covers both `streamAdapter` and `betaStreamAdapter`.

Without the fix:

```
toolu_AAA args=\"\"
toolu_BBB args=\`{\"refresh_id\":\"abc-{\"category\":\"tool_failure\",def\"}\"summary\":\"x\"}\`
JSON parse error: invalid character 'c' after object key:value pair
```

With the fix:

```
toolu_AAA args=\`{\"refresh_id\":\"abc-def\"}\`
toolu_BBB args=\`{\"category\":\"tool_failure\",\"summary\":\"x\"}\`
```

- [x] New regression tests added (\`TestParallelToolCallIDsAreNotCrossWired\`, \`TestBetaParallelToolCallIDsAreNotCrossWired\`)
- [x] Both new tests fail on \`main\` without the patch
- [x] Both new tests pass with the patch
- [x] Full \`./pkg/model/provider/anthropic/...\` suite passes
- [x] Full \`./pkg/runtime/...\` suite passes
- [x] Full \`./pkg/model/...\` suite passes
- [x] \`gofmt\` clean, \`go vet\` clean

## Follow-up

\`pkg/model/provider/bedrock/adapter.go\` carries the same single-\`currentToolID\` pattern and is likely vulnerable to the same bug on parallel Bedrock Converse tool calls. Not patched here to keep the diff focused — happy to file a follow-up PR if useful.

### DIFF
--- a/pkg/model/provider/anthropic/adapter.go
+++ b/pkg/model/provider/anthropic/adapter.go
@@ -99,9 +99,6 @@ func (a *streamAdapter) Recv() (chat.MessageStreamResponse, error) {
 	case anthropic.ContentBlockStartEvent:
 		switch block := eventVariant.ContentBlock.AsAny().(type) {
 		case anthropic.ToolUseBlock:
-			if a.toolIDByBlock == nil {
-				a.toolIDByBlock = map[int64]string{}
-			}
 			a.toolIDByBlock[eventVariant.Index] = block.ID
 			a.toolCall = true
 			toolCall := tools.ToolCall{
--- a/pkg/model/provider/anthropic/adapter.go
+++ b/pkg/model/provider/anthropic/adapter.go
@@ -21,14 +21,20 @@ type streamAdapter struct {
 
 	trackUsage bool
 	toolCall   bool
-	toolID     string
 	stopReason anthropic.StopReason
+	// toolIDByBlock maps a content block index to its tool_use block ID.
+	// Anthropic emits each tool_use in its own content block; subsequent
+	// input_json_delta events carry the block index (not the tool ID), so
+	// we must remember the ID per block to route partial JSON correctly
+	// when multiple tool calls stream in parallel.
+	toolIDByBlock map[int64]string
 }
 
 func (c *Client) newStreamAdapter(stream *ssestream.Stream[anthropic.MessageStreamEventUnion], trackUsage bool) *streamAdapter {
 	return &streamAdapter{
 		retryableStream: retryableStream[anthropic.MessageStreamEventUnion]{stream: stream},
 		trackUsage:      trackUsage,
+		toolIDByBlock:   map[int64]string{},
 	}
 }
 
@@ -93,10 +99,13 @@ func (a *streamAdapter) Recv() (chat.MessageStreamResponse, error) {
 	case anthropic.ContentBlockStartEvent:
 		switch block := eventVariant.ContentBlock.AsAny().(type) {
 		case anthropic.ToolUseBlock:
-			a.toolID = block.ID
+			if a.toolIDByBlock == nil {
+				a.toolIDByBlock = map[int64]string{}
+			}
+			a.toolIDByBlock[eventVariant.Index] = block.ID
 			a.toolCall = true
 			toolCall := tools.ToolCall{
-				ID:   a.toolID,
+				ID:   block.ID,
 				Type: "function",
 				Function: tools.FunctionCall{
 					Name: block.Name,
@@ -123,7 +132,7 @@ func (a *streamAdapter) Recv() (chat.MessageStreamResponse, error) {
 		case anthropic.InputJSONDelta:
 			inputBytes := deltaVariant.PartialJSON
 			toolCall := tools.ToolCall{
-				ID:   a.toolID,
+				ID:   a.toolIDByBlock[eventVariant.Index],
 				Type: "function",
 				Function: tools.FunctionCall{
 					Arguments: inputBytes,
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is correct and well-targeted. Replacing the single `toolID string` field with a `toolIDByBlock map[int64]string` keyed by content-block index is the right solution for routing `InputJSONDelta` events across parallel `tool_use` blocks. Both `streamAdapter` and `betaStreamAdapter` are patched consistently, and the new regression tests faithfully reproduce the interleaved-delta scenario described in the PR.

**Verification summary**

| Finding | Severity | Verdict |
|---|---|---|
| Silent empty-string ToolCall ID on missing block index | Medium | ✅ DISMISSED — Anthropic API guarantees `content_block_start` before any delta; retry path restarts from the beginning so indices are re-populated before deltas arrive |
| `toolIDByBlock` not cleared on stream retry | Medium | ✅ DISMISSED — retry replaces the stream from the beginning; all `content_block_start` events are re-delivered before any deltas, overwriting any stale entries |
| Redundant `nil`-guard inside `Recv()` | Low | Informational note (see inline comment) |

One low-severity observation is left as an inline comment for awareness. No blockers found.

pkg/model/provider/anthropic/adapter.go
```diff
@@ -92,10 +98,13 @@ func (a *streamAdapter) Recv() (chat.MessageStreamResponse, error) {
 	case anthropic.ContentBlockStartEvent:
 		switch block := eventVariant.ContentBlock.AsAny().(type) {
 		case anthropic.ToolUseBlock:
-			a.toolID = block.ID
+			if a.toolIDByBlock == nil {
+				a.toolIDByBlock = map[int64]string{}
+			}
```
dgageot:
```suggestion
```

pkg/model/provider/anthropic/beta_adapter.go
```diff
@@ -56,10 +60,13 @@ func (a *betaStreamAdapter) Recv() (chat.MessageStreamResponse, error) {
 	case anthropic.BetaRawContentBlockStartEvent:
 		switch block := eventVariant.ContentBlock.AsAny().(type) {
 		case anthropic.BetaToolUseBlock:
-			a.toolID = block.ID
+			if a.toolIDByBlock == nil {
+				a.toolIDByBlock = map[int64]string{}
+			}
```
dgageot:
```suggestion
```

dgageot (dismissed):

kemalaysegulonay-del (approved):

dgageot (dismissed):

gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the parallel `tool_use` cross-wiring bug by replacing the single `toolID string` field with a `toolIDByBlock map[int64]string` keyed by content-block index, applied consistently to both `streamAdapter` and `betaStreamAdapter`. Map initialization in both constructors (`newStreamAdapter`, `newBetaStreamAdapter`) prevents nil-map panics.

**Review summary**

| Area | Result |
|---|---|
| Core fix correctness | ✅ Map-per-block-index correctly routes `InputJSONDelta` events |
| Nil-map safety | ✅ Both constructors initialize the map |
| Type correctness | ✅ `int64` index matches the SDK struct field type |
| Test coverage | ✅ Both `streamAdapter` and `betaStreamAdapter` paths covered with interleaved-delta scenarios |
| No regressions introduced | ✅ |

**Observations (informational, not blocking)**

1. *Stale map on retry (non-issue):* `toolIDByBlock` is not explicitly cleared when `retryableStream` transparently swaps the underlying SSE stream on a context-length error. However, the SSE protocol guarantees that `ContentBlockStartEvent` always precedes any `ContentBlockDeltaEvent` for a given block index — so the new stream's start events overwrite stale entries before any delta events can read them. No practical cross-wiring risk.

2. *`toolCall` flag on retry (pre-existing, out of scope):* The `toolCall bool` field (pre-existing, not changed by this PR) is never reset if the retried stream contains no tool calls — potentially causing an incorrect `FinishReasonToolCalls`. This pre-dates this PR and is out of scope here.

3. *Bedrock follow-up:* The PR description already flags `pkg/model/provider/bedrock/adapter.go` as carrying the same single-`currentToolID` pattern. Worth tracking as a follow-up.

The regression tests (`TestParallelToolCallIDsAreNotCrossWired`, `TestBetaParallelToolCallIDsAreNotCrossWired`) directly exercise the failure mode described in the PR and provide good coverage for the fix.

PR #3026, opened 2026-06-08
feat: add --session-read-only flag to view sessions without sending messages

gtardif:
When running with --session-read-only, the TUI opens in read-only mode:
- The editor shows a "Session is read-only" placeholder
- Attempting to send a message shows a warning notification
- Slash commands (e.g. /exit, /compact) still work
- File attachments are also blocked
- Using --exec with --session-read-only returns an error

This is useful for reviewing past session conversations loaded from --session-db without accidentally sending new messages to the LLM.

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -69,6 +69,7 @@ type runExecFlags struct {
 	worktree          bool
 	worktreeName      string
 	worktreePR        string
+	sessionReadOnly   bool
 
 	// Exec only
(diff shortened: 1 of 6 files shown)


dgageot (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27130367500).

PR #796, opened 2025-11-14
Skip flaky tests

dgageot:

PR #3029, opened 2026-06-09
fix(evals): copy docker-agent binary + entrypoint in custom-base-image template

hamza-jeddad:
## Summary

Fixes #796.

`docker-agent eval` runs each eval case in a freshly built container. `pkg/evaluation/build.go` picks one of two embedded templates:

- `Dockerfile.template` (default — used when the eval has no `image:`)
- `Dockerfile.custom.template` (used when the eval sets `evals.image:`)

The **custom** template was missing the two things the default template provides: it never copied the `docker-agent` binary into the image and never set the `/run.sh … docker-agent run …` entrypoint wrapper. As a result the eval container inherited the base image's `ENTRYPOINT ["/docker-agent"]`, and `eval.go` appended the agent YAML path as CMD, producing:

```
running docker agent in container: container failed: exit status 1
(stderr: Error: unknown command "/configs/<agent>.yaml" for "docker-agent")
```

This broke **every** custom-base-image eval (e.g. task-style evals that set a base image), while plain evals (no `image:`) kept working. Note that PR #2779 only fixed the `/run.sh` `printf` generation in the *default* template, so it did not address this.

## Fix

Bring `Dockerfile.custom.template` to parity with the default template:
- `COPY --from=docker/docker-agent:edge /docker-agent /`
- create the `/run.sh` exec wrapper
- `ENTRYPOINT ["/run.sh", "/docker-agent", "run", "--exec", "--yolo", "--json"]`
- add the same telemetry-suppression env vars and a `custom` image label

`FROM {{.BaseImage}}` and the `CopyWorkingDir` conditional are preserved, so the change is fully data-compatible with `build.go` (no Go changes needed).

## Tests

Added `pkg/evaluation/dockerfile_template_test.go`:
- `TestDockerfileCustomTemplateParity` — asserts the custom template copies the binary and sets the `/run.sh` entrypoint (guards against this exact regression).
- `TestDockerfileTemplatesRender` — renders both templates across the `CopyWorkingDir` matrix.

`go test ./pkg/evaluation/` passes. The full `go test ./...` suite passes except one pre-existing, host-specific `pkg/sandbox` test failure unrelated to this change.

## Scope

Eval harness only. The published `docker/docker-agent` image, `docker-agent run`/`serve`, and the TUI are unaffected — `Dockerfile.custom.template` is referenced solely by `pkg/evaluation/build.go`.

### DIFF
--- a/pkg/evaluation/dockerfile_template_test.go
+++ b/pkg/evaluation/dockerfile_template_test.go
@@ -1,0 +1,71 @@
+package evaluation
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// renderTemplate is a small helper to render the embedded Dockerfile
+// templates with the same fields buildEvalImage populates.
+func renderTemplate(t *testing.T, custom, copyWorkingDir bool, baseImage string) string {
+	t.Helper()
+
+	var data struct {
+		CopyWorkingDir bool
+		BaseImage      string
+	}
+	data.CopyWorkingDir = copyWorkingDir
+	data.BaseImage = baseImage
+
+	tmpl := dockerfileTemplate
+	if custom {
+		tmpl = dockerfileCustomTemplate
+	}
+
+	var buf bytes.Buffer
+	require.NoError(t, tmpl.Execute(&buf, data))
+	return buf.String()
+}
+
+// TestDockerfileCustomTemplateParity guards against the regression in
+// https://github.com/docker/docker-agent/issues/796: the custom-base-image
+// template must copy the docker-agent binary and wrap it with the /run.sh
+// entrypoint, exactly like the default template. Without these, the eval
+// container inherits the base image's `ENTRYPOINT ["/docker-agent"]` and the
+// agent YAML path is passed as a bare subcommand, producing
+// `unknown command "/configs/<agent>.yaml"`.
+func TestDockerfileCustomTemplateParity(t *testing.T) {
+	t.Parallel()
+
+	out := renderTemplate(t, true /* custom */, true /* copyWorkingDir */, "python:3.12")
+
+	assert.Contains(t, out, "FROM python:3.12",
+		"custom template must use the provided base image")
+	assert.Contains(t, out, "COPY --from=docker/docker-agent:edge /docker-agent /",
+		"custom template must copy the docker-agent binary into the eval image")
+	assert.Contains(t, out, `ENTRYPOINT ["/run.sh", "/docker-agent", "run", "--exec", "--yolo", "--json"]`,
+		"custom template must set the /run.sh docker-agent run entrypoint")
+	assert.Contains(t, out, "/run.sh",
+		"custom template must create the /run.sh entrypoint wrapper")
+}
+
+// TestDockerfileTemplatesRender ensures both templates render with and without
+// a working directory copy for the fields buildEvalImage supplies.
+func TestDockerfileTemplatesRender(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/evaluation/dockerfile_template_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	out := renderTemplate(t, true /* custom */, true /* copyWorkingDir */, "python:3.12")
+
+	assert.Contains(t, out, "FROM python:3.12",
+		"custom template must use the provided base image")
+	assert.Contains(t, out, "COPY --from=docker/docker-agent:edge /docker-agent /",
+		"custom template must copy the docker-agent binary into the eval image")
+	assert.Contains(t, out, `ENTRYPOINT ["/run.sh", "/docker-agent", "run", "--exec", "--yolo", "--json"]`,
+		"custom template must set the /run.sh docker-agent run entrypoint")
+	assert.Contains(t, out, "/run.sh",
```
docker-agent, line 51: **[LOW] Redundant `/run.sh` assertion is subsumed by the ENTRYPOINT assertion**

The assertion on line 51 (`assert.Contains(t, out, "/run.sh", ...)`) can never fail independently — it will always pass if the ENTRYPOINT assertion on line 49 passes, since the ENTRYPOINT string itself contains `/run.sh`. The comment says this guards that "the /run.sh entrypoint wrapper" is *created*, but it only checks that the string `/run.sh` appears anywhere in the output.

To actually guard against a regression where the `RUN printf ... > /run.sh` creation command is missing (but the ENTRYPOINT line still references it), consider asserting on the `RUN printf` line directly:
```go
assert.Contains(t, out, "RUN printf", "custom template must contain the RUN printf command that creates /run.sh")
```

pkg/evaluation/dockerfile_template_test.go
```diff
@@ -0,0 +1,71 @@
...
+		"custom template must create the /run.sh entrypoint wrapper")
+}
+
+// TestDockerfileTemplatesRender ensures both templates render with and without
+// a working directory copy for the fields buildEvalImage supplies.
+func TestDockerfileTemplatesRender(t *testing.T) {
+	t.Parallel()
+
+	for _, custom := range []bool{false, true} {
```
docker-agent, line 60: **[LOW] Loop in `TestDockerfileTemplatesRender` lacks `t.Run` sub-tests**

The nested `for` loops iterate over 4 combinations (`custom × copyWorkingDir`), but no `t.Run(...)` is used to give each combination a name. When an assertion fails, the test output will only say `TestDockerfileTemplatesRender failed` — you won't know which combination (`custom=true, copyWorkingDir=false`, etc.) triggered the failure.

Consider wrapping the loop body:
```go
for _, custom := range []bool{false, true} {
    for _, copyWorkingDir := range []bool{false, true} {
        name := fmt.Sprintf("custom=%v/copyWorkingDir=%v", custom, copyWorkingDir)
        t.Run(name, func(t *testing.T) {
            out := renderTemplate(t, custom, copyWorkingDir, "alpine:latest")
            // assertions...
        })
    }
}
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly brings `Dockerfile.custom.template` to parity with the default template by adding the missing `COPY --from=docker/docker-agent:edge`, `RUN printf` wrapper, `ENTRYPOINT`, and telemetry-suppression env vars. The regression guard test (`TestDockerfileCustomTemplateParity`) and render test (`TestDockerfileTemplatesRender`) are a solid addition. Two minor test-quality nits below.

melmennaoui (approved):

PR #1250, opened 2026-01-05
Fix for emojis rendering

dgageot:
Fixes #1233

PR #3031, opened 2026-06-09
fix: block IPv6 6to4, NAT64, site-local and CGNAT ranges in IsPublicIP

ronan-thibaut-glitch:
## What

Extends `IsPublicIP` in `pkg/httpclient/ssrf.go` to block two categories of addresses that bypass the existing Go stdlib checks:

- CGNAT `100.64.0.0/10` (RFC 6598): not covered by `IsPrivate()`
- IPv6 prefix families that embed arbitrary IPv4 addresses:
  - `2002::/16` (RFC 3056, 6to4)
  - `64:ff9b::/96` (RFC 6052, NAT64 well-known)
  - `64:ff9b:1::/48` (RFC 8215, NAT64 local-use)
  - `fec0::/10` (RFC 3879, site-local, deprecated)

On dual-stack or NAT64-enabled hosts (default in IPv6-only GKE, AWS IPv6-only EC2, Azure IPv6 VMs), addresses like `64:ff9b::a9fe:a9fe` translate directly to `169.254.169.254` (cloud metadata). The existing blocklist did not catch these.

Also adds a `mustCIDR` helper that panics at startup on a malformed CIDR literal, replacing silent failure at runtime.

## Why

Same class of vulnerability as GHSA-r48c-v28r-pf6v (modelcontextprotocol/registry), fixed upstream in their PR #1250. The `IsPublicIP` function is the SSRF guard for all outbound HTTP in docker-agent (fetch tool, api tool, openapi tool, a2a tool, MCP OAuth, skills cache, config loading). All verified live to pass the old filter.

## Testing

29 new test cases in `TestIsPublicIP` and 5 new cases in `TestSSRFDialControl`, including concrete attack IPs (e.g. `2002:a9fe:a9fe::` which encodes `169.254.169.254`) and boundary cases confirming no over-blocking.

### DIFF
--- a/pkg/httpclient/ssrf.go
+++ b/pkg/httpclient/ssrf.go
@@ -28,10 +28,10 @@ var cgnatRange = mustCIDR("100.64.0.0/10")
 // blockedIPv6Prefixes are IPv6 ranges that embed or forward arbitrary IPv4
 // addresses and must therefore be treated as non-public.
 var blockedIPv6Prefixes = []*net.IPNet{
-	mustCIDR("2002::/16"),        // RFC 3056  – 6to4 (bits 16-47 = IPv4)
-	mustCIDR("64:ff9b::/96"),     // RFC 6052  – NAT64 well-known
-	mustCIDR("64:ff9b:1::/48"),   // RFC 8215  – NAT64 local-use
-	mustCIDR("fec0::/10"),        // RFC 3879  – site-local (deprecated)
+	mustCIDR("2002::/16"),      // RFC 3056  – 6to4 (bits 16-47 = IPv4)
+	mustCIDR("64:ff9b::/96"),   // RFC 6052  – NAT64 well-known
+	mustCIDR("64:ff9b:1::/48"), // RFC 8215  – NAT64 local-use
+	mustCIDR("fec0::/10"),      // RFC 3879  – site-local (deprecated)
 }
 
 // IsPublicIP reports whether ip is a routable public address. It rejects
--- a/pkg/httpclient/ssrf.go
+++ b/pkg/httpclient/ssrf.go
@@ -12,17 +12,51 @@ import (
 	"time"
 )
 
+// mustCIDR parses a CIDR string and panics if it is invalid. Intended for
+// package-level var initialisation so bad literals fail at startup.
+func mustCIDR(s string) *net.IPNet {
+	_, network, err := net.ParseCIDR(s)
+	if err != nil {
+		panic("httpclient: invalid CIDR " + s + ": " + err.Error())
+	}
+	return network
+}
+
+// cgnatRange is RFC 6598 (100.64.0.0/10). Go's IsPrivate does not cover it.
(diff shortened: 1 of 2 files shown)


dgageot (approved):

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR adds important SSRF protections for CGNAT and IPv6 embedding ranges. One confirmed security bypass was found in the new code: IPv4-mapped IPv6 addresses in the CGNAT range (`::ffff:100.64.x.x`) pass the custom `cgnatRange.Contains()` check due to a byte-length mismatch in Go's `net.IPNet.Contains()`. The stdlib methods (`IsPrivate`, `IsLoopback`, etc.) handle this normalization internally, but the new custom check does not.

pkg/httpclient/ssrf.go
```diff
@@ -12,17 +12,51 @@ import (
...
-		!ip.IsUnspecified()
+	if ip.IsLoopback() ||
+		ip.IsPrivate() ||
+		ip.IsLinkLocalUnicast() ||
+		ip.IsLinkLocalMulticast() ||
+		ip.IsMulticast() ||
+		ip.IsUnspecified() {
+		return false
+	}
+	if cgnatRange.Contains(ip) {
```
docker-agent, line 51: **[HIGH] IPv4-mapped IPv6 CGNAT addresses bypass `cgnatRange.Contains()` — SSRF bypass**

`net.ParseIP("::ffff:100.64.0.1")` returns a **16-byte** `net.IP`, but `mustCIDR("100.64.0.0/10")` stores a **4-byte** network. Go's `net.IPNet.Contains()` checks `len(ip) == len(network.IP)` first; a length mismatch causes it to return `false` unconditionally.

As a result, any IPv4-mapped address in the CGNAT range (e.g. `::ffff:100.64.0.1`) passes through all checks and `IsPublicIP` returns `true` — a live SSRF bypass.

**Why the existing test doesn't catch this:** `TestIsPublicIP_IPv4MappedIPv6` tests `::ffff:127.0.0.1`, `::ffff:10.0.0.1`, and `::ffff:169.254.169.254`, all of which are blocked by `IsLoopback()`, `IsPrivate()`, or `IsLinkLocalUnicast()` — those stdlib methods do call `To4()` internally. The CGNAT range is explicitly **not** covered by `IsPrivate()` (hence the custom check), so the normalization gap is left open.

**Suggested fix:**
```go
if ip4 := ip.To4(); ip4 != nil {
    if cgnatRange.Contains(ip4) {
        return false
    }
} else if cgnatRange.Contains(ip) {
    return false
}
```
Or more concisely:
```go
checkIP := ip
if ip4 := ip.To4(); ip4 != nil {
    checkIP = ip4
}
if cgnatRange.Contains(checkIP) {
    return false
}
```
Also add `{"::ffff:100.64.0.1", false}` to `TestIsPublicIP_IPv4MappedIPv6` to pin this behaviour.

PR #3035, opened 2026-06-09
fix: use SSRF-safe HTTP client for MCP OAuth authorization server metadata fetch

### LINKED ISSUE #3034: fix: SSRF protection missing on MCP OAuth metadata fetch and remote transport data plane
## Summary

Two code paths in docker-agent issue outbound HTTP requests without the
SSRF guard provided by `httpclient.NewSafeClient` / `NewSSRFSafeTransport`.
Both were identified during the audit that led to PR #3031.

## FINDING-005 -- `pkg/tools/mcp/oauth_login.go` OAuth metadata fetch (HIGH)

`PerformOAuthLogin` uses an unguarded `&http.Client{Timeout: 5s}` as
`oauth.metadataClient` for authorization server metadata discovery.

Attack path: the MCP server returns a `/.well-known/oauth-protected-resource`
response whose `authorization_servers` field is attacker-controlled. The
`metadataClient` follows that URL unconditionally, reaching any internal
host the attacker specifies (e.g. `http://169.254.169.254/`).

Fix: replace with `httpclient.NewSafeClient(5s, false)`. PR pending.

## FINDING-006 -- `pkg/tools/mcp/remote.go` data plane transport (LOW)

`headerTransport()` returns `http.DefaultTransport` unconditionally.
The `allowPrivateIPs` flag controls the OAuth helper client only -- it
has no effect on the data channel connection.

This is a contract inconsistency rather than a direct SSRF attack path:
the MCP server URL is operator-supplied (not attacker-supplied), so
pointing it at an internal address is self-inflicted. However, when
`allowPrivateIPs: false` is set, the documented behaviour is that private
addresses are blocked -- the data plane does not honour that contract.

Fix: make `headerTransport()` return `httpclient.NewSSRFSafeTransport()`
when `allowPrivateIPs` is false. Separate PR, lower urgency.

## References

- PR #3031 -- fixed `IsPublicIP` for IPv6 / CGNAT gaps (same audit)
- GHSA-r48c-v28r-pf6v -- upstream SSRF advisory

ronan-thibaut-glitch:
## What

Replaces the unguarded `&http.Client{Timeout: 5s}` used as
`oauth.metadataClient` in `PerformOAuthLogin` with
`httpclient.NewSafeClient(5s, false)`, preserving the original timeout
while adding the SSRF-safe transport and redirect cap.

## Why

`PerformOAuthLogin` fetches `/.well-known/oauth-protected-resource` from
the configured MCP server, then follows the `authorization_servers` URL
returned in that response. That second URL is attacker-controlled -- a
malicious MCP server can point it at an internal host (e.g.
`http://169.254.169.254/`). The unguarded `metadataClient` would fetch
it unconditionally.

Closes #3034.

## References

- Issue #3034
- PR #3031 -- fixed `IsPublicIP` for IPv6 / CGNAT gaps (same audit)
- GHSA-r48c-v28r-pf6v -- upstream SSRF advisory

### DIFF
--- a/pkg/tools/mcp/oauth_login.go
+++ b/pkg/tools/mcp/oauth_login.go
@@ -11,6 +11,8 @@ import (
 	"time"
 
 	"golang.org/x/oauth2"
+
+	"github.com/docker/docker-agent/pkg/httpclient"
 )
 
 // PerformOAuthLogin performs a standalone OAuth flow for the given MCP server URL.
@@ -19,7 +21,7 @@ import (
 func PerformOAuthLogin(ctx context.Context, serverURL string) error {
 	tokenStore := NewKeyringTokenStore()
 
-	o := &oauth{metadataClient: &http.Client{Timeout: 5 * time.Second}}
+	o := &oauth{metadataClient: httpclient.NewSafeClient(5*time.Second, false)}
 
 	// Derive the base origin (scheme + host) from the server URL.
 	// The well-known endpoints live at the origin, not under the SSE/path.
--- a/pkg/tools/mcp/oauth_login.go
+++ b/pkg/tools/mcp/oauth_login.go
@@ -37,7 +37,7 @@ func PerformOAuthLogin(ctx context.Context, serverURL string) error {
 	if err != nil {
 		return fmt.Errorf("failed to create resource metadata request: %w", err)
 	}
-	resp, err := http.DefaultClient.Do(resourceReq)
+	resp, err := o.metadataClient.Do(resourceReq)
 	if err != nil {
 		return fmt.Errorf("failed to fetch protected resource metadata: %w", err)
 	}


pkg/tools/mcp/oauth_login.go
```diff
@@ -19,7 +21,7 @@ import (
 func PerformOAuthLogin(ctx context.Context, serverURL string) error {
 	tokenStore := NewKeyringTokenStore()
 
-	o := &oauth{metadataClient: &http.Client{Timeout: 5 * time.Second}}
+	o := &oauth{metadataClient: httpclient.NewSafeClient(5*time.Second, false)}
```
docker-agent, line 24: **[HIGH] Incomplete SSRF fix — first metadata fetch still uses `http.DefaultClient`**

The safe client created here (`o.metadataClient`) is only used by the subsequent `o.getAuthorizationServerMetadata` call. The **first** outbound request in this function — the `GET /.well-known/oauth-protected-resource` fetch 16 lines below — still uses the bare `http.DefaultClient`:

```go
resp, err := http.DefaultClient.Do(resourceReq)  // line 40 — unguarded
```

This means the PR's SSRF protection is incomplete. Any `serverURL` whose hostname resolves to a private/internal address (e.g. `http://192.168.1.1/`, `http://169.254.169.254/`) will still trigger an unguarded outbound request during the first fetch. The fix should use `o.metadataClient.Do(resourceReq)` instead of `http.DefaultClient.Do(resourceReq)`.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

The PR introduces an SSRF-safe HTTP client for MCP OAuth metadata fetches, but the fix is **incomplete**. The newly created safe client (`o.metadataClient`) is only wired into the second fetch (`getAuthorizationServerMetadata`), while the first outbound request — `GET /.well-known/oauth-protected-resource` — still uses the unguarded `http.DefaultClient`. A `serverURL` pointing at an internal host will bypass SSRF protection on that initial call.

pkg/tools/mcp/oauth_login.go
ronan-thibaut-glitch: Fixed -- both fetches now use `o.metadataClient`. The agent config can be loaded from a remote URL or OCI reference, making `serverURL` attacker-influenced in those cases.

dgageot (approved):

pkg/tools/mcp/oauth_login.go
docker-agent: Confirmed fixed — both metadata fetches now use the safe client. Thanks for the additional context about remote config loading; that makes the attack surface much clearer and confirms why guarding both requests was essential.

PR #3036, opened 2026-06-09
feat: add 1Password CLI integration for secret resolution

dgageot:
LLM provider tokens and other secrets read through the environment provider chain can now be resolved from 1Password. When a resolved environment value starts with `op://`, docker-agent resolves it via the 1Password CLI (`op read`) before handing it to a model provider or tool.

The implementation adds a new `OnePasswordProvider` in `pkg/environment/onepassword.go` that decorates the existing provider chain. It resolves `op://` references and passes all other values through unchanged. The provider is wired into the default environment setup so it applies across OS environment variables, run secrets, credential helpers, Docker Desktop, pass, and keychain. The `op` binary is resolved lazily; if it's unavailable or a resolution fails, docker-agent logs a warning and treats the variable as unset rather than forwarding the raw `op://` reference.

As a hardening measure, the shared `lookupBinary` helper in `pkg/environment/cmd_provider.go` now rejects `exec.ErrDot` and non-absolute paths, preventing PATH hijacking (CWE-426). This also strengthens the existing pass, keychain, and credential-helper providers.

### DIFF
--- a/docs/guides/secrets/index.md
+++ b/docs/guides/secrets/index.md
@@ -1,6 +1,6 @@
 ---
 title: "Managing Secrets"
-description: "How to securely provide API keys and credentials to docker-agent using environment variables, env files, Docker Compose secrets, macOS Keychain, and pass."
+description: "How to securely provide API keys and credentials to docker-agent using environment variables, env files, Docker Compose secrets, macOS Keychain, pass, and 1Password references."
 permalink: /guides/secrets/
 ---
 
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The 1Password CLI integration is well-implemented. No bugs were found in the changed code.

**What was reviewed:**
- `pkg/environment/onepassword.go` — `OnePasswordProvider` decorator with lazy `op` binary lookup, correct pass-through for non-`op://` values, and safe error handling (returns unset rather than leaking raw `op://` references).
- `pkg/environment/cmd_provider.go` — hardened `lookupBinary` now correctly rejects `exec.ErrDot` and non-absolute paths (CWE-426 prevention). The logic is sound: the `if !errors.Is(err, exec.ErrNotFound) && !errors.Is(err, exec.ErrDot)` guard only warns on unexpected errors.
- `pkg/environment/default.go` — wrapping the entire `MultiProvider` chain with `NewOnePasswordProvider` ensures every resolved value (regardless of which provider supplied it) gets `op://` resolution applied uniformly.
- `pkg/environment/onepassword_test.go` — tests cover plain pass-through, successful resolution, missing variable, and failed resolution. `t.Context()` usage is idiomatic.
- `docs/guides/secrets/index.md` — documentation accurately reflects the implementation behavior including the failure mode (variable treated as unset).

**Security considerations verified:**
- `op://` references are never forwarded raw to model providers on resolution failure ✅
- `op` binary is resolved to an absolute path before execution ✅
- `exec.ErrDot` is treated as not-found (no CWD-relative binary execution) ✅
- `runCommand` uses `strings.TrimSpace` on output, preventing trailing-newline contamination of secrets ✅

gtardif (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27217686608).

PR #3041, opened 2026-06-09
Allow models.dev in sandbox proxy for model catalog resolution

dgageot:
## What

`docker agent run --sandbox` now always allowlists the `models.dev`
catalog host in the sandbox's default-deny network proxy, so the
in-sandbox agent can fetch the model list and resolve model metadata
(context limits, pricing, capabilities).

## Why

Every agent run resolves model metadata against `models.dev`, but the
sandbox proxy denies it by default. Without an explicit allow rule the
inner agent's first catalog lookup fails with a misleading
`403 Blocked by network policy`. This wires `models.dev` into the same
auto-allowlist machinery that already opens the models gateway and the
auto-installer's package hosts.

## Changes

- `pkg/modelsdev`: export `APIHost = "models.dev"` and derive
  `ModelsDevAPIURL` from it so the host and URL can't drift.
- `cmd/root/sandbox.go`: always include `modelsdev.APIHost` in
  `allowSandboxHosts`, and surface it via a new `printModelsDevAllowance`
  line alongside the gateway/tool-install output.
- `cmd/root/sandbox_test.go`: add `TestPrintModelsDevAllowance`.

## Validation

- `go build ./...`
- `go test ./cmd/root/ ./pkg/modelsdev/`
- `golangci-lint run` (0 issues)

### DIFF
--- a/cmd/root/sandbox.go
+++ b/cmd/root/sandbox.go
@@ -20,6 +20,7 @@ import (
 	"github.com/docker/docker-agent/pkg/config"
 	latestcfg "github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
+	"github.com/docker/docker-agent/pkg/modelsdev"
 	"github.com/docker/docker-agent/pkg/paths"
 	"github.com/docker/docker-agent/pkg/sandbox"
 	"github.com/docker/docker-agent/pkg/sandbox/kit"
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are clean, correct, and well-structured:

- **`pkg/modelsdev`**: Exporting `APIHost = "models.dev"` and deriving `ModelsDevAPIURL` from it is a sound refactor — it eliminates the possibility of the host and URL drifting apart.
- **`sandbox.go`**: Unconditionally prepending `modelsdev.APIHost` to the allowlist is the right approach since every run resolves model metadata, and the new `printModelsDevAllowance` helper follows the same pattern as the existing gateway/tool-install print functions.
- **`sandbox_test.go`**: `TestPrintModelsDevAllowance` is correctly structured with `t.Parallel()` and a `strings.Builder`.

No bugs found in the added code.

melmennaoui (approved):

PR #3042, opened 2026-06-09
fix: agent loses context and halts after first session compaction

### LINKED ISSUE #2871: Agent loses context and halts after first session compaction
### Description

After the first session compaction, the agent stops mid-task and responds as if it has no conversation history, outputting something like:

Image

This breaks the agent's current thought process entirely. The only workaround is to re-enter the original prompt manually, after which it manages to continue — but all in-progress state is lost.

### Steps to Reproduce

1. Start a `docker-agent` session with a multi-agent setup
2. Let the session run until the first compaction triggers
3. Observe the agent halt and output the "no conversation history" message
4. Re-enter the original prompt to confirm the agent can resume

### Expected Behavior

The agent should handle compaction transparently and continue its work without interruption or loss of context.

### Actual Behavior

- Agent halts immediately after the first compaction
- Outputs a confused "no history" message
- Breaks the thought process of the agent team
- Requires manual intervention to resume

### Additional Context

- Happens on the **first** compaction, not subsequent ones
- Appeared **recently** with no changes to the main `.yaml` config — this is a regression
- Re-prompting does allow the agent to continue, suggesting the compacted context is not fully lost — the agent just fails to resume autonomously

rumpl:
Fixes #2871

## Problem

After the first session compaction in a multi-agent run, the agent halts mid-task and replies as if it has no conversation history ("I understand you're looking for a session summary, but ... no previous conversation history visible").

## Root cause

Two compounding bugs, surfaced by ead97457 / 8dba51f4 (2026-05-18) which expanded when compaction activates — four days before the issue was filed:

1. **Phantom trigger in multi-agent runs**: `compactIfNeeded` estimated newly-added tokens via `sess.GetAllMessages()`, which recurses into sub-sessions. The content produced by a `transfer_task` child was attributed to the parent session even though it never enters the parent's prompt (`GetMessages` skips sub-session items). The phantom tokens triggered compaction of a parent conversation that was actually tiny; with everything fitting the keep budget, the split resolved to the "compact everything, keep nothing" sentinel — wiping the user's task and the in-flight tool exchange. The agent's next prompt was literally just `Session Summary: ...`, which models read as the user *asking* for a summary. This also explains the "first compaction only" symptom: the first compaction fires while the parent history is still tiny; after re-prompting, later compactions keep a real tail.

2. **Fixed budgets break small context windows**: `MaxSummaryTokens` (16k) and `maxKeepTokens` (20k) are absolute constants. For models whose window resolves from `provider_opts.context_size` and is ≤ ~16k, the summarizer's input budget went to zero — it received only its own prompts, fabricated a "no history" non-summary, and that text replaced the entire session history.

## Fix

- `Session.OwnMessages()` (no sub-session recursion) now drives the compaction trigger's token accounting, so sub-agent work no longer causes phantom parent compactions.
- Summary/keep budgets scale with the window (`min(16k, limit/4)` / `min(20k, limit/5)`); the scaled cap is also used for the summary call's `max_tokens`.
- Safety net: `RunLLM` no-ops when no conversation message fits the summarization budget, instead of running the summarizer on an empty conversation and wiping history with the result.
- `ComputeFirstKeptEntry` gains a `contextLimit` parameter so hook-supplied summaries share the same kept-tail policy.

## Tests

- `TestCompactIfNeeded_IgnoresSubSessionTokens` — regression test, verified to fail against the old trigger code.
- `TestCompactIfNeeded_TriggersOnOwnMessages` — large own tool results still trigger.
- `TestRunLLM_SmallContextWindow` — summarizer receives real conversation on an 8k window and a tail is kept.
- `TestRunLLM_NoConversationFits_NoOps` — empty summarizer input no-ops instead of wiping history.

`task build`, `task test`, `task lint` all pass (only pre-existing, environment-dependent `pkg/sandbox.TestExtraWorkspace` failure, which also fails on clean `main`).

## Note for reviewers

One residual hazard left untouched (documented contract defended in 1e9512e0): a *legitimately* triggered compaction whose whole conversation fits the keep budget (possible with image-heavy histories — token estimates ignore images) still drops the tail via the "compact everything" sentinel. Happy to follow up with a "keep the last user turn on threshold/overflow compaction" change if desired.

Assisted-By: docker-agent

### DIFF
--- a/pkg/runtime/compaction_trigger_test.go
+++ b/pkg/runtime/compaction_trigger_test.go
@@ -1,0 +1,106 @@
+package runtime
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/pkg/agent"
+	"github.com/docker/docker-agent/pkg/chat"
+	"github.com/docker/docker-agent/pkg/session"
+	"github.com/docker/docker-agent/pkg/team"
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+// TestCompactIfNeeded_IgnoresSubSessionTokens is a regression test for
+// issue #2871: in a multi-agent run, the tokens accumulated inside a
+// transfer_task sub-session were counted by the proactive compaction
+// trigger (GetAllMessages recurses into sub-sessions) even though they
+// never enter the parent's prompt (GetMessages skips sub-session items).
+// The phantom tokens made the parent compact its own tiny conversation;
+// with everything fitting the keep budget that meant "compact
+// everything, keep nothing" — the agent's next prompt was just the
+// summary and it halted with a confused "no conversation history" reply.
+func TestCompactIfNeeded_IgnoresSubSessionTokens(t *testing.T) {
+	prov := &mockProvider{id: "test/model", stream: &mockStream{}}
+	root := agent.New("root", "agent", agent.WithModel(prov))
+	tm := team.New(team.WithAgents(root))
+
+	rt, err := NewLocalRuntime(tm,
+		WithSessionCompaction(true),
+		WithModelStore(mockModelStoreWithLimit{limit: 100_000}))
+	require.NoError(t, err)
+
+	sess := session.New(session.WithUserMessage("build the app"))
+	messageCountBefore := len(sess.OwnMessages())
+
+	// Simulate a completed transfer_task tool call: a sub-session holding
+	// far more content than the parent's context limit, plus a small
+	// tool-result message on the parent itself.
+	sub := session.New(session.WithUserMessage("subtask"))
+	sub.AddMessage(session.NewAgentMessage("worker", &chat.Message{
+		Role:    chat.MessageRoleAssistant,
+		Content: strings.Repeat("z", 600_000), // ~150k estimated tokens
+	}))
+	sess.AddMessage(session.NewAgentMessage("root", &chat.Message{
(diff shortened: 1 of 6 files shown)


dgageot (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly addresses two compounding compaction bugs:

1. **Phantom token trigger** — switching from `GetAllMessages()` (which recurses into sub-sessions) to `OwnMessages()` (which does not) ensures sub-agent token counts no longer falsely trigger parent-session compaction.
2. **Fixed budget overflow** — scaling `MaxSummaryTokens` and `maxKeepTokens` proportionally to the context window (`limit/4` and `limit/5`) prevents the summarizer from consuming the entire budget on small-window models, and the `len(messages) <= 2` no-op guard correctly prevents a fabricated non-summary from replacing real session history.

**Verification summary:**
- The `ApplyCompaction` path only *appends* to `s.Messages`, so the `sess.OwnMessages()[messageCountBefore:]` slice in `compactIfNeeded` cannot panic (length is monotonically non-decreasing in a single-goroutine call chain).
- `OwnMessages()` excluding system-role items is intentional and consistent with `GetAllMessages()`; the invariant system messages in `GetMessages()` are built dynamically and were never stored in session items.
- All four new tests (`TestCompactIfNeeded_IgnoresSubSessionTokens`, `TestCompactIfNeeded_TriggersOnOwnMessages`, `TestRunLLM_SmallContextWindow`, `TestRunLLM_NoConversationFits_NoOps`) directly target the described regression scenarios.

No confirmed or likely bugs found in the changed code.

PR #3043, opened 2026-06-09
docs: fix stale defaults, wrong tool names, and missing CLI flags

dgageot:
A full review of the docs site against the current code turned up a number of claims that were false or out of date. This fixes all of them, verified against the CLI help output, `agent-schema.json`, and the relevant Go packages.

The biggest offenders were the config version (documented as 8, current is 10), the Anthropic/Bedrock thinking defaults (several pages contradicted each other; per `pkg/model/provider/defaults.go`, thinking is off unless `thinking_budget` is set and `interleaved_thinking` is then auto-enabled for Claude models), and the model-picker page, which documented a `pick_model` tool that does not exist — the real tools are `change_model` and `revert_model`, and the switch is immediate rather than next-turn.

Smaller fixes: the `serve a2a --agent` default is the team's first agent, not `root`; the filesystem post-edit hook docs referenced a nonexistent `patch_file` tool; the RAG pages used `batch_size` instead of `embedding_batch_size` and documented wrong defaults for chunk size and reranking `top_k`; and the CLI/TUI references were missing `--theme`, `--on-event`, `--tool-name`, `--mcp-keepalive`, `--attach`, and the `/skills` command. Docs-only change, no code touched.

### DIFF
--- a/docs/concepts/models/index.md
+++ b/docs/concepts/models/index.md
@@ -108,13 +108,13 @@ See the [Model Providers]({{ '/providers/overview/' | relative_url }}) section f
 
 Control how much the model "thinks" before responding:
 
-| Provider   | Format     | Values                                | Default      |
-| ---------- | ---------- | ------------------------------------- | ------------ |
-| OpenAI     | string     | `minimal`, `low`, `medium`, `high`    | `medium`     |
-| Anthropic  | int        | 1024–32768 tokens                     | 8192         |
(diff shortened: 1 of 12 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27233759595).

aheritier (approved):

PR #3045, opened 2026-06-10
Reusable editor

dgageot:


pkg/tui/components/editor/editor.go
```diff
@@ -176,6 +177,13 @@ func WithReadOnly() Option {
 	}
 }
 
+// WithPlaceholder sets the editor's placeholder text (shown while empty).
+func WithPlaceholder(placeholder string) Option {
```
docker-agent, line 181: **[MEDIUM] `WithPlaceholder` value is silently discarded when `SetRecording(false)` is called**

The new `WithPlaceholder` option stores the custom placeholder only in `e.textarea.Placeholder` at construction time. However, `SetRecording(false)` unconditionally overwrites it with the hard-coded default:

```go
// line 1469 — always reverts to the hard-coded string
e.textarea.Placeholder = "Type your message here…"
```

Any consumer that calls `New(..., WithPlaceholder("Ask anything..."))` and later invokes `SetRecording(false)` (e.g., when a voice-input session ends) will silently lose their custom placeholder. The fix is to store the custom placeholder in an `editor` struct field (e.g., `e.defaultPlaceholder`) and restore that field in `SetRecording` instead of the hard-coded literal.

docker-agent (commented):

One confirmed medium-severity issue found in the new `WithPlaceholder` option.

gtardif (approved):

PR #3047, opened 2026-06-10
Reusable editor (More)

dgageot:


aheritier (approved):

pkg/tui/components/editor/editor.go
```diff
@@ -1466,7 +1487,7 @@ func (e *editor) SetRecording(recording bool) tea.Cmd {
 		e.textarea.Placeholder = "🎤 Listening"
 		return e.tickRecordingDots()
 	}
-	e.textarea.Placeholder = "Type your message here…"
+	e.textarea.Placeholder = e.placeholder
```
docker-agent, line 1490: **[MEDIUM] `SetRecording(false)` overwrites the read-only placeholder with the default placeholder**

The new `e.placeholder` field (introduced in this PR) is restored here when recording ends. However, `WithReadOnly()` sets `e.textarea.Placeholder = "Session is read-only"` directly **without** updating `e.placeholder`. As a result, if `SetRecording(true)` is called on a read-only editor and then `SetRecording(false)` is called, this line will restore `e.placeholder` — which holds `defaultPlaceholder` ("Type your message here…") rather than the read-only text — silently removing the read-only indicator from the UI.

**Fix**: Update `WithReadOnly` to also set `e.placeholder`:
```go
func WithReadOnly() Option {
    return func(e *editor) {
        e.placeholder = "Session is read-only"
        e.textarea.Placeholder = "Session is read-only"
        e.textarea.KeyMap.InsertNewline.SetEnabled(false)
    }
}
```

docker-agent (commented):

One LIKELY issue was found in the changed code. One finding was dismissed after analysis confirmed the arithmetic is always safe.

PR #3048, opened 2026-06-10
Remove MCP non-working servers 

trungutt:
The Supabase and Tally entries in the MCP catalog do not work reliably through the OAuth flow. Until their OAuth integration is fixed, exposing them in the catalog leads users into a broken first-run experience, so drop them for now.

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/servers.json
+++ b/pkg/tools/builtin/mcpcatalog/servers.json
@@ -726,46 +726,6 @@
       }
     },
     {
-      "id": "supabase",
-      "title": "Supabase",
-      "description": "Manage Supabase projects, run database queries, edit schemas and configure edge functions and storage.",
-      "url": "https://mcp.supabase.com/mcp",
-      "transport": "streamable-http",
-      "category": "database",
-      "tags": [
-        "database",
-        "postgres",
-        "backend",
-        "auth",
-        "storage",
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR removes the `supabase` and `tally` entries from the MCP catalog due to broken OAuth flows. The JSON structure is valid after removal, comma handling is correct, and no other entries are affected.

**Note**: The top-level `count` field (line 6) still reads `35` but there are now `33` entries — consider updating it to stay in sync. This is a minor housekeeping item.

pkg/tools/builtin/mcpcatalog/servers.json
```diff
@@ -725,46 +725,6 @@
         ]
       }
     },
-    {
```
dgageot, line 728: @trungutt probably you'll need to also update `"count": 35,`

dgageot (changes requested):

rumpl:
Do we know why these don't work?

pkg/tools/builtin/mcpcatalog/servers.json
trungutt: Good catch — updated the count to 33 in 613a0d59c.

dgageot (approved):

PR #3049, opened 2026-06-10
fix: stop sidebar token usage panel flickering during sub-agent transfers

dgageot:
The sidebar's Token Usage panel flickered between two different numbers while a sub-agent ran. This happened because the panel picked which session to display using the `currentAgent` heuristic, which toggles rapidly as sub-agents are spawned and cleaned up. The fix tracks the active session via a stream stack instead: we push on `StreamStarted` and pop on `StreamStopped`, so the panel stably shows the active (sub-)session's tokens and context, then reverts to the parent once the sub-agent returns.

The second commit hardens this against drift: a new `ResetStreamTracking()` method (called from the chat page's `processMessage`, mirroring its existing `streamDepth=0` reset) plus clearing the stack on `LoadFromSession` prevents leaked entries from pinning the panel to a stale sub-session or growing unbounded when a stream ends without a balanced `StreamStoppedEvent` (for example, when a context cancel occurs without a corresponding `StreamCancelledMsg`).

All changes are confined to `pkg/tui/components/sidebar/` and `pkg/tui/page/chat/chat.go`, with updated and added unit tests. Linter and the full test suite both pass.

### DIFF
--- a/pkg/tui/components/sidebar/sidebar.go
+++ b/pkg/tui/components/sidebar/sidebar.go
@@ -58,6 +58,10 @@ type Model interface {
 	SetQueuedMessages(messages ...string)
 	GetSize() (width, height int)
 	LoadFromSession(sess *session.Session)
+	// ResetStreamTracking clears the active-stream stack so a new top-level run
+	// starts from a clean slate, even if a previous run's stream events were
+	// left unbalanced (e.g. cancelled without a StreamCancelledMsg).
+	ResetStreamTracking()
 	// HandleClick checks if click is on the star or title and returns true if handled
 	HandleClick(x, y int) bool
 	// HandleClickType returns the type of click (star, title, agent, or none).
@@ -455,8 +459,10 @@ func (m *model) LoadFromSession(sess *session.Session) {
 		}
 	}
 
-	// The restored session is the main session until a stream starts.
+	// The restored session is the main session until a stream starts. A freshly
+	// loaded session has no in-flight streams, so clear any stale stack entries.
 	m.rootSessionID = sess.ID
+	m.sessionStack = nil
 
 	// Load session title
 	if sess.Title != "" {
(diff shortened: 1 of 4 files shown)


pkg/tui/components/sidebar/sidebar.go
```diff
@@ -700,7 +715,13 @@ func (m *model) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
...
 		m.streamCancelled = false
 		m.workingAgent = msg.AgentName
-		m.currentSessionID = msg.SessionID
+		// Track the active session via a stack: the outermost stream owns the
+		// main session; nested sub-agent streams are pushed on top so their
+		// usage is shown while they run, then popped when they stop.
+		if len(m.sessionStack) == 0 {
+			m.rootSessionID = msg.SessionID
+		}
+		m.sessionStack = append(m.sessionStack, msg.SessionID)
```
docker-agent, line 724: **[MEDIUM] No deduplication guard in `StreamStartedEvent` handler — duplicate events cause permanent stack imbalance**

The new `sessionStack` append logic has no guard against duplicate events for the same session ID. If `StreamStartedEvent` fires twice for the same `msg.SessionID` (e.g. due to reconnect or replay by the runtime layer), the ID is pushed onto `sessionStack` twice:

```go
m.sessionStack = append(m.sessionStack, msg.SessionID) // no dedup check
```

When the session ends, `StreamStoppedEvent` pops only one entry (it pops unconditionally, not by matching the session ID), leaving a stale entry behind. `activeSessionID()` will then permanently return that session ID rather than falling back to `rootSessionID`, keeping the token panel pinned to a stale sub-session until the user presses ESC (`StreamCancelledMsg` nil-clears the stack) or `processMessage` calls `ResetStreamTracking()`.

**Suggestion:** Add a deduplication guard before the append, or pop by matching session ID in `StreamStoppedEvent`:

```go
// Option A: dedup on push
if len(m.sessionStack) == 0 || m.sessionStack[len(m.sessionStack)-1] != msg.SessionID {
    m.sessionStack = append(m.sessionStack, msg.SessionID)
}

// Option B: pop by matching ID in StreamStoppedEvent (more robust for unordered events)
case *runtime.StreamStoppedEvent:
    for i := len(m.sessionStack) - 1; i >= 0; i-- {
        if m.sessionStack[i] == msg.SessionID {
            m.sessionStack = append(m.sessionStack[:i], m.sessionStack[i+1:]...)
            break
        }
    }
```

docker-agent (commented):

The stack-based approach for tracking active stream sessions is a clean improvement over the previous sessionAgent map + currentSessionID approach. The core logic is sound: StreamStartedEvent pushes, StreamStoppedEvent pops with an underflow guard, StreamCancelledMsg nil-clears the entire stack, and LoadFromSession/ResetStreamTracking provide clean-slate reset paths. One structural weakness was identified.

trungutt (approved):

PR #3050, opened 2026-06-10
fix: add max_tokens floor for Anthropic when thinking is disabled

dgageot:
Newer adaptive-thinking Claude models (such as claude-fable-5) reason internally even when thinking is nominally disabled. With a 20-token budget from title generation's `WithNoThinking()` call, this internal reasoning consumes the entire allocation, leaving no tokens for visible output. The model returns empty text, causing title generation to fail.

The OpenAI provider already handles this by enforcing a 256-token output floor for reasoning models when thinking is disabled. This ensures hidden reasoning tokens can't starve visible output. Anthropic had no equivalent protection.

This fix adds `noThinkingMinOutputTokens = 256` and a `floorMaxTokensForNoThinking` helper in `pkg/model/provider/anthropic/thinking.go`, wired into `CreateChatCompletionStream` right after `adjustMaxTokensForThinking`. The floor applies only when thinking is disabled, ensuring it never overrides an explicit user-set cap during normal completions. It's intentionally gated on `NoThinking` rather than a model-name allowlist, so it covers new and unknown reasoning models automatically.

Test coverage includes scenarios for raising tiny caps to the floor, no-op above the floor, equal-to-floor, and thinking-enabled paths remaining untouched.

### DIFF
--- a/pkg/model/provider/anthropic/thinking_test.go
+++ b/pkg/model/provider/anthropic/thinking_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/model/provider/base"
+	"github.com/docker/docker-agent/pkg/model/provider/options"
 )
 
 func TestAnthropicThinkingDisplay(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/model/provider/anthropic/thinking_test.go
```diff
@@ -470,6 +471,55 @@ func TestCoerceAdaptiveThinking(t *testing.T) {
...
+
+	tests := []struct {
+		name      string
+		opts      options.ModelOptions
+		maxTokens int64
+		want      int64
+	}{
+		{
+			name:      "no-thinking tiny cap is raised to floor",
+			opts:      buildOpts(options.WithNoThinking(), options.WithMaxTokens(20)),
```
docker-agent, line 491: **[LOW] Dead option in test: `options.WithMaxTokens(20)` has no effect**

In the `"no-thinking tiny cap is raised to floor"` test case, `options.WithMaxTokens(20)` is passed to `buildOpts(...)` alongside `options.WithNoThinking()`. However, `floorMaxTokensForNoThinking` receives `maxTokens` as a plain `int64` argument and never reads `c.ModelOptions.MaxTokens()`. The `WithMaxTokens(20)` call is therefore dead code — it stores a value on `ModelOptions` that the function under test ignores entirely.

The test still passes because the struct field `maxTokens: 20` correctly drives the argument. However, the extraneous option creates a misleading impression that `floorMaxTokensForNoThinking` also respects the `ModelOptions.MaxTokens()` setting, which it does not. Suggest removing `options.WithMaxTokens(20)` from this test case to keep the opts minimal and accurate.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The floor logic is correctly placed, correctly gated on `NoThinking()`, and correctly skipped for thinking-enabled paths. The medium-severity hypothesis (conflict between `adjustMaxTokensForThinking` and `floorMaxTokensForNoThinking`) was verified and dismissed — even if both are set simultaneously, the floor only fires below 256 tokens, so any prior raise by `adjustMaxTokensForThinking` would already exceed it. One minor test clarity issue was flagged inline.

aheritier (approved):

PR #3051, opened 2026-06-10
feat: add title_model for delegating session-title generation

dgageot:
Adds a new `title_model` field to the agent configuration, allowing a model to declare which other model should be used for generating session titles. This lets a heavyweight primary model delegate the cheap title-generation call to a smaller or faster model. The value can be a named model from the `models` section or an inline `provider/model` spec; when omitted, title generation reuses the agent's own model.

The change touches the config schema (`pkg/config/latest`), agent initialization with a new `WithTitleModel` option and `TitleModel()` getter, title-model resolution in `pkg/teamloader`, and the title generators in `pkg/runtime` and `pkg/server/session_manager.go`. The `debug title` command now respects the `title_model` setting. A validation guard ensures `first_available` cannot be combined with `title_model`, since that resolution would otherwise silently drop the delegation.

Schema documentation and an example configuration have been added, along with test coverage for title-model resolution and the new validation constraint.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1527,6 +1527,14 @@
               "dmr/ai/qwen3"
             ]
           ]
+        },
+        "title_model": {
+          "type": "string",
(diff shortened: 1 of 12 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27271980173).

trungutt (approved):

PR #3052, opened 2026-06-10
fix: surface useful errors when session title generation fails

dgageot:
When session title generation failed, the error message was unhelpful — users would see something like `docker agent debug title: exit status 1` with no indication of what actually went wrong.

The fix improves error reporting in two ways. The `pkg/sessiontitle` generator now tags each failing model attempt with its model ID and aggregates all attempts via `errors.Join`, instead of keeping only the last error. The error message now reads like `all N title model(s) failed: model "X": <cause>`. The `docker agent debug title` command no longer double-wraps the error and includes the agent name for context.

A unit test asserts that the failure message lists every model and its cause. Changes validated with `task lint` (0 issues) and `task test` (all pass).

### DIFF
--- a/cmd/root/debug.go
+++ b/cmd/root/debug.go
@@ -225,7 +225,7 @@ func (f *debugFlags) runDebugTitleCommand(cmd *cobra.Command, args []string) (co
 
 	title, err := gen.Generate(ctx, "debug", []string{args[1]})
 	if err != nil {
-		return fmt.Errorf("generating title: %w", err)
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes correctly improve error reporting for session title generation failures. The  aggregation, per-model error tagging, and the removal of double-wrapping in the debug command are all implemented cleanly with no bugs detected.

trungutt (approved):

PR #3053, opened 2026-06-10
feat: add Claude Fable 5 support with refusal handling and server-side fallbacks

dgageot:
Claude Fable 5 ends turns with `stop_reason: "refusal"` when safety classifiers intervene. This change surfaces refusals as a distinct finish reason, allowing the runtime to terminate the turn properly and inform users why the model said nothing.

This PR adds three complementary features to handle Fable 5 gracefully. Model refusals are now plumbed through the Anthropic standard and Beta stream adapters (also mapping `max_tokens` to `length`), with Bedrock guardrail interventions mapped to the new `FinishReasonRefusal`. Server-side fallbacks are supported via a new `fallbacks` provider_opt that forwards a list of model IDs to Anthropic's fallback beta, so requests refused by safety classifiers are retried with backup models in a single round trip. Finally, the 200k context-window fallback used for brand-new models is made model-aware to avoid over-clamping `max_tokens` on Fable 5's larger 1M window.

All changes are backward-compatible and additive. Existing configurations continue to work unchanged.


rumpl:
> Claude Fable 5 ends turns with `stop_reason: "refusal"` when safety classifiers intervene.

How did you find this out you hacker you

docker-agent (commented):

pkg/runtime/loop.go
```diff
@@ -619,6 +619,14 @@ func (r *LocalRuntime) runTurn(
 	endStreamSpan()
 	slog.DebugContext(ctx, "Stream processed", "agent", a.Name(), "tool_calls", len(res.Calls), "content_length", len(res.Content), "stopped", res.Stopped)
 
+	// Surface refusals (e.g. Anthropic safety classifiers): the API returns a
+	// successful, often empty response that would otherwise look like the model
+	// silently said nothing.
+	if res.FinishReason == chat.FinishReasonRefusal {
```
docker-agent, line 625: **[MEDIUM] Refusal response not persisted to session history**

When the model returns `stop_reason: "refusal"`, `recordAssistantMessage` is called immediately after emitting the warning event. However, the existing empty-message guard in `recordAssistantMessage` skips recording when `res.Content == ""` and `len(res.Calls) == 0` — which is exactly the shape of a refusal response.

As a result, the session history contains **no trace of the refusal turn**: the assistant message is silently dropped. On the next turn, the model receives the user's prompt again with no context that it previously refused, potentially leading to repeated refusals or confusing behaviour. The user-facing warning event is emitted correctly, but session history (and any context sent to the model on follow-up turns) misses the refusal entirely.

Consider either (a) bypassing the empty-message guard for refusal turns (e.g., insert a placeholder `"[Model refused to respond]"` message), or (b) adding a special-case in `recordAssistantMessage` that persists the message when the `FinishReason` is `FinishReasonRefusal`.

trungutt (approved):

PR #3057, opened 2026-06-10
fix: prevent fork-mode skill commands from looping in TUI

dgageot:
## Summary

Typing a skill slash command (`/skillname`) in the TUI did nothing — the editor cleared and the skill never ran. The `--session-read-only` change had reordered `chatPage.handleSendMsg` so that `parseImmediateCommand` ran **before** the `BypassQueue` check.

Skill palette items are `Immediate` slash commands whose `Execute` re-emits a `SendMsg` carrying the **same** slash text (with `BypassQueue: true`). With the new ordering, that re-dispatched message was parsed again, re-matched its own palette item, and re-emitted itself — an infinite self-referential dispatch loop that never reached `processMessage` → `SkillCommandFork` / `ResolveInput`. (Built-in and agent commands don't loop because their `Execute` emits dedicated, non-slash message types.)

## Fix

Only skip re-parsing for `BypassQueue` messages, while still letting the read-only guard reject the resolved, model-bound work:

```go
if !msg.BypassQueue {
    if cmd := p.parseImmediateCommand(msg.Content); cmd != nil {
        return p, cmd
    }
}

if p.app != nil && p.app.IsReadOnly() {
    return p, notification.WarningCmd("Session is read-only. No new messages can be sent.")
}

if msg.BypassQueue || isBangCommand(msg.Content) {
    return p, p.processMessage(msg)
}
```

This mirrors the `if !msg.BypassQueue` guard that already exists in `processMessage`, so both dispatch paths now treat already-resolved messages consistently — rather than introducing a separate flag. It also fixes the read-only interaction: skill/agent commands are still correctly blocked in `--session-read-only` sessions instead of slipping past the guard.

> Note: #3058 fixes the same loop in `pkg/tui/commands` via a new `Item.SkipImmediateParse` flag on skill items. Both fixes are correct; this PR keeps the fix in `handleSendMsg` because it aligns with the existing `!msg.BypassQueue` guard in `processMessage` and generalizes to any re-dispatched message without a flag that would be redundant with `Immediate`. The strong end-to-end runtime tests below are adapted from the good ideas in that PR.

## Tests

- `TestQueueFlow_SkillCommand_DispatchesOnceWithoutLooping` — fast unit check that `Execute` runs exactly once across the two-hop dispatch.
- `TestReadOnly_RejectsBypassQueueCommands` — a resolved `BypassQueue` skill/agent command is still rejected in read-only mode (does not start a stream).
- `TestHandleSendMsg_ForkSkillRunsViaFork` / `TestHandleSendMsg_InlineSkillRunsViaResolveInput` — drive the real typed two-hop flow through `handleSendMsg` against a `skillDispatchRuntime` and assert the skill actually executes via the correct path: fork skills reach `RunSkillFork` with the parsed name/task; inline skills reach `ResolveInput`/`RunStream` with the `<skill name="...">` envelope and user request — and neither loops.

Each test was verified to fail against the buggy ordering and pass with the fix.

## Validation

- `go build ./...` ✓
- `go test ./pkg/tui/...` ✓ (incl. the new tests)
- `golangci-lint run ./pkg/tui/page/chat/` ✓ (0 issues)

### DIFF
--- a/pkg/tui/page/chat/chat.go
+++ b/pkg/tui/page/chat/chat.go
@@ -649,11 +649,19 @@ func (p *chatPage) handleSendMsg(msg msgtypes.SendMsg) (layout.Model, tea.Cmd) {
 		return p, core.CmdHandler(msgtypes.ExitSessionMsg{})
 	}
 
-	// Allow immediate slash commands (e.g. /exit, /compact) even in read-only mode
-	if cmd := p.parseImmediateCommand(msg.Content); cmd != nil {
-		return p, cmd
+	// Immediate UI slash commands (e.g. /exit, /compact) run even in read-only
+	// mode. A BypassQueue message has already been resolved (e.g. an agent
+	// command or fork-mode skill re-dispatching itself) and must skip parsing:
+	// re-parsing would match the same command again and loop forever.
+	if !msg.BypassQueue {
+		if cmd := p.parseImmediateCommand(msg.Content); cmd != nil {
+			return p, cmd
+		}
 	}
 
+	// Everything below hands work to the model, which read-only sessions must
+	// reject: normal input, bang commands, and resolved agent/skill commands
+	// flagged BypassQueue.
 	if p.app != nil && p.app.IsReadOnly() {
 		return p, notification.WarningCmd("Session is read-only. No new messages can be sent.")
 	}
--- a/pkg/tui/page/chat/queue_test.go
+++ b/pkg/tui/page/chat/queue_test.go
@@ -17,6 +17,7 @@ import (
 	mcptools "github.com/docker/docker-agent/pkg/tools/mcp"
 	"github.com/docker/docker-agent/pkg/tui/commands"
 	"github.com/docker/docker-agent/pkg/tui/components/sidebar"
+	"github.com/docker/docker-agent/pkg/tui/core"
 	"github.com/docker/docker-agent/pkg/tui/messages"
 	"github.com/docker/docker-agent/pkg/tui/service"
 )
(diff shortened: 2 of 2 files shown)


trungutt (dismissed):

pkg/tui/page/chat/chat.go
```diff
@@ -649,11 +649,19 @@ func (p *chatPage) handleSendMsg(msg msgtypes.SendMsg) (layout.Model, tea.Cmd) {
 		return p, core.CmdHandler(msgtypes.ExitSessionMsg{})
```
docker-agent, line 649: **[LOW] `exit`/`quit`/`:q` keyword check is not guarded by `!msg.BypassQueue`**

The `switch strings.TrimSpace(msg.Content)` block at the top of `handleSendMsg` fires unconditionally for all messages, including those with `BypassQueue: true`. This is inconsistent with the PR's intent: if a skill or agent command ever resolves to the literal string `"exit"`, `"quit"`, or `":q"`, the re-dispatched `BypassQueue` message would trigger `ExitSessionMsg` instead of reaching `processMessage`.

In practice the values are unlikely, but for correctness the guard could mirror the `parseImmediateCommand` check:

```go
if !msg.BypassQueue {
    switch strings.TrimSpace(msg.Content) {
    case "exit", "quit", ":q":
        return p, core.CmdHandler(msgtypes.ExitSessionMsg{})
    }
}
```

pkg/tui/page/chat/queue_test.go
```diff
@@ -144,6 +145,56 @@ func TestQueueFlow_BusyAgent_BangCommandBypassesQueue(t *testing.T) {
...
+	require.NotNil(t, cmd)
+	require.Equal(t, 1, calls)
+
+	redispatch, ok := cmd().(messages.SendMsg)
+	require.True(t, ok, "skill Execute should re-dispatch a SendMsg")
+	require.True(t, redispatch.BypassQueue, "re-dispatch must bypass the queue")
+
+	// Hop 2: the BypassQueue message must be processed directly, not re-parsed
+	// back into the command (which would invoke Execute again and loop).
+	_, cmd = p.handleSendMsg(redispatch)
```
docker-agent, line 264: **[LOW] Hop-2 assertion doesn't verify `processMessage` was actually reached**

After the `BypassQueue` re-dispatch (`hop 2`), the test asserts:
- `calls == 1` (Execute not re-invoked — loop is broken ✓)
- `p.messageQueue` is empty ✓

But it doesn't assert that `p.working` became `true` (or that the returned `cmd` is the output of `processMessage` rather than a no-op). A future regression that silently swallows the hop-2 message (drops it without looping *and* without processing it) would pass this test. Adding `assert.True(t, p.working)` after hop 2 would close that gap.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is correct and well-targeted. Wrapping `parseImmediateCommand` in `if !msg.BypassQueue` precisely breaks the infinite re-dispatch loop while keeping the read-only guard active for all messages (including `BypassQueue`). The two new regression tests cover both the loop scenario and the read-only rejection path, and the comment explaining the invariant is clear and helpful.

No high or medium severity issues found.

aheritier (dismissed):

gtardif (approved):

PR #3058, opened 2026-06-10
fix(tui): resolve skill slash commands instead of looping dispatch

### LINKED ISSUE #3054: Skill slash commands (e.g. /services, /but) do nothing — Immediate palette items re-dispatch the same SendMsg and never resolve
## Summary
Invoking a skill via its slash command (e.g. `/services` from a repo-local `.agents/skills`, or `/but` from `~/.agents/skills`) does nothing. The command **is** discovered — it appears in the command palette and as a `/` slash command — but pressing Enter just clears the editor; the skill body is never resolved or executed.

This is **not** a skill-discovery problem. Both skills load correctly and show in the palette. The defect is in slash-command dispatch.

## Environment
- docker-agent version: `v1.74.0-46-ga9d5ed574` (branch `gitbutler/workspace`)
- OS: macOS (arm64)
- Repro confirmed with skills present in both `~/Workspace/dev/docker/.agents/skills/services` and `~/.agents/skills/gitbutler` (skill name `but`).

## Steps to reproduce
1. Have a valid local skill discoverable (it shows in the palette / as a `/` command).
2. Select the skill command (e.g. `/services` or `/but`) and press Enter.
3. Observe: the editor clears and the prompt returns to empty. The skill never runs; no skill content is injected and no fork sub-session starts.

## Expected
The skill command should resolve: for inline skills, expand `SKILL.md` into the prompt (via `App.ResolveInput` → `ResolveSkillCommand`); for `context: fork` skills, start a sub-session (via `App.SkillCommandFork` → `RunSkillFork`).

## Root cause (self-referential Immediate dispatch loop)
Skill palette items (and agent-command items) are registered as `Immediate` slash commands whose `Execute` re-emits a `SendMsg` carrying the **same** slash text:

- `pkg/tui/commands/commands.go` — in `BuildCommandCategories`, each skill is added with `Immediate: true` and:
  `Execute: func(arg) tea.Cmd { input := "/"+skillName; ...; return core.CmdHandler(messages.SendMsg{Content: input, BypassQueue: true}) }`

But the chat page intercepts every `SendMsg` through the immediate-command parser **first**:

- `pkg/tui/page/chat/chat.go:653` — `handleSendMsg` calls `parseImmediateCommand(msg.Content)` before the `BypassQueue`/`processMessage` branch.
- `pkg/tui/page/chat/chat.go:635-639` — `parseImmediateCommand` delegates to `p.commandParser.Parse(content)`.
- `pkg/tui/tui.go:463` — the chat page's `commandParser` is built from `commands.NewParser(m.commandCategories()...)`, i.e. the **same** categories that include the `Skills` category.

So the flow for `/services` is:
1. `SendMsg{Content:"/services", BypassQueue:true}` reaches `handleSendMsg`.
2. `parseImmediateCommand("/services")` matches the skill's `Immediate` item.
3. Its `Execute` returns `core.CmdHandler(SendMsg{Content:"/services", BypassQueue:true})` — the same message.
4. That message comes back into `handleSendMsg` → step 2 again.

The message therefore **never reaches** `processMessage` (`chat.go:866`), so `SkillCommandFork` / `ResolveInput` (`chat.go:905-910`) are never called and the skill is never resolved. The editor is cleared (the message was "handled"), matching the reported symptom: command visible in palette/slash, Enter does nothing, prompt returns to empty.

Built-in `Immediate` commands (e.g. `/clear`, `/compact`) do **not** loop because their `Execute` emits a *different* message type (`ClearSessionMsg`, `CompactSessionMsg`, …) rather than re-emitting a `SendMsg` with the same slash text.

## Affected code references
- `pkg/tui/commands/commands.go` — Skills category items: `Immediate: true`, `Execute` → `SendMsg{Content: "/"+skillName, BypassQueue: true}` (same pattern for the Agent Commands category, which is likely affected too).
- `pkg/tui/page/chat/chat.go:653` — `handleSendMsg` runs `parseImmediateCommand` on the incoming `SendMsg` content before processing.
- `pkg/tui/page/chat/chat.go:635-639` — `parseImmediateCommand` → `commandParser.Parse`.
- `pkg/tui/page/chat/chat.go:866-911` — `processMessage` → `SkillCommandFork` / `ResolveInput` (the path that actually resolves skills, which is never reached).
- `pkg/tui/tui.go:463` — chat `commandParser` constructed from the skill-inclusive categories.

## Suggested fix direction (for maintainers)
Break the self-referential dispatch. Options:
- Have skill/agent-command `Execute` emit a dedicated message type (e.g. a `RunSkillMsg` / reuse `AgentCommandMsg`) that routes straight to `processMessage`/resolution, instead of re-emitting a `SendMsg` with the same slash text; or
- In `parseImmediateCommand`, do not re-match commands whose only action is to re-send the same slash content (i.e. skip skill/agent-command items there and let `processMessage` resolve them); or
- Guard `handleSendMsg` so a `SendMsg{BypassQueue:true}` originating from an immediate command is not fed back through `parseImmediateCommand`.

Label as a bug if labels are available. Report back the created issue number and URL.

aheritier:
## Summary
Fixes #3054 — skill slash commands (e.g. `/services`, `/but`) appeared in the command palette and as `/` commands but did nothing: pressing Enter just cleared the editor and the skill never resolved or executed.

## Root cause
Skill palette items were registered as `Immediate` slash commands whose `Execute` re-emitted a `SendMsg` carrying the **same** slash text. `chatPage.handleSendMsg` runs the immediate-command parser (`parseImmediateCommand`) on every inbound `SendMsg` *before* reaching `processMessage`, so a skill command re-matched its own palette item and re-sent the identical `SendMsg` — an infinite self-referential dispatch loop. The message therefore never reached `processMessage` → `SkillCommandFork` / `ResolveInput`, so the skill was never resolved.

Built-in commands (`/clear`, `/compact`, …) and agent commands do **not** loop because their `Execute` emits dedicated message types (`ClearSessionMsg`, `AgentCommandMsg`, …) that resolve to non-slash content. Only skills re-sent their own slash text.

## Fix
Add an explicit `Item.SkipImmediateParse` flag, set it on skill items, and have `Parser.Parse` skip flagged items. The skill `SendMsg` then falls through `handleSendMsg` to `processMessage`, which routes `context: fork` skills to `SkillCommandFork`/`RunSkillFork` and inline skills to `ResolveInput`. The palette `Execute` is unchanged (still invoked directly), so this fixes both the palette and typed-slash invocation paths. Change is contained to `pkg/tui/commands/commands.go` (one struct field, one item literal, one parser condition).

## Tests
- `pkg/tui/commands/commands_test.go`: `TestParseSkipsItemsMarkedSkipImmediateParse` (skill items return nil from the parser → no self-resend), `TestParseStillMatchesImmediateItems` (agent commands still match).
- `pkg/tui/page/chat/queue_test.go`: `TestHandleSendMsg_ForkSkillFallsThroughToRunSkillFork` and `TestHandleSendMsg_InlineSkillFallsThroughToResolveInput` (the skill `SendMsg` is not re-emitted and reaches the correct resolution path).

## Validation
- `task build` ✓
- targeted `go test ./pkg/tui/commands/... ./pkg/tui/page/chat/... ./pkg/app/...` ✓ (incl. the 4 new tests)
- `task lint` ✓ (0 issues, go.mod tidy)
- Full `task test` deferred to CI.

Assisted-By: docker-agent

### DIFF
--- a/pkg/tui/commands/commands.go
+++ b/pkg/tui/commands/commands.go
@@ -36,6 +36,11 @@ type Item struct {
 	// Immediate marks commands that should run as soon as they are submitted
 	// instead of being treated as ordinary queued chat input.
 	Immediate bool
+	// SkipImmediateParse excludes this item from Parser.Parse so the slash
+	// command falls through to the chat page's processMessage path. Skill
+	// items use this because their palette Execute re-emits the same SendMsg,
+	// which would otherwise be parsed again instead of resolved by the app.
+	SkipImmediateParse bool
 }
 
 func builtInSessionCommands() []Item {
@@ -557,12 +562,13 @@ func BuildCommandCategories(ctx context.Context, application *app.App) []Categor
 			description := toolcommon.TruncateText(skill.Description, 55)
 
 			skillCommands = append(skillCommands, Item{
-				ID:           "skill." + skillName,
-				Label:        skillName,
-				Description:  description,
-				Category:     "Skills",
-				SlashCommand: "/" + skillName,
-				Immediate:    true,
+				ID:                 "skill." + skillName,
+				Label:              skillName,
+				Description:        description,
+				Category:           "Skills",
+				SlashCommand:       "/" + skillName,
+				Immediate:          true,
+				SkipImmediateParse: true,
 				Execute: func(arg string) tea.Cmd {
 					input := "/" + skillName
 					if arg = strings.TrimSpace(arg); arg != "" {
@@ -620,7 +626,7 @@ func (p *Parser) Parse(input string) tea.Cmd {
 	// Search through all categories and commands
 	for _, category := range p.categories {
 		for _, item := range category.Commands {
-			if item.SlashCommand == cmd && item.Immediate {
+			if item.SlashCommand == cmd && item.Immediate && !item.SkipImmediateParse {
 				return item.Execute(arg)
 			}
 		}
--- a/pkg/tui/commands/commands_test.go
+++ b/pkg/tui/commands/commands_test.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"testing"
 
+	tea "charm.land/bubbletea/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the infinite dispatch loop for skill slash commands by introducing a `SkipImmediateParse` flag. The implementation is clean and the logic is sound across all dispatch paths (palette, direct typing, and queued messages).

**Reviewed:** `pkg/tui/commands/commands.go`, `pkg/tui/commands/commands_test.go`, `pkg/tui/page/chat/queue_test.go`

No high or medium severity issues found. Two minor observations were noted (both functionally harmless):
- Skill items retain `Immediate: true` even though they will never be matched by the parser — mildly misleading but has no runtime impact since `SkipImmediateParse` gates the match.
- No palette-visibility filtering was added (intentional — skills should still appear in the command palette; the fix targets parse-time dispatch only).

The four new tests accurately cover the key scenarios: skill items are skipped by the parser, built-in/agent commands still match, and fork/inline skills fall through to the correct runtime paths (`RunSkillFork`/`ResolveInput`).

dgageot:
This is fixed by https://github.com/docker/docker-agent/pull/3057

PR #3059, opened 2026-06-10
expose embeddable tui components

dgageot:
# refactor(tui): expose embeddable TUI components for downstream embedders

## What

Docker Sandboxes embeds a Gordon assistant inside its own TUI (`sbx` →
ctrl+k). It reuses docker-agent's runtime (teamloader/runtime/session) and
its runtime-agnostic TUI components (message, tool, markdown, editor,
animation), but it had to **duplicate** three pieces that were private or
implicit. This PR exposes them so embedders share the code instead of
porting it:

1. **`pkg/tui/components/toolconfirm`** (new package): the tool-confirmation
   *policy*, extracted from `pkg/tui/dialog/tool_confirmation.go`:
   - `Decision` enum + `Decision.Resume()` mapping to `runtime.ResumeRequest`
   - `BuildPermissionPattern` / `AlwaysAllowLabel` / `OptionsHelp`
   - `KeyMap` / `DefaultKeyMap`, `Title` / `Question` constants
   - `RejectionReasons()` presets (shared with the rejection-reason dialog)

   The permission pattern is the security-sensitive part: the pattern
   *shown* to the user ("always allow ls*") and the pattern *granted* to the
   runtime must come from the same code, in every UI that hosts a
   confirmation. Until now the embedder carried a verbatim copy.

2. **`service.StaticSessionState`**: a `SessionStateReader` with fixed,
   conservative values for rendering message/tool views outside the full
   TUI app. Replaces hand-rolled nine-method stubs that silently rot when
   the interface grows.

3. **`animation.Stopper` / `animation.StopView`**: the "views removed from
   the UI must have StopAnimation called or they leak tick subscriptions"
   contract, previously duplicated as private helpers in the `messages` and
   `reasoningblock` components (and again by embedders).

The last commit refactors `dialog/tool_confirmation.go` and
`dialog/tool_rejection_reason.go` onto the new package.

## Behavior

Strictly behavior-preserving for docker-agent's own TUI:
- `executeAction` emits byte-identical resume requests.
- `AlwaysAllowLabel` is equivalent to the old `alwaysAllowHelpText` for all
  reachable inputs (for non-shell tools the pattern is the tool name, which
  cannot contain `:cmd=`).
- Rejection presets unchanged in content and order; title, question, key
  bindings, and option ordering preserved.

### DIFF
--- a/pkg/tui/components/toolconfirm/toolconfirm.go
+++ b/pkg/tui/components/toolconfirm/toolconfirm.go
@@ -1,0 +1,167 @@
+// Package toolconfirm centralizes the tool-confirmation policy shared by
+// docker-agent's own confirmation dialog and embedders that bring their own
+// dialog framework (e.g. the Gordon assistant embedded in Docker Sandboxes):
+// the decision set and its runtime resume semantics, the "always allow"
+// permission-pattern construction, key bindings, and the user-facing strings.
+//
+// Keeping this policy in one runtime-agnostic place matters most for
+// BuildPermissionPattern: the pattern shown to the user ("always allow ls*")
+// and the pattern granted to the runtime must come from the same code, in
+// every UI that hosts a confirmation.
+package toolconfirm
+
+import (
+	"encoding/json"
+	"strings"
+
+	"charm.land/bubbles/v2/key"
+
+	"github.com/docker/docker-agent/pkg/runtime"
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+// User-facing strings of the confirmation prompt.
+const (
+	Title    = "Tool Confirmation"
+	Question = "Do you want to allow this tool call?"
+)
+
+// Decision is the user's answer to a tool confirmation.
+type Decision int
+
+const (
+	// Approve runs this one call.
+	Approve Decision = iota
+	// ApproveTool runs the call and always allows the tool, scoped by the
+	// permission pattern from BuildPermissionPattern.
+	ApproveTool
+	// ApproveSession runs the call and approves all tools for the rest of
+	// the session.
+	ApproveSession
+	// Reject rejects the call with an optional reason shown to the model.
+	Reject
+)
+
+// Resume translates the decision into the runtime resume request. pattern
+// is the permission pattern for ApproveTool (from BuildPermissionPattern);
+// reason is the optional rejection reason for Reject. Each is ignored by
+// the other decisions.
+func (d Decision) Resume(pattern, reason string) runtime.ResumeRequest {
+	switch d {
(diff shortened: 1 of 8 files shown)


aheritier (approved):

pkg/tui/components/toolconfirm/toolconfirm.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if toolName == "shell" {
+		var args struct {
+			Cmd string `json:"cmd"`
+		}
+		if err := json.Unmarshal([]byte(toolCall.Function.Arguments), &args); err == nil {
+			// First word of the command ("ls -la /tmp" -> "ls"); the
+			// trailing * matches any arguments.
+			if fields := strings.Fields(args.Cmd); len(fields) > 0 {
+				return toolName + ":cmd=" + fields[0] + "*"
```
docker-agent, line 77: **[MEDIUM] `BuildPermissionPattern` produces malformed glob patterns for shell commands starting with `[` or `[[`**

`BuildPermissionPattern` uses `fields[0]` (the first whitespace-split token) directly in a glob pattern without sanitizing glob-special characters:

```go
return toolName + ":cmd=" + fields[0] + "*"
```

A common bash construct like `[[ -f foo ]]` or `[ -z "$VAR" ]` has `[[` or `[` as its first token, producing patterns such as `shell:cmd=[[*` or `shell:cmd=[*`. These are malformed bracket expressions — `filepath.Match` returns `filepath.ErrBadPattern` for them. If the runtime's permission-matching code ignores that error (e.g. `if err == nil && matched`), the "always allow" grant the user just confirmed is silently stored but **never honored**: every subsequent `[[` invocation still prompts for confirmation.

**Impact**: A user clicks "always allow `[[*`", sees a success message, but the permission is effectively dead. The next `[[ -f config ]]` invocation re-prompts — an unexpected and confusing experience.

**Suggested fix**: Escape or reject glob-special characters in the extracted command token before building the pattern, e.g.:

```go
cmd := fields[0]
// Escape '[' and '?' which are special in filepath.Match
cmd = strings.ReplaceAll(cmd, "[", `\[`)
return toolName + ":cmd=" + cmd + "*"
```

or add a test and guard that falls back to the tool name when the first token is a glob meta-character.

docker-agent (commented):

This PR cleanly extracts tool-confirmation policy into `toolconfirm`, exposes `animation.StopView`/`Stopper`, and adds `StaticSessionState` for embedders. The refactoring is behavior-preserving for the existing TUI. One medium-severity issue was found in `BuildPermissionPattern`.

PR #3060, opened 2026-06-10
test: disable git commit signing in test helpers

dgageot:
Tests that bootstrap throwaway git repos and run `git commit` failed on hosts whose global git config enables commit signing (commit.gpgsign=true) without an available GPG secret key. This change disables commit signing in three test helpers, making the test suite deterministic regardless of the host's global git config.

The fix adds `git config commit.gpgsign false` to the test repository initialization in `pkg/snapshot/snapshot_test.go` (bootstrapRepo), `cmd/root/run_worktree_test.go` (createTestWorktree), and `pkg/hooks/builtins/snapshot_test.go` (snapshotBuiltinRepo). This pattern is already used elsewhere in the repo for similar purposes. The full test suite passes with `go test -count=1 ./...`.

### DIFF
--- a/cmd/root/run_worktree_test.go
+++ b/cmd/root/run_worktree_test.go
@@ -87,6 +87,7 @@ func createTestWorktree(t *testing.T) *worktree.Worktree {
 		{"init"},
 		{"config", "user.email", "test@example.com"},
 		{"config", "user.name", "Test User"},
+		{"config", "commit.gpgsign", "false"},
 	} {
 		cmd := exec.CommandContext(t.Context(), "git", append([]string{"-C", dir}, args...)...)
 		out, err := cmd.CombinedOutput()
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes correctly disable commit signing in three test helper functions by adding `git config commit.gpgsign false` after `git init` and before any commits. The fix is well-scoped:

- **Correct ordering**: The config command is placed after repo initialization and before the first commit in all three helpers.
- **Correct scope**: No explicit `--local`/`--global` flag is needed — because the helpers pass `-C <dir>` and the repos are freshly initialized temp dirs, git defaults to the local repo config, avoiding any global config pollution.
- **Consistent error handling**: The new lines follow the same pattern already used for `user.email`/`user.name` config in the same helpers.
- **Pattern reuse**: This matches the approach already used elsewhere in the repo for the same purpose.

No bugs or issues found in the changed code.

gtardif (approved):

PR #3067, opened 2026-06-10
Cycle model thinking level with shift+tab

rumpl:
Pressing Shift+Tab in the TUI now cycles the current model's thinking-effort level through its provider-specific range (e.g. off → minimal → low → medium → high → xhigh for OpenAI), wrapping around. The selected level is applied as a runtime model override for the active session.

The current level is shown inline in the sidebar's agent block as `<provider>/<model> • <thinking>`. To avoid a tool-count flicker, the change refreshes only agent/team info via a new lightweight Runtime.EmitAgentInfo instead of re-running full startup info (which re-discovers tools).

Models that don't support reasoning, and remote runtimes, are no-ops with an informational message.


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27310438861).

dgageot (approved):

PR #3070, opened 2026-06-11
fix: guard elicitation channel close against in-flight sends

### LINKED ISSUE #3069: panic: send on closed channel — elicitation events channel closed outside the bridge lock during stream teardown
## Summary

The runtime can crash with `panic: send on closed channel` when an in-flight elicitation send races the teardown of a stream's events channel. The root cause is a lifecycle defect in `elicitationBridge`: sends are synchronized against channel *swaps*, but the channel *close* happens outside the bridge's lock, so a close can land on a channel that a concurrent sender is still parked on.

This is a general concurrency bug, not specific to any one feature. It is reachable any time two stream lifecycles overlap — nested sub-sessions (background agents, transferred tasks, fork-mode skills) or a second top-level session created with `/fork` while the first is still running.

## Stack trace (abridged)

```
panic: send on closed channel

goroutine 253016 [running]:
runtime.(*elicitationBridge).send                pkg/runtime/elicitation.go:83
runtime.(*LocalRuntime).elicitationHandler       pkg/runtime/elicitation.go:136
userprompt.(*ToolSet).userPrompt                 pkg/tools/builtin/userprompt/userprompt.go:88
... toolexec dispatcher (runToolset → invoke → run → approveAndRun) ...
runtime.(*LocalRuntime).processToolCalls         pkg/runtime/tool_dispatch.go:51
runtime.(*LocalRuntime).runTurn                  pkg/runtime/loop.go:640
runtime.(*LocalRuntime).runStreamLoop            pkg/runtime/loop.go:434
created by runtime.(*LocalRuntime).RunStream in goroutine 253015
                                                 pkg/runtime/loop.go:205
```

The panic fires at `pkg/runtime/elicitation.go:83` — `b.ch <- ev` inside `elicitationBridge.send`, reached from a `userprompt` tool call running inside `processToolCalls` → `runTurn` → `runStreamLoop`.

## How elicitation channel ownership works

`elicitationBridge` owns the events channel that the MCP elicitation handler sends to. Each `RunStream` swaps in its own channel on entry and swaps the previous one back on exit, so nested streams don't lose the parent's pipe:

```go
// loop.go — runStreamLoop entry
prevElicitationCh := r.elicitation.swap(events)
```

```go
// elicitation.go — send and swap are serialized by an RWMutex
func (b *elicitationBridge) send(ev Event) error {
    b.mu.RLock()
    defer b.mu.RUnlock()
    if b.ch == nil {
        return errNoElicitationChannel
    }
    b.ch <- ev          // <-- panics if this channel was closed elsewhere
    return nil
}

func (b *elicitationBridge) swap(ch chan Event) chan Event {
    b.mu.Lock()
    defer b.mu.Unlock()
    prev := b.ch
    b.ch = ch
    return prev
}
```

The bridge's doc comment claims the RWMutex preserves the invariant that "the channel reference held by an in-flight sender stays open until that sender finishes." **That guarantee only holds against `swap`.**

## Root cause

The channel is closed in `finalizeEventChannel` (`pkg/runtime/loop.go`), which runs as the deferred teardown of every `runStreamLoop`:

```go
func (r *LocalRuntime) finalizeEventChannel(ctx, sess, reason, prevElicitationCh, events chan Event) {
    r.elicitation.swap(prevElicitationCh) // swap the ref out of the bridge
    defer close(events)                   // <-- closed OUTSIDE the bridge lock
    ...
    events <- StreamStopped(...)
}
```

`swap(prevElicitationCh)` removes `events` from the bridge, then `close(events)` runs with no coordination with `send`'s read-lock. So the sequence that panics is:

1. Stream A's `elicitationHandler` calls `send`, takes `RLock`, sees a valid `b.ch`, and parks on `b.ch <- ev` (buffer full or unbuffered).
2. Stream A (or the parent) reaches `finalizeEventChannel`, calls `swap(prev)` — which **does not block on the RLock** because the sender already passed the lock and is now blocked on the channel op, not the mutex — and then `close(events)`.
3. The parked send in step 1 resumes on a now-closed channel → panic.

The RWMutex serializes `send`↔`swap` but never gates `close`, so the "stays open until the sender finishes" invariant is not actually enforced.

## Triggers (in order of likelihood)

- Nested sub-sessions: background agents, `transfer_task`, fork-mode skills — each pushes/pops an elicitation channel, so teardown of an inner stream races a send on the bridge.
- `/fork` while a session is running: spawns a second top-level `RunStream` loop concurrently with the first, so two independent teardowns can interleave with an in-flight elicitation. This is the path that surfaced the crash above, but it is one trigger among several, not the cause.

## Existing test gap

`TestElicitationBridge_SwapWaitsForInflightSenders` proves `swap` waits for in-flight senders, but there is no coverage of `close` vs an in-flight send — exactly the unguarded path.

## Proposed fix

1. Move the channel close behind the bridge lock so it is mutually exclusive with `send` (a bridge-owned `restoreAndClose`/`closeChannel` that restores the previous channel, emits `StreamStopped`, and closes under the write lock).
2. Add a defensive `recover` in `send` that converts a send-on-closed-channel panic into `errNoElicitationChannel`, so any future regression degrades to a declined elicitation instead of crashing the process.
3. Add a `-race` regression test covering close-vs-in-flight-send.

## Environment

- Observed on macOS (arm64).
- Reproduced against the current `main` runtime.

aheritier:
## What

Fixes a `panic: send on closed channel` in the runtime's `elicitationBridge`. The bridge's `RWMutex` serializes elicitation `send` against channel `swap`, but the stream's events channel is closed in `finalizeEventChannel` (`pkg/runtime/loop.go`) via `defer close(events)` — **outside** the bridge lock. An in-flight elicitation `send` (e.g. from a `userprompt` tool call) can therefore race the channel close during stream teardown and crash the process.

The invariant the bridge's doc comment claims — "the channel reference held by an in-flight sender stays open until that sender finishes" — only held against `swap`, never against `close`.

This is a general lifecycle bug, not tied to one feature. It is reachable whenever two stream lifecycles overlap: nested sub-sessions (background agents, `transfer_task`, fork-mode skills) or a second top-level session created with `/fork` while the first is still running.

See #3069 for the full root-cause analysis, stack trace, and race timeline.

## Changes

- **`pkg/runtime/elicitation.go`**
  - Add a bridge-owned `restoreAndClose` that, under the bridge **write lock**, restores the previous elicitation channel, emits the final `StreamStopped` event, and closes the stream events channel — so close is now mutually exclusive with an in-flight `send`, exactly like `swap`.
  - Add a defensive `recover` in `send` that converts a send-on-closed-channel panic into `errNoElicitationChannel`, downgrading any future regression to a declined elicitation instead of a process crash.
  - Update the bridge concurrency doc comment to state that close is coordinated too.
- **`pkg/runtime/loop.go`**
  - `finalizeEventChannel` now closes the events channel through `restoreAndClose`, preserving the existing ordering (restore parent channel → `StreamStopped` → session-end hooks → close).
- **`pkg/runtime/elicitation_test.go`**
  - Add `-race` regression coverage for close-vs-in-flight-send (the existing `TestElicitationBridge_SwapWaitsForInflightSenders` only covered swap-vs-send).
  - Add coverage for the defensive closed-channel recovery path.

## Validation

| Command | Result |
|---|---|
| `task build` | ✅ pass |
| `go test ./pkg/runtime/...` | ✅ pass |
| `go test -race ./pkg/runtime/...` | ✅ pass |
| `task lint` | ✅ pass |

`task test` has one **unrelated** failure: `pkg/teamloader` example tests (`examples/dmr.yaml`, `examples/unload_on_switch.yaml`) attempt to auto-pull the Docker Model Runner model `ai/qwen3`, and the registry returned HTTP `416 Requested Range Not Satisfiable`. This is an environment/network issue, not related to this change; `pkg/runtime` passes within that same run.

Fixes #3069

### DIFF
--- a/pkg/runtime/elicitation.go
+++ b/pkg/runtime/elicitation.go
@@ -47,12 +47,13 @@ var errNoElicitationChannel = errors.New("no events channel available for elicit
 //
 // The bridge encapsulates a non-trivial concurrency contract: while a
 // caller holds a reference to the current channel and is in the middle
-// of sending an elicitation request, the swap-back must not race with
+// of sending an elicitation request, stream teardown must not race with
 // close(channel) on the inner stream. We achieve this by serializing
-// send and swap with an RWMutex held across the send. Pushing this into
-// a small standalone type keeps the contract testable in isolation
-// (with the race detector) without spinning up a runtime, and keeps
-// LocalRuntime free of the two raw fields it used to expose.
+// send, swap, and close with an RWMutex held across the channel
+// operation. Pushing this into a small standalone type keeps the
+// contract testable in isolation (with the race detector) without
+// spinning up a runtime, and keeps LocalRuntime free of the two raw
+// fields it used to expose.
 type elicitationBridge struct {
 	mu sync.RWMutex
 	ch chan Event
@@ -69,19 +70,36 @@ func (b *elicitationBridge) swap(ch chan Event) chan Event {
 }
 
 // send delivers ev to the current channel, holding the read lock across
-// the send. This blocks any concurrent swap until the send completes,
+// the send. This blocks concurrent teardown until the send completes,
 // preserving the invariant that the channel reference held by an
 // in-flight sender stays open until that sender finishes.
 //
-// Returns errNoElicitationChannel when no channel is configured.
-func (b *elicitationBridge) send(ev Event) error {
+// Returns errNoElicitationChannel when no channel is configured or when
+// a defensive recover catches an externally closed channel.
+func (b *elicitationBridge) send(ev Event) (err error) {
 	b.mu.RLock()
 	defer b.mu.RUnlock()
+	defer func() {
+		if recover() != nil {
+			err = errNoElicitationChannel
+		}
+	}()
 	if b.ch == nil {
 		return errNoElicitationChannel
 	}
 	b.ch <- ev
 	return nil
+}
+
+// restoreAndClose restores the previous stream channel, emits the final
+// event while no elicitation sender can target the closing channel, then
+// closes the current channel under the bridge write lock.
+func (b *elicitationBridge) restoreAndClose(current, previous chan Event, final Event) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.ch = previous
+	current <- final
+	close(current)
 }
 
 // ResumeElicitation sends an elicitation response back to a waiting
--- a/pkg/runtime/elicitation.go
+++ b/pkg/runtime/elicitation.go
@@ -91,14 +91,12 @@ func (b *elicitationBridge) send(ev Event) (err error) {
 	return nil
 }
 
-// restoreAndClose restores the previous stream channel, emits the final
-// event while no elicitation sender can target the closing channel, then
-// closes the current channel under the bridge write lock.
-func (b *elicitationBridge) restoreAndClose(current, previous chan Event, final Event) {
+// restoreAndClose restores the previous stream channel and closes the current
+// stream channel under the bridge write lock.
+func (b *elicitationBridge) restoreAndClose(current, previous chan Event) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	b.ch = previous
-	current <- final
 	close(current)
 }
 
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -162,7 +162,7 @@ func (r *LocalRuntime) emitHookDrivenShutdown(
 
 // finalizeEventChannel performs cleanup at the end of a RunStream goroutine:
 // restores the previous elicitation channel, emits the StreamStopped event,
-// fires hooks, and closes the events channel.
+// closes the events channel, and fires hooks.
 //
 // reason is one of the turnEndReason* constants and classifies how the
 // stream ended (e.g. "normal", "error", "canceled"). It is surfaced in
@@ -170,13 +170,6 @@ func (r *LocalRuntime) emitHookDrivenShutdown(
 // distinguish between successful completion, crashes, and user-initiated
 // stops without reverse-engineering reconnect failures.
 func (r *LocalRuntime) finalizeEventChannel(ctx context.Context, sess *session.Session, reason string, prevElicitationCh, events chan Event) {
-	// Swap back the parent's elicitation channel before closing this
-	// stream's channel. This prevents a send-on-closed-channel panic
-	// and restores elicitation for the parent session.
-	r.elicitation.swap(prevElicitationCh)
-
-	defer close(events)
-
 	a := r.resolveSessionAgent(sess)
 
 	// Execute session end hooks with a context that won't be cancelled so
@@ -186,7 +179,8 @@ func (r *LocalRuntime) finalizeEventChannel(ctx context.Context, sess *session.S
 	if ctx.Err() != nil && reason == "" {
 		reason = turnEndReasonCanceled
 	}
-	events <- StreamStopped(sess.ID, a.Name(), reason)
+
+	r.elicitation.restoreAndClose(events, prevElicitationCh, StreamStopped(sess.ID, a.Name(), reason))
 
 	r.executeOnUserInputHooks(ctx, sess.ID, "stream stopped")
 
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -161,8 +161,8 @@ func (r *LocalRuntime) emitHookDrivenShutdown(
 }
 
 // finalizeEventChannel performs cleanup at the end of a RunStream goroutine:
-// restores the previous elicitation channel, emits the StreamStopped event,
-// closes the events channel, and fires hooks.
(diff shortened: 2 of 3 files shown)


pkg/runtime/elicitation.go
```diff
@@ -69,21 +70,38 @@ func (b *elicitationBridge) swap(ch chan Event) chan Event {
...
 }
 
+// restoreAndClose restores the previous stream channel, emits the final
+// event while no elicitation sender can target the closing channel, then
+// closes the current channel under the bridge write lock.
+func (b *elicitationBridge) restoreAndClose(current, previous chan Event, final Event) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.ch = previous
+	current <- final
```
docker-agent: **[MEDIUM] Blocking channel send inside write lock — potential deadlock**

`restoreAndClose` acquires `b.mu.Lock()` (write lock) and then performs a blocking send `current <- final` while still holding that lock:

```go
func (b *elicitationBridge) restoreAndClose(current, previous chan Event, final Event) {
    b.mu.Lock()
    defer b.mu.Unlock()
    b.ch = previous
    current <- final  // blocks here while holding the write lock
    close(current)
}
```

`current` is the `events` channel buffered at `defaultEventChannelCapacity` (128). If that buffer is full when this send executes — possible under high-event throughput with a slow or cancelled consumer — the send will block indefinitely while the write lock is held. This prevents any concurrent `b.send()` call from acquiring the read lock, creating a secondary deadlock between the goroutine running teardown and any goroutine still trying to send elicitation events.

Consider using a non-blocking send with a fallback (`select { case current <- final: default: }`) or draining/closing the channel without sending the final event inside the lock.

pkg/runtime/loop.go
```diff
@@ -186,7 +179,8 @@ func (r *LocalRuntime) finalizeEventChannel(ctx context.Context, sess *session.S
 	if ctx.Err() != nil && reason == "" {
 		reason = turnEndReasonCanceled
 	}
-	events <- StreamStopped(sess.ID, a.Name(), reason)
+
+	r.elicitation.restoreAndClose(events, prevElicitationCh, StreamStopped(sess.ID, a.Name(), reason))
 
 	r.executeOnUserInputHooks(ctx, sess.ID, "stream stopped")
```
docker-agent, line 185: **[MEDIUM] Hook and telemetry calls now execute after the events channel is closed — ordering change**

Before this PR, `executeOnUserInputHooks` and `RecordSessionEnd` ran **before** `events` was closed (it was closed via `defer`). Now they run **after** `restoreAndClose` closes the channel:

```go
// New ordering:
r.elicitation.restoreAndClose(events, prevElicitationCh, StreamStopped(...))  // closes events here
r.executeOnUserInputHooks(ctx, sess.ID, "stream stopped")  // runs after close
r.telemetry.RecordSessionEnd(ctx)                          // runs after close
```

While the bridge correctly redirects future `b.send()` calls to `previous` (the parent channel) after `restoreAndClose` sets `b.ch = previous`, any hook that sends events through a different path (direct channel write, not through the bridge) targeting the now-closed `events` would panic. Beyond the panic risk, this is a semantic ordering change: consumers of the event channel will observe the channel close **before** these hooks fire, whereas previously hooks fired first. If any hook is intended to emit a final event that consumers should see before the stream is considered done, it will now silently drop that event.

docker-agent (commented):

This PR correctly identifies and partially fixes the send-on-closed-channel race by serializing close under the bridge write lock. Two concerns about the new `restoreAndClose` implementation are noted below.

pkg/runtime/elicitation.go
aheritier: Addressed: removed the lock-held send, switched StreamStopped to a non-blocking send, and added a full-buffer regression test.

pkg/runtime/loop.go
aheritier: Addressed: reordered the shutdown path so close happens last, after the hooks and telemetry.

docker-agent: Thanks for the fix! The reordering looks correct — hooks and telemetry now complete before the channel closes, preserving the original semantic guarantee that consumers observe hook-generated events before the stream is considered finished.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is well-structured and correctly addresses the  race condition in the elicitation bridge.

**What the PR does right:**
-  acquires the bridge **write lock** before restoring the previous channel and closing the current one — making close mutually exclusive with any in-flight  (which holds the read lock). The fix is minimal and surgical.
- The defensive  in  is correctly scoped: it only fires on panics while the read lock is held, wraps the result as a named return, and converts the panic to  rather than masking arbitrary panics at a higher level.
-  now calls  as the very last action, after hooks, , and telemetry have all completed — preserving the shutdown ordering invariant documented in project memory (hooks/telemetry must complete before resources are closed).
- The new tests exercise the exact race (close vs in-flight send) that the old  missed, and the  test covers the defensive recovery path directly.
- The non-blocking  emit replaces a direct channel send, correctly handling a full buffer without deadlocking (as the new  verifies).

dgageot (approved):

PR #3072, opened 2026-06-11
fix: guard type assertions in ScriptToolSet.Instructions() against missing description

### LINKED ISSUE #3071: panic in ScriptToolSet.Instructions when a script shell tool arg has no description
### Summary

`cagent` panics at session start with `interface conversion: interface {} is nil, not string` when a `script` toolset defines a shell tool whose arg omits `description`.

The crash is in `ScriptToolSet.Instructions()` (`pkg/tools/builtin/shell/script_shell.go:104`, v1.74.0):

```go
description := argDef.(map[string]any)["description"].(string)
```

If the arg definition has no `description` key, the map lookup returns `nil` and the type assertion panics. The outer `argDef.(map[string]any)` assertion can also panic if the arg definition is not a mapping (e.g. `name: string` shorthand or a null value).

### Reproduction

```yaml
agents:
  root:
    model: anthropic/claude-sonnet-4-6
    toolsets:
      - type: script
        shell:
          greet:
            cmd: echo hello ${name}
            args:
              name:
                type: string   # no description -> panic
            required: [name]
```

Start a session with any prompt. The runtime panics before the first model call, while building the system messages:

```
panic: interface conversion: interface {} is nil, not string

github.com/docker/docker-agent/pkg/tools/builtin/shell.(*ScriptToolSet).Instructions
        pkg/tools/builtin/shell/script_shell.go:104
github.com/docker/docker-agent/pkg/tools.GetInstructions
        pkg/tools/capabilities.go:72
github.com/docker/docker-agent/pkg/session.buildInvariantSystemMessages
        pkg/session/session.go:918
github.com/docker/docker-agent/pkg/session.(*Session).GetMessages
        pkg/session/session.go:1053
github.com/docker/docker-agent/pkg/runtime.(*LocalRuntime).runStreamLoop
        pkg/runtime/loop.go:261
```

### Why this is surprising

- The config loads and validates fine; `NewScript`/`validateConfig` accept it. The failure only surfaces mid-session, deep in message building, as a hard crash of the whole process.
- The rest of the toolset already treats `description` as optional: `Tools()` falls back with `cmp.Or(cfg.Description, ...)` and `defaultPropertyTypes()` uses a comma-ok assertion. `Instructions()` is the only code path that hard-requires it.

### Suggested fix

`description` should be either optional or rejected at config load time — anything but a runtime panic:

1. **Make it optional in `Instructions()`** (preferred, consistent with `Tools()`): use comma-ok assertions and fall back to an empty/`Runs: cmd` description when missing; skip or stringify non-map arg defs.
2. **Or validate at config load**: extend `validateConfig()` (or schema validation in `pkg/config`) to require a string `description` per arg and return a clear error like `tool 'greet' arg 'name' is missing a description`, so the config is refused up front instead of crashing the session.

Happy to provide more detail if useful.

simonferquel-clanker:
Fixes #3071.

`Instructions()` panicked with `interface conversion: interface {} is nil, not string` when a `script` toolset defined a shell tool whose arg omitted `description`.

**Root cause:** two chained unguarded type assertions:
```go
description := argDef.(map[string]any)["description"].(string)
```

**Fix:** replace both with comma-ok guards; fall back to empty string when `description` is absent or `argDef` is not a map. When description is empty the colon-space separator is also omitted to avoid cosmetic artefacts.

**Tests added:**
- `TestScriptToolSet_InstructionsNoDescription` — map argDef, missing description key
- `TestScriptToolSet_InstructionsNilArgDef` — nil argDef
- `TestScriptToolSet_InstructionsNonMapArgDef` — non-map argDef (bare string)

### DIFF
--- a/pkg/tools/builtin/shell/script_shell.go
+++ b/pkg/tools/builtin/shell/script_shell.go
@@ -101,12 +101,21 @@ func (t *ScriptToolSet) Instructions() string {
 		}
 
 		for argName, argDef := range tool.Args {
-			description := argDef.(map[string]any)["description"].(string)
+			description := ""
+			if m, ok := argDef.(map[string]any); ok {
+				if d, ok := m["description"].(string); ok {
+					description = d
+				}
+			}
 			required := ""
 			if slices.Contains(tool.Required, argName) {
 				required = " (required)"
 			}
-			fmt.Fprintf(&sb, "- `%s`: %s%s\n", argName, description, required)
+			if description != "" {
+				fmt.Fprintf(&sb, "- `%s`: %s%s\n", argName, description, required)
+			} else {
+				fmt.Fprintf(&sb, "- `%s`%s\n", argName, required)
+			}
 		}
 		sb.WriteString("\n")
 	}
--- a/pkg/tools/builtin/shell/script_shell_test.go
+++ b/pkg/tools/builtin/shell/script_shell_test.go
@@ -223,6 +223,66 @@ func TestScriptShellTool_RejectsNULInValue(t *testing.T) {
 	assert.Contains(t, result.Output, "NUL byte")
 }
 
+func TestScriptToolSet_InstructionsNoDescription(t *testing.T) {
+	shellTools := map[string]latest.ScriptShellToolConfig{
+		"greet": {
+			Cmd: "echo hello ${name}",
+			Args: map[string]any{
+				"name": map[string]any{
+					"type": "string",
+					// no "description" key — this must not panic
+				},
+			},
+			Required: []string{"name"},
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly guards both chained type assertions in `ScriptToolSet.Instructions()` with comma-ok patterns, falling back gracefully to an empty string when `argDef` is not a map, is nil, or when the `"description"` key is absent. The separator logic (omitting `": "` when description is empty) avoids cosmetic artefacts. All three new test cases accurately exercise the fixed code paths.

dgageot (approved):

PR #3075, opened 2026-06-11
Allow pulling agent from localhost http URL for local dev

gtardif:


pkg/config/sources.go
```diff
@@ -375,7 +375,7 @@ func validateAgentURL(rawURL string) error {
 	if err != nil {
 		return fmt.Errorf("invalid URL %q: %w", rawURL, err)
 	}
-	if u.Scheme != "https" {
+	if u.Scheme != "https" && (u.Hostname() != "localhost") {
```
docker-agent: **[HIGH] Scheme validation bypassed for *any* scheme on `localhost`, not just `http://`**

The condition `u.Scheme != "https" && (u.Hostname() != "localhost")` evaluates to `false` (skips the error) whenever the hostname is `localhost`, regardless of scheme. This means `ftp://localhost`, `file://localhost`, `javascript://localhost`, and any other scheme all pass validation.

The intent appears to be allowing only `http://localhost`, but the correct expression should be:
```go
if u.Scheme != "https" && !(u.Scheme == "http" && u.Hostname() == "localhost") {
```
or equivalently:
```go
isLocalhostHTTP := u.Scheme == "http" && u.Hostname() == "localhost"
if u.Scheme != "https" && !isLocalhostHTTP {
```

docker-agent: **[HIGH] `http://localhost` exemption is non-functional — blocked by `NewSSRFSafeTransport` at dial time**

Even though `validateAgentURL` now accepts `http://localhost`, the `Read()` method still wraps the HTTP client with `httpclient.NewSSRFSafeTransport()` (when `!a.unsafe`), which calls `SSRFDialControl` **after DNS resolution** and rejects loopback IPs via `ip.IsLoopback()` — covering both `127.0.0.1` and `::1`.

This means the PR's stated goal ("Allow pulling agent from localhost http URL for local dev") is silently non-functional on the `!a.unsafe` production path: `validateAgentURL` passes the URL, but the transport blocks the actual TCP connection. The user will get a cryptic SSRF dial error instead.

To make this work end-to-end, the SSRF transport also needs to be bypassed (or replaced) for `localhost` URLs, similar to how the `unsafe` flag works in tests.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

2 confirmed high-severity bugs in the single changed line. The localhost HTTP exemption has a logic error that allows any scheme (not just `http://`) for localhost, and the exemption is also non-functional end-to-end because `NewSSRFSafeTransport` still blocks loopback IPs at dial time.

pkg/config/sources.go
```diff
@@ -255,10 +255,17 @@ func (a urlSource) Read(ctx context.Context) ([]byte, error) {
...
 	client := httpclient.NewHTTPClient(ctx)
 	if !a.unsafe {
-		client = &http.Client{
-			Timeout:       60 * time.Second,
-			Transport:     httpclient.NewSSRFSafeTransport(),
-			CheckRedirect: httpclient.HTTPSOnlyRedirects(10),
+		if isLocalhostHTTP(a.url) {
+			client = &http.Client{
+				Timeout:       60 * time.Second,
+				CheckRedirect: httpclient.BoundedRedirects(10),
```
docker-agent: **[HIGH] SSRF bypass: `BoundedRedirects` allows following redirects to arbitrary non-localhost URLs**

The localhost HTTP branch creates an `http.Client` with only `BoundedRedirects(10)` and **no custom `Transport`** (Go's default transport is used). `BoundedRedirects` only caps the hop count — it does not restrict the redirect *target*. If a service running on `localhost` returns a `302 Location: http://169.254.169.254/latest/meta-data/` (or any RFC1918 / internal address), Go's HTTP client will follow it and reach that service, completely bypassing the SSRF protection.

By contrast, the non-localhost branch uses both `NewSSRFSafeTransport()` (blocks at dial time) **and** `HTTPSOnlyRedirects()` (blocks non-HTTPS redirect targets). The localhost branch has neither safeguard for redirect targets.

The comment in `validateAgentURL` acknowledges why the SSRF transport is skipped for the *initial* connection ("loopback is the whole point"), but that reasoning does not extend to redirect targets, which may not be loopback addresses.

**Suggested fix**: use a `LocalhostOnlyRedirects` policy that either rejects all redirects or restricts the redirect target to `localhost` as well:

```go
// Option A – reject all redirects from localhost (simplest, most secure)
CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
    return fmt.Errorf("redirects are not allowed for http://localhost sources")
},

// Option B – restrict redirect target to localhost only
CheckRedirect: func(req *http.Request, via []*http.Request) error {
    if len(via) >= 10 {
        return fmt.Errorf("stopped after 10 redirects")
    }
    if !isLocalhostHTTP(req.URL.String()) {
        return fmt.Errorf("refusing redirect to non-localhost URL %q", req.URL.Redacted())
    }
    return nil
},
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One confirmed high-severity SSRF bypass was found in the new localhost HTTP code path.

dgageot (approved):

PR #3077, opened 2026-06-11
Add Docker Desktop JWT when pulling agent from a .docker.com URL

gtardif:
Also, don't send desktop JWT to model gateway if it's not a .docker.com URL, or localhost (for dev purposes)

Signed-off-by: Guillaume Tardif <guillaume.tardif@gmail.com>

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -69,7 +69,7 @@ func Load(ctx context.Context, source Source) (*latest.Config, error) {
 //
 // This allows exiting early with a proper error message instead of failing later when trying to use a model or tool.
 func CheckRequiredEnvVars(ctx context.Context, cfg *latest.Config, modelsGateway string, env environment.Provider) error {
-	if modelsGateway != "" && environment.IsDockerURL(modelsGateway) {
+	if modelsGateway != "" && environment.IsTrustedDockerURL(modelsGateway) {
 		if jwt, _ := env.Get(ctx, environment.DockerDesktopTokenEnv); jwt == "" {
 			return errors.New("sorry, you first need to sign in Docker Desktop to use the Docker AI Gateway")
 		}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -69,7 +69,7 @@ func Load(ctx context.Context, source Source) (*latest.Config, error) {
 //
 // This allows exiting early with a proper error message instead of failing later when trying to use a model or tool.
 func CheckRequiredEnvVars(ctx context.Context, cfg *latest.Config, modelsGateway string, env environment.Provider) error {
-	if modelsGateway != "" {
+	if modelsGateway != "" && environment.IsDockerURL(modelsGateway) {
 		if jwt, _ := env.Get(ctx, environment.DockerDesktopTokenEnv); jwt == "" {
 			return errors.New("sorry, you first need to sign in Docker Desktop to use the Docker AI Gateway")
 		}
--- a/pkg/config/sources.go
+++ b/pkg/config/sources.go
@@ -340,8 +340,8 @@ func isGitHubURL(urlStr string) bool {
 	return slices.Contains(githubHosts, u.Host)
 }
 
-// isDockerURL is an alias for [environment.IsDockerURL] for local readability.
-var isDockerURL = environment.IsDockerURL
+// isTrustedDockerURL is an alias for [environment.IsTrustedDockerURL] for local readability.
+var isTrustedDockerURL = environment.IsTrustedDockerURL
 
 // addGitHubAuth adds GitHub token authorization to the request if:
 // - The URL is a GitHub URL
@@ -374,7 +374,7 @@ func (a urlSource) addDockerAuth(ctx context.Context, req *http.Request) {
 		return
 	}
 
-	if !isDockerURL(a.url) {
+	if !isTrustedDockerURL(a.url) {
 		return
 	}
 
--- a/pkg/config/sources.go
+++ b/pkg/config/sources.go
@@ -340,18 +340,8 @@ func isGitHubURL(urlStr string) bool {
 	return slices.Contains(githubHosts, u.Host)
 }
 
-// isDockerURL checks if the URL targets a .docker.com domain that should
-// receive the Docker Desktop JWT. The check matches the exact host
-// "docker.com" as well as any subdomain (e.g. "desktop.docker.com").
-// It performs strict hostname validation to prevent token leakage.
-func isDockerURL(urlStr string) bool {
-	u, err := url.Parse(urlStr)
-	if err != nil {
-		return false
-	}
-	host := u.Hostname()
-	return host == "docker.com" || strings.HasSuffix(host, ".docker.com")
-}
+// isDockerURL is an alias for [environment.IsDockerURL] for local readability.
+var isDockerURL = environment.IsDockerURL
 
 // addGitHubAuth adds GitHub token authorization to the request if:
 // - The URL is a GitHub URL
--- a/pkg/config/sources.go
+++ b/pkg/config/sources.go
@@ -250,8 +250,8 @@ func (a urlSource) Read(ctx context.Context) ([]byte, error) {
 		req.Header.Set("If-None-Match", cachedETag)
 	}
 
-	// Add GitHub token authorization for GitHub URLs
 	a.addGitHubAuth(ctx, req)
+	a.addDockerAuth(ctx, req)
 
 	client := httpclient.NewHTTPClient(ctx)
 	if !a.unsafe {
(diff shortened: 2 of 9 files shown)


dgageot (dismissed):

pkg/environment/docker-desktop.go
```diff
@@ -12,6 +14,22 @@ const (
 	DockerDesktopTokenEnv = "DOCKER_TOKEN"
 )
 
+// IsTrustedDockerURL checks if the URL targets a domain trusted to receive
+// the Docker Desktop JWT. It matches:
+//   - "docker.com" and any subdomain (e.g. "desktop.docker.com")
+//   - localhost addresses ("localhost", "127.0.0.1", "::1") for local development
+//
+// It performs strict hostname validation to prevent token leakage.
+func IsTrustedDockerURL(rawURL string) bool {
```
docker-agent, line 23: **[MEDIUM] JWT sent to any local process binding any port on localhost**

`IsTrustedDockerURL` returns `true` for `localhost`, `127.0.0.1`, and `::1` on *any* port. This means any process on the machine that binds to a loopback port (e.g., a dev dependency, local tool, or malicious package) and happens to receive a request to that URL will also receive the Docker Desktop JWT in the `Authorization` header.

The PR description confirms this is intentional for local development, and the function's doc comment documents it. The risk trade-off is accepted — but consider narrowing the exposed surface by restricting to specific ports (e.g., the Docker Desktop API port) or adding a note that this should only be used for dev builds, not in production agent config.

pkg/model/provider/anthropic/client.go
```diff
@@ -120,12 +127,14 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
...
 
-			client := anthropic.NewClient(
-				option.WithAuthToken(authToken),
-				option.WithAPIKey(authToken),
+			clientOptions := []option.RequestOption{
 				option.WithBaseURL(baseURL),
 				option.WithHTTPClient(httpclient.NewHTTPClient(ctx, httpOptions...)),
-			)
+			}
+			if authToken != "" {
```
docker-agent, line 134: **[MEDIUM] Anthropic SDK may silently send `ANTHROPIC_API_KEY` to a custom (non-Docker) gateway**

When `IsTrustedDockerURL(gateway)` is `false`, `authToken` stays empty and neither `option.WithAuthToken` nor `option.WithAPIKey` is appended to `clientOptions`. The Anthropic SDK's default behaviour then kicks in: it reads `ANTHROPIC_API_KEY` from the environment and sends it with every request.

If a user configures a self-hosted or third-party gateway URL (not `.docker.com`) and also has `ANTHROPIC_API_KEY` set in their environment, their API key will silently be forwarded to that gateway. Consider explicitly passing `option.WithAPIKey("")` (or `option.WithAuthToken("")`) to suppress the SDK's env-var fallback when the gateway is non-trusted:

```go
if authToken != "" {
    clientOptions = append(clientOptions, option.WithAuthToken(authToken), option.WithAPIKey(authToken))
} else {
    // Suppress SDK's automatic ANTHROPIC_API_KEY env-var lookup for non-Docker gateways
    clientOptions = append(clientOptions, option.WithAPIKey(""))
}
```

The same issue exists in `pkg/model/provider/openai/client.go:157` and `pkg/model/provider/gemini/client.go:165`.

pkg/environment/docker-desktop.go
docker-agent, line 23: **[MEDIUM] `IsTrustedDockerURL` does not enforce HTTPS — JWT could be sent over plain HTTP to a `.docker.com` gateway**

`IsTrustedDockerURL` checks the hostname but not the URL scheme. If a user configures a model gateway as `http://gateway.docker.com` (HTTP, not HTTPS), the function returns `true` and the JWT is injected into a plaintext HTTP request.

For the agent-pull path in `sources.go`, `validateAgentURL` provides an upstream guard that rejects non-HTTPS URLs (except `http://localhost`). But for the model provider gateway paths (anthropic, openai, gemini), there is no equivalent scheme enforcement before `IsTrustedDockerURL` is called.

Consider adding a scheme check:

```go
func IsTrustedDockerURL(rawURL string) bool {
    u, err := url.Parse(rawURL)
    if err != nil {
        return false
    }
    // Only trust HTTPS URLs (or http://localhost for local dev)
    if u.Scheme != "https" && u.Scheme != "http" {
        return false
    }
    host := u.Hostname()
    if host == "localhost" || host == "127.0.0.1" || host == "::1" {
        return true // local dev: allow http
    }
    // Require HTTPS for .docker.com domains
    if u.Scheme != "https" {
        return false
    }
    return host == "docker.com" || strings.HasSuffix(host, ".docker.com")
}
```

docker-agent (commented):

This PR correctly implements JWT gating — the Docker Desktop token is now only sent to `.docker.com` and localhost endpoints, rather than all gateways. The core logic is sound. Three medium-severity observations are noted inline.

pkg/environment/docker-desktop.go
docker-agent, line 23: **[MEDIUM] Docker Desktop JWT is sent to all localhost addresses, not just docker.com**

`IsTrustedDockerURL` intentionally includes `localhost`, `127.0.0.1`, and `::1` for local development. This means the Docker Desktop JWT will be attached to *any* service listening on loopback — not only Docker-controlled ones.

Concretely, both the agent-pull path (`addDockerAuth` in `sources.go`) and all three model-gateway clients (`openai`, `gemini`, `anthropic`) will forward the credential to any process bound to `localhost`. If a developer runs a third-party or test server on a loopback port (e.g. `http://localhost:8080`) and configures it as an agent source or a model gateway, their Docker Desktop JWT will be transmitted there.

The behavior is documented in the comment above the function, so it is intentional — but the JWT is a sensitive short-lived credential and the trust surface is broader than the `.docker.com` domain alone. Consider scoping the localhost exception to only the agent-pull path (where it makes the most sense) rather than the model-gateway path, or at minimum add a log warning when the JWT is sent to a non-docker.com host.

pkg/config/sources.go
```diff
@@ -362,6 +365,28 @@ func (a urlSource) addGitHubAuth(ctx context.Context, req *http.Request) {
 	slog.DebugContext(ctx, "Added GitHub token authorization to request", "url", a.url)
 }
 
+// addDockerAuth adds the Docker Desktop JWT to the request if:
```
docker-agent, line 368: **[LOW] Comment for `addDockerAuth` says ".docker.com domain" but the implementation also trusts localhost**

The function's doc-comment states:
```
// addDockerAuth adds the Docker Desktop JWT to the request if:
// - The URL targets a .docker.com domain
```
However, the underlying `isTrustedDockerURL` check also returns `true` for `localhost`, `127.0.0.1`, and `::1`, so the JWT will be sent to localhost URLs as well. The comment misrepresents the actual behavior and could mislead future maintainers into believing the localhost case is never triggered here.

Suggestion: update the comment to match reality, e.g.:
```
// - The URL targets a .docker.com domain or a localhost address (for local dev)
```

pkg/config/config.go
```diff
@@ -69,7 +69,7 @@ func Load(ctx context.Context, source Source) (*latest.Config, error) {
 //
 // This allows exiting early with a proper error message instead of failing later when trying to use a model or tool.
 func CheckRequiredEnvVars(ctx context.Context, cfg *latest.Config, modelsGateway string, env environment.Provider) error {
-	if modelsGateway != "" {
+	if modelsGateway != "" && environment.IsTrustedDockerURL(modelsGateway) {
```
docker-agent, line 72: **[LOW] Any localhost model gateway requires Docker Desktop sign-in, blocking non-Docker local gateways**

Because `IsTrustedDockerURL` returns `true` for localhost, `CheckRequiredEnvVars` will demand a Docker Desktop JWT for *any* `http://localhost:*` gateway — including purely local mock servers or third-party proxies that have no relationship with Docker auth.

A developer who points `DOCKER_GATEWAY` at their own local test server will receive:
> "sorry, you first need to sign in Docker Desktop to use the Docker AI Gateway"

...even though they never intended to use Docker's gateway. This is a friction point that may be hard to diagnose. If this is the intended behavior (localhost always implies a Docker gateway), a comment explaining the constraint would help; otherwise the check could be relaxed to only error when the token is actively needed (i.e. defer the check to the actual client call).

docker-agent (commented):

This PR correctly gates Docker Desktop JWT injection to  URLs (and localhost for dev), replacing the previous unconditional injection. The domain-matching logic in `IsTrustedDockerURL` is solid and well-tested. Three minor concerns noted below.

pkg/config/config.go
gtardif: This used to be blocked for any non empty gateway ; now it will be block users only if the token will be used later in the code when checking IsTrustedDockerURL.

docker-agent: Thanks for the context! I understand this is an improvement from blocking all gateways. However, the concern remains: because `IsTrustedDockerURL` includes localhost/127.0.0.1, a developer running their own local mock gateway (e.g., `http://localhost:8080`) will still hit the "sign in to Docker Desktop" error, even though their server has no relationship to Docker auth.

If localhost is intentionally treated as Docker-owned (perhaps because Docker Desktop runs local services there), that's a valid design choice — but it might help to document that constraint so users know localhost gateways are reserved for Docker-authenticated scenarios.

dgageot (approved):

PR #3080, opened 2026-06-11
Board/tui embedding helpers

dgageot:

PR #3081, opened 2026-06-11
feat(tui): expose read access to transcript messages

dgageot:
Follow-up to #3080 (TUI embedding seams for Gordon). The embedded transcript component now exposes a read-only `Messages()` accessor so embedders can observe conversation structure — host-side tests asserting on chat history, persistence layers reading state — without accessing private fields.

The accessor returns the backing slice as-is and the comment makes clear it must be treated as read-only: mutating entries would desync them from their rendered views, bloating the mutation surface. This keeps the component's contract clean while giving embedders the introspection they need.

Includes a unit test covering the empty case and the ordering of appended messages.

### DIFF
--- a/pkg/tui/components/transcript/transcript.go
+++ b/pkg/tui/components/transcript/transcript.go
@@ -264,3 +264,12 @@ func (t *Transcript) StopAnimations() {
 		animation.StopView(v)
 	}
 }
+
+// Messages returns the transcript's messages, oldest first. The slice is
+// the transcript's own backing store: callers must treat it as read-only
+// (mutating entries would desync them from their rendered views). It gives
+// embedders observability — host tests asserting on conversation structure,
+// or persistence of the chat — without growing the mutation surface.
+func (t *Transcript) Messages() []*types.Message {
+	return t.msgs
+}
--- a/pkg/tui/components/transcript/transcript_test.go
+++ b/pkg/tui/components/transcript/transcript_test.go
@@ -233,3 +233,17 @@ func TestTransferTaskGetsSeparator(t *testing.T) {
 	// Unlike other consecutive tool calls, a handoff stands apart.
 	assert.Contains(t, tr.Render(80), "\n\n")
 }
+
+func TestMessages(t *testing.T) {
+	tr := newTestTranscript()
+
+	assert.Empty(t, tr.Messages())
+
+	_ = tr.Append(types.User("hello"))
+	_ = tr.AppendToLastMessage(testAgent, "hi")
+
+	msgs := tr.Messages()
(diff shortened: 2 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes introduce a `Messages()` accessor on the `Transcript` component and a corresponding unit test. The implementation is intentionally simple: it returns the backing slice directly, with the trade-off explicitly documented in the comment (callers must treat it as read-only). No bugs were found in the changed code.

**Review coverage**: `pkg/tui/components/transcript/transcript.go`, `pkg/tui/components/transcript/transcript_test.go`

pkg/tui/components/transcript/transcript.go
```diff
@@ -264,3 +264,12 @@ func (t *Transcript) StopAnimations() {
...
 	}
 }
+
+// Messages returns the transcript's messages, oldest first. The slice is
+// the transcript's own backing store: callers must treat it as read-only
+// (mutating entries would desync them from their rendered views). It gives
+// embedders observability — host tests asserting on conversation structure,
+// or persistence of the chat — without growing the mutation surface.
+func (t *Transcript) Messages() []*types.Message {
+	return t.msgs
```
docker-agent, line 274: **[MEDIUM] `Messages()` returns the internal backing slice — callers can inadvertently corrupt Transcript state**

`Messages()` returns `t.msgs` directly. This exposes two hazards:

1. **Slice aliasing**: A caller that does `append(msgs, extra)` on the returned slice — without triggering a reallocation — will share the underlying array with the `Transcript`. If the `Transcript` later writes to the same indices (e.g. via `t.msgs[i]` assignment), the caller's copy is silently affected, and vice-versa. Similarly, re-slicing with `msgs[:0]` can zero out live entries.

2. **Pointer mutability**: Each element is a live `*types.Message`. Nothing stops a caller from doing `msgs[0].Content = "tampered"`, which modifies the message in place and desyncs it from its rendered view — exactly the hazard the comment warns against, but without enforcement.

A defensive copy makes the contract self-enforcing at negligible cost for typical transcript lengths:

```go
func (t *Transcript) Messages() []*types.Message {
    return append([]*types.Message(nil), t.msgs...)
}
```

If returning live pointers is intentional for performance, consider documenting in the type signature (e.g. returning `[]types.Message` value copies) or exposing a `ForEachMessage(func(*types.Message))` iterator that controls the call boundary.

docker-agent (commented):

melmennaoui (approved):

gtardif (commented):
LGTM :)

PR #3085, opened 2026-06-12
fix(mcpcatalog): mark semgrep server as oauth

trungutt:
The semgrep MCP server in the catalog was marked as `auth: none`, but the upstream server actually requires OAuth.

### DIFF
--- a/pkg/tools/builtin/mcpcatalog/servers.json
+++ b/pkg/tools/builtin/mcpcatalog/servers.json
@@ -695,7 +695,7 @@
       "readme": "http://desktop.docker.com/mcp/catalog/v3/readme/semgrep.md",
       "headers": {},
       "auth": {
-        "type": "none"
+        "type": "oauth"
       }
     },
     {


gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

**Summary**: This PR makes a single targeted fix — changing the `auth.type` for the semgrep MCP server entry from `"none"` to `"oauth"`. The change correctly aligns the catalog metadata with the upstream server's actual OAuth requirement, which will ensure users are offered the OAuth flow when adding the server.

**Analysis**:
- The change is minimal and correct: one field updated in one JSON entry
- Consistent with the catalog pattern for OAuth servers (same `auth: { type: "oauth" }` structure used by other OAuth entries)
- No other fields in the semgrep entry require modification alongside this auth type change
- The fix directly resolves the described user-facing issue (auth errors at connect time with no credential flow available)

No bugs or issues found in the changed code.

PR #3086, opened 2026-06-12
feat(runtime): discover gateway-served models for the model picker

### LINKED ISSUE #1606: Fetch models from models_gateway and prefer gateway-served models in picker
Summary
-------
When a `models_gateway` is configured, cagent currently assumes gateway supports core providers and shows a static list (configured models + models.dev catalog). It does not query the gateway for a live list of models. This causes incorrect/irrelevant models to appear in the model picker.

Proposal
--------
Add gateway model discovery when `models_gateway` is set: query `<gateway>/v1/models` (OpenAI-style) or a configurable metadata endpoint and use the returned model list to populate/suppress models in the picker.

Acceptance criteria
-------------------
- When `models_gateway` is set and `/v1/models` returns a list, the model picker shows only models served by that gateway (merged with explicitly defined models in config).
- Provide an opt-in config flag `models_gateway.discover: true/false` to control discovery.
- Add unit tests and an integration test that mocks gateway responses.
- Document behavior in `AGENTS.md` and `README.md`.

Notes
-----
- If gateway does not support `/v1/models`, fall back to current behavior.
- Consider caching the model list and exposing a refresh button in the TUI.

Sayt-0:
Closes #1606

## What

When a `models_gateway` is configured, docker-agent now queries the gateway's OpenAI-style `GET /v1/models` endpoint and populates the model picker with the models the gateway actually serves, merged with the models explicitly defined in the agent config. When the gateway doesn't expose `/v1/models` (or returns an empty/malformed list), it falls back to the previous behavior (models.dev catalog filtered by available credentials).

Discovery is automatic — no opt-in config flag.

## How

- **`pkg/modelsgateway`** (new): `ListModels()` queries `<gateway>/v1/models`:
  - 5s timeout so an unresponsive gateway doesn't block the picker
  - preserves the gateway URL's path prefix and query params
  - sends the Docker Desktop token as Bearer auth for trusted Docker URLs, mirroring the provider clients' gateway auth
- **`pkg/runtime/gateway_models.go`** (new): `buildGatewayChoices()` builds picker entries from the discovered IDs:
  - dedupes against configured models (which are always listed)
  - bare IDs (no `provider/` prefix) are routed through the `openai` provider since the gateway speaks the OpenAI protocol
  - filters out embedding models (by catalog family or model ID); enriches entries with models.dev metadata (pricing, context limits, modalities) best-effort
  - results — including failures — are cached for 5 minutes, coalesced with `singleflight` (double-checked inside the closure) so concurrent picker opens share one in-flight request and an unsupported gateway isn't re-queried on every open
- **`AvailableModels`**: prefers gateway discovery when `ModelsGateway` is set; falls back to `buildCatalogChoices` otherwise. Wire format unchanged (gateway entries reuse `is_catalog` so TUI grouping works as-is).

## Gateway-side fix (root cause of the staging discrepancy)

During testing, the Docker AI staging gateway advertised only 2 models on `/v1/models` while routing many more (its `restrict-models` circuit breaker is off, so the advertised list wasn't enforced). An earlier revision of this PR worked around that with a gateway/all-models toggle in the picker; that workaround is now **reverted** in favor of fixing the contract at the source:

- docker/gordon#852 (merged): `/v1/models` only advertises the curated list while the restriction is enforced; 404 otherwise → this PR's documented catalog fallback.
- docker/gordon#854: replaces that 404 with the models.dev catalog **filtered by the provider credentials the gateway holds** — so providers the gateway cannot route (e.g. mistral without `MISTRAL_API_KEY`) are never offered in the picker, and the catalog fallback here only remains for gateways without `/v1/models`.

**Deployment note**: best deployed gateway-first; reverse order is still safe (worst case: curated subset or catalog fallback).

## Tests

- `pkg/modelsgateway`: 8 unit tests with `httptest` mock gateways (success, path/query forwarding, auth header, missing token, 404, invalid JSON, unreachable gateway, invalid URL)
- `pkg/runtime`: 12 tests covering discovery, dedup with config models, catalog suppression, bare-ID routing, embedding filtering (ID and catalog family), metadata enrichment, fallback on 404/empty list, cache hit/expiry/negative caching, singleflight coalescing, and the TOCTOU double-check (deterministic, injected clock)
- Manual end-to-end against a local gordon proxy built from docker/gordon#852 (full middleware chain, stub auth): breaker ON → picker shows exactly the 2 advertised models; breaker OFF → `/v1/models` 404 → picker falls back to the full 124-model catalog

## Validation

- [x] `task build`
- [x] `task test` (only pre-existing `TestLoadExamples` DMR-pull failures, reproduced on `main`)
- [x] `task lint` — 0 issues
- [x] Docs: picker behavior documented under the TUI's *Runtime Model Switching* section (moved out of the CLI flags table per review)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27404625980).

Sayt-0:
## Update: gateway/all-models scope toggle in the picker

While testing against the Docker AI **staging** gateway, I found that its `/v1/models` endpoint advertises a **curated subset** (2 models: `anthropic/claude-haiku-4-5` "Fast" and `anthropic/claude-sonnet-4-5` "Deep") while the proxy actually **routes many more models** — a `chat/completions` call with `gpt-4o-mini` is forwarded to OpenAI just fine.

With the original behavior (gateway list *replaces* the catalog), those reachable-but-unadvertised models became undiscoverable in the picker. Since docker-agent has no reliable way to enumerate what a transparent proxy routes, strictly trusting `/v1/models` was too restrictive in practice. The proper fix is for gateways to advertise everything they serve, but until then the picker needed an escape hatch.

**What changed** (second commit):

- `ModelChoice` gains an `is_gateway` flag; the runtime now returns gateway-advertised entries **followed by the full catalog**, deduplicated. API consumers (`GET /api/sessions/:id/models`) can scope on `is_gateway` themselves.
- The TUI picker **defaults to the gateway scope** when gateway models are present (preserving the issue's acceptance criterion as the default UX) and `tab` toggles to the full list. Gateway entries get their own "Gateway models" section above "Other models", and the help bar advertises the toggle (`tab: show all models` / `show gateway models`).
- No behavior change when no gateway is configured or when the gateway doesn't expose `/v1/models`.

Verified end-to-end against the staging gateway: `total=124 config=1 gateway=2 catalog=121`, picker defaults to the 2 gateway models, `tab` reveals the rest.

pkg/runtime/gateway_models.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	c := &r.gatewayModels
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if !c.fetchedAt.IsZero() && now().Sub(c.fetchedAt) < gatewayModelsTTL {
+		return c.ids, c.err
+	}
+
```
docker-agent, line 59: **[MEDIUM] Mutex held across outbound network call — concurrent callers block for up to 5 s**

`listGatewayModels` acquires `c.mu` with `defer c.mu.Unlock()` and then makes a blocking network call to `modelsgateway.ListModels` (which carries a 5 s timeout). Any goroutine that concurrently calls `listGatewayModels` — e.g. the TUI opening the model picker a second time while the first request is still in flight — will stall for the remainder of that 5 s window.

A singleflight pattern (or lock-check-unlock-fetch-lock-store) would let concurrent callers coalesce on the single in-flight request without holding the mutex across I/O:

```go
// Option A: singleflight
ids, err, _ := sf.Do("gateway-models", func() (interface{}, error) {
    return modelsgateway.ListModels(ctx, ...)
})

// Option B: check with lock, fetch without, store with lock
c.mu.Lock()
if fresh { /* return cached */ c.mu.Unlock(); return ... }
c.mu.Unlock()
ids, err := modelsgateway.ListModels(ctx, ...)
c.mu.Lock()
c.ids, c.err, c.fetchedAt = ids, err, now()
c.mu.Unlock()
return ids, err
```

pkg/runtime/gateway_models.go
```diff
@@ -0,0 +1,120 @@
...
+		if _, err := latest.ParseModelRef(prov + "/" + model); err != nil {
+			continue
+		}
+		if isEmbeddingModel("", model) {
+			continue
+		}
+
+		ref := prov + "/" + model
+		if existingRefs[ref] {
+			continue
```
docker-agent, line 138: **[MEDIUM] Embedding filter ignores catalog `Family` field — embedding models may appear in picker**

`isEmbeddingModel("", model)` is called with an empty family string, so the check relies solely on the model's ID string (e.g. looking for `"embed"` in the name). The catalog lookup that would supply `m.Family` ("text-embedding" or similar) happens *after* this filter:

```go
if isEmbeddingModel("", model) {   // family="" — catalog Family not consulted
    continue
}
// ... later:
if m, err := r.modelsStore.GetModel(ctx, modelsdev.NewID(prov, model)); err == nil && m != nil {
    applyCatalogMetadata(&choice, m)  // Family available here, but too late
}
```

A gateway could serve a model like `openai/text-moderation-latest` or a vendor-specific embedding model whose ID contains no `"embed"` substring but whose catalog `Family` is `"text-embedding"`. That model would pass the filter and be added to the picker.

Consider moving the catalog lookup before the embedding check, or passing `m.Family` to `isEmbeddingModel` once the model is resolved:

```go
var family string
if r.modelsStore != nil {
    if m, _ := r.modelsStore.GetModel(ctx, modelsdev.NewID(prov, model)); m != nil {
        family = m.Family
    }
}
if isEmbeddingModel(family, model) {
    continue
}
```

docker-agent (commented):

Gateway model discovery feature looks solid overall. Found 2 medium-severity issues in the new `gateway_models.go` logic:

1. **Mutex held during network I/O** — concurrent picker opens stall for up to 5 s
2. **Embedding filter bypasses catalog Family** — models without 'embed' in their ID but with a text-embedding family can slip into the picker

Auth token gating (`IsTrustedDockerURL`), URL construction, fallback behavior, and cache TTL logic are all correct.

pkg/runtime/gateway_models.go
Sayt-0: Fixed in fe0e6e4bea3ec20867d484fb2f262360efabe9f5 — the fetch now runs outside the mutex, coalesced with `singleflight` so concurrent callers share a single in-flight request. Added `TestListGatewayModels_ConcurrentCallersCoalesce` (run with `-race`) asserting 8 concurrent callers produce exactly one HTTP request.

pkg/runtime/gateway_models.go
Sayt-0: Fixed in fe0e6e4bea3ec20867d484fb2f262360efabe9f5 — catalog metadata is now resolved *before* the embedding filter, so `isEmbeddingModel` receives the catalog `Family`. Added `TestAvailableModels_GatewayEmbeddingFilteredByCatalogFamily` covering a model with no 'embed' substring in its ID but a `text-embedding` family.

docker-agent: Perfect fix! The catalog metadata is now resolved before the filter, and the test case clearly demonstrates the scenario I was concerned about — a model like `some-vector-model` with no 'embed' substring in its ID but a `text-embedding` family is now correctly filtered out.

Thanks for addressing this and adding test coverage!

Sayt-0:
## Recap

**Original idea (#1606)**: when a `models_gateway` is configured, query its `GET /v1/models` endpoint and only show the models served by the gateway in the picker, instead of the full models.dev catalog.

**Why it was adjusted**: testing against the staging gateway revealed that its `/v1/models` only returns 2 models (Fast/Deep) even though the proxy routes many more (a `chat/completions` call with `gpt-4o-mini` goes through just fine). Strictly replacing the catalog with that list made all those models impossible to find in the picker.

**What was done in the end**: the picker shows the gateway-advertised models by default (the goal of the issue), but the full catalog stays accessible with `tab`. On the API side, gateway entries carry an `is_gateway` flag. No behavior change without a gateway or when `/v1/models` is not supported (catalog fallback as before).

docker-agent (commented):
### Assessment: 🟢 APPROVE

Gateway model discovery implementation looks solid. One concurrency note on the cache + singleflight pattern (see inline). All other findings are low-severity cosmetic/defensive code observations. No high-severity bugs found.

pkg/runtime/gateway_models.go
```diff
@@ -0,0 +1,148 @@
...
+
+	c.mu.Lock()
+	if !c.fetchedAt.IsZero() && now().Sub(c.fetchedAt) < gatewayModelsTTL {
+		ids, err := c.ids, c.err
+		c.mu.Unlock()
+		return ids, err
+	}
+	c.mu.Unlock()
+
+	v, err, _ := c.sf.Do("models", func() (any, error) {
```
docker-agent, line 60: **[MEDIUM] TOCTOU: cache-freshness check and `singleflight.Do` dispatch are not atomic**

`listGatewayModels` checks the cache under the lock, releases the lock, then calls `c.sf.Do("models", ...)`. A goroutine G3 that reads a stale/zero `fetchedAt` in the window *between* a just-completed singleflight write and its own cache check will call `singleflight.Do` after the group has already released the key, triggering a real HTTP fetch even though the cache was just populated.

The standard fix is a **double-checked lock** inside the singleflight closure:

```go
v, err, _ := c.sf.Do("models", func() (any, error) {
    // Re-check cache now that we hold the in-flight slot.
    c.mu.Lock()
    if !c.fetchedAt.IsZero() && now().Sub(c.fetchedAt) < gatewayModelsTTL {
        ids, err := c.ids, c.err
        c.mu.Unlock()
        return ids, err
    }
    c.mu.Unlock()

    ids, ferr := modelsgateway.ListModels(ctx, ...)
    c.mu.Lock()
    c.ids, c.err, c.fetchedAt = ids, ferr, now()
    c.mu.Unlock()
    return ids, ferr
})
```

Impact is low (extra redundant HTTP requests under burst concurrency, no data loss or crash), but the fix is straightforward and eliminates the superfluous gateway queries.

docs/features/cli/index.md
```diff
@@ -56,7 +56,7 @@ $ docker agent run [config] [message...] [flags]
...
 | `--env-from-file <path>`                | Load environment variables from file (repeatable)                                                                                         |
 | `--code-mode-tools`                     | Provide a single tool to call other tools via JavaScript (forces code-mode tools globally)                                                |
-| `--models-gateway <addr>`               | Route model traffic through a gateway. Also reads `DOCKER_AGENT_MODELS_GATEWAY` (legacy `CAGENT_MODELS_GATEWAY`) env var.                  |
+| `--models-gateway <addr>`               | Route model traffic through a gateway. Also reads `DOCKER_AGENT_MODELS_GATEWAY` (legacy `CAGENT_MODELS_GATEWAY`) env var. When the gateway exposes an OpenAI-style `/v1/models` endpoint, the model picker defaults to the models actually advertised by the gateway (merged with the models defined in the agent config); press `tab` in the picker to toggle between gateway-advertised models and the full catalog. When the gateway doesn't expose `/v1/models`, the picker falls back to the regular catalog. |
```
dgageot: Not sure this is the right place to put this information.

Sayt-0: Agreed — moved out of the CLI flags table in ce7e1d9e. The `--models-gateway` row is back to its terse one-liner, and the picker behavior is now documented where model picking lives: the *Runtime Model Switching* section of the TUI docs.

Note that the behavior itself also got simpler in the meantime: the gateway/all-models `tab` toggle was reverted (81817b5a) because the root cause — the staging gateway advertising a list it doesn't enforce — is fixed gateway-side in docker/gordon#852.

dgageot (approved):

PR #3088, opened 2026-06-12
fix: respect HOME env var when expanding ~ in paths

Divyansh670:
Fixes Windows test failure in pkg/tools/workingdir where HOME override was ignored because os.UserHomeDir() reads USERPROFILE instead.

### DIFF
--- a/pkg/path/home.go
+++ b/pkg/path/home.go
@@ -14,17 +14,25 @@ func ExpandHomeDir(path string) (string, error) {
 	if !isHomeDirPath(path) {
 		return path, nil
 	}
-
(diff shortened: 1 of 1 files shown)


Sayt-0:
Looks good. Thanks for the contribution!

Sayt-0 (approved):

Sayt-0:
One small thing: the commit (`c114ed67`) is unsigned. This repo requires commits to be signed with a GPG or SSH key. Could you re-sign and force-push? Thanks!

Divyansh670:
Done ✅.

PR #3090, opened 2026-06-12
feat: add options.WithHTTPTransportWrapper to inject HTTP middleware in provider clients

### LINKED ISSUE #3089: options.WithHTTPTransportWrapper: allow callers to inject HTTP middleware around all provider clients
## Problem

When embedding docker-agent programmatically (via `anthropic.NewClient`, `teamloader.Load`, etc.), there is currently no way to inject custom HTTP middleware into the transport chain used by provider clients — neither in direct mode nor in gateway/proxy mode.

Both the Anthropic and OpenAI provider clients build their HTTP client internally:

```go
// direct path (anthropic/client.go)
option.WithHTTPClient(httpclient.NewHTTPClient(ctx))

// gateway path (anthropic/client.go)
option.WithHTTPClient(httpclient.NewHTTPClient(ctx, httpOptions...))
```

The `options.ModelOptions` struct and `config.RuntimeConfig` carry no transport field, so there is no place to hook in.

## Use cases

- **Bearer-token injection for a private LLM proxy**: a private reverse proxy requires `Authorization: Bearer <token>` on every request. The token is known at client-construction time but must be injected regardless of whether the provider is in direct or gateway mode.
- **Custom observability / tracing**: adding spans or metrics around every LLM HTTP call, beyond what OTel wrapping already provides.
- **Request/response logging and debugging**: inspect raw HTTP traffic during development.
- **Retry/circuit-breaker policies**: wrap the transport with custom retry logic specific to the deployment.

All of these share the same shape: *wrap whatever transport docker-agent would build, apply some middleware, and delegate to the original*.

## Proposed API

Add a new `Opt` to the `options` package:

```go
// WithHTTPTransportWrapper registers a function that wraps the HTTP transport
// used by all provider clients. The function receives the transport that
// docker-agent built (including OTel instrumentation, SSE decompression fix,
// and Desktop proxy support) and must return a new RoundTripper that delegates
// to it. The wrapper is applied in both direct mode and gateway/proxy mode.
//
// Example — inject a bearer token on every outbound LLM request:
//
//	options.WithHTTPTransportWrapper(func(base http.RoundTripper) http.RoundTripper {
//	    return &bearerTransport{token: myToken, base: base}
//	})
func WithHTTPTransportWrapper(fn func(base http.RoundTripper) http.RoundTripper) Opt {
	return func(cfg *ModelOptions) {
		cfg.transportWrapper = fn
	}
}
```

And surface it via `ModelOptions`:

```go
func (o *ModelOptions) TransportWrapper() func(http.RoundTripper) http.RoundTripper {
	return o.transportWrapper
}
```

### Implementation sketch (Anthropic provider, direct path)

```go
httpClient := httpclient.NewHTTPClient(ctx)
if wrapper := globalOptions.TransportWrapper(); wrapper != nil {
	httpClient.Transport = wrapper(httpClient.Transport)
}
requestOptions := []option.RequestOption{
	option.WithHTTPClient(httpClient),
	...
}
```

Same two-liner applied to the gateway path. The same pattern applies symmetrically to the OpenAI provider (and any future providers that use `httpclient.NewHTTPClient`).

## Notes

- The wrapper function receives the **already-built** transport (post-OTel, post-SSE-filter, post-Desktop-proxy), so callers compose *on top of* docker-agent's chain rather than replacing it.
- A single `WithHTTPTransportWrapper` call is the common case; stacking multiple wrappers is possible by chaining the functions before passing them in.
- This is intentionally lower-level than `WithGateway` — it targets use cases where callers manage their own transport-level concerns that docker-agent has no opinion on.

simonferquel-clanker:
Closes #3089

Adds `options.WithHTTPTransportWrapper` — a new `Opt` that lets callers embedding docker-agent inject HTTP middleware into the transport chain of all provider clients.

**What changed**
- `options.ModelOptions`: new `transportWrapper` field, `WithHTTPTransportWrapper` Opt, `TransportWrapper()` accessor, `FromModelOptions` round-trip support
- Anthropic, OpenAI, Gemini (GeminiAPI backend): wrapper applied in both direct and gateway paths
- Gemini Vertex AI / Bedrock: wrapper cannot be applied (SDK-managed HTTP clients); a `slog.Warn` is emitted and limitations are documented
- All call sites guard nil returns from the wrapper (original transport preserved, warning logged)

**Tests**
- Unit tests in `options` package (field set/get, `FromModelOptions` round-trip)
- Direct-path and gateway-path integration tests in `anthropic`, `openai`, and `gemini` packages (httptest server + counting transport)
- Vertex AI no-op path verified in Gemini tests

**Deliberate decisions**
- The nil-return guard (`if wrapped := w(...); wrapped != nil`) appears at each call site rather than in a shared helper. This avoids adding a new exported symbol and keeps the pattern visible to reviewers; the six occurrences are mechanical.
- Bedrock and Gemini Vertex AI are documented as non-supported rather than wired, because those SDKs own their HTTP transport stack.

### DIFF
--- a/pkg/model/provider/bedrock/client.go
+++ b/pkg/model/provider/bedrock/client.go
@@ -18,6 +18,7 @@ import (
 	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
+	"github.com/docker/docker-agent/pkg/httpclient"
 	"github.com/docker/docker-agent/pkg/model/provider/base"
 	"github.com/docker/docker-agent/pkg/model/provider/options"
 	"github.com/docker/docker-agent/pkg/model/provider/providerutil"
@@ -64,7 +65,7 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		slog.WarnContext(ctx, "HTTP transport wrapper is set but not applied: Bedrock provider uses the AWS SDK HTTP client")
 	}
 
-	// Check for bearer token - use token_key if specified, otherwise try AWS_BEARER_TOKEN_BEDROCK.
+	// Check for bearer token
 	// Bearer token is optional: if not provided, falls back to standard AWS credential chain (SigV4).
 	//
 	// NOTE: Manual token handling is required because aws-sdk-go-v2's default credential chain
@@ -79,6 +80,28 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		}
 	} else {
 		bearerToken, _ = env.Get(ctx, "AWS_BEARER_TOKEN_BEDROCK")
+	}
+
+	// Build the docker-agent HTTP client (OTel instrumentation, SSE decompression,
+	// Desktop proxy support) so transport-level concerns apply to Bedrock too.
+	httpClient := httpclient.NewHTTPClient(ctx)
+
+	// If a bearer token is set, chain it on top of the base transport so auth
+	// headers are injected without replacing the rest of the transport stack.
+	if bearerToken != "" {
+		httpClient.Transport = &bearerTokenTransport{
+			token: bearerToken,
+			base:  httpClient.Transport,
+		}
+	}
+
+	// Apply the transport wrapper, if registered, over the full chain.
+	if w := globalOptions.TransportWrapper(); w != nil {
+		if wrapped := w(httpClient.Transport); wrapped != nil {
+			httpClient.Transport = wrapped
+		} else {
+			slog.WarnContext(ctx, "HTTP transport wrapper returned nil; using original transport")
+		}
 	}
 
 	// Build AWS config using default credential chain
@@ -98,19 +121,18 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		})
 	}
 
-	// If bearer token is set, use it instead of SigV4
+	// Inject our HTTP client (which carries OTel, SSE, bearer token if set, and
+	// any caller-registered transport wrapper) into the Bedrock runtime options.
 	if bearerToken != "" {
 		slog.DebugContext(ctx, "Bedrock using bearer token authentication")
 		clientOpts = append(clientOpts, func(o *bedrockruntime.Options) {
 			// Use anonymous credentials to skip SigV4 signing
 			o.Credentials = aws.AnonymousCredentials{}
-			// Add bearer token via custom HTTP client
-			o.HTTPClient = &http.Client{
-				Transport: &bearerTokenTransport{
-					token: bearerToken,
-					base:  http.DefaultTransport,
-				},
-			}
+			o.HTTPClient = httpClient
+		})
+	} else {
+		clientOpts = append(clientOpts, func(o *bedrockruntime.Options) {
+			o.HTTPClient = httpClient
 		})
 	}
 
--- a/pkg/model/provider/bedrock/client.go
+++ b/pkg/model/provider/bedrock/client.go
@@ -61,10 +61,6 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		opt(&globalOptions)
 	}
 
-	if globalOptions.TransportWrapper() != nil {
-		slog.WarnContext(ctx, "HTTP transport wrapper is set but not applied: Bedrock provider uses the AWS SDK HTTP client")
-	}
-
 	// Check for bearer token
 	// Bearer token is optional: if not provided, falls back to standard AWS credential chain (SigV4).
 	//
--- a/pkg/model/provider/bedrock/client.go
+++ b/pkg/model/provider/bedrock/client.go
@@ -60,6 +60,10 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		opt(&globalOptions)
 	}
 
+	if globalOptions.TransportWrapper() != nil {
+		slog.WarnContext(ctx, "HTTP transport wrapper is set but not applied: Bedrock provider uses the AWS SDK HTTP client")
+	}
+
 	// Check for bearer token - use token_key if specified, otherwise try AWS_BEARER_TOKEN_BEDROCK.
 	// Bearer token is optional: if not provided, falls back to standard AWS credential chain (SigV4).
 	//
--- a/pkg/model/provider/openai/client.go
+++ b/pkg/model/provider/openai/client.go
@@ -192,12 +192,12 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 	// Pre-create the WebSocket pool when the transport is configured.
 	// The pool is cheap (no connections opened until the first Stream call)
 	// and eager init avoids a data race on the lazy path.
-	if getTransport(cfg) == "websocket" && globalOptions.Gateway() == "" {
+	// WebSocket is also skipped when an HTTP transport wrapper is registered:
+	// gorilla/websocket dials raw TCP and never calls http.RoundTripper, so the
+	// wrapper cannot be applied. Fall back to SSE so the wrapper covers all calls.
+	if getTransport(cfg) == "websocket" && globalOptions.Gateway() == "" && globalOptions.TransportWrapper() == nil {
 		baseURL := cmp.Or(cfg.BaseURL, "https://api.openai.com/v1")
 		client.wsPool = newWSPool(httpToWSURL(baseURL), client.buildWSHeaderFn())
-		if globalOptions.TransportWrapper() != nil {
-			slog.WarnContext(ctx, "HTTP transport wrapper is set but not applied: WebSocket transport uses an SDK-managed connection")
-		}
 	}
 
 	return client, nil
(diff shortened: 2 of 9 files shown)


pkg/model/provider/openai/client.go
```diff
@@ -103,6 +103,13 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		clientOptions = append(clientOptions, option.WithMiddleware(oaistream.ErrorBodyMiddleware()))
 
 		httpClient := httpclient.NewHTTPClient(ctx)
+		if w := globalOptions.TransportWrapper(); w != nil {
+			if wrapped := w(httpClient.Transport); wrapped != nil {
+				httpClient.Transport = wrapped
+			} else {
+				slog.WarnContext(ctx, "HTTP transport wrapper returned nil; using original transport")
+			}
+		}
```
docker-agent, line 112: **[MEDIUM] Transport wrapper silently not applied to WebSocket pool**

The wrapper is correctly applied to the REST `httpClient` here, but when `transport: websocket` is configured (checked later at `getTransport(cfg) == "websocket"`), the `wsPool` is created via `newWSPool(httpToWSURL(baseURL), client.buildWSHeaderFn())` — using only auth headers, not the transport wrapper. Callers relying on the wrapper for observability, tracing, or auth injection will silently get no coverage over WebSocket streaming calls.

Both the Bedrock and Gemini Vertex AI paths emit a `slog.Warn` when the wrapper cannot be applied; this path should do the same. Consider adding near the `wsPool` creation:
```go
if w := globalOptions.TransportWrapper(); w != nil {
    slog.WarnContext(ctx, "HTTP transport wrapper is set but not applied: WebSocket transport uses an SDK-managed connection")
}
```

docker-agent (commented):

simonferquel-clanker:
Fixed in 59057b58:
- **Lint**: replaced `assert.Greater(t, x, int64(0))` with `assert.Positive(t, x)` in all three `wrap_transport_test` files (testifylint `negative-positive` rule).
- **WebSocket warning**: added `slog.WarnContext` adjacent to the `wsPool` creation when `globalOptions.TransportWrapper() != nil` — consistent with the Bedrock and Gemini Vertex AI paths.

pkg/model/provider/bedrock/client.go
```diff
@@ -60,6 +60,10 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		opt(&globalOptions)
 	}
 
+	if globalOptions.TransportWrapper() != nil {
+		slog.WarnContext(ctx, "HTTP transport wrapper is set but not applied: Bedrock provider uses the AWS SDK HTTP client")
```
rumpl: There really is no way to give bedrock an http?

simonferquel-clanker:
**Re: rumpl's comment on `bedrock/client.go:64`** — fixed in d7c54111.

Yes, the AWS SDK v2 absolutely supports a custom HTTP client: `bedrockruntime.Options.HTTPClient` accepts anything that implements `Do(*http.Request) (*http.Response, error)`, and `*http.Client` satisfies that interface.

The fix:
- Builds the docker-agent HTTP client (`httpclient.NewHTTPClient`) so Bedrock also gets OTel, SSE decompression, and user-agent transport.
- Chains `bearerTokenTransport` on top of that base when a bearer token is set (previously it wrapped `http.DefaultTransport`, bypassing OTel/SSE entirely — a bonus improvement).
- Applies the caller's transport wrapper over the full chain with the same nil-guard pattern used elsewhere.
- Injects the result via `o.HTTPClient = httpClient` in the `bedrockruntime.Options` closure.

The Bedrock limitation bullet has been removed from the `WithHTTPTransportWrapper` docstring.

pkg/model/provider/bedrock/client.go
simonferquel: I do not understand: I see the transport wrapper applied on line 99. Why do we have this warning ?

simonferquel-clanker:
Good catch — that `slog.Warn` at line 65 was a leftover from before the wrapper was wired in (d7c5411). Since the wrapper is now correctly applied further down via `bedrockruntime.Options.HTTPClient`, the upfront warning was wrong. Removed in b387a68d.

pkg/model/provider/openai/client.go
simonferquel: why can't we apply it to the websocket ?

docker-agent: Good question! Technically we **could** apply it, but it would require more work than just emitting the warning.

The WebSocket pool uses `gorilla/websocket.Dialer` (line 92 in ws_stream.go), which only accepts headers for the handshake — it doesn't expose an `http.RoundTripper` hook where the transport wrapper could be injected. To apply the wrapper, you'd need to either:
1. Use a custom dialer that wraps the underlying TCP connection (messy), or  
2. Switch to a WebSocket library that accepts an `http.Client` with a pluggable transport

Since WebSocket connections bypass the normal HTTP request/response cycle after the initial handshake, most observability wrappers won't capture the streaming data anyway — they'd only see the upgrade request. The warning (already present at line 175) lets users know their wrapper won't cover WebSocket traffic, which matches what Bedrock and Gemini do when SDK-managed clients can't be wrapped.

If you'd prefer to make the wrapper work, I'd suggest looking into whether `nhooyr.io/websocket` (which accepts `*http.Client`) would fit better. Otherwise the current approach (warn + document) seems reasonable.

simonferquel-clanker:
**Re: simonferquel's question about the WebSocket path** — the warning is accurate and must stay.

The transport wrapper is `func(http.RoundTripper) http.RoundTripper`. For REST calls the wrapper intercepts `http.RoundTripper.RoundTrip`, which is the right place for auth injection, tracing, etc.

For WebSocket, gorilla/websocket's `Dialer` never calls `RoundTrip` at all. Looking at its implementation:

```
// gorilla/websocket: Dialer.DialContext (client.go)
type Dialer struct {
    NetDialContext func(ctx context.Context, network, addr string) (net.Conn, error)
    NetDialTLSContext ...
    Proxy func(*http.Request) (*url.URL, error)
    TLSClientConfig *tls.Config
    // ...no http.RoundTripper field
}
```

It dials a raw TCP connection via `net.Dialer`, performs TLS itself (`tls.Client`), then writes the HTTP upgrade request as raw bytes over that socket. `http.RoundTripper` is completely bypassed — there is no intercept point for the wrapper in that code path.

The gorilla API only exposes `NetDialContext` (a `net.Conn` factory) as a hook, which is a different abstraction. We can't feed a `func(http.RoundTripper) http.RoundTripper` into it in any meaningful way.

Auth for the WebSocket handshake is already handled separately via `buildWSHeaderFn()` which attaches auth headers before the dial. The `slog.Warn` is kept so callers know upfront that their wrapper is not covering WebSocket streaming calls.

simonferquel:
OpenAI is supported trough cagent-proxy, including with injecting the docker-desktop bearer token. How does it work in that case ?

simonferquel-clanker:
Good question — this is actually a non-issue because **WebSocket and gateway are mutually exclusive in the existing code**.

Two guards enforce this:

**At client construction** (`NewClient`, line 195):
```go
if getTransport(cfg) == "websocket" && globalOptions.Gateway() == "" {
    client.wsPool = newWSPool(...)   // only created in direct mode
}
```

**At stream selection** (`CreateChatCompletionStream`, line 536):
```go
if transport == "websocket" && c.ModelOptions.Gateway() == "" {
    // use WebSocket
} else if transport == "websocket" {
    // Gateway is set → log debug + fall through to SSE
}
```

So when OpenAI is used through cagent-proxy (gateway configured), the code unconditionally falls back to SSE regardless of the `transport: websocket` setting. The Docker Desktop bearer token injection happens in the SSE gateway `clientFn` closure — that path uses our `httpclient.NewHTTPClient` transport chain, and the transport wrapper IS applied there.

The `slog.Warn` for the WebSocket case is therefore only ever emitted in **direct mode** (no gateway, no Docker Desktop JWT involved). There is no gap in the proxy/bearer-token scenario.

simonferquel:
Ok, do the same when a transport wrapper is specified then (fallback to sse automatically).

Vertex AI is not supported by the cagent proxy in some similar fashion ?

simonferquel-clanker:
Both addressed in 5bf51281.

**Request 1 — WebSocket auto-fallback to SSE when transport wrapper is set**

Done. Added `&& globalOptions.TransportWrapper() == nil` to both the `wsPool` creation guard in `NewClient` and the `if transport == "websocket"` branch in `CreateResponseStream`. When a wrapper is registered:
- `wsPool` is never created (no WebSocket connections will be opened)
- `CreateResponseStream` logs a debug message and falls straight through to the SSE path, where the wrapper is fully applied

The old `slog.Warn` is gone. A new test `TestNewClient_WebSocketFallsBackToSSEWhenTransportWrapperSet` asserts that `wsPool` is nil when a wrapper is set alongside `transport: websocket`.

---

**Request 2 — Vertex AI via cagent-proxy**

Same mutual-exclusivity pattern as WebSocket. Looking at `gemini/client.go`:

```go
if gateway := globalOptions.Gateway(); gateway == "" {
    // Direct path: may use Vertex AI if project/location are set
    // (httpClient = nil → slog.Warn when wrapper is set)
} else {
    // Gateway path: always uses BackendGeminiAPI with httpclient.NewHTTPClient
    // Vertex AI is never reached here
}
```

When a gateway is configured (cagent-proxy / Docker Desktop), the code always takes the `else` branch and builds a `genai.Client` with `BackendGeminiAPI` and our `httpclient.NewHTTPClient` transport. The Vertex AI path (`httpClient = nil`) is only reachable in direct mode (no gateway). So the existing `slog.Warn` for Vertex AI is accurate and only ever fires in direct mode where cagent-proxy is not involved — no gap.

simonferquel:
Ok, use the same logic as with openAI: if a custom wrapper is specified, use genai.Client with BackendGeminiAPI

simonferquel-clanker:
Both fixed in 05e69023.

**Lint** — replaced the `if/else if/else if` chain in `CreateResponseStream` with a `switch` statement to satisfy the `gocritic` `ifElseChain` rule.

**Vertex AI fallback** — mirrors the WebSocket → SSE pattern exactly:
```go
_, useVertexAIEnv := env.Get(ctx, "GOOGLE_GENAI_USE_VERTEXAI")
wantVertexAI := cfg.ProviderOpts["project"] != nil || cfg.ProviderOpts["location"] != nil || useVertexAIEnv
useVertexAI := wantVertexAI && globalOptions.TransportWrapper() == nil
```
When a transport wrapper is set and Vertex AI would normally be used, `useVertexAI` is false and the code falls through to the GeminiAPI branch (with `httpclient.NewHTTPClient` + wrapper applied). A debug log records the fallback. The old `slog.Warn` is removed.

`TestNewClient_TransportWrapperVertexAIWarns` replaced by `TestNewClient_TransportWrapperVertexAIFallsBackToGeminiAPI` which spins up an httptest server, sets `project`/`location` + `GOOGLE_API_KEY`, and asserts the wrapper's `RoundTrip` is called on a real stream.

gtardif (approved):

PR #3095, opened 2026-06-12
feat: add force_handoff for deterministic agent routing on natural stop

### LINKED ISSUE #3078: Forced handoff or transfer for more reliable workflows
## Overview

Add a `force_handoff: <agent_name>` parameter to agent configurations to guarantee a transition to a specific agent when the current one finishes, bypassing the LLM entirely.

## Motivation

Relying on LLMs to reliably call handoff or transfer tools is inherently flaky. Even with strict system prompts, LLMs can ignore the tool, hallucinate, or get stuck in loops. This parameter provides a native way to enforce deterministic routing.

## Use Cases

1. **Strict Pipelines:** Unconditional handoffs or transfer where autonomy isn't needed (e.g., `DataExtractor` always hands off to `Summarizer`).
2. **Setup/Teardown:** Guaranteed transitions like `SetupAgent` -> `ExecutionAgent` -> `CleanupAgent`.
3. **Escalation:** Hardcoded logic to ensure an issue reaches a `SpecialistAgent` without relying on the LLM's decision-making.

## Proposed Solution

Introduce a new field in the agent configuration:

```yaml
agents:
  researcher:
    model: ...
    force_handoff: writer_agent # or force_transfer: writer_agent

```

**Implementation:** When the current agent reaches a natural stop (e.g., generates a final text response), the orchestrator intercepts the finish state and automatically routes to the target agent, carrying over the context.

Sayt-0:
## What

Adds a `force_handoff` config parameter on agents. When set, the runtime intercepts the agent's natural stop (final text response) and **deterministically routes the conversation to the target agent**, bypassing the LLM's tool-calling entirely. The full conversation context carries over (same session, same handoff semantics as the `handoff` tool).

Closes #3078

## Why

With `handoffs`, the model *chooses* whether to call the `handoff` tool — and can simply forget to, silently breaking pipelines that depend on a strict order (e.g. extractor → summarizer). `force_handoff` makes such pipelines reliable by moving the routing decision from the model to the runtime.

## How

- **Config (latest schema only)**: new `force_handoff: <agent>` field on `AgentConfig`; frozen `vN` packages untouched. `agent-schema.json` updated (`TestSchemaMatchesGoTypes` passes).
- **Validation** (config load time):
  - target must exist locally or be an external reference (same rules as `handoffs`, incl. external-name collision check)
  - an agent cannot `force_handoff` to itself
  - `force_handoff` chains must not form a cycle (each agent has at most one outgoing edge, so detection is a linear walk)
- **Runtime**: in `runTurn`, on `res.Stopped`, if the agent has a `force_handoff` target the runtime fires `on_agent_switch` (new kind `force_handoff`), switches the current agent, injects an implicit user message (mirroring the `handoff` tool's notice, so the next model call doesn't start on a dangling assistant message) and continues the loop. `turn_end` reports the new reason `force_handoff`.
  - **Guard**: skipped for pinned sessions (background agents) — `resolveSessionAgent` would keep returning the pinned agent, turning the forced switch into an infinite stop/handoff loop.
- **Teamloader**: resolves the reference via the existing `resolveAgentRefs` (supports local + external agents).
- **Docs**: agents config reference, new *Forced Handoffs* section in multi-agent concepts, hooks reference (`agent_switch_kind`).
- **Example**: `examples/force_handoff.yaml` (two-stage extractor → summarizer pipeline).

## Tests

- `pkg/config`: validation table tests (missing target, self-reference, 2- and 3-agent cycles, valid chains, external refs, external-name collision)
- `pkg/runtime/force_handoff_test.go`:
  - `TestForceHandoff_RoutesToTargetOnNaturalStop` — target agent runs exactly once in the same session, sees the previous agent's output and the handoff notice, current agent ends up on the target
  - `TestForceHandoff_SkippedForPinnedSession` — pinned sessions never trigger the forced switch
- `examples/force_handoff.yaml` covered by `TestLoadExamples`

## Validation

- [x] `task build`
- [x] `task test` (only pre-existing, environment-dependent failures: `dmr.yaml` / `unload_on_switch.yaml` pull DMR models; they fail identically on a clean tree)
- [x] `task lint` — 0 issues

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -574,6 +574,10 @@
             "type": "string"
           }
         },
+        "force_handoff": {
+          "type": "string",
+          "description": "Name of an agent that unconditionally receives the conversation whenever this agent produces a final response, bypassing the LLM's tool-calling entirely. Guarantees deterministic routing for strict pipelines (e.g. extractor -> summarizer). Can be the name of an agent defined in this config or an external reference. Must not reference the agent itself, and force_handoff chains must not form a cycle."
+        },
         "add_date": {
           "type": "boolean",
           "description": "Whether to add date information"
--- a/docs/concepts/multi-agent/index.md
+++ b/docs/concepts/multi-agent/index.md
@@ -147,6 +147,34 @@ agents:
 
 </div>
 
+### Forced Handoffs
+
+With `handoffs`, the **model decides** whether to call the `handoff` tool — which means it can forget to, breaking pipelines that depend on a strict order. `force_handoff` removes that uncertainty: whenever the agent produces a final response, the **runtime itself** routes the conversation to the named agent, bypassing the LLM's tool-calling entirely. The full conversation context carries over.
+
+```yaml
+agents:
+  root:
+    model: anthropic/claude-sonnet-4-5
+    description: Extracts key facts from the input
+    instruction: |
+      Extract the key facts from the user's input as a bullet list.
+    force_handoff: summarizer
+
+  summarizer:
+    model: anthropic/claude-sonnet-4-5
+    description: Produces the final summary
+    instruction: |
+      Summarize the extracted facts for the user.
+```
+
+Rules enforced at config load time:
+
+- The target must be an agent defined in the config (or an external reference)
+- An agent cannot `force_handoff` to itself
+- Chains of `force_handoff` edges must not form a cycle (A → B → A is rejected)
+
+See <a href="https://github.com/docker/docker-agent/blob/main/examples/force_handoff.yaml"><code>examples/force_handoff.yaml</code></a> for a runnable example.
+
 ## Parallel Delegation with Background Agents
 
 `transfer_task` is **sequential** — the coordinator waits for the sub-agent to finish before continuing. When you need to fan out work to multiple agents at the same time, use the `background_agents` toolset instead.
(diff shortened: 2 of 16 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27425286205).

dgageot:
Nice, well-scoped feature — good validation (self-refs, cycles), reuse of existing handoff/session routing, and solid test + docs coverage.

A few comments:

1. **Steering-message ordering** (`pkg/runtime/loop.go`): `force_handoff` runs on the stop path before the steering-message recheck, so a late steering message gets handled by the handoff target instead of the original agent. Please either reorder or add a comment stating force-handoff intentionally wins.
2. **Docs**: after A → B, the session agent stays B — follow-up messages won't restart from A. Worth a short note since the "pipeline" framing may suggest otherwise.
3. **Minor**: `next != nil && sess.AgentName == ""` duplicates pinned-session knowledge (a `sess.IsPinned()` helper would be cleaner); the cycle-detection comment overstates linearity (worst case ~O(n²)); consider emitting an agent-switching event for forced handoffs for parity with explicit ones.

Approving with minor comments — only #1 needs clarification before merge.

dgageot (approved):

PR #3097, opened 2026-06-12
feat(effort): cycle thinking levels from per-model capabilities

rumpl:
Replace the per-provider thinking-level cycles with capability-driven selection. pkg/effort now exposes the generic level algebra: a three-state LevelMap, SupportedLevels (explicit-only top tiers xhigh and max), Clamp (nearest supported level, searching upward first), and NextSupportedLevel (wraparound cycling).

pkg/modelinfo builds the per-model level set: Claude models top out at high except Opus 4.6 (max) and Opus 4.7+/Fable (xhigh); OpenAI gets xhigh from gpt-5.2 on; Gemini and unknown providers keep their previous scales. CycleAgentThinkingLevel now advances through the levels the specific model accepts, so e.g. claude-sonnet is never offered xhigh.

### DIFF
--- a/pkg/effort/effort.go
+++ b/pkg/effort/effort.go
@@ -3,7 +3,10 @@
 // this package instead of hard-coding effort strings.
 package effort
 
-import "strings"
+import (
+	"slices"
+	"strings"
+)
 
 // Level represents a thinking effort level.
 type Level string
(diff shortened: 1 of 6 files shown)


pkg/effort/effort.go
```diff
@@ -75,63 +78,89 @@ func ValidNames() string {
...
 
-// NextThinkingLevel returns the level following current in the provider's
-// thinking cycle, wrapping back to the first level. When current is not in
-// the cycle the first level is returned.
-func NextThinkingLevel(providerType string, current Level) Level {
-	cycle := ThinkingCycle(providerType)
-	for i, l := range cycle {
-		if l == current {
-			return cycle[(i+1)%len(cycle)]
+// Clamp maps requested onto the nearest level in supported. When requested
```
docker-agent, line 121: **[MEDIUM] `Clamp` doc comment says "degrades" but upward-first scan can silently escalate effort**

The `Clamp` function scans **upward** (toward higher effort) before scanning downward. The doc comment says:

> "so a too-precise request degrades to the closest tier the model accepts"

But "upward first" means that if a user configured `Low` effort and the model's supported set is `[None, Medium, High]`, `Clamp` returns `Medium` — **silently escalating** the effort beyond what the user requested. This contradicts the documented semantics ("degrades") and has real cost implications (Anthropic bills based on thinking budget tier).

A concrete example not covered by existing tests:
```go
// User configures Low; model only supports None, Medium, High
Clamp([]Level{None, Medium, High}, Low) // returns Medium (escalation!)
// But the comment implies it should return None (nearest downward)
```

The upward-first behaviour is intentional for the cycling use-case (where you want to advance forward), but when Clamp is used to re-enter the cycle from a persisted configured level, silently upgrading to a higher — and more expensive — tier may surprise users. Consider either:
1. Flipping to downward-first so `Clamp` truly "degrades", or
2. Fixing the doc comment to accurately describe the upward-first escalation behaviour.

docker-agent (commented):

Reviewed **6 files** (pkg/effort, pkg/modelinfo, pkg/runtime). The capability-driven level algebra is clean and well-structured. One medium-severity semantic issue found in `Clamp`; all other findings were low-severity or dismissed.

aheritier (approved):

Issue #3104, opened 2026-06-13
TUI: improve readability of delegated / multi-agent execution


aheritier:
## Summary
When `root` delegates to a sub-agent and waits, the TUI is ambiguous about *who is working*, *who is waiting*, and *how many* agents are in flight — and it gets worse when multiple agents are involved. This issue groups a set of focused, independently-shippable improvements (see the linked phase issues).

## Background: two delegation mechanisms that render very differently
1. **`transfer_task` (synchronous "forwarding")** — `LocalRuntime.runForwarding` (`pkg/runtime/agent_delegation.go`) with `SwitchCurrentAgent: true`. The parent loop is **blocked** while the child runs; child events are **forwarded inline** into the same transcript. Multiple `transfer_task` calls in one assistant turn run **sequentially** (`Dispatcher.Process` iterates calls in order, `pkg/runtime/toolexec/dispatcher.go`).
2. **`run_background_agent` (asynchronous "collecting")** — `RunAgent` → `runCollecting` (`pkg/runtime/agent_delegation.go`) with `NonInteractive + PinAgent`, up to 20 concurrent. **Child events are dropped**; only the final result is collected. The human sees nothing live except the initial "task started" tool call and any later `view_background_agent` polls the model chooses to make.

## Concrete readability gaps
- The waiting **spinner is anonymous** — no "who's working / who's waiting" label (`pkg/tui/components/message/message.go`, `MessageTypeSpinner` renders a bare `spinner.View()`; added via `setPendingResponse` in `pkg/tui/page/chat/chat.go`).
- The **`transfer_task` card hardcodes a green ✓** regardless of status (`pkg/tui/components/tool/transfertask/transfertask.go` uses `styles.ToolCompletedIcon` unconditionally), so an in-progress delegation looks finished.
- The transcript is **flat**: `chatPage.streamDepth` is tracked but used only to balance spinner/cancel state, never to group or label sub-agent output.
- **`SubSessionCompletedEvent` is emitted by the runtime but not handled by the TUI** (no references under `pkg/tui/`).
- **Background-agent tools have no dedicated renderer** (`run/list/view/stop_background_agent` fall through to the default tool renderer) and the **sidebar tracks only one working agent** (`pkg/tui/components/sidebar/sidebar.go`).

## Proposed phases
- [x] Phase 1 — Make a single delegation unambiguous (`transfer_task`) — #3101
- [x] Phase 2 — Make sequential / multiple delegations legible — #3102
[...]

Issue #3103, opened 2026-06-13
TUI: background-agent visibility (+ optional depth nesting)


aheritier:
Part of #3104

`run_background_agent` work is essentially invisible in the UI: no live progress, no count, generic tool cards, and a single-agent sidebar. This is the strongest "multiple agents" case.

## Proposed changes
1. **Dedicated renderers for the background-agent tools.** Add `pkg/tui/components/tool/backgroundagent/` renderers for `run/list/view/stop_background_agent`, registered in `pkg/tui/components/tool/factory.go`, rendering compact cards (agent badge + status chip + elapsed) instead of raw text. Pure presentation; no runtime change.
2. **(Optional) Background-agent status surface.** A compact "N background agents running" indicator with per-task status. This needs a runtime seam because `runCollecting` deliberately drops child events — prefer emitting **coarse** lifecycle events (started/finished with task id + agent name + final status) without forwarding the full stream. ⚠️ Concurrency-sensitive (up to 20 goroutines); gate on maintainer approval.
3. **(Stretch) Depth-based indentation** of forwarded sub-agent messages using `streamDepth`. High blast radius in `pkg/tui/components/messages/` (width/wrap/selection/scroll math); only pursue if Phases 1–2 prove insufficient.

## Acceptance criteria
- Background-agent tool calls render as compact, readable cards (agent, task id, status, elapsed).
- (If approved) a live count/status of running background agents is visible.

## Validation
- `task build && task test && task lint` green.
- Renderer tests (ANSI-stripped: agent, task id, status, elapsed).

Issue #3102, opened 2026-06-13
TUI: make sequential / multiple delegations legible


aheritier:
Part of #3104

When several `transfer_task` calls happen in one turn (they run sequentially), the transcript is one undifferentiated stream and the sidebar shows only a single working agent, so the delegation structure is lost.

## Proposed changes
1. **Transcript delegation brackets.** On child stream start/stop, render theme-styled delimiter lines, e.g. `── root → librarian ─────` on entry and a matching completion marker on exit. Drive these from `AgentSwitchingEvent` (carries `FromAgent`/`ToAgent`) and by **handling `SubSessionCompletedEvent`** in `pkg/tui/page/chat/runtime_events.go` (currently unhandled). Reconcile with `messages.needsSeparator` (which already special-cases `transfer_task`).
2. **Sidebar delegation breadcrumb + background count.** In `pkg/tui/components/sidebar/sidebar.go`, extend the single `workingAgent` notion to show a short chain (e.g. `root ⏵ librarian`) when `streamDepth > 1`, plus a count badge when ≥1 background agents are active. Reuse `AgentAccentStyleFor` for per-agent color.

## Acceptance criteria
- Sequential delegations render as distinct, labeled sections.
- Returning control to the parent is visually marked.
- The sidebar shows the active delegation chain, not just the deepest single agent.

## Validation
- `task build && task test && task lint` green.
- Test that `SubSessionCompletedEvent` produces the completion delimiter and keeps `streamDepth` balanced.
- Integrate new transcript lines with existing separator/scroll/selection logic (extend `pkg/tui/components/messages` tests).

Issue #3101, opened 2026-06-13
TUI: make a single delegation unambiguous (transfer_task)


aheritier:
Part of #3104

While `root` waits on a delegated sub-agent, the user can't tell who is working or that the delegation is still in progress. The waiting spinner is anonymous and the `transfer_task` card always shows a green ✓.

## Proposed changes
1. **Status-aware `transfer_task` card.** In `pkg/tui/components/tool/transfertask/transfertask.go`, pick the icon from `msg.ToolStatus` (pending/running → spinner or `ToolPendingIcon`; completed → `ToolCompletedIcon` ✓; error → `ToolErrorIcon`) instead of the hardcoded ✓. The `Renderer` already receives the message and a `spinner.Spinner`, so a running delegation can animate. Mirror the status branching already used in `pkg/tui/components/toolcommon/common.go`.
2. **Name the waiting state.** Replace the anonymous pending spinner with a labeled one for delegations, e.g. `⟳ root → <child> (delegating…)`. Add a label-carrying spinner constructor in `pkg/tui/types/types.go`, render it in `pkg/tui/components/message/message.go` (`case types.MessageTypeSpinner`), and populate it from `handleStreamStarted` / `setPendingResponse` (`pkg/tui/page/chat/runtime_events.go`, `chat.go`) using the current/previous agent names and `streamDepth`.

## Acceptance criteria
- An in-progress `transfer_task` shows a non-✓ (animated/pending) indicator; completed shows ✓; error shows the error icon.
- The waiting spinner names the child agent (ideally parent → child).
- The labeled spinner stays animated/uncached (`messageModel.isSpinnerDriven()` still returns true for it).

## Validation
- `task build && task test && task lint` green.
- New render test for the `transfertask` renderer per `ToolStatus`.
- New `message` test that the labeled spinner renders the parent→child text and remains spinner-driven.

PR #3115, opened 2026-06-15
feat(tui): live status + labeled spinner for delegation (#3101)

aheritier:
## Summary

Makes a single `transfer_task` delegation legible at a glance. Two focused, presentation-only changes:

1. **Status-aware delegation card.** The `transfer_task` card hard-coded a green `✓`, so an in-flight delegation looked finished while the sub-agent was still running. The leading glyph is now driven by `ToolStatus`: an animated spinner while Running/Pending, `✓` on completion, `✗` on error. The icon stays a single fixed-width glyph (no elapsed-time suffix) so the wrapped task-text indent doesn't drift across statuses.
2. **Labeled "parent → child" pending spinner.** While the parent waited on a sub-agent, the pending spinner showed an anonymous whimsical word with no hint of who was working. For delegated streams (depth ≥ 2) it now reads `root → helper` in the child's accent color, via a new `types.SpinnerLabeled`. An `agentStack` is tracked in lockstep with `streamDepth`. The normal top-level turn keeps the playful spinner.

## Screenshots

_While `root` waits on `helper`: the card animates a spinner and the pending line reads **`root → helper`**. (The sidebar `root ▸ helper` breadcrumb is added in #3102 — the demo was recorded on the full stack.)_

phase1-labeled-spinner

## Tests

- `pkg/tui/components/tool/transfertask` — renderer asserts the spinner/✓/✗ glyph per `ToolStatus` + a fixed-icon-width invariant.
- `pkg/tui/components/message` — labeled spinner renders the `parent → child` text and stays spinner-driven (animated, uncached); a bare spinner is unchanged.
- `pkg/tui/page/chat` — `pendingSpinnerContext` depth logic (0/1/2/3).

`task build` / `task lint` clean; touched-package tests pass. (Repo-wide, the only failure is the pre-existing, unrelated `pkg/teamloader/TestLoadExamples`, a DMR model-pull 416 with no `pkg/tui` dependency.)

## Stack

Base of the delegation-readability stack (tracking #3104): **#3101 ← #3102 ← #3103**. Pure UI/UX; no config/schema changes.

Refs #3101

---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #3116
- <kbd>&nbsp;1&nbsp;</kbd> #3115 👈

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -55,7 +55,7 @@ type Model interface {
 	AddLoadingMessage(description string) tea.Cmd
 	ReplaceLoadingWithUser(content string, sessionPos int) tea.Cmd
 	AddErrorMessage(content string) tea.Cmd
-	AddAssistantMessage() tea.Cmd
+	AddAssistantMessage(sender, label string) tea.Cmd
 	AddCancelledMessage() tea.Cmd
 	AddWelcomeMessage(content string) tea.Cmd
 	AddOrUpdateToolCall(agentName string, toolCall tools.ToolCall, toolDef tools.Tool, status types.ToolStatus) tea.Cmd
@@ -1237,8 +1237,12 @@ func (m *model) AddShellOutputMessage(content string) tea.Cmd {
 	return m.addMessage(types.ShellOutput(content))
 }
 
-func (m *model) AddAssistantMessage() tea.Cmd {
-	return m.addMessage(types.Spinner())
+func (m *model) AddAssistantMessage(sender, label string) tea.Cmd {
+	m.removeSpinner()
+	if label == "" {
+		return m.addMessage(types.Spinner())
+	}
+	return m.addMessage(types.SpinnerLabeled(sender, label))
 }
 
 func (m *model) AddCancelledMessage() tea.Cmd {
--- a/pkg/tui/components/tool/transfertask/transfertask.go
+++ b/pkg/tui/components/tool/transfertask/transfertask.go
@@ -19,7 +19,7 @@ func New(msg *types.Message, sessionState service.SessionStateReader) layout.Mod
 	return toolcommon.NewBase(msg, sessionState, render)
 }
 
-func render(msg *types.Message, _ spinner.Spinner, _ service.SessionStateReader, width, _ int) string {
+func render(msg *types.Message, s spinner.Spinner, _ service.SessionStateReader, width, _ int) string {
 	var params transfertask.Args
 	if err := json.Unmarshal([]byte(msg.ToolCall.Function.Arguments), &params); err != nil {
 		return ""
@@ -29,8 +29,9 @@ func render(msg *types.Message, _ spinner.Spinner, _ service.SessionStateReader,
 		" calls " +
 		styles.AgentBadgeStyleFor(params.Agent).Render(params.Agent)
 
-	// Calculate the icon with its margin
-	icon := styles.ToolCompletedIcon.Render("✓")
+	// Status-aware icon: spinner while the delegation runs, ✓ on success, ✗ on error.
+	// Single glyph (no elapsed suffix) keeps the wrap math below stable.
+	icon := statusIcon(msg, s)
 	iconWithSpace := icon + " "
 	iconWidth := lipgloss.Width(iconWithSpace)
 
@@ -56,4 +57,22 @@ func render(msg *types.Message, _ spinner.Spinner, _ service.SessionStateReader,
 	}
 
 	return header + "\n\n" + taskContent.String()
+}
+
+// statusIcon picks the leading glyph for the delegation card from the tool
+// status: an animated spinner while the sub-agent runs, ✓ on success, ✗ on
+// error. The transfertask Base spinner is ModeSpinnerOnly, so s.View() is a
+// single 1-cell glyph whose width matches ✓/✗, keeping the task-text wrap math
+// stable (no elapsed-time suffix, unlike toolcommon.Icon).
+func statusIcon(msg *types.Message, s spinner.Spinner) string {
+	switch msg.ToolStatus {
+	case types.ToolStatusRunning, types.ToolStatusPending, types.ToolStatusConfirmation:
+		return styles.NoStyle.MarginLeft(2).Render(s.View())
+	case types.ToolStatusCompleted:
+		return styles.ToolCompletedIcon.Render("✓")
+	case types.ToolStatusError:
+		return styles.ToolErrorIcon.Render("✗")
+	default: // genuinely unknown/terminal states
+		return styles.ToolCompletedIcon.Render("✓")
+	}
 }
--- a/pkg/tui/page/chat/runtime_events.go
+++ b/pkg/tui/page/chat/runtime_events.go
@@ -202,6 +202,7 @@ func (p *chatPage) handleStreamStarted(msg *runtime.StreamStartedEvent) tea.Cmd 
 	slog.Debug("handleStreamStarted called", "agent", msg.AgentName, "session_id", msg.SessionID)
 	p.streamCancelled = false
 	p.streamDepth++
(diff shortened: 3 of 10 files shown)


docker-agent (commented):

Two findings in the labeled-spinner lifecycle: one medium (stale spinner persists after sub-agent completes) and one low (potential duplicate spinners from missing eviction in AddAssistantMessage). The status-aware delegation card in `transfertask.go` and the `agentStack`/`streamDepth` invariant look correct.

pkg/tui/page/chat/runtime_events.go
```diff
@@ -238,6 +239,9 @@ func (p *chatPage) handleStreamStopped(msg *runtime.StreamStoppedEvent) tea.Cmd
 	if p.streamDepth > 0 {
 		p.streamDepth--
 	}
+	if n := len(p.agentStack); n > 0 {
```
docker-agent: **[MEDIUM] Labeled spinner "parent → child" not removed when sub-agent stream ends**

When a sub-agent stream stops (depth 2→1), `handleStreamStopped` decrements both `streamDepth` and pops `agentStack` here — then immediately takes the early-return path (`if p.streamDepth > 0 { return ... }`) without calling `setPendingResponse(false)`. This leaves the labeled spinner (e.g., "root → researcher") visible in the message list even though the child agent has already finished.

**Trigger path:**
1. Sub-agent starts (depth 2) → `setPendingResponse(true)` adds a "root → researcher" labeled spinner
2. Sub-agent produces no direct content (tool-only responses); its spinner was never replaced by a content message
3. Sub-agent's `StreamStopped` fires, depth drops 2→1, `agentStack` is popped here
4. Early-return fires — labeled spinner is **not removed**, still reads "root → researcher"
5. Only when the parent next produces content or its own stream stops does the spinner finally disappear

**Fix:** Call `setPendingResponse(false)` (and optionally re-add a plain spinner for the still-running parent) before the early return:
```go
if p.streamDepth > 0 {
    p.setPendingResponse(false) // drop stale "root → child" label
    return tea.Batch(p.messages.ScrollToBottom(), sidebarCmd)
}
```

pkg/tui/components/messages/messages.go
```diff
@@ -1237,8 +1237,11 @@ func (m *model) AddShellOutputMessage(content string) tea.Cmd {
 	return m.addMessage(types.ShellOutput(content))
 }
 
-func (m *model) AddAssistantMessage() tea.Cmd {
-	return m.addMessage(types.Spinner())
+func (m *model) AddAssistantMessage(sender, label string) tea.Cmd {
```
docker-agent, line 1240: **[LOW] `AddAssistantMessage` appends a new spinner without evicting an existing one**

The updated `AddAssistantMessage` calls `addMessage` which appends to the message list without first calling `removeSpinner()`. If a spinner already exists (e.g., from a depth-1 `setPendingResponse(true)`) and a second `StreamStartedEvent` fires immediately (sub-agent, depth 2), the second `setPendingResponse(true)` → `AddAssistantMessage` call will **push a second spinner** rather than replace the first.

Only the tail is ever removed by `removeSpinner()`, so if two spinners accumulate, the depth-1 spinner becomes a phantom message that persists for the session.

In practice the parent usually emits content or a tool call before the sub-agent starts, replacing the first spinner; but this is not guaranteed by the runtime contract.

**Suggested safeguard:**
```go
func (m *model) AddAssistantMessage(sender, label string) tea.Cmd {
    m.removeSpinner() // evict any existing spinner before adding the new one
    if label == "" {
        return m.addMessage(types.Spinner())
    }
    return m.addMessage(types.SpinnerLabeled(sender, label))
}
```

pkg/tui/page/chat/runtime_events.go
```diff
@@ -238,6 +239,9 @@ func (p *chatPage) handleStreamStopped(msg *runtime.StreamStoppedEvent) tea.Cmd
...
+		p.agentStack = p.agentStack[:n-1]
```
docker-agent: **[MEDIUM] Asymmetric guards can cause `agentStack` and `streamDepth` to diverge on spurious `StreamStopped` events**

`streamDepth` and `agentStack` are popped with independent guards in `handleStreamStopped`:

```go
if p.streamDepth > 0 {
    p.streamDepth--
}
if n := len(p.agentStack); n > 0 {
    p.agentStack = p.agentStack[:n-1]
}
```

If a spurious or duplicate `StreamStoppedEvent` fires when `streamDepth` is already `0` (e.g. due to a runtime event replay or upstream bug), the `streamDepth` guard skips its decrement — but the `agentStack` guard still pops an entry because `len(agentStack) > 0`. After enough such events, `len(agentStack) < streamDepth`, and `pendingSpinnerContext()` returns stale or wrong `parent → child` labels.

**Suggested fix**: Couple the guards so they only fire together — for example, guard on `streamDepth > 0` and pop `agentStack` inside the same `if` block, with an optional `slog.Warn` when `streamDepth` and `len(agentStack)` are already mismatched:

```go
if p.streamDepth > 0 {
    p.streamDepth--
    if n := len(p.agentStack); n > 0 {
        p.agentStack = p.agentStack[:n-1]
    }
}
```

The cancel path (`p.streamDepth = 0; p.agentStack = nil`) always resets both correctly, so this only affects non-cancelled over-stop scenarios.

pkg/tui/components/tool/transfertask/transfertask.go
```diff
@@ -57,3 +58,19 @@ func render(msg *types.Message, _ spinner.Spinner, _ service.SessionStateReader,
...
+// error. The transfertask Base spinner is ModeSpinnerOnly, so s.View() is a
+// single 1-cell glyph whose width matches ✓/✗, keeping the task-text wrap math
+// stable (no elapsed-time suffix, unlike toolcommon.Icon).
+func statusIcon(msg *types.Message, s spinner.Spinner) string {
+	switch msg.ToolStatus {
+	case types.ToolStatusRunning, types.ToolStatusPending:
+		return styles.NoStyle.MarginLeft(2).Render(s.View())
+	case types.ToolStatusError:
+		return styles.ToolErrorIcon.Render("✗")
+	default: // Completed and any terminal/unknown state
```
docker-agent: **[LOW] `ToolStatusConfirmation` falls through to the ✓ (completed) glyph in `statusIcon`**

The `default` branch is commented as "Completed and any terminal/unknown state", but it also silently handles `ToolStatusConfirmation`:

```go
default: // Completed and any terminal/unknown state
    return styles.ToolCompletedIcon.Render("✓")
```

If a `transfer_task` call can ever enter `ToolStatusConfirmation` (awaiting user approval), its card would show ✓ — misleading the user into thinking the delegation finished when it is actually waiting for input.

If `ToolStatusConfirmation` is unreachable for `transfer_task` messages by design, adding an explicit `case types.ToolStatusCompleted:` and keeping `default` only for truly unknown states would make the intent clear and prevent silent mishandling of any future `ToolStatus` values added to the enum.

docker-agent (commented):

This PR adds two focused presentation changes to the TUI: a status-aware glyph on the `transfer_task` delegation card and a labeled `parent → child` pending spinner via a new `agentStack`. The implementation is clean and well-tested overall.

**1 MEDIUM** and **1 LOW** finding in the changed code:

- **[MEDIUM]** Asymmetric guards in `handleStreamStopped` can cause `agentStack` and `streamDepth` to diverge on spurious duplicate `StreamStopped` events, leading to stale `parent → child` spinner labels.
- **[LOW]** `ToolStatusConfirmation` silently falls through to the ✓ glyph in `statusIcon` — if that state is ever reachable for `transfer_task`, the card would show completed when it is actually awaiting confirmation.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27548968999).

dgageot (approved):

PR #3117, opened 2026-06-15
feat(tui): dedicated renderers for background-agent tools (#3103)

aheritier:
## Summary

The four background-agent tools (`run` / `list` / `view` / `stop_background_agent`) had no entry in the tool factory, so they fell through to `defaulttool` and rendered as raw JSON / text blobs. This gives them dedicated, presentation-only renderers so a dispatched fleet reads like delegations. Builds on #3102.

- **`run_background_agent`** → a delegation-style card `root dispatches worker` (accent badges) + the task text, led by the same status-aware fixed-width icon as the `transfer_task` card (#3101).
- **`view` / `stop_background_agent`** → `SimpleRendererWithResult` showing the task id (argument) + the formatted result.
- **`list_background_agents`** → shows the task listing. (The plan suggested `NoArgsRenderer`, but that drops the result — the only useful payload until a live surface lands — so the listing is preserved.)

No runtime change. The live "Background agents (N)" sidebar panel (Appendix C.2) is intentionally **out of scope** here and remains approval-gated.

## Screenshots

_`run_background_agent` fans out to several workers in parallel — each renders as a `root dispatches worker` card:_

phase3-dispatch

_`list_background_agents` shows the live fleet (three workers running concurrently):_

phase3-list

## Tests

- `pkg/tui/components/tool/backgroundagent` — `run` per-`ToolStatus` glyph + fixed-width invariant; `view`/`stop` show task id + result; `list` shows its result; ANSI-stripped.

`task build` / `task lint` clean; touched-package tests pass (same unrelated `teamloader` DMR failure as #3101).

## Stack

Top of the stack — stacked on **#3102** (← #3101). Tracking #3104.

Refs #3103


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR adds dedicated TUI renderers for the four background-agent tools (`run_background_agent`, `list_background_agents`, `view_background_agent`, `stop_background_agent`). The implementation correctly mirrors the `transfertask` sibling pattern, and factory registration is complete and correct — all four tool-name constants map to the right constructors.

**No high or medium severity bugs found.** The drafter identified five low-severity observations:

1. **`renderRun` uses `json.Unmarshal` directly** rather than the partial-JSON `toolcommon.ExtractField` path — streaming partial payloads produce an empty `params.Task` (card renders an orphaned icon). The `transfertask` sibling has the same limitation, so this is consistent with the codebase.
2. **`iconWidth` computed from `lipgloss.Width`** of an ANSI-styled string is correct today but would silently misalign wrapped-line indents if margin conventions changed.
3. **`renderList` intentionally surfaces error-status content** as a result, which is reasonable but asymmetric with `NoArgsRenderer`.
4. **`TestRunCard_StatusIcon` hard-codes spinner frame `"⠋"`**, creating a fragile coupling to spinner implementation details.
5. **`TestRunCard_IconWidthIsStable` does not assert `len(parts)==3`** before indexing `parts[len(parts)-1]`, so a degenerate render could silently measure the wrong column.

None of these warrant blocking the PR. Test coverage for status-icon transitions, icon-width stability, view/stop task-ID extraction, and list result visibility is solid.

pkg/tui/components/tool/backgroundagent/backgroundagent.go
```diff
@@ -0,0 +1,112 @@
...
+
+// NewRun renders run_background_agent as a delegation-style card:
+// "<sender> dispatches <agent>" plus the task text, led by a status-aware icon.
+func NewRun(msg *types.Message, sessionState service.SessionStateReader) layout.Model {
+	return toolcommon.NewBase(msg, sessionState, renderRun)
+}
+
+func renderRun(msg *types.Message, s spinner.Spinner, _ service.SessionStateReader, width, _ int) string {
+	var params agenttool.RunBackgroundAgentArgs
+	if err := json.Unmarshal([]byte(msg.ToolCall.Function.Arguments), &params); err != nil {
```
docker-agent: **[MEDIUM] `renderRun` uses raw `json.Unmarshal` instead of `toolcommon.ParseArgs`, rendering blank during argument streaming**

During LLM argument streaming, `msg.ToolCall.Function.Arguments` is emitted progressively (e.g., `{"agent":"res`). `json.Unmarshal` fails on partial JSON and the function returns `""`, so the delegation card is invisible until the arguments are fully received.

`toolcommon.ParseArgs[T]` (already in-repo) has a `tryFixPartialJSON` path that handles partial JSON — it can often extract fields like `agent` or `task` early enough to render a meaningful card during streaming.

The established `transfertask` renderer uses the same `json.Unmarshal` pattern, so this is consistent with the current codebase. However, since `toolcommon.ParseArgs` was specifically built to address this streaming gap, switching to it here (and backporting to `transfertask`) would improve the live-streaming UX for both delegation types.

```go
// current — blank until JSON is complete
if err := json.Unmarshal([]byte(msg.ToolCall.Function.Arguments), &params); err != nil {
    return ""
}

// suggested — partial cards during streaming
params, err := toolcommon.ParseArgs[agenttool.RunBackgroundAgentArgs](msg.ToolCall.Function.Arguments)
if err != nil {
    return ""
}
```

No behavioural regression risk; this is purely additive for streaming states.

docker-agent (commented):

One medium-severity finding in the new `backgroundagent` renderer package. The PR correctly mirrors the `transfertask` pattern and the factory registration is complete. The finding is a known streaming limitation that `toolcommon.ParseArgs` was designed to address.

dgageot:
## Code Review

Overall this looks well-scoped and solid — presentation-only renderers for the background-agent tools, following the existing `transfer_task`/delegation card patterns, with good test coverage. **Approve with minor nits.**

### Findings

**1. `ToolStatusConfirmation` rendered as completed** (minor correctness)
In `statusIcon`, the `default` case returns a green checkmark, so the non-terminal confirmation state would incorrectly render as success before the tool runs. Suggest handling it explicitly:
```go
case types.ToolStatusConfirmation:
    return styles.ToolPendingIcon.Render("?")
```

**2. `renderList` duplicates result-gating logic** (maintainability)
It manually gates the result on status, duplicating `toolcommon.SimpleRendererWithResult` (already used by `NewView`/`NewStop`). Could likely be simplified to match the others and keep gating centralized.

**3. `renderRun` silently returns empty on malformed args** (minor UX)
On `ParseArgs` failure it returns `""`, so genuinely malformed JSON makes the card disappear entirely. A minimal fallback (e.g. the tool name) would degrade more gracefully — acceptable if it matches existing `transfer_task` behavior.

**4. `HideToolResults` doesn't affect the run dispatch card** (likely intentional)
The run card always renders task/args, ignoring `HideToolResults()`. Consistent with the arguments-vs-results distinction, just worth confirming.

### Tests
Strong coverage: status icons, indentation/icon width stability, partial JSON streaming, task ID/result rendering, result gating, and ANSI stripping. Suggested addition: a test for `ToolStatusConfirmation`.

### Suggested follow-ups before merge
1. Handle `ToolStatusConfirmation` explicitly in `statusIcon`
2. Add a confirmation-status test case
3. (Optional) Simplify `renderList` via `toolcommon.SimpleRendererWithResult`
4. (Optional) Graceful fallback for malformed run args

PR #3116, opened 2026-06-15
feat(tui): bracket sub-agent delegations — return marker + sidebar breadcrumb (#3102)

aheritier:
## Summary

Makes nested / sequential delegations legible. Builds on #3101.

1. **Transcript return marker.** When a forwarded sub-agent (`transfer_task` / `run_skill`) hands control back, the only previous cue was the next parent badge, so a run of delegations read as one undifferentiated wall. A `↩ child → parent ───` divider (child's accent color) is now inserted when `AgentSwitching(false, child, parent)` fires. By design only the **exit** is bracketed — the existing `transfer_task` card remains the entry. Background agents and permanent handoffs never emit this event, so they're unaffected.
2. **Sidebar delegation breadcrumb.** When delegation depth > 1, a compact `root ▸ helper` breadcrumb renders under the Agents title (each name in its accent color, middle elided to width), backed by an `agentChain` tracked in lockstep with the sidebar's `sessionStack`. Click-zone alignment is preserved.

> The `+N background agents` count mentioned in the issue depends on the Phase 3 snapshot and is intentionally left as a `TODO(#3103)`.

## Screenshots

_Sidebar breadcrumb while `root` waits on `helper` (it disappears once control returns):_

phase2-breadcrumb

_Return marker when the sub-agent hands control back:_

phase2-return-marker

## Tests

- `message` — marker renders `↩ child → parent` and is static/cacheable (not spinner-driven).
- `messages` — `AddDelegationReturn` appends the marker.
- `page/chat` — `AgentSwitching{false}` inserts the marker; `{true}` / empty names do not.
- `sidebar` — `agentChain` stays balanced with `sessionStack` across push/pop/cancel/reset; breadcrumb renders only at depth > 1; click zones preserved.

`task build` / `task lint` clean; touched-package tests pass (same unrelated `teamloader` DMR failure as #3101).

## Stack

Stacked on **#3101**; base for **#3103**. Tracking #3104.

Refs #3102

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #3116 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #3115

PR #3118, opened 2026-06-15
feat(tui): live "Background agents (N)" sidebar panel

aheritier:
## Summary

Until now, a dispatched `run_background_agent` fleet was essentially invisible in the TUI: events are dropped by design (`runCollecting`), so the human only saw the initial "task started" line and whatever the model chose to poll. This adds a live, at-a-glance surface:

- **Runtime read-model.** `Handler.Snapshot() []TaskInfo` over the existing `tasks` concurrent map — strictly read-only, race-free, no new locks, never exposes `*task`, and **no change to execution / event-drop semantics**.
- **App→TUI accessor** via an optional interface, so a remote/client runtime that can't provide it simply reports none.
- **~1s poll** (`tea.Tick`) that runs while a stream is working **or** any task is still running (background tasks outlive the turn), and self-stops when idle.
- **Sidebar panel** `Background agents (N)`: one row per running task — `● <agent>  <elapsed>` (accent dot + name, right-aligned elapsed), hidden entirely when nothing runs.
- Fills the Phase-2 `+N background` breadcrumb count (the `TODO(#3103)` left in #3102).

### Why polling, not event-push
The background goroutines outlive the per-turn event sink (that's *why* events are dropped today), so there's no live stream to forward safely. Polling a lock-safe snapshot is simple, self-limiting (stops when empty), and adds no runtime coupling.

## Screenshots

c2-panel-running

_`Background agents (3)` panel while three researchers run in parallel — plus the Phase-2 breadcrumb gaining `+3 background` during a concurrent `transfer_task`:_

c2-panel-and-breadcrumb

_The panel is live: the count drains (3 → 1) and the section disappears as tasks finish:_

c2-panel-drain

## Tests

- `pkg/tools/builtin/agent` — `Snapshot()` table test (running/completed/stopped/empty); passes `go test -race`.
- `pkg/tui/page/chat` — poll idempotency, continue/stop lifecycle, running-task detection.
- `pkg/tui/components/sidebar` — panel render/hide, filter/sort/drain, `+N background`.

`task build` / `task lint` clean; touched-package tests green (only the unrelated `pkg/teamloader/TestLoadExamples` DMR 416 fails). Cleared local review (incl. a `-race` pass on the snapshot).

## Deviations (deliberate)
1. Rows omit a literal "running" word — only running tasks are listed, and the colored `●` conveys state.
2. Optional completion notice (C.2.3) skipped to keep the diff tight; trivial to add.
3. `+N background` is appended only to the depth>1 delegation breadcrumb; the standalone panel covers the no-delegation case.

## Stack
4th in the delegation-readability stack, stacked on **#3117** (Phase 3 C.1): #3115 ← #3116 ← #3117 ← this. Tracking #3104. C.3 (depth indentation) remains an out-of-scope stretch.

Refs #3103


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR adds a live **"Background agents (N)"** sidebar panel with a ~1s poll loop, a race-free `Snapshot()` read-model, and the Phase-2 breadcrumb count (`+N background`). The implementation is solid — no CONFIRMED or LIKELY bugs were found.

**Areas reviewed:**

- **`Snapshot()` race safety** — Fields read are immutable after task creation; atomic load for status. Race-free by design, confirmed clean. ✅
- **Poll lifecycle (`startBackgroundAgentPoll` / `handleBackgroundAgentPoll`)** — The `backgroundPollActive` guard works correctly in Bubbletea's serialised Elm loop. Each `tea.Tick` fires exactly once; the loop is strictly sequential (no stale in-flight ticks possible). ✅
- **`SetBackgroundAgents` no-op guard** — Correctly skips cache invalidation only when the panel was already empty before the call; drains properly when tasks finish. ✅
- **`delegationBreadcrumb` width arithmetic** — The `avail <= 0` guard fires before any `ansi.Truncate` call; no negative-width truncation possible. ✅
- **Optional `Snapshoter` interface assertion** (`app.go`) — Safe two-value type assertion (`l, ok := ...`), no panic path. ✅
- **Sidebar rendering** — `backgroundAgentsSection` gap calculation uses `lipgloss.Width` which correctly strips ANSI; layout is correct. ✅
- **`FormatDuration` with negative durations** — Clock skew could produce a display glitch (`-1s`) but no crash; acceptable for a live elapsed timer. ✅

No action required — the implementation matches the design intent described in the PR and the Phase 3 C.2 spec.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27543387890).

PR #3120, opened 2026-06-15
Better worktree handling

dgageot:
+ Create worktree sooner
+ Add documentation
+ Reopen sessions in their worktree
+ Let the user choose a worktree base

### DIFF
--- a/cmd/root/backend.go
+++ b/cmd/root/backend.go
@@ -138,11 +138,13 @@ func (b *localBackend) Spawner(rt runtime.Runtime) tui.SessionSpawner {
 // ResumeWorkingDir looks up the session named by --session and returns the
 // working directory it was created with. It opens (and shares) the same
 // session store CreateSession uses, so this peek does not pay for a second
-// connection. Any lookup miss — no --session, a relative ref, an unknown ID,
-// an empty stored dir, or a dir that no longer exists — returns ok=false and
-// leaves the run to its normal working-directory resolution.
+// connection. Both explicit IDs and relative refs (e.g. "-1" for the last
+// session) are honoured, so resuming the previous run reattaches to the
+// worktree it ran in. Any lookup miss — no --session, an unknown ID, an empty
+// stored dir, or a dir that no longer exists — returns ok=false and leaves the
+// run to its normal working-directory resolution.
 func (b *localBackend) ResumeWorkingDir(ctx context.Context) (string, bool) {
-	if b.flags.sessionID == "" || session.IsRelativeSessionRef(b.flags.sessionID) {
+	if b.flags.sessionID == "" {
 		return "", false
 	}
 
@@ -151,6 +153,8 @@ func (b *localBackend) ResumeWorkingDir(ctx context.Context) (string, bool) {
 		return "", false
 	}
 
+	// Relative refs (-1, -2, ...) resolve against the shared store here, which
+	// is already open at this point, so they reattach to their worktree too.
 	resolvedID, err := session.ResolveSessionID(ctx, store, b.flags.sessionID)
 	if err != nil {
 		return "", false
--- a/cmd/root/backend.go
+++ b/cmd/root/backend.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"os"
 	"sync"
 
 	"github.com/docker/docker-agent/pkg/config"
@@ -27,6 +28,14 @@ type backend interface {
 
 	CreateSessionRequest(workingDir string) runtime.CreateSessionRequest
 	CreateSession(ctx context.Context, loaded *teamloader.LoadResult, req runtime.CreateSessionRequest) (runtime.Runtime, *session.Session, func(), error)
+
+	// ResumeWorkingDir returns the working directory stored on the session
+	// this run would resume (--session naming an existing session), so a
+	// resumed run can reattach to the worktree it was created in without the
+	// caller re-passing --worktree (which would fail since the worktree
+	// already exists). Returns ok=false when there is no resumable session,
+	// it has no stored working directory, or that directory no longer exists.
+	ResumeWorkingDir(ctx context.Context) (dir string, ok bool)
 
 	Spawner(rt runtime.Runtime) tui.SessionSpawner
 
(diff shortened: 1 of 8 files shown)


docker-agent (commented):

3 findings: 1 LIKELY HIGH, 2 MEDIUM (1 CONFIRMED, 1 LIKELY)

pkg/worktree/worktree.go
```diff
@@ -278,6 +322,38 @@ func repoRoot(ctx context.Context, dir string) (string, error) {
...
+// A fetch failure is reported as [ErrInvalidBase].
+func fetchBase(ctx context.Context, root, base string) error {
+	remote, branch, ok := strings.Cut(base, "/")
+	if !ok {
+		return nil
+	}
+	if !isRemote(ctx, root, remote) {
+		return nil
+	}
+	if err := git(ctx, root, "fetch", remote, branch); err != nil {
```
docker-agent: **[LIKELY HIGH] `fetchBase`: plain `git fetch` may not update the remote-tracking ref used by `git worktree add`**

`fetchBase` runs `git fetch <remote> <branch>` (no explicit refspec). While modern git with a standard `+refs/heads/*:refs/remotes/origin/*` fetch refspec *does* propagate the update to `refs/remotes/origin/main` as a side-effect, this is not guaranteed for all git configurations (e.g., `--no-tags`, `tagOpt`, or custom refspecs). In non-standard configurations, `git fetch origin main` will write to `FETCH_HEAD` only; the subsequent `git worktree add -b <branch> <dest> origin/main` then resolves `origin/main` from the stale `refs/remotes/origin/main`, defeating the stated purpose of this function.

To make the behaviour explicit and configuration-independent, use a full refspec:

```go
git(ctx, root, "fetch", remote, branch+":refs/remotes/"+remote+"/"+branch)
```

This guarantees `refs/remotes/origin/main` is updated regardless of remote configuration.

cmd/root/backend.go
```diff
@@ -126,6 +135,36 @@ func (b *localBackend) Spawner(rt runtime.Runtime) tui.SessionSpawner {
...
 }
 
+// ResumeWorkingDir looks up the session named by --session and returns the
+// working directory it was created with. It opens (and shares) the same
+// session store CreateSession uses, so this peek does not pay for a second
+// connection. Any lookup miss — no --session, a relative ref, an unknown ID,
+// an empty stored dir, or a dir that no longer exists — returns ok=false and
+// leaves the run to its normal working-directory resolution.
+func (b *localBackend) ResumeWorkingDir(ctx context.Context) (string, bool) {
+	if b.flags.sessionID == "" || session.IsRelativeSessionRef(b.flags.sessionID) {
```
docker-agent: **[LIKELY MEDIUM] Relative session refs (e.g. `--session -1`) silently skip worktree reattachment**

`ResumeWorkingDir` returns `ok=false` immediately when `session.IsRelativeSessionRef(b.flags.sessionID)` is true. This means a user running:

```
docker agent run agent.yaml --session -1
```

will not be reattached to the git worktree that the previous session ran in — even though `-1` is a documented way to resume the last session. The session itself resumes, but the working directory defaults to the current directory instead of the saved worktree path.

The restriction exists because relative refs must be resolved against the session store, which isn't open yet at this point. However, the comment only documents this as an accepted miss without exploring whether the resolved absolute ID could be used for a follow-up reattachment. Consider resolving the relative ref first (using the already-shared store) before giving up:

```go
if b.flags.sessionID == "" {
    return "", false
}
// Resolve relative refs before deciding to skip
```

At minimum, this limitation should be documented in the CLI's `--session` flag help text so users know they must provide an absolute session ID to get worktree reattachment.

cmd/root/run.go
```diff
@@ -458,6 +490,39 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
...
+		wd = createdWorktree.Dir
+		f.runConfig.WorkingDir = createdWorktree.Dir
+	}
+
+	loadResult, err := b.LoadTeam(ctx, b.LoadTeamRequest())
+	if err != nil {
+		// The worktree was created before the load; tear it down so a load
+		// failure doesn't leave an orphaned worktree behind.
+		if createdWorktree != nil {
+			if rmErr := createdWorktree.Remove(ctx); rmErr != nil {
```
docker-agent: **[CONFIRMED MEDIUM] Worktree cleanup after `LoadTeam` failure uses a potentially-cancelled context, leaving orphaned worktrees**

When `LoadTeam` returns an error because the user pressed Ctrl-C (causing `ctx` to be cancelled), the cleanup path calls `createdWorktree.Remove(ctx)` with that already-cancelled context. Any git subprocess spawned by `Remove` will be immediately killed, so the worktree removal silently fails and the branch is left orphaned.

The same file already demonstrates the correct pattern later:

```go
// Correct — uses a fresh context immune to cancellation
f.cleanupWorktree(context.WithoutCancel(ctx), out, createdWorktree)
```

Apply the same fix here:

```go
if rmErr := createdWorktree.Remove(context.WithoutCancel(ctx)); rmErr != nil {
    slog.WarnContext(ctx, "Failed to remove worktree after load error", ...)
}
```

dgageot:
Addressed the review feedback (3 new commits, branch already on top of latest `main`):

- **`fetchBase` could branch from a stale ref** — switched to a fully-qualified refspec (`refs/heads/<branch>:refs/remotes/<remote>/<branch>`) so the remote-tracking ref is updated regardless of the repo's fetch configuration (custom refspecs, `--no-tags`, `tagOpt`, …). A bare `git fetch <remote> <branch>` only guarantees `FETCH_HEAD`. Added a regression test that drops `remote.origin.fetch` and verifies the worktree still starts at the latest remote commit.
- **Relative `--session` refs skipped worktree reattachment** — `ResumeWorkingDir` now resolves relative refs (e.g. `--session -1`) against the already-open shared store instead of bailing out, so resuming the last session reattaches to its worktree just like an explicit ID does.
- **Worktree cleanup used a cancelled context** — the post-load-failure teardown now uses `context.WithoutCancel(ctx)` so a Ctrl-C-cancelled `LoadTeam` doesn't immediately kill the `git worktree remove` subprocess and orphan the worktree, matching the interactive-exit cleanup path.

`task lint` and `task test` both pass.

gtardif (approved):

PR #3122, opened 2026-06-15
fix(agent): deduplicate tool names before dispatch

### LINKED ISSUE #2251: Duplicate tool names across toolsets cause Anthropic 400 errors
## Description

When an agent has multiple toolsets that produce tools with the same name, the Anthropic API rejects the request with HTTP 400: `"tools: Tool names must be unique."`. There is no deduplication anywhere in the tool collection pipeline.

Issue #1969 fixed this for the **specific case of multiple LSP toolsets** (via the LSP multiplexer in `teamloader.go`), but the general problem remains for all other toolset combinations.

**Observed in session:** `cca99566-4277-457a-92fe-cfd877825c0b`

**Error:**
```
all models failed: error receiving from stream: HTTP 400: POST
"https://ai-backend-service-stage.docker.com/proxy/v1/messages?beta=true": 400
Bad Request {"type":"error","error":{"type":"invalid_request_error",
"message":"tools: Tool names must be unique."}}
```

## Steps to Reproduce

**Option 1 — Duplicate built-in toolsets:**

```yaml
agents:
  root:
    model: anthropic/claude-sonnet-4-0
    instruction: You are a helpful assistant.
    toolsets:
      - type: filesystem
      - type: filesystem
```

Both filesystem toolsets produce identical tool names (`read_file`, `write_file`, `edit_file`, `directory_tree`, etc.).

**Option 2 — Two unnamed MCP servers exposing the same tool names:**

```yaml
agents:
  root:
    model: anthropic/claude-sonnet-4-0
    instruction: You are a helpful assistant.
    toolsets:
      - type: mcp
        command: npx
        args: ["-y", "@modelcontextprotocol/server-filesystem", "/tmp"]
      - type: mcp
        command: npx
        args: ["-y", "@modelcontextprotocol/server-filesystem", "/home"]
```

Both MCP servers expose identically-named tools and neither has a `name` prefix configured.

**Option 3 — MCP server colliding with a built-in:**

```yaml
agents:
  root:
    model: anthropic/claude-sonnet-4-0
    instruction: You are a helpful assistant.
    toolsets:
      - type: fetch
      - type: mcp
        command: npx
        args: ["-y", "@modelcontextprotocol/server-fetch"]
```

The built-in `fetch` toolset and the MCP fetch server both produce a tool named `fetch`.

## Root Cause

`Agent.Tools()` in `pkg/agent/agent.go` collects tools from all toolsets via simple `append` with no deduplication. No downstream code checks for uniqueness either:

1. `Agent.Tools()` (`pkg/agent/agent.go`) — appends from all toolsets, no dedup
2. `getTools()` (`pkg/runtime/loop.go`) — passes tools through as-is
3. `convertTools()` (`pkg/model/provider/anthropic/client.go`) — maps 1:1 to API params, no dedup
4. Anthropic API rejects the request

### Scenarios that produce duplicates

- Two MCP servers without a `name` prefix (or with the same prefix) exposing same-named tools
- An MCP server returning a tool that collides with a built-in name
- Same toolset type configured twice in agent config
- Dynamic MCP `ToolListChanged` notifications adding tools that collide with existing names

## Suggested Fix

Add deduplication in `Agent.Tools()` — the single collection point where all tools converge before being sent to any provider. Keep the first occurrence of each tool name and log a warning for dropped duplicates.

Sayt-0:
## What

Deduplicate tool names in `Agent.Tools()` so an agent never sends duplicate tool names to a provider.

Fixes #2251.

## Why

When an agent has multiple toolsets that emit the same tool name, `Agent.Tools()` (via `collectTools`) returned every copy verbatim — there was no dedup anywhere in the collection pipeline. The Anthropic API rejects such a request:

```
HTTP 400: {"type":"error","error":{"type":"invalid_request_error",
"message":"tools: Tool names must be unique."}}
```

Collisions happen in several real configs:
- the same built-in toolset type configured twice (e.g. two `filesystem` toolsets → `read_file`, `write_file`, …);
- two unnamed MCP servers exposing the same tool names;
- an MCP server whose tool name shadows a built-in (e.g. `fetch`).

#1969 fixed only the specific multi-LSP case via the LSP multiplexer; the general problem remained.

## How

`collectTools()` is the single point where tools from every toolset plus static tools converge before being handed to any provider, so dedup lives there. `dedupeToolsByName` keeps the **first** occurrence of each name, drops later collisions, and logs a `slog.Warn` per dropped tool. Deduplication is done in place on the locally-built slice, so a collision-free list is returned unchanged (and an empty list stays `nil`).

Warnings go through `slog.Warn` rather than the user-facing `AddToolWarning`: `collectTools()` runs every turn and warnings are drained per turn, so surfacing this to the user would repeat on every message.

## Tests

- `TestAgentTools`: new table case — two toolsets emitting the same names collapse to a unique set with no user-facing warnings.
- `TestAgentToolsDedupeKeepsFirstInOrder`: asserts the first occurrence wins and distinct names keep their first-seen order.

## Verification

Reproduced the root cause with a config of two `filesystem` toolsets:

- before: `Agent.Tools()` returned 18 tools / 9 distinct names (all duplicated);
- after: 9 tools / 9 distinct names, with a logged warning per dropped duplicate.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -377,11 +377,37 @@ func (a *Agent) collectTools(ctx context.Context) ([]tools.Tool, error) {
 
 	agentTools = append(agentTools, a.tools...)
 
+	// Collapse duplicate tool names to their first occurrence. Two toolsets of
+	// the same type, two unnamed MCP servers, or an MCP server whose tool name
+	// shadows a built-in can all surface the same name; providers such as
+	// Anthropic reject a request whose tool list contains duplicates ("tools:
+	// Tool names must be unique."). This is the single point where every tool
+	// converges before being handed to any provider. See #2251.
+	agentTools = a.dedupeToolsByName(ctx, agentTools)
+
 	if a.addDescriptionParameter {
 		agentTools = tools.AddDescriptionParameter(agentTools)
 	}
 
 	return agentTools, nil
+}
+
+// dedupeToolsByName keeps the first tool seen for each name and drops any later
+// collision, logging a warning per dropped tool. Deduplication happens in place
+// (the input slice is local to collectTools), so a collision-free list is
+// returned unchanged.
+func (a *Agent) dedupeToolsByName(ctx context.Context, agentTools []tools.Tool) []tools.Tool {
+	seen := make(map[string]struct{}, len(agentTools))
+	deduped := agentTools[:0]
+	for _, tool := range agentTools {
+		if _, ok := seen[tool.Name]; ok {
+			slog.WarnContext(ctx, "Dropping duplicate tool name from agent tool set", "agent", a.Name(), "tool", tool.Name)
+			continue
+		}
+		seen[tool.Name] = struct{}{}
+		deduped = append(deduped, tool)
+	}
+	return deduped
 }
 
 func (a *Agent) ToolSets() []tools.ToolSet {
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -149,6 +149,19 @@ func TestAgentTools(t *testing.T) {
 			wantToolCount: 0,
 			wantWarnings:  0,
 		},
+		{
+			// #2251: two toolsets emitting the same names (e.g. the same
+			// built-in type configured twice) must collapse to a unique set,
+			// otherwise Anthropic rejects the request with "Tool names must be
+			// unique." Deduplication is silent (slog only), so no warnings.
+			name: "duplicate names across toolsets are deduped",
+			toolsets: []tools.ToolSet{
+				newStubToolSet(nil, []tools.Tool{{Name: "read_file", Parameters: map[string]any{}}, {Name: "write_file", Parameters: map[string]any{}}}, nil),
+				newStubToolSet(nil, []tools.Tool{{Name: "read_file", Parameters: map[string]any{}}, {Name: "write_file", Parameters: map[string]any{}}}, nil),
+			},
+			wantToolCount: 2,
+			wantWarnings:  0,
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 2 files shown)


pkg/agent/agent.go
```diff
@@ -377,13 +377,39 @@ func (a *Agent) collectTools(ctx context.Context) ([]tools.Tool, error) {
...
+
 	if a.addDescriptionParameter {
 		agentTools = tools.AddDescriptionParameter(agentTools)
 	}
 
 	return agentTools, nil
 }
 
+// dedupeToolsByName keeps the first tool seen for each name and drops any later
+// collision, logging a warning per dropped tool. Deduplication happens in place
```
docker-agent, line 396: **[LOW] Fragile in-place slice reuse aliases the input backing array**

`deduped` is initialised as `agentTools[:0]`, which reuses the same backing array as the input slice. As elements are appended into `deduped`, they overwrite earlier positions in `agentTools`. This is safe today because `agentTools` is always a locally-constructed slice in `collectTools`, but the function signature offers no enforcement of that precondition. A future refactor that passes a slice backed by a shared array (e.g. a slice of `a.tools`) could silently corrupt the original.

Consider using a fresh allocation instead:
```go
deduped := make([]tools.Tool, 0, len(agentTools))
```
This removes the aliasing concern at negligible cost.

pkg/agent/agent_test.go
```diff
@@ -149,6 +149,19 @@ func TestAgentTools(t *testing.T) {
 			wantToolCount: 0,
 			wantWarnings:  0,
 		},
+		{
+			// #2251: two toolsets emitting the same names (e.g. the same
+			// built-in type configured twice) must collapse to a unique set,
+			// otherwise Anthropic rejects the request with "Tool names must be
+			// unique." Deduplication is silent (slog only), so no warnings.
+			name: "duplicate names across toolsets are deduped",
```
docker-agent, line 157: **[LOW] Dedup test does not verify that slog.Warn was actually emitted for dropped tools**

The test correctly asserts `wantWarnings: 0` (user-facing `AddToolWarning` is not called), but there is no assertion that `slog.WarnContext` fired for the two dropped tools. If the warn call were accidentally removed, the test would still pass, leaving the logging contract documented in the PR description untested.

Consider capturing the slog output (e.g. via `slog.New` with a test handler) and asserting that two warn records with the dropped tool names were emitted.

pkg/agent/agent_test.go
```diff
@@ -169,6 +182,38 @@ func TestAgentTools(t *testing.T) {
...
+
+	names := make([]string, len(got))
+	for i, tool := range got {
+		names[i] = tool.Name
+	}
+	assert.Equal(t, []string{"read_file", "fetch", "write_file"}, names,
+		"distinct names in first-seen order, no duplicates")
+
+	require.Len(t, got, 3)
+	assert.Equal(t, "from first", got[0].Description,
```
docker-agent, line 213: **[LOW] require.Len guard placed after assert.Equal and slice index access**

`assert.Equal(t, []string{...}, names, ...)` and the `got[0].Description` index read both appear before `require.Len(t, got, 3)`. If `got` has fewer than 3 elements, the safer pattern is to check the length first — a failing `require.Len` will abort the test immediately and prevent a potential panic on the index expression, as well as making the root cause (wrong length) the first diagnostic shown.

Reorder to `require.Len` first, then the `assert.Equal` and field access.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The deduplication logic in `dedupeToolsByName` is correct: it uses a seen-map, preserves insertion order, keeps the first occurrence, and routes warnings through `slog` rather than the user-facing warning queue. No high or medium severity issues were found. Three low-severity observations are left as inline comments (fragile backing-array alias, untested slog contract, and a minor test-guard ordering issue).

rumpl:
This seems wrong, we can't just remove tools like this, for example the jira toolset has a fetch tool, I have no idea what it does but it conflicts with our fetch tool

Sayt-0:
Ah yeah good catch, I hadn't thought about two different tools just sharing a name.Maybe I'll try prefixing the colliding ones instead so both stay usable, wdyt?

rumpl:
I would need to check, the issue I have with renaming tool is that an MCP server can also inject instructions into the agent's system prompt, and in those instructions it can say things like: "to do this, use the fecth tool", in that case the agent will use the wrong fetch tool if we rename it to "jira_fetch".

I don't really have a good solution for this

Sayt-0:
Looked into it more and I think I don't have a clean fix that covers every case too.

PR #3125, opened 2026-06-15
fix: run /shell in the active session's working directory

dgageot:
When running with `--worktree`, the current directory is changed for tools but the `/shell` slash command was still inheriting the process CWD. This meant launching an interactive shell via `/shell` would start in the original directory rather than the working directory where tools operate.

This fix updates the `startShell` handler in `pkg/tui/handlers.go` to set `cmd.Dir` to the active session's working directory (`runner.WorkingDir`). Now the interactive shell launched by `/shell` starts in the same directory as the tools, such as a worktree created by `--worktree`.

### DIFF
--- a/pkg/tui/handlers.go
+++ b/pkg/tui/handlers.go
@@ -785,6 +785,12 @@ func (m *appModel) startShell() (tea.Model, tea.Cmd) {
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	// Run the shell in the active session's working directory so it matches
+	// where the tools operate (e.g. the worktree created by --worktree),
+	// rather than inheriting the process CWD.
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Small, focused fix that correctly sets `cmd.Dir` to the active session's working directory when launching an interactive shell via `/shell`. The logic is straightforward and well-guarded with a nil and empty-string check.

Two minor edge-case observations (no action required):
- If `WorkingDir` is ever a relative path, it resolves against the process CWD rather than the worktree root — not a current issue but worth keeping in mind if the field is populated in new ways.
- No existence check before assigning `cmd.Dir`: if the worktree directory was deleted after the session started, `tea.ExecProcess` will fail silently (callback is `nil`). Acceptable for now given the use case.

gtardif (approved):

Issue #3123, opened 2026-06-15
TUI: right-panel (sidebar todo list) scrolling lags/hangs during active streaming when the list is long


aheritier:
## Summary
When the right-hand sidebar is long because it is rendering a large todo list, scrolling it becomes laggy/janky/unresponsive while the left-hand chat transcript is actively streaming (agent is "working"). Short lists are unaffected; the longer the todo list, the worse the lag. Scrolling becomes smooth again once the turn ends.

## Environment
- Component: TUI (`docker-agent run`, default Bubble Tea v2 TUI)
- Layout: wide window, vertical sidebar on the right (`sidebarVertical` mode)
- Repro evidence: local screen recording (2026-06-15)

## Steps to reproduce
1. Run an agent whose config enables the builtin `todo` toolset.
2. Have the agent create a large todo list (~50–150 items; long descriptions amplify the effect) via `create_todos`.
3. Keep the left panel streaming (a long assistant response, or a long-running tool whose output streams) so the working spinner is animating.
4. Maximize terminal height so the todo list overflows the viewport (forces the scrollbar / two-pass path).
5. Scroll the sidebar with the mouse wheel and by dragging the scrollbar.

**Expected:** sidebar scrolling stays smooth regardless of streaming activity.
**Actual:** scrolling lags/janks; worse the longer the list; smooth again once the agent stops working.

## Root cause (verified against current code)
The sidebar render cache is **defeated during streaming**. The single global animation coordinator emits a 14 FPS tick while any spinner is active, and the sidebar invalidates its *entire* render cache on every tick during a working turn — then fully rebuilds all sections (including the whole todo list) **twice** when the list overflows (two-pass scrollbar rendering). All of this runs on Bubble Tea's single `Update`/`View` goroutine, so coalesced scroll deltas only surface after each full rebuild lands.

Key evidence (`file:line`):
- 14 FPS tick: `pkg/tui/animation/coordinator.go` `tickLocked()` → `tea.Tick(time.Second/14, …)`
- Tick invalidates the whole sidebar cache while working: `pkg/tui/components/sidebar/sidebar.go:810-833` (`default:` branch — `m.workingAgent != "" …` → `needsInvalidate` → `invalidateCache()` at `:229`, which sets `cacheDirty = true`)
- Two-pass O(n) rebuild on overflow: `sidebar.go:921-945` (`verticalView()` calls `renderSections()` again when `totalLines > m.height`); `renderSections` at `:966`; todo render at `:986-987`
[...]

PR #3126, opened 2026-06-15
fix(tui): cache sidebar todo render to stop scroll lag while streaming

Sayt-0:
## What
The right-hand sidebar (todo list) scrolls laggily while the left chat panel is streaming, and the longer the todo list the worse it gets (#3123). It's smooth again once the turn ends.

## Why
While the agent works, the sidebar invalidates its whole render cache on every 14 FPS spinner tick and rebuilds every section, including the entire todo list, twice when it overflows (two-pass scrollbar layout). The todo render does O(n) word-wrap + styling per item with no memoization, so the per-frame cost scales with the list length and contends with scroll handling on the single render goroutine. The left (messages) panel already caches/virtualizes its build stage; the sidebar didn't.

## Fix
Memoize the todo render in the todo sidebar component, keyed by content width so it survives the two-pass width toggle, cleared when the todos change or the theme changes. Bounded to a few widths so a window resize can't grow it unbounded. Small, contained change in `todotool/sidebar.go` plus a one-line theme invalidation hook in `sidebar.go`.

## Before / after
Per-frame sidebar rebuild while streaming (`BenchmarkSidebarStreamingTodos`):

| todos | before (ms/frame) | after (ms/frame) | speedup |
|------:|------------------:|-----------------:|--------:|
| 10    | 1.05              | 0.36             | ~3x     |
| 50    | 3.65              | 0.36             | ~10x    |
| 150   | 10.19             | 0.36             | ~28x    |

At 150 todos, allocations drop too, and the cost stops scaling with the list:

| metric (150 todos) | before  | after  |
|--------------------|--------:|-------:|
| time / frame       | 10.19ms | 0.36ms |
| bytes / frame      | 7.5 MB  | 159 KB |
| allocs / frame     | 140855  | 2252   |
| scaling 150-vs-10  | ~10x    | ~1x (flat) |

## Tests
- new `BenchmarkSidebarStreamingTodos` (sidebar pkg) measuring the per-frame rebuild across todo counts
- new unit tests for the cache (todotool pkg): reflects todo changes, keyed by width, invalidates on theme, stays bounded
- full `pkg/tui/...` suite is green, gofmt clean

### DIFF
--- a/pkg/tui/components/sidebar/bench_test.go
+++ b/pkg/tui/components/sidebar/bench_test.go
@@ -1,0 +1,54 @@
+package sidebar
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/docker/docker-agent/pkg/session"
+	"github.com/docker/docker-agent/pkg/tools"
+	"github.com/docker/docker-agent/pkg/tools/builtin/todo"
+	"github.com/docker/docker-agent/pkg/tui/service"
+)
+
+// makeTodos builds n todo items whose descriptions are long enough to exercise
+// the word-wrap path in the sidebar's narrow content column.
+func makeTodos(n int) *tools.ToolCallResult {
+	statuses := []string{"completed", "in-progress", "pending"}
+	items := make([]todo.Todo, n)
+	for i := range items {
+		items[i] = todo.Todo{
+			ID:          fmt.Sprintf("t%d", i),
+			Status:      statuses[i%len(statuses)],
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

No bugs found in the changed code. The todo render cache implementation looks correct:
- Cache is properly keyed by content width (survives two-pass layout)
- Bounded map size prevents unbounded growth on window resize
- Cache is correctly cleared on todo content changes and theme changes
- Tests cover width-keying, invalidation on change, theme invalidation, and size bounds

dgageot (approved):

PR #3127, opened 2026-06-15
Secondary, non-alternate screen ui

rumpl:
I want a non-alternate screen UI, this is a first, not complete, implementation of a leaner ui for folks that don't need all the bells and whistles of our TUI

### DIFF
--- a/pkg/leantui/renderer.go
+++ b/pkg/leantui/renderer.go
@@ -1,0 +1,263 @@
+package leantui
+
+import (
+	"bufio"
+	"strings"
+
+	"github.com/charmbracelet/x/ansi"
+)
+
+const (
+	seqHideCursor            = "\x1b[?25l"
+	seqShowCursor            = "\x1b[?25h"
+	seqSyncStart             = "\x1b[?2026h"
+	seqSyncEnd               = "\x1b[?2026l"
+	seqEraseLine             = "\x1b[2K"
+	seqClearScreen           = "\x1b[2J\x1b[H\x1b[3J"
+	seqEnableBracketedPaste  = "\x1b[?2004h"
+	seqDisableBracketedPaste = "\x1b[?2004l"
+)
+
+// renderer draws the whole conversation as a single, growing array of lines and
+// keeps it in sync with the terminal using minimal, differential updates. It
+// writes to the normal screen buffer (never the alternate screen): content that
+// scrolls off the top becomes immutable terminal scrollback, exactly like a
+// regular program's output, while the bottom of the array (the input box and
+// status footer) is rewritten in place.
+//
+// The model mirrors how Claude Code / pi render: each frame the controller
+// produces the full set of lines; the renderer diffs them against the previous
+// frame and rewrites only the changed rows within the visible viewport, letting
+// the terminal scroll naturally when content is appended past the bottom.
+type renderer struct {
+	w *bufio.Writer
+
+	width  int
+	height int
+
+	prev        []string
+	viewportTop int // index in prev of the topmost visible row
+	cursorRow   int // buffer row the hardware cursor currently sits on
+	initialized bool
+	needsRedraw bool
+}
+
+func newRenderer(w *bufio.Writer, width, height int) *renderer {
+	return &renderer{w: w, width: width, height: height}
+}
+
+// setSize records a new terminal size and forces a clean repaint on the next
+// frame, since wrapping and the viewport both change with the dimensions.
+func (r *renderer) setSize(width, height int) {
+	r.width = width
+	r.height = height
+	r.needsRedraw = true
+}
+
+// repaint forces the next frame to clear the screen and redraw from scratch.
+func (r *renderer) repaint() {
+	r.needsRedraw = true
+}
+
+// frame reconciles the screen with newLines and places the hardware cursor at
+// (cursorLine, cursorCol), where cursorLine is an index into newLines.
+func (r *renderer) frame(newLines []string, cursorLine, cursorCol int) {
+	if len(newLines) == 0 {
+		newLines = []string{""}
+	}
+
+	if !r.initialized || r.needsRedraw {
+		r.fullRedraw(newLines, cursorLine, cursorCol, r.initialized)
+		r.initialized = true
+		r.needsRedraw = false
+		return
+	}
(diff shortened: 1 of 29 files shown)


docker-agent (commented):

This PR introduces a well-structured `pkg/leantui` package for a non-alternate-screen terminal UI. The code is generally clean, has good test coverage, and handles terminal raw mode, resize events, and incremental rendering. Two medium-severity issues were found in the changed code and warrant attention before merge.

**Summary of findings (2 LIKELY, medium severity):**
- `pkg/leantui/renderer.go` — Stale-row erasure inflates the internal cursor position counter (`cur`) beyond the viewport bottom, which can cause `moveCursor` to emit oversized cursor-up sequences and corrupt the display when content shrinks.
- `pkg/leantui/update.go` — `startCompact` does not pass the `cancel` function to `CompactSession`, unlike `startRun` and `startSkillFork`. The App cannot self-signal completion, which may leave `m.busy = true` and the UI locked in a spinner state after compaction finishes.

*Reviewed by cagent · 2 inline comment(s)*

pkg/leantui/renderer.go
```diff
@@ -0,0 +1,257 @@
...
+		if i > first {
+			b.WriteString("\r\n")
+		}
+		b.WriteString(seqEraseLine)
+		b.WriteString(newLines[i])
+	}
+	cur = renderEnd
+
+	// Clear stale trailing rows that remain when content shrank.
+	for i := len(newLines); i < len(r.prev); i++ {
```
docker-agent, line 127: **[MEDIUM] Stale-row erasure inflates `cur` beyond viewport, corrupting cursor position**

When content shrinks (e.g., a tool output block is compacted or replaced), the stale-trailing-rows erasure loop increments `cur` for every erased row:

```go
cur = renderEnd                         // last newly-written row

for i := len(newLines); i < len(r.prev); i++ {
    b.WriteString("\r\n")
    b.WriteString(seqEraseLine)
    cur++                               // cursor moves down with each \r\n
}

r.viewportTop = newViewportTop
cur = r.moveCursor(&b, cur, cursorLine, cursorCol)
```

Each `\r\n` physically moves the terminal cursor down, so `cur` correctly tracks where the cursor is — including past the last line of `newLines`. However, when the accumulated value of `cur` exceeds `r.viewportTop + r.height - 1` (the viewport bottom), `moveCursor` receives an out-of-viewport `from` value. If `moveCursor` does not clamp `from` before computing the delta to the target `cursorLine`, it will emit a cursor-up sequence with a magnitude that is too large, scrolling the display above the intended input row.

This manifests whenever:
1. `r.prev` is longer than `newLines` (content genuinely shrinks), **and**
2. The stale rows pushed `cur` past the viewport bottom, **and**
3. The target `cursorLine` is near the bottom of the visible area.

**Suggested fix**: cap `cur` at the viewport bottom before calling `moveCursor`, or clamp `from` inside `moveCursor` itself.

pkg/leantui/update.go
```diff
@@ -0,0 +1,510 @@
...
+	m.runCancel = cancel
+	m.busy = true
+	m.app.Run(runCtx, cancel, message, attachments)
+}
+
+func (m *model) startCompact(ctx context.Context, prompt string) {
+	runCtx, cancel := context.WithCancel(ctx)
+	m.runCancel = cancel
+	m.busy = true
+	m.app.CompactSession(runCtx, prompt)
```
docker-agent: **[MEDIUM] `startCompact` does not pass `cancel` to `CompactSession`, leaving the UI stuck in a busy state**

`startRun` and `startSkillFork` both pass the derived `cancel` function to the App so it can signal completion and free the UI:

```go
// startRun — correct
m.app.Run(runCtx, cancel, message, attachments)

// startSkillFork — correct
m.app.RunSkillFork(runCtx, cancel, name, task, nil)

// startCompact — missing cancel
m.app.CompactSession(runCtx, prompt)   // cancel is never given to the App
```

Because `CompactSession` never receives `cancel`, it cannot call it upon completion. `m.busy` is only cleared in `handleStreamStopped`, which fires from a `StreamStoppedEvent`. If the App relies on the passed `cancel` to trigger that event (or any equivalent termination signal), compaction will end silently without `m.busy` being reset — leaving the spinner spinning and blocking further user input until the user manually presses Ctrl+C.

**Suggested fix**: update the `CompactSession` signature to accept a `context.CancelFunc` (matching `Run`) and pass `cancel`:
```go
m.app.CompactSession(runCtx, cancel, prompt)
```

rumpl: Fixed in e2431226: `CompactSession` now accepts and stores the `context.CancelFunc`, and `startCompact` passes it just like `Run`/`RunSkillFork`. The app cancels the derived context when compaction finishes, and the lean TUI handles `SessionCompactionEvent` completion to clear `busy`, reset `runCancel`, and continue queued work.

pkg/leantui/renderer.go
rumpl: Fixed in e2431226: `moveCursor` now clamps the current row (`from`) to the visible viewport before computing the cursor movement, so stale-row erasure cannot produce oversized cursor-up sequences. I also added `TestRendererMoveCursorClampsCurrentRow` to cover the regression.

docker-agent: Perfect fix — clamping `from` inside `moveCursor` handles the root cause elegantly. The test case confirms that when `from = 100` and `viewportTop = 5`, the function now clamps to row 7 (bottom of a 3-line viewport) and emits `CursorUp(2)` instead of the buggy `CursorUp(95)`. This prevents the stale-row erasure loop from corrupting the cursor position when content shrinks.

Thanks for the quick turnaround!

pkg/leantui/update.go
docker-agent: Perfect — confirmed the fix at line 240. `startCompact` now passes `cancel` to `CompactSession`, matching the pattern in `startRun` and `startSkillFork`. The addition of `handleSessionCompaction` to explicitly handle the completion event and call `finishBusy` is a nice touch that ensures the UI state is properly reset regardless of how the compaction terminates. Thanks for the quick fix!

dgageot (approved):

PR #3128, opened 2026-06-15
perf(tui): skip sidebar scrollbar-probe pass on animation frames

dgageot:
While the agent is streaming, the 14 FPS spinner tick invalidates the sidebar render cache every frame. `verticalView()` then rebuilds all sidebar sections twice — the scrollbar "two-pass" (render without scrollbar to count lines, then re-render narrower if a scrollbar is needed) — even though the only thing that changed is a single fixed-width spinner glyph.

This splits cache invalidation in two: `invalidateCache()` for changes that can alter line count or scrollbar visibility (todos, sizing, agents, theme, …), which keeps the full two-pass render, and `invalidateAnimation()` (new) for the spinner-tick path. Because spinner glyphs are fixed single-cell width, the line count and scrollbar visibility cannot change, so `verticalView()` renders the sections once, reusing the cached scrollbar decision.

Profiling the real sidebar `View()` in a streaming state showed ~60% of per-frame CPU in `lipgloss.Style.Render` and ~45% in `ansi.stringWidth`, and the cost was doubled by the redundant second pass whenever a scrollbar is present. Per-frame sidebar render cost and allocations roughly halve while streaming (Apple M4 Max): at 50 todos, 4.35 ms → 2.15 ms and 51k → 25k allocations; at 120 todos, 9.49 ms → 4.67 ms and 116k → 58k allocations; at 300 todos, 22.2 ms → 11.2 ms and 282k → 141k allocations. This stacks with caching the todo render independently.

Output is unchanged. New `TestAnimationFastPathMatchesFullRebuild` asserts the fast path is byte-identical to a full two-pass rebuild across 0–300 todos; `TestAnimationFastPathTracksSpinnerFrame` asserts the fast path still reflects the advancing spinner. New before/after benchmarks included.

Relates to #3123.

### DIFF
--- a/pkg/tui/components/sidebar/sidebar.go
+++ b/pkg/tui/components/sidebar/sidebar.go
@@ -156,6 +156,7 @@ type model struct {
 	cachedWidth          int      // Width used for cached render
 	cachedNeedsScrollbar bool     // Whether scrollbar is needed for cached render
 	cacheDirty           bool     // True when cache needs rebuild
+	layoutDirty          bool     // True when a change may alter line count/scrollbar visibility (not just an animation frame)
 
 	// Agent click zones: maps content line index to agent name for click detection
 	agentClickZones map[int]string // content line -> agent name
@@ -189,6 +190,7 @@ func New(sessionState *service.SessionState) Model {
 		preferredWidth:   DefaultWidth,
 		titleInput:       ti,
 		cacheDirty:       true, // Initial render needed
+		layoutDirty:      true, // First render must probe scrollbar visibility
 	}
 	return m
 }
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The animation fast-path optimization is correct and well-structured. The two-level cache invalidation (`invalidateCache()` vs `invalidateAnimation()`) is sound: `layoutDirty` is initialized to `true`, only cleared after a full two-pass render, and never touched by `invalidateAnimation()`. The fast-path guard `!m.layoutDirty && len(m.cachedLines) > 0 && m.cachedWidth == contentWidthNoScroll` correctly prevents the optimization from firing before a valid scrollbar decision has been established. All layout-altering state changes continue to call `invalidateCache()`, which sets both flags.

The tests are well-designed: `TestAnimationFastPathMatchesFullRebuild` exercises 0–300 todos (spanning both scrollbar-absent and scrollbar-present states), and `TestAnimationFastPathTracksSpinnerFrame` confirms the fast path does not freeze the spinner glyph. Benchmarks are included for before/after comparison.

rumpl (approved):

aheritier (approved):

PR #3129, opened 2026-06-15
fix(runtime): finalize event channel on runStreamLoop early returns

### LINKED ISSUE #3073: events channel leaked and elicitation bridge left stale on early returns in runStreamLoop
## Summary

`runStreamLoop` swaps this stream's events channel into the elicitation bridge **before** the deferred `finalizeEventChannel` cleanup is registered. Two early-return paths sit in that gap. On those paths the stream exits without ever restoring the bridge or closing the events channel: the `observe` forwarder goroutine never terminates, the external consumer waits forever on a channel that never closes, and the elicitation bridge is left pointing at a dead stream's channel.

This is pre-existing (it predates #3069 / #3070) but belongs to the same elicitation-channel lifecycle family. Found while reviewing #3070.

## The gap

In `pkg/runtime/loop.go`, `runStreamLoop`:

```go
prevElicitationCh := r.elicitation.swap(events)   // bridge now points at this stream

...

agentTools, err := r.getTools(ctx, a, sessionSpan, sink, true)
if err != nil {
    sink.Emit(ErrorWithCode(...))
    return                                        // ← early return #1: no finalize
}

...

if lastMsg.Role == chat.MessageRoleUser {
    stop, msg, ctxMsgs := r.executeUserPromptSubmitHooks(...)
    if stop {
        ...
        r.emitHookDrivenShutdown(ctx, a, sess, msg, sink)
        return                                    // ← early return #2: no finalize
    }
    ...
}

...

var streamReason string
defer func() {                                    // ← finalize only registered HERE
    r.finalizeEventChannel(ctx, sess, streamReason, prevElicitationCh, events)
}()
```

## Consequences on the early-return paths

1. `events` is never closed → `observe`'s forwarder goroutine (`for event := range inner`) never exits → its `out` channel is never closed → any consumer doing `for range RunStream(...)` hangs forever. `Run()` is such a consumer.
2. The elicitation bridge keeps pointing at the dead stream's channel. A later elicitation `send` targets a channel nobody drains: with the buffer full it parks forever holding the bridge `RLock`, which then wedges the next stream's teardown (`restoreAndClose` blocks on the write lock).
3. For nested streams, the parent's elicitation channel is never restored.

## Proposed fix

Register the cleanup defer immediately after the swap, so every exit path finalizes:

```go
prevElicitationCh := r.elicitation.swap(events)

var streamReason string
defer func() {
    r.finalizeEventChannel(ctx, sess, streamReason, prevElicitationCh, events)
}()
```

`finalizeEventChannel` is already safe for these early paths (it emits `StreamStopped` non-blockingly and closes last, since #3070). The `streamReason` declaration just moves up with the defer.

Add a regression test: a `RunStream` whose `getTools` fails must still close its events channel (consumer's `for range` terminates) and restore the previous elicitation channel.

## Relations

- Same lifecycle family as #3069 (fixed by #3070); discovered during the #3070 review.

Sayt-0:
Fixes #3073

## Problem

`runStreamLoop` swaps this stream's events channel into the elicitation bridge **before** the `finalizeEventChannel` cleanup defer is registered. Two early-return paths sit in that gap:

- `getTools` failure
- a `user_prompt_submit` hook signalling termination

On those paths the stream exits without closing the events channel or restoring the bridge:

- `observe`'s forwarder goroutine (`for event := range inner`) never exits, so a `for range RunStream(...)` consumer hangs forever
- the elicitation bridge stays pointing at the dead stream's channel
- for nested streams, the parent's elicitation channel is never restored

## Fix

Register the cleanup defer immediately after the swap, so every exit path finalizes. `finalizeEventChannel` is already safe on these early paths (it emits `StreamStopped` non-blockingly and closes last, since #3070). The `streamReason` declaration just moves up with the defer.

## Test

Added a regression test (`TestRunStreamClosesChannelAndRestoresElicitationOnEarlyReturn`) that drives the reachable early return (a `user_prompt_submit` hook that stops the run) and asserts the consumer's range terminates and the previous elicitation channel is restored. Confirmed it hangs/fails without the fix.

Note: the `getTools` early return is not easily reachable today since `collectTools` swallows toolset listing errors, so the test exercises the hook-stop path. The same defer covers both.

## Checks

- `go build ./...`
- `go test -race ./pkg/runtime/ ./pkg/app/ ./pkg/a2a/ ./pkg/tui/...` all pass
- verified consumers of `StreamStopped` (sidebar, chat page, supervisor, a2a adapter) handle a stop without a preceding `StreamStarted` safely

### DIFF
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -227,6 +227,22 @@ func (r *LocalRuntime) runStreamLoop(ctx context.Context, sess *session.Session,
 	// parent's channel.
 	prevElicitationCh := r.elicitation.swap(events)
 
+	// streamReason records the exit reason from the final turn so
+	// finalizeEventChannel can surface it in the StreamStoppedEvent.
+	// It is updated by each turn via runTurn (passed by pointer).
+	//
+	// Register the cleanup defer immediately after the swap so every exit
+	// path finalizes, including the early returns below (tool setup
+	// failure, a user_prompt_submit hook signalling termination). Without
+	// this, those early returns leak the events channel: observe's
+	// forwarder goroutine never exits, a `for range RunStream(...)`
+	// consumer hangs forever, and the elicitation bridge is left pointing
+	// at this dead stream's channel.
+	var streamReason string
+	defer func() {
+		r.finalizeEventChannel(ctx, sess, streamReason, prevElicitationCh, events)
+	}()
+
 	a := r.resolveSessionAgent(sess)
 
 	// session_start fires once per RunStream. Its AdditionalContext
@@ -280,14 +296,6 @@ func (r *LocalRuntime) runStreamLoop(ctx context.Context, sess *session.Session,
 	}
 
 	sink.Emit(StreamStarted(sess.ID, a.Name()))
-
-	// streamReason records the exit reason from the final turn so
-	// finalizeEventChannel can surface it in the StreamStoppedEvent.
-	// It is updated by each turn via runTurn (passed by pointer).
-	var streamReason string
-	defer func() {
-		r.finalizeEventChannel(ctx, sess, streamReason, prevElicitationCh, events)
-	}()
 
 	if a.HasHarness() {
 		streamReason = r.runHarnessAgent(ctx, sess, a, slices.Concat(ls.sessionStartMsgs, ls.userPromptMsgs), sink)
--- a/pkg/runtime/elicitation_test.go
+++ b/pkg/runtime/elicitation_test.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"context"
 	"sync"
 	"testing"
 	"time"
@@ -9,6 +10,8 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/docker/docker-agent/pkg/agent"
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/hooks"
 	"github.com/docker/docker-agent/pkg/session"
 	"github.com/docker/docker-agent/pkg/team"
 )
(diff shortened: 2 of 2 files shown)


dgageot (approved):

docker-agent (commented):

The core fix — moving the defer immediately after the elicitation bridge swap — is correct and well-reasoned.

pkg/runtime/loop.go
```diff
@@ -227,6 +227,22 @@ func (r *LocalRuntime) runStreamLoop(ctx context.Context, sess *session.Session,
...
+	// It is updated by each turn via runTurn (passed by pointer).
+	//
+	// Register the cleanup defer immediately after the swap so every exit
+	// path finalizes, including the early returns below (tool setup
+	// failure, a user_prompt_submit hook signalling termination). Without
+	// this, those early returns leak the events channel: observe's
+	// forwarder goroutine never exits, a `for range RunStream(...)`
+	// consumer hangs forever, and the elicitation bridge is left pointing
+	// at this dead stream's channel.
+	var streamReason string
```
docker-agent, line 241: **[MEDIUM] `streamReason` is empty on both early-return paths, causing `StreamStoppedEvent` to carry a blank reason**

`streamReason` is declared as `""` (Go zero value) and only assigned via `streamReason = ls.exitReason` on the normal loop-exit path. The defer closure captures it, so on the two early-return paths:

1. **`getTools` failure** (a few lines below): returns before any assignment  
2. **`user_prompt_submit` hook stop** (~line 291): returns before any assignment  

`finalizeEventChannel` only substitutes a non-empty reason when `ctx.Err() != nil` (context cancellation). On hook-driven or tool-failure terminations the context is still alive, so `StreamStopped` is emitted with `reason == ""` — inconsistent with normal exits which always carry a non-empty `exitReason` constant.

**Suggested fix**: assign a meaningful reason constant before each early return, e.g.:
```go
if err != nil {
    sink.Emit(ErrorWithCode(ErrorCodeToolFailed, ...))
    streamReason = turnEndReasonError  // whichever constant fits
    return
}
```

docker-agent (commented):

The core fix — moving the defer immediately after the elicitation bridge swap — is correct. Two issues in the new early-return paths are worth addressing.

pkg/runtime/loop.go
```diff
@@ -281,14 +297,6 @@ func (r *LocalRuntime) runStreamLoop(ctx context.Context, sess *session.Session,
 
 	sink.Emit(StreamStarted(sess.ID, a.Name()))
```
docker-agent, line 298: **[MEDIUM] `StreamStopped` emitted without a preceding `StreamStarted` on `getTools` failure**

The new defer (registered earlier in the function) causes `finalizeEventChannel` to fire and emit `StreamStopped` on every exit path — including the `getTools` failure at ~line 264, which returns before reaching this `StreamStarted` emit. Consumers therefore receive `StreamStopped` with no matching `StreamStarted` on the tool-failure path.

The PR description states consumers were verified to handle this safely. However, since the events channel was never previously closed on this path (the root bug being fixed), consumers could not have been meaningfully exercised against it in practice. Any consumer that initialises state on `StreamStarted` before processing `StreamStopped` will silently misbehave.

**Suggestion**: Add a regression test for the `getTools` failure path (similar to the hook-stop test but with a broken toolset), or explicitly document that `StreamStopped` can arrive without `StreamStarted` as a protocol guarantee consumers must handle.

PR #3130, opened 2026-06-15
fix(skills): discover .agents/skills in non-git grouping parents

### LINKED ISSUE #3056: Repo-local `.agents/skills` anchored to nearest git root is invisible from inside sub-repos of a non-git parent (monorepo-grouping layout)
## Summary
Local skill discovery for repo-local `.agents/skills` is anchored to the **nearest enclosing git root**. In a layout where several independent git repositories are grouped under a **non-git parent directory**, a skill placed at the grouping-root's `.agents/skills` is only discovered when the agent's working directory is exactly that root. As soon as docker-agent runs from inside any sub-repo, the upward walk stops at the sub-repo's `.git` and never reaches the parent — so the grouping-level skill silently disappears.

Severity: low / latent. It does not bite when launching from the grouping root (so it's not the cause of #3054), but the behavior is surprising and undocumented.

## Filesystem setup (anonymized)
A common "grouping" layout — a plain directory that is **not** a git repo, containing multiple independent clones, each its own repo:

```
~/work/dev/org/                      <- grouping root, NOT a git repo (no .git)
├── .agents/
│   └── skills/
│       └── services/
│           └── SKILL.md             <- skill name: "services"  (grouping-level skill)
├── service-a/                       <- independent repo
│   └── .git/
├── service-b/                       <- independent repo
│   └── .git/
└── docker-agent/                    <- independent repo
    └── .git/
```

The intent: `services` is a cross-repo helper skill that should be available no matter which repo you're working in.

## What works vs. what doesn't

| Working directory when docker-agent is launched | `projectSearchDirs` walks                                  | Finds `~/work/dev/org/.agents/skills/services`? |
|--------------------------------------------------|------------------------------------------------------------|--------------------------------------------------|
| `~/work/dev/org` (grouping root, no `.git`)      | no git marker found → returns just `[cwd]`                  | ✅ yes (cwd == grouping root)                     |
| `~/work/dev/org/docker-agent`                    | hits `docker-agent/.git` immediately → stops there         | ❌ no (never reaches the parent)                  |
| `~/work/dev/org/docker-agent/pkg/skills`         | walks up, hits `docker-agent/.git` → stops there           | ❌ no                                             |
| `~/work/dev/org/service-a`                       | hits `service-a/.git` → stops there                        | ❌ no                                             |

So the grouping-level `services` skill is only visible from the one directory that is the grouping root itself; from inside any actual repo (the normal place to do work) it vanishes.

## Root cause
`projectSearchDirs` (in `pkg/skills/local.go`) walks **up** from cwd and stops at the **first** `.git` marker:

```go
for current := abs; ; {
    dirs = append(dirs, current)
    if hasGitMarker(current) {   // stops at the nearest git root
        gitRoot = current
        break
    }
    parent := filepath.Dir(current)
    if parent == current { break }
    current = parent
}
if gitRoot == "" {
    return []string{abs}         // not in a git repo: only cwd
}
```

`localSearchPaths` then scans `<dir>/.agents/skills` for each returned `dir`. Because the walk terminates at the nearest git boundary, it can never cross **above** a repo into a non-git parent. A skill living in that non-git parent is therefore unreachable from inside any sub-repo.

## Expected / discussion
At minimum the boundary behavior should be documented (it currently is not — see the doc comments below, which describe "from git root down to cwd" without noting the non-git-parent blind spot). Possible improvements for maintainers to weigh:
- Continue walking up past a git root to also include `.agents/skills` in non-git ancestor directories (up to `$HOME` or a sane ceiling), so grouping-root skills are picked up; and/or
- Provide an explicit, configurable extra search root (env var or config) for "umbrella"/grouping layouts; and/or
- Document the git-root anchoring clearly so users know to place shared skills under `~/.agents/skills` (the global recursive path), which is unaffected.

## Affected code references
- `pkg/skills/local.go` — `projectSearchDirs` (the upward walk that stops at the nearest `.git`) and `localSearchPaths` (which builds `.agents/skills` paths only from `projectSearchDirs`' result).
- `pkg/skills/local.go` — `localSearchPaths` doc comment: ".agents/skills` from the git root down to cwd (closest wins)".
- `pkg/skills/skills.go` — `Load` doc comment: ".agents/skills/` (flat, scanned from git root to cwd)".

## Note
Distinct from #3054 (that issue is a slash-command dispatch loop; this one is about discovery reachability in grouping layouts).

Sayt-0:
## What

`projectSearchDirs` stopped at the nearest `.git` when walking up from cwd, so a `.agents/skills` placed in a non-git "grouping" parent (e.g. `~/work/org` holding several independent repos) was invisible from inside any sub-repo. See #3056.

When cwd is inside `$HOME`, the search now walks up to (but not including) `$HOME` instead of stopping at the git root, so grouping-level skills are reachable from any sub-repo. `$HOME` itself is excluded because `~/.agents/skills` is already scanned as a global path.

## Design choice

The issue lists three options for maintainers to weigh. This PR takes option 1 (walk past the git root, with `$HOME` as the ceiling) plus the documentation fix. Reasoning for picking it over the others:

- Option 3 (document only) does not fix discovery, it just records the blind spot. The issue's intent is that a shared skill be available from any repo, and documentation alone does not make that true.
- Option 2 (env var / config search root) does fix it, but requires manual configuration of an extra root and introduces a new config surface (variable naming, plumbing, docs) that does not exist yet.
- Option 1 fits what the code already does: it extends the existing "closest wins" walk from root to cwd rather than adding a new mechanism. The `$HOME` ceiling is natural because `~/.agents/skills` is already the global path, so option 1 slots cleanly between global home skills and repo skills. It is also the smallest change for the value (one function, no new API or config).

If option 2 seems more appropriate, it can always be changed.

Precedence is unchanged for existing skills: ancestors are lower priority than deeper dirs, so a repo-level skill still wins (closest-to-cwd wins).

## Known boundary

When cwd is outside `$HOME` (e.g. `/opt/work/org/repo`), discovery stays anchored at the git root so it never climbs into system dirs like `/opt` or `/`. The issue only describes layouts under `$HOME`, so this is intentional rather than a gap.

## Tests

- `TestProjectSearchDirs/under_home_walks_past_git_root...`
- `TestLoad_AgentsSkillsFromNonGitGroupingParent` (reproduces the issue end to end)
- `TestLoad_AgentsSkillsPrecedence_GroupingOverridesGlobal`

All existing skills tests still pass.

Closes #3056

### DIFF
--- a/pkg/skills/local.go
+++ b/pkg/skills/local.go
@@ -43,7 +43,8 @@ type localSearchPath struct {
 //
 //  1. Global directories (under $HOME).
 //  2. .claude/skills under cwd (Claude project format, flat).
-//  3. .agents/skills from the git root down to cwd (closest wins).
+//  3. .agents/skills in each ancestor of cwd up to $HOME — or the enclosing
+//     git root when cwd is outside $HOME — down to cwd (closest wins).
 func localSearchPaths() []localSearchPath {
 	if kit := os.Getenv(KitDirEnv); kit != "" {
 		return []localSearchPath{{filepath.Join(kit, KitSkillsSubdir), true}}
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is well-scoped and logically sound. The home-walking loop in `projectSearchDirs` terminates correctly (the `IsWithin` precondition guarantees `home` is always an ancestor of `abs`, so the loop reaches `home` before the filesystem root; the `parent == current` guard handles the filesystem-root edge case in `gitAnchoredDirs`). `slices.Reverse` is placed correctly in both branches. The `abs != home` guard prevents double-counting `~/.agents/skills` in all practical paths, and the fallthrough to `gitAnchoredDirs` when outside `$HOME` preserves existing behavior for non-home layouts. Tests cover the new code path, the precedence rules, and the end-to-end reproduction.

pkg/skills/local.go
```diff
@@ -101,42 +102,62 @@ func homeSkillSearchPaths(home string) []localSearchPath {
...
 	if err != nil {
 		return []string{cwd}
 	}
 
-	// Walk up from cwd, recording each dir, until we either hit a git
-	// marker or run out of parents. This collects findGitRoot's result
-	// and the dir list in a single traversal.
+	// Walk up to $HOME only when cwd is inside it (and not $HOME itself, which
+	// is already covered by the global ~/.agents/skills scan).
+	if home := paths.GetHomeDir(); home != "" && abs != home && pathx.IsWithin(abs, home) {
```
aheritier, line 127: not sure if for windows we won't have the issue cc @ronan-thibaut-glitch

aheritier (approved):
it looks ok to me but I would prefer a review of @dgageot or @rumpl

pkg/skills/local.go
ronan-thibaut-glitch: I'll have a look at it

dgageot (approved):

PR #3135, opened 2026-06-15
fix(acp): respect workspace roots for Zed paths

rumpl:
Resolve ACP filesystem paths against the session cwd and advertised additional directories instead of re-rooting absolute paths under the agent process directory. This prevents temporary screenshot paths from creating project-local var/folders trees.

Also advertise inline image prompt support, avoid leaking blocked resource-link URIs into prompts, preserve filesystem tool options in the ACP registry, and return method-not-found for unsupported optional ACP calls.

### DIFF
--- a/pkg/acp/agent.go
+++ b/pkg/acp/agent.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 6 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27574912029).

krissetto (approved):

Issue #2788, opened 2026-05-13
feat: plan mode — /plan toggle, submit_plan tool, and plan_model for dual-model execution


aheritier:
## Problem

Agentic workflows that span many tool calls (write files, run commands, call GitHub APIs) give users no control point between "I described what I want" and "the agent already did it."

The only safeguard today is an instruction asking agents to present a plan and wait — a soft control that works most of the time but has no runtime enforcement, and cannot be toggled at runtime without modifying config files.

Two concrete gaps:

1. **No user-controlled mode switch.** There is no way to type `/plan` in the TUI to enter a careful, plan-then-approve mode for a risky task — and then `/plan off` to return to fluid execution for a simple one.
2. **No cost-optimised planning.** Planning requires deep reasoning (costly model, used once). Execution is mechanical (cheaper model, used N times). There is no way to route the planning turn to a different model than the execution turns.

---

## Prior Art

Every mature agentic system has converged on this pattern:

| System | Plan format | User toggle | Dual-model? |
|---|---|---|---|
| **Cursor** | Structured file list with diff preview | Ask vs Agent mode; `/multitask` slash command; per-subagent model setting (May 2026) | Yes — subagent model configurable separately from parent |
| **GitHub Copilot Workspace** | JSON `plan_action` function call | Phase gate: Task → Plan → Code; natural-language modify | Yes — o1-preview for planning, gpt-4o for execution |
| **OpenAI Operator** | JSON schema | Structured approval gate in UI | Yes — reasoning model for plan, fast model for execution |
| **Devin** | JSON step tree | Checkpoint-based pauses; user steers mid-execution | Implicit (planning phase uses extended thinking budget) |
| **Claude extended thinking** | `<thinking>` block + text response | Application-layer convention (no built-in gate) | Single model, separate token budget for reasoning |
| **LangGraph** | Free-form or JSON | `HumanNode` blocks graph traversal | Framework-dependent |

**Key pattern across all:** plan generation is separated from execution, the plan is shown to the user, and execution is gated on approval.

**Cursor model-routing (verified, May 2026 changelog):**
> "Added the ability to control Explore subagent behavior from settings: choose a specific model for Explore subagents to run on, inherit the same model as the parent agent, or disable Explore subagents altogether."
> "Added support for general model names for subagent configuration."
> "`/multitask` is now available in the editor for running async subagents."

**Anthropic extended thinking API (confirmed from SDK):**
```python
response = client.messages.create(
[...]

PR #3140, opened 2026-06-16
feat: add session-scoped plan/build mode

trungutt:
## Why

A planner sub-agent (e.g. the bundled `coder` agent's `planner`, `pkg/config/builtin-agents/coder.yaml`) puts planning under the **agent's** control. Plan mode puts it under the **user's** control: a session attribute the host (CLI, TUI, embedder) flips like a switch, and the model can't talk its way out of it.

That distinction matters because:

- **Works on any agent**, third-party or hand-rolled — no YAML edit needed to add a planner sub-agent.
- **Hard tool removal, not prompt drift.** `coder`'s planner today still ships full `filesystem` (incl. `write_file` / `edit_file`); its read-only-ness is prompt-only. Plan mode strips every tool without `Annotations.ReadOnlyHint` from what the model sees.
- **Covers runtime-added MCP tools** via the MCP-spec `ReadOnlyHint` annotation, with no YAML to update.
- **Mid-session toggle costs zero turns** — one PATCH and the next turn observes it (vs. a transfer + synthesised user prompt for sub-agent handoff).
- **One primitive across hosts**, instead of each host parsing agent topology or rolling its own deny-list.

Additive: planner sub-agents stay the right answer when planning belongs in the agent's design; plan mode is the user-driven counterpart.

## What

New per-session `Mode` (`build` default, `plan`). In plan mode the runtime:

1. Filters tools without `Annotations.ReadOnlyHint` from each turn's toolset (hard guarantee).
2. Splices a per-turn `<system-reminder>` telling the model to plan, not act (so it doesn't waste turns trying tools that vanished).

Sub-sessions (`transfer_task`, `run_skill`, `agent` background-agent) inherit the parent's mode so the filter applies across the whole delegation tree. Harness-backed agents — which own their own toolset and can't be filtered from the runtime — refuse plan mode with `code: "unsupported_mode"` rather than degrade plan mode to advisory.

API:

```
POST   /api/sessions               { ..., "mode": "plan" }   # rejects unknown values with 400
GET    /api/sessions/:id                                      # + snapshot
PATCH  /api/sessions/:id/mode      { "mode": "plan" }
```

`PATCH /mode` updates the live in-memory session if a runtime is attached and persists to the store, so the next turn picks it up directly. Persistence is a new `mode TEXT` column (migration 22); pre-existing rows normalize to `build` on read. The field is guarded by `s.mu` via `LoadMode` / `StoreMode` accessors so concurrent runtime reads and HTTP writes can't race.

Filter piggybacks on the existing `ExcludedTools` filter site in `runtime/loop.go`; reminder rides the existing turn-start system-message splice.

## Scope

In: runtime filter + reminder, sub-session mode propagation, harness refusal, API + DTOs, persistence + migration, user-facing API docs, unit + HTTP + race tests.

Out (left to hosts): TUI/leantui UX, CLI flags, "exit plan mode" tool — hosts can PATCH the endpoint directly.


pkg/server/session_manager.go
```diff
@@ -741,6 +746,29 @@ func (sm *SessionManager) UpdateSessionPermissions(ctx context.Context, sessionI
...
+// If the session is actively running, it also updates the in-memory session
+// object so the next turn's tool filter and plan-mode reminder see the new
+// mode without having to round-trip through the store.
+func (sm *SessionManager) UpdateSessionMode(ctx context.Context, sessionID string, mode session.Mode) error {
+	mode = session.NormalizeMode(mode)
+	sm.mux.Lock()
+	defer sm.mux.Unlock()
+
+	if rt, ok := sm.runtimeSessions.Load(sessionID); ok && rt.session != nil {
+		rt.session.Mode = mode
```
docker-agent: **[HIGH] Data race: `sess.Mode` written by `UpdateSessionMode` without synchronization visible to the run loop**

`UpdateSessionMode` writes `rt.session.Mode` while holding `sm.mux` (the `SessionManager`'s own mutex). The run-loop goroutine reads `sess.Mode` in `filterToolsForSession` and `planModeReminderMessages` without holding **any** lock — `Session.mu` only protects `Messages`, not `Mode`.

These two goroutines access the same `Mode` field concurrently without shared synchronization, which is a data race under Go's memory model and will be flagged by `go test -race`.

**Minimal fix**: Add a dedicated setter/getter on `Session` that uses `session.mu`:
```go
func (s *Session) SetMode(m Mode) {
    s.mu.Lock()
    defer s.mu.Unlock()
    s.Mode = m
}

func (s *Session) GetMode() Mode {
    s.mu.RLock()
    defer s.mu.RUnlock()
    return s.Mode
}
```
Then have `UpdateSessionMode` call `rt.session.SetMode(mode)` and the run-loop call `sess.GetMode()`.

pkg/runtime/harness.go
```diff
@@ -46,7 +46,8 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
 	}()
 
 	turnStartMsgs := r.executeTurnStartHooks(ctx, sess, a, events)
-	messages := sess.GetMessages(a, append(baseExtra, turnStartMsgs...)...)
+	planReminder := planModeReminderMessages(sess)
```
docker-agent: **[HIGH] Data race: `sess.Mode` read in `runHarnessAgent` without lock while `UpdateSessionMode` may write it concurrently**

`planModeReminderMessages(sess)` reads `sess.Mode` in the harness goroutine without holding any lock. This is the same root-cause data race as in `session_manager.go` — concurrent write from `UpdateSessionMode` (under `sm.mux`) and an unsynchronized read here. Go's race detector will catch this.

The same `SetMode`/`GetMode` accessor fix resolves this: replace the direct `sess.Mode` field read in `planModeReminderMessages` with `sess.GetMode()`.

pkg/server/session_manager.go
```diff
@@ -741,6 +746,29 @@ func (sm *SessionManager) UpdateSessionPermissions(ctx context.Context, sessionI
...
+		slog.DebugContext(ctx, "Updated mode for active session", "session_id", sessionID, "mode", mode)
+		return sm.sessionStore.UpdateSession(ctx, rt.session)
```
docker-agent, line 886: **[MEDIUM] `UpdateSessionMode` passes the live `rt.session` pointer to `UpdateSession` without cloning**

When a runtime session is active, `UpdateSessionMode` mutates `rt.session.Mode` and immediately calls `sm.sessionStore.UpdateSession(ctx, rt.session)` with the live pointer. The run-loop goroutine is concurrently modifying the same struct (appending messages, updating token counts). `SQLiteSessionStore.UpdateSession` reads all those fields to build its `UPDATE` statement, racing with in-flight mutations from the run loop.

Consider cloning the session after setting the mode (while still holding `sm.mux`):
```go
rt.session.Mode = mode
snap := rt.session.Clone()
return sm.sessionStore.UpdateSession(ctx, snap)
```
This matches how other methods in the codebase avoid reading partially-mutated live state.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

Two confirmed data races and one likely concurrent-write hazard in the new plan-mode machinery.

**Confirmed findings:**
- 🔴 CRITICAL (pkg/server/session_manager.go) — Data race: UpdateSessionMode writes rt.session.Mode under sm.mux while the run loop reads it without any lock (go test -race will catch this).
- 🔴 CRITICAL (pkg/runtime/harness.go) — Same data race: planModeReminderMessages reads sess.Mode in the harness goroutine without any lock.
- 🟡 MEDIUM (pkg/server/session_manager.go) — UpdateSessionMode passes the live rt.session pointer to UpdateSession without cloning, risking partial reads during an active stream.

**What is well done:** migration correctly numbered (22), mode normalizes on DB read via NormalizeMode, IsValid() gates the HTTP endpoint, Clone() propagates Mode, filter and reminder logic is clean and correct.

pkg/server/session_manager.go
trungutt: Accurate, but this matches the existing convention in `SessionManager`:

- `UpdateSessionTitle` writes `rt.session.Title = title` under `sm.mux` while the runtime reads `sess.Title` without lock (e.g. via `pkg/runtime/runtime.go:1005`).
- `ToggleToolApproval` writes `sess.ToolsApproved` under `sm.mux` while the runtime reads it without lock at `pkg/runtime/toolexec/dispatcher.go:323`.

`pkg/session/session.go` already exposes an `s.mu`-guarded `Usage()` accessor for `InputTokens`/`OutputTokens`, but it isn't mandatory either — most reads in `pkg/runtime/` go through the bare field (`hooks.go`, `event.go`, `loop.go:412`, `loop.go:896`, `loop_steps.go:169`). Only `session_compaction.go:112` uses the accessor.

So introducing an accessor just for `Mode` would be inconsistent with the surrounding code. The right cleanup is a uniform hardening pass — accessor methods for `Title`, `Mode`, `ToolsApproved`, and mandatory use in the runtime — driven by a `_race_test.go` that runs an HTTP PATCH against an in-flight `RunStream` so we can confirm we caught every read site. Happy to file that as a follow-up.

In the meantime the practical impact for `Mode` is bounded: read once per loop iteration, written on a rare user action, and `NormalizeMode` on the read side coerces any torn value to `ModeBuild`. Worst case is that a single turn uses the previous mode.

pkg/runtime/harness.go
trungutt: Same data race as in `UpdateSessionMode` above — see that thread for the analysis. This call site reads `sess.Mode` on the harness goroutine; the matching reader on the non-harness path is at `pkg/runtime/loop.go:562` (also a bare read).

Both readers and the writer in `SessionManager` follow the same convention as `Title` / `ToolsApproved` / `InputTokens` in the existing code, so I'd rather not single-case `Mode`. A uniform hardening pass with a race test is the right cleanup.

pkg/server/session_manager.go
trungutt: Same hazard exists today in `UpdateSessionTitle` (two lines down at L784, same `return sm.sessionStore.UpdateSession(ctx, rt.session)`). The store reads fields off the live pointer via `sessionPersistedFieldsOf`, so the partial-read risk covers every field — not just `Mode`.

Cloning before the store call is the right fix, but it should apply uniformly to `UpdateSessionTitle`, `UpdateSessionPermissions`, etc. at the same time, otherwise we'd just be moving the problem around. Happy to do that as part of the broader hardening follow-up alongside the accessor work for #1/#2.

docker-agent: Good catch—you're right that this is a broader issue affecting , , and anywhere else we pass  to the store while the run loop is concurrently mutating it.

Agree that a piecemeal fix here would just shift the problem around. Bundling the cloning fix with the broader accessor hardening makes sense—it'll ensure the store always sees consistent snapshots across all update paths.

Thanks for flagging the scope; I'll leave this open as a marker for the unified fix.

pkg/server/session_manager.go
docker-agent: You're absolutely right — I should have checked the surrounding code before flagging this in isolation. The pattern is consistent across `Title`, `ToolsApproved`, and `Mode`, and introducing an accessor just for `Mode` would be out of step with the rest of the codebase.

I agree with your assessment: the right fix is a unified hardening pass that covers all three fields, paired with a `_race_test.go` that exercises the HTTP PATCH + RunStream concurrency path so we can catch every read site. Doing it piecemeal here would leave the other fields still racy and create confusion about which convention to follow.

Filing this as a follow-up makes sense — the practical impact is low given the read frequency and value semantics you described.

pkg/runtime/harness.go
docker-agent: Fair point — you're right that singling out `Mode` with accessors when `Title`, `ToolsApproved`, and `InputTokens` follow the same bare-read convention would introduce an inconsistency. A uniform hardening pass (ideally gated by `go test -race`) that addresses all session field races together is the cleaner fix.

Thanks for the context about the matching reader at `loop.go:562` — I'll hold off on proposing per-field solutions until the broader session-synchronization cleanup lands.

rumpl:
Question, why can't "plan mode" be implemented with a second agent that the user can select?

trungutt:
> Question, why can't "plan mode" be implemented with a second agent that the user can select?

I updated the PR description

rumpl:
> A planner sub-agent (e.g. the bundled coder agent's planner, pkg/config/builtin-agents/coder.yaml) puts planning under the agent's control.

This is not true, you can always ctrl+number to switch to an agent

trungutt:
@rumpl It could, but it scales poorly: every agent (built-in, community, user-authored) would need a hand-maintained "planner" twin kept in lockstep with its prompt, toolsets, hooks and sub-agents. Plan mode is effectively "every agent gets a free read-only twin", derived from each tool's MCP `ReadOnlyHint` — so user-added MCP servers and third-party agents work without anyone authoring a paired variant.

It also keeps you on the same agent under a different constraint, vs a sibling-agent switch (which has handoff cost: transfer event, possibly synthesised user prompt, separate conversation thread).

rumpl:
Okay but... Not any agent can become a planner agent magically just by removing write tools. Its system prompt needs to be changed I would guess, also you are talking about ReadOnlyHint, but our builtin tools don't have it, do they?

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27621120785).

trungutt:
Two points:

**System prompt.** It is changed per turn — `planModeReminderMessages` in `pkg/runtime/plan_mode.go` splices a `<system-reminder>` telling the model it is in plan mode, must not edit/shell/execute, and should draft a plan. Rides the same turn-start splice as `turn_start` hook output, so it picks up the cache_control marker too. That's the soft layer; the tool filter is the hard layer.

**ReadOnlyHint on built-ins.** They do have it — 33 occurrences across every read-side tool:

- `pkg/tools/builtin/filesystem/filesystem.go:425,449,462,475,487` — read_file, read_multiple_files, list_directory, directory_tree, search_files_content
- `pkg/tools/builtin/shell/shell.go:592,602` — list_background_jobs, view_background_job
- `pkg/tools/builtin/todo/todo.go:306–341` — all four todo tools
- `pkg/tools/builtin/memory/memory.go:152,175` — get_memories, search_memories
- `pkg/tools/builtin/tasks/tasks.go:575,607,617` — get_task, list_tasks, next_task
- plus `think.go:54`, `api.go:187`, `handoff.go:31`, `transfertask.go:34`, `mcpcatalog.go:423,435`, `modelpicker.go:74,84`, `rag.go:164`, `userprompt.go:65`, `skills.go:366,382`, `deferred.go:151,163`, `agent.go:480,490`

Write-side tools (`write_file`, `edit_file`, `shell`, `run_background_job`, `add_memory`/`update_memory`/`delete_memory`, `run_skill`, `change_model`, `lsp_rename`/`_code_actions`/`_format`, `run_background_agent`) deliberately don't carry it — exactly what plan mode wants to strip. MCP tools get `ReadOnlyHint` from the spec itself, so user-added MCP servers and third-party tools are covered with no per-tool config.

**On the sibling-agent alternative more broadly** — beyond the points already in the description, it's painful operationally: every agent (built-in, community, user-authored) would need a hand-maintained planner twin kept in lockstep with its prompt, toolsets, hooks, and sub-agents. Two YAML surfaces per agent to keep in sync, every parent-agent change silently rots its planner sibling, and every third-party agent author has to do the same work. Plan mode collapses that to one knob across the fleet — including agents we don't author.

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27631088634).

aheritier:
It should cover most of #2788

dgageot:
## Review — `feat: add session-scoped plan/build mode`

Overall this is clean and consistent with existing patterns — it builds, has good baseline test coverage, and the migration is backwards compatible. One blocking issue and a couple of concerns below.

### 🔴 Blocking: plan mode is bypassable via delegated sub-agents

Delegation tools (`transfer_task`, `handoff`) are marked read-only, so they survive plan-mode filtering. But delegated child sessions don't appear to inherit the parent's mode — they default back to `build` and get the full mutating toolset. So a plan-mode agent can't call `write_file`/`shell` directly, but it *can* delegate to a sub-agent that can, which undermines the core safety guarantee of plan mode.

**Suggested fix:** propagate the parent mode when creating sub-sessions (e.g. `session.WithMode(parent.Mode)`) in `pkg/runtime/agent_delegation.go` / `pkg/runtime/loop.go`, and add a regression test asserting that a plan-mode parent produces plan-mode children with mutating tools filtered out.

If sub-agents are *intended* to run in build mode, please document that explicitly and soften the reminder text so it doesn't claim state-changing tools are unavailable.

### 🟡 Harness agents get the reminder but not the hard filtering

In `pkg/runtime/harness.go`, plan mode injects the reminder but doesn't seem to apply the same tool filtering as the streaming loop. If the harness manages its own toolset, plan mode becomes prompt-only there, making "Only read-only tools have been made available" misleading. Either apply the filtering or adjust the text and document the limitation.

### 🟡 Possible data race on `sess.Mode`

`UpdateSessionMode` writes the mode under the manager mutex, but runtime reads for filtering/reminders don't appear to use the same lock. Consider guarding `Mode` (or exposing locked accessors) and running the relevant tests with `-race`.

### ⚪ Minor

- **Inconsistent invalid-mode handling:** session creation normalizes invalid/empty mode → `build`, while the update endpoint rejects with `400`. Prefer making these consistent (rejecting both).
- **Mid-turn mode changes** apply on the next turn — matches the comments, not a bug, but worth documenting for API/UI callers.

### Test coverage

Good for core plumbing (filtering, reminder, persistence, migration, HTTP, in-memory). Would like to see before merge: a delegation-inheritance test, and a harness-mode behavior test (or explicit documentation that plan mode is advisory there). An optional concurrency/race test around mode updates would be nice.

### Recommendation

**Request changes** — the implementation is largely solid, but the delegated sub-agent bypass breaks the central guarantee of plan mode and should be fixed (or the guarantee redefined/documented) before merging. Looks close after that.

PR #3141, opened 2026-06-16
Use lipgloss and x/term for window size changes events

rumpl:


aheritier (dismissed):

pkg/leantui/terminal.go
```diff
@@ -36,52 +37,84 @@ func newTerminal(in, out *os.File) (*terminal, error) {
...
+
+	select {
+	case <-ch:
+	default:
 	}
-	if t.height <= 0 {
-		t.height = 24
+
+	select {
+	case ch <- size:
```
docker-agent: **[LOW] Silent resize event drop in `sendLatestResize` final fallback**

The three-step drain-and-resend idiom in `sendLatestResize` ends with a `default` branch that silently discards a resize event if the final send blocks. For a capacity-1 channel that was just drained in step 2 the send should always succeed — but the existence of the `default` path means a concurrent reader that drains the channel between step 2 and step 3 could cause the resize to be silently lost. In practice this race window is tiny, but the resulting behavior (terminal not reflowing) would be surprising to debug.

Consider removing the `default` branch from the final send, or adding a comment explaining the invariant that makes it safe to drop.

pkg/leantui/terminal.go
```diff
@@ -36,52 +37,84 @@ func newTerminal(in, out *os.File) (*terminal, error) {
 		return nil, fmt.Errorf("enabling raw mode: %w", err)
 	}
 
+	lipgloss.EnableLegacyWindowsANSI(out)
```
docker-agent: **[LOW] `EnableLegacyWindowsANSI` side-effects not reversed on initialization failure**

`lipgloss.EnableLegacyWindowsANSI(out)` is called after `term.MakeRaw` but before `cancelreader.NewReader`. If `cancelreader.NewReader` (or any later step) fails, `term.Restore` is called to undo raw mode — but any Windows console mode changes made by `EnableLegacyWindowsANSI` are not reversed. On non-Windows platforms this is a no-op, but on Windows the console could be left in an unexpected ANSI-enabled state after a failed init. The process typically exits on this error path so the practical impact is low, but symmetrically reversing the setup on failure would be more robust.

pkg/leantui/terminal.go
```diff
@@ -36,52 +37,84 @@ func newTerminal(in, out *os.File) (*terminal, error) {
...
-	t.width, t.height = t.querySize()
-	if t.width <= 0 {
-		t.width = 80
-	}
-	if t.height <= 0 {
-		t.height = 24
-	}
+	t.width, t.height = normalizeTerminalSize(t.querySize())
 
+	t.startResizeWatcher()
```
docker-agent: **[LOW] Goroutine started before initialization is complete — fragile ordering**

`t.startResizeWatcher()` launches a background goroutine before `writeString`/`flush` calls complete (and before `return`). There is no failure path after it today, so no goroutine is leaked currently. However, the pattern is fragile: any future error-returning code added after line 60 would leak the watcher goroutine because `t.stopResize` would never be closed. Consider moving `startResizeWatcher()` to the last step before `return t, nil`, or adding a deferred cleanup on error.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The polling-based resize watcher using a 250 ms ticker is a clean, cross-platform replacement for SIGWINCH-based detection. No high or medium severity bugs found. Three minor low-severity observations are left as inline comments.

melmennaoui (approved):

aheritier (approved):

PR #3146, opened 2026-06-16
fix(memory): sanitise reserved characters in default-path config segment

trungutt:
Refs: docker/desktop-feedback#449, https://docker.atlassian.net/browse/DDB-503

When an agent is loaded from an OCI reference, the config name passed
into the memory toolset's default-path branch carries the image tag's
`:` through to the directory name. `os.MkdirAll` then fails with
`ERROR_INVALID_NAME (123)` on NTFS because Windows reserves
`< > : " | ? * \ /` in any single path component, so the toolset
fails to initialise on every Windows host where an agent is consumed
that way.

The sanitisation is intentionally narrow: it only touches the segment
the memory toolset itself constructs, so the config name keeps its
shape everywhere else (logs, UI labels, …). The hash suffix already
embedded in the config name absorbs any loss from the replacement, so
this can't introduce collisions in practice.

### DIFF
--- a/pkg/tools/builtin/memory/memory.go
+++ b/pkg/tools/builtin/memory/memory.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/docker/docker-agent/pkg/config"
@@ -60,7 +61,7 @@ func CreateToolSet(toolset latest.Toolset, parentDir string, runConfig *config.R
 		if configName == "" {
 			configName = "default"
 		}
-		validatedMemoryPath = filepath.Join(paths.GetDataDir(), "memory", configName, "memory.db")
+		validatedMemoryPath = filepath.Join(paths.GetDataDir(), "memory", sanitizePathSegment(configName), "memory.db")
 	}
 
 	if err := os.MkdirAll(filepath.Dir(validatedMemoryPath), 0o700); err != nil {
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses `ERROR_INVALID_NAME (123)` on Windows NTFS when OCI image tags (containing `:`) are used as directory path segments.

**Review summary:**

- `sanitizePathSegment` properly replaces all Windows-reserved characters (`< > : " | ? * \ /`) and ASCII control characters (< 0x20, including NUL) with `_`
- Sanitization is applied to `configName` **before** `filepath.Join`, so no path-traversal via embedded `/` is possible
- The intentionally narrow scope (only the constructed segment, not the full config name) preserves labels/logs elsewhere — correct design
- The hash suffix in the config name prevents collisions after character replacement
- Test coverage is adequate for the sanitization logic

docker-agent (commented):
### Assessment: 🟢 APPROVE

The sanitisation logic is correct and complete. `sanitizePathSegment` covers all NTFS-reserved single-path-component characters (`< > : " | ? * \ /`) plus control characters below U+0020 using an idiomatic `strings.Map` — no regex, no allocation overhead. It is applied at exactly the right call-site (`CreateToolSet`'s default-path branch) and is invisible everywhere else (logs, UI labels). The test exercises a representative OCI-style config name and asserts the absence of each reserved character in the resulting directory entry.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The sanitization logic in `sanitizePathSegment` correctly covers all Windows NTFS reserved path-component characters (`< > : " | ? * \ /`) and control characters below `0x20`. The fix is well-scoped — it only touches the segment the memory toolset constructs, leaving the config name intact everywhere else. The test exercises the exact failure path (OCI-ref config name with `:`) on all OS platforms and validates the directory is created successfully.

**Low-severity notes** (no action required):

- **`sanitizePathSegment` misses DEL (0x7F)**: The `r < 0x20` guard stops at `0x1F`, leaving the DEL character unfiltered. DEL is non-printable and while not officially NTFS-reserved, it can confuse shell tooling. Given the stated intent to block control characters, extending the check to `r < 0x20 || r == 0x7F` would be more consistent.
- **Test checks absence not equality**: `TestCreateToolSet_DefaultPath_SanitisesReservedChars` asserts that each reserved character is absent from the directory name, but does not assert the full expected sanitized value (e.g., `oci_v8_______-deadbeef`). An equality assertion would catch regressions where the sanitizer over-replaces non-reserved characters.
- **Hash-suffix uniqueness is a convention, not enforced**: The comment in `sanitizePathSegment` claims "the hash suffix already in the config name preserves uniqueness." This holds only as long as callers always include a hash. If a plain name (no hash) is ever passed, two inputs differing only by reserved chars (e.g., `a:b` and `a/b`) would silently share the same `memory.db`. This is a latent risk worth documenting at the call site or enforcing in `CreateToolSet`.

None of these block merging. The core fix is correct and the test coverage is solid for the targeted scenario.

aheritier (approved):

dgageot (approved):

ryanexec:
sorry for the bad news, this is still an issue as of docker versions 4.79.0 and 4.80.0, with gordon version 1.79.0, but per the image below:

image

PR #3152, opened 2026-06-16
fix(runtime): persist sub-session transcript on error path in runCollecting

jedp-docker:
## Problem

`runCollecting` — the background-agent path used by `run_background_agent` — has the same structural bug fixed in #3151 for `runForwarding`: on `ErrorEvent` it returned early without calling `parent.AddSubSession`, silently dropping the sub-session from the parent's in-memory record.

The partial transcript of a failing background agent was invisible to any code walking `session.Messages`, and invisible to debug tooling inspecting the session store.

## Fix

Move `parent.AddSubSession(s)` unconditionally before the error gate, mirroring the pattern introduced in #3151.

## Known limitation

`runCollecting` has no `EventSink` parameter, so `SubSessionCompletedEvent` cannot be emitted here. The persistence observer pipeline (`PersistenceObserver.OnEvent`) is therefore not triggered and background-agent sub-sessions are still not written to the store. Fixing that requires threading `EventSink` through `RunParams` / the `Runner` interface and is left as a follow-up.

## Relationship to #3151

This PR is a companion to #3151 and should stack on top of it. The two PRs fix the same class of bug in the two sibling functions (`runForwarding` and `runCollecting`). They are separate because the `runCollecting` fix has a known limitation (no store persistence) that warrants independent review.

## Test plan

`TestRunAgentPersistsSubSessionOnError` mirrors the structure of `TestTransferTaskPersistsSubSessionOnError` from #3151. It constructs a background agent whose model always fails and asserts that `parent.Messages` contains a sub-session item after `RunAgent` returns an error — which fails without the fix.

### DIFF
--- a/pkg/runtime/agent_delegation.go
+++ b/pkg/runtime/agent_delegation.go
@@ -347,14 +347,22 @@ func (r *LocalRuntime) runCollecting(ctx context.Context, parent *session.Sessio
 	for range events {
 	}
 
+	// Persist the sub-session unconditionally — the partial transcript is
+	// the most valuable artifact for debugging a failed background agent.
+	// AddSubSession records it in-memory on both the success and error paths.
+	//
+	// Note: SubSessionCompletedEvent cannot be emitted here because
+	// runCollecting has no EventSink — the persistence observer pipeline
+	// (PersistenceObserver.OnEvent) is not triggered and the sub-session is
+	// not written to the store. Fixing that requires threading EventSink
+	// through RunParams / the Runner interface.
+	parent.AddSubSession(s)
+
 	if errMsg != "" {
 		return &agenttool.RunResult{ErrMsg: errMsg}
 	}
 
-	result := s.GetLastAssistantMessageContent()
-	parent.AddSubSession(s)
-
-	return &agenttool.RunResult{Result: result}
+	return &agenttool.RunResult{Result: s.GetLastAssistantMessageContent()}
 }
 
 // CurrentAgentSubAgentNames implements agenttool.Runner.
--- a/pkg/runtime/runtime_test.go
+++ b/pkg/runtime/runtime_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/docker/docker-agent/pkg/session"
 	"github.com/docker/docker-agent/pkg/team"
 	"github.com/docker/docker-agent/pkg/tools"
+	agenttool "github.com/docker/docker-agent/pkg/tools/builtin/agent"
 	mcptools "github.com/docker/docker-agent/pkg/tools/mcp"
 )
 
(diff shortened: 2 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly moves `parent.AddSubSession(s)` to run unconditionally before the error gate in `runCollecting`, mirroring the pattern already established in `runForwarding` (#3151). No bugs were found in the changed code.

**What looks good:**

- **`runCollecting` fix**: `parent.AddSubSession(s)` now always executes before the `errMsg != ""` check — the partial transcript is preserved on both the success and error paths. The inline comment accurately describes the known limitation (no `EventSink`, so `SubSessionCompletedEvent` is not emitted and the store is not written to).

- **`runForwarding` fix** (also in this diff): The refactor to capture `subSessionErr` and continue draining the channel is strictly better than the original early-return: it ensures all events (including late `StreamStoppedEvent` / notification events) reach the TUI sink, keeping `streamDepth` balanced and the user informed. The deliberate placement of `parent.ToolsApproved = s.ToolsApproved` **after** the error gate is a correct security improvement — a failed sub-session's tool approvals should not silently escalate the parent's permissions.

- **Tests**: Both `TestTransferTaskPersistsSubSessionOnError` and `TestRunAgentPersistsSubSessionOnError` are well-structured regression tests. They construct realistic failure scenarios, make targeted assertions on `sess.Messages`, and verify the required events appear in the sink.

**One minor observation (non-blocking):**

`TestTransferTaskPersistsSubSessionOnError` closes `evts` after `handleTaskTransfer` returns and then ranges over it to inspect events. This is safe today because `runForwarding` is fully synchronous (no goroutines outliving the function call), but would panic if `runForwarding` were ever made to emit events asynchronously after returning. Worth a brief comment in the test, but not a blocker.

docker-agent (commented):
### Assessment: 🟢 APPROVE

**PR**: fix(runtime): persist sub-session transcript on error path in `runCollecting`

The fix is correct and minimal. Moving `parent.AddSubSession(s)` unconditionally before the error gate mirrors the exact pattern used in the sibling `runForwarding` fix (#3151). On the error path, `errMsg` is set, the drain loop exits, and `AddSubSession` is now called before the early return — ensuring the partial transcript is always recorded in the parent's in-memory session.

**No high or medium severity issues found.** Two low-severity observations are noted below for informational purposes only (neither blocks merging).

**Logic correctness**: ✅ The reorder is safe. `errMsg` is populated from `ErrorEvent`, the drain loop exits, `AddSubSession` runs unconditionally, and then the error gate returns the correct result.

**Test quality**: ✅ `TestRunAgentPersistsSubSessionOnError` correctly models the failure path, asserts both the error surface (`require.NotEmpty(result.ErrMsg)`) and the sub-session presence in `sess.Messages`. Structure mirrors `TestTransferTaskPersistsSubSessionOnError` as stated in the PR description.

**Acknowledged limitation (documented)**: The comment in the code accurately describes that `SubSessionCompletedEvent` cannot be emitted here because `runCollecting` has no `EventSink` parameter. This means the `PersistenceObserver` pipeline is not triggered and the sub-session is not written to the store. This is a pre-existing limitation (not a regression introduced by this PR) and is appropriately flagged for follow-up work.

dgageot (approved):

PR #3154, opened 2026-06-17
fix(runtime): bound per-toolset tool listing during startup (#3137)

### LINKED ISSUE #3137: Tool loading hangs indefinitely when a toolset's Tools() blocks — sidebar "Loading tools…" never resolves and /quit shutdown delays
## Description

When a toolset's `Tools(ctx)` method blocks indefinitely (e.g., an LSP-based MCP server that fails to start or a stdio subprocess that hangs), the startup tool loading never completes. This causes two related problems:

1. **The sidebar permanently shows "Loading tools…"** because the final `ToolsetInfo` event with `Loading=false` is never emitted.
2. **The `/quit` exit path is delayed** by up to 30 seconds because `StopToolSets` has a timeout, and the process may not exit cleanly due to leaked goroutines.

## Root Cause

In [`pkg/runtime/runtime.go`](https://github.com/docker/docker-agent/blob/v1.79.0/pkg/runtime/runtime.go), the `emitToolsProgressively` function iterates over all toolsets and calls `toolset.Tools(ctx)` **without any per-toolset timeout**:

```go
func (r *LocalRuntime) emitToolsProgressively(...) {
    send(ToolsetInfo(0, true, agentName))  // ← tells sidebar "loading..."

    for _, toolset := range toolsets {
        ts, err := toolset.Tools(ctx)      // ← NO TIMEOUT — blocks forever if toolset hangs
        totalTools += len(ts)
        // ...
        send(ToolsetInfo(totalTools, !isLast, agentName))
    }

    send(ToolsetInfo(totalTools, false, agentName))  // ← never reached if blocked above
}
```

If a single toolset's `Tools()` call hangs:
- `emitToolsProgressively` never returns
- `ToolsetInfo(_, false, _)` is never sent, so the sidebar never updates
- The `startupEvents` channel in `app.New()` (via `EmitStartupInfo`) is never closed, leaking the reading goroutine
- On `/quit`, `stopToolSets(t)` is called in the deferred cleanup, which has a **30-second** `context.WithTimeout` (cmd/root/run.go ~line 1019). The process appears unresponsive during this window.

## Steps to Reproduce

1. Configure an agent YAML with a broken or unreachable MCP toolset (e.g., an LSP server that doesn't respond):
   ```yaml
   tools:
     my-lsp:
       mcp:
         command: npx
         args: ["-y", "@some/lsp-server", "--stdio"]
   ```
2. Run `docker agent run ./agent.yaml`
3. Observe the sidebar: "Loading tools…" stays forever.
4. Type `/quit` to exit.
5. Observe: the process hangs for ~30 seconds before exiting (or may appear to hang indefinitely if the LSP process doesn't respond to cancellation).

## Expected Behavior

1. Each toolset's `Tools()` call should have a reasonable timeout. If a toolset fails to enumerate its tools, it should be **skipped** with a warning log, not block the entire startup.
2. The sidebar should show the list of tools that loaded successfully, rather than an indefinite spinner.
3. `/quit` should exit promptly without waiting for hung toolset cleanup.

## Suggested Fix

1. **Add a per-toolset timeout** in `emitToolsProgressively`:

   ```go
   for i, toolset := range toolsets {
       toolCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
       ts, err := toolset.Tools(toolCtx)
       cancel()
       if err != nil {
           slog.WarnContext(ctx, "Toolset failed to load tools, skipping",
               "toolset", toolset.Name(), "error", err)
           continue  // skip this toolset, proceed to the next
       }
       totalTools += len(ts)
       send(ToolsetInfo(totalTools, !isLast, agentName))
   }
   ```

2. **Always send a terminal `ToolsetInfo`** even when some toolsets fail, so the sidebar resolves its loading state.

3. Consider reducing the `stopToolSets` timeout or making it respect context cancellation from the TUI shutdown signal.

## Environment

- **docker-agent version:** v1.79.0
- **OS:** Windows via WSL2 (Debian)
- **Go:** 1.26

Sayt-0:
Fixes #3137.

## what

`emitToolsProgressively` called `toolset.Tools(ctx)` with no timeout, so if a toolset's `Tools()` blocked forever (e.g. a wedged MCP stdio server that never answers `tools/list`), startup tool loading never finished. The terminal `ToolsetInfo{Loading:false}` was never sent, so the sidebar stayed on "Loading tools..." and `EmitStartupInfo`'s goroutine leaked.

This adds `listToolsWithTimeout`, which runs `Tools()` in a goroutine and bounds it with a per-toolset deadline (default 10s, configurable via `WithToolListTimeout`). I went with goroutine + select instead of a plain context timeout because the MCP client detaches ctx with `WithoutCancel`, so a wedged server might not honor cancellation. A timed-out toolset is skipped with a warning, and the terminal `ToolsetInfo` is always sent so the sidebar resolves and shows whatever loaded fine. The toolset is not torn down, so a slow-but-responsive one is listed again on the next turn.

## how it maps to the issue

the issue lists three expected behaviours:

1. per-toolset timeout, skip with a warning: done
2. sidebar shows the tools that loaded, not an infinite spinner: done
3. /quit exits promptly: mostly. the leaked startup goroutine that blocked it is gone now that `EmitStartupInfo` returns. i left the 30s `stopToolSets` bound alone since the issue lists it as "consider" and lowering it risks cutting off a legit graceful shutdown. happy to do it as a follow-up if you'd prefer.

## possible follow-up (not in this PR)

a server that wedges during `initialize`/`Start()` would stall startup the same way, and that path is also unbounded (`Connect` detaches ctx, the supervisor has no connect deadline). the same pattern would fix it. kept it out here to stay close to the issue's stated root cause (`Tools()`).

## test

added `TestEmitStartupInfo_SkipsToolsetWhoseListingHangs` (passes under `-race`). also reproduced by hand: hangs before, resolves after, using a hung toolset plus a tiny MCP stdio server that finishes initialize then never answers `tools/list`.

### DIFF
--- a/pkg/runtime/defaults.go
+++ b/pkg/runtime/defaults.go
@@ -30,3 +30,15 @@ const defaultMaxOverflowCompactions = 1
 // a notification from a server), so a slow or stuck server cannot be
 // allowed to wedge the caller indefinitely.
 const toolsChangedTimeout = 5 * time.Second
+
+// defaultToolListTimeout bounds how long EmitStartupInfo waits for a single
+// toolset to enumerate its tools while populating the sidebar. A toolset
+// whose Tools() blocks indefinitely — e.g. an MCP stdio subprocess that
+// never answers tools/list, possibly while also ignoring context
+// cancellation — would otherwise stall the whole startup: the terminal
+// ToolsetInfo{Loading:false} event is never emitted, so the sidebar stays
+// on "Loading tools…" forever and /quit appears to hang. A timed-out
+// toolset is skipped here; it stays usable for the actual agent turn, which
+// lists its tools again without this startup bound. Tests override it via
+// WithToolListTimeout to exercise the skip path without a real-time wait.
+const defaultToolListTimeout = 10 * time.Second
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -287,6 +287,12 @@ type LocalRuntime struct {
 	// succeeded" and "compaction exhausted" branches.
 	maxOverflowCompactions int
 
+	// toolListTimeout bounds how long EmitStartupInfo waits for a single
+	// toolset to enumerate its tools before skipping it, so one hung toolset
+	// cannot stall the sidebar's "Loading tools…" state forever. Defaults to
+	// defaultToolListTimeout; overridden via WithToolListTimeout.
+	toolListTimeout time.Duration
+
 	// pauseMu guards pauseCh.
 	pauseMu sync.Mutex
 	// pauseCh is non-nil and open while /pause has paused the run loop;
@@ -433,6 +439,19 @@ func WithMaxOverflowCompactions(n int) Opt {
 	}
 }
 
+// WithToolListTimeout overrides how long EmitStartupInfo waits for a single
+// toolset to enumerate its tools before skipping it. Defaults to
+// defaultToolListTimeout. A non-positive value is ignored so the default
+// stands. Tests pass a short timeout to exercise the skip path (a toolset
+// whose Tools() blocks) without a real-time wait.
+func WithToolListTimeout(d time.Duration) Opt {
+	return func(r *LocalRuntime) {
+		if d > 0 {
+			r.toolListTimeout = d
+		}
+	}
+}
+
 // WithRetryOnRateLimit enables automatic retry with backoff for HTTP 429 (rate limit)
 // errors when no fallback models are available. When enabled, the runtime will honor
 // the Retry-After header from the provider's response to determine wait time before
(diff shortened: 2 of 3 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27676729344).

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27699335024).

dgageot (approved):

PR #3157, opened 2026-06-17
fix(runtime): exempt list_background_agents from the loop-killer

mcavage-docker:
## Summary

The runtime loop detector exempts polling tools (`view_background_agent`, `view_background_job`) because models legitimately call them repeatedly while waiting for background tasks. `list_background_agents` was missing from that exempt set — even though:

- It's the natural reach-for tool when an agent has lost track of task IDs and needs to re-discover them.
- It's the **most** likely victim of the killer, because it takes zero arguments — every call is byte-identical, so the consecutive-duplicate counter trips fast.

Hit this in production: an agent waiting on a background research task fell into a `list_background_agents` poll loop and got terminated mid-turn. Five identical zero-arg calls is exactly the consecutive-duplicate killer's trigger pattern, but it's also the user's intent (poll status). The right fix is to mirror what we already do for the `view_*` tools.

## Changes

- `pkg/runtime/loop.go`: add `bgagent.ToolNameListBackgroundAgents` to the exempt list passed to `NewLoopDetector`, and update the surrounding comment.
- `pkg/runtime/toolexec/loop_detector.go`: update the `NewLoopDetector` docstring's example list to include `list_background_agents`.
- `pkg/runtime/toolexec/loop_detector_test.go`: add a regression test for repeated zero-arg `list_background_agents` calls (matches the existing `view_background_agent` / `view_background_job` exemption tests).

## Test plan

- [x] `go test ./pkg/runtime/toolexec/...` — passes
- [x] `go build ./pkg/runtime/...` — clean
- [ ] Reviewer confirms the exemption is the right call vs. the alternative (teaching agents to use `view_background_agent` instead). My read: both are right, but a heuristic safety net that punishes the natural status-check tool is a footgun.

### DIFF
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -303,16 +303,17 @@ func (r *LocalRuntime) runStreamLoop(ctx context.Context, sess *session.Session,
 	}
 
 	// Initialize consecutive duplicate tool call detector.
-	// Polling tools (view_background_agent, view_background_job) are
-	// expected to be called repeatedly with identical arguments while a
-	// background task is in progress. Exempt them so they never trigger
-	// the loop-termination path.
+	// Polling tools (view_background_agent, list_background_agents,
+	// view_background_job) are expected to be called repeatedly with
+	// identical arguments while a background task is in progress. Exempt
+	// them so they never trigger the loop-termination path.
 	loopThreshold := sess.MaxConsecutiveToolCalls
 	if loopThreshold == 0 {
 		loopThreshold = 5 // default: always active
 	}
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, well-scoped fix. The change correctly mirrors the existing exemption pattern for `view_background_agent` and `view_background_job` by adding `list_background_agents` to the loop-detector's exempt set. The constant name is consistent across production code and the new test. The regression test properly validates that repeated zero-arg `list_background_agents` calls don't trigger the loop killer.

dgageot (approved):

PR #3158, opened 2026-06-17
fix(providers): consume delta.reasoning in chat-completions stream adapter

### LINKED ISSUE #3145: Stream silently produces empty response when model outputs non-standard delta.reasoning SSE field (Qwen3 thinking mode)
## Description

When using a Qwen3-family model that streams thinking tokens via a non-standard delta.reasoning SSE field, Docker Agent's stream processor terminates in < 1ms and produces an empty response (tool_calls=0, content_length=0, stopped=true). The agent never executes any tool calls and returns no output.

## Expected Behavior

The agent calls list_directory and returns the result.

## Actual Behavior

The stream completes in < 1ms with no output:

```
level=DEBUG msg="OpenAI chat completion stream created successfully" model=Qwen3.5-397B-A17B
level=DEBUG msg="Processing stream" agent=root model=ovhcloud/Qwen3.5-397B-A17B
level=DEBUG msg="Stream processed" agent=root tool_calls=0 content_length=0 stopped=true
level=DEBUG msg="Skipping empty assistant message (no content and no tool calls)" agent=root
level=DEBUG msg="Conversation stopped" agent=root
```

## Steps to Reproduce

1. create a reproducer.yml file:
```yaml
providers:
  ovhcloud:
    api_type: openai_chatcompletions
    base_url: https://oai.endpoints.kepler.ai.cloud.ovh.net/v1
    token_key: OVH_AI_ENDPOINTS_ACCESS_TOKEN

models:
  ovh-qwen:
    provider: ovhcloud
    model: Qwen3.5-397B-A17B
    temperature: 0.6

agents:
  root:
    model: ovh-qwen
    instruction: |
      You are a helpful assistant.
    toolsets:
      - type: filesystem
```
ℹ️ This is reproducible without an API key, [OVHcloud AI Endpoints](https://www.ovhcloud.com/en/public-cloud/ai-endpoints/catalog/qwen-3-5-397b/) offers free access to this model at 2 req/s. ℹ️

2. run `docker agent run reproducer.yml "List the files in the current directory"`

 
## Environment

- Docker Agent version: v1.79.0
- OS: macOS Darwin 25.5.0
- Provider: api_type: openai_chatcompletions (OVHcloud AI Endpoints)
- Model: Qwen3.5-397B-A17B

## Error Output

```
level=DEBUG msg="OpenAI chat completion stream created successfully" model=Qwen3.5-397B-A17B
level=DEBUG msg="Processing stream" agent=root model=ovhcloud/Qwen3.5-397B-A17B
level=DEBUG msg="Stream processed" agent=root tool_calls=0 content_length=0 stopped=true
level=DEBUG msg="Skipping empty assistant message (no content and no tool calls)" agent=root
level=DEBUG msg="Conversation stopped" agent=root
```

## Screenshots

N/A

## Additional Context

Qwen3-family models in thinking mode stream reasoning tokens via a non-standard delta.reasoning field before producing any content or tool calls:

```{"choices":[{"index":0,"delta":{"role":"assistant","reasoning":"The user wants...","name":"assistant"}}]}```

Confirmed via curl: the model correctly outputs standard delta.tool_calls eventually, but only after all delta.reasoning chunks are done. Docker Agent's stream processor appears to fail silently on the unknown delta.reasoning field, terminating the stream before any tool calls or content is produced.

The standard delta.tool_calls format produced by the model is correct:
```{"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"id":"chatcmpl-tool-xxx","type":"function","function":{"name":"list_directory","arguments":""}}]}}]}```

### PR DESCRIPTION
**Pull request by Sayt-0**
Fixes #3145.

## TL;DR

Qwen3 thinking mode streams its thinking under a non-standard `delta.reasoning` SSE field. The shared OpenAI-compatible adapter only read `delta.reasoning_content`, so the thinking was dropped. A turn that **only reasons and then stops** ended up completely empty and silent.

## Symptom

```
Stream processed       tool_calls=0 content_length=0 stopped=true
Skipping empty assistant message (no content and no tool calls)
Conversation stopped
```

## Investigation

Hypotheses tested against the real OVHcloud Qwen3.5 stream:

| Hypothesis | Verdict |
|---|---|
| `delta.reasoning` breaks SSE/JSON decoding | ❌ reasoning+tool_call streams parse fine |
| SSE filter / HTTP client truncates the stream | ❌ body always complete (HTTP/1.0/1.1/2, byte-by-byte) |
| OVH `data: [DONE]`-only responses | ⚠️ real, but a rate-limit artifact, not this bug |
| **`delta.reasoning` silently dropped by the adapter** | ✅ **confirmed** |

Only a "reason, then stop" turn reproduces `content_length=0 / tool_calls=0 / stopped=true`.

## Root cause

| | |
|---|---|
| Field used by Qwen3 / OpenRouter / some vLLM/SGLang | `delta.reasoning` |
| Field read by docker-agent | `delta.reasoning_content` only |
| Result | reasoning dropped → empty assistant message → nothing shown |

## Fix

`pkg/model/provider/oaistream/adapter.go`: read `reasoning` in addition to `reasoning_content`.

- `reasoning_content` keeps priority when both are present
- non-string values (e.g. an object) are ignored

## Behaviour — same "reason, then stop" stream

| | Before | After |
|---|---|---|
| stdout | *(empty)* | thinking is streamed |
| reasoning captured | no | yes |

## Verify

`go test ./pkg/model/provider/oaistream/ -run TestStreamAdapter_ReasoningField` (new regression test, fails without the fix).

### DIFF
--- a/pkg/model/provider/oaistream/adapter.go
+++ b/pkg/model/provider/oaistream/adapter.go
@@ -67,15 +67,31 @@ func (a *StreamAdapter) Recv() (chat.MessageStreamResponse, error) {
 			a.lastFinishReason = finishReason
 		}
 
-		// Extract reasoning_content from ExtraFields since the OpenAI SDK
-		// does not yet have a dedicated field for it. Providers like DMR
-		// send reasoning tokens as a "reasoning_content" JSON field in the
-		// chat completion chunk delta.
+		// Extract reasoning text from ExtraFields since the OpenAI SDK does
+		// not have a dedicated field for it. OpenAI-compatible providers
+		// carry thinking tokens under different keys: DMR/DeepSeek use
+		// "reasoning_content", while Qwen3 thinking mode (e.g. via OVHcloud
+		// AI Endpoints), OpenRouter and several vLLM/SGLang builds use
+		// "reasoning". Without handling "reasoning", a Qwen3 turn that only
+		// reasons and then stops is dropped entirely and the user sees an
+		// empty reply. See https://github.com/docker/docker-agent/issues/3145.
 		var reasoningContent string
-		if ef, ok := choice.Delta.JSON.ExtraFields["reasoning_content"]; ok && ef.Raw() != "" {
-			// ef.Raw() returns the raw JSON value (e.g. `"some text"`), so
-			// we unmarshal it to get the plain Go string.
-			_ = json.Unmarshal([]byte(ef.Raw()), &reasoningContent)
+		for _, key := range []string{"reasoning_content", "reasoning"} {
+			ef, ok := choice.Delta.JSON.ExtraFields[key]
+			if !ok || ef.Raw() == "" {
+				continue
+			}
+			// ef.Raw() returns the raw JSON value (e.g. `"some text"`), so we
+			// unmarshal it to get the plain Go string. Some providers send a
+			// non-string "reasoning" (e.g. an object); ignore those and keep
+			// looking rather than surfacing a partial value.
+			if err := json.Unmarshal([]byte(ef.Raw()), &reasoningContent); err != nil {
+				reasoningContent = ""
+				continue
+			}
+			if reasoningContent != "" {
+				break
+			}
 		}
 
 		response.Choices[i] = chat.MessageStreamChoice{
--- a/pkg/model/provider/oaistream/adapter_test.go
+++ b/pkg/model/provider/oaistream/adapter_test.go
@@ -81,6 +81,53 @@ data: [DONE]
 	assert.ErrorIs(t, err, io.EOF)
 }
 
+// TestStreamAdapter_ReasoningField is a regression test for
+// https://github.com/docker/docker-agent/issues/3145. Qwen3 thinking mode (e.g.
+// via OVHcloud AI Endpoints), OpenRouter and several vLLM/SGLang builds stream
+// thinking tokens under a "reasoning" delta field (note the extra non-standard
+// "name" field too), not "reasoning_content". Before the fix the adapter only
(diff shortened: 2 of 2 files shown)


dgageot (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the Qwen3 `delta.reasoning` field being silently dropped by the `oaistream` adapter. The change is minimal, well-targeted, and the new regression test (`TestStreamAdapter_ReasoningField`) validates the fix.

**What was reviewed:**
- `pkg/model/provider/oaistream/adapter.go` — added `reasoning` as a fallback key with `reasoning_content` taking first-non-empty priority
- `pkg/model/provider/oaistream/adapter_test.go` — new test covering the `delta.reasoning` path

**No high or medium severity issues found.** A few minor observations (not blocking):

1. **Priority semantics** (`adapter.go`): The comment says `reasoning_content` has priority "when both are present," but the implementation uses *first-non-empty wins* — if a provider sends an explicit empty-string `reasoning_content` alongside a non-empty `reasoning`, the latter wins. This is arguably the right behavior (an empty string carries no useful content) but slightly inconsistent with the stated intent. Not a bug in practice.

2. **JSON `null` handling** (`adapter.go`): A `null` JSON value for either field unmarshal-succeeds to an empty string and falls through to the next key. This is harmless, but worth knowing: `null` and `""` are treated identically.

3. **Missing test cases** (`adapter_test.go`): No test for (a) both fields present in the same delta chunk, or (b) a non-string `reasoning` value (e.g. `{}`). These would strengthen confidence in the edge-case handling. Not required for this fix.

The core fix is sound and safe to merge.

PR #3159, opened 2026-06-17
fix(mcp): forward configured headers to OAuth discovery on the server host

### LINKED ISSUE #3148: Headers not passed to MCP ?
## Description

Following https://grafana.com/docs/grafana-cloud/machine-learning/assistant/configure/cloud-mcp/#client-configuration I tried to configure the MCP like this

```
grafana:
    remote:
      url: https://mcp.grafana.com/mcp
      transport_type: streamable
      headers:
        X-Grafana-URL: https://XXXXX.grafana.net/
```

It still ask for the grafana URL

## Expected Behavior

Based on the documentation it shouldn't

## Actual Behavior

The authentication screen ask for the instance URL

## Steps to Reproduce

1. Reference the MCP server as described above
2. Authenticate
3. The screen asks you the instance to use

## Environment

- **Docker Agent Version**: main
- **OS & Terminal**: macOS
- **Model Used**: sonnet

Sayt-0:
Fixes #3148

## Problem

A remote MCP server's `headers` were applied only to the main MCP channel, never to the OAuth flow. Grafana Cloud needs `X-Grafana-URL` on the OAuth discovery request to scope the flow to the right instance; without it the auth screen keeps prompting for the instance.

## Root cause

| Request | Transport used | Headers sent? |
|---|---|---|
| Main MCP channel | `oauthTransport.base` (header transport) | yes |
| OAuth discovery / token | `oauthHTTPClient` (bare singleton) | no |

The `protected-resource-metadata` request (the one Grafana uses to scope the instance) went out with none of the configured headers.

## Fix

- The OAuth HTTP client now forwards the configured headers...
- ...but only to the MCP server's own host. Requests to a third-party authorization server (URL taken from untrusted server metadata) get no headers, so `Authorization` / API keys can't leak.
- Preserved: SSRF protection, `allow_private_ips`, timeout, and `${headers.NAME}` placeholders. The shared client singleton is never mutated.

## Tests

| Test | Verifies |
|---|---|
| `...ScopesHeadersToMCPHost` | header reaches the MCP host, not a third party |
| `...NoHeadersReusesBaseClient` | no headers configured -> base client returned unchanged |
| `TestInitialize_CustomHeadersReachOAuthDiscovery` | real `Initialize()` path; reverting the fix makes it fail (reproduces #3148) |

Validation: `go build ./...`, `golangci-lint run ./pkg/tools/mcp/...` (0 issues), `go test -race ./pkg/tools/mcp/` all pass.

## Caveat

Scoping is on the exact host of the configured URL. If a server ever required the header on a host different from its MCP endpoint, the scope would need widening. Started strict, for safety.

### DIFF
--- a/pkg/tools/mcp/oauth_helpers.go
+++ b/pkg/tools/mcp/oauth_helpers.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/docker/docker-agent/pkg/browser"
 	"github.com/docker/docker-agent/pkg/httpclient"
+	"github.com/docker/docker-agent/pkg/upstream"
 )
 
 // oauthHTTPClient is the *http.Client used for outbound OAuth requests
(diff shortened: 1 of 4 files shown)


pkg/tools/mcp/oauth_helpers.go
```diff
@@ -40,6 +41,71 @@ func oauthHTTPClientForAllowPrivateIPs(allowPrivateIPs bool) *http.Client {
...
+// through base unchanged so configured credentials are never forwarded
+// off-host.
+type hostScopedHeaderTransport struct {
+	host        string
+	withHeaders http.RoundTripper
+	base        http.RoundTripper
+}
+
+func (t *hostScopedHeaderTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if strings.EqualFold(req.URL.Host, t.host) {
```
docker-agent: **[MEDIUM] Host comparison may silently drop headers when `rawURL` includes an explicit standard port**

`t.host` is set from `u.Host` where `u` is the parsed `rawURL`. Go's `url.Parse` preserves explicit ports verbatim, so `https://mcp.example.com:443/mcp` → `u.Host = "mcp.example.com:443"`. But the OAuth library will normally construct discovery requests using the normalized URL `https://mcp.example.com/...`, which yields `req.URL.Host = "mcp.example.com"` (no port). The `strings.EqualFold` comparison then fails and the headers are silently **not** forwarded — reproducing issue #3148 for users who happen to include an explicit standard port in their config.

Suggested fix — normalize the stored host using `net/http`'s own logic:

```go
import "net"

// strip the default port for the scheme so comparison is consistent
// with how Go's http package populates req.URL.Host
normalizeHost := func(host, scheme string) string {
    h, port, err := net.SplitHostPort(host)
    if err != nil {
        return host // no port present
    }
    if (scheme == "https" && port == "443") || (scheme == "http" && port == "80") {
        return h
    }
    return host
}

host: normalizeHost(u.Host, u.Scheme),
```

This is low-probability today (most users don't type `:443`) but the failure mode is silent — headers are dropped with no error, which is hard to debug.

docker-agent (commented):

One low-probability finding in newly added code: the host-scoping check may silently drop configured headers when the MCP server URL includes an explicit standard port (e.g., `:443`). The fix logic and security model (no credential leakage to third-party auth servers) are sound. All other security properties — SSRF protection, timeout, singleton non-mutation — are preserved correctly.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The host-scoping mechanism (`hostScopedHeaderTransport`) correctly forwards configured MCP server headers only to the MCP server's own host, preventing credential leakage to third-party authorization servers. The core security property is sound and well-tested.

**Minor observations** (no action required):

1. **`oauthHTTPClientWithHeaders` omits `Jar` from base client** (`oauth_helpers.go:82`) — The newly constructed `*http.Client` does not copy `base.Jar`. Today all base clients have a nil `Jar`, so this is harmless. If a cookie jar is ever attached to the base OAuth client in the future, it would silently disappear in the wrapped client.

2. **`req.URL.Host` may be empty on redirect requests** (`oauth_helpers.go:101`) — Go's HTTP client may reissue requests with `Request.URL.Host == ""` (host in `Request.Host`) on redirects. In that case `hostWithoutDefaultPort` returns `""`, the host check fails, and headers are silently dropped even if the redirect target is the MCP server itself. This only triggers on same-host redirects and is not the primary OAuth discovery path.

Both are low-risk edge cases that don't affect the primary use case described in this PR.

dgageot (approved):

Issue #1593, opened 2026-02-04
AWS Bedrock ValidationException: Protocol violation when assistant message contains both tool_calls and content


ThePlenkov:
**Summary:** cagent generates assistant messages with both tool_calls AND content, violating AWS Bedrock API protocol and causing ValidationException.

**Full technical details, reproduction steps, and suggested fix in first comment below.**

ThePlenkov:
## Full Technical Details with Reproduction Case

### Environment
- **cagent version**: Latest (via Docker)
- **OS**: Windows/Linux
- **Provider**: AWS Bedrock via LiteLLM proxy
- **Model**: `haiku` (Claude Haiku via bedrock/anthropic.claude-3-5-haiku-20241022-v1:0)
- **Setup**: Multi-agent coordinator with subagents using `transfer_task`

---

### The Problem

When cagent delegates tasks to subagents using `transfer_task`, it generates an assistant message that contains **BOTH** `tool_calls` AND `content` in the same turn. This violates the Anthropic/AWS Bedrock API protocol specification.

### Complete Error Message

```
error receiving from stream: POST "http://localhost:4000/chat/completions":
503 Service Unavailable {
  "message": "litellm.ServiceUnavailableError: litellm.MidStreamFallbackError: 
  litellm.APIConnectionError: APIConnectionError: OpenAIException - 
  Stream generation failed: ProviderException: (400, 
  \"AWS client error encountered: <class 'botocore.errorfactory.ValidationException'>: 
  An error occurred (ValidationException) when calling the ConverseStream operation 
  (reached max retries: 0): The number of toolResult blocks at messages.5.content 
  exceeds the number of toolUse blocks of previous turn.\"
}
```

---

### The Protocol Violation

Here's the **malformed message structure** that cagent generates (message index 5 in the conversation):

```json
{
  "role": "assistant",
  "content": "Let me proceed with analyzing the issue...",
  "tool_calls": [
    {
      "id": "tooluse_ABC123",
      "type": "function",
      "function": {
        "name": "transfer_task",
        "arguments": "{\"agent\": \"scout\", \"task\": \"Retrieve data from source A\", \"expected_output\": \"Summary of findings\"}"
      }
    },
    {
      "id": "tooluse_XYZ789",
      "type": "function",
      "function": {
        "name": "transfer_task",
        "arguments": "{\"agent\": \"coder\", \"task\": \"Analyze the implementation\", \"expected_output\": \"Technical analysis\"}"
      }
    }
[...]

PR #3162, opened 2026-06-18
fix(session): drop duplicate tool results in sanitizeToolCalls

Sayt-0:
## TL;DR

- Most of this is already fixed by #3001: `sanitizeToolCalls` drops orphaned tool results and injects placeholders for missing ones, keeping `tool_use`/`tool_result` balanced for strict providers (AWS Bedrock, including via a LiteLLM proxy). Related to #1676 and #1593.
- This PR is a small follow-up that closes the one variant #3001 did not cover: a duplicate `tool_result` for the same `tool_call_id`.

## Context

AWS Bedrock rejects any turn whose toolResult blocks outnumber the previous turn's toolUse blocks:

> The number of toolResult blocks ... exceeds the number of toolUse blocks of previous turn.

Known structural causes and coverage:

| Cause | Handled by | Status |
| --- | --- | --- |
| Result with no matching `tool_use` (orphan) | `sanitizeToolCalls` (#3001) | covered |
| Result for a `tool_use` that never produced one (missing) | placeholder injection (#3001) | covered |
| Second result for the same `tool_call_id` (duplicate) | this PR | new |

## Change

- Drop a `tool_result` whose `tool_call_id` already received a result; keep the first.
- ~3 lines in `sanitizeToolCalls`, same spirit as #3001.

## Test

- `TestSanitizeToolCalls_DropsDuplicateToolResult` (new)
- existing sanitize / orphan tests still pass

## Notes

- Duplicates can surface from provider or streaming quirks (tool-call deltas re-emitting an id), so this completes the guard rather than fixing a confirmed in-the-wild trigger.

### DIFF
--- a/pkg/session/sanitize_orphan_test.go
+++ b/pkg/session/sanitize_orphan_test.go
@@ -78,6 +78,37 @@ func TestSanitizeToolCalls_KeepsMissingResultBalanced(t *testing.T) {
 	assertToolPairingInvariant(t, out) // synthetic result injected
 }
 
+// TestSanitizeToolCalls_DropsDuplicateToolResult guards the remaining
+// toolResult/toolUse imbalance that orphan-dropping alone does not cover: a
+// second tool_result carrying the same tool_call_id. Bedrock counts it as an
+// extra toolResult block ("the number of toolResult blocks ... exceeds the
+// number of toolUse blocks of previous turn"), the same ValidationException
+// family as #1676 and #1593. Only the first result for a given tool_use should
+// survive.
+func TestSanitizeToolCalls_DropsDuplicateToolResult(t *testing.T) {
+	messages := []chat.Message{
+		{Role: chat.MessageRoleUser, Content: "hi"},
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change is correct and well-scoped. `sanitizeToolCalls` now drops duplicate `tool_result` messages for the same `tool_call_id` before the request reaches the provider, closing the last known structural cause of AWS Bedrock's toolResult/toolUse count mismatch (`ValidationException`).

**Production logic (session.go):** The new guard is placed after the existing orphan check and before `resultIDs` is updated, so it fires exactly when a second result for the same `tool_call_id` arrives within the same pending group. `flushPending` resets both maps, so the guard is naturally scoped to each turn — cross-turn duplicates are already handled as orphans. No logic errors found.

**Test (sanitize_orphan_test.go):** `TestSanitizeToolCalls_DropsDuplicateToolResult` exercises the new path end-to-end. The explicit `count != 1` assertion is the right guard for this case. The `assertToolPairingInvariant` helper is appropriately scoped to orphan detection and the test compensates for its narrower scope with the direct count check.

dgageot (approved):

PR #3163, opened 2026-06-18
feat(shell): opt-in sudo askpass flow (#1551)

### LINKED ISSUE #1551: Handling password protected commands
Sometimes the default agents will issue commands that require interactive user input, e.g. sudo somecommand.

After a few retries, the agent usually detects the command is hanging and falls back to providing manual instructions for the user to continue.

What would be an elegant way to support a one-time privilege escalation flow (e.g., prompt once for sudo approval/password, then continue non-interactively)?

Sayt-0:
Closes #1551.

## Problem

Shell commands that run `sudo` get no controlling terminal, so an interactive `sudo` hangs until the command times out and the agent falls back to printing manual instructions.

## Approach

Opt-in `sudo_askpass: true` on the `shell` toolset, bridging sudo's standard `SUDO_ASKPASS` mechanism to the existing elicitation prompt. Nothing changes unless the flag is set and a command actually runs `sudo`.

```yaml
toolsets:
  - type: shell
    sudo_askpass: true
```

Flow when a `sudo` command runs:

1. The toolset lazily starts a private unix-socket server in a `0700` temp dir, writes a `SUDO_ASKPASS` wrapper script, and injects the bridge env (`SUDO_ASKPASS`, `CAGENT_ASKPASS_SOCKET`, `CAGENT_ASKPASS_TOKEN`) only into that command.
2. A `sudo() { command sudo -A "$@"; }` function is prepended (POSIX shells) so sudo uses askpass instead of a TTY.
3. The hidden `__askpass` subcommand dials the socket; the server asks the user via a masked elicitation prompt and returns the password to sudo on stdout.

## Security

| Concern | Handling |
| --- | --- |
| Password exposure | Only transits the local socket and the helper stdout. Never logged, stored, or put on the command line. |
| Socket access | `0700` dir, single-user; request carries a `crypto/rand` token compared with `subtle.ConstantTimeCompare`. |
| Wrapper script | Executable path is single-quote escaped, so metacharacters cannot run as commands. |
| `__askpass` stdout | Forced onto the standalone command path even under the docker CLI plugin reexec env, so plugin usage output never corrupts the password sudo reads. |
| Fail-closed | Decline, cancel, error, headless, bad token, and missing env all make sudo abort, never a spurious success. |

## Limitations (documented)

- Unix only; no effect on Windows.
- Interactive runs only; non-interactive runs decline the prompt and sudo fails as before.
- Only a bare `sudo ...` in a POSIX shell is intercepted (`/usr/bin/sudo`, `env sudo`, nested scripts, and non-POSIX shells like `fish` are not).
- No cross-tool-call credential cache: each shell tool call is a fresh, TTY-less shell, so a prompt appears once per sudo-using command. Within a single command, sudo's own caching applies.

## Tests

- `pkg/tools/builtin/shell`: socket round-trip, decline, bad token, missing env, helper-death cancellation, `shQuote`, sudo wrapping, Elicitable discovery through the toolset wrappers.
- `pkg/config`: `sudo_askpass` validation (shell only) plus schema parity.
- `pkg/tui/dialog`: password field masked and not trimmed.
- `cmd/root`: `__askpass` routing and management-invocation gating.

Verified end to end with the built binary in both standalone and docker-plugin invocation, plus cross-compile for windows and js/wasm.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -2038,6 +2038,10 @@
           "type": "boolean",
           "description": "Opt in to dialling non-public IP addresses (valid for type 'fetch', 'api', 'openapi', 'a2a', and remote MCP toolsets). By default protected HTTP clients refuse connections \u2014 after DNS resolution, so DNS rebinding is also blocked \u2014 to loopback, RFC1918 private ranges, link-local (including the cloud metadata endpoint at 169.254.169.254), multicast and the unspecified address. Set this to true when an agent legitimately needs to call internal services. For fetch, 'allowed_domains' / 'blocked_domains' are evaluated independently and still apply."
         },
+        "sudo_askpass": {
+          "type": "boolean",
+          "description": "Opt in to a sudo privilege escalation flow for the shell toolset (only valid for type 'shell'). When enabled, sudo commands prompt the user for their password through the host UI via SUDO_ASKPASS; in non-interactive runs the prompt is declined automatically. Only a bare 'sudo ...' invocation in a POSIX shell is handled. No effect on Windows. Default false."
+        },
         "url": {
           "type": "string",
           "description": "URL for the a2a or openapi tool",
--- a/cmd/root/askpass.go
+++ b/cmd/root/askpass.go
@@ -1,0 +1,40 @@
+package root
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/docker/docker-agent/pkg/tools/builtin/shell"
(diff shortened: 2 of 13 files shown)


docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27763983496).

dgageot (approved):

PR #3167, opened 2026-06-19
feat: add adaptive RAG prefetching

### LINKED ISSUE #3164: feat: adaptive RAG prefetcher with topology-aware retrieval planning
## Summary

This proposal adds an adaptive RAG prefetcher and topology-aware retrieval planning layer for docker-agent. The idea comes from analyzing Aether-Lang's sparse attention graph, hierarchical block tree, adaptive epsilon governor, and drift detector, then translating the useful algorithms into docker-agent's Go RAG/runtime architecture without importing the Aether runtime.

## Motivation

Long-running agent sessions repeatedly query the same RAG sources, tool outputs, and nearby semantic regions. Today docker-agent indexes documents in the background and runs retrieval on demand, but it does not pre-warm likely next retrievals or adapt retrieval breadth based on topic drift. That leaves latency on the critical model/tool path and misses a chance to use existing turn signals.

## Proposed direction

- Add a small RAG prefetch subsystem that records recent user prompts, RAG queries, returned chunks, tool calls, and turn boundaries.
- Represent recent activity as lightweight topology metadata: centroid, radius, variance, concentration, and drift.
- Use an adaptive threshold to decide when the session is stable enough to prefetch nearby query candidates versus drifting enough to skip stale work.
- Warm query embeddings and optionally cached retrieval results for likely follow-up queries.
- Keep the feature opt-in and observable through debug logs/events/metrics.

## Candidate implementation surfaces

- `pkg/rag/manager.go`: coordinate prefetch lifecycle around real `Query` calls.
- `pkg/rag/strategy/vector_store.go`: add query embedding/result prewarm hooks without changing indexing semantics.
- `pkg/runtime/loop.go`: emit turn/user/tool signals to the prefetcher at safe boundaries.
- `pkg/config/latest/types.go` and `agent-schema.json`: add latest-only config.
- `docs/tools/rag/index.md` and examples: document the feature.

teerthsharma:
Closes #3164

## How this solves the issue

Issue #3164 proposed an adaptive RAG prefetcher that borrows the useful ideas from [Aether-Lang](https://github.com/teerthsharma/Aether-Lang), without importing Aether itself: topology metadata, adaptive drift detection, and conservative pre-warming for likely follow-up retrievals.

This PR implements that proposal in docker-agent's existing RAG architecture:

- Adds an opt-in `results.prefetch` configuration block so existing RAG behavior is unchanged by default.
- Adds `pkg/rag/prefetch`, a bounded in-memory prefetch layer that tracks normalized query shape, result count, average similarity, and centroid drift.
- Uses the drift score to decide when a session is stable enough to warm deterministic follow-up candidates, and suppresses background work during topic shifts.
- Serves exact repeated RAG queries from a cloned result cache, reducing repeated retrieval latency on the active path.
- Runs candidate prefetch in the background with timeout and in-flight de-duplication, so prefetch never blocks the foreground query.
- Wires prefetching at `pkg/rag/manager.go`, preserving the existing strategy, fusion, reranking, limit, full-content, and deduplication order.
- Clears cached state after explicit reindex checks to avoid holding stale results across index refreshes.
- Skips background candidate prefetch when reranking is enabled, keeping reranker error/fallback behavior tied to the user-visible foreground query.

## Issue acceptance criteria coverage

- Disabled/conservative by default: `results.prefetch.enabled` defaults to `false`.
- No Aether dependency: the PR translates the topology/drift idea into small Go primitives only.
- Bounded and non-blocking: cache size, candidate count, timeout, and in-flight de-duplication are all configurable.
- Drift invalidation: topology drift suppresses candidate generation when the query stream moves away from the current centroid.
- RAG manager integration: `Manager.Query` now checks the exact-query cache, stores final successful results, observes topology, and schedules background prefetches after cache misses.
- Documentation/example: RAG docs and `examples/rag/adaptive_prefetch.yaml` show the new config.

## Validation

- `go test ./pkg/rag/prefetch`
- `go test ./pkg/rag -run TestQueryUsesPrefetchCacheForRepeatedQuery`
- `go test ./pkg/config -run "TestJsonSchemaWorksForExamples|TestSchemaMatchesGoTypes"`
- `go run ./lint .`
- `go mod tidy --diff`
- `go build -v -o .\bin\docker-agent.exe .\main.go`

Local note: `task` and `golangci-lint` are not installed in this Windows environment, so `task lint`, `task test`, and `task build` could not run directly. Package-wide `go test ./...` on Windows still hits existing path/default-agent environment-sensitive tests unrelated to this branch.


Sayt-0:
Hello @teerthsharma ! The lint CI check is failing and blocking merge, could you take a look and fix it? Running task lint locally should help.

teerthsharma:
> Hello @teerthsharma ! The lint CI check is failing and blocking merge, could you take a look and fix it? Running task lint locally should help.

yep working on it

pkg/rag/prefetch/prefetch.go
```diff
@@ -0,0 +1,300 @@
...
+			break
+		}
+		if result.Similarity < p.cfg.MinSimilarity {
+			continue
+		}
+		name := sourceToken(result.Document.SourcePath)
+		if name == "" {
+			continue
+		}
+		candidate := strings.TrimSpace(base + " " + name)
```
Sayt-0: Candidates come out as `normalize(query) + space + filename-stem`, e.g. `rag manager` over `pkg/rag/manager.go` gives `rag manager manager`. Since `Get` needs an exact normalized match, a model-authored query would basically never emit that string, so this path seems to warm cache keys nothing later reads. A real-session number showing these candidates actually get hit would help.

pkg/rag/prefetch/prefetch.go
```diff
@@ -0,0 +1,300 @@
...
+	for i := range t.centroid {
+		t.centroid[i] += (point[i] - t.centroid[i]) * weight
+	}
+}
+
+func (t *tracker) stable(threshold float64) bool {
+	if t.seen < 2 {
+		return true
+	}
+	return t.drift <= threshold
```
Sayt-0: The drift gate looks like a near no-op at the defaults. `drift` is squared distance over `[len/256, tokens/32, results/32, avgSim]` against a cumulative-mean centroid (`weight := 1/seen` above), so it tracks distance from the all-time average rather than a recent shift, and sensitivity fades as `seen` grows. Crossing `drift_threshold: 0.8` needs roughly a 229-char or ~29-token jump in a single query, and the drifting test only passes by setting it to `0.0001`. An EWMA with a fixed alpha would track recent shifts better if this gate is meant to do anything.

pkg/rag/prefetch/prefetch.go
```diff
@@ -0,0 +1,300 @@
...
+	p.mu.Unlock()
+
+	go func() {
+		defer func() {
+			p.mu.Lock()
+			delete(p.inflight, key)
+			p.mu.Unlock()
+		}()
+
+		prefetchCtx, cancel := context.WithTimeout(ctx, p.cfg.Timeout)
```
Sayt-0: `prefetchCtx` derives from the foreground `Query` ctx, which is turn/request scoped (rag handler to dispatcher to loop to per-message ctx). It gets cancelled at turn end or on abort, so the background warm-up usually does not complete and an in-flight embedding call is aborted mid-way. `context.WithoutCancel(ctx)` is the pattern used for background work elsewhere (e.g. `pkg/tools/builtin/agent/agent.go`). Note the test uses `t.Context()`, which never cancels mid-flight, so it does not exercise this.

pkg/rag/manager.go
```diff
@@ -437,6 +455,7 @@ func (m *Manager) CheckAndReindexChangedFiles(ctx context.Context) error {
 			return fmt.Errorf("strategy %s failed: %w", strategyName, err)
 		}
 	}
+	m.prefetcher.Clear()
```
Sayt-0, line 492: This is the only `Clear()` call, and `Manager.CheckAndReindexChangedFiles` does not seem to have a non-test caller anywhere. The live refresh path is the watcher (`StartFileWatcher`, launched unconditionally in `pkg/tools/builtin/rag/rag.go`), which reindexes in place without clearing, and `Get` has no TTL or index-version check. So once prefetch is enabled, a cached query can keep returning pre-change results after a watched file updates, until it ages out of the window. That looks like a correctness regression. Maybe clear from the watcher reindex path, or observe an indexing-completed event.

docs/superpowers/plans/2026-06-18-adaptive-rag-prefetcher.md
```diff
@@ -0,0 +1,88 @@
+# Adaptive RAG Prefetcher Implementation Plan
```
Sayt-0: These `docs/superpowers/plans/*` and `docs/superpowers/specs/*` files look like agent planning/scaffolding (REQUIRED SUB-SKILL, DCO commit steps, etc.) rather than product docs, probably worth dropping from the diff.

Sayt-0 (changes requested):
Thanks for the PR. A few blocking-level doubts, happy to be wrong on any of these. For what it is worth, the exact-repeat cache idea on its own is reasonable, and the disabled-by-default no-op plus the clone-on-get/store are nicely careful. It is mostly the topology/candidate layer that seems like more than the use case needs.

Main points (details inline):

- **Adaptive layer may never fire.** Candidates are synthesized strings (`query + filename-stem`, e.g. `rag manager manager`) that an exact-match `Get` would basically never hit, and the drift gate looks like a no-op at the default `0.8` threshold. So the only part that can really fire is the plain exact-repeat cache.
- **Stale results once enabled.** Invalidation is wired to `CheckAndReindexChangedFiles`, which does not seem to have a non-test caller, while the file watcher reindexes in place without clearing. A cached query can keep returning pre-change results until it ages out of the window.
- **Background warm-up is tied to the request ctx.** Derived via `context.WithTimeout(ctx, ...)`, it usually gets cancelled at turn end and does not complete. `context.WithoutCancel` is the pattern used for background work elsewhere.
- **Repo hygiene.** `docs/superpowers/plans/*` and `docs/superpowers/specs/*` look like agent scaffolding rather than product docs.

A real-session benchmark showing actual cache hit-rate would help justify the complexity.

pkg/rag/manager.go
```diff
@@ -124,25 +127,56 @@ func New(_ context.Context, name string, config Config, strategyEvents <-chan ty
...
+		defer close(out)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case event, ok := <-in:
+				if !ok {
+					return
+				}
+				if event.Type == types.EventTypeIndexingComplete {
```
Sayt-0, line 173: Clear listens for EventTypeIndexingComplete ("indexing_complete"), but the live reindex never emits it: the vector store watcher emits the literal "indexing_completed" (trailing d) and the bm25 watcher emits no completion event, while CheckAndReindexChangedFiles still has no non-test caller. So Clear only runs once at initial indexing, when the cache is empty. Maybe hang it off the watcher reindex path, or add a TTL / index-version check on Get.

pkg/rag/prefetch/prefetch.go
```diff
@@ -0,0 +1,398 @@
...
+		return
+	}
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	clear(p.cache)
+	clear(p.inflight)
+	p.order = nil
+}
+
+func (p *Prefetcher) getTopologyLocked(key string) ([]database.SearchResult, bool) {
```
Sayt-0: On an exact miss this returns another query's cached results, with no rerank or re-score, so the scores belong to the other query. The gate (one filename-stem anchor + Jaccard >= 0.25) is loose, and the drift guard in front does not engage at the default threshold.

pkg/rag/prefetch/prefetch.go
```diff
@@ -0,0 +1,398 @@
...
+	return slices.Clone(results)
+}
+
+type tracker struct {
+	seen     int
+	centroid [4]float64
+	drift    float64
+}
+
+func (t *tracker) observe(query string, results []database.SearchResult) {
```
Sayt-0: EWMA fixes the shape but not the magnitude: the first drift is still the raw squared distance in the normalized point space, so 0.8 needs a ~229-char or ~29-token single-query jump, and the drifting test still forces 0.0001. A smaller threshold or different feature scale would make the gate actually engage.

pkg/rag/prefetch/proofs/TopologyHit.lean
```diff
@@ -0,0 +1,32 @@
+namespace RAGPrefetch
```
Sayt-0: These theorems are tautologies over opaque predicates and never reference the implementation, and the file is not built. The "prove" in the commit title oversells it. Probably worth dropping.

pkg/rag/prefetch/prefetch_test.go
```diff
@@ -0,0 +1,307 @@
...
+			continue
+		}
+		metrics.misses++
+		exactSeen[key] = struct{}{}
+		p.Store(turn.query, turn.results)
+	}
+	return metrics
+}
+
+func replayTrace() []replayTurn {
```
Sayt-0: The trace is hand-built so topology wins and the hit counts are hard-coded in the assertion, so this asserts the fixture rather than a real hit-rate. A trace captured from an actual session would be more convincing.

Sayt-0 (changes requested):
Thanks for the follow-up. The WithoutCancel fix and dropping the superpowers scaffolding look good, and the exact-repeat cache is the part that earns its keep.

Most of the earlier doubts still stand though, and the topology layer adds one. Blocking-level points (details inline):

- Invalidation still never fires on the live path: Clear listens for an event the watcher reindex does not emit, so a cached query can return stale results once enabled.
- Drift gate is still a no-op at the 0.8 default, so topic-shift suppression does not engage.
- The new topology Get can return another query's results, with no re-score.
- The Lean file proves trivial tautologies unrelated to the code, and the replay benchmark is a fixture shaped so topology wins, not a real hit-rate.

Net: the exact-repeat cache is fine on its own. The topology/drift/proof layer needs dropping or a real session hit-rate, and invalidation should hang off the watcher path (or a TTL on Get) before enabled is safe.

pkg/rag/prefetch/prefetch_test.go
teerthsharma: I will restructure a better translation of topology for rag for the next push

dgageot:
Let's close this PR for now. It looks like a way to advertise Aether-Lang.

teerthsharma:
Open to change wording or documentation style in any way possible I see it as a way to provide a different method then standard but I understand your decision

PR #3169, opened 2026-06-19
fix(modelsdev): skip models.dev fetch for custom providers (#3165)

### LINKED ISSUE #3165: Custom provider names trigger unexpected outbound lookup to models.dev, causing failures in internet-restricted environments
## Description
When defining a custom provider with an arbitrary name in the Docker Agent configuration, Docker Agent makes outbound HTTP requests to models.dev during model resolution. In network-restricted environments where direct internet access is not allowed, this causes timeouts or significant latency. Renaming the provider to a well-known name (e.g., openai) bypasses this behavior, suggesting that Docker Agent performs an external lookup only for unrecognized provider names.

## Expected Behavior
A custom provider defined with a base_url and token_key in the configuration should be used as-is, without any additional external HTTP calls to resolve provider metadata. The configuration should be entirely self-contained and work in air-gapped or internet-restricted environments.

## Actual Behavior
When the provider name is not a recognized built-in (e.g., mistral_gateway instead of openai), Docker Agent attempts to reach models.dev to resolve provider or model metadata. In environments where outbound internet access is blocked, this results in connection timeouts and severe latency, making the agent unusable.

## Steps to Reproduce
Define a custom provider with a non-standard name in your config.yaml
```
providers:
  mistral_gateway:
    base_url: https://your-internal-gateway.example.com/v1
    token_key: YOUR_API_KEY
models:
  my_model:
    provider: mistral_gateway
    model: Mistral:mistral-small-latest
```
Run Docker Agent in an environment where outbound HTTP to the internet (specifically models.dev) is blocked.
Observe connection timeouts or high latency when Docker Agent starts or processes the first request.
```
time=2026-06-18T15:35:53.095Z level=DEBUG msg="Failed to get model definition" error="failed to fetch from API and no cached data available: failed to fetch from API: Get \"https://models.dev/api.json\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
```
Rename the provider from mistral_gateway to openai (keeping base_url and token_key identical) and restart.
Observe that the agent starts and responds correctly with no external calls.

## Environment
Docker Agent Version: 
docker-agent version v1.79.0
Commit: a2c0cc8e880cbc6060ff6729f82236ecbbaa9ae2
OS & Terminal: Kubernetes pod
Model Used: Mistral:mistral-small-latest (via internal Gravitee API Gateway proxying Mistral)

Sayt-0:
## Problem

A custom provider — a name that is not a built-in/alias (e.g. `mistral_gateway`) defined with `base_url` + `token_key` — triggered a blocking `GET https://models.dev/api.json` (full ~3.4 MB catalog) during model resolution. In internet-restricted environments this hangs ~30s on every turn, making a self-contained custom-provider config unusable. Renaming to a known provider (e.g. `openai`) bypassed it.

Closes #3165.

## Fix

`modelsdev.Store` gains `WithKnownProvider`: a provider the predicate rejects is served cache-only and never fetches (missing cache → empty catalog → clean "provider not found"). The runtime's default store (`lazyModelStore`) and `RuntimeConfig.ModelsDevStore()` are wired with `provider.IsKnownProvider`, so only catalog providers (built-ins + aliases) can reach the network.

| Path | Before | After |
|---|---|---|
| `loop.go` `GetModel` (every turn) | fetch catalog for any provider | custom → no fetch |
| `session_compaction.go` `resolveContextLimit` | ~30s hang air-gapped | resolves instantly |
| known provider (openai/anthropic/bedrock…) | fetch | unchanged |

## Verification

Real binary, cold cache, custom-provider config from the issue:

| Binary | models.dev catalog fetched |
|---|---|
| `origin/main` | yes (~3.4 MB download) |
| this PR | no |

- `TestLazyModelStore_GatesCustomProvider` (runtime) reproduces the exact issue error without the fix, passes with it.
- `modelsdev` gate unit tests; `go build` / `vet` / `golangci-lint` / `-race` green.

## Known limitations (follow-ups, out of scope)

- TUI model picker / `models` CLI still fetch the full catalog (not the request-loop path).
- A custom provider without `provider_opts.context_size` has no context limit, so compaction stays disabled for it (inference unaffected — it routes via `base_url`).

### DIFF
--- a/pkg/config/runtime.go
+++ b/pkg/config/runtime.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
+	"github.com/docker/docker-agent/pkg/model/provider"
 	"github.com/docker/docker-agent/pkg/modelsdev"
 )
 
@@ -101,7 +102,9 @@ func (runConfig *RuntimeConfig) ModelsDevStore() (*modelsdev.Store, error) {
 		return runConfig.ModelsDevStoreOverride, nil
 	}
 	runConfig.modelsDevStoreOnce.Do(func() {
-		runConfig.modelsDevStore, runConfig.modelsDevStoreErr = modelsdev.NewStore()
+		runConfig.modelsDevStore, runConfig.modelsDevStoreErr = modelsdev.NewStore(
+			modelsdev.WithKnownProvider(provider.IsKnownProvider),
+		)
 	})
 	return runConfig.modelsDevStore, runConfig.modelsDevStoreErr
 }
--- a/pkg/modelsdev/store.go
+++ b/pkg/modelsdev/store.go
@@ -34,16 +34,25 @@ const (
 // The database is loaded on first access via GetDatabase and
 // then cached in memory for the lifetime of the Store.
 type Store struct {
-	cacheFile string
-	mu        sync.Mutex
-	db        *Database
+	cacheFile     string
+	knownProvider func(string) bool
+	mu            sync.Mutex
+	// db is the authoritative catalog from the full (fetch-eligible) load path.
+	db *Database
+	// cacheDB is a cache-only snapshot served to fetch-disallowed lookups. It is
+	// kept separate from db because it may be empty or stale and must never
+	// satisfy a later fetch-eligible lookup for a known provider; memoizing it
+	// keeps the hot path from re-reading and re-parsing the catalog file on
+	// every custom-provider resolution.
+	cacheDB *Database
 }
 
 // Opt configures a Store created with NewStore.
 type Opt func(*storeOptions)
 
 type storeOptions struct {
-	cacheFile string
+	cacheFile     string
+	knownProvider func(string) bool
 }
 
 // WithCache overrides the path of the on-disk cache file used by the Store.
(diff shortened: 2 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is well-scoped, correct, and thoroughly tested.

**What the PR does:** Adds a `WithKnownProvider` predicate to `modelsdev.Store` so that custom (non-catalog) providers are served from cache only and never trigger an outbound `GET models.dev/api.json`. The `lazyModelStore` and `RuntimeConfig.ModelsDevStore()` are both wired with `provider.IsKnownProvider`.

**Correctness review:**
- The mutex protocol is sound: `getDatabase` is called while holding `s.mu` only from within `GetDatabase`/`getProvider`, both of which acquire the lock before calling into `getDatabase`. No re-entrant locking issue.
- The `allowFetch=false` path deliberately skips memoization of `s.db` to avoid poisoning future fetch-eligible lookups for known providers — this is correct and clearly commented.
- `NewDatabaseStore` leaves `knownProvider` nil (fetch-eligible for all providers), which is intentional: it is used in tests with an in-memory database, not a live fetch path.
- The `loadDatabase` stale-cache logic change (`!allowFetch || time.Since(...) < refreshInterval`) correctly short-circuits on the first `err == nil` check when fetching is disallowed, so the secondary `if !allowFetch` block only handles the cache-miss case — logically consistent.
- `GetDatabase` (public API) always passes `allowFetch=true`, preserving backward compatibility for direct callers (e.g., TUI model picker, `models` CLI) that explicitly want the full catalog.
- `WithKnownProvider(nil)` stores `nil`, which is handled correctly: `s.knownProvider == nil` → `allowFetch = true` (same as no predicate).

**Tests:** `TestKnownProviderGatesFetch`, `TestNoPredicateAlwaysAllowsFetch`, and `TestLazyModelStore_GatesCustomProvider` together cover the regression, backward compatibility, and production store path. The use of an already-expired context deadline as a stand-in for "network unreachable" is a clean, deterministic approach.

dgageot (approved):

PR #3171, opened 2026-06-19
feat(embeddedchat): add headless chat session API

dgageot:
## Summary
- add `pkg/embeddedchat`, a headless wrapper for embedding docker-agent runtime conversations in non-docker-agent UIs
- support loading an agent source, streaming assistant text/tool events, confirming tool calls, restarting conversations, and closing active runs
- cover loading, streaming, confirmation, elicitation/max-iteration handling, error handling, concurrent send rejection, and cancellation semantics

## Tests
- `go test ./pkg/embeddedchat ./pkg/app ./pkg/runtime ./pkg/teamloader`

### DIFF
--- a/pkg/embeddedchat/embeddedchat.go
+++ b/pkg/embeddedchat/embeddedchat.go
@@ -78,9 +78,9 @@ type ToolActivity struct {
 
 // runtimeRunner is the subset of runtime.Runtime the headless session needs.
 type runtimeRunner interface {
-	RunStream(context.Context, *session.Session) <-chan dagentruntime.Event
-	Resume(context.Context, dagentruntime.ResumeRequest)
-	ResumeElicitation(context.Context, tools.ElicitationAction, map[string]any) error
+	RunStream(ctx context.Context, sess *session.Session) <-chan dagentruntime.Event
+	Resume(ctx context.Context, req dagentruntime.ResumeRequest)
+	ResumeElicitation(ctx context.Context, action tools.ElicitationAction, content map[string]any) error
 	Close() error
 }
 
--- a/pkg/embeddedchat/embeddedchat.go
+++ b/pkg/embeddedchat/embeddedchat.go
@@ -1,0 +1,296 @@
+// Package embeddedchat provides a small headless chat wrapper around the
+// docker-agent runtime for embedders that want to drive an agent from their
+// own UI instead of running docker-agent's Bubble Tea application.
+package embeddedchat
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+
+	dagentcfg "github.com/docker/docker-agent/pkg/config"
+	dagentruntime "github.com/docker/docker-agent/pkg/runtime"
+	"github.com/docker/docker-agent/pkg/session"
+	"github.com/docker/docker-agent/pkg/teamloader"
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+// Config describes an embedded agent session.
+type Config struct {
+	// AgentSource is the agent/team definition to load. Bytes sources are a
+	// good fit for embedders that ship a pinned agent in their binary.
+	AgentSource dagentcfg.Source
+	// RuntimeConfig is passed to the team loader. When nil, a zero runtime
+	// config is used.
+	RuntimeConfig *dagentcfg.RuntimeConfig
+	// ToolsetRegistry resolves toolsets declared by AgentSource. When nil,
+	// docker-agent's default registry is used.
+	ToolsetRegistry teamloader.ToolsetRegistry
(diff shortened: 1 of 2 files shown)


docker-agent (commented):

Found **2 medium** and **1 low** severity issues in the new `pkg/embeddedchat` package. All three are in the `Send`/`forwardEvents` concurrency path. The overall design is solid — the mutex/runID guard pattern is sound and the drain-on-cancel approach is correct in principle.

| Severity | Count |
|----------|-------|
| 🟡 Medium | 2 |
| ⚪ Low | 1 |

pkg/embeddedchat/embeddedchat.go
```diff
@@ -0,0 +1,296 @@
...
+		s.activeCancel = nil
+	}
+}
+
+// Send appends prompt to the conversation and streams the assistant reply.
+// The returned channel is closed after either a Done event or an Err event.
+// If ctx is cancelled, Send drains the runtime stream until it stops, but no
+// further events are delivered to the caller.
+func (s *Session) Send(ctx context.Context, prompt string) (<-chan Event, error) {
+	if s.rt == nil || s.session == nil {
```
docker-agent, line 219: **[MEDIUM] Data race: `s.session` read outside mutex before nil-check**

`s.session` is read at line 179 before the mutex is acquired at line 183. `Restart()` reassigns `s.session` under the mutex via `resetConversationLocked()`. A concurrent `Restart()` creates a Go data race — an unsynchronized read on one goroutine against a write on another, which is undefined behavior per the Go memory model.

While `resetConversationLocked` assigns a non-nil value (not nil), the race detector will still flag this. The guard should be moved inside the lock:

```go
s.mu.Lock()
if s.rt == nil || s.session == nil {
    s.mu.Unlock()
    return nil, errors.New("embeddedchat: session is not initialized")
}
// ... rest of lock-held logic
```

Note: `s.rt` is set once in `New()` and never reassigned, so the race on that field is only theoretical.

pkg/embeddedchat/embeddedchat.go
```diff
@@ -0,0 +1,296 @@
...
+		case out <- e:
+			return true
+		case <-ctx.Done():
+			return false
+		}
+	}
+
+	errSent := false
+	for event := range events {
+		if ctx.Err() != nil {
```
docker-agent, line 282: **[MEDIUM] Potential goroutine leak: cancelled context with pending `ToolCallConfirmation` not resumed**

When `ctx` is cancelled, the drain loop at line 232 hits `continue` for every remaining event — including `ToolCallConfirmationEvent`. This means `s.rt.Resume(...)` is never called for a pending confirmation.

If the runtime's `events` channel stays open until it receives a `Resume` call, and the runtime does not independently close the channel on context cancellation, `forwardEvents` will loop forever on a channel that never closes.

The `runCtx` (derived from `ctx`) is cancelled and passed to `RunStream`, so a well-implemented runtime _should_ close `events` when the context is done. However, the current code provides no safety net if that assumption is violated. Consider calling `Resume` for pending confirmations in the drain path:

```go
for event := range events {
    if ctx.Err() != nil {
        // Unblock any pending tool confirmation so the runtime can shut down
        if _, ok := event.(*dagentruntime.ToolCallConfirmationEvent); ok {
            s.rt.Resume(context.Background(), dagentruntime.ResumeReject("The run was aborted."))
        }
        continue
    }
    // ...
```

pkg/embeddedchat/embeddedchat.go
```diff
@@ -0,0 +1,296 @@
...
+			continue
+		}
+
+		switch e := event.(type) {
+		case *dagentruntime.ToolCallConfirmationEvent:
+			if errSent {
+				s.rt.Resume(ctx, dagentruntime.ResumeReject("The run was aborted."))
```
docker-agent, line 289: **[LOW] `Resume` called with potentially-cancelled context after error**

After an `ErrorEvent` sets `errSent = true`, subsequent `ToolCallConfirmationEvent`s call `s.rt.Resume(ctx, dagentruntime.ResumeReject(...))` using the caller's `ctx`. If that context is already cancelled (or becomes cancelled before `Resume` executes), the call may silently no-op depending on the runtime's implementation — leaving the runtime blocked on an unresolved confirmation.

Consider using `context.Background()` (or a detached context with a short deadline) here to ensure the rejection is always delivered regardless of the caller's context state:

```go
if errSent {
    s.rt.Resume(context.Background(), dagentruntime.ResumeReject("The run was aborted."))
    continue
}
```

The same applies to the `ResumeElicitation` call and the `MaxIterationsReachedEvent` branch.

trungutt (approved):

PR #3172, opened 2026-06-19
refactor: replace go-memoize and go-cache with internal memoize package

dgageot:
The codebase depended on two external memoization libraries: `github.com/kofalt/go-memoize` for single-flight de-duplication and `github.com/patrickmn/go-cache` for TTL-based caching. Both roles are now served by a small new internal package `pkg/memoize`, built on `golang.org/x/sync/singleflight`, eliminating unnecessary external dependencies.

The new `Memoizer[T]` type is generic, type-safe, and preserves all required semantics: single-flight de-duplication, TTL expiry, error non-caching, panic propagation, and correct handling of the `ttl==0` edge case (now treated as never-expire, matching go-cache's behavior). Expired entries are evicted lazily. Three call sites—`pkg/desktop/uuid.go`, `pkg/desktop/version.go`, and `pkg/remote/transport.go`—were migrated to the new API, removing `interface{}` boxing and type assertions in the process. Unit tests cover 100% of the memoize package and pass under `-race`.

This PR also fixes a pre-existing test isolation bug in `pkg/tui/components/reasoningblock` where two tests (`TestReasoningBlockCollapsedByDefaultFromSessionState` and `TestReasoningBlockCanDefaultExpandedFromSessionState`) read the developer's real `~/.config/cagent/config.yaml` via `userconfig.Get()`, causing spurious failures when `expand_thinking: true` is set. A new `useEmptyUserConfig` helper points paths to a clean temp directory for test isolation; those two tests are no longer parallelized since they mutate the global config-dir override.

Validation: `task build`, `task lint` (0 issues), and `task test` all pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,6 @@ require (
 	github.com/hashicorp/hcl/v2 v2.24.0
 	github.com/junegunn/fzf v0.73.1
 	github.com/k3a/html2text v1.4.0
-	github.com/kofalt/go-memoize v0.0.0-20240506050413-9e5eb99a0f2a
 	github.com/labstack/echo/v4 v4.15.4
 	github.com/mattn/go-isatty v0.0.22
 	github.com/mattn/go-runewidth v0.0.24
@@ -100,6 +99,7 @@ require (
 	github.com/mtibben/percent v0.2.1 // indirect
 	github.com/pb33f/jsonpath v0.8.2 // indirect
 	github.com/pb33f/ordered-map/v2 v2.3.1 // indirect
+	github.com/smartystreets/assertions v1.2.0 // indirect
 	github.com/standard-webhooks/standard-webhooks/libraries v0.0.1 // indirect
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.63.0 // indirect
 	golang.org/x/mod v0.36.0 // indirect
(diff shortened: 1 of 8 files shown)


gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed `pkg/memoize/memoize.go` (new generic `Memoizer[T]`), all three migrated call-sites (`pkg/desktop/uuid.go`, `pkg/desktop/version.go`, `pkg/remote/transport.go`), and the test-isolation fix in `pkg/tui/components/reasoningblock/reasoningblock_test.go`.

**What was checked:**
- TTL handling: `store()` uses `if m.ttl > 0` to set expiry — both `ttl ≤ 0` and `NoExpiration (-1)` correctly produce a zero `expires` (never-expire), matching the old `go-cache` behavior.
- Singleflight semantics: the double-check inside `group.Do` prevents redundant `fn()` calls after concurrent misses populate the cache. Error non-caching is preserved — errors are returned to all concurrent waiters without being stored.
- Type assertion safety: `result, _ := v.(T)` with the comma-ok form handles nil interface values correctly (covered by `TestMemoizeNilInterfaceValue`).
- Panic propagation: `singleflight` re-panics to all waiters; the test confirms the key is not cached after a panic.
- Call-site migration: all three sites preserve prior semantics (`NoExpiration` for UUID, 5-minute TTL for version, 1-minute TTL for transport) with no behavioral regression.
- Test isolation: `useEmptyUserConfig` correctly uses `atomic.Pointer`-backed `paths.SetConfigDir` with `t.Cleanup` reset; cross-package interference is impossible since each package's tests run in a separate binary.
- Dependency graph: `smartystreets/assertions` promoted to an explicit indirect dep is correct `go mod tidy` output (transitive path via `k3a/html2text → goconvey`).

PR #3173, opened 2026-06-19
fix(runtime): close cost-accounting blind spots (reasoning tokens, $0 spend leaks)

dgageot:
## Summary

Fixes three genuine cost-accounting/observability bugs in the agent runtime, surfaced by a cost analysis of a delegating multi-agent workload. Each fix is an isolated commit with a regression test that fails before the change.

All fixes are **behavior-preserving for existing cost numbers** — they add missing data/signals rather than changing the cost formula.

### `fix(anthropic): record reasoning tokens from streaming usage`
Extended-thinking tokens are billed inside `OutputTokens` but were never surfaced as `chat.Usage.ReasoningTokens`, so the `/cost` dialog and usage dashboards showed **zero reasoning for Claude** while OpenAI and Gemini reported it. Maps `OutputTokensDetails.ThinkingTokens` (a read-only decomposition of the already-billed output tokens) on both the standard and Beta streaming paths. Purely additive observability — the cost formula is unchanged. Extracted `usageFromDelta` / `betaUsageFromDelta` so the mapping is unit-testable without standing up an SSE stream.

### `fix(runtime): warn when an uncatalogued model bills $0 with token usage`
Per-message cost is computed only when the model is in the models.dev catalogue and carries a price table; otherwise it is silently `0` even though real tokens were spent. That makes an uncatalogued model look free, so any spend guardrail built on the recorded cost never trips and the leak is invisible. Extracts `computeMessageCost`, which now reports whether pricing was available, and emits a warning when tokens were spent but no price was found. The cost arithmetic is unchanged.

### `fix(runtime): persist run_background_agent sub-sessions to the store`
Background-agent sub-sessions were added to the parent's in-memory object but never written to the session store: `runCollecting` runs on a detached goroutine with no `EventSink`, so the `PersistenceObserver` never saw a `SubSessionCompleted` event. Their tokens and cost were therefore recorded nowhere durable — **$0 for work that actually ran**, escaping any spend accounting that reads the store. Persists the completed sub-session directly via the (concurrency-safe) store on both the success and error paths, using a non-cancellable context so a stopped task still records its transcript. Writing to the store directly rather than through the shared observer chain avoids racing the parent's live `RunStream`.

## Testing

- `task test` — full suite green
- `golangci-lint run ./...` — 0 issues
- Each fix verified red→green by reverting the production change with the new test in place
- Runtime + anthropic packages pass under `-race`

Rebased on latest `main`.

### DIFF
--- a/pkg/model/provider/anthropic/adapter.go
+++ b/pkg/model/provider/anthropic/adapter.go
@@ -143,18 +143,28 @@ func (a *streamAdapter) Recv() (chat.MessageStreamResponse, error) {
 	case anthropic.MessageDeltaEvent:
 		a.stopReason = eventVariant.Delta.StopReason
 		if a.trackUsage {
-			response.Usage = &chat.Usage{
-				InputTokens:       eventVariant.Usage.InputTokens,
-				OutputTokens:      eventVariant.Usage.OutputTokens,
-				CachedInputTokens: eventVariant.Usage.CacheReadInputTokens,
-				CacheWriteTokens:  eventVariant.Usage.CacheCreationInputTokens,
-			}
+			response.Usage = usageFromDelta(eventVariant.Usage)
 		}
 	case anthropic.MessageStopEvent:
 		response.Choices[0].FinishReason = finishReason(a.stopReason, a.toolCall)
 	}
 
 	return response, nil
+}
+
+// usageFromDelta maps the standard Messages API streaming usage onto chat.Usage.
+// ReasoningTokens comes from OutputTokensDetails.ThinkingTokens, which Anthropic
+// reports as a read-only decomposition of OutputTokens (thinking is already
+// billed inside output), so surfacing it never changes the cost computed in the
+// runtime — it only makes the otherwise-invisible thinking spend observable.
+func usageFromDelta(u anthropic.MessageDeltaUsage) *chat.Usage {
+	return &chat.Usage{
+		InputTokens:       u.InputTokens,
+		OutputTokens:      u.OutputTokens,
+		CachedInputTokens: u.CacheReadInputTokens,
+		CacheWriteTokens:  u.CacheCreationInputTokens,
+		ReasoningTokens:   u.OutputTokensDetails.ThinkingTokens,
+	}
 }
 
 // finishReason maps an Anthropic stop reason (received on message_delta) onto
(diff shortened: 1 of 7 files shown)


melmennaoui (dismissed):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The three cost-observability fixes are well-scoped, behavior-preserving, and correctly implemented. Each change is accompanied by a regression test. No high or medium severity bugs were found in the changed code.

**Summary of analysis:**

- **`fix(anthropic): record reasoning tokens from streaming usage`** — The `usageFromDelta` / `betaUsageFromDelta` extraction is clean. `ThinkingTokens` is correctly mapped to `ReasoningTokens` as a read-only decomposition of already-billed `OutputTokens`. No cost-formula impact.

- **`fix(runtime): warn when an uncatalogued model bills $0 with token usage`** — `computeMessageCost` and `usageHasTokens` are well-extracted. The `slog.Warn` correctly fires only when `usageHasTokens` is true and `!priced`. The `priced` bool is derived from whether `m != nil && m.Cost != nil` (implicitly, via `computeMessageCost`'s return). Logic is sound.

- **`fix(runtime): persist run_background_agent sub-sessions to the store`** — `persistBackgroundSubSession` uses `context.WithoutCancel` so a stopped task still writes its transcript. The event channel is fully drained before the persist call (the `for range events {}` loop at line 360 runs on both success and early-cancel paths), so no partially-written session state. `parent.ID` is an immutable field; safe to read from the background goroutine.

- **`test(reasoningblock): isolate user config`** — `paths.SetConfigDir` to a temp dir is the correct established pattern; fixes a legitimate test isolation flake.

**Minor observations (not actionable):**
- The `slog.Warn` for uncatalogued models (loop.go:834) logs `input_tokens` and `output_tokens` but not `reasoning_tokens`. For Anthropic thinking models, reasoning tokens are billed inside `output_tokens` anyway, so the omission doesn't affect spend accounting — but including it would aid future debugging.
- `usageHasTokens` (loop.go:895) does not check `ReasoningTokens`, which is consistent with Anthropic's billing model (thinking is billed inside `OutputTokens`). If that model changes, the check would need updating.

simonferquel (dismissed):

simonferquel (approved):

PR #3176, opened 2026-06-19
feat(provider): make openai, anthropic, google, and amazon-bedrock optional

dgageot:
Projects embedding docker-agent can now drop model providers they don't need, shrinking both the binary and transitive dependency graph. The largest win is dropping the AWS SDK (`aws-sdk-go-v2`) via the `docker_agent_no_bedrock` tag, which eliminates the entire Bedrock integration and its 61-package dependency tree.

The monolithic `pkg/model/provider/factory.go` was split into a shared dispatch core plus one self-registering file per optional provider, each gated by a build tag. Providers register into the dispatch map via `init()`, and a model whose provider was compiled out fails at construction time with a clear "not compiled into this build" error. All providers are included by default; pass a tag to opt out of unwanted ones.

Build flags (all included by default; use tags to exclude):

- `docker_agent_no_openai` — drops `github.com/openai/openai-go`
- `docker_agent_no_anthropic` — drops `github.com/anthropics/anthropic-sdk-go`
- `docker_agent_no_google` — drops `google.golang.org/genai` and Vertex AI / cloud auth stack
- `docker_agent_no_bedrock` — drops the `github.com/aws/aws-sdk-go-v2` stack

Example: `go build -tags 'docker_agent_no_bedrock docker_agent_no_openai' ...`

Tags are prefixed with `docker_agent_` because Go build tags are global to a build (not scoped per module), so a bare `no_openai` would collide with an embedding project's own tags. The rationale and full tag list are documented in the package doc via `go doc` / pkg.go.dev.

One caveat: anthropic and openai SDKs are also imported by the google provider's Vertex Model Garden support, so they only fully drop when combined with `docker_agent_no_google`. The `openai-go` and `genai` packages also appear elsewhere (`dmr/oaistream`, `chatserver`, `pkg/a2a`), so those won't leave the module graph entirely. The dmr provider and rule-based router remain always compiled in. Builds pass with every tag combination and without any tags; js/wasm still builds; lint is clean.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The provider opt-out refactoring is well-structured. The split from a monolithic `factory.go` into per-provider self-registering files (`factory_openai.go`, `factory_anthropic.go`, `factory_google.go`, `factory_bedrock.go`) is clean and correct. Build tag syntax (`//go:build !docker_agent_no_<provider>`) is syntactically valid. The `providerFactories` map is initialized before `init()` runs, and the `registerProviderFactory` panic-on-duplicate acts as an intentional safety net against misconfigured builds. All four providers are included by default; the `docker_agent_` prefix on tags avoids collision with embedding projects.

**Minor hygiene notes** (no action required):
- A few test files (`custom_provider_test.go`, etc.) lack build-tag guards matching their factory dependencies. This is harmless today since none of those tests call `New()` with a provider-specific path, but adding guards would make the intent explicit.
- `schema_test.go` always imports SDK sub-packages regardless of build tags, so test binaries won't fully tree-shake even when provider tags are set — this is expected given those sub-packages appear elsewhere in the module.

trungutt (approved):

PR #3177, opened 2026-06-19
refactor: replace bleve with lightweight BM25 matcher for model routing

dgageot:
The rule-based model router previously used the bleve full-text search library, which pulled in a large transitive dependency tree (zapx, go-faiss, geo, gtreap, mmap-go, scorch, and others) and crucially, could not cross-compile to WebAssembly due to its reliance on file I/O and memory mapping.

This change replaces bleve with a lightweight, pure-Go BM25 ranker that tokenizes example phrases, drops English stop words, indexes them by route, and returns the best-scoring match. At roughly 150 lines of code, it does what the router needs without pulling in heavy external dependencies or blocking WASM support.

The implementation includes proper BM25 scoring with IDF weighting, maintains a running total of document lengths for efficiency, and is thoroughly tested with unit tests covering tokenization, stop-word filtering, deduplication, best-route selection, and edge cases. The matcher is integrated into the rule-based client and now enables rule-based routing to work under js/wasm for the first time.


trungutt (dismissed):

pkg/model/provider/rulebased/matcher.go
```diff
@@ -0,0 +1,153 @@
...
+
+	termFreqs := make(map[string]int, len(terms))
+	for _, term := range terms {
+		termFreqs[term]++
+	}
+	for term := range termFreqs {
+		m.docFreq[term]++
+	}
+
+	m.docs = append(m.docs, document{
```
docker-agent, line 57: **[MEDIUM] Zero-length documents silently skew BM25 scoring and can cause NaN**

When `tokenize(text)` returns `nil` — which happens when an example phrase consists entirely of stop words (e.g. `"the and is"`) or is empty — `len(terms) == 0`, yet the document is still appended to `m.docs` with `length: 0`. This has two consequences:

1. **avgLen drift**: `totalLen` is unchanged but `len(m.docs)` grows, so `avgLen` is driven toward zero with each zero-length example added. For short routing phrases (typical 2–5 word examples), a handful of empty docs measurably skews the BM25 length normalization for all other documents.

2. **NaN propagation (edge case)**: If *all* indexed examples produce zero-length documents, `avgLen == 0.0`. In `score()`, line 97: `bm25B * float64(doc.length) / m.avgLen` becomes `0.75 * 0 / 0.0` which is `NaN` per IEEE 754. `NaN > 0.0` is `false`, so `found` stays `false` — routing silently falls back for every query with no error logged.

**Suggested fix**: Add an early return in `add()` when the example produces no terms:
```go
func (m *matcher) add(routeIndex int, text string) {
    terms := tokenize(text)
    if len(terms) == 0 {
        return // skip empty examples; they carry no routing signal
    }
    // ... rest of function unchanged
}
```

docker-agent (commented):

This PR cleanly replaces the heavy bleve dependency with a ~150-line pure-Go BM25 matcher, enabling js/wasm support. The implementation is well-structured and the test coverage is thorough. Two medium-severity edge cases were found in the BM25 scoring logic related to zero-length documents (examples that are entirely stop words).

pkg/model/provider/rulebased/matcher.go
```diff
@@ -0,0 +1,153 @@
...
+	var score float64
+	for _, term := range queryTerms {
+		freq, present := doc.termFreqs[term]
+		if !present {
+			continue
+		}
+
+		idf := math.Log(1 + (n-float64(m.docFreq[term])+0.5)/(float64(m.docFreq[term])+0.5))
+		tf := float64(freq)
+		norm := tf * (bm25K1 + 1) / (tf + bm25K1*(1-bm25B+bm25B*float64(doc.length)/m.avgLen))
```
docker-agent, line 103: **[MEDIUM] Division by zero (NaN) in score() when avgLen reaches 0.0**

In the BM25 norm formula on this line:
```go
norm := tf * (bm25K1 + 1) / (tf + bm25K1*(1-bm25B+bm25B*float64(doc.length)/m.avgLen))
```

If `m.avgLen == 0.0` (which occurs when all indexed examples are zero-length after stop-word filtering), `float64(doc.length) / m.avgLen` evaluates to `NaN` (IEEE 754: `0 / 0`). The `NaN` propagates through the entire expression, making `score` `NaN`, and `NaN > bestScore` is always `false`. The result is that `bestRoute` always returns `(0, false)` — silently routing every query to the fallback with no warning.

A guard on `m.avgLen` would make this safe:
```go
var lenNorm float64
if m.avgLen > 0 {
    lenNorm = bm25B * float64(doc.length) / m.avgLen
}
norm := tf * (bm25K1 + 1) / (tf + bm25K1*(1-bm25B+lenNorm))
```

The root fix (skipping zero-length documents in `add()`) also resolves this.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The bleve → BM25 replacement is clean, correct, and well-tested. No bugs were found.

**What was reviewed:**
- `pkg/model/provider/rulebased/matcher.go` — new BM25 ranker (highest-risk file)
- `pkg/model/provider/rulebased/client.go` — integration of the new matcher
- `pkg/model/provider/factory_js.go` — WASM enablement of rule-based routing
- `go.mod` / `go.sum` — bleve and transitive dependency removals

**BM25 implementation correctness:**
- IDF uses the smoothed Robertson-Walker variant (`log(1 + ...)`) which is always non-negative — correct.
- TF normalization formula matches the standard BM25 definition with k1=1.2, b=0.75 (textbook defaults).
- The `found` flag (`score > 0.0`) is sound: any term match produces a strictly-positive score, so there are no false negatives.
- `avgLen` division-by-zero is unreachable: stop-word-only documents produce zero-length term lists, so `docFreq` is never populated for them; query terms from such documents never match, meaning `score=0` and `found=false` before the NaN path is reached.

**Thread safety:** The matcher is write-only during `NewClient` construction and read-only thereafter. No concurrent writes occur; the existing `sync.RWMutex` on the `Client` guards `lastSelectedID` as before.

**WASM compatibility:** All bleve imports (`mmap-go`, `go-faiss`, `bbolt`, `zapx`) are removed. The new implementation is pure Go with no file I/O, making js/wasm compilation viable as described.

**Tests:** Unit tests cover tokenization, stop-word filtering, deduplication, BM25 scoring, best-route selection, edge cases (empty matcher, empty query, all-stop-word query), and the rare-term-wins IDF property.

dgageot:
Addressed the zero-length-document edge cases in 510d66706.

`matcher.add()` now early-returns when `tokenize()` yields no terms (empty or all-stop-word examples). This is the root fix for both comments:

- **avgLen skew**: empty docs are no longer appended, so they can't drag `avgLen` toward zero.
- **NaN propagation**: with no zero-length docs indexed, `avgLen` is only ever 0 when the matcher is empty (handled by the existing `len(m.docs) == 0` guard in `bestRoute`), so the `doc.length / avgLen` division in `score()` can never hit 0/0.

Added two tests: `TestMatcher_SkipsEmptyExamples` (empty/stop-word examples aren't indexed and avgLen stays correct) and `TestMatcher_AllEmptyExamples` (all-empty examples fall through to fallback instead of a NaN-driven false match). Tests pass under `-race`, the js/wasm build compiles, and lint is clean.

trungutt (approved):

gtardif (approved):

PR #3178, opened 2026-06-19
fix(modelinfo): offer the max effort tier in the Shift+Tab thinking cycle

### LINKED ISSUE #3100: Shift+Tab thinking cycle hides the max effort tier on Opus 4.7/4.8 (and Sonnet 4.6, Fable 5) — single-top-tier gating can't express xhigh + max
## Summary

The interactive **Shift+Tab thinking-level cycle** (TUI model picker) hides the `max` effort tier for several Claude models that actually support it — most visibly **Claude Opus 4.7 / 4.8**. The cycle offers `off / low / medium / high / xhigh` and never reaches `max`.

This is a **gating bug only**: authored `thinking_budget` values (e.g. `adaptive/max`) are still sent correctly to the API. It affects only the user-selectable Shift+Tab cycle.

## Reproduction

1. Configure an agent:
   ```yaml
   models:
     anthropic-large:
       provider: anthropic
       model: claude-opus-4-8
       thinking_budget: adaptive/max
   ```
2. Run the TUI; the header correctly shows `thinking auto adaptive`.
3. Press **Shift+Tab** to cycle thinking levels.
4. **Observed:** cycle is `off → low → medium → high → xhigh → (wrap)`. `max` is never offered.
5. **Expected:** `max` is offered, since Anthropic lists `max` as supported on Opus 4.8.

## Root cause

`anthropicTopEffort()` in `pkg/modelinfo/thinking_levels.go` returns a **single** highest effort `Level`, implicitly assuming the selectable tiers form a linear ladder. For Opus 4.7+ it returns `XHigh`:

```go
switch {
case minor >= 7:
    return effort.XHigh
case minor == 6:
    return effort.Max
default:
    return ""
}
```

`thinkingLevelMap()` then declares only that one tier as supported:

```go
m := effort.LevelMap{effort.Minimal: false}
if top := anthropicTopEffort(modelID); top != "" {
    m[top] = true
}
```

Because `Max` is an **explicit-only** tier (`explicitOnlyLevels = {XHigh, Max}` in `pkg/effort/effort.go`) and the map only declares `XHigh: true`, `SupportedLevels()` drops `Max`. Net result for Opus 4.7/4.8: `off / low / medium / high / xhigh`.

The deeper problem: **`xhigh` and `max` are independent capabilities, not a single ladder.** A single returned "top tier" cannot represent a model that supports *both* `xhigh` and `max` (Opus 4.7/4.8, Fable 5), nor distinguish it from a model that supports `max` *without* `xhigh` (Opus 4.6, Sonnet 4.6).

## Authoritative reference

Anthropic — **Effort**: https://platform.claude.com/docs/en/build-with-claude/effort

Per-level model availability, quoted from that page:

- **`max`** — Claude Fable 5, Claude Mythos 5, **Claude Opus 4.8**, Claude Mythos Preview, **Claude Opus 4.7**, Claude Opus 4.6, **Claude Sonnet 4.6**
- **`xhigh`** — Claude Fable 5, Claude Mythos 5, **Claude Opus 4.8**, **Claude Opus 4.7**

Resulting capability matrix:

| Model | low | medium | high | xhigh | max |
|---|---|---|---|---|---|
| Opus 4.5 | ✓ | ✓ | ✓ | — | — |
| Sonnet 4.6 | ✓ | ✓ | ✓ | — | ✓ |
| Opus 4.6 | ✓ | ✓ | ✓ | — | ✓ |
| Opus 4.7 | ✓ | ✓ | ✓ | ✓ | ✓ |
| Opus 4.8 | ✓ | ✓ | ✓ | ✓ | ✓ |
| Fable 5 | ✓ | ✓ | ✓ | ✓ | ✓ |

## Affected models (cycle gating incorrect)

- **Opus 4.7, Opus 4.8** — `max` missing (cycle caps at `xhigh`)
- **Sonnet 4.6** — `max` missing (cycle caps at `high`)
- **Fable 5** — `max` missing (cycle caps at `xhigh`)
- **Mythos 5 / Mythos Preview** — unrecognized (cycle caps at `high`)

Correct today: Opus 4.6 (`max`, no `xhigh`) and Opus 4.5 (`high`).

## Not affected (scope clarification)

- The request path in `pkg/model/provider/anthropic/thinking.go` sends `adaptive/<effort>` verbatim (e.g. effort `max`), so authored configs already behave correctly.
- `RejectsTokenThinking()` correctly coerces Opus 4.6/4.7/4.8 onto adaptive thinking.
- `agent-schema.json` already accepts `max` / `adaptive/max`.

## Suggested fix

- Replace the single-`Level` return of `anthropicTopEffort()` with independent capability flags (e.g. `supportsXHigh`, `supportsMax`), and have `thinkingLevelMap()` set both `effort.XHigh` and `effort.Max` in the `LevelMap` where applicable.
- Add table-driven tests covering each Claude tier (Opus 4.5 / 4.6 / 4.7 / 4.8, Sonnet 4.6, Fable 5, and the Mythos variants).

**Relevant files:**
- `pkg/modelinfo/thinking_levels.go`
- `pkg/effort/effort.go`
- `pkg/runtime/model_switcher.go` (consumer of `SupportedThinkingLevels`)

Sayt-0:
## What

The TUI **Shift+Tab thinking-level cycle** hid the `max` effort tier on several Claude models that support it — Opus 4.7/4.8 capped at `xhigh`, Sonnet 4.6 capped at `high`, and Fable 5 capped at `xhigh`.

Root cause: `anthropicTopEffort()` returned a **single** highest tier, implicitly treating the selectable tiers as a linear ladder. But `xhigh` and `max` are **independent capabilities**, not a ladder: Opus 4.6 / Sonnet 4.6 accept `max` *without* `xhigh`, while Opus 4.7+ / Fable 5 / Mythos 5 accept *both*. A single returned tier cannot express that, and because `max` is an explicit-only tier it was dropped by `effort.SupportedLevels()`.

This is a **selector-gating fix only**: the request path (`pkg/model/provider/anthropic/thinking.go`) already sends `adaptive/<effort>` verbatim, so authored `thinking_budget` values were never affected.

## Fix

- Replace `anthropicTopEffort` (one tier) with `anthropicTopEfforts`, which returns the full `{xhigh, max}` subset a model supports.
- Recognise the **Sonnet** family (4.6 → `max`) — previously only Opus/Fable were handled.
- Handle **Mythos** ids best-effort (no catalogue entry exists yet): every variant → `max`, the full release adds `xhigh`, the preview tops out at `max`.

## Capability matrix (now matches [Anthropic's effort docs](https://platform.claude.com/docs/en/build-with-claude/effort))

| Model | low | medium | high | xhigh | max |
|---|:---:|:---:|:---:|:---:|:---:|
| Opus 4.5 | ✓ | ✓ | ✓ | — | — |
| Sonnet 4.5 / Haiku | ✓ | ✓ | ✓ | — | — |
| Sonnet 4.6 | ✓ | ✓ | ✓ | — | ✓ |
| Opus 4.6 | ✓ | ✓ | ✓ | — | ✓ |
| Opus 4.7 | ✓ | ✓ | ✓ | ✓ | ✓ |
| Opus 4.8 | ✓ | ✓ | ✓ | ✓ | ✓ |
| Fable 5 | ✓ | ✓ | ✓ | ✓ | ✓ |
| Mythos 5 | ✓ | ✓ | ✓ | ✓ | ✓ |
| Mythos Preview | ✓ | ✓ | ✓ | — | ✓ |

## Tests

- **Unit** (`TestSupportedThinkingLevels`, `TestAnthropicTopEfforts`): all matrix models + bedrock-prefixed / dotted / date-stamped variants.
- **End-to-end** (`TestCycleAgentThinkingLevel_PerModelTopTier`): walks a full Shift+Tab cycle for all 9 models — the automated equivalent of the manual repro.

Verified: `gofmt` clean, `golangci-lint` 0 issues, `go vet` clean, all tests green.

## Out of scope (follow-ups surfaced by review)

- Docs (`docs/providers/anthropic/index.md`, `docs/guides/thinking/index.md`) still describe effort-based adaptive thinking as "Opus 4.6+ only" — predates this PR and concerns the API request path, not the cycle.
- `RejectsTokenThinking` enumerates only Opus 4.6/4.7/4.8 — the issue explicitly scopes that path as *not affected*; whether Sonnet 4.6 / Fable / Mythos reject token budgets needs authoritative confirmation before changing.

Fixes #3100

### DIFF
--- a/pkg/modelinfo/thinking_levels.go
+++ b/pkg/modelinfo/thinking_levels.go
@@ -25,7 +25,7 @@ func thinkingLevelMap(provider, modelID string) effort.LevelMap {
 		// The Anthropic effort scale starts at low ([effort.ForAnthropic]
 		// maps minimal onto low), so offering minimal would duplicate low.
 		m := effort.LevelMap{effort.Minimal: false}
-		if top := anthropicTopEffort(modelID); top != "" {
+		for _, top := range anthropicTopEfforts(modelID) {
 			m[top] = true
 		}
 		return m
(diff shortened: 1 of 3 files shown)


dgageot (approved):

pkg/modelinfo/thinking_levels.go
```diff
@@ -60,39 +60,92 @@ func providerFamily(providerType string) string {
...
+// date-stamped 4.0 ids such as "claude-opus-4-20250514", whose long digit run
+// is a date, not a minor version.
+func claudeFamilyMinor(m string) (family string, minor int, ok bool) {
+	for _, fam := range []string{"opus", "sonnet"} {
+		_, rest, found := strings.Cut(m, fam+"-4")
+		if !found {
+			continue
+		}
+		if rest == "" || (rest[0] != '-' && rest[0] != '.') {
+			return "", 0, false
```
docker-agent, line 140: **[LOW] Loop guard uses `return` instead of `continue`, preventing subsequent family checks**

The two guard branches inside the `for range` loop call `return "", 0, false` rather than `continue`. This means: if a model ID contains an `"opus"` substring that partially matches `fam+"-4"` but then fails the separator or digit-width check, the function exits entirely without trying `"sonnet"`.

For all real Claude IDs today this is harmless — each ID contains exactly one family name. But the structural issue means future composite or synthetic IDs would be silently mis-classified. Using `continue` would match the intent of the loop and be more defensive:

```go
if rest == "" || (rest[0] != '-' && rest[0] != '.') {
    continue  // try next family instead of returning
}
minor, width := leadingInt(rest[1:])
if width == 0 || width > 2 {
    continue  // same
}
```

Not blocking — no current model IDs are affected.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly replaces the single-return `anthropicTopEffort` with `anthropicTopEfforts` (returning a slice), properly handling the `xhigh` and `max` capability independence, adding Sonnet 4.6 and Mythos support. The capability matrix logic and tests are thorough and cover the documented cases well.

**One low-severity structural note** (no blocking issues):

In `claudeFamilyMinor`, the guard branches inside the `for range` loop use `return "", 0, false` instead of `continue`:

```go
if rest == "" || (rest[0] != '-' && rest[0] != '.') {
    return "", 0, false  // exits loop entirely instead of trying next family
}
minor, width := leadingInt(rest[1:])
if width == 0 || width > 2 {
    return "", 0, false  // same — stops after first family match attempt
}
```

If a model ID contains an `opus` prefix substring that fails the format guards, the `sonnet` family is never tried. In practice, real Claude IDs are single-family, so this is not a regression. A `continue` would make the intent explicit and guard against future ambiguity. Not blocking.

The test coverage is excellent — unit tests cover all 9 capability-matrix models plus Bedrock-prefixed, dotted, and date-stamped variants, and the new end-to-end `TestCycleAgentThinkingLevel_PerModelTopTier` is a solid regression guard for the Shift+Tab cycle.

PR #3180, opened 2026-06-19
Expose ParseTheme for in-memory/embedded themes

EronWright:
## What

Adds a public `styles.ParseTheme([]byte) (*Theme, error)` that decodes a theme
from raw YAML and merges it onto the default theme.

## Why

Themes are already partial overrides on top of `DefaultTheme()` — `loadBuiltinTheme`
and `loadThemeFrom` both do `mergeTheme(DefaultTheme(), &override)`, which is why a
bundled theme YAML only declares the colors it changes. That merge is the useful
part, but today it is reachable only through `LoadTheme`, which requires the theme
to exist on disk under a ref.

Downstream embedders that build a branded CLI/TUI on top of cagent want to ship a
custom theme compiled into the binary. Without a byte-level entrypoint the only
option is to unmarshal directly into a `Theme` and `ApplyTheme` it — which bypasses
the merge, so every unspecified field becomes a zero value. `ParseTheme` exposes the
existing merge behavior for in-memory/embedded themes, with no file round-trip.

It does not touch the theme cache and leaves `Ref` unset, since an in-memory theme
has no reference to reload from.

## Embed use case

```go
import (
    _ "embed"

    "github.com/docker/docker-agent/pkg/tui/styles"
)

//go:embed branded.yaml
var brandedTheme []byte

func applyBrandTheme() error {
    theme, err := styles.ParseTheme(brandedTheme)
    if err != nil {
        return err
    }
    styles.ApplyTheme(theme)
    return nil
}
```

## Example partial theme (overrides defaults)

Only the listed fields change; everything else is inherited from `DefaultTheme()`.

```yaml
# branded.yaml — a partial theme; unset fields fall back to the default theme.
name: Branded
colors:
  accent: "#FF5C00"        # override the accent
  brand: "#FF5C00"
  background: "#101014"    # override the background
markdown:
  heading: "#FF5C00"
# text_primary, success, error, chroma.*, etc. are NOT set here,
# so they keep the default theme's values.
```

## Test

`TestParseTheme_PartialOverrideMergesOntoDefault` asserts an overridden field takes
the new value while unspecified fields (`background`, `success`, `markdown.heading`)
match `DefaultTheme()`. Also covers name defaulting and invalid YAML.

### DIFF
--- a/pkg/tui/styles/theme.go
+++ b/pkg/tui/styles/theme.go
@@ -490,6 +490,30 @@ func LoadTheme(ref string) (*Theme, error) {
 	return theme, nil
 }
 
+// ParseTheme decodes a theme from raw YAML bytes and merges it onto the default
+// theme, so a partial theme need only specify the fields it wants to override.
+// This is the same merge behavior used when loading built-in and user themes
+// (see loadBuiltinTheme / loadThemeFrom), exposed for in-memory or embedded
+// themes that are not backed by a file on disk.
+//
+// Unlike LoadTheme, ParseTheme does not consult the theme cache and leaves Ref
+// unset, since an in-memory theme has no reference to reload from.
+func ParseTheme(data []byte) (*Theme, error) {
+	var override Theme
+	if err := yaml.Unmarshal(data, &override); err != nil {
+		return nil, fmt.Errorf("parsing theme: %w", err)
+	}
+
+	// mergeTheme inherits the default's Name when the override omits one, which
+	// would mislabel an unnamed custom theme as "Default". Name it explicitly.
+	merged := mergeTheme(DefaultTheme(), &override)
+	if override.Name == "" {
+		merged.Name = "custom"
+	}
+
+	return merged, nil
+}
+
 // getUserThemeFileInfo returns the path and modTime of a user theme file if it exists.
 // Returns empty path and zero time if the file doesn't exist.
 func getUserThemeFileInfo(ref string) (path string, modTime time.Time) {
--- a/pkg/tui/styles/theme_test.go
+++ b/pkg/tui/styles/theme_test.go
@@ -41,6 +41,43 @@ func TestDefaultTheme(t *testing.T) {
 	assert.NotEmpty(t, theme.Markdown.Heading)
 }
 
+func TestParseTheme_PartialOverrideMergesOntoDefault(t *testing.T) {
+	t.Parallel()
+
+	def := DefaultTheme()
+
+	// A partial theme that overrides only the accent color.
+	data := []byte("name: Branded\ncolors:\n  accent: \"#FF0000\"\n")
+
+	theme, err := ParseTheme(data)
+	require.NoError(t, err)
+	require.NotNil(t, theme)
+
+	// Overridden field takes the new value.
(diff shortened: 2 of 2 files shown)


rumpl (approved):

Sayt-0 (commented):
Nice work on this, the embed use case is well thought out and the merge semantics are right. One small bug worth fixing (verified against the head commit).

**`ParseTheme` returns `Ref = "default"` instead of leaving it unset, which contradicts its own docstring.**

In `pkg/tui/styles/theme.go`:

- The docstring (lines 499-500) promises ParseTheme "leaves Ref unset".
- But `merged := mergeTheme(DefaultTheme(), &override)` (line 509) inherits the base's `Ref`. `mergeTheme` does `result := *base` (line 618) and only overwrites `Version`/`Name`/`Colors`/`Chroma`/`Markdown`, never `Ref` (lines 617-633). `DefaultTheme()` sets `Ref = "default"` (line 204), so it flows straight through.
- `loadBuiltinTheme` (line 549) and `loadThemeFrom` (line 607) both fix this with `merged.Ref = ref` after the merge; `ParseTheme` just omits that line.

Impact after `ApplyTheme(ParseTheme(...))`:

- **Picker mislabels the active theme:** `handleOpenThemePicker` (`handlers.go:561-577`) sets `IsCurrent: ref == CurrentTheme().Ref`, and `CurrentTheme().Ref == "default"`, so the built-in default row shows as current instead of the branded one.
- **Cancel-preview discards the custom theme:** `originalRef` is `"default"`, so cancelling a preview runs `handleThemeCancelPreview` then `ApplyThemeRef("default")` (`handlers.go:623`), reloading the built-in default and silently dropping the in-memory theme.

Fix (`theme.go:509`):

```go
merged := mergeTheme(DefaultTheme(), &override)
merged.Ref = "" // in-memory theme has no reloadable reference
```

plus `assert.Equal(t, "", theme.Ref)` in the test.

(Non-blocking: even with `Ref=""`, cancel-preview can't *restore* an in-memory theme since `ApplyThemeRef("")` also falls back to the default. Clearing `Ref` fixes the docstring and the picker label.)

EronWright:
@Sayt-0 you're right, the interaction with the picker isn't good enough.  I think my true goal is to extend the built-ins in a first-class way, i.e. persistently selectable.  Here's what I have in mind in lieu of a raw `ParseTheme`, as a way to simply extend the set of built-in theme files:
```go
  //go:embed themes/*.yaml
  var myThemes embed.FS

  styles.AddThemeFS(myThemes)   // my CLI's themes are now built-ins (in addition to the normal built-ins)
```

WDYT?

EronWright:
Closing in favor of a more direct approach (#3182): a `RegisterBuiltinThemes(fs.FS)` that lets an embedder contribute additional built-in theme files through the same list/load/merge-onto-default pipeline cagent already uses for its bundled themes, rather than adding an in-memory parse primitive. This makes embedder themes first-class in the picker (listed, selectable, restorable, persistable) with no special-casing.

Thanks @rumpl and @Sayt-0 for the reviews — the Ref-handling catch fed directly into the new design.

PR #3184, opened 2026-06-20
refactor: make toolsets and providers explicit

dgageot:
The registry patterns for toolsets and providers have been implicit, relying on blank imports and init() functions to wire everything together. This refactor makes both registries explicit.

Providers now live behind a *provider.Registry which must be passed around explicitly. Similarly, toolsets are registered in pkg/teamloader/toolsets/NewDefaultToolsetRegistry(), removing the hidden coupling through blank imports in cmd/root/toolsets.go. The rag toolset, which pulls in cgo dependencies, is now available only when explicitly included via the registry.


Sayt-0 (approved):

docker-agent:
❌ **PR Review Failed** — The review agent encountered an error and could not complete the review. [View logs](https://github.com/docker/docker-agent/actions/runs/27868543685).

PR #3185, opened 2026-06-20
fix: seal MCP OAuth tokens with keyring-backed file

dgageot:
## Summary

- Replace the single keyring OAuth token bundle with a keyring-sealed encrypted file.
- Store only a fixed-size AES-256 key in the OS keyring; store the OAuth token bundle in an AES-GCM encrypted file under the config dir.
- Migrate existing legacy keyring entries safely, preserving both the previous bundled layout and the older per-resource layout.
- Harden persistence with cross-process locking, reload-before-write merge semantics, Windows-safe atomic file replacement, and unreadable-file safeguards.

### DIFF
--- a/pkg/tools/mcp/tokenstore_keyring.go
+++ b/pkg/tools/mcp/tokenstore_keyring.go
@@ -65,6 +65,7 @@ type KeyringTokenStore struct {
 
 	mu     sync.Mutex
 	cache  map[string]*OAuthToken
+	key    []byte // cached encryption key; fetched from the keyring once
 	loaded bool
 }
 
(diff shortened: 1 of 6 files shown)


pkg/tools/mcp/tokenstore_keyring.go
```diff
@@ -102,73 +137,263 @@ func (s *KeyringTokenStore) load() {
...
+		return s.key, nil
+	}
+
+	item, err := s.ring.Get(encryptionKeyItem)
+	switch {
+	case err == nil:
+		if len(item.Data) != encryptionKeySize {
+			return nil, fmt.Errorf("stored encryption key has wrong size %d", len(item.Data))
+		}
+		s.key = item.Data
```
docker-agent, line 179: **[LOW] Encryption key aliased from keyring `item.Data` — no defensive copy**

In `encryptionKey()`, when fetching an existing key from the keyring, `s.key` is assigned directly from `item.Data`:

```go
s.key = item.Data
```

This copies only the slice *header* (pointer, length, cap), not the underlying bytes. While Go's GC ensures the backing array won't be collected as long as `s.key` is live, the alias means any keyring backend that reuses or mutates its internal buffer after `Get()` returns could silently corrupt the cached key — causing all subsequent encrypt/decrypt operations to fail or produce wrong ciphertext.

In contrast, the newly-generated key path (line 193) does not have this issue since `key` is a freshly-allocated slice that is never shared.

**Suggestion**: Use a defensive copy to be safe across all possible keyring backends:
```go
s.key = bytes.Clone(item.Data)
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

The AES-256-GCM implementation is well-structured: fresh random nonces are generated per encryption call, authenticated decryption is used correctly (GCM tag verification via `cipher.AEAD.Open`), atomic file writes are handled by `natefinch/atomic` which correctly handles existing-file replacement on Windows, cross-process file locking uses `flock` (Unix) / `LockFileEx` (Windows), and the migration logic preserves tokens with a persist-before-delete strategy. One minor hardening opportunity was found.

Sayt-0 (approved):

PR #3186, opened 2026-06-20
fix: thread provider registry into RAG, eval, wasm and a2a/mcp/chat runtimes

Sayt-0:
## Context

Follow-up to #3184, which made the model provider registry explicit. The package-level `provider.New` / `provider.NewWithModels` now resolve against the slim default registry (`dmr` only on native, empty on js/wasm); the full provider set lives in `providers.NewDefaultRegistry()` and must be passed explicitly.

Several production paths still call the package-level helpers and were not updated by #3184. They silently fall back to the slim registry and fail at runtime with `unknown provider type "openai"` (or anthropic/google/bedrock). All of them compile cleanly, so neither the compiler nor CI caught the regression; it only surfaces when the path is exercised.

## Regressions fixed

| Severity | Path | Symptom before this PR |
|---|---|---|
| Blocker | `rag` toolset (`pkg/rag`) | A RAG toolset with a non-dmr embedding/reranking model (e.g. `openai/text-embedding-3-small`) fails at `Start()`. `model: auto` silently drops the openai candidate. |
| Blocker | js/wasm browser build (`cmd/wasm`) | Every model fails with `provider type "openai" is not registered`; the whole browser build was non-functional. |
| Warning | `eval` command (`pkg/evaluation`) | `eval` with a non-dmr judge (the default is `anthropic/...`) fails at judge construction. |
| Warning | a2a / mcp / chatserver runtimes | `type: model` hooks fail for non-dmr models (the runtime model hook got the slim registry). |

## Approach

- **rag**: thread a `*provider.Registry` through `ManagersBuildConfig` and `strategy.BuildContext`. The rag toolset creator passes `providers.NewDefaultRegistry()`. A nil registry still falls back to the slim default, so direct constructions are unaffected.
- **evaluation**: build the LLM judge through `providers.NewDefaultRegistry()`.
- **js/wasm**: repopulate `factory_js.go`'s default registry with the browser-capable providers (openai/anthropic/gemini), matching the pre-#3184 slim set. `cmd/wasm` is unchanged.
- **a2a / mcp / chatserver**: load via `LoadWithConfig` and pass `runtime.WithProviderRegistry(loadResult.ProviderRegistry)`, mirroring what `acp` and `new` already do.

## Tradeoff

`pkg/teamloader/toolsets` now imports `pkg/model/provider/providers` (for the rag creator), so the batteries-included toolset registry pulls the provider SDKs. This is the same coupling `defaults.Opts()` already has, and embedders building a slim toolset registry are unaffected. A more surgical alternative would thread the registry through the `ToolsetCreator` signature, but that churns a freshly-introduced public API.

## Verification

- `go build ./...` and `GOOS=js GOARCH=wasm go build ./cmd/wasm/...` pass.
- `go vet` and `go test` pass for `rag`, `tools/builtin/rag`, `teamloader`, `evaluation`, `a2a`, `mcp`, `chatserver`, `model/provider/...`.

### DIFF
--- a/pkg/mcp/server.go
+++ b/pkg/mcp/server.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/docker/docker-agent/pkg/agent"
 	"github.com/docker/docker-agent/pkg/config"
+	"github.com/docker/docker-agent/pkg/model/provider"
 	"github.com/docker/docker-agent/pkg/runtime"
 	"github.com/docker/docker-agent/pkg/session"
 	"github.com/docker/docker-agent/pkg/team"
@@ -92,10 +93,11 @@ func createMCPServer(ctx context.Context, agentFilename, agentName string, runCo
 		return nil, nil, err
 	}
 
-	t, err := teamloader.Load(ctx, agentSource, runConfig, loaderdefaults.Opts()...)
+	loadResult, err := teamloader.LoadWithConfig(ctx, agentSource, runConfig, loaderdefaults.Opts()...)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to load agents: %w", err)
 	}
+	t := loadResult.Team
 
 	cleanup := func() {
 		if err := t.StopToolSets(ctx); err != nil {
@@ -154,13 +156,13 @@ func createMCPServer(ctx context.Context, agentFilename, agentName string, runCo
 			OutputSchema: tools.MustSchemaFor[ToolOutput](),
 		}
 
-		mcp.AddTool(server, toolDef, CreateToolHandler(t, agentName))
+		mcp.AddTool(server, toolDef, CreateToolHandler(t, agentName, loadResult.ProviderRegistry))
 	}
 
 	return server, cleanup, nil
 }
 
-func CreateToolHandler(t *team.Team, agentName string) func(context.Context, *mcp.CallToolRequest, ToolInput) (*mcp.CallToolResult, ToolOutput, error) {
+func CreateToolHandler(t *team.Team, agentName string, providerRegistry *provider.Registry) func(context.Context, *mcp.CallToolRequest, ToolInput) (*mcp.CallToolResult, ToolOutput, error) {
 	return func(ctx context.Context, req *mcp.CallToolRequest, input ToolInput) (*mcp.CallToolResult, ToolOutput, error) {
 		slog.DebugContext(ctx, "MCP tool called", "agent", agentName, "message", input.Message)
 
@@ -183,6 +185,7 @@ func CreateToolHandler(t *team.Team, agentName string) func(context.Context, *mc
 			runtime.WithCurrentAgent(agentName),
 			runtime.WithNonInteractive(true),
 			runtime.WithTracer(otel.Tracer("cagent")),
+			runtime.WithProviderRegistry(providerRegistry),
 		)
 		if err != nil {
 			return nil, ToolOutput{}, fmt.Errorf("failed to create runtime: %w", err)
--- a/pkg/teamloader/toolsets/toolsets.go
+++ b/pkg/teamloader/toolsets/toolsets.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/docker/docker-agent/pkg/config"
 	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/model/provider/providers"
 	"github.com/docker/docker-agent/pkg/teamloader"
 	"github.com/docker/docker-agent/pkg/tools"
 	"github.com/docker/docker-agent/pkg/tools/a2a"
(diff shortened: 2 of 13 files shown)


pkg/mcp/server.go
```diff
@@ -154,13 +156,13 @@ func createMCPServer(ctx context.Context, agentFilename, agentName string, runCo
...
 
-		mcp.AddTool(server, toolDef, CreateToolHandler(t, agentName))
+		mcp.AddTool(server, toolDef, CreateToolHandler(t, agentName, loadResult.ProviderRegistry))
 	}
 
 	return server, cleanup, nil
 }
 
-func CreateToolHandler(t *team.Team, agentName string) func(context.Context, *mcp.CallToolRequest, ToolInput) (*mcp.CallToolResult, ToolOutput, error) {
+func CreateToolHandler(t *team.Team, agentName string, providerRegistry *provider.Registry) func(context.Context, *mcp.CallToolRequest, ToolInput) (*mcp.CallToolResult, ToolOutput, error) {
```
docker-agent, line 165: **[HIGH] Exported `CreateToolHandler` signature changed — e2e callers will fail to compile**

`CreateToolHandler` is a public exported function. Its signature was updated from:
```go
func CreateToolHandler(t *team.Team, agentName string) func(...)
```
to:
```go
func CreateToolHandler(t *team.Team, agentName string, providerRegistry *provider.Registry) func(...)
```

However, `e2e/mcp_test.go` still calls the old 2-argument form (lines 29 and 52):
```go
handler := mcp.CreateToolHandler(team, "root")   // compile error: too few arguments
handler := mcp.CreateToolHandler(team, "web")    // compile error: too few arguments
```

This will cause a build failure for any consumer of this package (including the repo's own e2e tests) that was not updated in this PR. The e2e tests also still use `teamloader.Load(...)` (which returns `*team.Team`) instead of `teamloader.LoadWithConfig(...)` (which returns `LoadResult`), so they would also need to be updated to extract a `ProviderRegistry` to pass here.

**Fix**: Update `e2e/mcp_test.go` to pass `nil` (or a real registry from `LoadWithConfig`) as the third argument:
```go
// Quick fix — use nil to retain slim-registry behavior in e2e tests:
handler := mcp.CreateToolHandler(team, "root", nil)
// Better fix — thread the registry through LoadWithConfig:
loadResult, err := teamloader.LoadWithConfig(ctx, agentSource, runConfig, loaderdefaults.Opts()...)
handler := mcp.CreateToolHandler(loadResult.Team, "root", loadResult.ProviderRegistry)
```

pkg/teamloader/toolsets/toolsets.go
```diff
@@ -91,7 +92,7 @@ func DefaultToolsetCreators() map[string]teamloader.ToolsetCreator {
 			return agenttool.CreateToolSet()
 		},
 		"rag": func(ctx context.Context, toolset latest.Toolset, parentDir string, runConfig *config.RuntimeConfig, _ string) (tools.ToolSet, error) {
-			return rag.CreateToolSet(ctx, toolset, parentDir, runConfig)
+			return rag.CreateToolSet(ctx, toolset, parentDir, runConfig, providers.NewDefaultRegistry())
```
docker-agent, line 95: **[MEDIUM] RAG toolset creator ignores threaded provider registry — custom providers lost**

`DefaultToolsetCreators()` hard-codes `providers.NewDefaultRegistry()` for RAG toolsets instead of forwarding whatever registry was threaded into the loader:

```go
"rag": func(ctx context.Context, toolset latest.Toolset, parentDir string, runConfig *config.RuntimeConfig, _ string) (tools.ToolSet, error) {
    return rag.CreateToolSet(ctx, toolset, parentDir, runConfig, providers.NewDefaultRegistry())
},
```

The `ToolsetCreator` function signature (defined as `func(ctx, toolset, parentDir, runConfig, agentName)`) does not include a `*provider.Registry` parameter, so there is no way to forward `LoadResult.ProviderRegistry` into the toolset even though the whole PR is aimed at threading it through.

As a result:
1. Any **custom providers** registered into the loader's registry (via `teamloader.WithProviderRegistry(customReg)`) are silently ignored when creating RAG toolsets — they will still get `providers.NewDefaultRegistry()` which won't include the custom entries.
2. A fresh registry is allocated on every toolset creation call, bypassing any shared-state caching the application might rely on.

For the standard case (no custom providers), this is functionally equivalent since `providers.NewDefaultRegistry()` is the full factory-registry. But it breaks the threading contract established by this PR for custom-provider configurations.

**Fix**: Either extend `ToolsetCreator` to accept a `*provider.Registry` parameter, or look up the provider registry from a context value rather than allocating fresh each time.

docker-agent (commented):
## Review: COMMENT
### Assessment: 🔴 CRITICAL

This PR correctly threads `*provider.Registry` through RAG, evaluation, wasm, a2a/mcp/chatserver to fix `unknown provider type` regressions. The approach is sound, but there is one build-breaking omission and one design gap in the threading contract.

**Summary of findings:**
1. 🔴 **CRITICAL** — `e2e/mcp_test.go` calls the old 2-argument `CreateToolHandler` — the e2e suite will not compile.
2. 🟡 **MEDIUM** — `DefaultToolsetCreators()` always allocates `providers.NewDefaultRegistry()` for RAG, ignoring any custom registry threaded through the loader.

PR #3188, opened 2026-06-20
refactor: decouple embedder dependencies

dgageot:
Packages like `pkg/runtime`, `pkg/model/provider`, and `pkg/tools/mcp` were transitively pulling in `openai-go` and `99designs/keyring` even for embedders that have no need for those dependencies. This made it expensive for downstream consumers to import any of these packages without dragging in heavyweight transitive deps.

DMR registration has moved to the explicit providers registry, so it is no longer wired in at import time. The MCP keyring token store has been extracted into an opt-in `keyringstore` sub-package, meaning callers that do not need keychain integration simply do not import it.

Validation:

```bash
go test ./...
```

```bash
# Both commands should return no output
go list -deps ./pkg/runtime ./pkg/model/provider ./pkg/tools/mcp | grep -E '99designs/keyring|openai-go'
go list -deps ./tui/gordon | grep 'local-replace'
```

No behaviour changes; this is a pure dependency boundary cleanup with no API surface modifications.

### DIFF
--- a/pkg/model/provider/factory.go
+++ b/pkg/model/provider/factory.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
-	"github.com/docker/docker-agent/pkg/model/provider/dmr"
 	"github.com/docker/docker-agent/pkg/model/provider/options"
 	"github.com/docker/docker-agent/pkg/model/provider/rulebased"
 )
@@ -22,8 +21,7 @@ type Registry struct {
 }
 
 func NewRegistry(factories map[string]Factory) *Registry {
-	copied := make(map[string]Factory, len(factories)+1)
-	copied["dmr"] = dmrFactory
+	copied := make(map[string]Factory, len(factories))
 	maps.Copy(copied, factories)
 	return &Registry{factories: copied}
 }
@@ -92,14 +90,13 @@ func (r *Registry) createDirectProvider(ctx context.Context, cfg *latest.ModelCo
 
 var defaultFactories map[string]Factory
 
+// Deprecated: the core registry intentionally contains no concrete SDK-backed
+// providers. Use providers.NewDefaultRegistry for Docker Agent's full provider
+// set, or NewRegistry with the concrete provider factories an embedder needs.
 func DefaultRegistry() *Registry {
 	return NewRegistry(defaultFactories)
 }
 
 func unknownProviderError(providerType string) error {
 	return fmt.Errorf("unknown provider type %q (register it with provider.NewRegistry or use providers.NewDefaultRegistry)", providerType)
-}
-
-func dmrFactory(ctx context.Context, cfg *latest.ModelConfig, _ environment.Provider, opts ...options.Opt) (Provider, error) {
-	return dmr.NewClient(ctx, cfg, opts...)
 }
--- a/pkg/tools/mcp/keyringstore/tokenstore_lock_unix.go
+++ b/pkg/tools/mcp/keyringstore/tokenstore_lock_unix.go
@@ -1,0 +1,22 @@
+//go:build unix
+
+package keyringstore
+
+import (
+	"errors"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+func lockExclusive(f *os.File) error {
+	err := unix.Flock(int(f.Fd()), unix.LOCK_EX|unix.LOCK_NB)
+	if errors.Is(err, unix.EWOULDBLOCK) || errors.Is(err, unix.EAGAIN) {
+		return errLockBusy
+	}
+	return err
+}
+
+func unlockFile(f *os.File) error {
+	return unix.Flock(int(f.Fd()), unix.LOCK_UN)
+}
--- a/pkg/tools/mcp/keyringstore/tokenstore_lock_windows.go
+++ b/pkg/tools/mcp/keyringstore/tokenstore_lock_windows.go
@@ -1,0 +1,39 @@
+//go:build windows
+
+package keyringstore
+
+import (
+	"errors"
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+const maxLockRange = ^uint32(0)
+
+func lockExclusive(f *os.File) error {
+	var ol windows.Overlapped
+	err := windows.LockFileEx(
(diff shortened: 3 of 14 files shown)


pkg/tools/mcp/tokenstore.go
```diff
@@ -17,6 +18,41 @@ type OAuthTokenStore interface {
 	RemoveToken(resourceURL string) error
 }
 
+// OAuthTokenStoreFactory constructs the process-wide persistent OAuth token store.
+type OAuthTokenStoreFactory func() OAuthTokenStore
+
+var (
+	defaultStoreMu sync.Mutex
+	defaultStore   = sync.OnceValue(func() OAuthTokenStore { return NewInMemoryTokenStore() })
```
docker-agent: **[MEDIUM] Init-order fragility: `mcp.NewKeyringTokenStore()` called before `keyringstore.init()` silently returns in-memory store**

`defaultTokenStore()` captures `store := defaultStore` under `defaultStoreMu`, then calls `store()` *after* releasing the lock. There is a narrow TOCTOU window:

1. Goroutine A calls `mcp.NewKeyringTokenStore()` → locks `defaultStoreMu`, captures the in-memory `OnceValue` as `store`, releases the lock.
2. `keyringstore.init()` runs `SetDefaultTokenStoreFactory`, replacing `defaultStore` with a new `sync.OnceValue(NewKeyringTokenStore)`.
3. Goroutine A calls `store()` — this invokes the **old** in-memory factory and seals it permanently for that call.

In the `cmd/root` binary the blank import `_ "keyringstore"` guarantees correct ordering before `main()` runs. However, if any package calls `mcp.NewKeyringTokenStore()` from its own `init()` without transitively importing `keyringstore`, it silently receives (and permanently seals) the in-memory store — OAuth tokens for that call site are never persisted, with no error or warning.

Consider adding a doc comment on `defaultStore` stating the ordering contract, or documenting that `Register()` must be called before any `NewKeyringTokenStore()` invocation.

pkg/tools/mcp/keyringstore/tokenstore_lock_unix.go
```diff
@@ -0,0 +1,17 @@
...
+package keyringstore
+
+import (
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+func lockExclusive(f *os.File) error {
+	return unix.Flock(int(f.Fd()), unix.LOCK_EX)
```
docker-agent: **[MEDIUM] `unix.Flock(LOCK_EX)` blocks indefinitely while `s.mu` is held**

`StoreToken` and `RemoveToken` both acquire `s.mu` and then call `lockTokenFile` → `lockExclusive` → `unix.Flock(fd, LOCK_EX)`. `LOCK_EX` without `LOCK_NB` is a blocking call that waits until the cross-process lock is available.

If another process holds the lock for an extended period (e.g. a concurrent docker-agent instance performing a write or migration), every goroutine in this process that also needs `s.mu` — including `GetToken` — will be stuck. This can stall the entire MCP OAuth subsystem in one process because of activity in another.

The standard remedy is to acquire the cross-process lock **before** the in-process mutex, or to use `LOCK_EX|LOCK_NB` with a retry loop bounded by a deadline so callers receive a timeout error rather than hanging indefinitely.

pkg/tools/mcp/keyringstore/tokenstore_lock_windows.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+const maxLockRange = ^uint32(0)
+
+func lockExclusive(f *os.File) error {
+	var ol windows.Overlapped
+	return windows.LockFileEx(
```
docker-agent: **[MEDIUM] `LockFileEx` blocks indefinitely while `s.mu` is held (missing `LOCKFILE_FAIL_IMMEDIATELY`)**

Same pattern as the Unix implementation: `windows.LockFileEx` with only `LOCKFILE_EXCLUSIVE_LOCK` (without `LOCKFILE_FAIL_IMMEDIATELY`) is a blocking call. `StoreToken`/`RemoveToken` hold `s.mu` while this cross-process lock acquisition blocks.

If any other process holds the file lock (e.g. a second docker-agent instance mid-write), all goroutines in this process needing `s.mu` — including `GetToken` — are stalled until the other process releases its lock, with no timeout or error path.

Consider passing `LOCKFILE_EXCLUSIVE_LOCK | LOCKFILE_FAIL_IMMEDIATELY` and retrying with back-off up to a deadline, or restructuring so the in-process mutex is released before the file lock is acquired.

pkg/model/provider/factory.go
```diff
@@ -22,8 +21,7 @@ type Registry struct {
 }
 
 func NewRegistry(factories map[string]Factory) *Registry {
-	copied := make(map[string]Factory, len(factories)+1)
-	copied["dmr"] = dmrFactory
+	copied := make(map[string]Factory, len(factories))
```
docker-agent, line 24: **[MEDIUM] `NewRegistry` no longer hard-wires `dmr` — `DefaultRegistry()` now returns an empty registry**

The previous `NewRegistry` always seeded `copied["dmr"] = dmrFactory` regardless of what `factories` was passed. That hard-wiring is now removed: `DefaultRegistry()` calls `NewRegistry(nil)`, which returns a registry with zero entries.

Any existing caller that invokes `provider.DefaultRegistry()` directly (internal or downstream) and passes the result to `New`/`NewWithModels` will fail immediately:

```
unknown provider type "dmr" (register it with provider.NewRegistry or use providers.NewDefaultRegistry())
```

This is likely the intended design — callers should migrate to `providers.NewDefaultRegistry()` — but there is no deprecation notice on `DefaultRegistry()`, and the `if r == nil { r = DefaultRegistry() }` guard inside `createDirectProvider` is now dead code (a method on `*Registry` can never have a nil receiver in normal usage), so it provides no fallback.

Suggest adding a `// Deprecated: use providers.NewDefaultRegistry()` doc comment on `DefaultRegistry()` so IDEs and `go vet` surface the migration requirement at the call site rather than at runtime.

docker-agent (commented):

The dependency decoupling is well-structured: the `keyringstore` sub-package extraction is clean, build-tag coverage across unix/windows/js is complete, and the `providers` package correctly re-registers all factories including `dmr`. Four medium-severity issues were found in the new code.

PR #3189, opened 2026-06-20
refactor: decouple embedder deps and register keyring store explicitly

dgageot:
Embedder packages like `pkg/runtime`, `pkg/model/provider`, and `pkg/tools/mcp` were transitively pulling in `openai-go` and `99designs/keyring` even for consumers that have no need for those dependencies. This made any downstream import of these packages unexpectedly heavyweight.

The MCP OAuth token store now defaults to an in-memory implementation. The OS-keyring-backed store lives in its own `pkg/tools/mcp/keyringstore` sub-package and is registered explicitly via `keyringstore.Register()`, called once from `cmd/root.Execute()` at CLI startup. No `init()` side effects, no blank imports. The OAuth token-file locking logic has been moved into `keyringstore` alongside the implementation it belongs to, and the now-dead `tokenstore_lock*.go` files in `pkg/tools/mcp` have been removed.

No behaviour changes for end users; this is a dependency boundary cleanup with no API surface modifications.

### DIFF
--- a/pkg/tools/mcp/tokenstore.go
+++ b/pkg/tools/mcp/tokenstore.go
@@ -1,7 +1,9 @@
 package mcp
 
 import (
+	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/docker/docker-agent/pkg/concurrent"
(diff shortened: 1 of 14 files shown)


pkg/tools/mcp/tokenstore.go
```diff
@@ -17,6 +19,41 @@ type OAuthTokenStore interface {
...
+	defaultStoreMu.Lock()
+	defer defaultStoreMu.Unlock()
+	defaultStore = sync.OnceValue(factory)
+}
+
+func defaultTokenStore() OAuthTokenStore {
+	defaultStoreMu.Lock()
+	store := defaultStore
+	defaultStoreMu.Unlock()
+	return store()
```
docker-agent: **[MEDIUM] TOCTOU window: `defaultTokenStore` releases mutex before invoking `store()`**

`defaultTokenStore()` correctly copies `defaultStore` under the mutex, but then calls `store()` *after* releasing the lock (line 47). Between the `Unlock()` and the `store()` invocation, a concurrent call to `SetDefaultTokenStoreFactory()` could replace `defaultStore` with a new `sync.OnceValue`. The local `store` variable still points to the old `sync.OnceValue`, so this caller gets the old (possibly in-memory) store while all future callers get the new keyring-backed store. This creates two live, divergent `OAuthTokenStore` instances in the same process.

In practice, `Register()` is called once at the top of `Execute()` before any MCP activity, so the race doesn't fire in normal CLI use. However, embedders or test helpers that call `NewKeyringTokenStore()` concurrently with `SetDefaultTokenStoreFactory()` would be affected. Consider holding the lock through the `store()` call, or using a `sync.Once`-guarded initialisation that cannot be swapped after first use.

pkg/tools/mcp/tokenstore.go
```diff
@@ -17,6 +19,41 @@ type OAuthTokenStore interface {
...
+var (
+	defaultStoreMu sync.Mutex
+	defaultStore   = sync.OnceValue(NewInMemoryTokenStore)
+)
+
+// SetDefaultTokenStoreFactory installs the process-wide persistent token-store
+// factory used by NewKeyringTokenStore and remote MCP toolsets. Embedders that
+// do not need persistent MCP OAuth storage can avoid importing any OS keyring
+// implementation; docker-agent's CLI registers one from pkg/tools/mcp/keyringstore.
+func SetDefaultTokenStoreFactory(factory OAuthTokenStoreFactory) {
```
docker-agent, line 40: **[MEDIUM] Late `Register()` call can permanently lock in the in-memory store**

`defaultStore` in `pkg/tools/mcp/tokenstore.go` is initialised to `sync.OnceValue(NewInMemoryTokenStore)`. `SetDefaultTokenStoreFactory` *replaces* the variable with a brand-new `sync.OnceValue(factory)`, but it cannot un-invoke an already-executed `sync.OnceValue`.

If any code path calls `mcp.NewKeyringTokenStore()` (→ `defaultTokenStore()` → old `defaultStore()`) **before** `keyringstore.Register()` is called in `Execute()`, the old `sync.OnceValue` memoises the in-memory store. `Register()` then installs a new `sync.OnceValue`, so future callers get the keyring store — but the early caller already holds a permanent reference to the in-memory instance. The two instances diverge silently: tokens written through one are invisible to the other.

The CLI is safe today because `Register()` is the first thing in `Execute()`. The risk is that any package-level `var` initialiser or `TestMain` that touches `mcp.NewKeyringTokenStore()` before `Execute()` would permanently shadow the keyring store. A guard such as `panic`-on-double-init, or a `sync.Once` that prevents `SetDefaultTokenStoreFactory` from being called after first use, would make this invariant explicit.

docker-agent (commented):

This PR cleanly decouples keyring/openai-go from embedder packages by moving the OS-keyring token store into its own `keyringstore` sub-package with explicit `Register()` registration. The overall design is sound: atomic file writes, AES-256-GCM encryption at rest, cross-process file locking, legacy-token migration, and the `testing.Testing()` guard (idiomatic Go 1.21+ pattern, intentionally used here) are all correct.

Two medium findings in the new `defaultStore` swap mechanism are worth addressing to prevent subtle divergence issues for embedders.

| Severity | Finding |
|----------|---------|
| 🟡 Medium | TOCTOU window: mutex released before `store()` is called — can yield two live store instances |
| 🟡 Medium | Late `Register()` call permanently locks in the in-memory store if any caller precedes it |

aheritier (approved):

PR #3196, opened 2026-06-22
fix: restore provider registry after cf5a430d2 emptied the default

dgageot:
Commit cf5a430d2 (`refactor: make toolsets and providers explicit`) emptied `provider.DefaultRegistry()` and required callers to pass the full provider set explicitly via `providers.NewDefaultRegistry()`. The main agent model loading and runtime model switcher were updated at that time, but three code paths still fell through to the now-empty default registry and failed at runtime with `unknown provider type "openai"`.

This commit restores the three broken paths. RAG toolsets now receive the resolved provider registry via a new `ProviderRegistry` field on `config.RuntimeConfig` (with a nil-safe `ProviderRegistryOrDefault()` getter); `teamloader.LoadWithConfig` populates it, and `pkg/rag/builder.go` and `pkg/rag/strategy/strategy.go` consume it instead of the empty default. The js/wasm build gets its own restored `defaultFactories` in `pkg/model/provider/factory_js.go` (openai, anthropic, google) because `pkg/model/provider/providers` cannot compile to wasm — it pulls in cloud SDKs that have no wasm-safe build. The eval judge in `pkg/evaluation/eval.go` previously called the empty default directly; `cmd/root/eval.go` now sets `runConfig.ProviderRegistry = providers.NewDefaultRegistry()` and `createJudgeModel` uses `runConfig.ProviderRegistryOrDefault()`.

No behavior changes outside these three error paths. The fix is self-contained and all existing tests continue to pass.

### DIFF
--- a/pkg/config/runtime.go
+++ b/pkg/config/runtime.go
@@ -24,6 +24,12 @@ type RuntimeConfig struct {
 	modelsDevStore         *modelsdev.Store
 	modelsDevStoreErr      error
 	modelsDevStoreOnce     sync.Once
+
+	// ProviderRegistry instantiates model providers for toolsets that build
+	// providers at load time (e.g. RAG embeddings/reranking). It is populated
+	// by the team loader with the same registry used for agent models. When
+	// nil, ProviderRegistryOrDefault falls back to provider.DefaultRegistry.
+	ProviderRegistry *provider.Registry
 }
 
 type Config struct {
@@ -78,6 +84,7 @@ func (runConfig *RuntimeConfig) Clone() *RuntimeConfig {
 		ModelsDevStoreOverride: runConfig.ModelsDevStoreOverride,
 		modelsDevStore:         store,
 		modelsDevStoreErr:      storeErr,
+		ProviderRegistry:       runConfig.ProviderRegistry,
 	}
 	clone.envProviderOnce.Do(func() {})    // mark as resolved
 	clone.modelsDevStoreOnce.Do(func() {}) // mark as resolved
@@ -107,6 +114,18 @@ func (runConfig *RuntimeConfig) ModelsDevStore() (*modelsdev.Store, error) {
 		)
 	})
 	return runConfig.modelsDevStore, runConfig.modelsDevStoreErr
+}
+
+// ProviderRegistryOrDefault returns the configured provider registry, or the
+// package default registry when none was set (including when the receiver is
+// nil). The default registry only contains providers the core package can
+// expose without optional SDK dependencies, so callers that need the full
+// provider set must ensure the team loader populated ProviderRegistry.
+func (runConfig *RuntimeConfig) ProviderRegistryOrDefault() *provider.Registry {
+	if runConfig != nil && runConfig.ProviderRegistry != nil {
+		return runConfig.ProviderRegistry
+	}
+	return provider.DefaultRegistry()
 }
 
 func (runConfig *RuntimeConfig) EnvProvider() environment.Provider {
--- a/pkg/rag/builder.go
+++ b/pkg/rag/builder.go
@@ -27,9 +27,11 @@ type ManagersBuildConfig struct {
 }
 
 // NewProvider creates a model provider using the build config's environment,
-// gateway, and custom provider settings.
+// gateway, and custom provider settings. It uses the provider registry carried
+// by RuntimeConfig (populated by the team loader with the full provider set);
+// without it, model creation fails with "unknown provider type".
 func (c ManagersBuildConfig) NewProvider(ctx context.Context, cfg *latest.ModelConfig) (provider.Provider, error) {
-	return provider.New(ctx, cfg, c.Env,
+	return c.RuntimeConfig.ProviderRegistryOrDefault().New(ctx, cfg, c.Env,
 		options.WithGateway(c.ModelsGateway),
 		options.WithProviders(c.Providers))
 }
--- a/pkg/rag/strategy/strategy.go
+++ b/pkg/rag/strategy/strategy.go
@@ -26,9 +26,11 @@ type BuildContext struct {
 }
 
 // NewProvider creates a model provider using the build context's environment,
-// gateway, and custom provider settings.
+// gateway, and custom provider settings. It uses the provider registry carried
+// by RuntimeConfig (populated by the team loader with the full provider set);
+// without it, model creation fails with "unknown provider type".
 func (c BuildContext) NewProvider(ctx context.Context, cfg *latest.ModelConfig) (provider.Provider, error) {
-	return provider.New(ctx, cfg, c.Env,
+	return c.RuntimeConfig.ProviderRegistryOrDefault().New(ctx, cfg, c.Env,
 		options.WithGateway(c.ModelsGateway),
 		options.WithProviders(c.Providers))
 }
--- a/pkg/teamloader/teamloader.go
+++ b/pkg/teamloader/teamloader.go
@@ -180,6 +180,9 @@ func LoadWithConfig(ctx context.Context, agentSource config.Source, runConfig *c
 	// Make model definitions available to toolset creators (e.g., RAG reranking)
 	runConfig.Models = cfg.Models
 	runConfig.Providers = cfg.Providers
+	// Share the resolved provider registry so toolsets that build providers at
+	// load time (e.g. RAG embeddings/reranking) use the same one as agent models.
+	runConfig.ProviderRegistry = loadOpts.providerRegistry
 
 	// Load agents
 	parentDir := cmp.Or(agentSource.ParentDir(), runConfig.WorkingDir)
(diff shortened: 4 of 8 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The three broken provider-registry paths are correctly restored. The nil-safe `ProviderRegistryOrDefault()` getter, the new `ProviderRegistry` field on `config.RuntimeConfig`, and the `teamloader.LoadWithConfig` wiring all look sound.

**Key areas checked:**
- **`teamloader.LoadWithConfig` recursive safety**: `loadExternalAgent` explicitly forwards the parent's registry via `WithProviderRegistry(loadOpts.providerRegistry)`, so recursive loads for external agent refs correctly propagate the full registry — no overwrite with the empty default.
- **js/wasm `factory_js.go`**: The missing `vertexai.IsModelGardenConfig` branch is intentional and documented; Vertex AI cloud SDKs don't compile to wasm, and the comment in the new file makes this explicit.
- **Eval judge path** (`cmd/root/eval.go`): `ProviderRegistry` is set before `createJudgeModel` is invoked — ordering is correct.
- **RAG builder/strategy**: Both now consume `runConfig.ProviderRegistryOrDefault()`, which falls back to `DefaultRegistry()` (populated in wasm, empty in non-wasm — but `LoadWithConfig` always sets it before RAG toolsets are created).

melmennaoui (dismissed):

pkg/rag/builder.go
```diff
@@ -27,9 +27,11 @@ type ManagersBuildConfig struct {
 }
 
 // NewProvider creates a model provider using the build config's environment,
-// gateway, and custom provider settings.
+// gateway, and custom provider settings. It uses the provider registry carried
+// by RuntimeConfig (populated by the team loader with the full provider set);
+// without it, model creation fails with "unknown provider type".
 func (c ManagersBuildConfig) NewProvider(ctx context.Context, cfg *latest.ModelConfig) (provider.Provider, error) {
-	return provider.New(ctx, cfg, c.Env,
+	return c.RuntimeConfig.ProviderRegistryOrDefault().New(ctx, cfg, c.Env,
```
docker-agent, line 34: **[MEDIUM/LIKELY] Silent fallback to empty provider registry when `RuntimeConfig.ProviderRegistry` is unset**

`c.RuntimeConfig.ProviderRegistryOrDefault()` falls back to `provider.DefaultRegistry()` when `RuntimeConfig` is nil or its `ProviderRegistry` field is not set. In non-js builds, `provider.DefaultRegistry()` returns an empty registry (no factories registered), so any call to `NewProvider` in that state will produce `"unknown provider type"` errors — exactly the runtime failure this PR is fixing elsewhere.

All current callers go through `loaderdefaults.Opts()` (which injects `providers.NewDefaultRegistry()`), so the happy path is fine. However, the fallback `provider.DefaultRegistry()` returns an empty registry in non-js builds, so any future call site that constructs `ManagersBuildConfig` without going through `teamloader.LoadWithConfig` + `loaderdefaults.Opts()` will fail silently.

Consider logging a warning or returning an explicit error if `RuntimeConfig` is nil, rather than silently falling back to an unusable registry.

pkg/rag/strategy/strategy.go
```diff
@@ -26,9 +26,11 @@ type BuildContext struct {
 }
 
 // NewProvider creates a model provider using the build context's environment,
-// gateway, and custom provider settings.
+// gateway, and custom provider settings. It uses the provider registry carried
+// by RuntimeConfig (populated by the team loader with the full provider set);
+// without it, model creation fails with "unknown provider type".
 func (c BuildContext) NewProvider(ctx context.Context, cfg *latest.ModelConfig) (provider.Provider, error) {
-	return provider.New(ctx, cfg, c.Env,
+	return c.RuntimeConfig.ProviderRegistryOrDefault().New(ctx, cfg, c.Env,
```
docker-agent, line 33: **[MEDIUM/LIKELY] Same silent empty-registry fallback as in `pkg/rag/builder.go`**

`c.RuntimeConfig.ProviderRegistryOrDefault()` falls back to `provider.DefaultRegistry()` (empty in non-js builds) when `RuntimeConfig` is nil or `ProviderRegistry` is unset. The same analysis applies as the parallel issue in `pkg/rag/builder.go`: all current production paths are safe because all callers go through `loaderdefaults.Opts()`, but any future caller that constructs `BuildContext` without a populated `RuntimeConfig` will silently receive a broken provider registry and fail at model creation time with `"unknown provider type"`.

pkg/teamloader/teamloader.go
```diff
@@ -153,6 +153,9 @@ func LoadWithConfig(ctx context.Context, agentSource config.Source, runConfig *c
 	// Make model definitions available to toolset creators (e.g., RAG reranking)
 	runConfig.Models = cfg.Models
 	runConfig.Providers = cfg.Providers
+	// Share the resolved provider registry so toolsets that build providers at
+	// load time (e.g. RAG embeddings/reranking) use the same one as agent models.
+	runConfig.ProviderRegistry = loadOpts.providerRegistry
```
docker-agent, line 185: **[MEDIUM/LIKELY] `loadOpts.providerRegistry` may be empty when `WithProviderRegistry` is not passed; callers get a non-nil but unusable registry**

`loadOpts.providerRegistry` is initialized to `provider.DefaultRegistry()` (an empty registry in non-js builds) before opts are applied. `runConfig.ProviderRegistry` is then set to this value. Because `ProviderRegistryOrDefault()` checks only `!= nil`, a caller that omits `WithProviderRegistry(...)` will silently receive an empty (non-nil) registry and fail at RAG model creation time with `"unknown provider type"` — the same failure this PR intends to fix.

All current callers pass `loaderdefaults.Opts()` (which includes `WithProviderRegistry(providers.NewDefaultRegistry())`), so this is safe today. However the default is a footgun: consider initializing `loadOpts.providerRegistry` to `nil` instead of `provider.DefaultRegistry()`, so that `ProviderRegistryOrDefault()` can provide a meaningful fallback, or enforce that callers must always pass `WithProviderRegistry`.

docker-agent (commented):

This PR correctly restores provider registry wiring for three broken code paths (RAG builder/strategy, eval judge, js/wasm build). The structural fix is sound and all current callers are wired correctly via loaderdefaults.Opts().

Three related medium-severity concerns were identified around the empty-registry fallback behavior: ProviderRegistryOrDefault() silently falls back to an empty DefaultRegistry() in non-js builds, and teamloader.LoadWithConfig initializes loadOpts.providerRegistry to that same empty registry before callers can override it. No current production path triggers these issues, but they are latent traps for future callers.

aheritier (approved):

aheritier (approved):

aheritier (dismissed):

docker-agent (commented):

This PR correctly restores three broken provider-registry code paths (RAG toolsets, js/wasm build, eval judge) after the cf5a430d2 refactor emptied `provider.DefaultRegistry()`. The targeted fixes are sound and the known callers are properly wired. One concern is worth addressing: the new `ProviderRegistryOrDefault()` getter introduced in this PR falls back to the intentionally-empty default registry, creating a trap for future callers.

pkg/config/runtime.go
```diff
@@ -109,6 +116,18 @@ func (runConfig *RuntimeConfig) ModelsDevStore() (*modelsdev.Store, error) {
 	return runConfig.modelsDevStore, runConfig.modelsDevStoreErr
 }
 
+// ProviderRegistryOrDefault returns the configured provider registry, or the
+// package default registry when none was set (including when the receiver is
+// nil). The default registry only contains providers the core package can
+// expose without optional SDK dependencies, so callers that need the full
+// provider set must ensure the team loader populated ProviderRegistry.
+func (runConfig *RuntimeConfig) ProviderRegistryOrDefault() *provider.Registry {
```
docker-agent, line 124: **[MEDIUM] `ProviderRegistryOrDefault()` silently falls back to an empty registry**

This PR introduces `ProviderRegistryOrDefault()` as a nil-safe convenience getter, but its fallback — `provider.DefaultRegistry()` — returns an empty registry (no factories registered) on non-wasm platforms, because `var defaultFactories map[string]Factory` in `factory.go` is `nil` after the cf5a430d2 refactor.

The getter name implies "you'll get something useful by default," which is misleading: any code path that calls `ProviderRegistryOrDefault()` without first having `ProviderRegistry` wired (e.g. a future caller, a test that constructs `RuntimeConfig{}` directly, or a CLI entry point not updated in this PR) will silently receive an empty registry and fail at runtime with `unknown provider type "openai"` — the exact error this PR is meant to fix.

The known callers added in this PR are correctly wired (eval path sets `ProviderRegistry = providers.NewDefaultRegistry()`, teamloader initialises `loadOpts.providerRegistry`), but the "safe fallback" framing of the getter risks silent regressions as the code evolves.

**Suggestion**: Consider either (a) panicking (or returning an error) when the fallback would be empty, (b) falling back to `providers.NewDefaultRegistry()` instead, or (c) renaming to `ProviderRegistryOrEmpty()` to make the semantics explicit.

aheritier (approved):

PR #3199, opened 2026-06-22
feat(server): expose session forking over HTTP

trungutt:
## Why

Downstream HTTP consumers want to let users branch a conversation from a specific user turn — "rewind to this message and try a different question" — without losing the history that came before. Today the only way to do this over the API is to read out the full message list and `POST /api/sessions` a brand-new session with replayed messages, which is lossy (no metadata carry-over, no shared title lineage) and forces every client to re-implement the deep-copy semantics.

`session.BranchSession` already exists in-tree and is exercised by the TUI for the same purpose, but it isn't reachable over HTTP. This PR exposes it.

## What the endpoint does

```
POST /api/sessions/:id/fork
  body: { "message_index": <int> }
  200:  api.SessionResponse for the new session
  400:  index out of range / not a user message / falls inside a sub-session
  404:  parent session not found
```

`message_index` is a 0-based index into the flat, user-visible message list (the same shape `GET /api/sessions/:id` returns). The fork **excludes** the message at that index, so the new session contains messages `[0, message_index)` of the parent. This lets a client prefill the excluded user message into its own input UI for the user to edit and resubmit.

```
parent.messages:   [u0]──[a0]──[u1]──[a1]──[u2]──[a2]
                                 ▲
                            fork here (message_index = 2)

forked.messages:   [u0]──[a0]
forked.title:      "<parent.title> (fork 1)"  →  "(fork 2)" on repeat forks
```

## Design notes

- **User-message-only fork point.** Branching mid-tool-call would dangle a `tool_use` without its matching `tool_result` in the forked session and break the next provider call. The handler validates the resolved item's role and returns 400 otherwise.
- **Index translation.** `Session.GetAllMessages()` returns a flat list (skipping system messages, flattening sub-sessions) while `BranchSession` expects an index into `Session.Messages` (the `Item` slice). The new handler bridges the two via `flatMessageIndexToItemIndex`, so the API stays stable if `Item` ever grows new variants.
- **Title series kept separate.** `BranchSession`'s callers (the TUI's branch-from-edit + fork-session handlers) keep their `(branched)` / `(branch N)` titles. The new HTTP path uses `(fork N)` so consumers can tell the two flows apart in a sessions list. The two functions share their body via a private `branchSessionWithTitle` helper.
- **No persisted parent link.** Migration 016 once carried `branch_parent_*` columns and migration 019 dropped them. Adding them back is out of scope for this PR — clients that need provenance can stash it in `metadata` for now.

## Notes for reviewers

- The deep-copy path is `session.BranchSession`, which already has regression tests for `MultiContent` pointer aliasing, sub-session cloning, and message ID resets. The new code re-uses that.
- Fields `BranchSession` deliberately does **not** copy (`Cost`, `InputTokens`/`OutputTokens` are recomputed; `MaxConsecutiveToolCalls`/`MaxOldToolCallTokens` are not preserved) are unchanged here — happy to revisit in a follow-up if downstream needs them.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -256,10 +256,9 @@ func (s *Server) forkSession(c echo.Context) error {
 	forked, err := s.sm.ForkSession(c.Request().Context(), c.Param("id"), req.MessageIndex)
 	if err != nil {
 		switch {
-		case errors.Is(err, ErrForkInvalidMessage):
-			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
-		case strings.Contains(err.Error(), "out of range"),
-			strings.Contains(err.Error(), "sub-session"):
+		case errors.Is(err, ErrForkInvalidMessage),
+			errors.Is(err, ErrForkOutOfRange),
+			errors.Is(err, ErrForkInSubSession):
 			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
 		}
 		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("failed to fork session: %v", err))
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -77,6 +77,7 @@ func (s *Server) registerRoutes() {
 	group.PATCH("/sessions/:id/starred", s.setSessionStarred)
 	group.GET("/sessions/:id/models", s.getSessionModels)
 	group.POST("/sessions", s.createSession)
+	group.POST("/sessions/:id/fork", s.forkSession)
 	group.DELETE("/sessions/:id", s.deleteSession)
 	group.POST("/sessions/:id/agent/:agent", s.runAgent)
 	group.POST("/sessions/:id/agent/:agent/:agent_name", s.runAgent)
@@ -238,6 +239,43 @@ func (s *Server) createSession(c echo.Context) error {
 	}
 
 	return c.JSON(http.StatusOK, sess)
+}
+
+// forkSession creates a new session whose history is a deep copy of an
+// existing session up to (but excluding) the message at MessageIndex. The
+// fork point must be a user message; the new session uses a
+// fork-numbered title derived from the parent and starts with no runtime
+// attached. Clients are expected to prefill the excluded user message into
+// their chat input for the user to edit and resubmit.
+func (s *Server) forkSession(c echo.Context) error {
+	var req api.ForkSessionRequest
+	if err := c.Bind(&req); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("invalid request body: %v", err))
+	}
+
+	forked, err := s.sm.ForkSession(c.Request().Context(), c.Param("id"), req.MessageIndex)
+	if err != nil {
+		switch {
+		case errors.Is(err, ErrForkInvalidMessage):
+			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+		case strings.Contains(err.Error(), "out of range"),
+			strings.Contains(err.Error(), "sub-session"):
+			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+		}
+		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("failed to fork session: %v", err))
+	}
+
+	return c.JSON(http.StatusOK, api.SessionResponse{
+		ID:            forked.ID,
+		Title:         forked.Title,
+		CreatedAt:     forked.CreatedAt,
+		Messages:      forked.GetAllMessages(),
+		ToolsApproved: forked.ToolsApproved,
+		InputTokens:   forked.InputTokens,
+		OutputTokens:  forked.OutputTokens,
+		WorkingDir:    forked.WorkingDir,
+		Permissions:   forked.Permissions,
+	})
 }
 
 func (s *Server) getSession(c echo.Context) error {
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -375,10 +375,26 @@ func (sm *SessionManager) CreateSession(ctx context.Context, sessionTemplate *se
 	return sess, sm.sessionStore.AddSession(ctx, sess)
 }
 
-// ErrForkInvalidMessage is returned when a fork is requested at a message
-// index that does not point at a user-role message (e.g. an assistant turn,
-// a tool response, or a sub-session item).
-var ErrForkInvalidMessage = errors.New("fork point must be a user message")
+// Sentinel errors returned by ForkSession. They are matched with
+// errors.Is by the HTTP handler to classify failures as 400 vs 500, so
+// rewording the error string later is safe — the classification stays
+// pinned to the sentinel rather than to the message text.
+var (
+	// ErrForkInvalidMessage is returned when the resolved fork point is
+	// not a user-role message (e.g. an assistant turn, a tool response,
+	// or a sub-session item).
+	ErrForkInvalidMessage = errors.New("fork point must be a user message")
+
+	// ErrForkOutOfRange is returned when the requested message index is
+	// outside the parent session's visible message list.
+	ErrForkOutOfRange = errors.New("fork message index out of range")
+
+	// ErrForkInSubSession is returned when the requested message index
+	// falls inside a sub-session, where positional forking would be
+	// ambiguous (a sub-session can carry many messages but the fork
+	// model is "stop before this user turn in the parent's timeline").
+	ErrForkInSubSession = errors.New("fork message index falls inside a sub-session")
+)
 
 // ForkSession creates a new session whose history is a deep copy of the
 // parent session up to (but excluding) the message at userMessageIndex, with
@@ -387,10 +403,17 @@ var ErrForkInvalidMessage = errors.New("fork point must be a user message")
 // indexing that api.SessionResponse.Messages exposes to clients — so callers
 // do not need to know about the underlying Item slice.
 //
-// The message at userMessageIndex must be a user message; otherwise
-// ErrForkInvalidMessage is returned. The fork itself does not include that
-// message, so a client can prefill it into its chat input to let the user
-// edit and resubmit.
+// The message at userMessageIndex must be a user-role message in the
+// parent's visible list; values outside `[0, visibleMessageCount)` return
+// ErrForkOutOfRange and non-user messages return ErrForkInvalidMessage.
+// The fork itself does not include that message, so a client can prefill
+// it into its chat input to let the user edit and resubmit.
+//
+// There is intentionally no "full clone" shortcut on this endpoint: every
+// fork must anchor at a real user turn so the resulting history ends
+// cleanly between a user message and its (excluded) follow-up. Callers
+// that need a whole-session duplicate should issue a separate request
+// targeting the index of the most recent user message.
 func (sm *SessionManager) ForkSession(ctx context.Context, sessionID string, userMessageIndex int) (*session.Session, error) {
 	parent, err := sm.sessionStore.GetSession(ctx, sessionID)
 	if err != nil {
@@ -402,11 +425,9 @@ func (sm *SessionManager) ForkSession(ctx context.Context, sessionID string, use
 		return nil, err
 	}
 
-	if itemIndex < len(parent.Messages) {
-		item := parent.Messages[itemIndex]
-		if item.Message == nil || item.Message.Message.Role != chat.MessageRoleUser {
-			return nil, ErrForkInvalidMessage
-		}
+	item := parent.Messages[itemIndex]
+	if item.Message == nil || item.Message.Message.Role != chat.MessageRoleUser {
+		return nil, ErrForkInvalidMessage
 	}
 
 	forked, err := session.ForkSession(parent, itemIndex)
@@ -423,11 +444,12 @@ func (sm *SessionManager) ForkSession(ctx context.Context, sessionID string, use
 // flatMessageIndexToItemIndex maps an index in the flat user-visible message
 // list (Session.GetAllMessages — which skips system messages and flattens
 // sub-sessions) into an index in the parent's Session.Messages Item slice.
-// Returns an error if the index is out of range or lands inside a
-// sub-session, where positional forking would be ambiguous.
+// Returns ErrForkOutOfRange if the index is outside the visible list, or
+// ErrForkInSubSession if the index lands inside a sub-session, where
+// positional forking would be ambiguous.
 func flatMessageIndexToItemIndex(s *session.Session, flatIdx int) (int, error) {
 	if flatIdx < 0 {
-		return 0, fmt.Errorf("message index %d out of range", flatIdx)
+		return 0, fmt.Errorf("%w: %d", ErrForkOutOfRange, flatIdx)
 	}
 	seen := 0
 	for i, item := range s.Messages {
@@ -445,16 +467,16 @@ func flatMessageIndexToItemIndex(s *session.Session, flatIdx int) (int, error) {
 		case item.IsSubSession():
 			subCount := len(item.SubSession.GetAllMessages())
 			if flatIdx-seen < subCount {
-				return 0, fmt.Errorf("cannot fork at message index %d: position falls inside a sub-session", flatIdx)
+				return 0, fmt.Errorf("%w at index %d", ErrForkInSubSession, flatIdx)
 			}
 			seen += subCount
 		}
 	}
-	// Allow flatIdx == seen, meaning "after the last message" — a full clone.
-	if flatIdx == seen {
-		return len(s.Messages), nil
-	}
-	return 0, fmt.Errorf("message index %d out of range", flatIdx)
+	// flatIdx must point at an existing visible message. Values at or
+	// past the end (including the "after the last message" full-clone
+	// shortcut that earlier revisions allowed) are now rejected so the
+	// user-role check in ForkSession can never be skipped.
+	return 0, fmt.Errorf("%w: %d", ErrForkOutOfRange, flatIdx)
 }
 
 // GetSessions retrieves all sessions.
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -19,6 +19,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/docker/docker-agent/pkg/api"
+	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/concurrent"
 	"github.com/docker/docker-agent/pkg/config"
 	"github.com/docker/docker-agent/pkg/runtime"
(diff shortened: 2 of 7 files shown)


pkg/server/session_manager.go
```diff
@@ -374,6 +375,88 @@ func (sm *SessionManager) CreateSession(ctx context.Context, sessionTemplate *se
...
+	if err != nil {
+		return nil, err
+	}
+
+	itemIndex, err := flatMessageIndexToItemIndex(parent, userMessageIndex)
+	if err != nil {
+		return nil, err
+	}
+
+	if itemIndex < len(parent.Messages) {
```
docker-agent: **[MEDIUM] Full-clone boundary bypasses user-message role validation**

When `message_index` equals the total number of visible messages, `flatMessageIndexToItemIndex` returns `itemIndex == len(parent.Messages)` (the intentional "full clone" path). The subsequent guard:

```go
if itemIndex < len(parent.Messages) {
    item := parent.Messages[itemIndex]
    if item.Message == nil || item.Message.Message.Role != chat.MessageRoleUser {
        return nil, ErrForkInvalidMessage
    }
}
```

evaluates to `false`, so no role check is performed. A caller can pass any value that resolves to a full clone — even if conceptually the last real message is an assistant turn — and receive a silent success instead of `ErrForkInvalidMessage`.

The API doc comment says `MessageIndex` must point at a user message. Clients that pass `message_index = N` (total count) expecting a strict "fork before this user turn" contract may get a full clone silently instead of a validation error, which could confuse callers about what the new session contains.

**Recommendation**: Either document explicitly that `message_index == total_count` is a special "full clone" shortcut that skips role validation, or add a check that the last item in the parent is a user message when the full-clone path is taken.

pkg/server/server.go
```diff
@@ -240,6 +241,43 @@ func (s *Server) createSession(c echo.Context) error {
...
+	if err := c.Bind(&req); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("invalid request body: %v", err))
+	}
+
+	forked, err := s.sm.ForkSession(c.Request().Context(), c.Param("id"), req.MessageIndex)
+	if err != nil {
+		switch {
+		case errors.Is(err, ErrForkInvalidMessage):
+			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+		case strings.Contains(err.Error(), "out of range"),
```
docker-agent: **[LOW] Fragile string-matching to classify fork errors as HTTP 400 vs 500**

The error dispatch uses `strings.Contains` on error message text to identify 400-worthy errors:

```go
case strings.Contains(err.Error(), "out of range"),
    strings.Contains(err.Error(), "sub-session"):
    return echo.NewHTTPError(http.StatusBadRequest, err.Error())
```

These strings (`"out of range"`, `"sub-session"`) come from `flatMessageIndexToItemIndex`, which is a package-private function. If those error messages are ever rephrased — even trivially (e.g., `"index out of bounds"`, `"inside a sub-session"`) — the handler will silently start returning HTTP 500 for what are actually client input errors, breaking API contracts without any compile-time warning.

**Recommendation**: Introduce sentinel errors (e.g., `ErrForkOutOfRange`, `ErrForkInSubSession`) alongside `ErrForkInvalidMessage`, and use `errors.Is` throughout. This makes the 400 vs 500 classification robust to future refactoring.

pkg/session/branch.go
```diff
@@ -191,6 +203,27 @@ func generateBranchTitle(parentTitle string) string {
...
+// generateForkTitle creates a title for a forked session based on the parent
+// title using "(fork N)" suffixes that increment on repeated forks.
+// If the parent has no title, returns empty string (will trigger auto-generation).
+// "<title>" -> "<title> (fork 1)"; "<title> (fork N)" -> "<title> (fork N+1)".
+func generateForkTitle(parentTitle string) string {
+	if parentTitle == "" {
+		return ""
+	}
+
+	if idx := strings.LastIndex(parentTitle, "(fork "); idx >= 0 {
```
docker-agent: **[LOW] `generateForkTitle` silently drops title content after `(fork N)` in mid-title positions**

`strings.LastIndex(parentTitle, "(fork ")` finds the *last* occurrence of the substring, not just a trailing suffix. For a title like `"Project (fork 1) note"`:

1. `idx` points at `(fork 1)` mid-title
2. `suffix = "(fork 1) note"`
3. `fmt.Sscanf(suffix, "(fork %d)", &n)` **succeeds** with `n=1` — Go's `Sscanf` does not require consuming the full string, so `1)` matches and ` note` is silently ignored
4. `baseTitle = strings.TrimRight("Project ", " \t")` = `"Project"`
5. Result: `"Project (fork 2)"` — the ` note` portion is **permanently lost**

This can happen with any user-generated title that contains `(fork N)` anywhere other than the very end (e.g., a project named `"Q1 (fork 2) Analysis"`).

**Recommendation**: Anchor the match so it only fires on a genuine trailing suffix — for example, require `idx + len(suffix) == len(parentTitle)` after scanning, or use a regexp with `$`:

```go
var forkSuffixRe = regexp.MustCompile(`^(.*?)\s+\(fork (\d+)\)$`)
```

docker-agent (commented):

Reviewed **feat(server): expose session forking over HTTP** (PR #3199 by `trungutt`).

2 verified findings (1 medium, 2 low). 1 medium hypothesis was **dismissed** (nil-guard on `IsSubSession()` is implicit and correct). 1 low finding (data race) was not reported per severity threshold.

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🟡 Medium | `session_manager.go:405` | Full-clone boundary silently bypasses user-message role validation |
| 2 | 🔵 Low | `server.go:261` | String-matching on error text to classify 400 vs 500 (fragile) |
| 3 | 🔵 Low | `branch.go:215` | `generateForkTitle` drops title content when `(fork N)` appears mid-title |

pkg/server/session_manager.go
trungutt: Good catch — fixed in 629d83df. I went with "tighten the contract" rather than "document the shortcut": the full-clone path is now rejected with `ErrForkOutOfRange`, so a caller has to pass the index of a real user turn. That eliminates the silent role-check bypass entirely. Whole-session duplicates are still expressible by passing the index of the most recent user message (the TUI keeps its own in-process full-clone path via `BranchSession(s, len(s.Messages))` for sessions that legitimately want a 1:1 copy).

pkg/server/server.go
trungutt: Fixed in 629d83df. Introduced `ErrForkOutOfRange` and `ErrForkInSubSession` alongside the existing `ErrForkInvalidMessage`; the handler now classifies via `errors.Is` over all three sentinels. `flatMessageIndexToItemIndex` wraps the sentinels with `%w` so future error rephrasing keeps the 400 status. Added an HTTP test pinning the out-of-range path as 400 so a regression here would fail the build.

pkg/session/branch.go
trungutt: Fixed in 629d83df. Switched to an end-anchored regexp (`^(.*?)[ \t]*\(fork (\d+)\)$`) so a mid-title `(fork N)` is not detected as a suffix. Added a regression test for `"Q1 (fork 2) Analysis"` and for a benign title-ending `(notes)`.

Worth flagging: `generateBranchTitle` (TUI's branch-from-edit flow) has the same `LastIndex+Sscanf` pattern and the same gap, but that's pre-existing code this PR only refactored at the call-site level. Leaving it alone here to keep the diff scoped — happy to do the matching fix in a follow-up.

pkg/server/session_manager.go
docker-agent: Perfect — removing the full-clone shortcut entirely eliminates the bypass and makes the contract unambiguous: every fork must target a real visible user message. The approach of surfacing distinct error sentinels for the HTTP handler to match with errors.Is is cleaner than string-matching and future-proof for error message rewording.

Thanks for the thorough fix!

pkg/session/branch.go
docker-agent: Perfect — the end-anchored regexp fixes the issue cleanly, and the test cases for `"Q1 (fork 2) Analysis"` and `"My Session (notes)"` confirm both the regression and the benign-suffix scenario.

Thanks for flagging the parallel issue in `generateBranchTitle`. I agree with keeping this PR scoped; the TUI's branch flow isn't touched by this change, and a follow-up makes it easier to review/test independently. If you do open a follow-up, the same regexp approach would work there (maybe `branchSuffixRe` for `(branch N)` and a straightforward `HasSuffix` for the `(branched)` case).

pkg/server/server.go
docker-agent: Perfect! The fix is exactly what I recommended:

✅ **Sentinel errors introduced** (, , )  
✅ **Handler uses ** for 400 classification — immune to message rewording  
✅ ** wraps sentinels with ** — preserves classification through wrapping  
✅ **HTTP test added** () — pinning the 400 status prevents silent regression

This eliminates the fragility completely. API contracts are now enforced at the type level, and future refactoring of error messages won't break HTTP status classification. Thank you for the thorough implementation!

pkg/server/session_manager.go
```diff
@@ -374,6 +375,110 @@ func (sm *SessionManager) CreateSession(ctx context.Context, sessionTemplate *se
...
+// The fork itself does not include that message, so a client can prefill
+// it into its chat input to let the user edit and resubmit.
+//
+// There is intentionally no "full clone" shortcut on this endpoint: every
+// fork must anchor at a real user turn so the resulting history ends
+// cleanly between a user message and its (excluded) follow-up. Callers
+// that need a whole-session duplicate should issue a separate request
+// targeting the index of the most recent user message.
+func (sm *SessionManager) ForkSession(ctx context.Context, sessionID string, userMessageIndex int) (*session.Session, error) {
+	parent, err := sm.sessionStore.GetSession(ctx, sessionID)
```
docker-agent, line 427: **[LIKELY / MEDIUM] ForkSession may inherit a stale title when auto-title generation is in-flight**

`ForkSession` always reads the parent session from the store:

```go
parent, err := sm.sessionStore.GetSession(ctx, sessionID)
```

However, `generateTitle` runs in a background goroutine (launched at line ~681):

```go
go sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
```

This goroutine updates `sess.Title` in memory and then calls `sm.sessionStore.UpdateSession` — but only after the title has been generated (an async LLM call). There is a window between when the goroutine starts and when it persists the new title, during which `sm.sessionStore.GetSession` returns the old/empty title.

If a client POSTs to `/api/sessions/:id/fork` while title generation is still running (i.e., immediately after the first exchange in the parent session), the forked session will be titled `" (fork 1)"` (empty base) or inherit the placeholder title instead of the final auto-generated one.

In contrast, `GetSessionSnapshot` explicitly prefers the live in-memory session to avoid exactly this issue:
```go
if rs, ok := sm.runtimeSessions.Load(id); ok {
    sess = rs.session  // freshest title
}
```

**Suggested fix**: Apply the same live-session-first pattern in `ForkSession` — load from `runtimeSessions` when available, fall back to the store otherwise. If the live session is used, call `session.ForkSession` with the live session's `Title` field only (to avoid races on `Messages`, which belongs to the runtime).

pkg/server/session_manager.go
```diff
@@ -374,6 +375,110 @@ func (sm *SessionManager) CreateSession(ctx context.Context, sessionTemplate *se
...
+	if item.Message == nil || item.Message.Message.Role != chat.MessageRoleUser {
+		return nil, ErrForkInvalidMessage
+	}
+
+	forked, err := session.ForkSession(parent, itemIndex)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := sm.sessionStore.AddSession(ctx, forked); err != nil {
```
docker-agent, line 447: **[LIKELY / MEDIUM] Concurrent fork requests can produce sessions with duplicate titles**

`ForkSession` performs a read-then-write on the session store without holding `sm.mux`:

```go
func (sm *SessionManager) ForkSession(...) (*session.Session, error) {
    parent, err := sm.sessionStore.GetSession(ctx, sessionID)  // no sm.mux
    // ...
    forked, err := session.ForkSession(parent, itemIndex)      // title computed here
    // ...
    if err := sm.sessionStore.AddSession(ctx, forked); err != nil {  // no sm.mux
        return nil, err
    }
}
```

Two simultaneous `POST /api/sessions/:id/fork` requests will both read `parent.Title = "foo"`, both compute `generateForkTitle("foo") = "foo (fork 1)"`, and both call `AddSession` — creating two separate sessions both named `"foo (fork 1)"`. Subsequent forks of either will then produce `"foo (fork 2)"` names that also collide.

By contrast, other mutating methods (`DeleteSession`, `UpdateSessionTitle`) hold `sm.mux` for exactly this reason.

While sessions are keyed by UUID so there is no functional correctness issue, duplicate titles can confuse users browsing their session list and break the invariant documented in the PR description ("(fork 1)", "(fork 2)" counting).

**Suggested fix**: Wrap the `GetSession` → `ForkSession` → `AddSession` sequence under `sm.mux.Lock()`, consistent with the pattern used by `DeleteSession`.

pkg/session/branch.go
```diff
@@ -21,7 +41,7 @@ func BranchSession(parent *Session, branchAtPosition int) (*Session, error) {
 	}
 
 	branched := New()
-	copySessionMetadata(branched, parent, generateBranchTitle(parent.Title))
+	copySessionMetadata(branched, parent, titleFn(parent.Title))
```
docker-agent, line 44: **[LIKELY / MEDIUM] copySessionMetadata silently drops MaxConsecutiveToolCalls and MaxOldToolCallTokens**

The `branchSessionWithTitle` function (called by both `BranchSession` and the new `ForkSession`) invokes `copySessionMetadata` here, which does not copy these two session-level configuration fields:

```go
// copySessionMetadata omits:
// MaxConsecutiveToolCalls  — NOT copied (zeroed in forked session)
// MaxOldToolCallTokens     — NOT copied (zeroed in forked session)

// But MaxIterations IS copied:
dst.MaxIterations = src.MaxIterations
```

In contrast, `Clone()` (the full deep-copy) correctly copies both:

```go
MaxConsecutiveToolCalls: s.MaxConsecutiveToolCalls,
MaxOldToolCallTokens:    s.MaxOldToolCallTokens,
```

`MaxConsecutiveToolCalls` controls how many consecutive tool calls are permitted before the agent is interrupted — a safety rail that a user or operator may have deliberately configured. `MaxOldToolCallTokens` controls context-management budget. A forked session silently resets both to `0` (which may mean "unlimited" or "use built-in default"), causing the fork to behave differently from its parent with no indication to the user or API caller.

Since `MaxIterations` — a structurally similar limit — *is* copied, the exclusion appears inconsistent rather than principled. The PR description mentions "happy to revisit in a follow-up," but this creates a silent behavioral divergence from the parent today.

**Suggested fix**: Add `dst.MaxConsecutiveToolCalls = src.MaxConsecutiveToolCalls` and `dst.MaxOldToolCallTokens = src.MaxOldToolCallTokens` to `copySessionMetadata`, or add a comment explaining the intentional reset.

docker-agent (commented):

3 medium-severity findings in the new `ForkSession` code path. The core index translation, error sentinel wiring, and deep-copy logic are sound. Issues centre on title staleness under async title generation, missing mutex around the fork store sequence, and silently dropped session config fields.

pkg/server/session_manager.go
trungutt: Acknowledging but deferring this one.

The race is real but narrow: a client has to POST `/fork` during the window between `generateTitle`'s in-memory `Title = ...` write and its `UpdateSession` persist. In practice the downstream UI that motivates this PR only renders the fork affordance for sessions whose messages have been loaded from `GET /sessions/:id` (i.e. post-stream, typically post-title-gen), so the race is hard to hit from there.

The proposed fix (load from `runtimeSessions` when attached, else from the store) has a subtlety: we can't cleanly take `Title` from the live in-memory session while taking `Messages` from the store snapshot — they could be from different consistency points, and reading `sess.Title` without holding `Session.mu` is itself racy with `generateTitle`'s writer. A proper fix needs to hold `Session.mu.RLock()` on the parent for the full clone, which is a bigger restructuring than this PR.

Worst-case outcome today is a fork titled `" (fork 1)"` — ugly but recoverable via `PATCH /sessions/:id/title`. Tracking as a follow-up; would rather land that with the matching `Session.mu` change in `BranchSession` than half-fix it here.

pkg/server/session_manager.go
trungutt: Good call — fixed in ddc183a8. `ForkSession` now holds `sm.mux` for the full `GetSession` → `session.ForkSession` → `AddSession` sequence, matching `DeleteSession` and `RunSession`. Two concurrent fork requests on the same parent now serialise, so the title computation sees a consistent parent state and the auto-numbered counter advances cleanly (`(fork 1)` then `(fork 2)`) instead of producing collision pairs.

pkg/session/branch.go
trungutt: Fair pushback — fixed in ddc183a8. Added the two fields to `copySessionMetadata` with a comment explaining the rationale (`Clone()` already preserves them; `MaxIterations` is preserved; the omission was inconsistent rather than principled). Added a regression test that pins all three limits on a fork.

This also fixes the same behavioural drift in the TUI's branch-from-edit flow, which shares the helper. That's a TUI-side change in this PR, but it's a strict bug-fix (the rails the user/operator configured no longer silently disappear on branch), so I think it's fine to land together.

docker-agent: Thank you for the quick fix! I've reviewed commit ddc183a8 and the approach looks solid.

Adding the two fields to `copySessionMetadata` with the clarifying comment resolves the inconsistency, and the regression test ensures all three limits (MaxIterations, MaxConsecutiveToolCalls, MaxOldToolCallTokens) are preserved across forks. You're absolutely right that the TUI branch-from-edit flow benefits from the same fix — it's a bug fix that restores user-configured safety rails, so landing it together makes sense.

The rationale is clear: these are session-level configuration values that the operator or user explicitly set, and there's no principled reason to treat them differently from MaxIterations. Appreciate the thoughtful response and the test coverage.

pkg/server/session_manager.go
docker-agent: Fair — thanks for the detailed context. The locking subtlety (can't safely cherry-pick Title from the live session without holding Session.mu for the full clone) is a good point, and deferring until you can land the proper Session.mu restructure for both ForkSession and BranchSession makes sense.

Given the narrow window and the UI's typical access pattern (fork affordance only shown post-load), treating this as a follow-up is reasonable. Tracking it so the fix lands with the matching concurrency primitives.

pkg/server/session_manager.go
docker-agent: Perfect — thanks for the quick fix! The lock now serializes the read-fork-write sequence, so concurrent fork requests will see each other's title updates and the (fork N) counter will increment cleanly instead of producing collisions. This matches the pattern used by DeleteSession and UpdateSessionTitle, so the concurrency semantics are consistent across all session-mutating operations.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR exposes `session.ForkSession` over HTTP with solid implementation across the stack.

**What was reviewed:** 7 files — `pkg/api/types.go`, `pkg/server/server.go`, `pkg/server/server_test.go`, `pkg/server/session_manager.go`, `pkg/server/session_manager_test.go`, `pkg/session/branch.go`, `pkg/session/branch_test.go`

**No bugs found.** Key correctness properties verified:

- **Mutex serialization**: `sm.mux.Lock()` is held across the full `GetSession → ForkSession → AddSession` sequence, preventing concurrent forks from racing to the same parent title counter (e.g., two requests both seeing no existing forks and both computing "(fork 1)"). ✅
- **Safety-rail limits preserved**: `copySessionMetadata` now includes `MaxConsecutiveToolCalls` and `MaxOldToolCallTokens`, matching `Clone()` behavior. User-configured tool-call cutoff and token truncation budget carry over to the fork. ✅
- **Index translation correctness**: `flatMessageIndexToItemIndex` correctly mirrors `GetAllMessages`'s system-message skip rule, and the off-by-one boundary (equal to visible count → `ErrForkOutOfRange`) prevents bypassing the role check. ✅
- **Error classification**: Sentinel errors (`ErrForkInvalidMessage`, `ErrForkOutOfRange`, `ErrForkInSubSession`) are matched with `errors.Is` so future rewording of error strings cannot silently flip HTTP status codes. ✅
- **Fork title regex**: `forkSuffixRe` is anchored at the end of the string (`$`) so a mid-title "(fork N)" (e.g. "Q1 (fork 2) Analysis") is not treated as a suffix — the trailing content is preserved. ✅
- **Test coverage**: Unit tests cover happy path, non-user message rejection, out-of-range (negative, equal-to-count, past-end), deep-copy isolation, system-message skip, sub-session boundary, and safety-rail preservation. End-to-end server test covers the full HTTP path including persistence verification. ✅

dgageot (approved):

pkg/server/server.go
```diff
@@ -240,6 +241,42 @@ func (s *Server) createSession(c echo.Context) error {
...
+	if err := c.Bind(&req); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("invalid request body: %v", err))
+	}
+
+	forked, err := s.sm.ForkSession(c.Request().Context(), c.Param("id"), req.MessageIndex)
+	if err != nil {
+		switch {
+		case errors.Is(err, ErrForkInvalidMessage),
+			errors.Is(err, ErrForkOutOfRange),
+			errors.Is(err, ErrForkInSubSession):
```
docker-agent, line 261: **[HIGH] Missing 404 response when parent session does not exist — returns 500 instead**

The `forkSession` error switch handles `ErrForkInvalidMessage`, `ErrForkOutOfRange`, and `ErrForkInSubSession` (all → 400), but has no case for `session.ErrNotFound`. When a client POSTs to `/api/sessions/<unknown-id>/fork`, `sm.sessionStore.GetSession` returns `session.ErrNotFound`, `ForkSession` propagates it unwrapped (`return nil, err`), and the handler falls through to the 500 branch — contradicting the PR spec ("404: parent session not found") and the pattern every other session handler follows.

**Fix** — add a `session.ErrNotFound` case to the switch:

```go
switch {
case errors.Is(err, session.ErrNotFound):
    return echo.NewHTTPError(http.StatusNotFound, err.Error())
case errors.Is(err, ErrForkInvalidMessage),
    errors.Is(err, ErrForkOutOfRange),
    errors.Is(err, ErrForkInSubSession):
    return echo.NewHTTPError(http.StatusBadRequest, err.Error())
}
```

A matching test case in `TestServer_ForkSession` would pin this behaviour.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One confirmed high-severity bug in the newly introduced `forkSession` handler: a missing 404 path causes the server to return 500 Internal Server Error when the parent session ID does not exist. All other details of the PR (index translation, title generation, deep-copy, concurrency) look correct.

PR #3206, opened 2026-06-22
fix(dmr): detect and use locally-installed Docker Model Runner models

### LINKED ISSUE #2799: No model providers found - but Docker Model Runner already setup with models
**OS** macOS 26.4.1 / **Docker Agent** v1.54.0

1. When I ran `docker-agent` for the first time, I was prompted to pull `ai/qwen3:latest`. First I declined because I already have some models installed locally via Docker Model Runner.

After declining, the CLI showed:

```sh
No model providers available.
```

This seems unexpected. I would assume Docker Model Runner acts as a local model provider service and that docker agent would connect to it automatically.

2. To further test the provider availability I downloaded `ai/qwen3:latest` – started `docker-agent`, but when I tried to list the models from Docker Model Runner (ctrl+m) the UI just showed …

```
No models available for selection
```

… which is a bit confusing, since at least `ai/qwen3:latest` should be selectable, even if its already active. And as you can see in the screenshot: Docker Model Runner is listed as the Provider `dmr` of the qwen model. Proof that DMR is already active as an provider but still not discovered by `docker-agent` when invoked with `ctrl+m`.

Image

## Expected Behavior

1. Models available through Docker Model Runner should be detected and listed automatically.
2. List Models works correctly when inside of `docker-agents`

## Enhancement Request

For local agentic workflows, please consider supporting LM Studio and other local endpoints as well 😋🙏

### Docker Agent Output

```
$ docker agent
Model ai/qwen3:latest not found locally.
Do you want to pull it now? ([y]es/[n]o): n
No model providers available.
To fix this, you can:
  - Install Docker Model Runner: https://docs.docker.com/ai/model-runner/get-started/
  - Configure an API key for a cloud provider:
    - anthropic: ANTHROPIC_API_KEY
    - openai: OPENAI_API_KEY
    - google: GOOGLE_API_KEY (or GEMINI_API_KEY, GOOGLE_GENAI_USE_VERTEXAI)
    - mistral: MISTRAL_API_KEY
    - amazon-bedrock: AWS_ACCESS_KEY_ID (or AWS_PROFILE, AWS_ROLE_ARN, AWS_BEARER_TOKEN_BEDROCK)
```
---

### Docker Model Runner: Models Are Available

Image

Sayt-0:
## Summary

Fixes #2799. Docker Model Runner (DMR) models pulled locally were never enumerated, causing two reported problems:

| Symptom | Cause | Handled by |
|---|---|---|
| Declining the `ai/qwen3:latest` pull showed "No model providers available." although DMR had other models | Auto-selection hard-coded the DMR provider to `DefaultModels["dmr"]` (`ai/qwen3:latest`); declining the pull failed selection | Auto-selection now prefers an already-pulled model |
| ctrl+m showed "No models available for selection" although DMR had models | The picker only listed configured models plus the models.dev catalog; DMR locals are not in the catalog | DMR models are now discovered and listed |

## Changes

| Area | File | Change |
|---|---|---|
| DMR discovery | `pkg/model/provider/dmr/list.go` (new) | `ListModels` queries DMR's OpenAI `/models` endpoint, reusing the same endpoint/baseURL/httpClient resolution as the inference client (handles `MODEL_RUNNER_HOST` and the Docker Desktop unix-socket transport) |
| Auto-selection | `pkg/config/auto.go` | `AutoModelConfig` takes a `DMRModelLister`; `pickDMRAutoModel` prefers the configured default (including under a non-default tag), else the first non-embedding installed model, else the static default |
| Fallback error | `pkg/config/auto.go`, `pkg/teamloader/teamloader.go` | `AutoModelFallbackError` carries the cause (e.g. a declined pull), unwraps it, and suggests `docker model pull` |
| Model picker | `pkg/runtime/dmr_models.go` (new), `model_switcher.go` | `buildDMRChoices` mirrors the gateway discovery path (TTL cache + singleflight), filters embedding models, dedupes against configured models, and groups under "Other models" |
| Wiring | `runtime.go`, `teamloader.go`, `cmd/root/models.go` | Lister defaulted in `NewLocalRuntime`; the run path passes `dmr.ListModels`; `docker agent models` passes nil to stay side-effect-free |

## Behavior

- DMR model IDs contain slashes (`ai/qwen3:latest`); the picker ref is `dmr/<id>` and round-trips through `ParseModelRef` (first-slash split) back to `{provider: dmr, model: ai/qwen3:latest}`, so selecting one does not re-trigger a pull.
- Runtimes constructed directly (tests) get a nil lister and therefore no DMR entries, keeping existing picker assertions deterministic.
- When no model is available at all, the message now names the underlying cause and points at `docker model pull`.

## Tests

New coverage (35 cases): DMR `/models` parsing and the exported `ListModels` entry point (`MODEL_RUNNER_HOST` + unreachable), auto-selection across installed/default/tagged/embedding-only/error cases, the cause-carrying fallback error, and `buildDMRChoices` including catalog-family embedding filtering, dedup, and cache TTL/failure behavior.

`go build ./...`, `go vet`, `gofmt`, and `golangci-lint` are clean on the changed packages.

## Not included

The enhancement request for LM Studio and other local endpoints is left as a separate follow-up; this PR is scoped to the two reported DMR defects.

### DIFF
--- a/pkg/config/auto.go
+++ b/pkg/config/auto.go
@@ -3,11 +3,20 @@ package config
 import (
 	"context"
 	"fmt"
+	"log/slog"
+	"slices"
 	"strings"
 
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
 )
+
+// DMRModelLister returns the IDs of the models currently available to Docker
+// Model Runner (i.e. pulled locally). It is injected so DMR discovery can be
+// stubbed in tests and disabled by callers that must stay side-effect-free:
+// `docker agent models` passes nil to avoid shelling out to `docker model`,
+// while the agent run path passes dmr.ListModels.
+type DMRModelLister func(ctx context.Context) ([]string, error)
 
 // providerConfig defines a cloud provider and how to detect/describe its API keys.
 type providerConfig struct {
@@ -37,8 +46,15 @@ var cloudProviders = []providerConfig{
 }
 
 // AutoModelFallbackError is returned when auto model selection fails because
-// no providers are available (no API keys configured and DMR not installed).
-type AutoModelFallbackError struct{}
+// no model could be initialized (no API keys configured and no usable Docker
+// Model Runner model, e.g. DMR not installed or the pull was declined).
+type AutoModelFallbackError struct {
+	// Cause is the underlying provider-initialization error, when available
+	// (for example "model pull declined by user"). It is surfaced in the
+	// message so the user understands why selection fell through, and exposed
+	// via Unwrap for errors.Is/As callers.
+	Cause error
+}
 
 func (e *AutoModelFallbackError) Error() string {
 	var hints []string
@@ -46,11 +62,21 @@ func (e *AutoModelFallbackError) Error() string {
 		hints = append(hints, fmt.Sprintf("    - %s: %s", p.name, p.hint))
 	}
 
-	return "No model providers available.\n\nTo fix this, you can:\n" +
-		"  - Install Docker Model Runner: https://docs.docker.com/ai/model-runner/get-started/\n" +
-		"  - Configure an API key for a cloud provider:\n" +
-		strings.Join(hints, "\n")
+	var b strings.Builder
+	if e.Cause != nil {
+		fmt.Fprintf(&b, "Could not initialize the auto-selected model: %v\n\n", e.Cause)
+	}
+	b.WriteString("No model is currently available.\n\nTo fix this, you can:\n")
+	b.WriteString("  - Pull a Docker Model Runner model, e.g. `docker model pull ai/qwen3`\n")
+	b.WriteString("  - Install Docker Model Runner: https://docs.docker.com/ai/model-runner/get-started/\n")
+	b.WriteString("  - Configure an API key for a cloud provider:\n")
+	b.WriteString(strings.Join(hints, "\n"))
+	return b.String()
 }
+
+// Unwrap exposes the underlying initialization error so callers can inspect it
+// with errors.Is/errors.As.
+func (e *AutoModelFallbackError) Unwrap() error { return e.Cause }
 
 var DefaultModels = map[string]string{
 	"openai":         "gpt-5",
@@ -84,7 +110,7 @@ func AvailableProviders(ctx context.Context, modelsGateway string, env environme
 	return providers
 }
 
-func AutoModelConfig(ctx context.Context, modelsGateway string, env environment.Provider, defaultModel *latest.ModelConfig) latest.ModelConfig {
+func AutoModelConfig(ctx context.Context, modelsGateway string, env environment.Provider, defaultModel *latest.ModelConfig, dmrLister DMRModelLister) latest.ModelConfig {
 	// If user specified a default model config, use it (with defaults for unset fields)
 	if defaultModel != nil && defaultModel.Provider != "" && defaultModel.Model != "" {
 		result := *defaultModel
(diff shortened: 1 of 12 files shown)


docker-agent (commented):

Two issues found in the new DMR discovery code. The medium-severity singleflight/context bug can suppress DMR model listings for all users for up to one minute after a single cancelled request; the low-severity embedding filter gap is a latent defect in the repo-prefix auto-selection fallback.

pkg/runtime/dmr_models.go
```diff
@@ -0,0 +1,154 @@
...
+
+	v, err, _ := c.sf.Do("models", func() (any, error) {
+		// Double-check the cache now that we hold the in-flight slot: a caller
+		// that read a stale cache right before a concurrent singleflight
+		// completed would otherwise trigger a redundant fetch.
+		if ids, ok, err := readFresh(); ok {
+			return ids, err
+		}
+
+		ids, err := r.dmrModelLister(ctx)
```
docker-agent, line 69: **[MEDIUM] Singleflight closure captures caller's context — a cancelled picker open poisons the cache for ~1 minute**

`listDMRModels` passes the first caller's `ctx` directly to `r.dmrModelLister`:

```go
ids, err := r.dmrModelLister(ctx)   // ctx from the *first* sf.Do caller
c.mu.Lock()
c.ids, c.err, c.fetchedAt = ids, err, now()   // error is cached here
c.mu.Unlock()
return ids, err
```

Two compounding problems arise from this design:

1. **Spurious errors for concurrent waiters.** `singleflight.Do` broadcasts the first caller's result to every goroutine waiting on the same key. If the first caller's context is cancelled (e.g. the user closes the model picker mid-request), `dmrModelLister` returns `context.Canceled`. All concurrent waiters receive that error even though their own contexts are still alive.

2. **Error is cached for the full TTL.** The code sets `c.fetchedAt = now()` unconditionally, so the `context.Canceled` result is stored and replayed to every subsequent caller for up to one minute (the `dmrModelsTTL`). During that window DMR models disappear from the picker for all users — recoverable only by waiting out the TTL.

**Suggested fix:** Use a context that is not tied to any single caller's lifetime inside the singleflight body. The simplest approach is `context.WithoutCancel(ctx)` (Go 1.21+), which propagates values but detaches cancellation:

```go
v, err, _ := c.sf.Do("models", func() (any, error) {
    if ids, ok, err := readFresh(); ok {
        return ids, err
    }
    ids, err := r.dmrModelLister(context.WithoutCancel(ctx))
    // ...
})
```

Additionally, consider not caching errors (or using a shorter error TTL) so a transient failure or cancellation doesn't block discovery for a full minute.

pkg/config/auto.go
```diff
@@ -97,13 +123,89 @@ func AutoModelConfig(ctx context.Context, modelsGateway string, env environment.
...
+	// who do have ai/qwen3:latest.
+	if slices.Contains(installed, defaultModel) {
+		return defaultModel
+	}
+
+	// The default model pulled under a different tag (e.g. ai/qwen3:Q4_K_M)
+	// still satisfies "prefer the default", so match on the repository.
+	defaultRepo := dmrModelRepo(defaultModel)
+	for _, m := range installed {
+		if dmrModelRepo(m) == defaultRepo {
```
docker-agent, line 171: **[LOW] Repo-prefix fallback in `pickDMRAutoModel` skips the embedding-model filter**

The general fallback loop (used when no exact tag match is found) correctly calls `looksLikeEmbeddingModel`:

```go
for _, m := range installed {
    if !looksLikeEmbeddingModel(m) {
        return m   // safe: chat models only
    }
}
```

But the repo-prefix loop that runs first does not apply the same guard:

```go
defaultRepo := dmrModelRepo(defaultModel)
for _, m := range installed {
    if dmrModelRepo(m) == defaultRepo {
        return m   // no embedding check here
    }
}
```

If an embedding-only variant that shares the same repository prefix as the default model is installed (e.g. a future `ai/qwen3-embed:latest` published under the `ai/qwen3` namespace), auto-selection could return it as the chat model. The agent would then start and fail when it tries to use a text-embedding model for conversation.

The fix is a one-liner:

```go
for _, m := range installed {
    if dmrModelRepo(m) == defaultRepo && !looksLikeEmbeddingModel(m) {
        return m
    }
}
```

dgageot (approved):

PR #3210, opened 2026-06-23
fix: add idle timeout and context cancellation to model stream reads

### LINKED ISSUE #3209: Session hangs indefinitely on stalled model stream; graceful stop fails and requires SIGKILL
## Summary
A `docker-agent run` session became permanently stuck after a sub-agent's LLM streaming request to the model gateway stalled and never returned. The process consumed 0% CPU with all threads parked in `pthread_cond_wait` (idle, not spinning). Attempting to stop the session did not terminate it; `SIGTERM` was ignored and only `SIGKILL` worked. This indicates the streaming HTTP read has no read/stream deadline and that graceful shutdown blocks waiting on the wedged read.

## Environment
- docker-agent run via Docker Desktop 4.81.0 (gordonTag v9-dev)
- macOS 26.5.1 (arm64)
- Multi-agent config (root agent delegating to sub-agents over MCP stdio)

## What happened
1. Root agent transferred a task to a sub-agent (`notion`).
2. The sub-agent opened an Anthropic `claude-sonnet-4-6` streaming chat completion to the model gateway.
3. The streaming response never arrived. A single outbound TCP connection to the gateway (`:443`) remained `ESTABLISHED` with no data flowing.
4. The debug log froze mid-request and never advanced again (~13 minutes idle when diagnosed).
5. User pressed stop; the session did not exit.

## Diagnosis / evidence
- Process state `S+`, 0% CPU; all goroutines blocked in `pthread_cond_wait` → idle wait, not a busy loop or CPU deadlock.
- One stuck `ESTABLISHED` TCP socket to the model gateway with no traffic.
- Debug log last line: opening the Anthropic streaming request, then silence.
- `SIGTERM` had no effect (graceful shutdown was itself blocked on the hung read); `SIGKILL` terminated it cleanly.

## Expected behavior
- The streaming model client should enforce a read/stream timeout (and honor context cancellation) so a stalled or half-open upstream connection fails fast instead of hanging the goroutine forever.
- Pressing stop / sending SIGTERM should cancel in-flight requests and shut the session down gracefully without requiring SIGKILL.

## Suggested fixes
- Add a read/stream deadline (idle timeout) to the streaming HTTP client used for model completions.
- Wire request context cancellation into the stream read loop so shutdown/stop interrupts a blocked read.
- Consider a watchdog that surfaces "no stream activity for N seconds" to the user and aborts the request.

## Notes
- A separate Docker Desktop `docker-agent serve api` background process was unaffected; only the interactive `docker-agent run` session was impacted.
- A local debug log capturing the hang exists at `/Users/arnaud/.cagent/cagent.debug-20260623-090647-2067.log` (not attached; available on request).

aheritier:
## Problem

A `docker-agent run` session hung permanently when a streaming LLM request to the model gateway stalled. The outbound TCP connection stayed ESTABLISHED but no data was flowing. Debug logs froze mid-request, CPU dropped to 0%, and SIGTERM was ignored — only SIGKILL worked.

**Root cause** (confirmed by code analysis):

The SSE stream-reading loop in `handleStream` blocked indefinitely in `stream.Recv()` → `bufio.Scanner.Scan()` → `body.Read()` → `syscall.Read`. There was no idle/read deadline and no explicit context-cancellation check between chunks, so:

1. A half-open TCP connection (remote side stopped sending data without closing) would park the goroutine forever.
2. SIGTERM (which cancels the root context) only helped if Go's `net/http` transport's `readLoop` goroutine happened to race ahead and call `pc.close()` before the `body.Read` blocked. Under scheduler contention or proxy configurations this race could be lost, making graceful shutdown unreliable.

### `pkg/runtime/streaming.go`

- **Idle timeout**: `stream.Recv()` is moved into a dedicated reader goroutine. The main loop selects between the result channel, `ctx.Done()`, and a **5-minute idle timer**. The timer resets on every received chunk. If no chunk arrives within the window the stream is declared stalled.
- **cancelStream hook**: on idle timeout, `cancelStream(errStreamIdle)` is called. This cancels the HTTP-request child context, which propagates to Go's `net/http` transport's `readLoop`; `pc.close()` closes the TCP socket and unblocks the goroutine that is stuck in `body.Read()`.
- **Context cancellation**: the `ctx.Done()` select arm returns `ctx.Err()` immediately, making SIGTERM / Ctrl-C terminate the stream without racing against the transport.
- **Goroutine cleanup**: a `done` channel (closed via `defer`) signals the reader goroutine to exit on any return path, preventing leaks.

### `pkg/runtime/fallback.go`

- Creates a `context.WithCancelCause` child context (`streamCtx`) for each stream attempt.
- Passes `streamCtx` to `CreateChatCompletionStream` **and** `handleStream`.
- The `streamCancel` function is the `cancelStream` hook described above.

## Tests

Two new tests in `pkg/runtime/streaming_test.go`:

| Test | What it verifies |
|------|-----------------|
| `TestHandleStream_IdleTimeout` | A never-sending stream times out in 50 ms; `errStreamIdle` is returned; `cancelStream` is called |
| `TestHandleStream_ContextCancellation` | A never-sending stream is interrupted by context cancellation in 20 ms; `context.Canceled` is returned |

## Validation

- `task build` ✅
- `task test` ✅ (pre-existing `TestLoadExamples/{dmr,unload_on_switch}` failures confirmed unrelated)
- `task lint` ✅ (0 new golangci-lint issues; 2 pre-existing custom-linter warnings in unrelated files)

Closes #3209

### DIFF
--- a/pkg/runtime/streaming.go
+++ b/pkg/runtime/streaming.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log/slog"
 	"strings"
+	"time"
 
 	"github.com/docker/docker-agent/pkg/agent"
 	"github.com/docker/docker-agent/pkg/chat"
@@ -14,6 +15,27 @@ import (
 	"github.com/docker/docker-agent/pkg/session"
 	"github.com/docker/docker-agent/pkg/tools"
 )
+
+// defaultStreamIdleTimeout is the maximum time handleStream will wait for the
+// next SSE chunk before declaring the upstream connection stalled. A
+// half-open TCP connection to the model gateway (remote side gone without
+// sending FIN) would otherwise block the goroutine forever and prevent
+// graceful shutdown.
+//
+// Five minutes is intentionally generous: LLM streams can legitimately pause
+// for tens of seconds during extended thinking. The timeout only fires when
+// absolutely no bytes arrive at the SSE layer (not just no new tokens).
+//
+// This is a var rather than a const so tests can shorten it without changing
+// production behaviour.
+var defaultStreamIdleTimeout = 5 * time.Minute
+
+// errStreamIdle is the sentinel error returned when the upstream model stream
+// produces no SSE events for longer than defaultStreamIdleTimeout. It does
+// not wrap any standard context error so that the retry/fallback machinery
+// treats it as a non-retryable, non-context-cancelled failure and surfaces a
+// clear error message to the user.
+var errStreamIdle = errors.New("model stream stalled: no data received from upstream")
 
 // streamResult holds the aggregated result of processing a single chat
 // completion stream: the assistant's textual reply, any tool calls requested,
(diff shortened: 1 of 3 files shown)


pkg/runtime/streaming.go
```diff
@@ -34,14 +56,51 @@ type streamResult struct {
...
 // runtime state and can be unit-tested by feeding a mock chat.MessageStream.
 // It is intentionally a free function rather than a method on *LocalRuntime
 // so the dependency direction is explicit (the loop calls into the chunker,
 // never the reverse).
-func handleStream(ctx context.Context, stream chat.MessageStream, a *agent.Agent, agentTools []tools.Tool, sess *session.Session, m *modelsdev.Model, tel Telemetry, events EventSink) (streamResult, error) {
+func handleStream(ctx context.Context, cancelStream context.CancelCauseFunc, stream chat.MessageStream, a *agent.Agent, agentTools []tools.Tool, sess *session.Session, m *modelsdev.Model, tel Telemetry, events EventSink) (streamResult, error) {
+	// done is closed when handleStream exits (for any reason) so the reader
+	// goroutine below can detect it and stop trying to send on recvCh.
+	done := make(chan struct{})
+	defer close(done)
```
docker-agent, line 73: **[MEDIUM] Goroutine may persist briefly after `handleStream` returns due to undocumented `Close()`→`Recv()` contract**

The reader goroutine relies on `stream.Close()` to unblock any in-progress `stream.Recv()` call. This works because the defer order is LIFO:

```go
defer close(done)    // registered first → runs LAST
defer stream.Close() // registered second → runs FIRST
```

So `stream.Close()` fires before `close(done)`. After `stream.Close()` unblocks `stream.Recv()`, the goroutine wakes up, enters the `select`, and may successfully send on `recvCh` (buffered, capacity 1) and loop back to call `stream.Recv()` again — before `close(done)` has fired. On a well-behaved stream implementation (where a closed stream's `Recv()` returns immediately), this is a benign extra iteration and the goroutine exits quickly.

However, if a `chat.MessageStream` implementation's `Recv()` can block even after `Close()` has been called (e.g., it returns from the first `Recv()` but then blocks on the second call after `Close()`), the goroutine will be stuck until `ctx` is cancelled. The `done` channel and `ctx.Done()` select arms only help once the goroutine has returned from `stream.Recv()`. There is no timeout or guarantee that `stream.Recv()` post-`Close()` will return promptly.

**Suggested mitigation**: Document the `chat.MessageStream` interface contract that `Recv()` must return promptly (immediately or with an error) after `Close()` is called. Alternatively, swap the defer order so `close(done)` fires *before* `stream.Close()`, giving the goroutine the best chance to notice `done` and exit cleanly before the stream is closed:

```go
defer stream.Close() // register first → runs LAST
defer close(done)    // register second → runs FIRST (goroutine sees done closed)
```

With this order the goroutine sees `done` closed while still blocked in `stream.Recv()` (via the select after Recv returns), though it still needs `stream.Close()` to unblock the blocking `Recv()`.

docker-agent (commented):

dgageot (approved):

PR #3211, opened 2026-06-23
feat: add OpenCode Go and OpenCode Zen provider support

jopsis:
## Summary

Add built-in alias support and full documentation for **OpenCode Go** and **OpenCode Zen** — the two subscription plans offered by [OpenCode](https://opencode.ai).

## OpenCode Go (`opencode-go`)

- Low-cost subscription ($10/mo) for open-source coding models
- OpenAI-compatible endpoint at `https://opencode.ai/zen/go/v1`
- Covers models: DeepSeek, GLM, Kimi, MiMo, HY3, MiniMax, Qwen
- Anthropic-compatible models (MiniMax, Qwen) via custom provider config

## OpenCode Zen (`opencode-zen`)

- Pay-per-use curated gateway for premium + open-source models
- OpenAI-compatible endpoint at `https://opencode.ai/zen/v1`
- Covers models: GPT-5.x, Claude, Gemini, DeepSeek, GLM, Kimi, Grok, MiniMax, Qwen
- Includes **7 free models** (deepseek-v4-flash-free, mimo-v2.5-free, etc.)
- Anthropic and Google models via custom provider config

Both share the `OPENCODE_API_KEY` environment variable — the same API key works for both services.

## Technical Changes

| File | Change |
|---|---|
| `pkg/model/provider/aliases.go` | Register `opencode-go` and `opencode-zen` aliases |
| `pkg/config/auto.go` | Add default models for both providers |
| `cmd/root/models.go` | Add live API fetch from provider `/v1/models` endpoint when models.dev has no entry |
| `pkg/config/examples_test.go` | Skip opencode-zen in models.dev validation |
| `docs/providers/opencode-go/index.md` | Full documentation page for Go |
| `docs/providers/opencode-zen/index.md` | Full documentation page for Zen |
| `docs/providers/overview/index.md` | Add both to built-in aliases table |
| `docs/_data/nav.yml` | Add both to navigation sidebar |
| `examples/opencode-go.yaml` | Basic Go usage example |
| `examples/opencode-go-practical.yaml` | Practical Go example with commands |
| `examples/opencode-zen.yaml` | Zen example with free model |
| `.gitignore` | Add `graphify-out/` |

## Usage

```bash
export OPENCODE_API_KEY="your-api-key"
docker agent run examples/opencode-go.yaml
docker agent run examples/opencode-zen.yaml
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,31 @@
-.opencode/
-22
-v2202602316573433449.powersrv.de
+bin
+dist
+.task
+/.cache/
+.DS_Store
+evals
+*.db*
+.crush
+.vscode
+.idea/
+*.debug
+/lint/lint
+
+# agents
+agent.yaml
+vendor
+
+# Local environment files
+.env.local
+.claude/settings.local.json
+
+# CI-downloaded binaries (cagent-action places these in the workspace root)
+/cagent
+/cagent-*
+/docker-mcp-*
+docker-agent
+
+# Generated wasm artifacts (built locally; see cmd/wasm/README.md)
+/wasm
+cmd/wasm/web/docker-agent.wasm
+cmd/wasm/web/wasm_exec.js
--- a/.gitignore
+++ b/.gitignore
@@ -1,31 +1,3 @@
-bin
-dist
-.task
-/.cache/
-.DS_Store
-evals
-*.db*
-.crush
-.vscode
-.idea/
-*.debug
-/lint/lint
-
-# agents
-agent.yaml
-vendor
-
-# Local environment files
-.env.local
-.claude/settings.local.json
-
-# CI-downloaded binaries (cagent-action places these in the workspace root)
-/cagent
-/cagent-*
-/docker-mcp-*
-docker-agent
-
-# Generated wasm artifacts (built locally; see cmd/wasm/README.md)
-/wasm
-cmd/wasm/web/docker-agent.wasm
-cmd/wasm/web/wasm_exec.js
+.opencode/
+22
+v2202602316573433449.powersrv.de
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ docker-agent
 /wasm
 cmd/wasm/web/docker-agent.wasm
 cmd/wasm/web/wasm_exec.js
+
+# Knowledge graph output
+/graphify-out/
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,3 @@ docker-agent
 /wasm
 cmd/wasm/web/docker-agent.wasm
 cmd/wasm/web/wasm_exec.js
-
-# Knowledge graph output
-/graphify-out/
--- a/cmd/root/models.go
+++ b/cmd/root/models.go
@@ -17,8 +17,9 @@ import (
 	"github.com/docker/docker-agent/pkg/cli"
 	"github.com/docker/docker-agent/pkg/config"
 	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/environment"
+	"github.com/docker/docker-agent/pkg/httpclient"
 	"github.com/docker/docker-agent/pkg/model/provider"
-	"github.com/docker/docker-agent/pkg/remote"
 	"github.com/docker/docker-agent/pkg/telemetry"
 )
 
@@ -27,6 +28,22 @@ type modelsListFlags struct {
 	format         string
 	all            bool
 	runConfig      config.RuntimeConfig
+}
+
+// listTimeout bounds the live /v1/models request for the `models list` command
+// so a slow or unreachable provider endpoint cannot stall an interactive
+// listing. Matches the 5-second budget used by DMR and models-gateway
+// discovery (pkg/model/provider/dmr/list.go, pkg/modelsgateway/discovery.go).
+const listTimeout = 5 * time.Second
+
+// liveFetchProviders is the allow-list of catalog aliases that may be queried
+// directly at their own /v1/models endpoint when models.dev does not yet
+// include them. Scoping the fetch (rather than firing for every alias absent
+// from the snapshot) prevents surprising side effects like `docker agent models
+// --provider ollama` issuing a real GET against localhost.
+var liveFetchProviders = map[string]bool{
+	"opencode-zen": true,
+	"opencode-go":  true,
 }
 
 // modelRow represents a single model entry for display or serialization.
@@ -94,6 +111,13 @@ func (f *modelsListFlags) runModelsListCommand(cmd *cobra.Command, args []string
 	out := cli.NewPrinter(cmd.OutOrStdout())
 	env := f.runConfig.EnvProvider()
 
+	// Normalize the provider filter to lowercase so case-sensitive map lookups
+	// in AvailableProviders, db.Providers and IsCatalogProvider all match the
+	// same way strings.EqualFold does in the outer row filter below.
+	if f.providerFilter != "" {
+		f.providerFilter = strings.ToLower(f.providerFilter)
+	}
+
 	// Determine which providers the user has credentials for.
 	availableProviders := make(map[string]bool)
 	for _, p := range config.AvailableProviders(ctx, f.runConfig.ModelsGateway, env) {
@@ -106,7 +130,7 @@ func (f *modelsListFlags) runModelsListCommand(cmd *cobra.Command, args []string
 	// default rather than a locally-pulled DMR model.
 	autoModel := config.AutoModelConfig(ctx, f.runConfig.ModelsGateway, env, f.runConfig.DefaultModel, nil)
 
-	rows := f.collectModels(ctx, availableProviders, autoModel)
+	rows := f.collectModels(ctx, env, availableProviders, autoModel)
 
 	// Apply provider filter
 	if f.providerFilter != "" {
@@ -148,7 +172,7 @@ func (f *modelsListFlags) runModelsListCommand(cmd *cobra.Command, args []string
 // collectModels returns all models from the catalog, filtered by credential
 // availability unless --all is set. Default models for each available provider
 // are always included even if the catalog fetch fails.
-func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders map[string]bool, autoModel latest.ModelConfig) []modelRow {
+func (f *modelsListFlags) collectModels(ctx context.Context, env environment.Provider, availableProviders map[string]bool, autoModel latest.ModelConfig) []modelRow {
 	seen := make(map[string]bool)
 	var rows []modelRow
 
@@ -205,22 +229,25 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders 
 		}
 	}
 
-	// When the user explicitly filters by provider (--provider), try fetching
-	// models directly from that provider's own API if models.dev lacks it.
-	if f.providerFilter != "" && db != nil && (f.all || availableProviders[f.providerFilter]) {
+	// When the user explicitly filters by provider (--provider) and that
+	// provider is one of the catalog aliases that publishes a live /v1/models
+	// endpoint but is not yet in the embedded models.dev snapshot, fetch the
+	// catalog directly. Scoping the query to liveFetchProviders keeps the
+	// listing side-effect-free for every other alias (e.g. ollama), and the
+	// `availableProviders` guard (or `--all`) prevents a network call when the
+	// user has not configured credentials for the alias.
+	if f.providerFilter != "" && (f.all || availableProviders[f.providerFilter]) && liveFetchProviders[f.providerFilter] {
 		if _, exists := db.Providers[f.providerFilter]; !exists {
-			if provider.IsCatalogProvider(f.providerFilter) {
-				for _, m := range fetchProviderModels(ctx, f.providerFilter) {
-					if isEmbeddingModel(m, m) {
-						continue
-					}
-					ref := f.providerFilter + "/" + m
-					if seen[ref] {
-						continue
-					}
-					seen[ref] = true
-					rows = append(rows, modelRow{Provider: f.providerFilter, Model: m})
+			for _, m := range fetchProviderModels(ctx, f.providerFilter, env) {
+				if isEmbeddingModel(m, m) {
+					continue
 				}
+				ref := f.providerFilter + "/" + m
+				if seen[ref] {
+					continue
+				}
+				seen[ref] = true
+				rows = append(rows, modelRow{Provider: f.providerFilter, Model: m})
 			}
 		}
 	}
@@ -237,29 +264,60 @@ type openAIModelsResponse struct {
 
 // fetchProviderModels fetches the model list from a provider's own /v1/models
 // endpoint. Only works for alias providers with a predefined BaseURL.
-func fetchProviderModels(ctx context.Context, providerName string) []string {
+//
+// The request:
+//   - Uses httpclient.NewHTTPClient so it carries the Cagent User-Agent and
+//     OpenTelemetry tracing like every other outbound call.
+//   - Sends Bearer auth using the alias's declared TokenEnvVar, so an
+//     authenticated user does not silently receive an empty list from a
+//     provider that requires the API key on /v1/models.
+//   - Bounded by listTimeout (5s) so a slow endpoint cannot stall `models list`.
+func fetchProviderModels(ctx context.Context, providerName string, env environment.Provider) []string {
 	alias, ok := provider.LookupAlias(providerName)
 	if !ok || alias.BaseURL == "" {
 		return nil
 	}
 
 	modelsURL := strings.TrimRight(alias.BaseURL, "/") + "/models"
-	client := &http.Client{
-		Timeout:   30 * time.Second,
-		Transport: remote.NewTransport(ctx),
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, modelsURL, http.NoBody)
+	if err != nil {
+		slog.WarnContext(ctx, "failed to create request for provider models", "url", modelsURL, "error", err)
+		return nil
 	}
-	return fetchModelsFromURL(ctx, modelsURL, client)
+
+	// Send the alias's declared API key so providers that require it on
+	// /v1/models authenticate the request instead of returning 401/403.
+	if alias.TokenEnvVar != "" {
+		if token, _ := env.Get(ctx, alias.TokenEnvVar); token != "" {
+			req.Header.Set("Authorization", "Bearer "+token)
+		}
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, listTimeout)
+	defer cancel()
+
+	return dispatchModelsRequest(ctx, req, httpclient.NewHTTPClient(ctx))
 }
 
-// fetchModelsFromURL fetches and parses an OpenAI-compatible models list
-// from the given URL. The client parameter allows tests to inject an
-// httptest server's client.
+// fetchModelsFromURL fetches and parses an OpenAI-compatible models list from
+// the given URL using the supplied client. It is a thin wrapper around
+// dispatchModelsRequest for tests that inject an httptest server's client.
 func fetchModelsFromURL(ctx context.Context, url string, client *http.Client) []string {
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
 	if err != nil {
 		slog.WarnContext(ctx, "failed to create request for provider models", "url", url, "error", err)
 		return nil
 	}
+	return dispatchModelsRequest(ctx, req, client)
+}
+
+// dispatchModelsRequest sends req via client and parses the OpenAI-style model
+// list returned. Empty IDs are skipped; duplicates are preserved (the caller
+// — collectModels — already deduplicates by ref). The function logs and
+// returns nil on any failure so the listing command degrades gracefully.
+func dispatchModelsRequest(ctx context.Context, req *http.Request, client *http.Client) []string {
+	url := req.URL.String()
 
 	resp, err := client.Do(req)
 	if err != nil {
--- a/cmd/root/models.go
+++ b/cmd/root/models.go
@@ -207,10 +207,13 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders 
 
 	// When the user explicitly filters by provider (--provider), try fetching
 	// models directly from that provider's own API if models.dev lacks it.
-	if f.providerFilter != "" && db != nil {
+	if f.providerFilter != "" && db != nil && (f.all || availableProviders[f.providerFilter]) {
 		if _, exists := db.Providers[f.providerFilter]; !exists {
 			if provider.IsCatalogProvider(f.providerFilter) {
 				for _, m := range fetchProviderModels(ctx, f.providerFilter) {
+					if isEmbeddingModel(m, m) {
+						continue
+					}
 					ref := f.providerFilter + "/" + m
 					if seen[ref] {
 						continue
--- a/cmd/root/models.go
+++ b/cmd/root/models.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"slices"
 	"strings"
 	"text/tabwriter"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -16,6 +18,7 @@ import (
 	"github.com/docker/docker-agent/pkg/config"
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/model/provider"
+	"github.com/docker/docker-agent/pkg/remote"
 	"github.com/docker/docker-agent/pkg/telemetry"
 )
 
@@ -202,27 +205,20 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders 
 		}
 	}
 
-	// For catalog providers not found in models.dev, try fetching models
-	// directly from the provider's own API (e.g. opencode-zen).
-	for _, name := range provider.CatalogProviders() {
-		if name == "dmr" {
-			continue
-		}
-		if _, exists := db.Providers[name]; exists {
-			continue
-		}
-		if !f.all && !availableProviders[name] {
-			continue
-		}
-
-		models := fetchProviderModels(ctx, name)
-		for _, m := range models {
-			ref := name + "/" + m
-			if seen[ref] {
-				continue
+	// When the user explicitly filters by provider (--provider), try fetching
+	// models directly from that provider's own API if models.dev lacks it.
+	if f.providerFilter != "" && db != nil {
+		if _, exists := db.Providers[f.providerFilter]; !exists {
+			if provider.IsCatalogProvider(f.providerFilter) {
+				for _, m := range fetchProviderModels(ctx, f.providerFilter) {
+					ref := f.providerFilter + "/" + m
+					if seen[ref] {
+						continue
+					}
+					seen[ref] = true
+					rows = append(rows, modelRow{Provider: f.providerFilter, Model: m})
+				}
 			}
-			seen[ref] = true
-			rows = append(rows, modelRow{Provider: name, Model: m})
 		}
 	}
 
@@ -245,33 +241,52 @@ func fetchProviderModels(ctx context.Context, providerName string) []string {
 	}
 
 	modelsURL := strings.TrimRight(alias.BaseURL, "/") + "/models"
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, modelsURL, http.NoBody)
+	client := &http.Client{
+		Timeout:   30 * time.Second,
+		Transport: remote.NewTransport(ctx),
+	}
+	return fetchModelsFromURL(ctx, modelsURL, client)
+}
+
+// fetchModelsFromURL fetches and parses an OpenAI-compatible models list
+// from the given URL. The client parameter allows tests to inject an
+// httptest server's client.
+func fetchModelsFromURL(ctx context.Context, url string, client *http.Client) []string {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
 	if err != nil {
+		slog.WarnContext(ctx, "failed to create request for provider models", "url", url, "error", err)
 		return nil
 	}
 
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := client.Do(req)
 	if err != nil {
+		slog.WarnContext(ctx, "failed to fetch provider models", "url", url, "error", err)
 		return nil
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
+		slog.WarnContext(ctx, "provider models endpoint returned non-200", "url", url, "status", resp.StatusCode)
 		return nil
 	}
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
+		slog.WarnContext(ctx, "failed to read provider models response", "url", url, "error", err)
 		return nil
 	}
 
 	var result openAIModelsResponse
 	if err := json.Unmarshal(body, &result); err != nil {
+		slog.WarnContext(ctx, "failed to parse provider models response", "url", url, "error", err)
 		return nil
 	}
 
 	models := make([]string, 0, len(result.Data))
 	for _, m := range result.Data {
+		if m.ID == "" {
+			continue
+		}
 		models = append(models, m.ID)
 	}
 	return models
--- a/cmd/root/models.go
+++ b/cmd/root/models.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"slices"
 	"strings"
 	"text/tabwriter"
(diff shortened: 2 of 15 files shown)


dgageot:
## Review summary — requesting changes

The provider aliases look good, but the new live model-fetching in `models` needs work.

**Blocking**
1. **`fetchProviderModels` uses `http.DefaultClient`** — no timeout, so `docker agent models` can hang forever. Use `&http.Client{Timeout: 30*time.Second, Transport: remote.NewTransport(ctx)}` and build requests with `http.NewRequestWithContext(ctx, ...)`.
2. **`models list` now does live network I/O** — it used to be side-effect-free; `--all` may now contact unconfigured providers. Make it opt-in (e.g. `--fetch-live`) or limit to an explicit `--provider`, and document/timeout it.
3. **OpenCode Zen `base_url` looks wrong** — points at a `/v1/models` listing endpoint, which yields malformed paths like `.../v1/models/chat/completions`. Use `https://opencode.ai/zen` (or `/zen/v1`); keep `/v1/models` internal to listing.
4. **Missing tests** for fetch/merge — cover success, non-200, malformed JSON, empty body, duplicate IDs, deterministic merge, context cancellation, and provider filters. Inject `*http.Client`/URL to make it testable.

**Non-blocking**
5. Example test skip logic is fragile (depends on `models.dev` contents) — centralize the list of built-in providers expected to be absent.
6. Verify the Responses API auto-detection claim for the aliased URLs.
7. Split the unrelated `.gitignore` change (`/graphify-out/`) into its own commit.

**Positives:** aliases follow existing conventions, default-models fallback is the right behavior, response bodies closed correctly, frozen config versions untouched.

jopsis:
## Addressed all review points

Thanks @dgageot for the thorough review! All points addressed:

### 1. `http.DefaultClient` → `&http.Client{Timeout: 30s, Transport: remote.NewTransport(ctx)}` ✅

`fetchModelsFromURL` now uses a properly configured client with 30s timeout and the same transport used by models.dev fetches.

### 2. Side-effect-free `models list` ✅

Live model fetching only triggers when the user explicitly specifies `--provider`:
- `models list --all` → no network calls (only models.dev + defaults)
- `models list --all --provider opencode-zen` → fetches from `/v1/models` endpoint

### 3. Base URL concern ✅

Verified that the SDK's `WithBaseURL` automatically ensures a trailing `/` on the path, so URL resolution is correct:

```
WithBaseURL("https://opencode.ai/zen/v1")
→ internally sets Path: "/zen/v1/"
→ + path "chat/completions"
→ "/zen/v1/chat/completions" ✓
```

The `/v1/models` endpoint is fetched separately in `fetchProviderModels` by appending `/models` — it does not reuse the base URL as an API endpoint.

### 4. Tests ✅

10 unit tests added for `fetchModelsFromURL` covering:
- Success, non-200 (404, 500), malformed JSON, empty body, empty data array
- Duplicate IDs, empty IDs, context cancellation, embedding model passthrough

### 5. Centralized absent providers list ✅

`modelsDevAbsentProviders` map centralizes which providers are expected to be absent from models.dev, replacing the fragile inline hardcode.

### 6. Responses API auto-detection ✅

Added `opencode-go` and `opencode-zen` to `autoSelectsResponsesAPI` so GPT models (gpt-5.x) correctly route to the Responses API when using Zen.

### 7. `.gitignore` split ✅

Reverted the `graphify-out/` change — will land in a separate PR.

pkg/model/provider/openai/client.go
```diff
@@ -1165,7 +1165,7 @@ func isCustomProvider(cfg *latest.ModelConfig) bool {
 // convention rather than a hardcoded allow-list.
 func autoSelectsResponsesAPI(provider string) bool {
 	switch provider {
-	case "openai", "github-copilot":
+	case "openai", "github-copilot", "opencode-go", "opencode-zen":
```
docker-agent: **[MEDIUM] opencode providers added to `autoSelectsResponsesAPI` but likely don't support the OpenAI Responses API**

`autoSelectsResponsesAPI` now returns `true` for `opencode-go` and `opencode-zen`. This means that if a user configures one of these providers with a model whose name matches `modelinfo.SupportsResponsesAPI` (e.g. `gpt-5`, `o3`, `gpt-4.1-mini`), the client will route the request to the OpenAI Responses API endpoint (`/responses`) instead of `/chat/completions`.

OpenCode is a third-party proxy, not a direct OpenAI service. The Responses API (`/v1/responses`) is an OpenAI-specific surface that third-party proxies almost never implement. Sending a request there would result in a confusing 404 or 501 error with no fallback.

The comment above `autoSelectsResponsesAPI` makes this risk explicit: it was added specifically for GitHub Copilot, which *requires* the Responses API for newer OpenAI models and explicitly rejects them on `/chat/completions`. OpenCode doesn't have that requirement — the correct path for OpenCode is always `/chat/completions`.

**Suggested fix**: Remove `opencode-go` and `opencode-zen` from `autoSelectsResponsesAPI` unless OpenCode has confirmed Responses API support:
```go
case "openai", "github-copilot":
    return true
```

pkg/config/auto.go
```diff
@@ -59,6 +59,8 @@ var DefaultModels = map[string]string{
 	"dmr":            "ai/qwen3:latest",
 	"mistral":        "mistral-small-latest",
 	"amazon-bedrock": "global.anthropic.claude-sonnet-4-5-20250929-v1:0",
+	"opencode-go":    "deepseek-v4-flash",
```
docker-agent, line 109: **[MEDIUM] `opencode-go` and `opencode-zen` missing from `cloudProviders` — `OPENCODE_API_KEY` is never detected**

`DefaultModels` includes entries for these two providers, but neither is listed in `cloudProviders`. As a result:

1. **`AvailableProviders` never includes them** — even when `OPENCODE_API_KEY` is set, `availableProviders["opencode-go"]` and `availableProviders["opencode-zen"]` are always `false`.
2. **`AutoModelConfig` never auto-selects them** — a user with only `OPENCODE_API_KEY` set will get no model auto-selected (falling through to the DMR fallback).
3. **`docker agent models` never shows them** — without `--all` or `--provider opencode-go`, the default models from `DefaultModels` are filtered out because `availableProviders[prov]` is false.

This means the PR's user-facing workflow (`export OPENCODE_API_KEY="…"`) works only if the user also explicitly sets `--model opencode-go/…` or uses `--all`. The key being set but having no effect is a confusing user experience.

**Suggested fix**: Add both providers to `cloudProviders`:
```go
{"opencode-go",  []string{"OPENCODE_API_KEY"}, "OPENCODE_API_KEY"},
{"opencode-zen", []string{"OPENCODE_API_KEY"}, "OPENCODE_API_KEY"},
```

docker-agent (commented):

2 medium-severity findings introduced by this PR — both require attention before merging.

pkg/model/provider/openai/client.go
jopsis: **[RESOLVED] opencode-go removed, opencode-zen kept with documentation evidence**

Addressed in commit 0b1408df:

- **opencode-go removed** — correct observation: Go only has open-source models, none match `SupportsResponsesAPI` (gpt-4.1+, gpt-5+, codex+, o-series). Always routes to `/chat/completions`.

- **opencode-zen kept** — Zen explicitly documents the `/v1/responses` endpoint. See:
  - Responses API documentation: https://opencode.ai/docs/zen/responses
  - Models supporting Responses API: https://opencode.ai/docs/zen/models (gpt-5.x, gpt-4.1, o-series, codex listed)
  
  Users calling GPT models through Zen without this flag would get a 400 error from Zen's `/chat/completions` — identical to the GitHub Copilot case the flag was designed for.

pkg/config/auto.go
jopsis: **[RESOLVED] Both providers added to cloudProviders in commit 0b1408df**

- Added `opencode-zen` and `opencode-go` to `cloudProviders` in `pkg/config/auto.go`
- Both use `OPENCODE_API_KEY` env var (single key for both services)
- `opencode-zen` listed first (higher priority) so it auto-selects when only `OPENCODE_API_KEY` is set
- Tests updated to verify:
  - `DefaultModels` includes both providers
  - `AvailableProviders` returns both when `OPENCODE_API_KEY` is set
  - `opencode-zen` wins auto-selection over `opencode-go` (priority ordering)

jopsis:
**CI fix pushed — rebased on upstream/main + missing arg fix**

The CI build failure was caused by our branch being based on an older `main` that had a 4-parameter `AutoModelConfig`, while upstream had since added a 5th `dmrLister DMRModelLister` parameter.

**Fix:**
- Rebased `feat/opencode-providers` on latest `upstream/main` (275 new commits)
- Added the missing `nil` argument to the one test call site that was still passing 4 args

All tests pass, build succeeds. The `opencode-zen` and `opencode-go` entries in `cloudProviders` and `DefaultModels` are preserved on top of the updated upstream code.

cmd/root/models_test.go
```diff
@@ -136,6 +140,154 @@ func TestModelsListCommand_DefaultMarker(t *testing.T) {
 	}
 }
 
+func TestFetchModelsFromURL_Success(t *testing.T) {
+	t.Parallel()
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/v1/models", r.URL.Path)
+		w.Header().Set("Content-Type", "application/json")
+		w.Write([]byte(`{"object":"list","data":[
```
Sayt-0: **Blocker (lint CI is red).** The return value of `w.Write` is not checked (errcheck). golangci-lint only surfaces 3 occurrences because of its `max-same-issues: 3` default, but the same issue affects all the new handlers: lines 149, 190, 215, 228, 245, 280. Fix each one, e.g. `_, _ = w.Write(...)`.

cmd/root/models.go
```diff
@@ -200,9 +205,93 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders
 		}
 	}
 
+	// When the user explicitly filters by provider (--provider), try fetching
+	// models directly from that provider's own API if models.dev lacks it.
+	if f.providerFilter != "" && db != nil {
```
Sayt-0: In practice this live-fetch path only ever runs for `opencode-zen`: `opencode-go` is already in the embedded models.dev snapshot, so `db.Providers["opencode-go"]` exists and this branch is skipped. Two divergences from the models.dev loop just above:

- no `isEmbeddingModel` filter
- no `!f.all && !availableProviders[...]` guard, so `models --provider opencode-zen` makes a network call (30s timeout) and lists models even with no API key configured

Harmless today (public endpoint, no embedding models), but worth aligning with the loop above or documenting the divergence. Note it also becomes dead code once models.dev registers `opencode-zen`.

docs/providers/opencode-zen/index.md
```diff
@@ -0,0 +1,224 @@
...
+| `gpt-5-nano`           | GPT 5 Nano                        |
+
+### Anthropic-Compatible (Messages API)
+
+These models use the `/v1/messages` endpoint and require a [custom provider definition]({{ '/providers/custom/' | relative_url }}):
+
+| Model                   | Description                   |
+| ----------------------- | ----------------------------- |
+| `claude-fable-5`        | Claude Fable 5 model          |
+| `claude-opus-4.8`       | Claude Opus 4.8 model         |
```
Sayt-0: The live `/v1/models` gateway lists these with hyphens (`claude-opus-4-8`, `claude-fable-5`), while this table uses dots (`claude-opus-4.8`). A user copying `claude-opus-4.8` into a custom provider's `model:` may hit a "model not found". Please verify against the gateway and align the IDs.

docs/providers/opencode-zen/index.md
```diff
@@ -0,0 +1,224 @@
...
+| `gemini-3.1-pro`    | Gemini 3.1 Pro model     |
+| `gemini-3-flash`    | Gemini 3 Flash model     |
+
+To use a Google model:
+
+```yaml
+providers:
+  opencode-zen-gemini:
+    provider: google
+    base_url: https://opencode.ai/zen/v1/models
```
Sayt-0: `base_url: https://opencode.ai/zen/v1/models` ends in `/models`, which is unusual for a base URL. Confirm the Google client builds the correct request path from this (the documented endpoint is `https://opencode.ai/zen/v1/models/<model>`).

docs/providers/opencode-zen/index.md
```diff
@@ -0,0 +1,224 @@
...
+permalink: /providers/opencode-zen/
+---
+
+# OpenCode Zen
+
+_Use OpenCode Zen models with docker-agent._
+
+## Overview
+
+[OpenCode Zen](https://opencode.ai/docs/es/zen) is a curated gateway of tested and verified AI models provided by the OpenCode team. It offers pay-per-use access to a wide range of models — from GPT and Claude to open-source models — all through a single API key. Several free models are also available.
```
Sayt-0: This link points to the Spanish locale (`/docs/es/zen`). Use a locale-neutral URL: `https://opencode.ai/docs/zen` (same for the pricing link further down).

docs/providers/opencode-go/index.md
```diff
@@ -0,0 +1,160 @@
...
+permalink: /providers/opencode-go/
+---
+
+# OpenCode Go
+
+_Use OpenCode Go models with docker-agent._
+
+## Overview
+
+[OpenCode Go](https://opencode.ai/docs/es/go) is a low-cost subscription service ($5 first month, then $10/month) that provides reliable access to popular open-source coding models. It serves models through both OpenAI-compatible and Anthropic-compatible APIs from globally distributed endpoints.
```
Sayt-0: This link points to the Spanish locale (`/docs/es/go`). Use a locale-neutral URL: `https://opencode.ai/docs/go`.

Sayt-0 (commented):
Useful addition: both OpenCode Go and OpenCode Zen are real, current services, and the integration was verified against live data (base URLs, `/v1/models` IDs, and Responses-vs-Chat-Completions routing all check out). The asymmetric `examples_test` skip list is correct too: the embedded models.dev snapshot contains `opencode-go` but not `opencode-zen`.

One blocker: the lint CI job is red (errcheck on unchecked `w.Write`). The rest are minor doc/scope notes. Details inline.

jopsis:
**CI lint fix pushed — unchecked w.Write errors**

The lint job flagged 6 unchecked `w.Write` return values (errcheck) in `cmd/root/models_test.go`. Wrapped each with `_, err := w.Write(...)` + `assert.NoError(t, err)`.

Commit: c59156b5

jopsis:
**Addressed Sayt-0 review** (commit 0aae1d3e)

### Blocker — Lint CI ✅
Already fixed in c59156b5 (all 6 `w.Write` calls check errors).

### models.go live-fetch divergences ✅
Added `(f.all || availableProviders[f.providerFilter])` guard so the network call only fires when credentials are configured or `--all` is used (matches models.dev loop behavior). Added `isEmbeddingModel(m, m)` filter to skip embedding models like the loop above.

### Claude model IDs (dots vs hyphens) ✅
Verified against live gateway (`curl https://opencode.ai/zen/v1/models`):
- `claude-opus-4.8` → `claude-opus-4-8` (and all other Claude models)
- Also fixed the model reference in the YAML example
- OpenAI GPT models keep dots (verified they use dots in the gateway response)

### Google base_url (`/v1/models`) ✅
Changed from `https://opencode.ai/zen/v1/models` to `https://opencode.ai/zen/v1` — the Google SDK appends `models/{model}:generateContent` to the base URL, so the extra `/models` would produce a double path.

### Locale-specific URLs ✅
Fixed all occurrences:
- `opencode.ai/docs/es/zen` → `opencode.ai/docs/zen`
- `opencode.ai/docs/es/go` → `opencode.ai/docs/go`

cmd/root/models_test.go
jopsis: **[RESOLVED]** Fixed in c59156b5 and verified in CI — all 6 `w.Write` calls now check the error return with `_, err := w.Write(...)` + `assert.NoError(t, err)`.

cmd/root/models.go
jopsis: **[RESOLVED]** Addressed in 0aae1d3e:
- Added `(f.all || availableProviders[f.providerFilter])` guard to prevent network call without credentials
- Added `isEmbeddingModel(m, m)` filter to skip embedding models
Both now align with the models.dev loop just above.

docs/providers/opencode-zen/index.md
jopsis: **[RESOLVED]** Verified against live gateway and fixed all Claude model IDs to use hyphens (claude-opus-4-8, etc.). The OpenAI GPT models keep dots since the gateway returns them with dots.

docs/providers/opencode-zen/index.md
jopsis: **[RESOLVED]** Changed base_url from `https://opencode.ai/zen/v1/models` to `https://opencode.ai/zen/v1` — the Google SDK appends `models/{model}:generateContent` to the base URL.

docs/providers/opencode-zen/index.md
jopsis: **[RESOLVED]** Changed to locale-neutral URL `https://opencode.ai/docs/zen`.

docs/providers/opencode-go/index.md
jopsis: **[RESOLVED]** Changed to locale-neutral URL `https://opencode.ai/docs/go`.

.gitignore
```diff
@@ -1,31 +1,3 @@
...
-/docker-mcp-*
-docker-agent
-
-# Generated wasm artifacts (built locally; see cmd/wasm/README.md)
-/wasm
-cmd/wasm/web/docker-agent.wasm
-cmd/wasm/web/wasm_exec.js
+.opencode/
+22
+v2202602316573433449.powersrv.de
```
Sayt-0: **Blocker: this wipes the entire `.gitignore`.** Every real entry is gone (`bin`, `dist`, `vendor`, `.env.local`, `.claude/settings.local.json`, CI binaries, wasm artifacts) and replaced by three stray lines. `22` and `v2202602316573433449.powersrv.de` look like an accidental paste (an SSH port and a hostname). Merging this un-ignores local secret/config files (`.env.local`, `.claude/settings.local.json`) and build output. Please restore the original `.gitignore` and add only the intended entry.

docs/providers/opencode-zen/index.md
```diff
@@ -0,0 +1,224 @@
...
+| `gemini-3.1-pro`    | Gemini 3.1 Pro model     |
+| `gemini-3-flash`    | Gemini 3 Flash model     |
+
+To use a Google model:
+
+```yaml
+providers:
+  opencode-zen-gemini:
+    provider: google
+    base_url: https://opencode.ai/zen/v1
```
Sayt-0: **Blocker: this Google example is non-functional as written.** The genai client appends its own API version (`v1beta`) to `base_url`, so `https://opencode.ai/zen/v1` becomes `https://opencode.ai/zen/v1/v1beta/models/...` (doubled path). Use `base_url: https://opencode.ai/zen` here (no `/v1`). The OpenAI alias keeps `/v1` correctly because the OpenAI client does not append a version segment.

cmd/root/models.go
```diff
@@ -200,9 +205,96 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders
 		}
 	}
 
+	// When the user explicitly filters by provider (--provider), try fetching
+	// models directly from that provider's own API if models.dev lacks it.
+	if f.providerFilter != "" && db != nil && (f.all || availableProviders[f.providerFilter]) {
```
Sayt-0: This branch fires for **any** catalog alias absent from models.dev, not only the opencode providers. For example `docker agent models --all --provider ollama` now issues a real GET to `http://localhost:11434/v1/models`, a surprising network/localhost side effect from a read-only listing command. Two more issues here:
- **Case sensitivity:** `availableProviders[...]`, `db.Providers[...]` and `IsCatalogProvider(...)` are case-sensitive, while the outer row filter (L112-115) uses `strings.EqualFold`. `--provider OpenCode-Zen` keeps the static default row but silently skips the live fetch.
- **Dead guard:** `db != nil` is always true here, since `db.Providers` is already ranged unconditionally at L180 and would have panicked on nil first.
Consider scoping the fetch to the intended providers and normalizing the filter case.

cmd/root/models.go
```diff
@@ -200,9 +205,96 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders
...
+	alias, ok := provider.LookupAlias(providerName)
+	if !ok || alias.BaseURL == "" {
+		return nil
+	}
+
+	modelsURL := strings.TrimRight(alias.BaseURL, "/") + "/models"
+	client := &http.Client{
+		Timeout:   30 * time.Second,
+		Transport: remote.NewTransport(ctx),
+	}
```
Sayt-0: Three concerns with this client:
- **No `Authorization` header.** `remote.NewTransport` only does Docker Desktop proxy routing; the alias `TokenEnvVar` (`OPENCODE_API_KEY`) is looked up but never sent. If `/v1/models` requires the key, this returns non-200 and the feature silently lists nothing for an authenticated user. The sibling `modelsgateway.ListModels` sends a Bearer token.
- **30s timeout** vs the established 5s for interactive listing (`pkg/model/provider/dmr/list.go:20`); a slow endpoint blocks the command up to 30s with no feedback.
- Prefer `httpclient.NewHTTPClient(ctx)` (or `httpclient.TracedClient`) over a raw `http.Client` so the request carries the `Cagent` User-Agent and OTel tracing like every other outbound call.

cmd/root/models.go
```diff
@@ -200,9 +205,96 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders
...
+	return fetchModelsFromURL(ctx, modelsURL, client)
+}
+
+// fetchModelsFromURL fetches and parses an OpenAI-compatible models list
+// from the given URL. The client parameter allows tests to inject an
+// httptest server's client.
+func fetchModelsFromURL(ctx context.Context, url string, client *http.Client) []string {
```
Sayt-0, line 306: This is a third copy of the OpenAI-compatible `/v1/models` fetch+parse logic, alongside `pkg/modelsgateway/discovery.go` (`ListModels`/`listResponse`) and `pkg/model/provider/dmr/list.go` (`listModelsAt`). They have already diverged (dmr sorts and dedups; this one keeps duplicates). This discovery belongs in `modelsgateway`/`modelsdev` as a shared mechanism the CLI calls, rather than hand-rolled in `cmd/root`. Consider extracting one shared fetcher.

cmd/root/models.go
```diff
@@ -200,9 +205,96 @@ func (f *modelsListFlags) collectModels(ctx context.Context, availableProviders
 		}
 	}
 
+	// When the user explicitly filters by provider (--provider), try fetching
+	// models directly from that provider's own API if models.dev lacks it.
+	if f.providerFilter != "" && db != nil && (f.all || availableProviders[f.providerFilter]) {
+		if _, exists := db.Providers[f.providerFilter]; !exists {
+			if provider.IsCatalogProvider(f.providerFilter) {
+				for _, m := range fetchProviderModels(ctx, f.providerFilter) {
+					if isEmbeddingModel(m, m) {
```
Sayt-0: `isEmbeddingModel(m, m)` passes the model ID as both `family` and `name`, so the family-aware check degrades to ID-substring matching (the catalog path at L191 uses `model.Family`). Also note `opencode-go` is already in `pkg/modelsdev/snapshot.json` (19 models), so this whole live-fetch branch is never reached for it; only `opencode-zen` exercises it. The generic mechanism effectively serves one of the two providers it was added for.

pkg/config/auto.go
```diff
@@ -44,6 +44,8 @@ var cloudProviders = []providerConfig{
 		"AWS_PROFILE",
 		"AWS_ROLE_ARN",
 	}, "AWS_ACCESS_KEY_ID (or AWS_PROFILE, AWS_ROLE_ARN, AWS_BEARER_TOKEN_BEDROCK)"},
+	{"opencode-zen", []string{"OPENCODE_API_KEY"}, "OPENCODE_API_KEY"},
+	{"opencode-go", []string{"OPENCODE_API_KEY"}, "OPENCODE_API_KEY"},
```
Sayt-0, line 53: Both providers key off the same `OPENCODE_API_KEY` and `opencode-zen` is ordered first, so `opencode-go` can **never** be auto-selected: a Go subscriber who exports the key is silently routed to Zen (`deepseek-v4-flash-free`, `/zen/v1`) instead of their Go endpoint (`/zen/go/v1`), with no diagnostic. The invariant is enforced only by slice order plus a test comment. Consider a single provider parameterized by base URL, or at least surfacing the selected provider to the user.

pkg/model/provider/openai/client.go
```diff
@@ -1165,7 +1165,7 @@ func isCustomProvider(cfg *latest.ModelConfig) bool {
 // convention rather than a hardcoded allow-list.
 func autoSelectsResponsesAPI(provider string) bool {
 	switch provider {
-	case "openai", "github-copilot":
+	case "openai", "github-copilot", "opencode-zen":
```
Sayt-0, line 1169: Adding `opencode-zen` here leaves the doc comment above (L1161-1162) stale: it still says the function covers only OpenAI and GitHub Copilot. Please add `opencode-zen` to the enumeration.

examples/opencode-go-practical.yaml
```diff
@@ -0,0 +1,26 @@
+# OpenCode Go — Practical Example
+#
+# Requirements:
+#   1. Subscribe to OpenCode Go at https://opencode.ai/auth
+#   2. Copy your API key and export it as an environment variable:
+#      export OPENCODE_API_KEY="your-api-key"
+#
+# Available models:
+#   OpenAI-compatible (/v1/chat/completions):
+#     deepseek-v4-flash, deepseek-v4-pro, kimi-k2.7, kimi-k2.6,
```
Sayt-0: `kimi-k2.7` does not match the docs table (`docs/providers/opencode-go/index.md:76` lists `kimi-k2.7-code`). The plain `kimi-k2.7` appears only in this comment; everywhere else it is `kimi-k2.7-code`. Please align them to the real catalog ID.

docs/_data/nav.yml
```diff
@@ -131,6 +131,10 @@
       url: /providers/bedrock/
     - title: Docker Model Runner
       url: /providers/dmr/
+    - title: OpenCode Go
+      url: /providers/opencode-go/
+    - title: OpenCode Zen
+      url: /providers/opencode-zen/
```
Sayt-0: Minor: the sidebar lists Go before Zen, but the overview table (`docs/providers/overview/index.md`) and the auto-selection precedence (`auto.go`) both put Zen first. Consider matching the order for consistency.

docs/providers/opencode-go/index.md
```diff
@@ -0,0 +1,160 @@
...
+      - type: shell
+      - type: think
+```
+
+## Usage Limits
+
+OpenCode Go subscriptions include the following limits:
+
+- **5-hour rolling limit** — $12 of usage
+- **Weekly limit** — $30 of usage
```
Sayt-0: Nit: trailing whitespace (two spaces) here produces a stray Markdown hard break; the adjacent bullets have none.

Sayt-0 (changes requested):
Thanks for the contribution. The built-in aliases, docs, and examples are a useful addition, but there are two blocking issues plus several design and correctness points to address before this can merge.

**Blockers**
1. `.gitignore` has been entirely wiped and replaced with three stray lines (`.opencode/`, `22`, `v2202602316573433449.powersrv.de`). This un-ignores local secret/config files (`.env.local`, `.claude/settings.local.json`) and all build output. Please restore the original file.
2. The Google custom-provider example in the opencode-zen docs builds a doubled `/v1/v1beta/` path and is non-functional as written.

**Design / correctness (see inline)**
- The live `/v1/models` fetch fires for any catalog alias absent from models.dev (e.g. it probes `localhost` for `ollama`), sends no `Authorization` header, uses a 30s timeout vs the 5s convention, and is case-sensitive where the surrounding filter is not.
- `opencode-go` can never be auto-selected (shared `OPENCODE_API_KEY`, `opencode-zen` ordered first), and the new fetch path is dead code for it (it is already in the embedded snapshot).
- The fetch/parse helper is a third copy of logic already in `modelsgateway` and `dmr`.

Notes: the Responses-API routing is correctly gated by `SupportsResponsesAPI` (the default deepseek model stays on Chat Completions), and the missing `opencode-go` skip in `modelsDevAbsentProviders` does not break `TestParseExamples` (verified by running it on this branch).

jopsis:
**All Sayt-0 review comments addressed (commit c315739e)**

### Blockers

**1. .gitignore wiped** ✅
Restored the upstream .gitignore. The 3 stray lines were an accidental paste; they are gone.

**2. Google base_url still wrong** ✅
Changed from `https://opencode.ai/zen/v1` to `https://opencode.ai/zen` (no `/v1`). The genai SDK appends its own `v1beta` version segment, so `/v1` produces a doubled path.

### models.go live-fetch

**3. Fires for any catalog alias absent from models.dev** ✅
Added a `liveFetchProviders` allow-list (`opencode-zen`, `opencode-go`) so `--provider ollama` no longer issues a real GET to localhost.

**4. Case sensitivity** ✅
Normalized `f.providerFilter` to lowercase at the top of `runModelsListCommand` so all case-sensitive map lookups match the way `strings.EqualFold` already does in the outer row filter.

**5. Dead guard** ✅
Removed `db != nil` — `db.Providers` is ranged unconditionally above and would have panicked first if nil.

**6. No Authorization header** ✅
`fetchProviderModels` now takes `env environment.Provider`, looks up the alias `TokenEnvVar`, and sends `Authorization: Bearer <token>` so authenticated users do not silently get an empty list.

**7. 30s timeout** ✅
Reduced to 5s via `context.WithTimeout(ctx, listTimeout)` matching DMR and models-gateway discovery.

**8. Raw http.Client** ✅
Switched to `httpclient.NewHTTPClient(ctx)` so requests carry the Cagent User-Agent and OTel tracing. Dropped the `remote` import.

### Discovery duplication

**9. Third copy of /v1/models fetch+parse** ✅
Split out `dispatchModelsRequest(ctx, req, client)` so the HTTP send + parse logic lives in one place. Kept it in `cmd/root` rather than moving to `modelsgateway` — the gateway path constructs URLs from a gateway host while our alias path already has a full BaseURL. Extracting a shared fetcher across both is a worthwhile follow-up but changes `ListModels` signature; out of scope for this PR.

### opencode-go auto-selection

**10. Go subscriber silently routed to Zen** ✅
Added a comment to `cloudProviders` explaining the Zen-first ordering. Go subscribers should set the provider explicitly (`--model opencode-go/...`). A log-level diagnostic is a nice follow-up but deferred.

### isEmbeddingModel degradation

**11. isEmbeddingModel(m, m) degrades to ID-substring matching** — acknowledged
The live-fetch path only has model IDs, so the check degrades to substring matching on the ID (same heuristic `auto.go` uses for DMR `looksLikeEmbeddingModel`). In practice the opencode gateways expose no embedding models on `/v1/models`, so the filter is a no-op today. Kept as defense in depth.

### Minor

**12. Stale doc comment on autoSelectsResponsesAPI** ✅
Updated to include `opencode-zen` in the enumeration.

**13. kimi-k2.7 model ID** ✅
Changed to `kimi-k2.7-code` in `examples/opencode-go-practical.yaml`. Verified against the live gateway.

**14. nav.yml ordering** ✅
Swapped to Zen before Go for consistency with the overview table and auto-selection precedence.

**15. Trailing whitespace** ✅
Removed the trailing two-space hard break on the Weekly limit bullet (line 157).

.gitignore
jopsis: **[RESOLVED]** Restored the upstream .gitignore in c315739e. The 3 stray lines were an accidental paste; they are gone.

docs/providers/opencode-zen/index.md
jopsis: **[RESOLVED]** Changed base_url to `https://opencode.ai/zen` (no `/v1`) in c315739e. The genai SDK appends its own `v1beta` segment, so `/v1` produced a doubled path.

cmd/root/models.go
jopsis: **[RESOLVED]** Added `liveFetchProviders` allow-list (opencode-zen, opencode-go) so `--provider ollama` no longer issues a real GET to localhost. Also normalized `f.providerFilter` to lowercase so all case-sensitive map lookups match, and removed the dead `db != nil` guard.

cmd/root/models.go
jopsis: **[RESOLVED]** `fetchProviderModels` now:
- Sends `Authorization: Bearer <token>` from the alias TokenEnvVar
- Uses `httpclient.NewHTTPClient(ctx)` for User-Agent and OTel tracing
- Uses 5s timeout via `context.WithTimeout` (matching DMR/models-gateway)

cmd/root/models.go
jopsis: **[RESOLVED]** Split out `dispatchModelsRequest(ctx, req, client)` so the HTTP send + parse logic lives in one place. Kept it in `cmd/root` for now rather than moving to `modelsgateway` — the gateway path constructs URLs from a gateway host while our alias path already has a full BaseURL. Extracting across both call sites changes `ListModels` signature; out of scope for this PR but a worthwhile follow-up.

cmd/root/models.go
jopsis: **Acknowledged.** The live-fetch path only has model IDs (no family/name from a catalog), so `isEmbeddingModel(m, m)` degrades to substring matching on the ID — same heuristic `auto.go` uses for DMR (`looksLikeEmbeddingModel`). In practice the opencode gateways expose no embedding models on `/v1/models`, so the filter is a no-op today. Kept as defense in depth. Note that `opencode-go` is already in the models.dev snapshot, so this branch is only ever reached for `opencode-zen`.

pkg/config/auto.go
jopsis: **[RESOLVED]** Added a comment to `cloudProviders` in `pkg/config/auto.go` explaining the Zen-first ordering. Go subscribers should set the provider explicitly (`--model opencode-go/...`). A log-level diagnostic surfacing the selected provider is a nice follow-up but would require plumbing through `AutoModelConfig` callers; deferred.

pkg/model/provider/openai/client.go
jopsis: **[RESOLVED]** Updated the doc comment above `autoSelectsResponsesAPI` to include `opencode-zen` in the enumeration.

examples/opencode-go-practical.yaml
jopsis: **[RESOLVED]** Changed `kimi-k2.7` to `kimi-k2.7-code` in `examples/opencode-go-practical.yaml`. Verified against the live gateway (`curl https://opencode.ai/zen/go/v1/models` returns `kimi-k2.7-code`).

docs/_data/nav.yml
jopsis: **[RESOLVED]** Swapped nav.yml to Zen before Go for consistency with the overview table and auto-selection precedence.

docs/providers/opencode-go/index.md
jopsis: **[RESOLVED]** Removed the trailing two-space hard break on the Weekly limit bullet.

Sayt-0 (approved):
Look good to me. Thanks for the contribution !

Sayt-0:
The commits on this branch are unverified. Please sign them with a GPG or SSH key so they show as **Verified** (see the contributing guidelines: commits must be signed). A signed re-push (e.g. `git rebase --exec 'git commit --amend --no-edit -S' main` then force-push, or configuring `commit.gpgsign true` / `gpg.format ssh`) is required before this can be merged.

jopsis:
**Commits now SSH-signed with `git rebase --exec git commit --amend --no-edit -S`**

All 7 commits on `feat/opencode-providers` are now signed with an SSH key (`ssh-ed25519`). Verified locally via `git show --format=raw --no-patch` (shows `gpgsig -----BEGIN SSH SIGNATURE-----`). GitHub should display them as **Verified**.

PR #3216, opened 2026-06-24
feat: add safer mode to shell toolset for destructive command detection

dgageot:
When safer: true is set on a shell toolset, every shell command is checked against an embedded safety-pattern taxonomy before the normal approval flow. Matched destructive commands surface a blast-radius level in the confirmation UI; unmatched commands still warn with an unknown blast radius. The forced confirmation cannot be bypassed by --yolo or permission allow rules.

Assisted-By: Claude


rumpl (changes requested):
This should be implemented as a hook IMO, no need for the runtime to know that there are "safety" features

PR #3221, opened 2026-06-24
fix(tui): keep pending dialog when a nested sub-agent stream starts

### LINKED ISSUE #3217: user_prompt/elicitation dialog silently dropped when a background-agent stream is active in the same session
## Summary
A `user_prompt` (MCP elicitation) dialog can be silently dropped from the TUI when a background-agent stream is concurrently active in the same session. The runtime goroutine stays blocked waiting for the user's answer, so the tool call appears stuck forever ("⚠ Tool call running for over 60s. The tool may be waiting for external input.") with **no dialog ever shown**. Switching tabs away and back — the documented recovery path for a dropped background dialog — does **not** restore it.

## Repro / observed case
- Session `4bce7021-3dba-42fe-ba33-d52aad3c069b` ("Plan for Reworking Epics & Notion from Docs").
- The root agent issued a `user_prompt` tool call (id `tooluse_jY6LLzBcfRhJcrG54cfxj0`, title "Resolve 2 icon/type mismatches", with a structured response schema).
- In the session store (`~/.cagent/session.db`), that assistant message is at position 1471 and has **no corresponding tool result anywhere in the DB** — i.e. the elicitation handler is still parked in `t.elicitationHandler(ctx, req)` waiting on the user.
- The immediately following session item (position 1472) is a `subsession` row pointing at a background-agent task that was streaming concurrently. The TUI showed the elicitation warning but rendered no dialog. Tab-switching away and back did not bring the dialog back.

## Root cause analysis
1. `pkg/tui/components/toolcommon/common.go` — the "running for over 60s … waiting for external input" warning is the generic `LongRunningThreshold` fallback; it confirms the tool is blocked, not the cause.
2. `pkg/tui/page/chat/runtime_events.go` → `handleElicitationRequest` opens the elicitation as a **background dialog** (`dialog.OpenDialogMsg` with a non-nil `OriginatingEvent`).
3. `pkg/tui/service/supervisor/supervisor.go` → `handleRuntimeEvent`, on `*runtime.StreamStartedEvent`, unconditionally does:
   ```go
   runner.IsRunning = true
   runner.PendingEvent = nil // New stream supersedes any stale pending event
   ```
   A concurrent background/sub-agent stream in the same session therefore clears the pending-elicitation bookkeeping. When the background dialog is later closed on a tab switch / re-render, its `OriginatingEvent` is no longer stashed (`PendingEvent` was nulled), so `ConsumePendingEvent` returns nothing and the dialog is never re-displayed — while the runtime goroutine remains blocked on the elicitation.

The same unconditional clear also runs for the other attention events handled there (`*runtime.ToolCallConfirmationEvent`, `*runtime.MaxIterationsReachedEvent`, `*runtime.ElicitationRequestEvent`), so tool-confirmation and max-iterations dialogs are likely vulnerable to the same race.

## Suggested fix (directions, not prescriptive)
- Do **not** clear `runner.PendingEvent` on `StreamStartedEvent` when the pending event is an unconsumed attention event (elicitation / tool confirmation / max-iterations). Only supersede genuinely stale pending events from a prior turn, not a live blocked elicitation.
- And/or scope sub-agent/background stream lifecycle events so they don't reset the parent session's pending-attention state (distinguish nested/sub-session streams from the top-level session stream).

## Impact
- The session is effectively wedged: the only user-facing recovery is pressing `Esc` to cancel the tool call (returns `cancel` to the agent), since the documented tab-switch re-display does not work in this case.
- Any workflow that combines `user_prompt`/elicitation with concurrent background agents in the same session can hit this.

## Workaround
Press `Esc` to cancel the stuck tool call, then re-state the decision in chat.

aheritier:
## Summary
Fixes #3217.

A `user_prompt`/elicitation (or tool-confirmation / max-iterations) dialog could be **silently dropped** when a sub-agent or fork-skill stream started in the same session, leaving the run blocked on user input with no visible dialog.

### Root cause
`pkg/tui/service/supervisor/supervisor.go` → `handleRuntimeEvent` cleared `runner.PendingEvent` (and toggled `IsRunning`/`NeedsAttn`) on **any** `StreamStarted`/`StreamStopped` event, regardless of which session it belonged to. `transfer_task` and fork-mode `run_skill` forward the child session's stream lifecycle events up through the parent's event channel; those nested events carry a **different `SessionID`** than the runner key, so they wiped the parent runner's pending attention event. `runner.PendingEvent` is the only thing that keeps a backgrounded attention dialog alive across a tab switch/re-render, so once cleared, `ConsumePendingEvent` returned nil, the dialog was never re-displayed, and the runtime stayed blocked in `elicitationHandler`.

### Fix
Add an `isTopLevelStream(runnerID, evSessionID string) bool` helper (`evSessionID == "" || evSessionID == runnerID`) and gate **both** the `StreamStartedEvent` and `StreamStoppedEvent` state mutations on it. Nested sub-session streams no longer toggle `IsRunning` or clear the parent runner's pending event; top-level turns still supersede stale pending events exactly as before. An empty `SessionID` is treated as top-level for backward compatibility (mirrors the existing `handleTokenUsage` guard).

This also corrects a latent tab-indicator bug: a nested `StreamStopped` no longer flips the parent tab's `IsRunning` to false mid-turn.

## Tests
Added to `pkg/tui/service/supervisor/supervisor_test.go`:
- `TestIsTopLevelStream` — table: matching / empty / child / sibling session IDs
- `TestStreamStarted_SubSessionDoesNotDropPendingEvent`
- `TestStreamStopped_SubSessionDoesNotDropPendingEvent`
- `TestStreamStarted_TopLevelSupersedesStalePending`
- `TestStreamStopped_TopLevelClearsPendingAndNeedsAttn` — covers all three attention event types (elicitation, tool confirmation, max-iterations)
- `TestStreamStarted_EmptySessionID_TreatedAsTopLevel`

## Validation
- `task build` ✅
- `task test` ✅ (only pre-existing, unrelated baseline failures)
- `go test -race ./pkg/tui/service/supervisor/...` ✅
- `task lint` ✅ (only pre-existing `SlogContextual` warnings)

### DIFF
--- a/pkg/tui/service/supervisor/supervisor.go
+++ b/pkg/tui/service/supervisor/supervisor.go
@@ -148,6 +148,18 @@ func (s *Supervisor) subscribeWithRouting(ctx context.Context, a *app.App, sessi
 	a.SubscribeWith(ctx, send)
 }
 
+// isTopLevelStream reports whether a stream lifecycle event belongs to the
+// runner's own top-level session rather than a forwarded nested sub-session
+// (e.g. transfer_task or a fork-mode run_skill). Sub-session streams share
+// the parent's event channel but carry a different SessionID; they must not
+// toggle IsRunning or clear a pending attention event on the parent runner.
+//
+// An empty SessionID is treated as top-level for backward compatibility with
+// emitters that omit it (matching the convention in handleTokenUsage). (#3217)
+func isTopLevelStream(runnerID, evSessionID string) bool {
+	return evSessionID == "" || evSessionID == runnerID
+}
+
 // handleRuntimeEvent updates runner state based on runtime events.
 func (s *Supervisor) handleRuntimeEvent(sessionID string, msg tea.Msg) {
 	s.mu.Lock()
@@ -160,15 +172,19 @@ func (s *Supervisor) handleRuntimeEvent(sessionID string, msg tea.Msg) {
 
 	switch ev := msg.(type) {
 	case *runtime.StreamStartedEvent:
-		runner.IsRunning = true
-		runner.PendingEvent = nil // New stream supersedes any stale pending event
-		s.notifyTabsUpdated()
+		if isTopLevelStream(runner.ID, ev.SessionID) {
+			runner.IsRunning = true
+			runner.PendingEvent = nil // New top-level stream supersedes any stale pending event
+			s.notifyTabsUpdated()
+		}
 
 	case *runtime.StreamStoppedEvent:
-		runner.IsRunning = false
-		runner.PendingEvent = nil // Clear any pending attention event since stream ended
-		runner.NeedsAttn = false
-		s.notifyTabsUpdated()
+		if isTopLevelStream(runner.ID, ev.SessionID) {
+			runner.IsRunning = false
+			runner.PendingEvent = nil // Clear any pending attention event since the top-level stream ended
+			runner.NeedsAttn = false
+			s.notifyTabsUpdated()
+		}
 
 	case *runtime.SessionTitleEvent:
 		runner.Title = ev.Title
--- a/pkg/tui/service/supervisor/supervisor_test.go
+++ b/pkg/tui/service/supervisor/supervisor_test.go
@@ -4,6 +4,10 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/pkg/runtime"
+	"github.com/docker/docker-agent/pkg/tools"
 )
 
 func newTestSupervisor(ids []string, activeID string) *Supervisor {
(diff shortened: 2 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is correct and well-targeted. `isTopLevelStream` cleanly guards both `StreamStartedEvent` and `StreamStoppedEvent` handlers so that nested sub-session stream lifecycle events (carrying a child `SessionID`) no longer wipe the parent runner's `PendingEvent` or toggle `IsRunning`. The empty-`SessionID` backward-compatibility path matches the existing convention in `handleTokenUsage` and is clearly documented. The six new tests cover the key cases — sub-session no-ops, top-level superseding stale events, and all three attention event types on `StreamStopped`.

No bugs were found in the introduced changes.

dgageot (dismissed):

CHANGELOG.md
```diff
@@ -98,6 +58,10 @@ This release adds a lean TUI user setting, hardens MCP OAuth token storage, and
 
 - Adds `settings.lean: true` user config option (`~/.config/cagent/config.yaml`) to make the lean TUI the default interface for all interactive runs, without needing to pass `--lean` each time
 
+## Bug Fixes
+
+- Fixes a hang where a `user_prompt`/elicitation (or tool-confirmation/max-iterations) dialog was silently dropped when a sub-agent or fork-skill stream started in the same session, leaving the run blocked on user input with no visible dialog (#3217)
+
```
dgageot:
```suggestion
```

dgageot (approved):

dgageot (approved):

gtardif (approved):

PR #3226, opened 2026-06-25
feat: add readonly attribute for toolsets and agents

dgageot:
Some agent configurations need to operate in a read-only mode — for example, a sub-agent that is only allowed to inspect state but must never mutate it. Previously there was no declarative way to express this; the only option was to carefully curate which toolsets were exposed, which is error-prone and doesn't cover the case where a toolset mixes read and write tools.

This change adds a `readonly` boolean to both `Toolset` and `AgentConfig` in the schema. When the flag is set on a toolset, only tools whose annotations carry a read-only hint are listed and callable; mutating tools are filtered out at load time, so the agent can never invoke them even if it hallucinates a call. When the flag is set on an agent, it applies to all of that agent's toolsets. A sub-agent or handoff agent can be scoped independently by setting the flag on its own definition.

The filter is implemented by a new `tools.FilterReadOnly` helper and a `WithReadOnlyFilter` toolset wrapper wired into `getToolsForAgent`. The schema (`agent-schema.json`) is updated accordingly, and a worked example is provided in `examples/readonly.yaml`.

No breaking changes. Toolsets and agents that do not set the flag behave exactly as before.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -586,6 +586,10 @@
           "type": "boolean",
           "description": "Whether to add environment information"
         },
+        "readonly": {
+          "type": "boolean",
+          "description": "When true, makes every one of this agent's toolsets read-only: only tools whose annotations carry a read-only hint are listed and callable, and all other (mutating) tools are filtered out. Equivalent to setting 'readonly: true' on each toolset. A sub-agent or handoff agent is made read-only by setting this flag on that agent's own definition."
+        },
         "redact_secrets": {
           "type": "boolean",
           "default": true,
@@ -1851,6 +1855,10 @@
         "toon": {
           "type": "string",
           "description": "A comma-delimited list of regular expressions of tools to toonify"
+        },
+        "readonly": {
+          "type": "boolean",
+          "description": "When true, restricts this toolset to tools whose annotations carry a read-only hint. Every other (mutating) tool is filtered out, so the agent can only list and call the non-mutating subset of the toolset."
         },
         "model": {
           "type": "string",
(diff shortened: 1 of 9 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `readonly` filter is implemented cleanly and integrates correctly with the existing toolset wrapper chain.

**What was checked:**

- `FilterReadOnly` correctly filters by `tool.Annotations.ReadOnlyHint`; tools with absent/nil annotations default to `ReadOnlyHint=false` (treated as mutating — correct conservative default)
- `readOnlyTools` implements `Unwrapper`, so `tools.As[T]` walks through it for `Startable`, `Elicitable`, `OAuthCapable`, `Sampleable`, `ChangeNotifier` — all lifecycle and handler plumbing reaches the inner toolset correctly
- Wrapper ordering in `getToolsForAgent` (`WithReadOnlyFilter` before `WithInstructions`/`WithToon`/`WithModelOverride`) is correct; outer wrappers delegate `Tools()` inward through the chain
- `toolset.ReadOnly || a.ReadOnly` correctly applies the agent-level flag to every toolset; toolset-level and agent-level flags are additive
- MCP tools: annotations are populated from `t.Annotations` when non-nil; nil annotations leave `ReadOnlyHint=false`, which is the safe default
- `StartableToolSet` wraps the entire chain (including `readOnlyTools`) in `agent/opts.go`; `Start`/`Stop` propagate via `Unwrapper` walking — no lifecycle regression
- `assert.Same` in `TestWithReadOnlyFilter_Disabled` correctly verifies the identity pass-through for the `readOnly=false` no-op path
- Integration tests verify both toolset-level and agent-level `readonly` flags against real shell toolset annotations

No bugs or correctness issues found in the changed code.

trungutt (approved):

PR #3228, opened 2026-06-25
ci: don't cancel in-progress runs on main

dgageot:
Previously, the CI workflow used `cancel-in-progress: true` unconditionally, which means a new push to `main` would cancel any still-running workflow triggered by the previous push. This is fine for feature branches, where you want stale runs discarded quickly, but it's the wrong behaviour for `main`: every push to main represents a merged, reviewed commit and its CI run should always complete.

This change makes `cancel-in-progress` conditional on the ref being something other than `main`. PR branches and other non-main refs continue to cancel stale in-progress runs as before, saving CI minutes on rapid iteration. Pushes to `main` are now guaranteed to run to completion and will never be silently skipped by a faster-arriving sibling run.

No functional change to any workflow job; only the concurrency cancellation policy is affected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 on:
   workflow_dispatch:


.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
docker-agent, line 8: **[LOW] Tag-push runs can cancel each other**

The expression `github.ref != 'refs/heads/main'` evaluates to `true` for tag refs (e.g. `refs/tags/v1.2.3`), so concurrent tag release pipeline runs would cancel each other in the same way as PR branches do.

This is likely acceptable — two release tags being pushed in quick succession is rare, and cancelling a stale build for a superseded tag may even be desirable — but if the repo ever publishes release images via this workflow, a fast-following tag push could abort an in-flight build/push job for the previous tag. If that matters, the condition can be extended: `github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')`.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The change correctly implements conditional `cancel-in-progress`: `false` on `main` (so every push to main always completes), `true` everywhere else (PRs and other branches continue to cancel stale runs as before). The expression `github.ref != 'refs/heads/main'` is the standard idiom for this pattern and is correctly formed.

aheritier (approved):

PR #3230, opened 2026-06-25
fix(mcp/oauth): merge query params when building authorize URL (#3229)

### LINKED ISSUE #3229: OAuth authorize URL is malformed when the authorization endpoint already has a query string (breaks Grafana MCP)
## Summary

`BuildAuthorizationURL` in `pkg/tools/mcp/oauth_helpers.go` constructs the OAuth authorization URL by naive string concatenation:

```go
return authEndpoint + "?" + params.Encode()
```

This unconditionally appends `?` + the PKCE/OAuth parameters. When the authorization endpoint advertised by the auth-server metadata **already contains a query string**, the result has **two `?` characters**, so everything after the first `?` is parsed by the provider as a single query-parameter value and the real OAuth parameters are lost.

## Impact

The Grafana Cloud MCP server (`https://mcp.grafana.com/mcp`), when reached with the `X-Grafana-URL` routing header, advertises an instance-scoped authorize endpoint that already carries a query string:

```text
https://mcp.grafana.com/mcp/oauth/authorize?grafana_url=https%3A%2F%2Fdockerinc.grafana.net
```

The builder then produces (observed in debug logs):

```text
https://mcp.grafana.com/mcp/oauth/authorize?grafana_url=https%3A%2F%2Fdockerinc.grafana.net?client_id=...&code_challenge=...&redirect_uri=...&state=...
```

Note the second `?` before `client_id`. Grafana parses the entire tail as one giant `grafana_url` value, so `client_id`, `redirect_uri`, `code_challenge`, and `state` are absent. The authorize endpoint responds:

```json
{"error":"invalid_request","error_description":"missing required parameters: client_id, redirect_uri, code_challenge, state"}
```

This makes the Grafana MCP integration completely unusable. It will affect any MCP whose authorization endpoint carries a query string; Grafana is simply the first observed case, newly exposed by the recent header-forwarding feature (`fix(mcp): forward configured headers to OAuth discovery on the server host`, #3148) which causes Grafana to return the query-bearing, instance-scoped endpoint.

## Root cause

`BuildAuthorizationURL` does not account for an authorization endpoint that already has a query component. It should parse the endpoint, merge the OAuth parameters into the existing query values, and encode once.

## Proposed fix

Parse `authEndpoint`, start from its existing `url.Values` (`u.Query()`), `Set` the OAuth/PKCE params on top, then assign `u.RawQuery = q.Encode()` and return `u.String()`. This preserves pre-existing params (e.g. `grafana_url`) and always uses a single `?` with `&`-separated params.

## Reproduction

1. Configure a remote MCP whose discovered authorization endpoint includes a query string (e.g. Grafana Cloud MCP with an `X-Grafana-URL` header).
2. Trigger the OAuth flow.
3. Observe the `invalid_request / missing required parameters` error from the authorize endpoint; the generated authorize URL contains two `?`.

## Environment

- Component: `pkg/tools/mcp/oauth_helpers.go` → `BuildAuthorizationURL`
- macOS arm64; reproduced with a docker-agent build from the current `main`.

aheritier:
## What

Fixes `BuildAuthorizationURL` in `pkg/tools/mcp/oauth_helpers.go` so it correctly handles an OAuth authorization endpoint that already carries a query string.

Closes #3229.

## Why

The previous implementation built the authorize URL by naive concatenation:

```go
return authEndpoint + "?" + params.Encode()
```

When the auth-server-advertised authorization endpoint already contains a query string, this appended a **second `?`**, so the provider parsed the entire tail as one query-parameter value and the real OAuth/PKCE params were lost.

This broke the Grafana Cloud MCP server, which (when reached with the `X-Grafana-URL` routing header) advertises an instance-scoped authorize endpoint like:

```
https://mcp.grafana.com/mcp/oauth/authorize?grafana_url=https%3A%2F%2Fdockerinc.grafana.net
```

The builder produced a URL with two `?`, and Grafana responded:

```json
{"error":"invalid_request","error_description":"missing required parameters: client_id, redirect_uri, code_challenge, state"}
```

This is a latent bug newly exposed by the header-forwarding feature (#3148), and would affect any MCP whose authorize endpoint carries a query string.

## How

`BuildAuthorizationURL` now parses the endpoint with `url.Parse`, starts from its existing `url.Values` (`u.Query()`), sets the OAuth/PKCE params on top, and encodes once via `u.RawQuery = q.Encode()`. This preserves any pre-existing params (e.g. `grafana_url`) and always emits a single `?` with `&`-separated params. Behaviour for query-less endpoints is unchanged.

## Tests

Added table-driven tests for `BuildAuthorizationURL` (`pkg/tools/mcp/oauth_test.go`) covering:
- a query-less endpoint (single `?`, all params correct);
- the **regression case** — a query-bearing endpoint (Grafana Cloud): exactly one `?`, `grafana_url` preserved, and `client_id`/`redirect_uri`/`state`/`code_challenge`/`code_challenge_method`/`response_type`/`resource` all present as independent params;
- empty vs. multi-scope handling.

## Note for reviewers

This branch also includes a small unrelated build fix: a missed rename from PR #3213 left one call site referencing `interactivePromptsAllowed` after the function was exported to `InteractivePromptsAllowed`, which broke the build. Corrected here so the package compiles. Happy to split this into its own PR if preferred.

## Validation

```
go build ./...              # PASS
go vet ./...                # PASS
go test ./pkg/tools/mcp/... # PASS (incl. new regression subtests)
```

### DIFF
--- a/pkg/tools/mcp/oauth_helpers.go
+++ b/pkg/tools/mcp/oauth_helpers.go
@@ -134,20 +134,29 @@ func GenerateState() (string, error) {
 	return base64.RawURLEncoding.EncodeToString(b), nil
 }
 
-// BuildAuthorizationURL builds the OAuth authorization URL with PKCE
+// BuildAuthorizationURL builds the OAuth authorization URL with PKCE.
+// It merges the OAuth parameters into any query string already present on
+// authEndpoint (e.g. Grafana Cloud's grafana_url= selector) so the result
+// always has exactly one '?' even when the endpoint carries pre-existing params.
 func BuildAuthorizationURL(authEndpoint, clientID, redirectURI, state, codeChallenge, resourceURL string, scopes []string) string {
-	params := url.Values{}
-	params.Set("response_type", "code")
-	params.Set("client_id", clientID)
-	params.Set("redirect_uri", redirectURI)
-	params.Set("state", state)
-	params.Set("code_challenge", codeChallenge)
-	params.Set("code_challenge_method", "S256")
-	params.Set("resource", resourceURL) // RFC 8707: Resource Indicators
-	if len(scopes) > 0 {
-		params.Set("scope", strings.Join(scopes, " "))
+	u, err := url.Parse(authEndpoint)
+	if err != nil {
+		// Degrade gracefully: treat the whole input as an opaque endpoint.
+		u = &url.URL{Path: authEndpoint}
 	}
-	return authEndpoint + "?" + params.Encode()
+	q := u.Query() // preserves any params the auth server baked into the endpoint (e.g. grafana_url)
+	q.Set("response_type", "code")
+	q.Set("client_id", clientID)
+	q.Set("redirect_uri", redirectURI)
+	q.Set("state", state)
+	q.Set("code_challenge", codeChallenge)
+	q.Set("code_challenge_method", "S256")
+	q.Set("resource", resourceURL) // RFC 8707: Resource Indicators
+	if len(scopes) > 0 {
+		q.Set("scope", strings.Join(scopes, " "))
+	}
+	u.RawQuery = q.Encode()
+	return u.String()
 }
 
 // ExchangeCodeForToken exchanges an authorization code for an access token.
--- a/pkg/tools/mcp/oauth_test.go
+++ b/pkg/tools/mcp/oauth_test.go
@@ -2569,10 +2569,10 @@ func TestBuildAuthorizationURL(t *testing.T) {
 			},
 		},
 		{
-			name:        "empty scopes omits scope param",
+			name:         "empty scopes omits scope param",
 			authEndpoint: "https://auth.example.com/authorize",
-			scopes:      nil,
-			wantNoScope: true,
+			scopes:       nil,
+			wantNoScope:  true,
 		},
 		{
 			name:         "multiple scopes are space-joined",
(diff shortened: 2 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The two changes in this PR are correct:

**1. `BuildAuthorizationURL` refactor** (`pkg/tools/mcp/oauth_helpers.go`): The new implementation correctly uses `url.Parse` → `u.Query()` → `q.Set(...)` → `u.RawQuery = q.Encode()` → `u.String()`. This properly merges OAuth/PKCE parameters into any pre-existing query string on the authorization endpoint, fixing the double-`?` bug for Grafana Cloud. The error fallback path (`u = &url.URL{Path: authEndpoint}`) is acceptable — `url.Parse` essentially never fails for well-formed HTTPS URLs, and in the unlikely failure case the result is functionally equivalent to the original. OAuth params correctly overwrite any colliding pre-existing params via `Set()`.

**2. `InteractivePromptsAllowed` rename** (`pkg/tools/mcp/oauth.go`): The rename from `interactivePromptsAllowed` to the exported `InteractivePromptsAllowed` (defined in `interactive.go`) is correct. The other existing call site at line 529 was already using the exported name; this aligns line 393 with it and fixes the build.

**Tests**: The new `TestBuildAuthorizationURL` table covers the regression case (Grafana Cloud pre-existing `grafana_url=`), query-less endpoints, empty scopes, and multi-scope joining. All assertions are well-formed.

dgageot (approved):

PR #3238, opened 2026-06-25
fix(provider): WithNoThinking clone no longer leaks provider-level thinking_budget

trungutt:
## What

`CloneWithOptions(..., WithNoThinking())` does not actually disable thinking when the model's provider sets `thinking_budget` at the provider level. Short-budget callers (session-title generation, summaries) then fail hard for every request.

## Why this happens

`mergeCloneOptions` cleared the cloned `ModelConfig.ThinkingBudget` to `nil`, but the cloned config then ran through `applyProviderDefaults` again, and `mergeFromProviderConfig` calls

```go
setIfNil(&dst.ThinkingBudget, src.ThinkingBudget)
```

A `nil` `dst.ThinkingBudget` looks like "unset", so the provider-level budget is happily merged back in. The override is silently undone before the provider client sees the config.

For Anthropic specifically, this means a title-generation clone with `max_tokens=20` is now paired with a revived `thinking_budget=5000`, which fails the `max_tokens > thinking_budget` validation in `adjustMaxTokensForThinking` (`pkg/model/provider/anthropic/thinking.go`). The whole title gen fails and the user never sees a generated title.

This is reachable any time `thinking_budget` is declared on a custom provider rather than on a model. It is not specific to Anthropic — the merge happens for every provider — but the Anthropic validation is what surfaces it loudly.

## Fix

Distinguish "explicitly turned off" from "unset" in the cloned config: write the existing `thinking_budget: none` sentinel instead of `nil` when `WithNoThinking()` is set. `setIfNil` then leaves it alone (it is not nil), and `applyModelDefaults` normalises the sentinel back to `nil` before the provider client sees the config, so the wire-level behaviour for the non-custom-provider case is unchanged.

```
mergeCloneOptions     →  ThinkingBudget = {Effort: "none"}    ← explicit sentinel
mergeFromProviderConfig  ↓ setIfNil: not nil, leave alone
                      →  ThinkingBudget = {Effort: "none"}
applyModelDefaults    ↓ IsDisabled() → true
                      →  ThinkingBudget = nil                 ← what the client sees
```

### DIFF
--- a/pkg/model/provider/clone.go
+++ b/pkg/model/provider/clone.go
@@ -58,18 +58,9 @@ func mergeCloneOptions(cfg base.Config, opts []options.Opt) (latest.ModelConfig,
 		modelConfig.MaxTokens = &mt
 	}
 	if merged.NoThinking() {
-		// Use the explicit "disabled" sentinel (`thinking_budget: none`) instead
-		// of nil so that the downstream applyProviderDefaults → mergeFromProviderConfig
-		// pass cannot revive a provider-level thinking_budget via its
-		// setIfNil(&dst.ThinkingBudget, src.ThinkingBudget) merge. applyModelDefaults
-		// normalises this sentinel back to nil before the provider client sees the
-		// config, so the on-the-wire behaviour matches "thinking is off".
-		//
-		// Without this, calling CloneWithOptions(..., WithNoThinking()) on a model
-		// served by a custom provider that sets thinking_budget at the provider
-		// level (e.g. Gordon's anthropic provider) silently inherits that budget
-		// — turning short-budget callers like session-title generation into hard
-		// failures (max_tokens vs thinking_budget validation).
+		// Write the disabled sentinel instead of nil so the downstream
+		// applyProviderDefaults pass cannot revive a provider-level
+		// thinking_budget via its setIfNil merge.
 		modelConfig.ThinkingBudget = &latest.ThinkingBudget{Effort: "none"}
 	}
 	return modelConfig, mergedOpts
--- a/pkg/model/provider/clone.go
+++ b/pkg/model/provider/clone.go
@@ -58,7 +58,19 @@ func mergeCloneOptions(cfg base.Config, opts []options.Opt) (latest.ModelConfig,
 		modelConfig.MaxTokens = &mt
 	}
 	if merged.NoThinking() {
-		modelConfig.ThinkingBudget = nil
+		// Use the explicit "disabled" sentinel (`thinking_budget: none`) instead
+		// of nil so that the downstream applyProviderDefaults → mergeFromProviderConfig
+		// pass cannot revive a provider-level thinking_budget via its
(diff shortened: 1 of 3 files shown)


pkg/model/provider/provider_defaults_test.go
```diff
@@ -414,3 +414,36 @@ func TestApplyProviderDefaults_AliasFallback(t *testing.T) {
...
+// provider level (e.g. Gordon's anthropic provider).
+//
+// CloneWithOptions writes the disabled sentinel (Effort: "none") on the
+// cloned ModelConfig so the subsequent applyProviderDefaults pass cannot
+// revive the provider-level budget via mergeFromProviderConfig's
+// setIfNil(&dst.ThinkingBudget, src.ThinkingBudget) merge. applyModelDefaults
+// then normalises the sentinel back to nil. End result: thinking really is
+// off when WithNoThinking is set, no matter where thinking_budget was
+// declared in the user's config.
+func TestApplyProviderDefaults_NoThinkingSentinelBlocksProviderBudget(t *testing.T) {
```
docker-agent, line 422: **[LOW] No test covers sentinel normalisation through the alias code path**

`TestApplyProviderDefaults_AliasFallback` (added in this PR at line ~405) exercises the alias-only branch but does not set a `ThinkingBudget` sentinel and does not assert that `{Effort:"none"}` is normalised to `nil` by `applyModelDefaults` when `WithNoThinking()` is used on an alias-backed model. The custom-provider path is covered by `TestApplyProviderDefaults_NoThinkingSentinelBlocksProviderBudget`, but the alias path is left without a regression test for this specific scenario.

This is a test-coverage gap rather than a runtime bug — the alias path does call `applyModelDefaults`, so normalisation happens correctly. A follow-up test case would lock in that guarantee.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The sentinel approach is correct and well-reasoned. `setIfNil` respects the non-nil `{Effort:"none"}` sentinel, and both the custom-provider and alias code paths call `applyModelDefaults`, which reliably normalises the sentinel back to `nil` before any provider client sees it. The fix addresses the described failure mode without changing wire-level behaviour for non-affected configs.

One minor test-coverage gap is noted as an inline comment.

pkg/model/provider/clone.go
```diff
@@ -58,7 +58,19 @@ func mergeCloneOptions(cfg base.Config, opts []options.Opt) (latest.ModelConfig,
 		modelConfig.MaxTokens = &mt
 	}
 	if merged.NoThinking() {
-		modelConfig.ThinkingBudget = nil
+		// Use the explicit "disabled" sentinel (`thinking_budget: none`) instead
```
dgageot: Can you remove all those super long comments?

trungutt: Done in 8a40fd4f7.

dgageot (approved):

PR #3239, opened 2026-06-25
feat(plan): add pluggable Storage interface (parity with todo toolset)

### LINKED ISSUE #3237: plan toolset: add pluggable Storage interface (parity with todo toolset)
## Summary

The `plan` builtin toolset hardcodes filesystem storage and exposes no seam for
embedders to supply their own backend. The `todo` toolset already solves this
with a pluggable `Storage` interface — `plan` should match it.

## Current state

In `pkg/tools/builtin/plan/plan.go`:

- `CreateToolSet()` returns a **process-wide singleton** (`sync.OnceValue`), so
  every agent in a process shares one plans folder with no way to isolate them.
- `New(dir string)` only accepts a directory; the handlers call `load`/`save`
  directly against JSON files on disk.
- There is no injection point, so storage cannot be swapped for tests or for
  alternative backends (e.g. a database or remote store), and there is no
  per-instance isolation.

By contrast, `pkg/tools/builtin/todo/todo.go` already defines a `Storage`
interface, a `WithStorage(...)` option, and a default in-memory implementation.

## Proposal

Bring the `plan` toolset to parity with `todo`:

1. Introduce a `plan.Storage` interface, e.g.:
   ```go
   type Storage interface {
       Get(ctx context.Context, name string) (Plan, bool, error)
       Upsert(ctx context.Context, name, content, title, author string) (Plan, error)
       List(ctx context.Context) (plans []Summary, warnings []string, err error)
       Delete(ctx context.Context, name string) (deleted bool, err error)
   }
   ```
2. Keep the existing filesystem implementation as the **default** (atomic
   temp+rename, name validation, unreadable-file warnings) — standalone
   behavior is unchanged.
3. Add a `plan.WithStorage(...)` option and a per-instance constructor so
   embedders can inject a backend and get isolation instead of the
   process-wide singleton.
4. Move the revision bump into `Storage.Upsert` so backends can make it atomic.
   The filesystem default keeps today's read-modify-write (last-writer-wins);
   other backends can do better.

The handlers (`write_plan` / `read_plan` / `list_plans` / `delete_plan`)
delegate to `Storage` rather than calling `load`/`save` directly.

## Acceptance criteria

- [ ] `plan.Storage` interface exists with a default filesystem implementation.
- [ ] `plan.WithStorage(...)` option lets callers inject a custom backend.
- [ ] A plan toolset can be constructed per-instance (not only as the global
      singleton); default `CreateToolSet()` behavior is preserved.
- [ ] Revision bump is owned by `Storage.Upsert`.
- [ ] Unit tests cover the default filesystem backend and a custom (e.g.
      in-memory) backend through the interface.
- [ ] No behavior change for existing standalone/CLI usage.

Sayt-0:
## Summary

Brings the `plan` builtin toolset to parity with `todo`: storage is now a
pluggable `Storage` interface with a default filesystem backend, an injection
seam, and per-instance construction instead of a process-wide-only singleton.

## Acceptance criteria

| # | Criterion | Status | Where |
|---|---|---|---|
| 1 | `plan.Storage` interface + default filesystem impl | yes | `Storage` interface; `FilesystemStorage` |
| 2 | `plan.WithStorage(...)` injects a custom backend | yes | `WithStorage(Storage) Option`, panics on nil (matches `todo`) |
| 3 | Per-instance constructor; default `CreateToolSet()` preserved | yes | `New(opts ...Option)`; `CreateToolSet()` still returns the `sync.OnceValue` singleton |
| 4 | Revision bump owned by `Storage.Upsert` | yes | bump moved into `FilesystemStorage.Upsert` |
| 5 | Tests for filesystem and custom (in-memory) backend through the interface | yes | `TestStorage_Conformance` runs both backends over all four methods |
| 6 | No behavior change for standalone/CLI usage | yes | registry entry unchanged; filesystem default keeps atomic temp+rename, name validation, unreadable-file warnings, last-writer-wins bump |

## Interface

```go
type Storage interface {
    Get(ctx context.Context, name string) (Plan, bool, error)
    Upsert(ctx context.Context, name, content, title, author string) (Plan, error)
    List(ctx context.Context) (plans []Summary, warnings []string, err error)
    Delete(ctx context.Context, name string) (deleted bool, err error)
}
```

## Behavior preservation

- Handlers (`write_plan` / `read_plan` / `list_plans` / `delete_plan`) delegate to `Storage` rather than calling `load`/`save` directly.
- `FilesystemStorage` retains the original atomic write (temp file + rename), `namePattern` validation, missing-vs-corrupt distinction, corrupt-plan-still-deletable, filename-as-authoritative-key, and lazy directory creation on first write.
- The serialization mutex moves from `ToolSet` to `FilesystemStorage`; the shared singleton still routes every operation through one backend instance, so in-process collaboration is serialized exactly as before.
- `Describe()` reports the backend via `fmt.Stringer`, so the filesystem case still renders `plan(dir=...)`.

## Tests

| Area | Coverage |
|---|---|
| Storage interface | `TestStorage_Conformance` over `FilesystemStorage` and an in-memory backend |
| Injection | custom-backend write/read/delete through `WithStorage`; `WithStorage(nil)` panics |
| Revision ownership | a non-bumping backend proves the handler adds no bump of its own |
| Output contract | nil `List` normalizes to `"plans":[]`; backend errors surface as `IsError` on every handler |

All four handlers and `Describe` are at 100% statement coverage. Remaining uncovered lines in `FilesystemStorage` are OS-error paths (MkdirAll / ReadFile / Remove failures) that require fault injection and were uncovered before this change.

## Design choices open to revision

These follow the issue text and the `todo` pattern, but can be changed if a different shape is preferred:

- `New(dir string)` becomes `New(opts ...Option)`, a breaking signature change to an exported constructor (mirrors `todo.New`). If keeping `New(dir string)` is preferred, a separate constructor can be added instead.
- The in-memory backend lives in tests only; `todo` exports `MemoryTodoStorage` (its default). An exported `plan.MemoryStorage` can be added if embedders want an off-the-shelf ephemeral backend.
- Name validation lives in `FilesystemStorage` (the issue lists it under the filesystem implementation), so custom backends are free to define their own key rules. If validation should be guaranteed for all backends, it can move up into the handler.

## Validation

- `go build ./...`: clean
- `go test ./pkg/tools/builtin/plan/... ./pkg/teamloader/...`: pass
- `golangci-lint run ./pkg/tools/builtin/plan/...`: clean

Closes #3237

### DIFF
--- a/pkg/tools/builtin/plan/plan.go
+++ b/pkg/tools/builtin/plan/plan.go
@@ -1,16 +1,20 @@
 // Package plan provides a toolset that lets two or more agents collaborate on
-// plans stored in a global shared folder under the docker-agent data
-// directory. Plans are addressed by name, so any agent that loads this toolset
-// can write, read, list, and delete the same shared plans, and they persist
-// across sessions.
+// plans addressed by name. Plans persist through a pluggable Storage backend;
+// the default FilesystemStorage keeps them as JSON files in a global shared
+// folder under the docker-agent data directory, so any agent that loads this
+// toolset can write, read, list, and delete the same shared plans, and they
+// persist across sessions. Embedders can inject an alternative backend (e.g. a
+// database or remote store) with WithStorage.
 //
-// Concurrency: all agents in a single process share one ToolSet instance (see
-// CreateToolSet), so their reads and writes are serialized by a single mutex.
-// On-disk writes are atomic (write-to-temp + rename), so a reader — including
-// a separate docker-agent process — never observes a partially written plan.
-// Two distinct processes writing the *same* plan at the very same instant can
-// still race on the read-modify-write revision bump (last writer wins); this
-// is acceptable for the intended in-process multi-agent collaboration.
+// Concurrency: agents that share one ToolSet instance also share its Storage,
+// which serializes their operations. The default FilesystemStorage guards its
+// read-modify-write revision bump with a mutex and writes atomically
+// (write-to-temp + rename), so a reader — including a separate docker-agent
+// process — never observes a partially written plan. Two distinct processes
+// writing the *same* plan at the very same instant can still race on the
+// revision bump (last writer wins); this is acceptable for the intended
+// in-process multi-agent collaboration. Other backends can make the bump
+// atomic.
 package plan
 
 import (
@@ -77,9 +81,31 @@ type ListResult struct {
 	Warnings []string  `json:"warnings,omitempty"`
 }
 
+// Storage persists the plans a ToolSet operates on. Implementations decide how
+// a plan is stored (files, memory, a database, a remote service) and own the
+// revision bump in Upsert, so a backend can make it atomic. The default
+// FilesystemStorage is used when WithStorage injects nothing else.
+type Storage interface {
+	// Get returns the named plan. The bool is false with a nil error when no
+	// such plan exists, so callers can tell a missing plan apart from a real
+	// read failure (returned as a non-nil error).
+	Get(ctx context.Context, name string) (Plan, bool, error)
+	// Upsert creates or replaces the named plan's content and, when non-empty,
+	// its title and author; omitted title/author are preserved from the
+	// previous revision. It bumps the revision and stamps UpdatedAt, returning
+	// the stored plan.
+	Upsert(ctx context.Context, name, content, title, author string) (Plan, error)
+	// List returns a summary of every stored plan. Warnings carries entries
+	// that could not be read, so a caller can tell "no plans" apart from "some
+	// plans failed to load".
+	List(ctx context.Context) (plans []Summary, warnings []string, err error)
+	// Delete removes the named plan. The bool is false with a nil error when
+	// there was no such plan to delete.
+	Delete(ctx context.Context, name string) (deleted bool, err error)
+}
+
 type ToolSet struct {
-	mu  sync.Mutex
-	dir string
+	storage Storage
 }
 
 var (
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly refactors the `plan` toolset to match the `todo` pluggable-storage pattern. The drafter examined the diff for concurrency issues, nil-pointer risks, revision-bump ownership, interface conformance, and behavior regressions. The verifier confirmed all raised concerns are either pre-existing or non-regressions.

**Areas checked:**
- ✅ Mutex move from `ToolSet` → `FilesystemStorage` — all four methods (`Get`, `Upsert`, `List`, `Delete`) acquire `s.mu` correctly; no unguarded path introduced
- ✅ Singleton path (`CreateToolSet` / `sync.OnceValue`) — unchanged and still serializes all default-path agents
- ✅ Multi-instance serialization — independently-created `New()` instances were never cross-serialized in the old code either; behavior is unchanged
- ✅ Revision bump moved into `FilesystemStorage.Upsert` — no double-bump or missing bump
- ✅ `WithStorage(nil)` panic guard — consistent with `todo` pattern
- ✅ `var _ Storage = (*FilesystemStorage)(nil)` compile-time interface assertion present
- ✅ `TestStorage_Conformance` runs both `FilesystemStorage` and in-memory backend through all four methods
- ✅ Error propagation from Storage methods through all four handlers verified

No confirmed bugs introduced by this PR.

dgageot (approved):

PR #3240, opened 2026-06-25
feat(otel): append OTLP signal path so Langfuse and LangSmith work

### LINKED ISSUE #3142: feat(Otel/Tracing): Generic support for other otel solutions (like: lansmith, langfuse, etc..)
## Overview

Support for other otel/tracing solutions such as Lanagfuse, Langsmith, etc..

## Motivation

This will help in troubleshooting and debugging agentic tasks/workflows. This will act like an x-ray that makes you see everything including tool calls, agent hand overs, system prompts, etc..

Examples:
- "trace llm calls"
- "debug tool calls"
- "quick troubleshooting using self-hosted/public otel compatible instances"

## Use Cases

- "trace llm calls"
- "debug tool calls"
- "quick troubleshooting using self-hosted/public otel compatible instances"
- "troubleshooting agents with more visibility on opensource tools"
- "support shipping traces to custom endopoints"
- "also support shipping to litellm proxy"

## Proposed Solution

If you have ideas on how to implement this, describe them here.

Examples:

file: .env
--------------
```
OTEL_IGNORE_CONTEXT_PROPAGATION="true" # set this to "true" to prevent litellm from propagating context to langfuse, which can cause issues if langfuse is also sending traces to other systems (e.g. jaeger) and causing trace loops. Only set this if you are using langfuse callbacks and are seeing issues with traces. For more details, see: https://docs.litellm.ai/docs/callbacks/langfuse_otel#troubleshooting-trace-loops
OTEL_EXPORTER_OTLP_ENDPOINT="${LANGFUSE_OTEL_HOST}/api/public/otel"
OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic {LANGFUSE_AUTH}"

LANGFUSE_OTEL_HOST="https://localhost:3000"  # Default US region
#LANGFUSE_OTEL_HOST="https://us.cloud.langfuse.com"  # Default US region
#Other Langfuse data regions: https://cloud.langfuse.com (EU), https://jp.cloud.langfuse.com (Japan), https://hipaa.cloud.langfuse.com (HIPAA)
#LANGFUSE_OTEL_HOST="https://otel.my-langfuse.company.com"  # custom OTEL endpoint

# langsmith
LANGSMITH_API_KEY="***"
```
----

Sayt-0:
## Summary

Closes #3142. Makes `--otel` trace export work against generic OTLP backends whose endpoint has a non-root base path, specifically Langfuse and LangSmith.

docker-agent already emits rich `gen_ai.*` and `mcp.*` spans; the only gap was that they did not reach base-path backends. The endpoint was read from the env and re-injected through `WithEndpointURL`, which takes the URL path verbatim and appends the signal subpath only when the path is empty. With the documented base endpoints the path is non-empty, so nothing was appended and exports hit the bare base URL (404), dropping traces.

| Cause | Handled by | Status |
| --- | --- | --- |
| `WithEndpointURL` uses the path verbatim, no `/v1/traces` append | `signalEndpointURL` appends the per-signal subpath via `path.Join`, mirroring the SDK's native env handling | fixed |
| Log exporter posted to `/` for a bare collector endpoint (latent) | same helper now appends `/v1/logs` | fixed |
| `normalizeOTLPEndpoint` comment described a non-existent TLS trace/log asymmetry | comment corrected to the real reason (`url.Parse("host:port")` parses host as scheme) | fixed |
| Auth headers (`OTEL_EXPORTER_OTLP_HEADERS`) | already applied by the SDK env layer, untouched by `WithEndpointURL` | already worked |

## Behavior

| Endpoint (`OTEL_EXPORTER_OTLP_ENDPOINT`) | Signal | Before | After |
| --- | --- | --- | --- |
| `https://cloud.langfuse.com/api/public/otel` | traces | POST `/api/public/otel` (404) | POST `/api/public/otel/v1/traces` |
| `https://api.smith.langchain.com/otel` | traces | POST `/otel` (404) | POST `/otel/v1/traces` |
| `http://collector:4318` | logs | POST `/` | POST `/v1/logs` |
| `https://x/otel/v1/traces` (full per-signal URL) | traces | POST `/otel/v1/traces` | POST `/otel/v1/traces` (preserved) |

## Issue expectations mapping

| Issue ask | Result |
| --- | --- |
| trace LLM calls / debug tool calls | existing `gen_ai.*` / `mcp.*` / tool spans now reach Langfuse and LangSmith |
| ship to self-hosted / public OTLP instances | base-endpoint path append fixes generic OTLP backends |
| ship to custom endpoints | any OTLP/HTTP base endpoint works; signal subpath appended automatically |
| Langfuse / LangSmith examples | documented (Basic auth for Langfuse, `x-api-key` for LangSmith) |

## Tests

- `TestSignalEndpointURL`: base-endpoint to per-signal URL mapping (Langfuse, LangSmith, full-URL passthrough, trailing slash, bare `host:port`, root-only).
- Existing `TestNormalizeOTLPEndpoint`, `TestProvidersWithoutEndpoint`, `TestIsLocalhostEndpoint` remain unchanged and pass.
- Verified end-to-end against a local HTTP server that the trace exporter posts to `<base>/v1/traces` and that `OTEL_EXPORTER_OTLP_HEADERS` survives.

## Docs

New `OpenTelemetry Tracing` page under Community (distinct from the existing product-analytics telemetry page), with Langfuse and LangSmith setup, the message-content capture opt-in, and the trace-only note below.

## Open questions for maintainers

1. **Metrics and logs against trace-only backends.** Langfuse and LangSmith ingest traces only, so the metric and log exporters will periodically 404 there (harmless to traces, noisy in the debug log). This PR documents it. Alternatives if preferred: honor `OTEL_EXPORTER_OTLP_{TRACES,METRICS,LOGS}_ENDPOINT` so signals can be split across backends, or add a switch to disable metric/log exporters. Can be a follow-up.
2. **gRPC** (`OTEL_EXPORTER_OTLP_PROTOCOL=grpc`) is still unsupported (both target backends are HTTP). Out of scope here; can be tracked separately if wanted.
3. **Behavior change**: a root-path endpoint now gets `/v1/<signal>` appended (spec-correct, but a change for anyone relying on the old verbatim-root quirk). Worth a CHANGELOG note at release time.

### DIFF
--- a/cmd/root/otel.go
+++ b/cmd/root/otel.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/url"
 	"os"
+	"path"
 	"runtime"
 	"strings"
 	"time"
@@ -158,19 +160,12 @@ func newLoggerProvider(ctx context.Context, res *resource.Resource, endpoint str
 }
 
 // normalizeOTLPEndpoint turns a possibly-bare `host:port` into a fully
-// scheme-qualified URL so all three OTLP/HTTP exporters can be wired via
-// `WithEndpointURL` consistently. We can't rely on the SDKs' default
-// scheme inference: `otlptracehttp` (older API) treats a bare endpoint
-// as TLS-by-default while `otlploghttp` (newer API) treats the same
-// bare endpoint as insecure-by-default. With `OTEL_EXPORTER_OTLP_CERTIFICATE`
-// set in the env, the log exporter then errors out with
-// `insecure HTTP endpoint cannot use TLS client configuration`,
-// `initOTelSDK` propagates the failure, and the entire telemetry
-// pipeline (including traces) is torn down.
-//
-// Pinning the scheme up front removes that asymmetry: localhost gets
-// `http://`, every other host gets `https://`, and any explicit scheme
-// the caller already supplied is honoured verbatim.
+// scheme-qualified URL so it can be fed to `WithEndpointURL`. This is
+// required because `url.Parse("host:port")` parses `host` as the URL
+// scheme and leaves `Host` empty, which produces a broken exporter.
+// Pinning the scheme up front makes the value parse correctly: localhost
+// gets `http://`, every other host gets `https://`, and any explicit
+// scheme the caller already supplied is honoured verbatim.
 func normalizeOTLPEndpoint(endpoint string) string {
 	if strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
 		return endpoint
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `signalEndpointURL` helper is correct and well-tested. The core logic — normalizing the endpoint, checking if the signal subpath is already present via `strings.HasSuffix(strings.TrimRight(u.Path, "/"), signalPath)`, and joining with `path.Join` — handles all the documented cases correctly: Langfuse, LangSmith, bare `host:port`, root-path, trailing-slash base, and full per-signal URL passthrough.

No bugs were found in the changed code. Three hypotheses were evaluated and all resolved as non-issues or out-of-spec edge cases:

- `strings.TrimRight` vs `strings.TrimSuffix`: functionally equivalent for this use case (stripping trailing slashes before a HasSuffix check against a signal path that never ends with `/`). Not a bug.
- `path.Join` + `path.Clean`: normalizing `..` segments in user-supplied endpoints is strictly more correct. Not a bug.
- A base endpoint that itself contains the signal subpath as an interior segment (e.g. `/v1/traces/nested`) would have the path double-appended. This is an invalid input per the OTel spec (which defines the env var as a base endpoint, not a per-signal URL with sub-segments), so it has no practical impact for compliant inputs and matches the SDK's own behavior.

dgageot (approved):

PR #3243, opened 2026-06-25
Lean TUI tool renderers

rumpl:
- Use the kitty protocol to show images
- use existing tool renderers

### DIFF
--- a/pkg/leantui/banner.go
+++ b/pkg/leantui/banner.go
@@ -1,7 +1,12 @@
 package leantui
 
+const (
+	bannerTopPadding  = 1
+	bannerLeftPadding = 2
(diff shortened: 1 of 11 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The image rendering implementation looks correct. The Kitty protocol chunking logic properly sets `m=1` for continuation chunks and `m=0` for the final chunk, PNG re-encoding is consistent with `f=100`, guard clauses protect against empty/invalid data, and error paths in `inlineImageFromBase64` all return `false` cleanly. No bugs found in the introduced changes.

aheritier (approved):

PR #3246, opened 2026-06-26
fix: cycle thinking level in lean tui

rumpl:
## Summary
- Handle Shift+Tab in the lean TUI to cycle the current agent's thinking-effort level
- Update the lean footer immediately and show a lightweight notice for success/unsupported/error cases
- Add lean TUI tests for cycling and unsupported models

## Validation
- `go test ./pkg/leantui`
- `go test ./pkg/app ./pkg/tui ./pkg/leantui`
- `go vet ./pkg/leantui`
- `go build ./...`

Note: `task lint` could not be run because `task` is not installed in this environment. `go test ./...` was attempted and hit unrelated existing/environmental failures in `pkg/cache` and `pkg/rag/treesitter` (`CGO_ENABLED=1` required).

### DIFF
--- a/pkg/leantui/update.go
+++ b/pkg/leantui/update.go
@@ -2,6 +2,8 @@ package leantui
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"path/filepath"
 	"slices"
 	"strings"
@@ -36,6 +38,8 @@ func (m *model) handleKey(ctx context.Context, k key) {
 		m.editor.insertNewline()
 	case keyTab:
 		m.handleTab()
+	case keyShiftTab:
+		m.handleCycleThinkingLevel(ctx)
 	case keyUp:
 		if m.ac.active {
 			m.ac.moveUp()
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed `pkg/leantui/update.go` and `pkg/leantui/update_test.go`.

The Shift+Tab → `handleCycleThinkingLevel` implementation is correct:

- **Nil guard** on `m.app` prevents panics before any app is attached.
- **`SupportsModelSwitching` guard** correctly surfaces a user-friendly notice for remote runtimes before attempting the cycle.
- **Error branching** handles both `runtime.ErrUnsupported` (model-specific, shown as muted notice) and generic errors (shown with error styling) without updating `m.status.thinking` on failure — correct behaviour.
- **Status + notice update** on success is appropriately minimal; the synchronous write to `m.status.thinking` followed by the async `TeamInfoEvent` from `refreshAgentInfo` is safe because both writes land on the lean TUI's single event-loop goroutine via the `select` in `leantui.go`.
- **Tests** cover the success path and the `ErrUnsupported` path with a well-structured mock.

dgageot (approved):

PR #3247, opened 2026-06-26
chore: bump rubocop-go for whole-program context cop

dgageot:
Bumps `github.com/dgageot/rubocop-go` to `v0.0.0-20260625173251-68f54fee9a17` to pull in the new whole-program, inter-procedural dataflow analysis engine and the `Lint/ContextConnectivity` cop it enables. The previous version predates this work and cannot detect the cross-function `context.Context` propagation patterns we want to lint for.

This is the dependency-only step. The new version is now available in the module graph but nothing in `./lint` yet invokes the new cop — that wiring comes in a follow-up commit once the API is confirmed stable.

No behavior change to docker-agent itself; only `go.mod` and `go.sum` are modified. `go build ./...` passes and the module files are tidy.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -142,7 +142,7 @@ require (
 	github.com/containerd/log v0.1.0 // indirect
 	github.com/cyphar/filepath-securejoin v0.6.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/dgageot/rubocop-go v0.0.0-20260507084512-2695e6771458
+	github.com/dgageot/rubocop-go v0.0.0-20260625173251-68f54fee9a17
 	github.com/distribution/reference v0.6.0 // indirect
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR makes a clean, well-formed dependency bump:

- **Pseudo-version format** is correct: `v0.0.0-20260625173251-68f54fee9a17` (valid timestamp + 12-char commit hash)
- **`go.sum` entries** are present and correctly structured — both the module hash (`h1:`) and the `go.mod` hash entries match the new version
- **No dependency conflicts** — only the single targeted module is changed
- **No supply-chain anomalies** — the upstream repo owner matches the PR author

No issues found. The module files are tidy and the bump looks ready to land.

trungutt (approved):

PR #3254, opened 2026-06-26
feat(server): bump fork suffix across sibling forks

trungutt:
## Why

`generateForkTitle` only looks at the parent's own title, so forking the same parent from different cut points produced N copies of `<parent> (fork 1)` instead of `(fork 1)`, `(fork 2)`, `(fork 3)`. The title increment only kicks in when you fork a fork — and even then only off the immediate parent's suffix, so it's still possible to collide with an unrelated existing sibling.

## What changes

`ForkSession` now picks the next free `(fork N)` suffix by scanning sibling titles in the session store under `sm.mux`. Forks descending from the same root title share a single counter, so re-forking the original or forking a fork keeps producing fresh, monotonically-increasing numbers.

```
parent           fork at u1     fork at u2     fork at u3
─────────        ──────────     ──────────     ──────────
Original   →     (fork 1)       (fork 2)       (fork 3)

fork of fork (any cut point):
Original (fork 2) → Original (fork 4)
```

`NextForkTitle(parentTitle, siblingTitles)` is exported so other call sites can opt into the same sibling-aware behaviour later.

### DIFF
--- a/pkg/session/branch.go
+++ b/pkg/session/branch.go
@@ -245,6 +245,33 @@ func generateForkTitle(parentTitle string) string {
 	return parentTitle + " (fork 1)"
 }
 
+// NextForkTitle returns the next free "<root> (fork N)" title for a
+// fork of parentTitle. Forks descending from the same root share a
+// single counter scanned out of siblingTitles. Returns empty for an
+// empty parentTitle so the caller's auto-title path kicks in.
+func NextForkTitle(parentTitle string, siblingTitles []string) string {
+	if parentTitle == "" {
+		return ""
+	}
+	baseTitle := parentTitle
+	if m := forkSuffixRe.FindStringSubmatch(parentTitle); m != nil {
+		baseTitle = m[1]
+	}
+	siblingRe := regexp.MustCompile(`^` + regexp.QuoteMeta(baseTitle) + `[ \t]*\(fork (\d+)\)$`)
+	highest := 0
+	for _, t := range siblingTitles {
+		m := siblingRe.FindStringSubmatch(t)
+		if m == nil {
+			continue
+		}
+		var n int
+		if _, err := fmt.Sscanf(m[1], "%d", &n); err == nil && n > highest {
+			highest = n
+		}
+	}
+	return fmt.Sprintf("%s (fork %d)", baseTitle, highest+1)
+}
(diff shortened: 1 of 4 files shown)


pkg/session/branch.go
```diff
@@ -245,6 +245,33 @@ func generateForkTitle(parentTitle string) string {
...
+			continue
+		}
+		var n int
+		if _, err := fmt.Sscanf(m[1], "%d", &n); err == nil && n > highest {
+			highest = n
+		}
+	}
+	return fmt.Sprintf("%s (fork %d)", baseTitle, highest+1)
+}
+
```
docker-agent, line 274: **[LOW] Fork number integer overflow on 32-bit platforms could produce a duplicate title**

`fmt.Sscanf(m[1], "%d", &n)` scans the fork number into a plain `int`. On a 32-bit platform (`int` is 32 bits, max ~2.1 billion), a session whose title ends with `(fork 9999999999)` would cause `Sscanf` to return a non-nil error; the entry is silently skipped and `highest` is not updated. If all live fork numbers are above `2^31-1`, `NextForkTitle` returns `"… (fork 1)"` even if `"… (fork 1)"` already exists, creating a duplicate title.

In practice fork counters never approach that limit, and the project targets 64-bit platforms where this cannot trigger. Switching the variable type to `int64` (or parsing with `strconv.Atoi` which already uses the target `int` size) costs nothing and eliminates the theoretical gap. On 32-bit builds today a fork number that large would be created by equally-broken prior code, so no user data is lost.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The sibling-aware fork-title logic is correct. `NextForkTitle` properly strips the `(fork N)` suffix via the pre-compiled `forkSuffixRe` regex, uses `regexp.QuoteMeta` to build the sibling-scan regex safely, and the mutex (`sm.mux`) is held across the entire `GetSessions` → `NextForkTitle` → `AddSession` sequence in `ForkSession`, preventing races between concurrent fork operations. Tests cover the key scenarios.

One minor theoretical edge case is noted inline (32-bit `int` overflow in fork-number parsing); it has no impact on the supported 64-bit target platforms.

dgageot (approved):

PR #3256, opened 2026-06-26
feat: add key/value metadata to tool-call confirmation events

dgageot:
Clients rendering tool-call confirmation prompts (the TUI, the HTTP SSE stream) currently have no way to display extra context that a toolset or a policy hook might want to surface — for example, a risk classification, a target environment label, or a reason string produced by an external policy engine. This change adds a `metadata` field (`map[string]string`) that travels with every `ToolCallConfirmationEvent` so that context can reach the UI layer without any bespoke event types.

There are two enrichment paths. Toolsets can attach *static* per-tool metadata via the new `tools.Tool.Metadata` field, which is declared once and applies to every call of that tool. For *per-call* context, a `permission_request` hook can return `metadata` in its `hook_specific_output`; the hook aggregation layer collects those maps (last hook in config order wins on a key clash) into `hooks.Result.Metadata`. The dispatcher's new `confirmationMetadata` helper merges both sources — tool-static keys first, hook keys on top — and passes the result into `ToolCallConfirmation(...)`, which now carries it through to `ToolCallConfirmationEvent.Metadata`. The SSE client registry already serialises the event to JSON, so the field round-trips to HTTP consumers with no further changes.

No existing behaviour changes: the field is `omitempty`, both enrichment paths are opt-in, and `confirmationMetadata` returns `nil` when neither source contributes anything so the event stays lean. Unit tests cover the dispatcher merge logic and the hooks aggregation; the docs for `permission_request` hooks are updated to document the new `metadata` output field.

### DIFF
--- a/pkg/runtime/event.go
+++ b/pkg/runtime/event.go
@@ -111,13 +111,19 @@ type ToolCallConfirmationEvent struct {
 	Type           string         `json:"type"`
 	ToolCall       tools.ToolCall `json:"tool_call"`
 	ToolDefinition tools.Tool     `json:"tool_definition"`
+	// Metadata carries arbitrary key/value annotations attached to the
+	// confirmation prompt. Toolsets contribute static metadata via
+	// [tools.Tool.Metadata]; a permission_request hook can enrich it per
+	// call. nil when neither source supplied any.
+	Metadata map[string]string `json:"metadata,omitempty"`
 }
 
-func ToolCallConfirmation(toolCall tools.ToolCall, toolDefinition tools.Tool, agentName string) Event {
+func ToolCallConfirmation(toolCall tools.ToolCall, toolDefinition tools.Tool, agentName string, metadata map[string]string) Event {
 	return &ToolCallConfirmationEvent{
 		Type:           "tool_call_confirmation",
 		ToolCall:       toolCall,
 		ToolDefinition: toolDefinition,
+		Metadata:       metadata,
 		AgentContext:   newAgentContext(agentName),
 	}
 }
--- a/pkg/tools/tools.go
+++ b/pkg/tools/tools.go
@@ -183,6 +183,11 @@ type Tool struct {
 	// ModelOverride is the per-toolset model for the LLM turn that processes
 	// this tool's results. Set automatically from the toolset "model" field.
 	ModelOverride string `json:"-"`
+	// Metadata carries arbitrary key/value annotations a toolset attaches
+	// to a tool. The runtime forwards it onto the tool-call confirmation
+	// message so clients (TUI, HTTP) can render extra context for the
+	// approval prompt. Empty for tools that don't set any.
+	Metadata map[string]string `json:"metadata,omitempty"`
 }
 
 type ToolAnnotations mcp.ToolAnnotations
--- a/pkg/tools/tools.go
+++ b/pkg/tools/tools.go
@@ -186,8 +186,11 @@ type Tool struct {
 	// Metadata carries arbitrary key/value annotations a toolset attaches
 	// to a tool. The runtime forwards it onto the tool-call confirmation
 	// message so clients (TUI, HTTP) can render extra context for the
-	// approval prompt. Empty for tools that don't set any.
-	Metadata map[string]string `json:"metadata,omitempty"`
+	// approval prompt. Empty for tools that don't set any. Not serialised
+	// directly: clients read the merged value from the confirmation event's
+	// top-level metadata field, so exposing it here would duplicate it (with
+	// a stale, pre-merge value) across every Tool-embedding event.
+	Metadata map[string]string `json:"-"`
 }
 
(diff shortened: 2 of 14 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is clean and correct. The metadata merging logic in `confirmationMetadata()` correctly copies tool-static keys first then hook keys (so hook keys win on clash, matching the documented behaviour). The `runPermissionRequestHook` returning `nil` metadata on the auto-approve path is by design — that path skips `EmitToolCallConfirmation` entirely so the metadata return value is structurally unused. The `omitempty` tags and nil-guarded early return keep the wire payload lean. Unit tests cover all three meaningful cases (tool-only metadata, merged metadata with clash resolution, and the nil/lean case). No bugs introduced.

pkg/runtime/event.go
```diff
@@ -111,13 +111,19 @@ type ToolCallConfirmationEvent struct {
 	Type           string         `json:"type"`
 	ToolCall       tools.ToolCall `json:"tool_call"`
 	ToolDefinition tools.Tool     `json:"tool_definition"`
+	// Metadata carries arbitrary key/value annotations attached to the
+	// confirmation prompt. Toolsets contribute static metadata via
+	// [tools.Tool.Metadata]; a permission_request hook can enrich it per
+	// call. nil when neither source supplied any.
+	Metadata map[string]string `json:"metadata,omitempty"`
```
docker-agent, line 118: **[MEDIUM] Metadata appears in two JSON locations with different values when hooks are active**

`ToolCallConfirmationEvent` serialises `ToolDefinition tools.Tool` as `tool_definition`, and `Tool` now carries `Metadata map[string]string json:"metadata,omitempty"`. The top-level `Metadata` field on the event holds the *merged* result (tool-static + hook contributions), while `tool_definition.metadata` holds only the *unmerged* tool-static subset.

For a tool whose metadata is enriched by a `permission_request` hook, an HTTP/SSE consumer will see two `metadata` keys with different content:

```json
{
  "type": "tool_call_confirmation",
  "tool_definition": {
    "metadata": {"danger": "high"}          // tool-static only
  },
  "metadata": {"danger": "high", "reason": "policy-x"}  // merged (hook wins)
}
```

A client reading `tool_definition.metadata` would miss the hook-contributed keys. Given that the PR description specifically calls out the SSE stream as a consumer ("The SSE client registry already serialises the event to JSON, so the field round-trips to HTTP consumers"), this inconsistency is worth calling out.

Consider removing the JSON tag from `Tool.Metadata` (tag it `json:"-"`) or documenting explicitly that consumers should always prefer the top-level `metadata` field over `tool_definition.metadata`.

pkg/tools/tools.go
```diff
@@ -183,6 +183,11 @@ type Tool struct {
 	// ModelOverride is the per-toolset model for the LLM turn that processes
 	// this tool's results. Set automatically from the toolset "model" field.
 	ModelOverride string `json:"-"`
+	// Metadata carries arbitrary key/value annotations a toolset attaches
+	// to a tool. The runtime forwards it onto the tool-call confirmation
+	// message so clients (TUI, HTTP) can render extra context for the
+	// approval prompt. Empty for tools that don't set any.
+	Metadata map[string]string `json:"metadata,omitempty"`
```
docker-agent: **[MEDIUM] `Tool.Metadata` json tag exposes static metadata in all Tool-embedding events, not just confirmation events**

Adding `json:"metadata,omitempty"` to `Tool.Metadata` causes the field to appear in the JSON representation of *every* `tools.Tool` that gets serialised — including `ToolCallEvent.ToolDefinition`, `ToolCallOutputEvent.ToolDefinition`, `ToolCallResponseEvent.ToolDefinition`, and others (grep for `ToolDefinition tools.Tool` in `event.go`).

This means:
- `ToolCallEvent` (fires as the tool begins executing) will now include `tool_definition.metadata`
- `ToolCallResponseEvent` (fires when the tool finishes) will also include it

These events are received by TUI, HTTP SSE, `acp`, `cli/runner`, `leantui`, and third-party SDK consumers (e.g. `examples/golibrary`). Adding an unexpected field to existing serialised event types is a breaking change for any consumer that uses strict JSON schema validation.

If the intent is to carry metadata only on the confirmation event, consider tagging `Tool.Metadata` as `json:"-"` and adding the merged metadata *only* to `ToolCallConfirmationEvent.Metadata` (which already holds the merged value). The confirmation event still carries everything clients need.

docker-agent (commented):

Two medium-severity findings in the new metadata serialisation surface. Core merge logic (hook keys win on clash, nil-safe maps, correct `maps.Copy` order) is correct and well-tested. The concerns are about the JSON wire shape exposed to HTTP/SSE consumers.

aheritier (dismissed):

melmennaoui (approved):

PR #3257, opened 2026-06-26
feat(config): accept ${env.X} in toolset env values (#2615)

dgageot:
Toolset `env` values previously only understood the shell-style `${X}` expansion syntax. When an agent config used the JS-template form `${env.X}` — which already works in path fields like `working_dir` and `path` — the runtime would reject it with a hard "environment variable not set" error, silently breaking configs that mixed both styles. This is the first part of fixing the inconsistency tracked in #2615.

The fix extracts a shared `NormalizeEnvRefs` helper in `pkg/path/expand.go` that rewrites `${env.X}` to `${X}` before expansion, then calls it from `pkg/environment/expand.go` ahead of `os.Expand`. Richer JS expressions such as `${env.VAR || 'default'}` are intentionally *not* evaluated in env values — only the plain `${env.X}` alias is handled. The variable-expansion quick-reference table in `docs/configuration/overview/index.md` has been updated to reflect that env values now accept both forms.

A related cleanup drops the now-obsolete startup warning that flagged `${env.X}` in toolset env values. The warning helper for shell and hook fields — which are forwarded verbatim to `exec.Cmd` with no expansion at all — was renamed `warnPathField` → `warnExecField` and its message corrected; the old message incorrectly told users to switch to `${X}` or `$X`, neither of which is substituted in those fields.

### DIFF
--- a/docs/configuration/overview/index.md
+++ b/docs/configuration/overview/index.md
@@ -260,7 +260,7 @@ Applies to:
 
 - `agents.<name>.toolsets[*].working_dir` (MCP, LSP)
 - `agents.<name>.toolsets[*].path` (memory, tasks)
-- `agents.<name>.toolsets[*].env` values (MCP, shell, script, LSP) — these go through `os.Expand`, not the JS evaluator, so `${env.X}` is **not** recognized here.
+- `agents.<name>.toolsets[*].env` values (MCP, shell, script, LSP)
 - The `~` prefix is also accepted in any path-like field documented as such.
 
 ```yaml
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly implements `NormalizeEnvRefs` to rewrite `${env.X}` → `${X}` before `os.Expand`, and integrates it cleanly in both `pkg/path/expand.go` and `pkg/environment/expand.go`. The regex replacement string `${\}` is valid Go regexp syntax and produces the expected output. The warning removal and `warnPathField` → `warnExecField` rename are consistent with the change. No bugs introduced by this PR were found.

aheritier (approved):

PR #3258, opened 2026-06-26
ci: parallelize build-image across native runners

dgageot:
## Problem

The `build-image` job is the long pole in CI at **~8 minutes** on every PR. Measured on real GitHub Actions runners, the root cause is that both `linux/amd64` and `linux/arm64` are built on a **single 2-vCPU runner**, so the two CPU-bound cgo/zig cross-compiles (~390s and ~400s each) run concurrently and starve each other for CPU.

## Approach

I benchmarked alternatives on real GHA runners (no-cache builds):

| Strategy | Wall-clock |
|---|---|
| Current (both platforms, 1 x86 runner) | **473s (7.9m)** |
| Split platforms across 2 x86 runners (cross-compile) | 261s (4.3m) |
| Split platforms on **native** runners (amd64 → `ubuntu-latest`, arm64 → `ubuntu-24.04-arm`) | **250s (4.2m)** ✅ |

> **Rejected:** simply disabling CGO. `pkg/rag/treesitter` uses the `go-tree-sitter` C library for code-aware RAG chunking; building with `CGO_ENABLED=0` silently degrades that feature at runtime. Not behavior-preserving.

## Changes

- **`build-image`** (PR job): now a matrix building each platform on its own native runner in parallel → **~47% faster (473s → 250s)**. Cache scopes are now per-platform.
- **`build-and-push-image`** (main/tags): split into per-platform native-runner builds that push **by digest** (with `sbom` + `provenance` attestations), plus a new **`merge-and-push-image`** job that assembles the multi-arch manifest via `docker buildx imagetools create` — Docker's official recommended pattern.

## Validation

- The full **push-by-digest + imagetools merge** flow was validated end-to-end against a local registry: the merged image is a correct multi-arch index containing **both `linux/amd64` and `linux/arm64`**, each with its **attestation manifest preserved** — behavior-identical to the original single-job `build-push-action`.
- `actionlint` and `scripts/workflow-lint.sh` both pass clean.
- `task lint` passes with **0 issues**; `task test` passes the full suite.

The only thing not exercised locally is the actual Docker Hub push (no credentials), but the registry interaction and manifest assembly are proven against a real registry.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,18 @@ jobs:
 
   build-image:
     if: github.event_name == 'pull_request'
-    runs-on: ubuntu-latest
+    # Build each platform on its own native runner, in parallel. Building both
+    # platforms on a single runner serializes two CPU-bound cgo/zig compiles on
+    # 2 vCPUs (~8 min); splitting across native runners cuts wall-clock to ~4 min.
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-latest
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The parallelization approach is sound. The push-by-digest + `imagetools create` pattern is Docker's official recommended approach for multi-arch manifest assembly, and the implementation follows it correctly.

**Reviewed areas:**
- ✅ **Cache scope with `/`** — `buildx-${{ matrix.platform }}` (e.g., `buildx-linux/amd64`) is the canonical pattern shown in Docker's own `build-push-action` documentation. BuildKit's GHA cache backend handles the slash correctly.
- ✅ **merge-and-push-image dependency safety** — When `fail-fast: false` and a matrix leg fails, GitHub Actions marks the whole matrix job as failed and the downstream `merge-and-push-image` job is **skipped** (not run), so no partial manifest can be pushed.
- ✅ **Digest export robustness** — The `if-no-files-found: error` guard on the upload-artifact step provides a reliable safety net if the digest output were ever unexpectedly empty.
- ✅ **Native runner selection** — `ubuntu-latest` for amd64 and `ubuntu-24.04-arm` for arm64 is correct; no cross-compile overhead.
- ✅ **PLATFORM_PAIR env var** — Correctly strips the `/` for artifact names (`linux/amd64` → `linux-amd64`), avoiding artifact name issues.
- ✅ **Attestations** — `sbom: true` and `provenance: mode=max` are preserved on the per-platform push steps.

melmennaoui (approved):

Issue #2812, opened 2026-05-18
Steer messages need a framing axis so callers can control how the model reads them


trungutt:
## Background

Today `Runtime.Steer` (and `POST /api/sessions/:id/steer`) drains queued
messages into `sess.Messages` as **plain user messages** — no wrapping,
no envelope. This was locked in by #2492 and #2518, both of which solved
real scheduling and tokenisation bugs. Those fixes should stay.

But the storage choice — plain text — was bundled in alongside the
scheduling fix, and it is the only "shape" available to callers. That
single shape collapses two very different user intents into the same
on-the-wire representation, and the model can't tell them apart.

## The problem

When a steered message arrives mid-turn, the user (or programmatic
caller) is trying to express one of two things:

1. **"Add this to your work, but finish what you're doing first."**
   *(common interactive case: "also use pytest")*

2. **"Drop what you're doing and do this instead."**
   *(explicit pivot: "stop, the bug is in db.ts, not api.ts")*

Both arrive at the model as bare text appended to the session. The model
has no signal to distinguish them, so it picks one — and empirically it
picks **"drop what you're doing"** most of the time, because a fresh
user turn after a tool batch reads as a new instruction. The "finish
first, then address" interpretation is essentially unreachable today.

A third axis exists for programmatic callers (chains, channel bridges,
AHP sessions) where neither preamble is wanted — they really do want a
plain user message. So the answer is not "always wrap" — it's
"let the caller pick".

## Illustration of the current behavior

```
Setup: agent is editing api.ts. User types "also update the tests".

         ┌──────────┐                              ┌─────────┐
         │  edit    │ ─── steer arrives ──────────▶│ LLM #3  │
         │  api.ts  │     drained, plain text      │         │
         │  ✓ done  │                              │         │
         └──────────┘                              └─────────┘
                                                       ▲
                                model sees this as the
                                last user turn:
                                ┌─────────────────────────┐
                                │ also update the tests   │
                                └─────────────────────────┘

What happens: model pivots to "update the tests", abandoning the
              partly-finished api.ts work. The user expected it
              to finish the bug fix first.
```

There is no API path today to get the "finish first" interpretation,
even though it's the more common interactive intent.

## Proposed change

Add an explicit **framing** field to `QueuedMessage` and to the
`/steer` HTTP payload. Framing is orthogonal to scheduling — the
three drain sites from #2492 and the newline behavior from #2518 are
unchanged.

```
┌─────────────────────────────────────────────────────────────────┐
[...]

PR #3264, opened 2026-06-26
feat(runtime): add framing axis to steered messages

Sayt-0:
## Summary

Adds an explicit **framing** axis to steered messages so callers can control how the model reads a mid-turn injection. Framing is orthogonal to scheduling: the three drain sites from #2492 and the newline behavior from #2518 are unchanged, and the `user_steering_messages_submit` hook still receives raw, unframed text.

Until now a drained steer message was always injected as plain user text, which the model tends to read as "drop what you're doing". The "finish first, then address" intent was effectively unreachable.

## The three framings

| `framing` | What the model sees | For |
|---|---|---|
| `plain` | bare user turn (previous behavior) | programmatic callers (chains, channels, bridges) |
| `instruction` | `<system-reminder>` asking it to finish the current task first, then address the message | interactive course corrections ("also do Y") |
| `replacement` | `<system-reminder>` asking it to abandon the current task and address the message instead | explicit pivots ("stop, do X instead") |

## Wire shape

Framing is request-level on the HTTP API and applies to every message in the request:

```
POST /api/sessions/:id/steer
{ "messages": [{ "content": "use pytest, not unittest" }], "framing": "instruction" }
```

An unrecognized value returns `400`. An omitted value defaults per the table below.

## Issue expectations mapped

| Issue #2812 expectation | Implementation | Status |
|---|---|---|
| `framing` field on `QueuedMessage` | `pkg/runtime/message_queue.go` (`Framing` type + field + `Valid()`) | done |
| `framing` on the `/steer` payload | `pkg/api/types.go` (`SteerSessionRequest.Framing`) | done |
| three framings plain / instruction / replacement | `frameSteeredMessage` in `pkg/runtime/loop.go` | done |
| default `POST /steer` = instruction | resolved at drain via `resolveSteerFraming` | done |
| default `Runtime.Steer(QueuedMessage{})` = instruction | zero value resolves to instruction | done |
| default `POST /followup` = plain (unchanged) | follow-up path ignores framing | done |
| `Runtime.FollowUp` stays as is | unchanged, always plain | done |
| scheduling drain sites (#2492) unchanged | only rendering changed, not drain sites | done |
| newline behavior (#2518) unchanged | newline applied after framing per message | done |
| hooks see raw text | hook input built from raw content before framing | done |

## Behavior change to weigh

Per the issue's defaults table, the default for `POST /steer` and zero-value `Runtime.Steer` changes from plain to **instruction**, so existing steer callers now get the `<system-reminder>` envelope by default. Programmatic callers that want bare text opt into `plain`. If a different default is preferred, only `resolveSteerFraming` needs to change.

## Verification

- New unit + integration tests in `pkg/runtime/steer_framing_test.go`, plus strengthened existing steer/follow-up tests; scheduling and newline tests pinned to `plain` to isolate them from rendering.
- `go test` (including `-race`) on `pkg/runtime`, `pkg/server`, `pkg/api`, `pkg/mcp` passes; `golangci-lint` reports no issues on the touched packages.
- Verified end-to-end against a real model:

| framing | persisted envelope | observed model behavior |
|---|---|---|
| plain | none | treated as a normal user message |
| instruction | present | finished the original task, then addressed the new one |
| replacement | present | abandoned the original task and pivoted to the new one |

## Docs

`docs/features/api-server/index.md` gains a "Steering a running turn" section documenting the field, defaults, and curl examples. `agent-schema.json` is not affected (this is an HTTP API change, not an agent-config feature).

### DIFF
--- a/docs/features/api-server/index.md
+++ b/docs/features/api-server/index.md
@@ -54,7 +54,7 @@ All endpoints are under the `/api` prefix.
 | `POST`   | `/api/sessions/:id/resume`          | Resume a paused session (after tool confirmation)       |
 | `POST`   | `/api/sessions/:id/tools/toggle`    | Toggle auto-approve (YOLO) mode                         |
 | `POST`   | `/api/sessions/:id/elicitation`     | Respond to an MCP tool elicitation request              |
-| `POST`   | `/api/sessions/:id/steer`           | Inject messages into a running turn (pre-empts current) |
+| `POST`   | `/api/sessions/:id/steer`           | Inject messages into a running turn, with a `framing` axis (see [Steering a running turn](#steering-a-running-turn)) |
 | `POST`   | `/api/sessions/:id/followup`        | Enqueue messages to run after the current turn finishes (supports an `Idempotency-Key` — see [Idempotent follow-ups](#idempotent-follow-ups)). |
 | `GET`    | `/api/sessions/:id/models`          | List available models for the session's current agent   |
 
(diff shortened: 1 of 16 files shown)


rumpl:
What's the usecase for this? Does this need to be in the runtime? If a client wants a different framing they can inject whatever in the steering message

dgageot:
This original issue has been closed

PR #3265, opened 2026-06-26
fix(tui): restore ctrl+1..ctrl+9 agent quick-switch shortcuts

dgageot:
The `ctrl+1` through `ctrl+9` shortcuts for switching between agents stopped working on terminals that support the Kitty keyboard protocol. Under that protocol, key events carry a `Text` field, which caused `tea.KeyPressMsg.String()` to return the bare digit (e.g. `"1"`) instead of `"ctrl+1"`, so the string-based match in `parseCtrlNumberKey` silently failed.

`parseCtrlNumberKey` now inspects the `Mod` and `Code`/`BaseCode` fields of the key event directly rather than parsing the result of `String()`. It prefers `BaseCode` — the PC-101 layout key — so the shortcut works correctly on international keyboard layouts too. Lock-state modifiers (`CapsLock`, `NumLock`, `ScrollLock`) that the Kitty protocol passes along in the `Mod` field are masked out before the comparison, so the shortcut fires even when Caps Lock or Num Lock is on.

The test suite for this helper was rewritten to construct `tea.KeyPressMsg` values directly instead of relying on string parsing, with new cases covering the Kitty `Text` regression, `BaseCode`/international layout, and lock-modifier scenarios.

### DIFF
--- a/pkg/tui/tui.go
+++ b/pkg/tui/tui.go
@@ -2027,11 +2027,24 @@ func (m *appModel) handleKeyPress(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
 	return m, nil
 }
 
-// parseCtrlNumberKey checks if msg is ctrl+1 through ctrl+9 and returns the index (0-8), or -1 if not matched
+// parseCtrlNumberKey checks if msg is ctrl+1 through ctrl+9 and returns the index (0-8), or -1 if not matched.
+//
+// It inspects the key's modifiers and code directly rather than msg.String():
+// terminals with keyboard enhancements (e.g. the Kitty protocol) populate the
+// key's Text field, which makes String() return the bare digit ("1") instead of
+// "ctrl+1", silently breaking string-based matching.
 func parseCtrlNumberKey(msg tea.KeyPressMsg) int {
-	s := msg.String()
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The rewrite of `parseCtrlNumberKey` in `pkg/tui/tui.go` is correct and well-structured:

- **Modifier mask** (`msg.Mod&^lockModifiers != tea.ModCtrl`) correctly strips lock-state bits before the equality check, so the shortcut fires with CapsLock/NumLock on and rejects any combination that includes Shift, Alt, or Meta alongside Ctrl.
- **BaseCode fallback** follows the standard bubbletea pattern — `BaseCode == 0` means "not set", so the fallback to `Code` is safe and correctly handles international keyboard layouts.
- **Type correctness** — `Code` and `BaseCode` are `rune` (int32), making the `>= '1'`/`<= '9'` comparisons and `code - '1'` arithmetic correct.
- **Test coverage** in `tui_helpers_test.go` covers the Kitty-protocol `Text`-field regression, international layout via `BaseCode`, lock-modifier scenarios, and the relevant negative cases.

aheritier (approved):

PR #3266, opened 2026-06-26
refactor(tui): remove redundant agent-switch and thinking-level toasts

dgageot:
Two toast notifications in the TUI were firing on actions whose result is already visible in the interface without them. The "Switched to agent" success toast appeared every time the user switched the active agent, even though the chat view header immediately reflects the new agent. The "✻ Thinking: low" info toast appeared on every thinking-level cycle, even though the sidebar already shows the current effort level. Neither notification added information the user couldn't already see.

Both toasts have been removed from `handleSwitchAgent` and `handleCycleThinkingLevel` in `pkg/tui/handlers.go`. All other behavior is unchanged: the underlying state mutations, keybindings, and visual indicators in the chat and sidebar remain exactly as before.

No breaking changes. The only observable difference is the absence of those two transient pop-ups.

### DIFF
--- a/pkg/tui/handlers.go
+++ b/pkg/tui/handlers.go
@@ -323,10 +323,8 @@ func (m *appModel) handleSwitchAgent(agentName string) (tea.Model, tea.Cmd) {
 		return m, notification.ErrorCmd(fmt.Sprintf("Failed to switch to agent '%s': %v", agentName, err))
 	}
 	m.sessionState.SetCurrentAgentName(agentName)
-	return m, tea.Batch(
-		m.updateChatCmd(messages.SessionToggleChangedMsg{}),
-		notification.SuccessCmd(fmt.Sprintf("Switched to agent '%s'", agentName)),
-	)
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Clean refactor that removes two redundant toast notifications. Both  and  correctly preserve their error paths, state mutations, and return semantics. The  simplification in  and the  in  are both valid Bubble Tea patterns with no behavioral regression.

docker-agent (commented):
### Assessment: 🟢 APPROVE

Clean refactor that removes two redundant toast notifications. Both `handleSwitchAgent` and `handleCycleThinkingLevel` correctly preserve their error paths, state mutations, and return semantics. The `tea.Batch` simplification in `handleSwitchAgent` and the `return m, nil` in `handleCycleThinkingLevel` are both valid Bubble Tea patterns with no behavioral regression.

aheritier (approved):

PR #3269, opened 2026-06-26
ci: install Task from prebuilt binary instead of compiling

dgageot:
The `build-and-test` job spent ~51s on `go install github.com/go-task/task/v3/cmd/task@latest`, which compiles Task from source on every run.

This switches to `arduino/setup-task`, which downloads a pinned prebuilt Task binary, removing the compile step.

- Action is pinned by 40-char SHA (`b91d5d2…` = `v2.0.0`), per repo policy.
- `version: 3.51.1` is an exact pin; the action re-adds the `v` prefix itself.

**Verified on CI** (this PR's runs): the *Install Task* step dropped from **~51s → 1s**, with build-and-test still passing.

Validated locally with `actionlint` and `scripts/workflow-lint.sh`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,9 @@ jobs:
           cache: true
 
       - name: Install Task
-        run: go install github.com/go-task/task/v3/cmd/task@latest
+        uses: arduino/setup-task@b91d5d2c96a56797b48ac1e0e89220bf64044611 # v2.0.0
+        with:
+          version: 3.51.1
 
       - name: Create bin directory
         run: mkdir -p "$HOME/bin"


.github/workflows/ci.yml
```diff
@@ -55,7 +55,10 @@ jobs:
           cache: true
 
       - name: Install Task
-        run: go install github.com/go-task/task/v3/cmd/task@latest
+        uses: arduino/setup-task@b91d5d2c96a56797b48ac1e0e89220bf64044611 # v2.0.0
```
docker-agent, line 58: **[LOW] Task pinned at 3.51.1 instead of latest — manual bump required for future upgrades**

Task is now hard-pinned at `3.51.1` rather than `@latest`. This is intentional for reproducibility and removes non-determinism from CI, but it introduces an upgrade-hygiene burden: the version must be manually bumped whenever a newer Task release is needed. Consider tracking this via Dependabot or Renovate so upgrades don't get forgotten.

.github/workflows/ci.yml
```diff
@@ -55,7 +55,10 @@ jobs:
...
+        with:
+          version: 3.51.1
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
docker-agent: **[LOW] repo-token passes GITHUB_TOKEN to a third-party action — confirm SHA resolves to v2.0.0**

`repo-token: ${{ secrets.GITHUB_TOKEN }}` is passed to `arduino/setup-task` to avoid GitHub API rate-limiting when fetching release metadata. This is standard practice and doesn't grant elevated permissions. The SHA pin (`b91d5d2c96a56797b48ac1e0e89220bf64044611`) mitigates supply-chain risk, but it's worth confirming this SHA actually resolves to the expected `v2.0.0` tag on the `arduino/setup-task` repo before merging (e.g., `gh api repos/arduino/setup-task/git/ref/tags/v2.0.0`).

docker-agent (commented):
### Assessment: 🟢 APPROVE

The change is structurally correct — `uses:` syntax is valid, the SHA pin and version comment are internally consistent (`b91d5d2c96a56797b48ac1e0e89220bf64044611` / `# v2.0.0`), and `repo-token` is used appropriately. Switching from `go install … @latest` to a prebuilt binary download is a clean, well-scoped improvement. No high or medium severity issues found.

gtardif (approved):

PR #3270, opened 2026-06-26
docs: document URL-opening /commands and clean up /feedback and /bug

dgageot:
PR #3261 introduced a `url` field for agent `/commands` that opens a link in the user's default browser instead of sending a prompt to the agent. This PR adds the missing documentation for that feature and reverts the `/feedback` and `/bug` slash command definitions that #3261 had added but that don't belong in the default command set.

The docs now cover the full command format in `docs/configuration/agents/index.md`: the existing prompt-style commands, the new URL command variant, supported URI schemes (including custom ones like `docker-desktop://`), the `{{session_id}}` placeholder that is URL-query-escaped at invocation time, and a note that URL commands are TUI-only. The slash-commands reference in `docs/features/tui/index.md` has also been updated to reflect this.

On the code side, `pkg/tui/commands/commands.go` drops the `SlashCommand` and `Immediate` fields from the `/feedback` and `/bug` entries so they remain command-palette-only, and the corresponding parser subtests are removed from `commands_test.go`.


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly reverts the `/feedback` and `/bug` slash command definitions from #3261, leaving those commands as palette-only items. No new bugs are introduced by the changes. The documentation additions for the `url` field are accurate and well-structured.

Sayt-0 (approved):

PR #3272, opened 2026-06-26
feat(config): support external instruction files via instruction_file

### LINKED ISSUE #1666: Support external instruction files for better configuration maintainability
## Problem

Currently, agent instructions must be embedded inline in the YAML configuration file, which forces mixing configuration concerns (infrastructure: models, providers, tools) with behavioral content (agent instructions).

```yaml
agents:
  coordinator:
    model: claude-haiku
    instruction: |
      You are a coordinator agent that...
      [potentially 100+ lines of instruction text inline]
```

This creates several maintainability issues:

1. **Version Control**: Diffs mix infrastructure changes with instruction updates (noisy)
2. **Team Collaboration**: Single large file difficult to review and merge
3. **Instruction Reuse**: Can't template or version instructions independently
4. **Editing Risk**: Updating instructions risks YAML syntax errors

## Solution

Support external instruction file references:

```yaml
agents:
  coordinator:
    model: claude-haiku
    instruction_file: instructions/coordinator.md
```

Or with remote references for versioning:

```yaml
agents:
  coordinator:
    model: claude-haiku
    instruction_ref: docker://registry.example.com/instructions:coordinator-v1.0
```

## Benefits

- ✅ Separates configuration from content (clean separation of concerns)
- ✅ Safe editing (no YAML syntax risk)
- ✅ Meaningful version control diffs
- ✅ Team collaboration (separate files = fewer merge conflicts)
- ✅ Enables instruction templating and versioning
- ✅ Supports advanced patterns (instruction versioning, A/B testing)

### Option 1: File References (Recommended)
```yaml
instruction_file: path/to/instructions/agent.md
```
- Load from local file system
- Relative to config file location
- Simple, immediate benefit

### Option 2: Remote References
```yaml
instruction_ref: docker://registry/instructions:agent-v1.0
```
- Load from OCI artifact registry
- Enable versioning and rollback
- More complex but powerful

### Backward Compatibility
- Keep `instruction:` field working for inline text
- New `instruction_file:` parameter optional
- No breaking changes

## Use Case

Multi-agent setups with coordinator and specialist agents need clear separation between infrastructure configuration and behavioral specifications.

## Acceptance Criteria

- [ ] `instruction_file` parameter supported in agent configuration
- [ ] Supports relative paths (relative to config file)
- [ ] File content loaded at startup
- [ ] Clear error messages if file not found
- [ ] Backward compatible with inline `instruction:` parameter
- [ ] Documentation updated with examples
- [ ] Tests cover file loading scenarios

Sayt-0:
## Summary

Adds an optional `instruction_file` field on agent config so an agent's instruction can live in its own file (relative to the config file's directory) instead of being inlined in the YAML. This separates infrastructure configuration from behavioral content, which keeps version-control diffs focused, reduces merge conflicts on shared configs, and avoids YAML syntax risk when editing long prompts.

Scope is Option 1 (local file references) from the issue. The inline `instruction` field stays fully supported and is not deprecated.

```yaml
agents:
  coordinator:
    model: openai/gpt-5-mini
    description: Routes work between specialist agents
    instruction_file: instructions/coordinator.md
```

## Acceptance criteria

| Criterion | Status | Notes |
| --- | --- | --- |
| `instruction_file` parameter supported | yes | New field on the latest `AgentConfig` |
| Relative paths (relative to config file) | yes | Resolved against `Source.ParentDir()` |
| File content loaded at startup | yes | Resolved in `config.Load`, between migration and validation |
| Clear error messages if file not found | yes | Errors name the agent and the path |
| Backward compatible with inline `instruction` | yes | Both fields supported; setting both is rejected |
| Documentation updated with examples | yes | New docs section plus a runnable example |
| Tests cover file loading scenarios | yes | See test list below |

## Design decisions

- **Resolution point.** `config.Load` is the single choke point for every consumer (run, teamloader, OCI push, sandbox kit), so resolution happens there. The whole pipeline only ever sees the inlined `Instruction`; no downstream change is needed.
- **Self-containment.** After reading the file into `Instruction`, `InstructionFile` is cleared. This keeps the in-memory config self-contained and is required for the marshalling round-trip test (`TestParseExamplesAfterMarshalling` reloads via a parentless bytes source).
- **Path safety.** The reference must be a local relative path inside the config directory. Absolute paths and `..` traversal are rejected via `filepath.IsLocal`, and reads are confined with `os.OpenRoot` (TOCTOU-safe, blocks symlink escapes). This mirrors the existing HCL `file()` helper and `fileSource.Read`.
- **Mutual exclusivity.** `instruction` and `instruction_file` cannot both be set.
- **Parentless sources.** OCI/URL/bytes sources have no directory to resolve against, so `instruction_file` is rejected there with a clear message. `share push` re-serializes the resolved config when `instruction_file` is used, so pushed artifacts inline the contents and stay self-contained.

## Files

| File | Change |
| --- | --- |
| `pkg/config/latest/types.go` | New `InstructionFile` field (latest config only; v0..v10 frozen) |
| `pkg/config/config.go` | `resolveInstructionFiles` called from `Load` |
| `agent-schema.json` | `instruction_file` added to the `AgentConfig` definition |
| `pkg/oci/package.go` | Re-serialize resolved config when `instruction_file` is used |
| `examples/instruction_file.yaml`, `examples/instructions/*.md` | Runnable coordinator/specialist example |
| `pkg/config/instruction_file_test.go`, `pkg/oci/package_test.go` | Tests |
| `docs/configuration/agents/index.md` | External Instruction Files section plus reference row |

Test coverage: resolution from a relative path, file in a subdirectory, missing file, `..` traversal rejected, absolute path rejected, symlink escape rejected, mutual exclusivity, parentless source rejected, empty-string treated as unset, and OCI push inlining (loads from the pushed bytes without the local file).

## Validation

- `golangci-lint run ./pkg/config/... ./pkg/oci/...` reports no issues.
- `go build ./...` passes (the `lint/` tooling module needs a network fetch unavailable in the sandbox and is unrelated).
- `pkg/config` and `pkg/oci` suites pass. The only failures in the local run are `TestURLSource_Read_RejectsLocalAddresses`, which are network-dependent SSRF tests unrelated to this change (they time out instead of rejecting because outbound network is blocked in the sandbox).

## Possible follow-up

Option 2 from the issue (`instruction_ref` for OCI artifact references) is not included here and can be added later if maintainers want remote versioning.

Closes #1666

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -561,6 +561,10 @@
           "type": "string",
           "description": "Instructions for the agent"
         },
+        "instruction_file": {
+          "type": "string",
+          "description": "Path to a file, relative to the config file's directory, whose contents become the agent's instruction. Loaded at startup. Mutually exclusive with 'instruction'. Must be a local relative path inside the config directory (absolute paths and '..' traversal are rejected). Only supported for local file-based configs, not OCI/URL sources."
+        },
         "harness": {
           "$ref": "#/definitions/HarnessConfig",
           "description": "External coding harness to run this agent with instead of a docker-agent model provider"
--- a/docs/configuration/agents/index.md
+++ b/docs/configuration/agents/index.md
@@ -16,7 +16,8 @@ agents:
   agent_name:
     model: string # Required: model reference
     description: string # Required: what this agent does
-    instruction: string # Required: system prompt
+    instruction: string # Required (unless instruction_file): system prompt
+    instruction_file: string # Optional: load the system prompt from a file relative to this config (mutually exclusive with instruction)
     sub_agents: [list] # Optional: local or external sub-agent references
     toolsets: [list] # Optional: tool configurations (use `type: rag` for RAG sources)
     fallback: # Optional: fallback config
@@ -81,7 +82,8 @@ agents:
 | --------------------------- | ------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `model`                     | string  | ✓        | Model reference. Either inline (`openai/gpt-5`) or a named model from the `models` section.                                                                              |
 | `description`               | string  | ✓        | Brief description of the agent's purpose. Used by coordinators to decide delegation.                                                                                          |
-| `instruction`               | string  | ✓        | System prompt that defines the agent's behavior, personality, and constraints.                                                                                                |
+| `instruction`               | string  | ✓        | System prompt that defines the agent's behavior, personality, and constraints. Required unless `instruction_file` is set.                                                      |
+| `instruction_file`          | string  | ✗        | Path to a file (relative to the config file's directory) whose contents become the agent's instruction, loaded at startup. Mutually exclusive with `instruction`. Must be a local relative path inside the config directory (absolute paths and `..` traversal are rejected). Only supported for local file-based configs, not OCI/URL sources. See [External Instruction Files](#external-instruction-files) below. |
 | `sub_agents`                | array   | ✗        | List of agent names or external OCI references this agent can delegate to. Supports local agents, registry references (e.g., `agentcatalog/pirate`), and named references (`name:reference`). Automatically enables the `transfer_task` tool. Pin external OCI references to a digest (`name@sha256:…`) to skip the per-run registry lookup that tag references incur. See [External Sub-Agents]({{ '/concepts/multi-agent/#external-sub-agents-from-registries' | relative_url }}). |
 | `toolsets`                  | array   | ✗        | List of tool configurations. See [Tool Config]({{ '/configuration/tools/' | relative_url }}).                                                                                                        |
 | `fallback`                  | object  | ✗        | Automatic model failover configuration.                                                                                                                                       |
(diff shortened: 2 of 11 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `instruction_file` feature is well-implemented. The security design is sound:
- `filepath.IsLocal` provides a first-pass rejection of `..` traversal and absolute paths
- `os.OpenRoot` confines reads to the config directory, blocking symlink escapes at the OS level (TOCTOU-safe)
- Parentless sources (OCI/URL/bytes) are correctly rejected with clear error messages
- Mutual exclusivity between `instruction` and `instruction_file` is correctly enforced
- The OCI push re-serialization path correctly inlines resolved instructions before pushing

Two hypotheses about the OCI re-serialization path were investigated and dismissed:
1. **HCL + `instruction_file`**: An HCL source fails at the initial `yaml.Unmarshal` version-probe (line 43–45), returning an explicit error before `configUsesInstructionFile` is ever called — no silent bypass.
2. **`goccy/go-yaml` type assertion**: `goccy/go-yaml v1.19.2` decodes string-keyed YAML mappings as `map[string]interface{}`, so the `probe["agents"].(map[string]any)` assertion succeeds correctly for valid YAML configs.

One low-severity observation (not a bug): `os.OpenRoot(parentDir)` is called inside the per-agent loop. Opening the directory root once before the loop and reusing it for all agents would be slightly more efficient, but this is a style/performance note rather than a correctness issue and does not warrant blocking the PR.

dgageot (approved):

PR #3275, opened 2026-06-26
docs(runtime): document StreamStopped ordering and teardown trade-offs (#3074)

### LINKED ISSUE #3074: decide StreamStopped ordering vs session-end hooks and document teardown trade-offs from #3070
## Summary

Follow-up to #3070 (fix for #3069). The fix is correct, but it made two deliberate behavior trade-offs around `StreamStopped` that deserve an explicit decision and documentation:

1. **Ordering change:** `StreamStopped` is now emitted **before** session-end hooks run. Before #3070 it was emitted after `executeSessionEndHooks` (and before `executeOnUserInputHooks`/telemetry).
2. **Delivery change:** `StreamStopped` is now emitted **non-blockingly** — it is dropped when the events buffer is full and the consumer has gone away (previously a blocking send: guaranteed delivery, but it could deadlock teardown — the very issue #3070 fixed).

## 1. Ordering: StreamStopped before session-end hooks

Current order in `finalizeEventChannel` (`pkg/runtime/loop.go`):

```go
nonBlocking(&channelSink{ch: events}).Emit(StreamStopped(...))  // 1. emit
r.executeSessionEndHooks(context.WithoutCancel(ctx), sess, a)   // 2. hooks
r.executeOnUserInputHooks(ctx, sess.ID, "stream stopped")       // 3. hooks
r.telemetry.RecordSessionEnd(ctx)                                // 4. telemetry
r.elicitation.restoreAndClose(events, prevElicitationCh)         // 5. close (last)
```

Consumers (TUI, A2A forwarders, persistence observers) now observe `StreamStopped` while session-end hooks are still executing. If any consumer treats the **event** (rather than the channel close) as "all cleanup done", that is a semantic shift. The channel close — which terminates `for range` — still happens last and remains the authoritative terminal signal.

**Option A (recommended):** move the emit to just before `restoreAndClose` (after hooks/telemetry). One-line move, restores the pre-#3070 semantics, keeps close-last:

```go
r.executeSessionEndHooks(...)
r.executeOnUserInputHooks(...)
r.telemetry.RecordSessionEnd(ctx)
nonBlocking(&channelSink{ch: events}).Emit(StreamStopped(...))
r.elicitation.restoreAndClose(events, prevElicitationCh)
```

**Option B:** keep the current order and document that the channel close, not the event, is the terminal signal.

Either way, audit consumers of `StreamStoppedEvent` for assumptions about hook completion.

## 2. Documentation: accepted trade-offs in restoreAndClose

Two deliberate behaviors from #3070 are currently undocumented and should be captured in doc comments so future maintainers don't "fix" them:

- `restoreAndClose` **waits** for a parked in-flight elicitation send (the sender holds the bridge `RLock`). If the consumer is truly gone and both buffers are full, teardown blocks instead of panicking — a goroutine leak is the accepted alternative to crashing the process. Document this in the `restoreAndClose` comment.
- `StreamStopped` is **best-effort** under buffer-full/no-consumer conditions: it is dropped rather than deadlocking teardown. Document this in the `finalizeEventChannel` comment (and in `StreamStoppedEvent`'s doc if consumers are expected to rely on it).

## Scope

- One-line emit reorder (if Option A) in `pkg/runtime/loop.go`
- Doc comments in `pkg/runtime/elicitation.go` and `pkg/runtime/loop.go`
- Optional: a test asserting `StreamStopped` is the last event delivered before close

## Relations

- Follow-up to #3070 / #3069.

Sayt-0:
## What

Follow-up to #3074, the documentation and decision debt left by #3070 (fix for #3069). No production behavior change: the only code edits are doc comments; the runtime ordering is kept as-is (Option B) and a regression test is added.

## Decision: Option B (keep current order)

`StreamStopped` stays emitted before the session-end hooks and telemetry, rather than moving it after (Option A). An audit of every `StreamStoppedEvent` consumer (TUI chat, sidebar, supervisor, leantui, A2A adapter, API server) shows none rely on session-end hooks having completed, so the order is safe, and emit-before-hooks keeps the TUI responsive (spinner stop and queue drain are not delayed behind user-configured `session_end` hooks). The authoritative terminal signal remains the channel close, not the event. Full rationale and the consumer audit are in #3074.

This is opened as a draft because #3074 listed Option A as the recommended option; a maintainer decision on B versus A would unblock marking it ready. Option A is a one-line move and the docs plus test apply unchanged either way.

## Changes

| File | Change |
|---|---|
| `pkg/runtime/elicitation.go` | Doc on `restoreAndClose`: close-under-write-lock is the #3069 fix, and the accepted trade-off that teardown blocks (goroutine leak) instead of panicking when a parked sender cannot drain |
| `pkg/runtime/loop.go` | Doc on `finalizeEventChannel`: ordering decision (#3074) and best-effort delivery; inline note on the non-blocking emit warning against reintroducing the #3070 deadlock |
| `pkg/runtime/event.go` | Doc on `StreamStoppedEvent`: best-effort delivery, channel close is the terminal signal, hooks run after the event |
| `pkg/runtime/elicitation_test.go` | New `TestLocalRuntime_FinalizeEventChannelStreamStoppedIsLastBeforeClose`: asserts exactly one StreamStopped and that it is the last event delivered before the channel close |

## Issue scope mapping

| Issue item | Status |
|---|---|
| Ordering decision | Option B, documented |
| Document `restoreAndClose` leak-over-panic trade-off | done |
| Document StreamStopped best-effort delivery | done |
| Test: StreamStopped last before close | done |

## Validation

| Command | Result |
|---|---|
| `task build` | pass |
| `golangci-lint run ./pkg/runtime/...` | 0 issues |
| `go test ./pkg/runtime/` | pass |
| `go test -race ./pkg/runtime/ -run 'Elicitation\|FinalizeEventChannel\|RunStreamClosesChannel'` | pass |

Note: `go test -race ./pkg/runtime/` (full package) reports a data race in `TestHandleStream_ContextCancellation`. It is pre-existing and unrelated to this change (reproduced on a tree with these edits stashed) and lives in the `stalledStream.Close()` test double in `streaming_test.go`, which is invoked concurrently without synchronization. It does not surface under the standard `task test` run (no `-race`). Out of scope here; a separate fix can add a mutex or atomic to `stalledStream.Close()`.

## Relations

- Resolves #3074
- Follow-up to #3070 / #3069

### DIFF
--- a/pkg/runtime/elicitation.go
+++ b/pkg/runtime/elicitation.go
@@ -92,7 +92,17 @@ func (b *elicitationBridge) send(ev Event) (err error) {
 }
 
 // restoreAndClose restores the previous stream channel and closes the current
-// stream channel under the bridge write lock.
+// stream channel under the bridge write lock, so the close is mutually
+// exclusive with an in-flight send. This is the #3069 fix: close can no longer
+// race a parked sender and panic with "send on closed channel".
+//
+// Accepted trade-off (do not "fix" by dropping the lock): holding the write
+// lock makes restoreAndClose wait for any in-flight send to finish, because
+// send holds the read lock across "b.ch <- ev". If the stream consumer has
+// gone away and current is full (or unbuffered), that parked send never
+// drains, so this call blocks on Lock forever and the teardown goroutine
+// leaks. A leaked goroutine is the deliberate, accepted alternative to
+// crashing the whole process with a send-on-closed-channel panic.
 func (b *elicitationBridge) restoreAndClose(current, previous chan Event) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
--- a/pkg/runtime/elicitation_test.go
+++ b/pkg/runtime/elicitation_test.go
@@ -236,6 +236,49 @@ func TestLocalRuntime_FinalizeEventChannelDoesNotDeadlockWhenBufferFullAndConsum
 	assert.Zero(t, stopped, "StreamStopped should be dropped instead of blocking when the buffer is full")
 }
 
+// TestLocalRuntime_FinalizeEventChannelStreamStoppedIsLastBeforeClose pins the
+// Option B decision from #3074: StreamStopped is emitted before the session-end
+// hooks and telemetry run (so the UI reacts immediately), but it must still be
+// the final event a consumer observes before the channel close that terminates
+// its range. Session-end hooks dispatch with a nil event sink, so nothing is
+// delivered onto the stream after StreamStopped and the close is the terminal
+// signal. If a future change emits onto the events channel after StreamStopped
+// (for example by handing the events sink to a session_end hook), this fails.
+func TestLocalRuntime_FinalizeEventChannelStreamStoppedIsLastBeforeClose(t *testing.T) {
+	t.Parallel()
+
+	rt := newElicitationTestRuntime(t)
+	sess := session.New()
+	events := make(chan Event, defaultEventChannelCapacity)
+	parent := make(chan Event, 1)
+
+	// Seed events that stand in for the stream's prior output so asserting
+	// StreamStopped is *last* is a real ordering check, not merely "it was the
+	// only event delivered".
+	events <- Error("prior stream output 1")
+	events <- Error("prior stream output 2")
+	rt.elicitation.swap(events)
+
+	rt.finalizeEventChannel(t.Context(), sess, turnEndReasonNormal, parent, events)
+
+	var delivered []Event
+	for ev := range events {
+		delivered = append(delivered, ev)
+	}
+
+	require.NotEmpty(t, delivered, "expected events delivered before the channel close")
+
+	var stopped int
+	for _, ev := range delivered {
+		if _, ok := ev.(*StreamStoppedEvent); ok {
+			stopped++
+		}
+	}
+	assert.Equal(t, 1, stopped, "exactly one StreamStopped should be delivered")
+	assert.IsType(t, &StreamStoppedEvent{}, delivered[len(delivered)-1],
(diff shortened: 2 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a documentation-only PR adding doc comments to three runtime files and one regression test. No production logic changes.

**Drafter findings**: The single medium-severity hypothesis (StreamStopped potentially dropped due to buffer exhaustion in the new test) was **dismissed** by verification: `defaultEventChannelCapacity = 128` leaves 126 free slots after the 2 pre-seeded events, so the non-blocking emit succeeds without issue.

The remaining low-severity observations (doc understatement about best-effort delivery, latent test fragility around synchronous channel close) are pre-existing design characteristics, not problems introduced by this PR.

The new test correctly pins the Option B ordering decision from #3074: StreamStopped is emitted before session-end hooks, it is the last event before channel close, and the assertion structure is sound.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR is a documentation-only change — no production behaviour is altered. The new doc comments on `restoreAndClose`, `StreamStoppedEvent`, and `finalizeEventChannel` accurately describe the existing code, correctly document the #3069/#3070/#3074 design decisions, and follow the project's "explain the **why**" comment philosophy from AGENTS.md.

The new regression test (`TestLocalRuntime_FinalizeEventChannelStreamStoppedIsLastBeforeClose`) correctly exercises the ordering guarantee using a buffered channel (`defaultEventChannelCapacity = 128`) and a real runtime instance, with no race conditions identified.

No bugs introduced by the `+` lines were found.

aheritier (approved):

PR #3277, opened 2026-06-26
feat(modelsdev): embed models.dev catalog snapshot as binary fallback

dgageot:
Without a network connection and with no on-disk cache, the agent currently falls back to a small set of hardcoded model defaults. This means a fresh install in an air-gapped or offline environment has no useful model catalog at all. This change embeds a build-time snapshot of the full models.dev catalog into the binary as a last-resort fallback, so a fresh binary always ships with a complete, usable catalog even when nothing else is available.

The snapshot (`pkg/modelsdev/snapshot.json`, ~1.5 MB compact JSON) is generated by a small tool in `pkg/modelsdev/internal/gen` that fetches `models.dev/api.json` and re-marshals it through the trimmed `Database` type to keep the file small. It is regenerated via `task update-models` or `go generate ./pkg/modelsdev/...`. A weekly GitHub Actions workflow (`.github/workflows/update-models.yml`) keeps the snapshot current by opening an automated PR — this avoids stale data without tying reproducibility to a live network fetch at build time. `loadDatabase` now returns both a database and an *authoritative* flag; the `Store` only memoizes results that are authoritative (cache hit or live fetch), so a transient fetch failure does not pin the embedded fallback for the `Store`'s lifetime — a subsequent lookup will retry once the network recovers.

Because the snapshot reliably carries Claude Fable and Opus 4.6–4.8 with their correct 1 M context windows, the hardcoded `DefaultClaudeContextLimit` special-casing in the Anthropic client has been removed. The client now falls back to a conservative 200 k floor only when no catalog entry exists at all. The snapshot freshness test (`TestSnapshotDateIsFresh`) is opt-in via the `CHECK_MODELS_SNAPSHOT_FRESHNESS` env var so it does not become a CI time-bomb on forks or stale branches after 90 days; a `task check-models-fresh` target runs it explicitly, while a separate always-on test validates that the embedded date is well-formed.


pkg/modelsdev/snapshot.go
```diff
@@ -0,0 +1,58 @@
...
+
+// embeddedSnapshot returns the catalog baked into the binary at build time.
+// The JSON is parsed once and cached. A malformed snapshot yields an empty
+// (non-nil) database rather than a panic, so a bad embed degrades to
+// "provider not found" instead of crashing the process.
+func embeddedSnapshot() *Database {
+	snapshotOnce.Do(func() {
+		var providers map[string]Provider
+		if err := json.Unmarshal(snapshotJSON, &providers); err != nil {
+			snapshotDB = &Database{}
```
docker-agent, line 45: **[MEDIUM] Silent snapshot parse failure degrades 1M-context models to 200k with no warning**

When `json.Unmarshal` fails here, `snapshotDB` is set to `&Database{}` (an empty catalog) and the error is swallowed with no log output at all. Because this PR removes the hardcoded `DefaultClaudeContextLimit` fallback (which returned 1 M tokens for Claude Fable and Opus 4.6/4.7/4.8), a malformed `snapshot.json` — e.g. from a partial `go generate` run — would now silently cause those models to report only 200 k context, triggering incorrect retry clamping without any observable signal for operators.

The comment says the intent is to "degrade to provider not found instead of crashing", which is correct, but a `slog.Warn` (not just silence) would make a bad build artifact discoverable:

```go
if err := json.Unmarshal(snapshotJSON, &providers); err != nil {
    slog.Warn("embedded models.dev snapshot is malformed; context-window limits will be wrong",
        "error", err)
    snapshotDB = &Database{}
    return
}
```

This is low-probability in practice (the snapshot is tested by `TestEmbeddedSnapshotParses`), but the observability gap is new because the hardcoded 1M safety net no longer exists.

docker-agent (commented):

One finding worth addressing before merge. Overall the design is sound — the `authoritative` flag correctly prevents pinning the embedded fallback in `s.db`, and the snapshot/embed wiring is clean. The generator tool, memoization logic, and test coverage look correct.

melmennaoui (approved):

PR #1436, opened 2026-01-20
runtime: add session-level observability plumbing

Pnkcaht:
---

## What I did

Introduced session-level observability plumbing to collect and emit high-level interaction and performance metrics during agent execution.

This change adds support for tracking session metrics such as user messages, assistant messages, tool calls, tool errors, execution start/end times, and token usage totals. These metrics are emitted as **ephemeral runtime events** and are **not persisted nor used for billing**, enabling better observability without impacting existing session storage semantics.

The implementation wires metrics collection through the session manager, runtime events, and protocol layer, making the data available to consumers such as the TUI and future UI integrations.

## Related issue

Related to the feature request for observability and interaction summaries:  
Refs #1345

> This PR lays the groundwork for surfacing interaction and performance summaries, but does not yet implement a full UI summary view.

## What’s missing / follow-up work

This PR intentionally focuses on **infrastructure and data flow**. The following items are out of scope and left for follow-up changes:

- TUI summary panel or end-of-session report
- Aggregated success-rate or outcome classification
- Persistence of observability metrics across sessions
- Export to external observability systems (OpenTelemetry, Prometheus, etc.)
- UI/UX decisions around how and when summaries should be displayed

These can now be implemented incrementally on top of the metrics exposed here.

### Observability Architecture Overview

The provided diagram illustrates the new session-level observability plumbing. Key highlights of the flow include:

- Lifecycle Tracking: Upon triggering execution, the Session Manager initializes session metrics (marking StartedAt) and manages the execution lifecycle.

- Real-time Collection: A dedicated Goroutine monitors the event stream from the LLM Provider. It intercepts assistant messages, tool calls, and tool errors to increment performance counters in real-time.

- Ephemeral Events: These metrics are emitted as structured Runtime Events via the Protocol Layer (Protobuf) for consumption by the TUI/UI.

- Zero Persistence: While the core session state is saved to the Bank (database), the metrics are explicitly excluded from JSON serialization (json:"-"), ensuring they remain ephemeral and do not impact storage or billing.

- Final Handshake: At the end of the execution, EndedAt is recorded, and a final TokenUsageEvent is emitted to provide a complete summary of the interaction.

image

## Before

- No unified mechanism to observe session-level interaction metrics
- Token usage was available, but interaction context was fragmented
- No clear extension point for summaries or dashboards

## After

- Session-level interaction and performance metrics are collected consistently
- Metrics are emitted as structured runtime events
- Consumers (TUI / future UI) can safely display or aggregate observability data
[...]

Issue #1345, opened 2026-01-11
Stats on the usage along with performance + interaction summary ->success rate, tool calls etc


Raveendiran-RR:
It would be nice to have a summary of the the performance and interaction like in the screenshot

Requested feature
<img width="1200" height="578" alt="Image" src="https://github.com/user-attachments/assets/2c0c3d45-dee6-4d0d-ba17-66dabaaaec84" />

Currently available 

<img width="427" height="412" alt="Image" src="https://github.com/user-attachments/assets/9ddb8494-92f0-456d-8690-f99ffd0c730b" />

**Why you'd like to see it**

Tell us why it's important for you.
- This helps in improved observability metrics
- With this observability metrics, cagent would be more production ready
- Once we find significant success, UI or Extension for Cagent will be an awesome thing to have

**Workarounds?**

Are you using any workarounds at the moment? If so, tell us about them
None as of now.

**Additional context**

Any other info you consider useful can be included here
None as of now

ghost:
Hi @Raveendiran-RR , I'll take a look and send a pull request about it.

dgageot:
I'm going to close this issue for now. We have a /cost command that prints a lot of this

PR #3278, opened 2026-06-26
feat(cli): print session summary at end of exec runs

Sayt-0:
## Summary

Adds an end-of-run session summary for non-interactive (`--exec`) runs. Until now `--exec` printed no usage recap, while the TUI only exposed the on-demand `/cost` dialog. The recap covers interaction stats (tool calls, success rate), per-model token usage, total cost, and cache savings, matching the request in #1345.

The data is derived from the existing session through a new `Session.Stats()` helper, so no new tracking pipeline or protocol events are introduced.

## Sample output

```
─── Session summary ───
Session:       7532d5a9-92e9-4197-ae26-34855d7167f7
Duration:      1s
Requests:      2
Tool calls:    1 (1 ok, 0 failed)
Success rate:  100.0%

Model usage
MODEL           REQS   INPUT   CACHE READ   OUTPUT
openai/gpt-4o      2   1,600            0       20

Tokens:  1,600 in / 20 out
Cost:    $0.0042
```

## Mapping to the request (#1345)

| Requested field | Status |
|---|---|
| Session ID | done |
| Tool calls with success/failure split | done |
| Success rate | done |
| Per-model usage (requests, input, cache reads, output) | done |
| Token totals and cache savings | done |
| Wall-clock duration | done |
| Agent-active / API-time / tool-time split | deferred (needs runtime timing instrumentation) |
| User Agreement | dropped (ambiguous metric) |

## Design notes

- The recap is written to **stderr**, so stdout stays machine-consumable. The existing exec stdout contract (asserted by the e2e `TestExec_*` tests) is unchanged.
- It is skipped in JSON output mode, where structured events are emitted instead.
- It is a no-op when the session recorded no measurable activity.
- No observability stack is added. General observability remains the role of OTEL, consistent with the feedback on the earlier closed PR #1436.

## Validation

| Check | Result |
|---|---|
| `task build` | pass |
| `golangci-lint` on `pkg/session`, `pkg/cli` | 0 issues |
| Unit tests (`pkg/session`, `pkg/cli`) | pass |
| e2e (`TestExec_*`) | pass (stdout unchanged) |
| Manual real run (replay cassette, plain and tool-call) | recap on stderr, stdout clean |

## Out of scope / possible follow-ups

- API-time vs tool-time split (requires lightweight runtime timing).
- A flag to suppress the recap (currently it can be silenced with `2>/dev/null`).
- Surfacing the same stats in JSON output mode.

### DIFF
--- a/pkg/session/stats.go
+++ b/pkg/session/stats.go
@@ -1,0 +1,190 @@
+package session
+
+import (
+	"cmp"
+	"slices"
+	"time"
+
+	"github.com/docker/docker-agent/pkg/chat"
+)
+
+// Stats aggregates high-level interaction and usage metrics for a session.
+// It is derived data computed by walking the session and its sub-sessions; it
+// is never persisted and carries no billing authority. It backs end-of-run
+// summaries such as the non-interactive (--exec) recap.
+type Stats struct {
+	// ID is the session identifier.
+	ID string
+
+	// Duration is the wall-clock span between the first and last message,
+	// approximated from message timestamps (see Session.Duration).
+	Duration time.Duration
+
+	// Requests is the number of assistant responses that reported usage.
+	Requests int
+
+	// ToolCalls is the number of completed tool calls (tool-role results).
+	ToolCalls int
+
+	// ToolErrors is the number of completed tool calls that failed.
+	ToolErrors int
+
+	// InputTokens counts new (uncached) input tokens.
+	InputTokens int64
+
+	// CachedInput counts input tokens served from cache (cache reads).
+	CachedInput int64
+
+	// CacheWrite counts input tokens written to the cache.
+	CacheWrite int64
+
+	// OutputTokens counts generated output tokens.
+	OutputTokens int64
+
+	// ReasoningTokens counts reasoning tokens (subset of output for some models).
+	ReasoningTokens int64
+
+	// Cost is the total session cost in dollars.
+	Cost float64
+
+	// Models holds the per-model breakdown, sorted by cost descending.
+	Models []ModelStats
+}
+
+// ModelStats holds the per-model usage breakdown within a session.
+type ModelStats struct {
+	Model        string
+	Requests     int
+	InputTokens  int64 // new (uncached) input tokens
+	CachedInput  int64 // cache reads
+	OutputTokens int64
+	Cost         float64
+}
+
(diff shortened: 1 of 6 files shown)


dgageot:
@Sayt-0 add a `--summary` to enable this feature

pkg/session/stats.go
```diff
@@ -0,0 +1,190 @@
...
+
+	modelIndex := make(map[string]int)
+
+	var walk func(sess *Session)
+	walk = func(sess *Session) {
+		for _, item := range sess.snapshotItems() {
+			switch {
+			case item.IsMessage():
+				m := item.Message.Message
+				st.Cost += m.Cost
```
docker-agent, line 113: **[MEDIUM] Potential cost double-counting in `Stats()` remote-mode fallback**

`st.Cost += m.Cost` is accumulated unconditionally for every message during the `walk` (line 113), including assistant messages where `m.Usage == nil`. However, `st.Requests` is only incremented when `m.Usage != nil`. In remote mode, usage is stored in `MessageUsageHistory` rather than on the message object — so it is architecturally possible for an assistant message to carry `m.Cost > 0` while `m.Usage == nil`.

When this happens:
1. The walk adds the message's cost to `st.Cost` (via line 113).
2. After the walk, `st.Requests == 0` → the remote-mode fallback fires.
3. The fallback also adds `r.Cost` for the same call from `MessageUsageHistory`.

The cost is counted twice. The fix is either to skip the message-level cost accumulation when `m.Usage == nil` (since that cost will be covered by the fallback path), or to zero out `st.Cost` before entering the fallback loop:

```go
// Option A — only accumulate m.Cost when usage is present:
case chat.MessageRoleAssistant:
    if m.Usage != nil {
        st.Cost += m.Cost  // move cost inside the usage guard
        st.Requests++
        // ...
    }

// Option B — reset before fallback:
if st.Requests == 0 && len(s.MessageUsageHistory) > 0 {
    st.Cost = 0  // discard costs collected without usage data
    for _, r := range s.MessageUsageHistory { ... }
}
```

docker-agent (commented):

One medium-severity issue found in the new `Session.Stats()` implementation. The overall PR structure is sound — stderr output, JSON mode guard, no-op for empty sessions, and correct token math are all well-implemented.

PR #3279, opened 2026-06-26
refactor: thread context throughout and enforce ContextConnectivity lint rule

dgageot:
This branch integrates the `Lint/ContextConnectivity` cop from `rubocop-go` as a whole-program static analysis pass and then resolves every violation it finds. The cop detects places where a `context.Context` is available but a derived or background context is used instead — patterns that silently lose deadlines, cancellation signals, and trace spans.

The bulk of the work is mechanical but intentional: context threading was pushed down into the `Runtime` interface and its implementations, through RAG embedding and chunking helpers, into the SQLite memory backend (now lazily opened with the caller's context), through the telemetry `TrackSynchronous` path, into proxy-allowlist resolution at dial time, and across dozens of test harnesses that previously created detached roots with `context.Background()`. Places where a genuinely detached context is correct — shutdown flushes, background notification playback — are suppressed with `context.WithoutCancel` derivation, which was added to the `rubocop-go` dependency in the same batch.

No behaviour changes are intended. `task --force lint` and `task --force test` both pass.

### DIFF
--- a/pkg/acp/agent.go
+++ b/pkg/acp/agent.go
@@ -126,7 +126,7 @@ func (a *Agent) Initialize(ctx context.Context, params acp.InitializeRequest) (a
 }
 
 // newRuntime creates a new runtime using the default agent.
-func (a *Agent) newRuntime(workingDir string) (runtime.Runtime, *agent.Agent, error) {
+func (a *Agent) newRuntime(ctx context.Context, workingDir string) (runtime.Runtime, *agent.Agent, error) {
 	if a.team == nil {
 		return nil, nil, errors.New("agent not initialized")
 	}
@@ -148,7 +148,7 @@ func (a *Agent) newRuntime(workingDir string) (runtime.Runtime, *agent.Agent, er
 		opts = append(opts, runtime.WithWorkingDir(workingDir))
 	}
 
-	rt, err := runtime.New(a.team, opts...)
+	rt, err := runtime.New(ctx, a.team, opts...)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -202,7 +202,7 @@ func (a *Agent) NewSession(ctx context.Context, params acp.NewSessionRequest) (a
 		return acp.NewSessionResponse{}, err
 	}
 
-	rt, defaultAgent, err := a.newRuntime(workingDir)
+	rt, defaultAgent, err := a.newRuntime(ctx, workingDir)
 	if err != nil {
 		return acp.NewSessionResponse{}, err
 	}
@@ -332,7 +332,7 @@ func (a *Agent) ResumeSession(ctx context.Context, params acp.ResumeSessionReque
 		return acp.ResumeSessionResponse{}, err
 	}
 
-	rt, _, err := a.newRuntime(sess.WorkingDir)
+	rt, _, err := a.newRuntime(ctx, sess.WorkingDir)
 	if err != nil {
 		return acp.ResumeSessionResponse{}, err
 	}
--- a/pkg/httpclient/ssrf.go
+++ b/pkg/httpclient/ssrf.go
@@ -118,20 +118,17 @@ func NewSSRFSafeTransport() *http.Transport {
 			"type", fmt.Sprintf("%T", http.DefaultTransport))
 		t = &http.Transport{Proxy: http.ProxyFromEnvironment}
 	}
-	t.DialContext = func(ctx context.Context, network, address string) (net.Conn, error) {
-		proxies := proxyDialAllowlist(ctx)
-		dialer := &net.Dialer{
-			Timeout:   30 * time.Second,
-			KeepAlive: 30 * time.Second,
-			Control: func(network, address string, c syscall.RawConn) error {
-				if _, ok := proxies[canonicalHostPort(address)]; ok {
-					return nil
-				}
-				return SSRFDialControl(network, address, c)
-			},
-		}
-		return dialer.DialContext(ctx, network, address)
-	}
+	proxies := proxyDialAllowlist()
+	t.DialContext = (&net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+		Control: func(network, address string, c syscall.RawConn) error {
+			if _, ok := proxies[canonicalHostPort(address)]; ok {
+				return nil
+			}
+			return SSRFDialControl(network, address, c)
+		},
+	}).DialContext
 	return t
 }
 
@@ -160,7 +157,9 @@ func canonicalHostPort(address string) string {
 // HTTP_PROXY / HTTPS_PROXY / ALL_PROXY environment variables resolve to.
 // Hostname-based proxies are resolved to all of their A/AAAA records so
 // the comparison can be done against the post-resolution dial address.
-func proxyDialAllowlist(ctx context.Context) map[string]struct{} {
+// The result is a snapshot taken at call time — safe to capture in a
+// dialer's Control closure.
+func proxyDialAllowlist() map[string]struct{} {
 	out := map[string]struct{}{}
 	raw := []string{
 		os.Getenv("HTTP_PROXY"), os.Getenv("http_proxy"),
@@ -168,7 +167,7 @@ func proxyDialAllowlist(ctx context.Context) map[string]struct{} {
 		os.Getenv("ALL_PROXY"), os.Getenv("all_proxy"),
 	}
 	for _, s := range raw {
-		for addr := range proxyHostPorts(ctx, s) {
+		for addr := range proxyHostPorts(s) {
 			out[addr] = struct{}{}
 		}
 	}
@@ -179,7 +178,7 @@ func proxyDialAllowlist(ctx context.Context) map[string]struct{} {
 // accepted by net/http.ProxyFromEnvironment: a full URL or a bare host[:port]
 // in which case http:// is implied) and returns each "ip:port" it can be
 // reached at. A nil/empty input yields an empty set.
-func proxyHostPorts(ctx context.Context, spec string) map[string]struct{} {
+func proxyHostPorts(spec string) map[string]struct{} {
 	if spec == "" {
 		return nil
 	}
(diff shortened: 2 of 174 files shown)


aheritier:

pkg/tools/builtin/shell/askpass.go
```diff
@@ -437,16 +439,16 @@ func (h *shellHandler) askpassActive() bool {
...
 // The mutex makes lazy start safe against concurrent commands and a concurrent
 // stopAskpass.
-func (h *shellHandler) ensureAskpass() *askpassServer {
+func (h *shellHandler) ensureAskpass(ctx context.Context) *askpassServer {
 	h.askpassMu.Lock()
 	defer h.askpassMu.Unlock()
 	if h.askpassStarted {
 		return h.askpass
 	}
 	h.askpassStarted = true
```
docker-agent, line 448: **[HIGH] `ensureAskpass` permanently disables askpass if the first call's context is cancelled**

`h.askpassStarted` is set to `true` before `startAskpassServer` is called, and is never reset on failure. If the context passed to `ensureAskpass` on the first-ever sudo command is cancelled before `lnConfig.Listen` completes, `startAskpassServer` returns an error and `h.askpass` remains `nil`. All subsequent shell commands then receive `nil` from `ensureAskpass` (the guard `if h.askpassStarted { return h.askpass }` short-circuits), permanently disabling the askpass server for the lifetime of the `shellHandler` — sudo commands will run without the password-prompt helper for the rest of the session.

Before this PR, `startAskpassServer` was called with `context.Background()` (uncancellable), so this failure mode did not exist. The fix is to either (a) reset `h.askpassStarted = false` in the error path so the next command retries, or (b) call `startAskpassServer` with a long-lived context (e.g., `context.WithoutCancel(context.Background())` or a handler-lifetime context) rather than the per-request context.

pkg/httpclient/ssrf.go
```diff
@@ -118,17 +118,20 @@ func NewSSRFSafeTransport() *http.Transport {
...
-		KeepAlive: 30 * time.Second,
-		Control: func(network, address string, c syscall.RawConn) error {
-			if _, ok := proxies[canonicalHostPort(address)]; ok {
-				return nil
-			}
-			return SSRFDialControl(network, address, c)
-		},
-	}).DialContext
+	t.DialContext = func(ctx context.Context, network, address string) (net.Conn, error) {
+		proxies := proxyDialAllowlist(ctx)
```
docker-agent: **[MEDIUM] `proxyDialAllowlist` (potentially blocking DNS) called on every TCP dial**

`proxyDialAllowlist(ctx)` is now called inside the `DialContext` closure, meaning it runs on every individual TCP connection the transport makes — including connection-pool replenishment, retries, and keep-alive re-dials. The function reads 6 environment variables and, for hostname-based proxy specs, calls `proxyHostPorts` which performs a DNS lookup with a 5-second timeout to resolve each proxy hostname to its IP addresses.

Before this PR the allowlist needed a `ctx` to pass to `proxyHostPorts`, which forced the call into the per-dial closure. The fix is to pre-compute the allowlist once at transport-construction time (environment variables and proxy hostnames rarely change at runtime) and capture the result in the closure, or to use `sync.Once` / a short-TTL cache. If env-var changes must be picked up at dial time, a `sync.Once`-per-transport or a cached value with a TTL would bound the DNS overhead.

pkg/acp/agent.go
```diff
@@ -332,7 +332,7 @@ func (a *Agent) ResumeSession(ctx context.Context, params acp.ResumeSessionReque
 		return acp.ResumeSessionResponse{}, err
 	}
 
-	rt, _, err := a.newRuntime(sess.WorkingDir)
+	rt, _, err := a.newRuntime(ctx, sess.WorkingDir)
```
docker-agent, line 335: **[MEDIUM] `ResumeSession` passes request-scoped context to `newRuntime` without `context.WithoutCancel`**

The runtime created here is stored in `a.sessions` (a long-lived map) and is expected to serve the resumed session indefinitely. However, it is constructed with `ctx` — the HTTP handler's request context, which is cancelled as soon as the RPC call returns.

If `runtime.New` / `newRuntime` stores the context for use in background goroutines (e.g. tool list refresh via `emitToolsChanged`, which is now wired through `r.ctx()`), the runtime's background work will be cancelled when the RPC handler returns. The parallel paths `CreateSession` and `StartSession` (line 205) exhibit the same pattern, so this may be intentional, but the PR's stated policy for "genuinely detached" objects is to wrap with `context.WithoutCancel`. Compare with `NewLocalRuntime`, which already wraps: `ctx: func() context.Context { return context.WithoutCancel(ctx) }` — so the runtime itself does the right thing internally. If `newRuntime` calls `NewLocalRuntime` the internal closure will protect background work; the only concern is if any state is derived from the raw `ctx` before that wrapping.

pkg/runtime/runtime.go
```diff
@@ -531,19 +532,20 @@ func WithHooksRegistry(reg *hooks.Registry) Opt {
...
 // This is useful for testing or when persistence is handled externally.
-func NewLocalRuntime(agents *team.Team, opts ...Opt) (*LocalRuntime, error) {
+func NewLocalRuntime(ctx context.Context, agents *team.Team, opts ...Opt) (*LocalRuntime, error) {
 	defaultAgent, err := agents.DefaultAgent()
 	if err != nil {
 		return nil, err
 	}
 
 	r := &LocalRuntime{
+		ctx:                    func() context.Context { return context.WithoutCancel(ctx) },
```
docker-agent, line 548: **[MEDIUM] `context.WithoutCancel` does not strip deadlines — `r.ctx()` inherits the caller's deadline**

Go's `context.WithoutCancel` (Go 1.21+) strips the cancellation signal and cause but explicitly preserves all other context properties, including deadlines. From the Go docs: _"WithoutCancel returns a copy of parent that is not canceled when parent is canceled. The returned context returns no Deadline …"_ — wait, actually it DOES carry the deadline if the parent has one; it only removes the cancel chain.

If `NewLocalRuntime` is ever called with a context that has a deadline (e.g. a request-scoped context with a 30-second timeout), `r.ctx()` will return a context that expires at that same deadline. Background work derived from `r.ctx()` — most notably `emitToolsChanged` which calls `context.WithTimeout(r.ctx(), toolsChangedTimeout)` — will use a base context that may already be expired or nearly expired, silently truncating tool-list refreshes throughout the session.

The fix is to use `context.WithoutCancel(context.Background())` or to derive from a process-lifetime context rather than from the caller's context when constructing the stored background context.

pkg/runtime/runtime.go
```diff
@@ -1673,8 +1681,8 @@ func (r *LocalRuntime) Resume(_ context.Context, req ResumeRequest) {
 // Steer enqueues a user message for urgent mid-turn injection into the
 // running agent loop. The message will be picked up after the current batch
 // of tool calls finishes but before the loop checks whether to stop.
-func (r *LocalRuntime) Steer(msg QueuedMessage) error {
-	if !r.steerQueue.Enqueue(context.Background(), msg) {
+func (r *LocalRuntime) Steer(ctx context.Context, msg QueuedMessage) error {
+	if !r.steerQueue.Enqueue(ctx, msg) {
```
docker-agent, line 1693: **[MEDIUM] Context cancellation silently misreported as "queue full" in `Steer`/`FollowUp`**

Before this PR, `Enqueue` was called with `context.Background()` (uncancellable), so `Enqueue` returning `false` always meant the queue was actually full. Now the caller's `ctx` is threaded in. If `MessageQueue.Enqueue` blocks waiting for capacity and respects context cancellation (returning `false` on `ctx.Done()`), a cancelled context will cause the error message `"steer queue full"` or `"follow-up queue full"` to be returned — masking the real cause (cancellation) as a capacity error.

This matters for callers that inspect the error or log it for debugging: a cancelled follow-up (e.g. from a request that timed out) will look like a capacity problem in logs, making operational diagnosis harder. If `Enqueue` is truly non-blocking (returns immediately if the channel buffer is full), the concern only applies to concurrent cancellation racing the send, which is less likely. Consider returning `ctx.Err()` when the context is done, or documenting that cancellation maps to the same "full" error.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR threads `context.Context` throughout the codebase — a well-structured refactor that correctly uses `context.WithoutCancel` for detached lifecycle objects. One confirmed high-severity regression and three medium-severity issues were found in the newly introduced `+` lines. The bulk of the mechanical context threading (test harnesses, logging, RAG pipeline) looks correct.

dgageot:
Rebased on top of `origin/main` (conflicts resolved in `pkg/tui/handlers.go`, keeping both the new timing instrumentation and the threaded `m.ctx()`). Addressed the review feedback as follows:

**Fixed (valid):**
- **[HIGH] askpass permanently disabled after a cancelled first command** — `ensureAskpass` now resets `askpassStarted = false` in the error path so a later command retries instead of running the rest of the session without the sudo password helper.
- **[MEDIUM] proxy allowlist DNS lookups on every dial** — reverted to resolving the allowlist once at transport-construction time rather than inside `DialContext`. This avoids repeating blocking DNS lookups on connection-pool refills, retries and keep-alive re-dials. The lone `context.Background()` root is annotated for the `ContextConnectivity` cop.

**Not changed (not valid):**
- **`ResumeSession` passes request ctx to `newRuntime`** — `NewLocalRuntime` already wraps with `context.WithoutCancel` internally and the raw `ctx` is only used synchronously during construction; identical to the existing `NewSession`/`CreateSession` paths.
- **`context.WithoutCancel` inherits the deadline** — the comment is self-contradictory; empirically (Go 1.26) `WithoutCancel` *strips* the deadline (`Deadline()` returns `ok=false`), so `r.ctx()` has no deadline to truncate background work.
- **Cancellation misreported as "queue full"** — the default `inMemoryMessageQueue.Enqueue` ignores the context and is strictly non-blocking, so a cancelled context can never be reported as a capacity error. No behaviour change from before.

`task build`, `task lint` (incl. the `ContextConnectivity` cop) and `task test` are all green.

aheritier:

dgageot:
Rebased again on top of the latest `origin/main` and resolved the new conflict in `pkg/tui/page/chat/queue_test.go` (kept the threaded `t.Context()` / `app.New(ctx, …)` signatures).

The rebase auto-merge silently dropped the `context.Context` parameter from `app.New` while keeping its `ctx`-using body, so I restored the signature and re-threaded the caller's context through every call site (CLI, examples, and tests). I also annotated the build-time root context in the new `pkg/modelsdev/internal/gen` snapshot generator (landed from `main`) for the `ContextConnectivity` cop — it's a standalone `main` with no parent context.

The earlier review feedback remains addressed as before:

**Fixed (valid):**
- **[HIGH] askpass permanently disabled after a cancelled first command** — `ensureAskpass` resets `askpassStarted = false` in the error path.
- **[MEDIUM] proxy allowlist DNS lookups on every dial** — resolved once at transport-construction time; the lone `context.Background()` root is annotated for the cop.

**Not changed (not valid):** the `ResumeSession`, `WithoutCancel`-deadline, and "queue full" comments — same rationale as my previous reply (verified again: `NewLocalRuntime` wraps with `WithoutCancel` internally, `WithoutCancel` strips the deadline, and `inMemoryMessageQueue.Enqueue` ignores ctx and is non-blocking).

`task build`, `task lint` (incl. the `ContextConnectivity` cop) and `task test` are all green.

gtardif (approved):

PR #3281, opened 2026-06-27
telemetry: don't record error on chat span after successful LLM completion

jedp-docker:
## Summary

- `handleStream` returns the moment it sees a terminal `finish_reason` (a deliberate latency optimisation — don't block the next turn on trailing `message_stop`/EOF). The deferred `stream.Close()` then closes the HTTP/2 body while the reader goroutine is still blocked in `Recv()`, causing that read to return `"http2: response body closed"` (or `"context canceled"` if the per-attempt `streamCancel` fires first).
- `instrumentedStream` was recording this transport-level teardown as a span error even when the completion had fully succeeded — every affected span also carries `gen_ai.response.finish_reasons=["stop"]` and full token usage, confirming the response was complete.
- This inflates the `docker-agent` error rate on every LLM call. Observed consistently across all `harbor_watch.investigation` traces in Tempo.

**Fix:** add `finishReasonSeen bool` to `instrumentedStream` (set-once on the single-consumer `Recv` path). Once a terminal finish reason is delivered, skip `RecordError` for subsequent transport errors — they're benign teardown, not failures. `endOnce` is still called so the span closes.

## Test plan

- [ ] Verify existing `genai` package tests pass: `go test ./pkg/telemetry/genai/...`
- [ ] In Tempo: after deploy, confirm `chat claude-*` spans with `finish_reasons=stop` no longer appear as `STATUS_CODE_ERROR`
- [ ] Confirm spans that fail *before* a finish reason (genuine mid-stream errors) still record errors

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pkg/telemetry/genai/stream.go
+++ b/pkg/telemetry/genai/stream.go
@@ -67,6 +67,13 @@ type instrumentedStream struct {
 	ended       bool
 	innerClosed bool
 
+	// finishReasonSeen is set to true the first time Recv delivers a
+	// terminal finish reason. After that, errors from the underlying
+	// transport (e.g. "http2: response body closed" when the consumer
+	// closes the body before draining to io.EOF) are benign teardown,
+	// not real failures, and should not be recorded on the span.
+	finishReasonSeen bool
+
 	// capture buffers the streamed deltas for emission as
 	// `gen_ai.output.messages` on Close. Filled only when content
 	// capture is opted in (`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true`)
@@ -82,14 +89,29 @@ func (s *instrumentedStream) Recv() (chat.MessageStreamResponse, error) {
 	resp, err := s.inner.Recv()
 	if err != nil {
 		// io.EOF is the normal stream terminator and is not an error
-		// for the span's purposes — End handles closing.
-		// For non-EOF errors we end the span here too: callers that
-		// abandon the stream after an error (a common pattern for
-		// network failures) would otherwise leak the span and skip the
-		// duration metric. Close remains idempotent so the canonical
-		// `defer Close()` path still works.
+		// for the span's purposes — End handles closing. For other
+		// errors we end the span here too so callers that abandon the
+		// stream on failure don't leak the span or skip the duration
+		// metric; Close remains idempotent so the canonical
+		// `defer Close()` path still works. We skip RecordError when a
+		// terminal finish reason was already delivered: the completion
+		// succeeded, and the subsequent transport error (typically
+		// "http2: response body closed") is benign teardown from the
+		// consumer closing the body before draining to io.EOF.
+		//
+		// Some providers pack a finish reason and a transport error into
+		// the same Recv call. Scan resp before checking finishReasonSeen
+		// so that case is also treated as a successful completion.
+		for i := range resp.Choices {
+			if resp.Choices[i].FinishReason != "" {
+				s.span.AddFinishReason(string(resp.Choices[i].FinishReason))
+				s.finishReasonSeen = true
+			}
+		}
 		if !errors.Is(err, io.EOF) {
-			s.span.RecordError(err, ClassifyError(err))
+			if !s.finishReasonSeen {
+				s.span.RecordError(err, ClassifyError(err))
+			}
 			s.endOnce()
 		}
 		return resp, err
@@ -111,6 +133,7 @@ func (s *instrumentedStream) Recv() (chat.MessageStreamResponse, error) {
 	for i := range resp.Choices {
 		if resp.Choices[i].FinishReason != "" {
 			s.span.AddFinishReason(string(resp.Choices[i].FinishReason))
+			s.finishReasonSeen = true
 		}
 	}
 	if resp.Usage != nil {
--- a/pkg/telemetry/genai/genai_test.go
+++ b/pkg/telemetry/genai/genai_test.go
@@ -8,6 +8,10 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/codes"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
 
 	"github.com/docker/docker-agent/pkg/chat"
 )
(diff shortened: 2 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is correct and well-scoped. The `finishReasonSeen` flag is set on the single-consumer `Recv` path — no concurrency risk — and the teardown-suppression logic correctly gates only `RecordError`, leaving `endOnce()` unconditional so spans always close. All terminal finish reasons are covered (the flag trips on any non-empty `FinishReason`, not just `"stop"`). The explanatory comment accurately describes the transport-teardown scenario and the intended invariant.

No bugs found in the introduced code.

pkg/telemetry/genai/stream.go
```diff
@@ -82,14 +89,19 @@ func (s *instrumentedStream) Recv() (chat.MessageStreamResponse, error) {
...
+		// stream on failure don't leak the span or skip the duration
+		// metric; Close remains idempotent so the canonical
+		// `defer Close()` path still works. We skip RecordError when a
+		// terminal finish reason was already delivered: the completion
+		// succeeded, and the subsequent transport error (typically
+		// "http2: response body closed") is benign teardown from the
+		// consumer closing the body before draining to io.EOF.
 		if !errors.Is(err, io.EOF) {
-			s.span.RecordError(err, ClassifyError(err))
+			if !s.finishReasonSeen {
```
docker-agent, line 112: **[HIGH] Data race on `finishReasonSeen`: read/write without synchronization while concurrent `Close`+`Recv` is an intended usage pattern**

`finishReasonSeen` is read at line 102 (`if !s.finishReasonSeen`) and written at line 126 (`s.finishReasonSeen = true`) with no lock held, yet the existing `mu` comment explicitly says it protects against "a Recv that errors concurrently with the consumer's Close". The typical teardown path this PR fixes is: consumer goroutine calls `Close()` while another goroutine is blocked in `s.inner.Recv()`. Once the transport error fires, that goroutine reads `finishReasonSeen` at line 102. If a separate `Recv` call is concurrently at line 126 writing `true`, or if the Go memory model doesn't guarantee visibility without synchronization, this is an unsynchronized concurrent read/write — a classic data race detectable by `go test -race`.

All other lifecycle flags (`ended`, `innerClosed`) are already protected by `s.mu`. `finishReasonSeen` should be too. Two clean options:

```go
// Option A: atomic bool (zero overhead, no lock widening)
finishReasonSeen atomic.Bool

// read:
if !s.finishReasonSeen.Load() {
    s.span.RecordError(err, ClassifyError(err))
}
// write:
s.finishReasonSeen.Store(true)
```

```go
// Option B: hold s.mu around both the read and the write
// (read — inside the err != nil branch)
s.mu.Lock()
seen := s.finishReasonSeen
s.mu.Unlock()
if !seen {
    s.span.RecordError(err, ClassifyError(err))
}

// (write — inside the choices loop)
s.mu.Lock()
s.finishReasonSeen = true
s.mu.Unlock()
```

pkg/telemetry/genai/stream.go
```diff
@@ -82,14 +89,19 @@ func (s *instrumentedStream) Recv() (chat.MessageStreamResponse, error) {
 	resp, err := s.inner.Recv()
 	if err != nil {
```
docker-agent, line 90: **[MEDIUM] finish_reason + error in the same `Recv` call bypasses the new suppression logic**

The early-return on `if err != nil` (line 90) means the choices loop at line 123 is never reached when `Recv` returns both a non-nil error and a populated `resp.Choices[0].FinishReason` simultaneously. In that case `finishReasonSeen` stays `false` and the error is recorded on the span, negating the fix.

Most providers (OpenAI, Anthropic) deliver `finish_reason` on a successful chunk and then return EOF on the next call, so this scenario may be rare in practice. But the protocol doesn't prohibit it (e.g., a proxy or gateway could close the connection in the same read that delivers the final chunk), and the guard logic silently fails if it happens.

A defensive approach is to scan `resp.Choices` for a finish reason before checking `finishReasonSeen`:

```go
if err != nil {
    // Check resp even on error — some providers can return a finish
    // reason in the same call that also signals a transport close.
    for i := range resp.Choices {
        if resp.Choices[i].FinishReason != "" {
            s.finishReasonSeen = true // (or atomic store)
            break
        }
    }
    if !errors.Is(err, io.EOF) {
        if !s.finishReasonSeen {
            s.span.RecordError(err, ClassifyError(err))
        }
        s.endOnce()
    }
    return resp, err
}
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

The fix correctly identifies and addresses the false-positive span error problem. However, the new `finishReasonSeen` field introduces a data race — it is read and written on the hot path without the `mu` lock that already guards the other lifecycle flags (`ended`, `innerClosed`). A secondary edge case exists where a provider returns a finish reason and a transport error in the same `Recv` call, which the current guard logic misses.

pkg/telemetry/genai/stream.go
jedp-docker: finishReasonSeen has no data race in practice. It is read and written exclusively inside Recv(), which is called by a single consumer goroutine — the reader loop in handleStream. The goroutine that runs concurrently (Close()/endOnce() via defer stream.Close()) never touches finishReasonSeen; that's precisely why ended and innerClosed need mu and this field doesn't. This matches the existing unlocked mutations in Recv() — span.MarkChunk(), SetResponseID(), AddFinishReason(), RecordUsage() — which are all single-consumer for the same reason. Adding mu to only finishReasonSeen would be inconsistent with that established pattern.

pkg/telemetry/genai/stream.go
jedp-docker: Updated in the latest push: the error branch now scans resp.Choices for a finish reason before checking finishReasonSeen, so a provider that packs both into a single Recv call is handled correctly. A third test (TestWrapStream_FinishReasonAndErrorSameRecv) covers this case.

docker-agent: Perfect, thanks for the thorough fix! The error branch now correctly scans `resp.Choices` before consulting `finishReasonSeen` (lines 106–111), so the edge case where a provider packs both finish_reason and a transport error into the same call is handled defensively. The test coverage for this scenario is solid too — `errorStream.finishWithError` mode delivers both in a single `Recv`, and the assertion confirms the span status stays `Unset` rather than erroring.

This resolves the issue. ✅

dgageot (approved):

PR #3282, opened 2026-06-27
session: add generate_title opt-out for headless callers

jedp-docker:
## Summary

- Programmatic / headless clients (e.g. Harbor Watch) create short-lived sessions that are deleted after use and whose title is never displayed in any UI.
- The background LLM call that generates a session title on first message is pure waste for these callers: ~16% of calls fail with `context canceled` (the turn ends before the ~1s Haiku call completes), producing false error spans on `sessiontitle.generate`; the ~84% that succeed write a title that is read by nobody and deleted seconds later.
- There is currently no way for a caller to opt out of title generation.

**Fix:** add `GenerateTitle *bool` to `Session` (JSON: `generate_title`, `omitempty`). `nil` and `true` preserve the existing default (generate). `false` skips title generation for that session. The gate is a one-line addition to `needsTitle` in `RunSession` — no other codepath is affected. Extra JSON fields are silently ignored by existing deployed versions, so clients can send `generate_title: false` before this ships without breaking anything.

## Test plan

- [ ] `go test ./pkg/server/...` passes
- [ ] Create a session with `{"generate_title": false}`, run a turn, confirm no `sessiontitle.generate` span appears in Tempo
- [ ] Create a session without the field, confirm title generation still works as before

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -371,6 +371,7 @@ func (sm *SessionManager) CreateSession(ctx context.Context, sessionTemplate *se
 	if len(sessionTemplate.CustomModelsUsed) > 0 {
 		sess.CustomModelsUsed = append([]string(nil), sessionTemplate.CustomModelsUsed...)
 	}
+	sess.DisableTitle = sessionTemplate.DisableTitle
 
 	return sess, sm.sessionStore.AddSession(ctx, sess)
 }
@@ -657,7 +658,8 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 	// avoid a data race with UpdateSessionTitle, which takes sm.mux and
 	// writes to sess.Title concurrently with the goroutine's read.
 	titleToEmit := sess.Title
-	needsTitle := titleToEmit == "" && len(userMessages) > 0 && titleGen != nil
+	needsTitle := titleToEmit == "" && len(userMessages) > 0 && titleGen != nil &&
+		!sess.DisableTitle
 
 	go func() {
 		// Defers run LIFO: close(streamChan) last, so by the time the
@@ -1299,6 +1301,7 @@ func (sm *SessionManager) SetSessionAgentModel(ctx context.Context, sessionID, m
 		OutputTokens:            sess.OutputTokens,
 		Cost:                    sess.Cost,
 		Starred:                 sess.Starred,
+		DisableTitle:            sess.DisableTitle,
 	}
 
 	// Clone the maps/slices under sm.mux to avoid data races
--- a/pkg/server/session_manager_test.go
+++ b/pkg/server/session_manager_test.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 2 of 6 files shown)


pkg/server/session_manager.go
```diff
@@ -657,7 +657,8 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 	// avoid a data race with UpdateSessionTitle, which takes sm.mux and
 	// writes to sess.Title concurrently with the goroutine's read.
 	titleToEmit := sess.Title
-	needsTitle := titleToEmit == "" && len(userMessages) > 0 && titleGen != nil
+	needsTitle := titleToEmit == "" && len(userMessages) > 0 && titleGen != nil &&
+		(sess.GenerateTitle == nil || *sess.GenerateTitle)
```
docker-agent: **[MEDIUM] `GenerateTitle` is not propagated from the session template in `CreateSession`, silently discarding the opt-out**

The guard added here correctly skips title generation when `sess.GenerateTitle` is `false`. However, `CreateSession` — which is the function that turns an API request body into a persisted `Session` — does not copy `GenerateTitle` from `sessionTemplate` to the new session. As a result, a headless caller that POSTs `{"generate_title": false}` will receive a session whose `GenerateTitle` field is `nil`, and `RunSession` will see `nil`, treat it as "generate", and fire the title-generation call anyway. The opt-out feature is effectively a no-op for all API callers.

Fix: after `sess := session.New(opts...)` in `CreateSession`, add:
```go
if sessionTemplate.GenerateTitle != nil {
    sess.GenerateTitle = sessionTemplate.GenerateTitle
}
```
(or introduce a `WithGenerateTitle` opt, consistent with the existing pattern for other fields)

docker-agent (commented):

The `needsTitle` guard itself is correct — the nil/false/true pointer semantics are sound and there is no nil-dereference risk. JSON serialization (`omitempty` on `*bool`) is also correct. However, the feature does not work end-to-end because `CreateSession` does not propagate the new field from the session template.

**Secondary note (low):** `SetSessionAgentModel` contains a hand-rolled field-by-field `Session` clone that now omits `GenerateTitle`. If a headless session's model is changed via that path, the `GenerateTitle` preference could be silently dropped on the next store write.

pkg/server/session_manager.go
jedp-docker: Inverted the flag to DisableTitle.

docker-agent: Great solution! Inverting to `DisableTitle` fixes the propagation issue cleanly: Go's zero value for `bool` is `false`, so omitting the field now correctly defaults to "generate title", and an explicit `{"disable_title": true}` in the request body propagates through `CreateSession` (line 331) and is correctly honored at line 661.

The original `GenerateTitle *bool` pattern would have required three-state logic (`nil` / `true` / `false`) everywhere; this is much simpler.

jedp-docker:
Superseded by #3310, which solves the same problem by wiring up the existing `title` field on `POST /api/sessions` — a pre-supplied title causes `needsTitle` to short-circuit, skipping LLM generation entirely. Closing in favour of that approach.

PR #3284, opened 2026-06-27
feat(lint): wire whole-program ContextConnectivity cop

dgageot:


pkg/runtime/remote_runtime.go
```diff
@@ -101,14 +101,22 @@ func (r *RemoteRuntime) resolvedAgent(ctx context.Context) (string, latest.Agent
...
-func (r *RemoteRuntime) CurrentAgentName() string {
+func (r *RemoteRuntime) CurrentAgentName(ctx context.Context) string {
 	if r.currentAgent != "" {
 		return r.currentAgent
 	}
-	r.resolvedDefaultOnce.Do(func() {
-		r.resolvedDefault, _ = r.resolvedAgent(context.Background())
-	})
-	return r.resolvedDefault
+	r.resolvedDefaultMu.Lock()
```
docker-agent, line 108: **[MEDIUM] Mutex held across blocking network call; failed lookups never cached**

This PR replaces `sync.Once` with an explicit mutex (`resolvedDefaultMu`) for caching the remote agent name, but the mutex is held while `resolvedAgent` performs a blocking network call (`r.client.GetAgent`). This means any concurrent caller waiting on the lock is blocked for the full duration of the network round-trip — potentially several seconds under poor network conditions.

Additionally, if the network call returns an empty name (i.e., fails to resolve), `r.resolvedDefault` stays `""` and every subsequent call re-issues the full network request while holding the lock — the once-semantics are lost for the error path.

With `sync.Once`, goroutines block only until the single inflight call completes, and the result (success or failure) is cached forever. Suggested fix: either keep `sync.Once` for the first lookup or check-and-set `resolvedDefault` outside the critical section, using a separate flag to distinguish "not yet tried" from "tried and empty".

pkg/runtime/runtime.go
```diff
@@ -531,19 +532,20 @@ func WithHooksRegistry(reg *hooks.Registry) Opt {
...
 // This is useful for testing or when persistence is handled externally.
-func NewLocalRuntime(agents *team.Team, opts ...Opt) (*LocalRuntime, error) {
+func NewLocalRuntime(ctx context.Context, agents *team.Team, opts ...Opt) (*LocalRuntime, error) {
 	defaultAgent, err := agents.DefaultAgent()
 	if err != nil {
 		return nil, err
 	}
 
 	r := &LocalRuntime{
+		ctx:                    func() context.Context { return context.WithoutCancel(ctx) },
```
docker-agent, line 548: **[MEDIUM] context.WithoutCancel inherits construction-time deadline into long-lived background operations**

`context.WithoutCancel` strips the cancellation signal but **preserves the parent's deadline**. By storing `func() context.Context { return context.WithoutCancel(ctx) }` at construction time, every call to `r.ctx()` returns a context whose deadline is the deadline of whatever `ctx` was passed to `New()`.

If the caller passes a short-lived context (e.g., an HTTP request context with a 30 s timeout, or a test context), the stored closure will start returning expired contexts before the runtime is done with background work. Operations like `emitToolsChanged` that call `context.WithTimeout(r.ctx(), toolsChangedTimeout)` will observe a base context whose deadline has already passed, causing the timeout to fire instantly.

The previous code used `context.Background()`, which has no deadline. Suggested fix: if the intent is only to strip cancellation while inheriting values (trace spans, logger), use `context.WithoutCancel(ctx)` eagerly at construction and store the resulting context value rather than a closure — or create a fresh `context.Background()`-rooted context and copy over the relevant values.

pkg/runtime/session_compaction.go
```diff
@@ -253,7 +253,7 @@ func providerContextLimit(p provider.Provider) int64 {
 // which avoids creating an import cycle on [pkg/runtime].
 func (r *LocalRuntime) runCompactionAgent(ctx context.Context, a *agent.Agent, sess *session.Session) error {
 	t := team.New(team.WithAgents(a))
-	rt, err := New(t, WithSessionCompaction(false))
+	rt, err := New(ctx, t, WithSessionCompaction(false))
```
docker-agent, line 256: **[MEDIUM] Compaction runtime inherits session context deadline via context.WithoutCancel**

`New(ctx, t, WithSessionCompaction(false))` now stores `context.WithoutCancel(ctx)` inside the freshly created compaction `LocalRuntime`. `context.WithoutCancel` strips the cancellation signal but **preserves the parent's deadline**.

If the session `ctx` carries a deadline (e.g., imposed by an outer HTTP handler or session lifecycle timeout), the compaction runtime's background context inherits that deadline. Background work inside the compaction runtime — tool-change notifications, OAuth flows, etc. — may then fail with `context deadline exceeded` mid-run, even though the compaction itself has not timed out.

Previously `context.Background()` was used, which has no deadline. Suggested fix: call `context.WithoutCancel(ctx)` on the ctx before passing it to `New()` (at the `runCompactionAgent` call site rather than inside `New`), or strip the deadline explicitly — though Go's standard library has no built-in for that, so a `context.Background()`-rooted context with only the values copied is the safest approach.

docker-agent (commented):

This PR systematically wires the `ContextConnectivity` linter cop across the codebase, replacing `context.Background()`/`context.TODO()` with caller-supplied contexts and adding `context.WithoutCancel()` for background operations. The approach is generally sound, but three medium-severity issues were found in the context propagation changes.

PR #3287, opened 2026-06-27
lint: add WrapErrors cop

dgageot:
`errorlint` catches places where a wrapped error is compared or type-asserted without `errors.Is`/`errors.As`, but it does not check the formatting site itself. Code like `fmt.Errorf("failed: %v", err)` passes `errorlint` cleanly even though it discards the error chain. The `WrapErrors` cop closes that gap.

The cop walks every `fmt.Errorf` call in the codebase and reports any call that passes an `error`-typed argument without a `%w` verb in the format string. It is intentionally narrow: it only flags arguments whose type is the built-in `error` interface (so struct fields named `Error` or string-typed values are not caught), and it skips calls that already contain a `%w` verb. A follow-up fix tightens the verb check so that the `%%w` escape sequence — a literal percent sign followed by `w` — is not mistaken for a real wrap verb.

The codebase already wraps real errors correctly, so the cop lands as a zero-noise guardrail: 1310 files analyzed, no offenses.

### DIFF
--- a/lint/wrap_errors.go
+++ b/lint/wrap_errors.go
@@ -5,7 +5,6 @@ import (
 	"go/token"
 	"go/types"
 	"strconv"
-	"strings"
 
 	"github.com/dgageot/rubocop-go/cop"
 )
@@ -45,7 +44,7 @@ var WrapErrors = &cop.Func{
 			if !ok {
 				return
 			}
-			if strings.Contains(format, "%w") {
+			if hasWrapVerb(format) {
 				return // already wrapping at least one error
 			}
 			for _, arg := range call.Args[1:] {
@@ -56,6 +55,26 @@ var WrapErrors = &cop.Func{
 			}
 		})
 	},
+}
+
+// hasWrapVerb reports whether format contains a real %w verb, ignoring
+// escaped percent signs. A naive strings.Contains(format, "%w") would treat
+// the literal "%%w" (an escaped percent followed by the letter w) as a
+// wrapping verb and silence a genuinely unwrapped error.
+func hasWrapVerb(format string) bool {
+	for i := 0; i < len(format); i++ {
+		if format[i] != '%' || i+1 >= len(format) {
+			continue
+		}
+		if format[i+1] == '%' {
+			i++ // consume the escaped percent so "%%w" is not read as a verb
+			continue
+		}
+		if format[i+1] == 'w' {
+			return true
+		}
+	}
+	return false
 }
 
 // stringLit returns the unquoted value of a string-literal expression.
--- a/lint/wrap_errors.go
+++ b/lint/wrap_errors.go
@@ -61,6 +61,10 @@ var WrapErrors = &cop.Func{
 // escaped percent signs. A naive strings.Contains(format, "%w") would treat
 // the literal "%%w" (an escaped percent followed by the letter w) as a
 // wrapping verb and silence a genuinely unwrapped error.
+//
+// Flags, width, precision, and argument indices between the percent and the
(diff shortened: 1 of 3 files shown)


lint/wrap_errors.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+		})
+	},
+}
+
+// hasWrapVerb reports whether format contains a real %w verb, ignoring
+// escaped percent signs. A naive strings.Contains(format, "%w") would treat
+// the literal "%%w" (an escaped percent followed by the letter w) as a
+// wrapping verb and silence a genuinely unwrapped error.
+func hasWrapVerb(format string) bool {
```
docker-agent, line 68: **[LOW] hasWrapVerb does not recognise `%w` with format flags, width, precision, or argument indices**

The current implementation only matches the two-character sequence `%w` (after skipping `%%` escapes). Go's `fmt` package accepts verbs with modifiers such as `%-10w`, `%[1]w`, `%3.2w`, etc. If a caller ever writes one of these forms, `hasWrapVerb` returns `false`, the cop treats the call as unwrapped, and flags a false positive — or, conversely, fails to skip a genuinely-wrapped call.

In practice `%w` with flags is rare (and `go vet` would likely warn about unusual modifiers), so the real-world impact is minimal. A more robust scan would skip any characters between `%` and the final verb letter before checking for `w`. For example:

```go
// skip optional flags, width, precision, and [n] argument index
for i < len(format) && !isAlpha(format[i]) {
    i++
}
if i < len(format) && format[i] == 'w' {
    return true
}
```

This is a low-priority hardening item, not a correctness blocker for the common case.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The WrapErrors cop is correctly implemented. The core logic — detecting `fmt.Errorf` calls that pass an `error`-typed argument without `%w` — works as intended: `isErrorType` correctly type-asserts to `*types.Named` (which is how Go's `go/types` represents the built-in `error` interface), and the `%%w` escape handling in `hasWrapVerb` is sound. Test coverage is good, including the escaped-percent edge case and the deliberate policy of allowing mixed `%w`/`%v` calls when at least one error is wrapped.

One low-priority hardening opportunity is noted inline.

aheritier:

dgageot:
Addressed the inline review feedback: `hasWrapVerb` now skips flags, width, precision, and `[n]` argument indices between the `%` and the verb letter, so forms like `%-10w`, `%[1]w`, and `%3.2w` are correctly recognised as wrap verbs. Added a table-driven test covering these modifier forms.

Also rebased on top of `origin/main` (resolved the cop-list conflict in `lint/main.go`). `task lint` (1316 files, no offenses), `task test`, and `task build` all pass.

Sayt-0 (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The `WrapErrors` cop implementation is clean and correct. The drafter raised one high-severity hypothesis — that `isErrorType` would always return `false` because `error` is not a `*types.Named` — but this was **dismissed** by the verifier: in Go's `go/types`, the predeclared `error` type _is_ a `*types.Named` (declared in the universe scope with `Obj().Pkg() == nil`), so the type assertion and subsequent `Obj().Name() == "error"` + `Obj().Pkg() == nil` checks are a valid, if unconventional, way to identify the built-in error interface. No bugs were found in the changed code.

**Coverage**: `lint/main.go`, `lint/wrap_errors.go`, `lint/wrap_errors_test.go` — 223 diff lines reviewed, all findings verified.

PR #3288, opened 2026-06-27
lint: enforce deferred mutex unlocks

dgageot:
Scattered across the codebase were terminal `mutex.Unlock()` calls that weren't deferred — written by hand as the last statement before a return, or factored inconsistently. While functionally correct today, this style is fragile: any future edit that adds an early return or a new code path silently loses the unlock.

This change converts those terminal unlocks to `defer mu.Unlock()` immediately after the corresponding `Lock()`, making the critical-section boundary explicit and safe by construction. It also introduces a new `DeferMutexUnlock` lint cop in `lint/` that detects the pattern automatically, so new instances can't sneak in unnoticed.

The cop is deliberately conservative: it ignores short critical sections (≤3 statements) where deferring would obscure rather than clarify, skips cases where the mutex is relocked inside the same block, avoids flagging defers that already exist for the same lock, and does not fire when intervening side-effectful returns would change observable behaviour. `task lint` and `task test` both pass cleanly.

### DIFF
--- a/lint/defer_mutex_unlock.go
+++ b/lint/defer_mutex_unlock.go
@@ -1,0 +1,190 @@
+package main
+
+import (
+	"go/ast"
+	"go/token"
+
+	"github.com/dgageot/rubocop-go/cop"
+)
+
+// DeferMutexUnlock catches manual mutex unlocks that are safe and clearer as a
+// defer immediately after the corresponding lock.
+//
+// The cop is deliberately conservative. It ignores intentionally short critical
+// sections (work after unlock), unlock/relock patterns, nested lock scopes, and
+// terminal returns with side-effectful result expressions. Those patterns would
+// change behavior if rewritten to defer because Go evaluates return expressions
+// before running deferred calls.
+var DeferMutexUnlock = &cop.Func{
+	Meta: cop.Meta{
+		Name:        "Lint/DeferMutexUnlock",
+		Description: "use defer immediately after mutex locks when the whole remaining scope is the critical section",
+		Severity:    cop.Warning,
+	},
+	Run: func(p *cop.Pass) {
+		ast.Inspect(p.File, func(n ast.Node) bool {
+			var body *ast.BlockStmt
+			switch fn := n.(type) {
+			case *ast.FuncDecl:
+				body = fn.Body
+			case *ast.FuncLit:
+				body = fn.Body
+			default:
+				return true
+			}
+			checkDeferMutexUnlockBody(p, body)
+			return true
+		})
+	},
+}
+
+func checkDeferMutexUnlockBody(p *cop.Pass, body *ast.BlockStmt) {
+	if body == nil {
+		return
+	}
+	for i, stmt := range body.List {
+		lock, ok := lockCall(stmt)
+		if !ok || lock.recv == "" {
+			continue
+		}
+		if i+1 < len(body.List) && isDeferUnlock(body.List[i+1], lock) {
+			continue
+		}
+		if isSafeManualUnlockAtEnd(body.List, i, lock) {
+			p.Reportf(stmt,
+				"%s.%s() is only released at the end of this scope; use `defer %s.%s()` immediately after locking",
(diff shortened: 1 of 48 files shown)


docker-agent (commented):

The bulk of this PR is a straightforward and correct mechanical refactor — converting terminal  calls to  immediately after each lock. All analysed cases (closures in , -inside-, -case scope) are semantically equivalent to the pre-PR code.

One correctness gap was found in the **new lint cop** itself.

lint/defer_mutex_unlock.go
```diff
@@ -0,0 +1,205 @@
...
+	return sel.Sel.Name == lock.unlockMethod() && selectorReceiver(sel.X) == lock.recv
+}
+
+func isSafeManualUnlockAtEnd(stmts []ast.Stmt, lockIdx int, lock mutexCall) bool {
+	unlockIdx := -1
+	for i := lockIdx + 1; i < len(stmts); i++ {
+		if stmtContainsDefer(stmts[i]) {
+			return false
+		}
+		name, recv, ok := selectorCallStmt(stmts[i])
```
docker-agent, line 130: **[MEDIUM] Lint cop false positive: nested `mu.Unlock()` inside `if`/`for` body is invisible to the safety check**

`isSafeManualUnlockAtEnd` scans the top-level statement list using `selectorCallStmt`, which only matches a bare `*ast.ExprStmt` (i.e. a call that is directly a statement). An unlock call nested inside an `if`-block body, a `for`-body, or a `switch`-case is an `*ast.IfStmt`/`*ast.ForStmt` at the top level — `selectorCallStmt` returns `false` for those, so the nested unlock is completely invisible to the scan.

`stmtContainsDefer` uses `ast.Inspect` recursively but only looks for `*ast.DeferStmt` nodes; it does not detect nested unlock calls.

**Consequence — a concrete false-positive pattern:**
```go
mu.Lock()
if cond {
    mu.Unlock()          // intentional early unlock: releases lock BEFORE expensive work
    expensiveWork()      // must run WITHOUT the lock
    return earlyResult
}
doWork()                 // legitimate critical-section work
mu.Unlock()              // top-level terminal unlock — cop sees this and flags it
```
`isSafeManualUnlockAtEnd` finds Lock at idx 0, an `*ast.IfStmt` at idx 1 (invisible via `selectorCallStmt`), and the terminal `mu.Unlock()` at idx 2 (last position) → returns `true` → cop emits "use `defer mu.Unlock()`". Following that advice causes `expensiveWork()` to run with the lock held in the `cond=true` path — a behavioral regression that can introduce lock contention or, if `expensiveWork()` itself acquires the same mutex, a deadlock.

**Fix:** Before returning `true`, walk each intermediate statement with `ast.Inspect` and check whether any nested `selectorCallStmt` matches the same receiver + unlock method. If a nested unlock is found, return `false`.

aheritier:

dgageot:
Addressed the review feedback and rebased on top of `origin/main`.

**Fix for the lint cop false positive** (nested `mu.Unlock()` invisible to the safety check):

`isSafeManualUnlockAtEnd` only inspected top-level statements via `selectorCallStmt`, so an unlock nested inside an `if`/`for`/`switch` body was invisible. The flagged pattern below would have been incorrectly reported, and following the advice would have run `expensiveWork()` while holding the lock:

```go
mu.Lock()
if cond {
    mu.Unlock()
    expensiveWork()
    return earlyResult
}
doWork()
mu.Unlock()
```

Intermediate compound statements are now walked with `ast.Inspect` (new `stmtContainsLockOp` helper): if any nested lock/unlock of the same receiver is found, the terminal unlock is no longer considered the sole release and the cop stays silent. Verified the false positive is gone while genuine cases are still flagged.

Also resolved the rebase conflict in `lint/main.go` (kept both the new constructor cops from `main` and `DeferMutexUnlock`). `task lint`, `task test`, and `task build` all pass.

gtardif (approved):

PR #3290, opened 2026-06-28
feat(hooks): cap oversized tool result payloads

rumpl:
Install an always-on `tool_response_transform` builtin that caps large tool outputs from Docker Agent's filesystem and shell toolsets. When one of those tool results exceeds either configured limit (more than 2,000 lines or more than 50 KiB / 51,200 bytes), the hook writes the full payload to a per-session temp file and replaces the result with a notice plus a bounded tail of the original output.

The replacement tail is constrained by both limits: last 2,000 lines, capped at 50 KiB.

The same builtin is also wired as a `session_end` hook to clean up the per-session temp directory.

This preserves access to the complete result for debugging while keeping subsequent model calls, persisted sessions, and UI events from carrying huge payloads for the most common high-output built-in toolsets.


pkg/hooks/builtins/limit_large_tool_results.go
```diff
@@ -0,0 +1,108 @@
...
+	}, nil
+}
+
+func writeLargeToolResult(payload string) (string, error) {
+	dir := filepath.Join(os.TempDir(), "docker-agent-tool-results")
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return "", err
+	}
+
+	file, err := os.CreateTemp(dir, "tool-result-*.txt")
```
docker-agent, line 100: **[MEDIUM] Resource leak — temp files for large tool results are never deleted**

`writeLargeToolResult` creates a new file under `os.TempDir()/docker-agent-tool-results` on every oversized tool result, but there is no corresponding cleanup path — no deferred removal, no session-end hook that purges the directory, and no TTL-based eviction. Tool outputs from file-read, search, or shell commands routinely exceed the 128 KiB threshold, so in a long-running agent session this directory can grow without bound and eventually exhaust disk space.

The OS does not guarantee cleanup of files in `os.TempDir()` between runs on Linux (only at reboot for `/tmp` on some configurations; not on macOS or Windows).

**Suggested fix**: register a cleanup via a `session_end` hook (or a `defer os.RemoveAll(dir)` scoped to the session) so files are removed when the session ends, or switch to `os.CreateTemp("", ...)` in the default system temp dir and remove each file after the LLM has processed the turn.

pkg/hooks/builtins/limit_large_tool_results.go
```diff
@@ -0,0 +1,108 @@
...
+		_ = os.Remove(path)
+		return "", closeErr
+	}
+	return path, nil
+}
+
+func tailLargeToolResult(payload string) string {
+	tail := lastLines([]byte(payload), largeToolCallResultTailLines)
+	if len(tail) > largeToolCallResultTailBytes {
+		tail = tail[len(tail)-largeToolCallResultTailBytes:]
```
docker-agent: **[MEDIUM] UTF-8 boundary not respected when byte-slicing the tail**

`tailLargeToolResult` performs a raw byte slice to cap the tail at `largeToolCallResultTailBytes` (32 KiB):

```go
tail = tail[len(tail)-largeToolCallResultTailBytes:]
```

`tail` is a `[]byte` derived from a UTF-8 string. The slice offset is a byte position, not a rune boundary, so the cut can land in the middle of a multi-byte character (e.g. a 3-byte CJK glyph or a 4-byte emoji). Go's `string([]byte)` conversion passes bytes through as-is; it does **not** validate or repair UTF-8. The resulting string is then embedded in a message sent to the LLM API. Some providers enforce strict UTF-8 in their JSON payloads and will reject — or silently mangle — the request.

**Suggested fix**:
```go
import "unicode/utf8"

// Advance past any incomplete leading rune after the byte cut
for len(tail) > 0 && !utf8.RuneStart(tail[0]) {
    tail = tail[1:]
}
```

pkg/hooks/builtins/limit_large_tool_results.go
```diff
@@ -0,0 +1,108 @@
...
+
+	path, err := writeLargeToolResult(payload)
+	if err != nil {
+		slog.WarnContext(ctx, "Failed to write large tool call result to temp file", "error", err)
+		return nil, nil
+	}
+
+	tail := tailLargeToolResult(payload)
+	updated := fmt.Sprintf(
+		"Tool call result was too large (%d bytes; limit %d bytes). The full result is available in a file: %s\n\nShowing the last %d lines (up to %d bytes):\n\n%s",
```
docker-agent, line 62: **[MEDIUM] Full tool payload path exposed to LLM — prompt-injection exfiltration risk**

The message returned to the LLM explicitly includes the local filesystem path of the temp file holding the full, unredacted payload:

```
"The full result is available in a file: <path>"
```

This is intentional by design, but it creates a prompt-injection vector: a malicious tool response could instruct the agent to subsequently call a file-read tool with this path, re-exposing the entire oversized payload that the hook was meant to suppress. The feature's purpose is to **cap** what the LLM sees, but by advertising the file path the hook inadvertently offers a bypass.

If the intent is only to prevent context-window bloat (not to hide sensitive content), the current design is fine and the risk is acceptable. If there are cases where the capped content may be sensitive (e.g. secrets surfacing in a large shell output), consider either omitting the path from the LLM-visible message or introducing a flag to control path disclosure.

docker-agent (commented):

pkg/hooks/builtins/limit_large_tool_results.go
rumpl: Addressed in 5fa2388. Large tool results are now stored under a per-session temp directory (`$TMPDIR/docker-agent-tool-results/<session-id>/`), and the same builtin is also auto-injected on `session_end` to remove that session directory. I added coverage in `TestLimitLargeToolResultsCleansSessionTempDir` to verify the cleanup path.

pkg/hooks/builtins/limit_large_tool_results.go
rumpl: Addressed in 5fa2388. After applying the byte cap, the tail is advanced to a UTF-8 rune boundary before converting back to string, so we do not emit invalid UTF-8. I also added `TestLimitLargeToolResultsPreservesUTF8Tail` for this case. While touching this code I also updated the backward scan to use `slices.Backward`, which fixed the lint warning.

pkg/hooks/builtins/limit_large_tool_results.go
rumpl: Thanks for calling this out. I left the path in the LLM-visible notice intentionally because the requested behavior was to tell the agent/user that the full result is available in a temp file. This feature is meant to cap context/window bloat rather than provide a secrecy boundary; secret scrubbing remains the responsibility of the existing `redact_secrets` hook. I did narrow the lifetime by moving files into per-session temp directories and cleaning them up on `session_end` in 5fa2388.

docker-agent: Thanks for the clarification and the lifecycle improvement in 5fa2388. With the separation of concerns clear (this hook handles bloat, `redact_secrets` handles secrecy), the current design makes sense — exposing the path lets the agent retrieve the full result when it's actually needed, rather than blindly truncating potentially important data.

The per-session temp directory with `session_end` cleanup is a nice touch that tightens the exposure window. This design is intentional and well-reasoned; no further action needed.

pkg/hooks/builtins/limit_large_tool_results.go
docker-agent: Perfect! The fix looks good — trimToRuneStart advances past any incomplete leading bytes after the byte cap, ensuring the tail starts at a valid UTF-8 boundary. The test using repeated 3-byte CJK characters confirms that the result round-trips cleanly through strings.ToValidUTF8.

Nice bonus cleaning up the lint warning with slices.Backward while you were here. Thanks for addressing this!

pkg/hooks/builtins/limit_large_tool_results.go
docker-agent: Excellent — that's a complete fix. The per-session subdirectory scoping prevents session ID collisions, and the auto-injection on `session_end` (line 33-37 in the updated code) ensures cleanup happens reliably without requiring embedders to wire it manually.

The test coverage in `TestLimitLargeToolResultsCleansSessionTempDir` verifies both that files are created under the session path and that they're removed when the session ends — exactly what was missing in the original implementation. Thanks for the thorough fix!

docker-agent (commented):
### Assessment: 🟢 APPROVE

The new `limit_large_tool_results` builtin hook is well-structured. Core logic for line counting, byte-capping, UTF-8 rune boundary alignment, temp-file lifecycle, and hook registration is correct. All medium-severity hypotheses were dismissed after verification — `ApplyAgentDefaults` is called exactly once per agent with a fresh config, so no duplicate hook registration occurs in practice. No high or medium severity bugs were introduced by this PR.

dgageot (approved):

PR #3292, opened 2026-06-29
plan: per-session markdown toolset + exit_plan_mode marker + fix per-turn agent on warm runtimes

trungutt:
Two small changes that go together.

## 1. Simpler plan tools

When the user wants the agent to think before it acts, they ask it to write a plan first. The current plan toolset was built for two agents collaborating on the same shared document — with names, revision numbers, status fields. That's too much for the common case, which is "one agent writes a plan, the user looks at it, another agent runs it".

This PR replaces it with three tools, no configuration needed:

- **`write_plan(content)`** — saves the plan as Markdown for the current chat. One plan per chat.
- **`read_plan()`** — reads it back.
- **`exit_plan_mode()`** — the agent calls this when it thinks the plan is final. It just says "the plan is ready" and stops.

`exit_plan_mode` deliberately does **not** switch agents or pop a confirmation dialog. Whatever is using cagent — a CLI, a chat UI with a "switch to build mode" button, a server that routes the next message automatically — decides what to do next. That way the same tool works for all of them.

Two small extras: a `plan_updated` event so UIs can show the plan as it's written, and old plan files (>30 days) get cleaned up automatically.

## 2. Bug fix: switching agents mid-chat now works

If a host lets the user pick a specific agent for the next message, that pick was only honored on the **first** message. After that the server quietly ignored it and kept running whichever agent was active.

The most visible symptom: a "plan mode" toggle worked once, then was stuck on — toggling it off did nothing.

Now the server checks which agent the host wants on every message and switches when it differs. An empty pick means "the team's default agent".

## Compatibility

The old tool names (`list_plans`, `delete_plan`, `set_plan_status`, …) stay as unused string constants so the existing TUI still builds. The runtime no longer offers them to the model, so no agent can call them. Cleaning up the TUI is a follow-up.


aheritier:

simonferquel:
This unwire the changes we made to make it cross-session able. If it gets done, we are probably not going to leverage it in AP

trungutt:
@simonferquel So I make another PR https://github.com/docker/docker-agent/pull/3305, leaving `plan` untouched

trungutt:
Closed, in favor of https://github.com/docker/docker-agent/pull/3305

PR #3295, opened 2026-06-29
feat: builtin hooks now honor working_dir and env

dgageot:
Of the three hook types (`command`, `builtin`, `model`), only `command` hooks honored the per-hook `working_dir` and `env` fields. The `HookDefinition` struct shares those fields across all hook types, they pass validation, and they're even flagged by the config-expansion linter — but the builtin factory silently discarded the `HandlerEnv` it received and never forwarded an override directory. A user setting `working_dir` on a `builtin` hook got no effect; the same for per-hook env overrides.

Builtin hooks now resolve `working_dir` with the same `hookWorkingDir` logic command hooks use (an absolute path wins outright; a relative path joins onto the executor directory) and repoint `Input.Cwd` before invoking the `BuiltinFunc`. The resolved env is carried on the context via a new `EnvFromContext` accessor. The git builtins read that env and pass it to `exec.Cmd.Env`; a `nil` value leaves `Env` unset so git still inherits `os.Environ()`, preserving prior behavior exactly. The `working_dir` override is applied only when set, so dispatcher-supplied cwd (e.g. the `worktree_create` event) continues to win when no per-hook override is configured. The `model` hook type still ignores both fields by design — it doesn't shell out.

No existing behavior changes: in production `WithEnv` is never called, so `hookEnv(nil, nil)` returns `nil` and `cmd.Env` stays unset.

### DIFF
--- a/pkg/hooks/builtin_test.go
+++ b/pkg/hooks/builtin_test.go
@@ -3,6 +3,8 @@ package hooks
 import (
 	"context"
 	"errors"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly wires `working_dir` and `env` fields through to builtin hooks, mirroring the existing command-hook logic. The core implementation is sound:

- `hookWorkingDir` and `hookEnv` are reused from the command-hook path — consistent behavior.
- `withHookEnv` / `EnvFromContext` use an unexported struct key (`hookEnvKey{}`) to avoid context-key collisions — correct pattern.
- `nil` env semantics are preserved end-to-end: builtins without overrides keep `nil`, so `cmd.Env` stays unset and git inherits `os.Environ()` — matching the PR's stated intent.
- `builtinHandler.Run` only overrides `in.Cwd` when `h.workingDir != ""`, so dispatcher-supplied cwd (e.g. `worktree_create`) continues to win when no per-hook override is configured — correct guard.
- Tests cover the key behavioral contracts: override applies when set, original cwd is preserved when unset, and env overrides flow through to git commands.

No high- or medium-severity issues were found. The implementation is clean and the change is well-scoped.

melmennaoui (approved):

PR #3299, opened 2026-06-29
refactor: receiver-based Coordinator and parallel provider tests

dgageot:
The animation `Coordinator` type was previously accessed only through package-level functions that held a single global instance. This made the code harder to test in isolation and prevented running tests in parallel, since all callers shared state.

`Coordinator`'s methods are now defined directly on the receiver so each caller can hold its own instance. The package-level shims (`Start`, `Stop`, `Add`, etc.) are kept intact for backward compatibility, delegating to a default instance. Provider tests have been updated to construct per-test registries instead of relying on shared global state, and all updated test functions call `t.Parallel()`.

Validated with `go build ./...`, `golangci-lint run`, and `task test`.

### DIFF
--- a/pkg/model/provider/factory_compat_test.go
+++ b/pkg/model/provider/factory_compat_test.go
@@ -6,32 +6,19 @@ import (
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/environment"
 	"github.com/docker/docker-agent/pkg/model/provider/options"
-	"github.com/docker/docker-agent/pkg/model/provider/rulebased"
 	"github.com/docker/docker-agent/pkg/model/provider/vertexai"
 )
 
 type providerFactory = Factory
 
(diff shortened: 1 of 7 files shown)


Sayt-0 (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, well-executed refactor. The `Coordinator` receiver-based conversion is correct — mutex usage, counter semantics, and the `globalCoordinator` initialization are all safe. The package-level backward-compat shims faithfully delegate to the singleton. The provider test changes correctly eliminate global mutable state (`geminiClientFactory`/`vertexClientFactory` vars) in favor of the per-test `googleFactoryWith(...)` closure constructor, and all helper functions (`testRegistry`, `createDirectProvider`, `resolveRoutedModel`) have been migrated to receiver-based calls consistently across all test files. No compile errors, no data races introduced.

Issue #3298, opened 2026-06-29
Make the model stream/first-token timeout configurable — long local-model (DMR) prefill aborts with "unexpected end of JSON input"


matteobisi:
### Summary

A single large prompt to a **local model via Docker Model Runner** fails with:

```
model failed: error receiving from stream: unexpected end of JSON input
```

It fails consistently at **~30 seconds**, while the model is still **prefilling** the prompt (no token streamed yet). The same request to the same DMR endpoint via `curl` succeeds — so the backend is healthy and the abort is client-side.

This does **not** appear to be the 5-minute idle timeout added in #3210: that timer resets on each received chunk, whereas here no chunk ever arrives and the abort is at ~30s, not 5 min. It's also distinct from the remote-provider / SSE-format causes in #749 (MCP), #1956 / #2349 (OpenRouter) and #2300 (OVH). The trigger here is simply a **long, healthy prefill on a local model** exceeding some ~30s client-side limit.

### Environment

- docker agent / cagent **v1.73.0**
- Docker Model Runner, model `huggingface.co/bartowski/qwen2.5-14b-instruct-gguf` (Qwen2.5-14B, Q4_K_M)
- macOS, Apple Silicon, 32 GB RAM

### Repro

1. Point an agent at a DMR local model (`provider: dmr`).
2. Send one large prompt (~48 KB / ~14K tokens — e.g. a whole meeting transcript to summarize) via `docker agent run --exec`.
3. The model prefills for ~30s with no streamed token, then the call aborts with the error above.

### Evidence that it's a client-side timeout, not the model/backend

Measured request→error gap (from `--debug` logs):

| condition | gap | result |
|-----------|-----|--------|
| cold (model not loaded) | 34.3s | fail |
| warm (model pre-loaded, prefill only) | 30.2s | fail |

Pre-warming the model removed only the load time (~4s); the call still aborts at ~30s, so it's the **prefill duration**, not loading.

Crucially, hitting the **same DMR endpoint directly with `curl`** (`stream: true`, same model, same ~48–60 KB prompt) **succeeds** — `curl` has no such client deadline. So Model Runner handles the request fine; docker agent abandons the stream before the first token.

### Expected

A long but progressing prefill should not be aborted. Local models routinely have tens-of-seconds time-to-first-token on large prompts.

### Ask

[...]

PR #3302, opened 2026-06-29
fix(modelerrors): retry stream truncation and clarify the error (#3298)

Sayt-0:
## Summary

Fixes the failure mode in #3298 where a long local-model (DMR) prefill aborts with `model failed: error receiving from stream: unexpected end of JSON input`.

Investigation showed the issue's premise (a fixed ~30s client deadline in docker-agent) does not hold: docker-agent imposes no sub-5-minute stream deadline on either DMR endpoint path, on `main` or on v1.73.0. The real, fixable problem is error handling: when an intermediary drops a healthy-but-silent prefill connection, the resulting error is classified inconsistently and surfaced as a raw decoder string.

| Aspect | Before | After |
|---|---|---|
| `connection reset` mid-stream | retryable | retryable (unchanged) |
| `unexpected EOF` mid-stream | non-retryable | retryable |
| `unexpected end of JSON input` | non-retryable | retryable |
| truncated-JSON variant | matched `"invalid"` to non-retryable | retryable |
| user-facing message | raw decoder error | actionable guidance |

## Root cause

A long prefill on a local model streams no bytes until the first token. An idle/connection timeout in the model server, or in a proxy in front of it (on Docker Desktop the DMR `model-runner.docker.internal` endpoint is reached through the Docker socket experimental endpoint), can drop that idle connection before the first token. `curl` against a direct endpoint avoids the proxy hop and succeeds.

The dropped connection surfaces from the openai-go SSE decoder as one of `unexpected EOF`, `unexpected end of JSON input`, `connection reset by peer`, or `invalid character ...`, all wrapped as `error receiving from stream: <x>`. Only the `connection reset` variant was retryable; the truncation variants were classified non-retryable (one even matched the `"invalid"` non-retryable pattern), so the request failed hard with a cryptic message.

## Change

`pkg/modelerrors`:

- Add `streamTruncationPatterns` + `IsStreamTruncationError` for the `unexpected EOF` / `unexpected end of JSON input` shapes (context cancel/deadline explicitly excluded so user/idle-timeout teardown stays non-retryable).
- Classify these as retryable in `isRetryableModelError`, checked before the pattern loops so the truncated-JSON variant is not mis-caught by `"invalid"`.
- Return an actionable message from `FormatError` instead of the raw decoder error.

Why retrying helps: a truncation mid-prefill is transient. The backend's prompt (KV) cache makes a subsequent identical request's prefill fast enough to beat the idle limit. Verified on a real DMR instance (`ai/smollm2`): an identical prompt's time-to-first-byte collapses from ~0.55s to ~0.00s once cached, and the cache survives a connection aborted mid-prefill.

## Tests

- `pkg/modelerrors`: new cases in `TestIsRetryableModelError` (the exact wrapped strings, plus a genuinely-malformed payload that must stay non-retryable), a dedicated `TestIsStreamTruncationError`, and a `FormatError` truncation case.
- End-to-end against a fake DMR server through `docker agent run --exec`:

| Scenario | Result |
|---|---|
| First attempt truncated, backend serves normally after | retried and recovered (was: immediate non-retryable failure) |
| Every attempt truncated | retried with backoff, then the actionable message |

`task build` passes; `golangci-lint run ./pkg/modelerrors/...` reports 0 issues. (`task lint` could not run fully in the sandbox: the custom `lint/` tool fails to fetch a dependency over TLS, unrelated to this change.)

## Residual considerations

- The ~30s limit itself lives upstream (model server / proxy), not in docker-agent, so a client-side timeout knob would not prevent the drop. The deeper fix (SSE keep-alive during prefill, or a higher idle timeout) belongs in Docker Model Runner.
- DMR/llama.cpp uses round-robin slot caching, so the first retry may land on a cold slot and re-incur the prefill; later attempts hit the warm slot. The default 2 retries (3 attempts) usually covers this, but a configuration with more slots or a shorter idle limit could still exhaust attempts. If preferred, the retry count for stream-truncation specifically can be raised in a follow-up.
- The "idle between bytes" semantics requested in the issue already exists on `main` via the 5-minute idle timer (added in #3210, after v1.73.0). Exposing that timer as configurable (per-model YAML field or env var) is a separate change and can be added if maintainers want it.

Refs #3298

### DIFF
--- a/pkg/modelerrors/modelerrors.go
+++ b/pkg/modelerrors/modelerrors.go
@@ -400,6 +400,47 @@ var retryablePatterns = []string{
 	"internal_error",        // HTTP/2 INTERNAL_ERROR (stream-level)
 }
 
+// streamTruncationPatterns matches errors where the response body was cut off
+// mid-stream, before a clean end-of-stream marker. For local backends (e.g.
+// Docker Model Runner / llama.cpp) this most often happens when a large
+// prompt's prefill takes long enough with no bytes flowing that an idle or
+// connection timeout in the model server -- or a proxy in front of it -- drops
+// the connection before the first token is streamed (see issue #3298). The
+// connection reset / closed-by-peer variants are already covered by
+// [retryablePatterns]; these two cover the clean-FIN-mid-event shapes that
+// surface from the SSE/JSON decoder instead of the socket layer.
+//
+// Patterns MUST be lowercase: they are compared against a lowercased error
+// message (see [IsStreamTruncationError]).
+var streamTruncationPatterns = []string{
+	"unexpected eof",               // io.ErrUnexpectedEOF: HTTP body cut mid-frame
+	"unexpected end of json input", // json.Unmarshal on a truncated SSE event
+}
+
+// IsStreamTruncationError reports whether err indicates the model stream was
+// cut off before a clean completion: the connection was dropped or the body
+// was truncated mid-stream. See [streamTruncationPatterns] for why this happens
+// and why it is treated as retryable.
+//
+// Context cancellation/deadline is excluded: that is the caller tearing the
+// request down (e.g. the idle-timeout cancel or a user Ctrl+C), not an upstream
+// drop, and must stay non-retryable.
+func IsStreamTruncationError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	for _, p := range streamTruncationPatterns {
+		if strings.Contains(msg, p) {
+			return true
+		}
+	}
+	return false
+}
+
 // nonRetryablePatterns contains error message substrings that indicate a
 // permanent/non-retryable failure. Numeric status codes (429, 401, etc.) are
 // handled separately by extractHTTPStatusCode.
@@ -468,6 +509,17 @@ func isRetryableModelError(err error) bool {
 			slog.Debug("Network timeout error, retryable", "error", err)
 			return true
 		}
+	}
+
+	// A stream truncated mid-response is transient: the upstream (or a proxy)
+	// dropped a connection that was making progress, most often during a long
+	// prefill with no bytes flowing yet. Retrying usually succeeds because the
+	// backend's prompt cache makes the second prefill fast enough to beat the
+	// idle limit. Checked before the pattern loops below so the truncated-JSON
+	// variant isn't mis-classified by the "invalid" non-retryable pattern.
+	if IsStreamTruncationError(err) {
+		slog.Debug("Stream truncated mid-response, retryable", "error", err)
+		return true
 	}
 
 	errMsg := strings.ToLower(err.Error())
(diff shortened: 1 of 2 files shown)


pkg/modelerrors/modelerrors.go
```diff
@@ -588,6 +640,14 @@ func FormatError(err error) string {
 			"Try running /compact to reduce the conversation size, or start a new session."
 	}
 
+	if IsStreamTruncationError(err) {
+		return "The connection to the model closed before it produced any output. " +
```
docker-agent: **[MEDIUM] FormatError truncation message incorrectly claims no output was produced**

The message `"The connection to the model closed before it produced any output"` is factually inaccurate for mid-stream truncations. Both matched patterns — `"unexpected eof"` (`io.ErrUnexpectedEOF`, HTTP body cut mid-frame) and `"unexpected end of json input"` (json decoder on a truncated SSE event) — can fire after partial tokens have already been streamed to the user, not only during the silent prefill phase.

A user who received several sentences of output before the connection dropped would see a message claiming the model produced nothing, which is incorrect. The follow-up guidance (`"try a shorter prompt"`) is also only appropriate for the prefill-timeout case; it does not apply to a mid-stream drop.

Suggested fix — soften the wording so it covers both scenarios:

```go
return "The connection to the model was closed unexpectedly before it completed its response. " +
    "With local models (e.g. Docker Model Runner), this can happen when an idle or " +
    "connection timeout in the model server (or a proxy in front of it) drops a silent " +
    "prefill before the first token, or cuts a long response mid-stream. " +
    "Try a shorter prompt, or raise the model server's idle/keep-alive timeout."
```

docker-agent (commented):

One medium-severity issue found in the newly introduced `FormatError` truncation message. The error detection and retry logic are correct.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR's error-handling changes in `pkg/modelerrors` are logically sound and well-structured. Reviewed the full diff and source.

**What was checked:**
- `IsStreamTruncationError`: correctly excludes `context.Canceled`/`context.DeadlineExceeded` via `errors.Is` before substring matching; patterns are pre-lowercased and the error message is lowercased before comparison — consistent with existing conventions in this package.
- Placement in `isRetryableModelError`: the truncation check fires *before* the pattern loops, so `"unexpected end of json input"` cannot be mis-caught by the `"invalid"` non-retryable substring — the stated fix is implemented correctly.
- `FormatError`: the truncation branch is correctly ordered after the overflow switch and before the raw `err.Error()` fallback.
- Test coverage: nil, wrapped errors, mixed-case, `context.Canceled`/`context.DeadlineExceeded` (non-retryable), clean EOF (non-retryable), and a genuine malformed-payload case that must stay non-retryable are all covered.
- No nil pointer dereferences, race conditions, or resource leaks introduced.

**One low-severity observation (not a blocker):**
The substring patterns (`"unexpected eof"`, `"unexpected end of json input"`) could theoretically over-match atypical proxy error messages that happen to contain those phrases in unrelated contexts. This is an inherent tradeoff of the substring-matching approach used throughout this package — not a new problem introduced by this PR — and the patterns chosen are narrow enough to be practically safe.

No bugs introduced by the PR's `+` lines were found.

dgageot (approved):

PR #3305, opened 2026-06-29
feat(session_plan): per-session markdown plan toolset alongside plan

trungutt:
**Alternative to #3292.** #3292 narrows the existing `plan` toolset to a single per-session use case, which removes the cross-session collaboration capability that Agentic Platform relies on ([Simon's comment](https://github.com/docker/docker-agent/pull/3292#issuecomment-4834033004)). This PR keeps both use cases by adding a new toolset rather than rewriting the existing one.

## What this PR does

Adds a `session_plan` toolset alongside the existing `plan` toolset:

| Toolset | Purpose | Storage |
|---|---|---|
| `plan` *(existing, unchanged)* | Shared, named, cross-session plans for multi-agent collaboration. Pluggable Storage backend. | `<data-dir>/plans/<name>.json` |
| `session_plan` *(new)* | One ephemeral markdown plan per session, plus an `exit_plan_mode` marker. | `<data-dir>/session_plans/<session-id>.md` |

The new toolset exposes three tools — `write_session_plan(content)`, `read_session_plan()`, `exit_plan_mode()` — with deliberately distinct names so an agent that enables both toolsets has no collisions.

Handlers live runtime-side (`pkg/runtime/sessionplan_handlers.go`) next to `handoff` and `transfer_task`, because they need the live `session.Session`. The toolset itself is a metadata stub.

`exit_plan_mode` is a pure marker: it validates the plan exists and returns "Plan ready for review." It does **not** switch agents or pop a confirmation dialog. Routing the next turn is the host's concern — a CLI that prints tool results inline, a chat UI with a plan-mode toggle, and a server that auto-routes through a configured handoff can all consume the same signal without one stepping on the other.

A `session_plan_updated` event fires on every successful `write_session_plan` so UIs can render the plan inline without polling. Stale plan files (>30 days) get cleaned up once per process.

## Compared to #3292

| | #3292 | This PR |
|---|---|---|
| Existing `plan` toolset | rewritten (cross-session capability lost) | untouched |
| New `session_plan` toolset | n/a | added |
| Tool names | `write_plan` / `read_plan` / `exit_plan_mode` | `write_session_plan` / `read_session_plan` / `exit_plan_mode` |
| Per-turn agent fix on warm runtimes | included | not included (orthogonal — can be cherry-picked separately) |
| AP use case preserved | no | yes |

If this PR lands, #3292 should be closed and downstream consumers (Gordon's planner, the Desktop UI renderer) updated to call the renamed tools.


pkg/runtime/sessionplan_handlers.go
```diff
@@ -0,0 +1,63 @@
...
+// the host. Switching agents is the host's decision — the runtime does not
+// call setCurrentAgent here so a CLI that prints results inline, a chat UI
+// with a mode toggle, and a server with a configured handoff can all consume
+// the same marker without one stepping on the other.
+func (r *LocalRuntime) handleExitPlanMode(_ context.Context, sess *session.Session, _ tools.ToolCall, _ EventSink) (*tools.ToolCallResult, error) {
+	if _, _, err := sessionplan.ReadContent(sessionplan.DefaultDir(), sess.ID); err != nil {
+		if errors.Is(err, sessionplan.ErrPlanNotFound) {
+			return tools.ResultError("no plan to mark ready; call write_session_plan before exit_plan_mode"), nil
+		}
+		return nil, err
```
docker-agent, line 63: **[MEDIUM] `handleExitPlanMode` does not handle `ErrInvalidSessionID` gracefully, unlike `handleReadSessionPlan`**

`ReadContent` can return `ErrInvalidSessionID` (from `Path()`) as a distinct named error before any file I/O occurs. `handleReadSessionPlan` explicitly catches it and returns a graceful `tools.ResultError(...)`, but `handleExitPlanMode` lets it fall through to `return nil, err`, which propagates as a hard internal error that is fatal to the agent loop.

The fix is a direct mirror of the pattern already used in the sibling handlers:

```go
func (r *LocalRuntime) handleExitPlanMode(_ context.Context, sess *session.Session, _ tools.ToolCall, _ EventSink) (*tools.ToolCallResult, error) {
	if _, _, err := sessionplan.ReadContent(sessionplan.DefaultDir(), sess.ID); err != nil {
		if errors.Is(err, sessionplan.ErrPlanNotFound) {
			return tools.ResultError("no plan to mark ready; call write_session_plan before exit_plan_mode"), nil
		}
		if errors.Is(err, sessionplan.ErrInvalidSessionID) {
			return tools.ResultError(err.Error()), nil
		}
		return nil, err
	}
	return tools.ResultSuccess("Plan ready for review."), nil
}
```

docker-agent (commented):

pkg/runtime/sessionplan_handlers.go
docker-agent, line 63: **[MEDIUM] `handleExitPlanMode` missing `ErrInvalidSessionID` check causes internal error instead of user-facing message**

`handleWriteSessionPlan` and `handleReadSessionPlan` both guard against `ErrInvalidSessionID` with a `tools.ResultError(err.Error()), nil` return — a clean, user-visible tool result. `handleExitPlanMode` only guards `ErrPlanNotFound`; an invalid session ID from `ReadContent` falls through to `return nil, err`, propagating it as an internal runtime error rather than a user-facing one.

In practice the runtime generates UUID session IDs that always pass the regex, but the PR docs explicitly note that custom embedder IDs could be hand-crafted. The other two handlers account for this; `handleExitPlanMode` should too.

Suggested fix:
```go
if _, _, err := sessionplan.ReadContent(sessionplan.DefaultDir(), sess.ID); err != nil {
    if errors.Is(err, sessionplan.ErrPlanNotFound) {
        return tools.ResultError("no plan to mark ready; call write_session_plan before exit_plan_mode"), nil
    }
    if errors.Is(err, sessionplan.ErrInvalidSessionID) {
        return tools.ResultError(err.Error()), nil
    }
    return nil, err
}
```

docker-agent (commented):

trungutt:
Addressed: \`handleExitPlanMode\` now guards \`ErrInvalidSessionID\` the same way \`handleWriteSessionPlan\` and \`handleReadSessionPlan\` do. Squashed into the single commit on the branch.

dgageot (approved):

PR #3307, opened 2026-06-29
refactor(sidebar): redesign agents panel to uniform two-line entries

dgageot:
The agents panel in the TUI sidebar had two distinct rendering modes: a multi-line "focus card" for the current agent and compact "roster rows" for the others. This asymmetry made the layout shift noticeably when switching agents and required maintaining separate layout computation paths.

This change replaces both modes with a single `renderAgentLine` that produces a consistent two-line entry for every agent. Line 1 shows the agent name in its accent color, a thinking-effort badge right-aligned in a shared column, and a `^N` switch shortcut flush against the right edge. Line 2 shows the provider/model string, indented and left-truncated so the informative tail is always visible. The current agent is marked with ▶ (or the spinner while working); other agents pad the marker column so names stay aligned. Agent descriptions are removed from the sidebar — they remain accessible in the agent-details dialog.

Click zones are updated so both rendered lines of an entry are owned by the agent, while the blank separator lines between entries remain unclickable. The `^N` shortcut displayed matches the existing `ctrl+1`..`ctrl+9` keyboard handler.

Removed now-unused helpers — `renderAgentCard`, `renderAgentRow`, `computeRosterLayout`, `cardThinkingLine`, `renderTreeNodes`, `wrapDescription`, `ellipsizePlain`, `rowShortcutCell`, `rightAlign` — and the `rosterLayout` type. Added small `padLeft`/`padRight`/`badgeColumnWidth` helpers and moved badge-column-width computation into `agentInfo` so it is derived once rather than per render. Layout constants `rowShortcutWidth`/`rowIndentWidth` are replaced by `agentMarkerWidth`/`agentShortcutWidth`.

One pre-existing edge case (not a regression): below the `MinWidth` auto-collapse threshold combined with very long agent names, `tab.Render` may wrap body lines and misalign click zones.

### DIFF
--- a/pkg/tui/components/sidebar/agent_click_test.go
+++ b/pkg/tui/components/sidebar/agent_click_test.go
@@ -92,10 +92,10 @@ func TestSidebar_AgentClickZones_EveryRenderedLineMapped(t *testing.T) {
 	}
 	assert.Positive(t, counts["agent1"], "agent1 should own rendered lines")
 	assert.Positive(t, counts["agent2"], "agent2 should own rendered lines")
-	// agent2 is a non-current roster agent: its row spans two lines (name+badge
+	// agent2 is a non-current roster agent: its entry spans two lines (name+badge
 	// then the indented model), and BOTH must map to it so a click on either
 	// switches to the agent.
-	assert.Equal(t, 2, counts["agent2"], "a roster agent owns both of its two row lines")
+	assert.Equal(t, 2, counts["agent2"], "a roster agent owns both of its two entry lines")
 
 	// The number of click zones equals the number of owned (non-blank) lines:
 	// every owned line is clickable.
--- a/pkg/tui/components/sidebar/agent_panel_test.go
+++ b/pkg/tui/components/sidebar/agent_panel_test.go
@@ -14,18 +14,18 @@ import (
 	"github.com/docker/docker-agent/pkg/tui/styles"
 )
 
-// newAgentPanelSidebar builds a sidebar whose current agent and roster are set,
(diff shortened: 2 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This refactor replaces the two-mode rendering system (`renderAgentCard` + `renderAgentRow`) with a unified `renderAgentLine` that produces consistent two-line entries for all agents. The new implementation is clean and logically sound.

**Width/layout calculations** — The `nameWidth` formula is internally consistent with the `right`-column formula. The `- 1` separator correctly accounts for the literal space between badge and shortcut columns; the gap always resolves to exactly `minGap=2`. No off-by-one errors.

**Left-truncation of model strings** — `TruncateTextLeft(modelText, max(1, contentWidth-agentMarkerWidth))` correctly limits model text to the available space, producing a total line width equal to `contentWidth`.

**Click zone mapping** — The `agentLineOwners` approach is preserved. `buildAgentClickZones` correctly applies the `tabHeaderLines=2` offset and skips blank owner entries. The updated click tests thoroughly verify the new 2-lines-per-agent layout.

**Badge column width** — `badgeColumnWidth` now includes the current agent (unlike the old `computeRosterLayout` which skipped it), which is correct since the current agent is now rendered as a regular line.

**Marker logic** — The spinner/`▶` precedence in the switch is correct: working agents (any) show a spinner; the current non-working agent shows `▶`; all others get blank padding.

No actionable bugs were found in the introduced code.

gtardif (approved):

PR #3310, opened 2026-06-29
feat(server): accept title in POST /api/sessions to skip LLM title generation

zampani-docker:
## Problem

`POST /api/sessions` accepts a `title` field in the request body (via the existing `json:"title"` tag on `session.Session`), but `CreateSession` silently drops it — the field is deserialized but never forwarded into the new session via `WithTitle`.

`RunSession` already has a skip condition for title generation:
```go
// session_manager.go
needsTitle := titleToEmit == "" && len(userMessages) > 0 && titleGen != nil
```
where `titleToEmit := sess.Title`. If `sess.Title` is non-empty before `RunSession` is called, the LLM title-generation call is skipped and the existing title is re-emitted instead.

## Fix

Wire up the missing assignment, mirroring the adjacent `WorkingDir` pattern:

```go
if title := strings.TrimSpace(sessionTemplate.Title); title != "" {
    opts = append(opts, session.WithTitle(title))
}
```

## Motivation

The LLM title-generation call (`sessiontitle.generate` span) runs in a goroutine but sends its `session_title` event through the same SSE stream — the connection stays open until the goroutine finishes (~2–3s). For callers that create ephemeral sessions with a known purpose and immediately delete them, this is wasted LLM inference on every invocation. Pre-supplying a title at creation is the correct fix with no client-side workarounds needed.

## Verification

```
# Before patch
POST /api/sessions {}
→ {"id": "...", "title": ""}   ← needsTitle=true, LLM fires

# After patch  
POST /api/sessions {"title":"harbor-watch"}
→ {"id": "...", "title": "harbor-watch"}  ← needsTitle=false, skipped
```

Verified locally by building the patched binary and confirming the `title` field is preserved on the session after creation.

## Testing

Added `TestSessionManager_CreateSession_CarriesTitle` with three subtests covering the non-empty, whitespace-only (TrimSpace guard), and empty cases.

### DIFF
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -339,6 +339,13 @@ func (sm *SessionManager) CreateSession(ctx context.Context, sessionTemplate *se
 		session.WithToolsApproved(sessionTemplate.ToolsApproved),
 	)
 
+	// Carry a caller-supplied title (from the POST /api/sessions request body)
+	// into the new session. When set, RunSession's needsTitle check skips the
+	// LLM title-generation call and re-emits this title instead.
+	if title := strings.TrimSpace(sessionTemplate.Title); title != "" {
+		opts = append(opts, session.WithTitle(title))
+	}
+
 	if wd := strings.TrimSpace(sessionTemplate.WorkingDir); wd != "" {
 		absWd, err := filepath.Abs(wd)
 		if err != nil {
--- a/pkg/server/session_models_test.go
+++ b/pkg/server/session_models_test.go
@@ -658,3 +658,32 @@ func TestApplyStoredOverrides(t *testing.T) {
 		})
 	})
 }
+
+func TestSessionManager_CreateSession_CarriesTitle(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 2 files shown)


dgageot (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The change is correct and well-scoped. The  field wiring mirrors the existing  pattern exactly, the  guard handles the whitespace-only edge case, and the test covers all three relevant subtests (non-empty, whitespace-only, empty). No bugs or logic errors were found in the added code.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The change is correct and well-tested. The `WithTitle` wiring precisely mirrors the adjacent `WorkingDir` pattern, the `strings.TrimSpace` guard correctly handles the whitespace-only case, and the three subtests in `TestSessionManager_CreateSession_CarriesTitle` cover all meaningful branches (non-empty title, whitespace-only, empty string). No logic errors, resource leaks, concurrency issues, or security concerns were found in the introduced lines.

PR #3311, opened 2026-06-29
test: enable t.Parallel() on isolated unit tests

dgageot:
## What

Adds `t.Parallel()` to a set of pure, well-isolated unit tests that were previously running serially.

Parallelism was added at both the top-level and subtest/table-driven levels for:

| Package | Files |
|---------|-------|
| `pkg/session` | `truncate_test.go`, `session_options_test.go`, `branch_test.go`, `clone_test.go`, `sanitize_orphan_test.go`, `store_persisted_fields_test.go`, `get_messages_tool_truncation_test.go`, `migrations_unit_test.go`, `store_memory_test.go` |
| `pkg/tools` | `schema_test.go` |
| `pkg/tui/styles` | `composite_test.go` |
| `pkg/runtime` | `cost_test.go` |

## Why

These tests are pure / self-contained: they build their own inputs (or open a private in-memory SQLite DB per test) and share no global mutable state, so they are safe to run in parallel and shorten total test time.

## Scope / what was deliberately left out

Tests that use `t.Setenv`, `t.Chdir`, or mutate global state (e.g. `pkg/path`, `pkg/shellpath`, `pkg/userid`, theme state in `pkg/tui/styles/hooks_test.go`) were **not** changed — `t.Parallel()` is incompatible with those.

## Validation

- `go test ./pkg/session/... ./pkg/tools/ ./pkg/tui/styles/ ./pkg/runtime/` — pass
- `go test -race -count=5` on the affected packages — pass
- `golangci-lint` — 0 issues
- `gofmt` — clean

> Note: a pre-existing data race in `TestHandleStream_ContextCancellation` (`pkg/runtime`) exists independently on `main` and is unrelated to this change (verified by stashing these edits).

### DIFF
--- a/pkg/runtime/cost_test.go
+++ b/pkg/runtime/cost_test.go
@@ -13,6 +13,7 @@ import (
 // nil/unpriced branches are exercised by TestComputeMessageCost in
 // after_llm_call_test.go, which shares the same computeMessageCost source.
 func TestUsageHasTokens(t *testing.T) {
+	t.Parallel()
 	assert.False(t, usageHasTokens(nil), "nil usage has no tokens")
 	assert.False(t, usageHasTokens(&chat.Usage{}), "zero usage has no tokens")
 	assert.True(t, usageHasTokens(&chat.Usage{InputTokens: 1}))
--- a/pkg/session/branch_test.go
+++ b/pkg/session/branch_test.go
@@ -10,6 +10,7 @@ import (
 )
 
 func TestGenerateBranchTitle(t *testing.T) {
+	t.Parallel()
 	tests := []struct {
 		name        string
 		parentTitle string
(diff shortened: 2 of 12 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR mechanically adds `t.Parallel()` to 12 test files across `pkg/session`, `pkg/tools`, `pkg/tui/styles`, and `pkg/runtime`. Full analysis found no bugs introduced by these changes:

- **Loop variable capture** — No risk: each table-driven subtest reads struct fields by value and does not take pointers to the range variable. Go 1.22+ loop-variable semantics further eliminate any residual concern.
- **Shared mutable state** — Each parallel subtest constructs its own local data. No shared variables are written inside the new parallel goroutines.
- **Database handles** — SQLite-based tests (`store_memory`, `migrations_unit`, `store_persisted_fields`) each open their own private in-memory database inside the test body, so no handle is shared across parallel tests.
- **`t.Setenv` / `t.Chdir` / global state** — None present in the changed tests, consistent with the PR's stated scope.
- **Parent `t` use after completion** — No subtest closes over a parent `t` in a way that could be called after the parent returns.
- **`t.Cleanup` / `defer` ordering** — No modifications to cleanup paths.

No correctness issues found. The changes safely shorten test cycle time.

gtardif (approved):

PR #3312, opened 2026-06-29
fix: data race in stalledStream.Close using sync.Once

dgageot:
The race detector flagged a data race in `TestHandleStream_ContextCancellation`: `stalledStream.Close()` could be called concurrently from the test goroutine and from `handleStream`'s deferred `stream.Close()`, both reading and writing the same `closed bool` field without synchronisation.

This replaces the unguarded `closed bool` with a `sync.Once`, so the channel close happens exactly once regardless of which caller reaches it first. As a bonus, this also prevents a potential double-close panic on the channel in cases where both callers race to close it.

Verified clean under `go test -race -count=20` on the affected package.

### DIFF
--- a/pkg/runtime/streaming_test.go
+++ b/pkg/runtime/streaming_test.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"context"
 	"io"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `sync.Once` fix is correct and complete. The race detector finding is properly addressed:

- **Double-close safety**: `sync.Once.Do` guarantees the channel is closed exactly once regardless of concurrency — no more data race on the `closed bool` field, and no risk of a double-close panic.
- **Correct capture**: The closure `func() { close(s.unblock) }` correctly captures `s` (pointer receiver), referencing the right channel.
- **No remaining races**: `unblock` is initialized once in `newStalledStream()` and never reassigned; concurrent receives on it are safe. No other shared mutable state exists on `stalledStream`.
- **Test intent preserved**: Both `TestHandleStream_IdleTimeout` and `TestHandleStream_ContextCancellation` continue to exercise the concurrent-`Close` path they were designed to test.
- **No goroutine leaks**: No new goroutines introduced; `Recv()` unblocks cleanly on channel close as before.

No issues found. Clean fix.

gtardif (approved):

PR #3316, opened 2026-06-30
refactor: extract themeRegistry struct for parallel-safe tests

dgageot:
## Summary
- Extract theme registry state into an isolated `themeRegistry` struct
- Keep package-level APIs backed by a default registry
- Allow registry-focused theme tests to run safely in parallel

## Validation
- task test
- task build
- task lint

### DIFF
--- a/pkg/tui/styles/theme.go
+++ b/pkg/tui/styles/theme.go
@@ -32,20 +32,37 @@ type themeCacheEntry struct {
 	path    string    // For user themes: file path; for built-in: empty
 }
 
-var (
-	themeCache   = make(map[string]*themeCacheEntry)
-	themeCacheMu sync.RWMutex
+// themeRegistry holds the mutable theme state: embedder-registered theme
+// sources and the lazy caches derived from them. The package-level theme API
+// is backed by a single process-wide defaultRegistry, but tests can construct
+// isolated instances with newThemeRegistry to exercise registration and
+// loading in parallel without racing on shared globals.
+type themeRegistry struct {
+	cacheMu sync.RWMutex
+	cache   map[string]*themeCacheEntry
+	// generation changes whenever cached entries are invalidated, so in-flight
+	// loads can avoid repopulating stale data after registration.
+	generation uint64
 
-	// builtinRefsCache caches the list of built-in theme refs (they never change at runtime)
-	builtinRefsCache   []string
-	builtinRefsCacheOK bool
-	builtinRefsCacheMu sync.Mutex
+	// builtinRefs caches the list of built-in theme refs (they never change at
+	// runtime); builtinRefsOK guards its lazy population.
+	refsMu        sync.Mutex
+	builtinRefs   []string
+	builtinRefsOK bool
 
-	// extraThemeFSes holds additional theme sources contributed by embedders via
+	// extraFSes holds additional theme sources contributed by embedders via
 	// RegisterBuiltinThemes. They are consulted alongside the bundled themes.
-	extraThemeFSes   []fs.FS
-	extraThemeFSesMu sync.RWMutex
-)
+	extraMu   sync.RWMutex
+	extraFSes []fs.FS
+}
+
+// newThemeRegistry returns an empty registry with an initialized theme cache.
+func newThemeRegistry() *themeRegistry {
+	return &themeRegistry{cache: make(map[string]*themeCacheEntry)}
+}
+
+// defaultRegistry backs the package-level theme API.
+var defaultRegistry = newThemeRegistry()
 
 // RegisterBuiltinThemes adds an additional source of built-in themes from fsys.
 // It must contain theme files at "themes/<ref>.yaml" (or .yml) — the same layout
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor cleanly extracts  state into an isolated struct while keeping the package-level API intact. Concurrency handling (mutex usage, generation-based cache invalidation, lock ordering) is consistent and correct. The test isolation strategy — using a non-parallel test to swap the package-level `defaultRegistry` with a fresh instance — is idiomatic Go and safe under the sequential test scheduler.

pkg/tui/styles/theme_test.go
```diff
@@ -72,26 +73,50 @@ func TestRegisterBuiltinThemes_Integration(t *testing.T) {
 	assert.Equal(t, "embedder", CurrentTheme().Ref)
 }
 
+// TestRegisterBuiltinThemes_PackageAPI verifies the exported wrapper functions
+// use the package-wide default registry correctly.
+func TestRegisterBuiltinThemes_PackageAPI(t *testing.T) { //nolint:paralleltest // Temporarily swaps package-wide defaultRegistry.
+	previous := defaultRegistry
+	defaultRegistry = newThemeRegistry()
```
docker-agent, line 80: **[HIGH] Unsynchronised write to `defaultRegistry` races with parallel tests**

`TestRegisterBuiltinThemes_PackageAPI` performs a plain (lock-free) assignment to the package-level `defaultRegistry` variable:

```go
previous := defaultRegistry
defaultRegistry = newThemeRegistry()      // ← bare write, no mutex
t.Cleanup(func() { defaultRegistry = previous })
```

In Go's test runner, top-level parallel tests (those that have already called `t.Parallel()`) are placed in a pool and released to run concurrently while sequential tests continue executing. Tests such as `TestAllBuiltinThemes_LoadSuccessfully` and `TestLoadTheme_Default` both call `t.Parallel()` and invoke package-level wrappers (`LoadTheme`, `ListThemeRefs`, `IsBuiltinTheme`) that read `defaultRegistry` without synchronisation. This creates a genuine data race detectable by `go test -race`.

The `//nolint:paralleltest` comment documents the intent to run sequentially with respect to future _new_ tests, but it does not prevent already-released parallel tests from racing against this write.

**Suggested fix**: protect `defaultRegistry` access with an `atomic.Pointer[themeRegistry]` or a `sync.RWMutex`, so the swap and all reads are race-free, or move this test into a `TestMain` guard that serialises the swap before any parallel tests start.

docker-agent (commented):

This refactor cleanly extracts  and enables parallel tests. Two issues were found in the new code:

1. **[HIGH]**  writes to the package-level  without synchronisation — detectable by `go test -race` when parallel tests are running concurrently.
2. **[MEDIUM]** In `LoadTheme`'s retry loop, `isBuiltin` is determined before the generation is snapshotted, creating a narrow TOCTOU window where a stale built-in classification can be cached with a fresh generation number.

pkg/tui/styles/theme.go
```diff
@@ -536,72 +574,81 @@ func LoadTheme(ref string) (*Theme, error) {
...
-
-	// For user themes, check if file exists and get modTime
-	var userThemePath string
-	var userModTime time.Time
-	if !isBuiltin {
-		userThemePath, userModTime = getUserThemeFileInfo(baseRef)
-	}
+	for {
+		// Determine if this should load from built-in or user themes
+		isBuiltin := !forceUserTheme && r.IsBuiltinTheme(baseRef)
```
docker-agent, line 579: **[MEDIUM] TOCTOU: `isBuiltin` determined before generation snapshot, stale value can be cached**

In `LoadTheme`'s retry loop, `isBuiltin` is evaluated before the generation is snapshotted:

```go
for {
    isBuiltin := !forceUserTheme && r.IsBuiltinTheme(baseRef)  // ← acquires/releases refsMu
    // ...
    r.cacheMu.RLock()
    generation := r.generation   // ← snapshot taken AFTER isBuiltin check
    cached, hasCached := r.cache[ref]
    r.cacheMu.RUnlock()
    // ... later: store entry with snapshotted generation
}
```

If `RegisterBuiltinThemes` runs between the `IsBuiltinTheme` call and the `r.cacheMu.RLock()` snapshot, the generation will have been incremented. The loop's generation-comparison guard at store time will then see a fresh generation (matching the post-increment value it just read), and cache the stale `isBuiltin=false` result with that generation. Subsequent `LoadTheme` calls will serve this incorrect cache entry — treating a newly-registered built-in theme as a user theme — until the next explicit cache invalidation.

The impact is bounded to the narrow window during startup when `RegisterBuiltinThemes` races a concurrent `LoadTheme`, but it is a logical correctness issue introduced by the new extraction that is worth closing.

**Suggested fix**: snapshot the generation first (`r.cacheMu.RLock()` / read `generation` / `r.cacheMu.RUnlock()`), _then_ call `r.IsBuiltinTheme`, so that any intervening `RegisterBuiltinThemes` increments the generation to a value higher than the snapshot — causing the guard to detect the race and retry.

melmennaoui (approved):

PR #3317, opened 2026-06-30
refactor(runregistry): introduce Registry struct and tighten dir permissions

dgageot:
The `pkg/runregistry` package previously relied on a process-global directory path set via `paths.SetDataDir`, which made it impossible to run its tests in parallel — each test case mutated shared state and raced against the others.

This change converts the package to an explicit `Registry` struct that carries its own `dir` field and an injectable `alive func(pid int) bool` liveness hook. The five package-level functions (`Write`, `List`, `Latest`, `Find`, `Dir`) become methods on `*Registry`. Two constructors are provided: `New(dir)` for tests and callers that need a specific root, and `Default()` which roots at `<data dir>/runs` — the same location as before. Call sites in `cmd/root/run_listen.go` and `cmd/root/mcp.go` are updated to use `Default()`, and the test file is rewritten so every test creates its own isolated `Registry` instance and runs with `t.Parallel()`.

A code-review finding is also addressed: `Write` now explicitly `chmod`s the registry directory to `0o700` after `MkdirAll`, because `MkdirAll` only sets the mode on *newly created* directories — a pre-existing directory with looser permissions would otherwise let other local users enumerate live PIDs and addresses. A regression test (`TestWrite_TightensExistingDirectoryPermissions`) covers this path, and a `//nolint:gosec` comment suppresses the false-positive G302 warning on the `0o700` constant (a directory needs the execute bit; owner-only is the intent).

### DIFF
--- a/pkg/runregistry/registry.go
+++ b/pkg/runregistry/registry.go
@@ -69,7 +69,7 @@ func (r *Registry) Write(rec Record) (func(), error) {
 	// MkdirAll only applies the mode to directories it creates, so an
 	// already-existing dir may be group/world-readable. Tighten it explicitly
 	// to keep live PIDs/addresses unreadable by other local users.
-	if err := os.Chmod(r.dir, 0o700); err != nil {
+	if err := os.Chmod(r.dir, 0o700); err != nil { //nolint:gosec // directory needs execute bit; 0700 is owner-only
 		return nil, fmt.Errorf("restricting run registry dir: %w", err)
 	}
 
--- a/pkg/runregistry/registry.go
+++ b/pkg/runregistry/registry.go
@@ -29,9 +29,30 @@ type Record struct {
 	StartedAt time.Time `json:"started_at"`
 }
 
+// Registry persists and queries run records stored as per-pid JSON files under
+// a single directory. Construct one with [New] or [Default]; methods carry no
+// shared mutable state, so distinct instances (e.g. one per test) are fully
+// independent and safe to use concurrently.
+type Registry struct {
+	dir string
+	// alive reports whether a pid is live; overridable so tests need not rely
+	// on real process state.
+	alive func(pid int) bool
+}
+
+// New returns a Registry that stores records under dir.
+func New(dir string) *Registry {
+	return &Registry{dir: dir, alive: pidAlive}
+}
+
+// Default returns a Registry rooted at the user's data dir (<data dir>/runs).
+func Default() *Registry {
+	return New(filepath.Join(paths.GetDataDir(), "runs"))
+}
+
 // Dir is the directory holding run records.
-func Dir() string {
-	return filepath.Join(paths.GetDataDir(), "runs")
+func (r *Registry) Dir() string {
+	return r.dir
 }
 
 // Write atomically persists a record for the current process and returns a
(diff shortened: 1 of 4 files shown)


pkg/runregistry/registry.go
```diff
@@ -41,12 +62,18 @@ func Dir() string {
...
-func Write(rec Record) (func(), error) {
-	if err := os.MkdirAll(Dir(), 0o700); err != nil {
+func (r *Registry) Write(rec Record) (func(), error) {
+	if err := os.MkdirAll(r.dir, 0o700); err != nil {
 		return nil, fmt.Errorf("creating run registry dir: %w", err)
 	}
+	// MkdirAll only applies the mode to directories it creates, so an
+	// already-existing dir may be group/world-readable. Tighten it explicitly
+	// to keep live PIDs/addresses unreadable by other local users.
+	if err := os.Chmod(r.dir, 0o700); err != nil { //nolint:gosec // directory needs execute bit; 0700 is owner-only
```
docker-agent, line 72: **[MEDIUM] `os.Chmod` failure on pre-existing registry dir causes `Write` to always fail for non-owner callers**

The newly-added `os.Chmod(r.dir, 0o700)` call (line 72) returns a hard error if the calling process does not own the registry directory. Before this PR there was no `Chmod` call, so `Write` succeeded as long as the process could write to the directory. This PR introduces a regression for at least one realistic scenario:

> A daemon (or install script) creates `<datadir>/runs` as `root`, then the process drops privileges. When `Write` is called under the unprivileged UID, `os.Chmod` returns `EPERM` because only the owner can change permissions — and `Write` fails permanently with `"restricting run registry dir: operation not permitted"`.

The underlying goal (tighten permissions on a pre-existing, world-readable dir) is sound. A more resilient approach would be to treat a `Chmod` error as a non-fatal warning — log it but continue writing the record. The security posture only degrades if the directory is pre-existing *and* loose; a process that can't `chmod` it also can't make it looser, so the worst case is the pre-existing state, not a new vulnerability.

Suggested fix:
```go
if err := os.Chmod(r.dir, 0o700); err != nil {
    // Best-effort: if we don't own the dir we can't tighten it,
    // but we can still write our record.
    log.Debugf("run registry: could not tighten dir permissions: %v", err)
}
```

docker-agent (commented):

pkg/runregistry/registry.go
docker-agent, line 72: **[MEDIUM] Hard-error on `os.Chmod` failure makes the registry unusable when the directory is owned by a different user**

`os.Chmod(r.dir, 0o700)` returns `EPERM` whenever the caller does not own the directory (e.g., the directory was previously created by `root`, a different uid in a container, or any privileged-upgrade workflow). When that happens, `Write` returns a non-nil error on every subsequent call, permanently breaking the registry for the current user until an admin manually fixes ownership.

Before this PR, `Write` only called `os.MkdirAll`; a pre-existing directory with looser permissions would simply be used as-is, so this is a **new hard-failure path introduced by this PR**.

Suggested fix — treat `Chmod` failures as non-fatal, since the security intent (preventing enumeration by *other* unprivileged users) is already served by `MkdirAll(0o700)` for newly-created directories, and a directory owned by root is unlikely to be readable by arbitrary local users anyway:

```go
// Best-effort: tighten permissions on pre-existing directories.
// Ignore EPERM — if we don't own the dir we can't chmod it, but that
// scenario (root-owned dir, non-root caller) still provides adequate isolation.
_ = os.Chmod(r.dir, 0o700) //nolint:gosec
```

Alternatively, check `os.Stat` and only call `Chmod` when the current mode is wider than `0o700`:

```go
if info, err := os.Stat(r.dir); err == nil && info.Mode().Perm() > 0o700 {
    _ = os.Chmod(r.dir, 0o700) //nolint:gosec
}
```

docker-agent (commented):

The refactor from package-level globals to a `Registry` struct is clean and the parallel-test rewrite is correct. One concern with the new `os.Chmod` call is flagged below.

melmennaoui (approved):

PR #3318, opened 2026-06-30
feat(worktree): add WithRoot option to decouple from global data dir

dgageot:
The `pkg/worktree` tests could not run with `t.Parallel()` because `Create` and `CreatePR` resolved the worktree destination by reading the process-global `paths.GetDataDir()`. Tests that needed a different root had to call `paths.SetDataDir`, which meant concurrent tests would clobber each other's state.

This adds a `WithRoot(dir)` functional option to both `Create` and `CreatePR`. Internally, a `resolveCreateConfig` helper builds a `createConfig` that defaults `root` to `paths.GetDataDir()` then applies any supplied options, so production callers in `cmd/root/run.go` are unchanged. The worktree tests now pass `WithRoot(t.TempDir())`, drop the global `SetDataDir` mutation, and run with `t.Parallel()`. The `createTestWorktree` helper in `cmd/root/run_worktree_test.go` was updated the same way.

No behavior change for production code; the only risk is the new variadic parameter on `CreatePR`, which is backwards-compatible.

### DIFF
--- a/cmd/root/run_worktree_test.go
+++ b/cmd/root/run_worktree_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/docker/docker-agent/pkg/cli"
-	"github.com/docker/docker-agent/pkg/paths"
 	"github.com/docker/docker-agent/pkg/worktree"
 )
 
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `WithRoot` functional-option refactor is clean and correct. The `resolveCreateConfig` helper properly defaults `root` to `paths.GetDataDir()` before applying caller-supplied options, preserving existing production behaviour. The variadic addition to `CreatePR` is backwards-compatible. Tests drop the global `SetDataDir` mutation and now run with `t.Parallel()` — exactly the right fix. No bugs, resource leaks, or behavioral regressions were found in the changed lines.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are correct and well-structured. The `resolveCreateConfig` helper properly initialises defaults before applying caller-supplied options, `WithRoot` guards against empty strings to preserve the default, and both `Create` and `CreatePR` correctly thread `cfg.root` through to their `dest` path. No behaviour change for production callers; the variadic addition to `CreatePR` is backwards-compatible. Tests now run safely in parallel.

melmennaoui (approved):

PR #3319, opened 2026-06-30
refactor: remove process-global clock/ID vars from pkg/session

dgageot:
Global mutable variables in `pkg/session` — a `nowFn` pointing to `time.Now` and a `newIDFn` wrapping a UUID generator — were shared state that tests mutated directly. This made the test suite inherently sequential in that package and introduced data-race potential under `-race`.

The fix moves those dependencies inside each `Session` as unexported `clock func() time.Time` and `idgen func() string` fields, exposed through `now()` and `newID()` accessor methods that fall back to `time.Now` and a random UUID when nil. Existing `&Session{}` literals and JSON-deserialized sessions continue to work without any call-site changes. Two new constructor options, `WithClock` and `WithIDGen`, allow deterministic injection in tests. `New()` now applies all options before generating the ID and `CreatedAt` timestamp, so `WithID` can still short-circuit ID generation as before. Explicit-timestamp message constructors (`UserMessageAt`, `SystemMessageAt`, `ImplicitUserMessageAt`) were added so the originals delegate to them rather than capturing `time.Now` at call time; `buildSessionSummaryMessages` became a method on `*Session` to use `s.now()`. The clock tests in `pkg/session/clock_test.go` were rewritten with local `fixedClock`/`stubIDs` helpers and `t.Parallel()`, so that file now runs its cases in parallel cleanly.

No behaviour is changed for callers. The package passes `go test -race ./pkg/session/...` with all tests marked `t.Parallel()`.

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -718,19 +718,19 @@ type Opt func(s *Session)
 
 func WithUserMessage(content string) Opt {
 	return func(s *Session) {
-		s.AddMessage(UserMessage(content))
+		s.AddMessage(UserMessageAt(s.now(), content))
 	}
 }
 
 func WithImplicitUserMessage(content string) Opt {
 	return func(s *Session) {
-		s.AddMessage(ImplicitUserMessage(content))
+		s.AddMessage(ImplicitUserMessageAt(s.now(), content))
 	}
 }
 
 func WithSystemMessage(content string) Opt {
 	return func(s *Session) {
-		s.AddMessage(SystemMessage(content))
+		s.AddMessage(SystemMessageAt(s.now(), content))
 	}
 }
 
@@ -831,10 +831,13 @@ func WithID(id string) Opt {
 	}
 }
 
-// WithClock injects the time source used for the session's CreatedAt and for
-// timestamping messages it generates (summaries, compaction input). Primarily
-// for tests that need a deterministic clock; production code leaves it unset
-// and falls back to time.Now.
+// WithClock injects the time source used for the session's CreatedAt, for
+// timestamping messages it generates (summaries, compaction input), and for
+// messages added through WithUserMessage/WithSystemMessage/
+// WithImplicitUserMessage. Because those message options read the clock when
+// they run, WithClock must precede them in the option list (the natural
+// ordering). Primarily for tests that need a deterministic clock; production
+// code leaves it unset and falls back to time.Now.
 func WithClock(now func() time.Time) Opt {
 	return func(s *Session) {
 		s.clock = now
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -17,14 +17,8 @@ import (
 	"github.com/docker/docker-agent/pkg/tools"
 )
 
-// nowFn returns the current time. Indirected through a package-level variable
-// so that tests can install a deterministic clock via setNowForTest.
-var nowFn = time.Now
-
-// newIDFn returns a fresh session ID. Indirected through a package-level
-// variable so that tests can install a deterministic ID generator via
-// setIDForTest.
-var newIDFn = func() string { return uuid.New().String() }
+// defaultNewID returns a fresh random session ID.
+func defaultNewID() string { return uuid.New().String() }
 
 const (
 	// toolContentPlaceholder is the text used to replace truncated tool content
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, well-structured refactor. Moving `nowFn`/`newIDFn` from process-global state into per-`Session` `clock`/`idgen` fields is the right design: the nil-fallback `now()` and `newID()` accessors ensure zero-value `Session{}` structs and JSON-deserialized sessions continue to work without changes, the option application order (options first, then ID/`CreatedAt` generation) correctly enables `WithClock` and `WithIDGen` to influence construction, and `WithID` still short-circuits ID generation as before.

The `*At` variants (`UserMessageAt`, `SystemMessageAt`, `ImplicitUserMessageAt`) are a nice addition for callers that need explicit timestamps without going through the session clock. The `WithClock` docstring correctly scopes the clock to session-generated timestamps (summaries, compaction), not to caller-supplied messages added via `WithUserMessage` etc. — that's consistent and intentional.

The rewritten clock tests with `t.Parallel()`, `fixedClock`, and `stubIDs` helpers are straightforward and correct.

docker-agent (commented):

One medium-severity behavioral gap was found in this refactor: the `With*Message` option helpers (`WithUserMessage`, `WithSystemMessage`, `WithImplicitUserMessage`) bypass the per-session clock introduced by `WithClock`, stamping message timestamps with `time.Now()` rather than the injected clock. See inline comment for details.

The core mechanics of the refactor — nil-fallback `now()`/`newID()` accessors, option application ordering in `New()`, delegation chains in the `*At` constructors, and `buildSessionSummaryMessages` becoming a method — are all correct.

pkg/session/session.go
```diff
@@ -248,18 +252,30 @@ type Message struct {
...
+// ImplicitUserMessageAt is like ImplicitUserMessage but stamps the message with
+// an explicit creation time, letting callers (and tests) avoid the wall clock.
+func ImplicitUserMessageAt(createdAt time.Time, content string) *Message {
+	msg := UserMessageAt(createdAt, content)
 	msg.Implicit = true
 	return msg
 }
 
 func UserMessage(content string, multiContent ...chat.MessagePart) *Message {
+	return UserMessageAt(time.Now(), content, multiContent...)
```
docker-agent, line 267: **[MEDIUM] `WithUserMessage`/`WithSystemMessage`/`WithImplicitUserMessage` bypass the per-session clock injected via `WithClock`**

This PR removes the process-global `nowFn` and replaces it with per-session clock injection via `WithClock`. However, the pre-existing `WithUserMessage`, `WithSystemMessage`, and `WithImplicitUserMessage` option functions call the package-level `UserMessage()`/`SystemMessage()`/`ImplicitUserMessage()` free functions, which this PR changed to call `UserMessageAt(time.Now(), ...)` — hardcoding `time.Now()` rather than delegating to `s.now()`.

**Result**: A test that does `New(WithClock(fixedClock), WithUserMessage("hello"))` will get the session's `CreatedAt` stamped with `fixedClock`, but the message's `CreatedAt` will still use the real wall clock. The gap didn't exist before this PR because `nowFn` was a global that applied everywhere (including inside `UserMessage`); now that the global is gone, the `With*Message` options are left out of the per-session clock contract.

**Trigger**: Any caller combining `WithClock` and `WithUserMessage`/`WithSystemMessage`/`WithImplicitUserMessage` in the same `New(...)` call.

**Suggested fix**: Update `WithUserMessage`, `WithSystemMessage`, and `WithImplicitUserMessage` to use `s.now()` directly instead of delegating to the free functions:
```go
func WithUserMessage(content string) Opt {
    return func(s *Session) {
        s.AddMessage(UserMessageAt(s.now(), content))
    }
}
func WithSystemMessage(content string) Opt {
    return func(s *Session) {
        s.AddMessage(SystemMessageAt(s.now(), content))
    }
}
func WithImplicitUserMessage(content string) Opt {
    return func(s *Session) {
        s.AddMessage(ImplicitUserMessageAt(s.now(), content))
    }
}
```
Note: this requires `WithClock` to be listed before `WithUserMessage` in the `opts` slice, which is the natural ordering (clock injected first, messages added after).

melmennaoui (dismissed):

trungutt (approved):

PR #3321, opened 2026-06-30
refactor: introduce Resolver type in userid and inject cagent ID source in httpclient for parallel-safe tests

dgageot:
The `pkg/userid` and `pkg/httpclient` tests could not use `t.Parallel()` safely because they both relied on package-global mutable state: `userid` had a cached UUID and a process-wide `paths.SetConfigDir` override, while `httpclient` stamped the `X-Cagent-Id` header by calling the global `userid.Get` directly inside `WithProxiedBaseURL`.

`pkg/userid` now exposes a `Resolver` type that owns its own directory function and an in-memory cache guarded by a mutex. Each test constructs an independent `Resolver` via `New(t.TempDir())` and runs in parallel without interfering with others. The package-level `Get()` delegates to a `defaultResolver` whose directory function is `paths.GetConfigDir`, so production behavior is completely unchanged. The now-unused `ResetForTests()` and `reset()` helpers were removed.

`pkg/httpclient` gains a dependency-injection seam: `HTTPOptions` carries a `cagentID func() string` that defaults to `userid.Get`. The header stamping was moved from `WithProxiedBaseURL` into `RoundTrip`, next to the existing `X-Cagent-Session-Id` logic and gated on the same `X-Cagent-Forward` header, which preserves behavior for all real callers. An unexported `withCagentIDSource(fn)` option lets tests inject a deterministic value, so those tests can also run with `t.Parallel()` without touching global state.

Both packages pass `go test -race` and the full `task test` suite; `task lint` reports no new issues.

### DIFF
--- a/pkg/httpclient/client.go
+++ b/pkg/httpclient/client.go
@@ -19,13 +19,20 @@ import (
 type HTTPOptions struct {
 	Header http.Header
 	Query  url.Values
+
+	// cagentID resolves the persistent install UUID stamped as
+	// `X-Cagent-Id` on gateway-bound requests. It defaults to
+	// [userid.Get]; tests inject their own source via
+	// [withCagentIDSource] to stay independent of global state.
+	cagentID func() string
 }
 
 type Opt func(*HTTPOptions)
 
 func NewHTTPClient(ctx context.Context, opts ...Opt) *http.Client {
 	httpOptions := HTTPOptions{
-		Header: make(http.Header),
+		Header:   make(http.Header),
+		cagentID: userid.Get,
 	}
 
 	for _, opt := range opts {
@@ -125,14 +132,16 @@ func WithProxiedBaseURL(value string) Opt {
 		o.Header.Set("X-Cagent-Arch", runtime.GOARCH)
 		o.Header.Set("X-Cagent-Runtime", "cagent")
 		o.Header.Set("X-Cagent-Runtime-Version", version.Version)
+	}
+}
 
-		// Stamp the persistent UUID identifying this cagent install so
-		// the gateway can correlate calls coming from the same client
-		// across sessions and processes. Same value as the `user_uuid`
-		// telemetry property; the gateway is free to ignore it.
-		if id := userid.Get(); id != "" {
-			o.Header.Set("X-Cagent-Id", id)
-		}
+// withCagentIDSource overrides the source of the `X-Cagent-Id` header.
+// Unexported because it exists solely so tests can supply a
+// deterministic, isolated [userid.Resolver] instead of the global
+// default.
+func withCagentIDSource(fn func() string) Opt {
+	return func(o *HTTPOptions) {
+		o.cagentID = fn
 	}
 }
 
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This refactor correctly replaces package-global mutable state with a per-instance `Resolver` type and a dependency-injection seam in `httpclient`, making both packages safe for `t.Parallel()` tests.

**What was reviewed:**
- `pkg/userid/userid.go` — `Resolver` struct, mutex-guarded cache, `New()` constructor, `defaultResolver`, package-level `Get()`
- `pkg/httpclient/client.go` — `cagentID func() string` field, default injection in `NewHTTPClient`, header stamping moved to `RoundTrip`, `withCagentIDSource` option
- Both test files

**No bugs found.** Key observations:
- The `if u.httpOptions.cagentID != nil` guard in `RoundTrip` is correct defensive coding — a `userAgentTransport` constructed outside `NewHTTPClient` would have a nil `cagentID`.
- `defaultResolver` captures `paths.GetConfigDir` as a function pointer (lazy), so it correctly honours any `SetConfigDir` override active at call time — until the first successful `Get()` call caches the result. This is the expected lifetime for production use.
- No mutex-copy hazard: `New()` returns `*Resolver`; `HTTPOptions` holds only a `func() string`, no embedded mutex.
- Header stamping behaviour is preserved: `X-Cagent-Id` is still only stamped on gateway-bound requests (gated on `X-Cagent-Forward`), now evaluated at `RoundTrip` time instead of option-apply time.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This refactor is well-structured and achieves its goal of making both `pkg/userid` and `pkg/httpclient` tests parallel-safe without touching global state.

**What was reviewed:**
- `pkg/userid`: `Resolver` type with `sync.Mutex`-guarded cache, `New(dir)` constructor, `defaultResolver` package-level delegation
- `pkg/httpclient`: `cagentID func() string` field in `HTTPOptions`, header-stamping moved to `RoundTrip`, `withCagentIDSource` test-injection seam

**No bugs found.** All hypotheses examined were dismissed:

1. **`X-Cagent-Id` gating change** — `X-Cagent-Forward` is exclusively set by `WithProxiedBaseURL` in the production codebase, so moving the stamping from option-setup time into `RoundTrip` is behaviorally equivalent. Per-request stamping in `RoundTrip` is actually the more correct placement.

2. **`withCagentIDSource(nil)`** — The nil-guard in `RoundTrip` is intentional and handles this safely. The function is unexported and exists solely for tests; no production path can trigger it accidentally.

3. **Mutex held across file I/O** — Standard lazy-initialization pattern; no deadlock risk, no data race, minor one-time contention on first call after which the cache is warm.

**Concurrency correctness**: `defaultResolver` is initialized as a package-level var with zero-value `sync.Mutex{}` (valid in Go) and `cached: ""` (correctly triggers re-read on first call). The design is sound.

Sayt-0 (approved):

PR #3322, opened 2026-06-30
feat(1password): cache op:// resolutions and coalesce concurrent lookups

dgageot:
Every time an agent environment variable resolved to an `op://` reference, the `op` CLI was spawned unconditionally. In workloads that reference the same secret many times—or that have concurrent tool calls touching the same variable—this produced redundant subprocess invocations and, in some edge cases, leaked the raw `op://` string to downstream providers when resolution failed.

This change introduces two complementary fixes. First, resolved references are now memoized so the `op` CLI is called at most once per distinct `op://` reference per process lifetime; successful resolutions and deterministic failures (missing binary, bad reference) are cached, while failures caused by a cancelled or expired context are deliberately *not* cached so a later call can retry. Second, `golang.org/x/sync/singleflight` is used to coalesce concurrent resolutions of the *same* reference—preventing redundant `op` invocations under parallelism while still allowing different references to resolve in parallel. The `Get` method also now returns `("", true)` instead of `("", false)` when a reference cannot be resolved, so a raw `op://` string is never passed to downstream providers as a literal value. Finally, the raw reference is no longer logged on failure to avoid leaking vault/item/field metadata into log output.

No breaking changes; the public API and behavior for callers that were already getting a resolved secret are unchanged.

### DIFF
--- a/pkg/environment/onepassword.go
+++ b/pkg/environment/onepassword.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log/slog"
 	"strings"
+	"sync"
 )
 
 // onePasswordPrefix marks an environment value as a 1Password secret reference
@@ -19,6 +20,12 @@ type OnePasswordProvider struct {
 	// resolve turns a "op://..." reference into its secret value. It is a field
 	// so tests can inject a fake resolver without relying on the `op` binary.
 	resolve func(ctx context.Context, reference string) (string, bool)
+
+	// cache memoizes resolved references so the same secret is not looked up
+	// (and the `op` CLI not spawned) repeatedly. Failed resolutions are cached
+	// as an empty string too, since they are equally deterministic per run.
+	mu    sync.Mutex
+	cache map[string]string
 }
 
 type OnePasswordNotAvailableError struct{}
@@ -35,6 +42,7 @@ func NewOnePasswordProvider(provider Provider) Provider {
 	return &OnePasswordProvider{
 		provider: provider,
 		resolve:  resolveOnePasswordReference,
+		cache:    make(map[string]string),
 	}
 }
 
@@ -54,11 +62,29 @@ func (p *OnePasswordProvider) Get(ctx context.Context, name string) (string, boo
 		return value, found
 	}
 
-	resolved, ok := p.resolve(ctx, value)
-	if !ok {
-		slog.WarnContext(ctx, "Failed to resolve 1Password secret reference", "name", name)
-		return "", false
+	// Always report the variable as found: returning the raw "op://" reference
+	// would leak it to downstream providers, so an unresolved reference becomes
+	// an empty value instead.
+	return p.resolveCached(ctx, value), true
+}
+
+func (p *OnePasswordProvider) resolveCached(ctx context.Context, reference string) string {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.cache == nil {
+		p.cache = make(map[string]string)
+	}
+	if resolved, ok := p.cache[reference]; ok {
+		return resolved
 	}
 
-	return resolved, true
+	resolved, ok := p.resolve(ctx, reference)
+	if !ok {
+		slog.WarnContext(ctx, "Failed to resolve 1Password secret reference; using empty value", "reference", reference)
+		resolved = ""
+	}
+
+	p.cache[reference] = resolved
+	return resolved
 }
--- a/pkg/environment/onepassword.go
+++ b/pkg/environment/onepassword.go
@@ -5,6 +5,8 @@ import (
 	"log/slog"
 	"strings"
 	"sync"
+
+	"golang.org/x/sync/singleflight"
 )
 
 // onePasswordPrefix marks an environment value as a 1Password secret reference
@@ -21,9 +23,15 @@ type OnePasswordProvider struct {
 	// so tests can inject a fake resolver without relying on the `op` binary.
 	resolve func(ctx context.Context, reference string) (string, bool)
 
-	// cache memoizes resolved references so the same secret is not looked up
-	// (and the `op` CLI not spawned) repeatedly. Failed resolutions are cached
-	// as an empty string too, since they are equally deterministic per run.
+	// group coalesces concurrent resolutions of the same reference so the `op`
+	// CLI is spawned at most once per reference, while still allowing distinct
+	// references to resolve in parallel.
+	group singleflight.Group
+
+	// cache memoizes resolved references. Successful results and deterministic
+	// failures (op missing, bad reference) are cached so we don't keep spawning
+	// the CLI; transient failures (e.g. context cancellation) are not cached so
+	// a later call can retry.
 	mu    sync.Mutex
 	cache map[string]string
 }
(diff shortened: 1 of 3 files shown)


pkg/environment/onepassword.go
```diff
@@ -54,11 +70,52 @@ func (p *OnePasswordProvider) Get(ctx context.Context, name string) (string, boo
...
+	return p.resolveCached(ctx, value), true
+}
+
+func (p *OnePasswordProvider) resolveCached(ctx context.Context, reference string) string {
+	if cached, ok := p.cachedValue(reference); ok {
+		return cached
 	}
 
-	return resolved, true
+	resolved, _, _ := p.group.Do(reference, func() (any, error) {
```
docker-agent, line 84: **[MEDIUM] Singleflight coalesces concurrent callers but shares a single context — a cancelled caller's ctx can cause all coalesced callers to receive an empty value**

When two goroutines call `resolveCached` concurrently for the same reference, `singleflight.Group.Do` runs the work function only once — using whichever goroutine's closure was captured first. If that goroutine happens to carry a cancelled or expired `ctx`, `p.resolve(ctx, reference)` will immediately fail with a context error. The cache-skip guard (`if ctx.Err() != nil { return value, nil }`) prevents the empty result from being *persisted*, but all coalesced callers still receive the empty string for *this* invocation — including goroutines that had a perfectly healthy context.

The fix is to decouple the resolution from any single caller's cancellation deadline. Use `context.WithoutCancel(ctx)` (Go 1.21+) when calling `p.resolve`, so the underlying `op read` subprocess is not cancelled by any one caller's context:

```go
resolved, _, _ := p.group.Do(reference, func() (any, error) {
    if cached, ok := p.cachedValue(reference); ok {
        return cached, nil
    }
    // Use a detached context so a cancelled caller cannot abort
    // a resolution that other concurrent callers are waiting for.
    value, ok := p.resolve(context.WithoutCancel(ctx), reference)
    ...
})
```

This also simplifies the "don't cache transient failures" check: with `WithoutCancel`, cancellation from an individual caller never propagates into the singleflight work, so all failures become deterministic and can be cached (or handled with a shorter retry window).

pkg/environment/onepassword.go
```diff
@@ -54,11 +70,52 @@ func (p *OnePasswordProvider) Get(ctx context.Context, name string) (string, boo
...
 	}
 
-	resolved, ok := p.resolve(ctx, value)
-	if !ok {
-		slog.WarnContext(ctx, "Failed to resolve 1Password secret reference", "name", name)
-		return "", false
+	// Always report the variable as found: returning the raw "op://" reference
+	// would leak it to downstream providers, so an unresolved reference becomes
+	// an empty value instead.
+	return p.resolveCached(ctx, value), true
```
docker-agent, line 76: **[MEDIUM] `("", true)` on resolution failure silently injects an empty string as the secret value — callers have no way to distinguish "not an op:// reference" from "resolution failed"**

Returning `(value, true)` regardless of whether resolution succeeded means every caller of `Get` sees `found=true` even when the `op` CLI is missing, the reference is malformed, or permissions are denied. The raw `op://` string is correctly suppressed, but the silent empty-string substitution trades one failure mode for another: instead of the agent surfacing an error, downstream code (e.g. an HTTP client, a DB driver) receives an empty bearer token or password and fails in ways that are much harder to diagnose.

The previous code returned `("", false)` on failure, which let callers detect the missing value and potentially refuse to start or log a clear error. Consider returning `("", false)` for deterministic failures (binary not found, bad reference) while keeping `("", true)` only for the transient/context-cancelled path where the reference *is* valid but temporarily unresolvable:

```go
value, ok := p.resolve(ctx, reference)
if !ok {
    if ctx.Err() != nil {
        // Transient: suppress the raw ref but signal not-found so the
        // caller can surface an appropriate error.
        return "", nil   // don't cache; caller receives ("", true) from Get
    }
    // Deterministic failure: cache it so we don't keep retrying, but
    // signal the caller that resolution failed.
    p.storeValue(reference, "")
    return "", errors.New("deterministic resolution failure")
}
```

At a minimum, the behavior change (silently returning empty string vs. `found=false`) warrants a note in the PR that callers which previously handled `found=false` from a 1Password provider will now silently receive empty strings.

docker-agent (commented):

Two medium-severity issues in the new caching/singleflight logic introduced by this PR.

pkg/environment/onepassword.go
docker-agent, line 84: **[HIGH] singleflight closure captures first caller's context; cancelled context poisons all coalesced waiters**

`singleflight.Group.Do` runs only the **first** goroutine's closure — secondary waiters block and share its result. Because the closure captures `ctx` from the first caller, if that context is cancelled or times out while `op` is running:

1. `p.resolve(ctx, reference)` fails due to the cancelled context
2. `ctx.Err() != nil` → caching is **skipped** (correct: transient)
3. `""` is returned to **all** coalesced waiters — including secondary callers whose own contexts are still valid

Those secondary callers had perfectly live contexts but receive an empty string, silently losing their secret value. There is no retry path because their contexts were never used.

**Suggested fix:** Pass a context that is not tied to any single request into the closure — for example use `context.WithoutCancel(ctx)` (Go 1.21+) or a long-lived background context stored on the provider, so resolution completes even if the triggering request is cancelled:

```go
resolved, _, _ := p.group.Do(reference, func() (any, error) {
    // Use a detached context so a single caller's cancellation does not
    // abort resolution for all coalesced waiters.
    resolveCtx := context.WithoutCancel(ctx)
    value, ok := p.resolve(resolveCtx, reference)
    ...
})
```

pkg/environment/onepassword.go
```diff
@@ -54,11 +70,52 @@ func (p *OnePasswordProvider) Get(ctx context.Context, name string) (string, boo
...
+			slog.WarnContext(ctx, "Failed to resolve 1Password secret reference; using empty value")
+			value = ""
+			// Don't cache failures caused by a cancelled/expired context: those
+			// are transient and a later call should be allowed to retry.
+			if ctx.Err() != nil {
+				return value, nil
+			}
+		}
+
+		p.storeValue(reference, value)
```
docker-agent, line 100: **[MEDIUM] Deterministic failures permanently cached as empty string with no TTL or retry**

When `p.resolve` returns `(_, false)` and `ctx.Err() == nil` (e.g. bad vault name, malformed reference), `""` is written to the cache and all future lookups for that reference return immediately without ever calling `op` again. In a long-running daemon this creates a permanent silent failure:

- `op` binary is not yet on `PATH` at startup → first lookup caches `""` → `op` is installed later → lookups never recover
- A typo in the vault path is fixed in config → lookups still return the old cached `""`

The operator has no signal that the secret is missing — the variable simply appears empty.

**Suggested mitigation:** At minimum, emit a `slog.WarnContext` or `slog.ErrorContext` when returning a **cached** empty value (not just on first failure), so it is visible in logs. A more robust option is to not cache `""` for infrastructure failures (binary missing) separately from reference failures, or to add a simple retry window (e.g. re-resolve after a fixed interval).

docker-agent (commented):
### Assessment: 🔴 CRITICAL

trungutt (approved):

PR #3323, opened 2026-06-30
fix: make chmod in Registry.Write best-effort to avoid breaking non-owner callers

dgageot:
When `Registry.Write` creates the runs directory it calls `os.Chmod` to tighten permissions to `0o700`. That worked fine when the directory was created fresh, but it regresses for callers that don't own the directory — for example, when a root-owned `<datadir>/runs` directory already exists and a privilege-dropped process tries to write into it. In that case `os.Chmod` returns `EPERM`, which previously caused every `Write` call to fail permanently.

The fix makes the chmod best-effort. `os.Stat` is checked first: if the directory already has `0o700` permissions the chmod is skipped entirely. If it can't be skipped and `os.Chmod` fails, the error is now logged via `slog.Warn` rather than surfaced as a hard error. A failed `os.Stat` is also logged so the caller has visibility into filesystem surprises. The comment was also softened to avoid over-claiming that a non-owned directory is unlikely to be readable by other users — that depends entirely on the umask and parent directory ACLs.

No breaking changes. The behavioral delta only affects the already-broken case where chmod was returning an error; callers that succeed today continue to behave identically.

### DIFF
--- a/pkg/runregistry/registry.go
+++ b/pkg/runregistry/registry.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"slices"
@@ -68,9 +69,15 @@ func (r *Registry) Write(rec Record) (func(), error) {
 	}
 	// MkdirAll only applies the mode to directories it creates, so an
 	// already-existing dir may be group/world-readable. Tighten it explicitly
-	// to keep live PIDs/addresses unreadable by other local users.
-	if err := os.Chmod(r.dir, 0o700); err != nil { //nolint:gosec // directory needs execute bit; 0700 is owner-only
-		return nil, fmt.Errorf("restricting run registry dir: %w", err)
+	// to keep live PIDs/addresses unreadable by other local users. This is
+	// best-effort: a dir we don't own (e.g. created by root before dropping
+	// privileges) can't be chmod'd, but such a dir is also unlikely to be
+	// readable by arbitrary local users, so we log and carry on rather than
+	// breaking Write permanently.
+	if info, err := os.Stat(r.dir); err == nil && info.Mode().Perm() != 0o700 {
+		if err := os.Chmod(r.dir, 0o700); err != nil { //nolint:gosec // directory needs execute bit; 0700 is owner-only
+			slog.Warn("Could not tighten run registry dir permissions", "dir", r.dir, "error", err)
+		}
 	}
 
 	path := filepath.Join(r.dir, strconv.Itoa(rec.PID)+".json")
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change correctly converts the hard-failing `os.Chmod` into a best-effort stat-then-chmod with structured `slog.Warn` logging. The logic is sound:

- `MkdirAll` errors are still propagated as hard failures (unchanged) ✅
- `info.Mode().Perm()` correctly masks to the low 9 bits — the `!= 0o700` comparison is accurate ✅
- `slog.Warn` calls use the structured key-value API correctly (`"dir"`, `"error"` keys) ✅
- The skip-if-already-correct optimization avoids unnecessary syscalls on the happy path ✅

Two low-severity observations (no action required):

1. **TOCTOU between `os.Stat` and `os.Chmod`**: A narrow window exists where another process could change directory permissions between the two calls. Given that `chmod` is now explicitly best-effort and the PR's stated design intent, the blast radius is minimal.

2. **`os.Stat` failure skips permission verification silently**: If `Stat` fails after a successful `MkdirAll`, the code logs a warning and continues. The subsequent `WriteAtomic` will surface any real filesystem problem — this is consistent with the best-effort framing and acceptable.

pkg/runregistry/registry.go
```diff
@@ -68,9 +69,17 @@ func (r *Registry) Write(rec Record) (func(), error) {
...
+	// to keep live PIDs/addresses unreadable by other local users. This is
+	// best-effort: a dir we don't own (e.g. created by root before dropping
+	// privileges) can't be chmod'd, so we log and carry on rather than breaking
+	// Write permanently — the directory's existing permissions still apply.
+	switch info, err := os.Stat(r.dir); {
+	case err != nil:
+		slog.Warn("Could not stat run registry dir to verify permissions", "dir", r.dir, "error", err)
+	case info.Mode().Perm() != 0o700:
+		if err := os.Chmod(r.dir, 0o700); err != nil { //nolint:gosec // directory needs execute bit; 0700 is owner-only
+			slog.Warn("Could not tighten run registry dir permissions", "dir", r.dir, "error", err)
```
docker-agent, line 81: **[MEDIUM] Warn log omits actual directory permissions, making the security state undiagnosable**

When `os.Chmod` fails (e.g. the process doesn't own the directory), `slog.Warn` is called but the log entry does not include the directory's *current* permissions. This makes it impossible for an operator to distinguish between two very different situations:

- **(a) benign** — the directory is already `0o700` (correct), owned by root, and a non-owner process just couldn't re-chmod it. No real security issue.
- **(b) insecure** — the directory is `0o755` or `0o777` (world-readable), and chmod failed because the caller doesn't own it. Live PIDs and socket addresses are now enumerable by any local user, silently violating the security invariant documented in the comment above.

Before this PR a failed chmod was a hard error, so callers could never proceed into a world-readable registry directory unknowingly. Now they can — but the log message gives operators no way to tell which case they're in.

**Suggested fix** — include the actual permissions in the warning:
```go
if err := os.Chmod(r.dir, 0o700); err != nil {
    slog.Warn("Could not tighten run registry dir permissions",
        "dir", r.dir,
        "current_perm", info.Mode().Perm(),  // ← add this
        "error", err)
}
```
This lets operators immediately see whether the directory is at a safe `0o700` or something more permissive.

docker-agent (commented):

trungutt (approved):

PR #3324, opened 2026-06-30
fix: eliminate themeRegistry data race and TOCTOU window

dgageot:
This is a follow-up to #3316, which extracted a `themeRegistry` struct for parallel-safe theme tests. Two issues were identified in review that warranted a targeted fix.

The first was a data race: the package-level `defaultRegistry` was a plain `var *themeRegistry`, so any goroutine reading through the exported wrapper functions while the package-API test swapped the pointer in a parallel test would trigger an unsynchronized write detectable under `go test -race`. `defaultRegistry` is now an `atomic.Pointer[themeRegistry]`; all exported wrappers call `.Load()` and the test uses `.Store()` to swap registries safely.

The second was a TOCTOU window in `themeRegistry.LoadTheme`: the `generation` snapshot was taken at cache-read time, *after* the `isBuiltin` classification. If a generation bump occurred between the two reads, the resulting cache entry could store a stale built-in classification under a fresh generation number. Moving the snapshot to before the `isBuiltin` check closes the window.

### DIFF
--- a/pkg/tui/styles/theme.go
+++ b/pkg/tui/styles/theme.go
@@ -61,8 +61,14 @@ func newThemeRegistry() *themeRegistry {
 	return &themeRegistry{cache: make(map[string]*themeCacheEntry)}
 }
 
-// defaultRegistry backs the package-level theme API.
-var defaultRegistry = newThemeRegistry()
+// defaultRegistry backs the package-level theme API. It is an atomic pointer so
+// tests can swap in an isolated registry without racing package-level callers
+// that read it through the exported wrapper functions.
+var defaultRegistry = func() *atomic.Pointer[themeRegistry] {
+	p := &atomic.Pointer[themeRegistry]{}
+	p.Store(newThemeRegistry())
+	return p
+}()
 
 // RegisterBuiltinThemes adds an additional source of built-in themes from fsys.
 // It must contain theme files at "themes/<ref>.yaml" (or .yml) — the same layout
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes correctly address both issues described in the PR.

** migration**: All exported wrapper functions now call  before use, and the test swaps the registry via  in a race-safe manner. The  restore is also atomic. No unsynchronized pointer reads remain in the changed lines.

**TOCTOU fix in **: Moving the generation snapshot to before the  classification is the right approach. If a concurrent  bumps the generation between the snapshot and the cache store, the store-time generation check () will detect the staleness and retry, so no stale entries can persist under a mismatched generation number.

No issues found in the introduced code.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The two targeted fixes introduced by this PR are correct and well-reasoned.

**atomic.Pointer migration** — Converting `defaultRegistry` from a plain `*themeRegistry` to `*atomic.Pointer[themeRegistry]` (initialised via an IIFE) eliminates the unsynchronised write that was detectable under `-race` when a test swapped the pointer. All seven exported wrapper functions (`RegisterBuiltinThemes`, `InvalidateThemeCache`, `ListThemeRefs`, `listBuiltinThemeRefs`, `LoadTheme`, `loadBuiltinTheme`, `IsBuiltinTheme`) correctly call `.Load()` before use, and the test cleanup correctly calls `.Store(previous)` to restore the original registry.

**TOCTOU fix in `LoadTheme`** — Moving the generation snapshot to before `IsBuiltinTheme` is the right approach. The snapshot now covers the classification call, so any concurrent `RegisterBuiltinThemes` that bumps the generation between the snapshot and the store-time check will be caught, forcing a retry with the refreshed classification. The two-lock pattern (one acquire for `generation`, a separate acquire for the cache read) leaves a narrow window, but analysis of `InvalidateThemeCache` — which replaces the entire cache map rather than clearing individual entries — shows that a stale cache hit through that window is not possible; the store-time `generation != generation` guard handles every interleaving correctly.

No bugs were found in the changes introduced by this PR.

trungutt (approved):

PR #3326, opened 2026-06-30
refactor: replace test-mutated globals with dependency injection

dgageot:
## What

Replaces several package-level globals that tests mutated at runtime with proper dependency injection. Tests no longer swap shared globals, so they can run with `t.Parallel()` safely and there is no hidden cross-test coupling.

## Why

We had a recurring pattern where production code reached for a package-level singleton (an HTTP client, a fetch function, a clock, a catalog loader) and tests overrode it via `var x = ...; x = stub; t.Cleanup(...)` or a dedicated `SetXForTest` helper. That pattern forces sequential tests, is racy under `-race` if a goroutine ever reads the global concurrently, and exercises a wiring that only exists in tests. Injecting the dependency removes the global entirely.

## Changes

Five globals converted (production behavior unchanged — the default still wires the original implementation):

- **`pkg/modelsdev`** — removed `var fetchCatalog`; added a `Store.fetch` field + `WithFetcher(...)` option. `loadDatabase` takes a `fetcher` parameter. Tests inject a stub fetcher.
- **`pkg/toolinstall`** — removed `var doInstallFn`; `resolve`/`safeInstall` take an `installFunc`, with `EnsureCommand` supplying the real `doInstall`.
- **`pkg/hooks/builtins`** — removed `var httpPostClient` and `SetHTTPPostClientUnsafeForTest`; `Register` now takes functional `Option`s (`WithHTTPPostClient(...)`), and `http_post` is a closure over an `httpDoer`. The SSRF-safe client remains the production default; only tests inject the unsafe client. Deleted `export_test.go` and `main_test.go`.
- **`pkg/tui/components/reasoningblock`** — replaced package-level `var nowFunc` with a per-instance `Model.now` clock field defaulting to `time.Now`.
- **`pkg/gateway`** — introduced a `Loader` type (memoization via `sync.Once` + injectable fetch), `NewLoader()`/`NewStaticLoader()`, and context plumbing `WithLoader`/`loaderFrom` with a package-level `defaultLoader`. Removed the `catalogOnce`/`catalogCached`/`catalogOverride` globals and deleted `testing.go`. Production call sites use the default loader via context transparently.

A follow-up commit hardens the new seams: nil-safe `WithLoader`/`loaderFrom`, a zero-value-safe `Loader.load` and `Store.fetcher()`, and a corrected godoc placement — all defensive, no behavior change.

## Validation

- `task lint` — golangci-lint, custom cops, and `go mod tidy` all clean
- `task test` — full suite passes
- `go test -race` on the affected packages passes; the converted tests now run `t.Parallel()`


docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor cleanly converts five package-level globals to dependency injection across `pkg/gateway`, `pkg/toolinstall`, `pkg/hooks/builtins`, `pkg/tui/components/reasoningblock`, and `pkg/modelsdev`. Production behaviour is preserved — every call site wires the original implementation as the default.

**What was reviewed:**
- Nil-safety of new zero-value code paths (`Store.fetcher()`, `Loader.load`, `loaderFrom`)
- Concurrency correctness of `sync.Once` memoization in the new `gateway.Loader`
- Singleflight cross-test contamination via the residual package-level `installGroup`
- Data-race exposure on `Model.now` in `reasoningblock`
- Default wiring correctness in all five refactored packages

**Findings:** All hypotheses examined were either non-issues (code paths that can never be reached) or correctly scoped (e.g., `TestResolve_ConcurrentPanic_DoesNotCrash` does not call `t.Parallel()`, so the shared `installGroup` cannot cause cross-test contamination in a standard `go test ./...` run). No CONFIRMED or LIKELY bugs were found in the introduced code.

Sayt-0 (approved):

PR #3327, opened 2026-06-30
fix(#3145): empty agent response from multiple system messages on OpenAI-compatible providers

Sayt-0:
## Summary

Agents backed by an OpenAI-compatible thinking provider (reported with OVHcloud's Qwen3.5-397B-A17B over `openai_chatcompletions`) could return an empty response and stop silently as soon as a toolset was attached. The root cause is that the provider receives more than one system message and replies with an empty stream.

## Root cause

docker-agent builds one system message per source:

- the agent instruction
- each toolset's instructions (`pkg/session/session.go`)
- a handoff prompt in multi-agent teams

OVHcloud's Qwen3.5 endpoint replies with `HTTP 200` and an immediate `data: [DONE]` (empty stream) whenever a request carries **more than one** system message. docker-agent records this as `tool_calls=0 content_length=0 finish_reason=null` and skips the empty assistant message, so the agent appears to do nothing.

Bisected live against the real endpoint. The system-message count was the only variable that flipped the behaviour:

| Request | Result |
|---|---|
| 2 system messages | empty `[DONE]` |
| 2 system messages merged into 1 | streams normally |
| 1 system message | streams normally |
| 9 tools, no system message | streams normally |
| union-type tool schemas / `parallel_tool_calls` toggled | no effect |

Note: the endpoint also emits non-standard `delta.reasoning` and `delta.name` fields, but the stream adapter parses content correctly, so those are not the cause.

## The fix

`convertMessages` on the chat-completions path now calls `oaistream.MergeConsecutiveMessages`, coalescing consecutive same-role (system/user) messages into one. The DMR client already does this for the same reason; the chat-completions path was the only one that skipped it. The change is semantically equivalent for providers that accept multiple system messages (OpenAI, Azure, and so on).

## Also in this PR

`fix(runtime): surface empty/reasoning-only turns ...` keeps a safety net: when a turn genuinely comes back empty (rate limit, output token limit, reasoning-only), a warning is emitted instead of stopping silently.

## How it was tested

| Step | Result |
|---|---|
| Live reproduction on OVHcloud Qwen3.5-397B-A17B with the reporter's config + filesystem toolset | empty response, reproduced |
| Same config after the fix | agent calls `list_directory` then `read_file` then answers; 3 requests, 1 system message each |
| New unit test `TestChatCompletions_MergesConsecutiveSystemMessages` | asserts 2 system messages collapse to 1 on the wire |
| `go test ./pkg/model/... ./pkg/runtime/...` | green |

## Out of scope / follow-up

Session-title generation can still come back empty on OVHcloud's Qwen3.5: thinking cannot be disabled there via `chat_template_kwargs` (rejected with `HTTP 400`) and is unbounded, so a `max_tokens=20` title call spends its whole budget on reasoning. The only working lever found is `reasoning_effort: "none"`. That is a separate, cosmetic issue left for a follow-up.

Re #3145


pkg/runtime/loop.go
```diff
@@ -721,6 +721,36 @@ func (r *LocalRuntime) runTurn(
 	if res.FinishReason == chat.FinishReasonRefusal {
 		slog.WarnContext(ctx, "Model refused to respond", "agent", a.Name(), "model", modelID.String(), "session_id", sess.ID)
 		events.Emit(Warning(fmt.Sprintf("Model %s refused to respond (stop reason: refusal).", modelID.String()), a.Name()))
+	} else if strings.TrimSpace(res.Content) == "" && len(res.Calls) == 0 {
```
docker-agent, line 724: **[MEDIUM] Empty reasoning-only turn emits warning but continues looping — potential infinite loop**

When a thinking-mode model (e.g. OVHcloud Qwen3) returns an empty turn with `finish_reason=null` and `res.Stopped=false`, the new `else if` block (lines 724–753) correctly emits a warning, but does not `return turnExit`. Execution falls through to:

1. `recordAssistantMessage` — no-op (empty turn is skipped)
2. `processToolCalls(… res.Calls …)` — no-op (empty slice)
3. The `res.Stopped` check is skipped because `Stopped == false`
4. `return turnContinue` — re-enters the model call with **identical messages**

The loop detector (`ls.loopDetector`) only fires on repeated identical *tool calls*, so it provides no protection here. The outer `runStreamLoop` only exits on context cancellation or a `turnExit` signal. A provider that persistently returns empty, non-stopped responses (exactly the OVHcloud scenario this PR targets) would therefore spin forever, emitting a warning on each iteration and incurring unbounded API cost, until an external context deadline fires.

Consider returning `turnExit` from inside the empty-turn block (or adding a consecutive-empty-turn counter with a threshold) so the loop self-terminates:

```go
} else if strings.TrimSpace(res.Content) == "" && len(res.Calls) == 0 {
    // ... build warning string ...
    events.Emit(Warning(warning, a.Name()))
    endReason = turnEndReasonEmptyTurn
    return turnExit  // don't re-enter with identical messages
}
```

docker-agent (commented):

The fix correctly addresses the root cause (multiple system messages on chat-completions path) and the `MergeConsecutiveMessages` merge logic is sound for the homogeneous-format case docker-agent actually emits (all-string system messages). One concern was found in the new empty-turn handling in `loop.go`.

**Reviewed**: pkg/cli/printer.go · pkg/cli/runner.go · pkg/model/provider/openai/client.go · pkg/model/provider/openai/system_message_merge_test.go · pkg/runtime/loop.go · pkg/runtime/runtime_test.go

pkg/runtime/loop.go
Sayt-0: Thanks, the underlying fragility is real, but the specific scenario as described does not actually loop.

For a truly empty turn (no content, no tool calls, including the OVHcloud bare `[DONE]` case this PR targets), `handleStream` sets `Stopped=true` on both return paths: the early return uses `len(toolCalls)==0`, and the bare-EOF return uses `fullContent.Len()==0 && len(toolCalls)==0`. So `res.Stopped` is true whenever the empty-turn warning fires, and `runTurn` takes the `if res.Stopped` exit instead of falling through to `turnContinue`. Verified live against OVHcloud Qwen3.5-397B-A17B: the empty turn ended the run after a single request, no loop.

There was a narrower real hole though: whitespace-only content with a bare EOF. The warning uses a trimmed-empty check while the no-output test used `Len()`, so `"\n\n"` gave `Stopped=false` while the warning still fired, which would re-enter. Fixed in 0f612b48 by aligning the no-output test with the trimmed-empty check, plus a `handleStream` regression test.

docker-agent: I appreciate the detailed correction! You're absolutely right — I missed that `handleStream` already sets `Stopped=true` for the truly empty cases this PR targets, so the loop scenario I described doesn't actually occur in practice.

The whitespace-only gap you identified is the real issue: content like `"\n\n"` would satisfy `fullContent.Len() > 0` (keeping `Stopped=false`) but fail the `strings.TrimSpace(res.Content) == ""` check in the warning block, creating the exact re-entry path I was worried about. Your fix in 0f612b48 to align both checks on the trimmed-empty condition closes that hole cleanly, and the regression test ensures it stays closed.

Thanks for the clarification and the fix — the fragility is now addressed at the right layer.

dgageot (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR is a clean, well-scoped fix. The two changes work together correctly:

**System-message merge** ():  is properly gated behind the `openai_chatcompletions` api_type check, leaving first-class providers (OpenAI, Azure, Mistral, etc.) untouched. The DMR client already applied this merge, so the chat-completions path is now consistent.

**Empty-turn surfacing** (, ): The `strings.TrimSpace` alignment between `stoppedDueToNoOutput` in `streaming.go` and the new empty-turn guard in `loop.go` ensures whitespace-only responses always set `Stopped=true`, preventing re-entry. The new warning branch is correctly placed in the `else if` chain (after the refusal branch), so refusals and empty turns are reported distinctly without overlap. The `WarningEvent` is handled in `runner.go` for non-TUI runs.

No logic errors, race conditions, or edge case gaps found in the introduced changes.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR is a clean, well-scoped fix. The two changes work together correctly:

**System-message merge** (`pkg/model/provider/openai/client.go`): `MergeConsecutiveMessages` is properly gated behind the `openai_chatcompletions` api_type check, leaving first-class providers (OpenAI, Azure, Mistral, etc.) untouched. The DMR client already applied this merge, so the chat-completions path is now consistent.

**Empty-turn surfacing** (`pkg/runtime/loop.go`, `pkg/runtime/streaming.go`): The `strings.TrimSpace` alignment between `stoppedDueToNoOutput` in `streaming.go` and the new empty-turn guard in `loop.go` ensures whitespace-only responses always set `Stopped=true`, preventing re-entry. The new warning branch is correctly placed in the `else if` chain (after the refusal branch), so refusals and empty turns are reported distinctly without overlap. The `WarningEvent` is handled in `runner.go` for non-TUI runs.

No logic errors, race conditions, or edge case gaps found in the introduced changes.

PR #3330, opened 2026-06-30
refactor: replace single-variable mutexes with atomic types

dgageot:
Each of the three structs changed here used a `sync.Mutex` or `sync.RWMutex` to protect exactly one mutable field — all other fields were written once at construction and never changed afterward. When a mutex guards a single variable, the standard library's `sync/atomic` package offers a simpler and more self-documenting fit: the data and its synchronization are expressed as one value rather than two separate declarations.

`lspSession` in `pkg/tools/builtin/lsp/lsp_lifecycle.go` had a `closed bool` protected by a plain mutex. It is now an `atomic.Bool`, with `CompareAndSwap` in `Close()` to preserve idempotency and `Load()` in the wait goroutine to check the flag. `Client` in `pkg/model/provider/rulebased/client.go` had a `lastSelectedID` of type `modelsdev.ID` behind an `RWMutex`; it is now an `atomic.Pointer[modelsdev.ID]`. `agentRouter` in `pkg/runtime/agent_router.go` had a `current string` behind an `RWMutex`; it is now an `atomic.Pointer[string]`.

All changes are behavior-preserving. The suite passes under `go test -race` and the linter reports no new issues.

### DIFF
--- a/pkg/model/provider/rulebased/client.go
+++ b/pkg/model/provider/rulebased/client.go
@@ -13,7 +13,7 @@ import (
 	"fmt"
 	"log/slog"
 	"slices"
-	"sync"
+	"sync/atomic"
 
 	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/config/latest"
@@ -43,11 +43,13 @@ type ProviderFactory func(ctx context.Context, modelSpec string, models map[stri
 type Client struct {
 	base.Config
 
-	routes         []Provider
-	fallback       Provider
-	matcher        *matcher
-	mu             sync.RWMutex
-	lastSelectedID modelsdev.ID // ID of the provider selected by the most recent call
+	routes   []Provider
+	fallback Provider
+	matcher  *matcher
+	// lastSelectedID holds the ID of the provider selected by the most recent
+	// call. A plain atomic pointer suffices because it is the only mutable
+	// field; everything else is set once at construction and read-only after.
+	lastSelectedID atomic.Pointer[modelsdev.ID]
 }
 
 // NewClient creates a new rule-based routing client.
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor correctly replaces single-variable mutexes with atomic types across all three files:

- **** →  with  in  for idempotency and  in the wait goroutine — both uses are correct.
- **** →  — nil checks are in place before dereferencing the loaded pointer.
- **** →  in  — nil is handled correctly.

The removed mutexes each guarded exactly one mutable field (as described), so no other fields are left unprotected. The changes are behavior-preserving and idiomatic Go.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor correctly replaces single-variable mutexes with atomic types across all three files. No bugs were found in the changed code.

- **`lspSession.closed` → `atomic.Bool`**: `CompareAndSwap(false, true)` in `Close()` correctly preserves idempotency; `Load()` in the wait goroutine is the right read primitive.
- **`lastSelectedID` → `atomic.Pointer[modelsdev.ID]`**: nil is handled before dereferencing the loaded pointer.
- **`current` → `atomic.Pointer[string]` in `agentRouter`**: nil is handled correctly on load.

The removed mutexes each guarded exactly one mutable field, so no other fields are left without synchronization. The changes are behavior-preserving and idiomatic Go.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor correctly replaces mutex-protected single fields with `sync/atomic` types across all three files.

**`lspSession.closed` → `atomic.Bool`**: The `CompareAndSwap(false, true)` in `Close()` correctly preserves idempotency. The happens-before ordering (`CAS → processCancel → process exit → cmd.Wait()`) ensures the wait goroutine observes `closed=true` after a signal-induced exit.

**`Client.lastSelectedID` → `atomic.Pointer[modelsdev.ID]`**: Taking the address of the local `selectedID` variable correctly heap-escapes it via the Go compiler. The nil-check on `Load()` properly mirrors the old zero-value behavior on first call.

**`agentRouter.current` → `atomic.Pointer[string]`**: `newAgentRouter` always calls `Store` before the struct is shared, and the nil guard in `Name()` is a correct and harmless safety net.

No bugs introduced. All changes are behavior-preserving.

Sayt-0 (approved):

PR #3332, opened 2026-06-30
refactor: replace test-overridden globals with parallel-safe injection

dgageot:
## What

Removes a cluster of mutable package-level global variables that tests were overriding in place. Overriding a shared global forces the tests that do it to run serially; replacing each global with either an explicit parameter or a field on the owning struct lets those tests run with `t.Parallel()` and no shared-state mutation.

This is a behavior-preserving refactor — production defaults are unchanged. Each commit is independently green (build, `go test -race`, `golangci-lint`, and the repo's custom cops).

## Why

Several packages carried `var x = …` globals whose only reason for being a `var` (not a `const` or a struct field) was so a test could reassign them via `orig := x; x = …; t.Cleanup(func(){ x = orig })`. That pattern:

- is not race-safe and silently blocks `t.Parallel()` on the affected tests;
- spreads a half-dozen slightly different idioms for the same need;
- leaves test-only seams (and the occasional exported `Reset*ForTest`) in production code.

## How

Two consistent patterns, matching what the codebase already does well elsewhere (`pkg/userid.Resolver`, `pkg/paths`):

1. **Explicit parameter** when the consumer is a free function.
2. **Field on the owning struct + a zero-value-means-default accessor** when the consumer is a method, so existing struct literals (including hand-built test ones) keep working untouched.

## Commits

| Commit | Package | Global removed | Pattern | Tests unblocked |
|--------|---------|----------------|---------|-----------------|
| `refactor(toolinstall)` | `pkg/toolinstall` | 4 × `max*` limit vars | `limits` value type + `Registry.extractor()` accessor | 3 |
| `refactor(runtime)` | `pkg/runtime` | `defaultStreamIdleTimeout` | const default + `handleStream` parameter | 2 |
| `refactor(mcp)` | `pkg/tools/mcp` | `unmanagedOAuthWaitTimeout` | const default + per-`oauthTransport` field | 1 |
| `refactor(notification)` | `pkg/tui/components/notification` | `timerDuration` | `Manager.autoHideDuration` field, defaulted in `New()` | 6 |
| `refactor(tui)` | `pkg/tui` | `exitFunc`, `shutdownTimeout` | per-`appModel` fields + accessors | 10 |
| `refactor(cmd/root)` | `cmd/root` | `loadUserConfig` | `userConfigLoader` parameter; tests inject `MapEnvProvider` | gateway/default-model tests |

**~22 tests** that previously had to run serially now run in parallel.

## Safety notes

- The zip/tar-bomb extraction bounds, the OAuth wait timeout, and the stream idle timeout all keep their exact production defaults.
- The "zero-value-means-default" accessors were checked against every production construction site: `Registry` is only built via `NewRegistry()` and `Manager` only via `New()`, so the default is always applied in production. The zero-value fallback only ever benefits test literals.

## Test plan

- `task build`
- `go test -race ./...` for each touched package
- `task lint` (golangci-lint + 18 custom cops) — clean

### DIFF
--- a/cmd/root/flags.go
+++ b/cmd/root/flags.go
@@ -26,7 +26,7 @@ const (
 )
 
 func addRuntimeConfigFlags(cmd *cobra.Command, runConfig *config.RuntimeConfig) {
-	addGatewayFlags(cmd, runConfig)
+	addGatewayFlags(cmd, runConfig, userconfig.Load)
 	cmd.PersistentFlags().StringSliceVar(&runConfig.EnvFiles, "env-from-file", nil, "Set environment variables from file")
 	cmd.PersistentFlags().BoolVar(&runConfig.GlobalCodeMode, "code-mode-tools", false, "Provide a single tool to call other tools via Javascript")
 	cmd.PersistentFlags().StringVar(&runConfig.WorkingDir, "working-dir", "", "Set the working directory for the session (applies to tools and relative paths)")
@@ -73,11 +73,12 @@ func canonize(endpoint string) string {
 	return strings.TrimSuffix(strings.TrimSpace(endpoint), "/")
 }
 
-// loadUserConfig is the function used to load user configuration.
-// It can be overridden in tests.
-var loadUserConfig = userconfig.Load
+// userConfigLoader loads the user configuration. Production passes
+// [userconfig.Load]; tests inject a stub so they neither touch the real
+// config file nor share a mutable package global (and can run in parallel).
+type userConfigLoader func() (*userconfig.Config, error)
 
-func addGatewayFlags(cmd *cobra.Command, runConfig *config.RuntimeConfig) {
+func addGatewayFlags(cmd *cobra.Command, runConfig *config.RuntimeConfig, loadUserConfig userConfigLoader) {
 	cmd.PersistentFlags().StringVar(&runConfig.ModelsGateway, flagModelsGateway, "", "Set the models gateway address")
 
 	persistentPreRunE := cmd.PersistentPreRunE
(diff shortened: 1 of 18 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, well-structured refactor that replaces mutable package-level globals with injected parameters and struct fields, enabling parallel test execution without any behavioral regressions.

**What was checked:**
- `pkg/toolinstall/registry.go` — `extractor()` zero-value-means-default accessor: the `limits{}` zero check correctly distinguishes "unset" from "configured", and the comment explicitly documents the design. No test constructs partial limits, so the theoretical edge case does not apply. ✅
- `pkg/tui/components/notification/notification.go` — `Manager.autoHideDuration` zero-value for tests: the comment documents that tests intentionally use zero duration for prompt timer resolution, while all production code goes through `New()` which sets `defaultDuration`. ✅
- `pkg/runtime/streaming.go` — `defaultStreamIdleTimeout` replaced with const + parameter: production defaults preserved. ✅
- `pkg/tools/mcp/oauth.go` — `unmanagedOAuthWaitTimeout` moved to `oauthTransport` field with zero-value accessor: production default preserved. ✅
- `pkg/tui/tui.go` — `exitFunc` and `shutdownTimeout` moved to `appModel` fields with accessor defaults: production behavior unchanged. ✅
- `cmd/root/flags.go` / `cmd/root/models.go` — `loadUserConfig` replaced with injected `userConfigLoader`: production path still calls the real loader. ✅

No bugs were introduced by this PR.

Sayt-0 (approved):

Issue #2611, opened 2026-05-01
File picker `alt+h` / `alt+i` shortcuts not working on macOS inside `sbx` — Option key intercepted before reaching the sandbox


steilerDev:
## Summary

The file picker shortcuts for toggling hidden files (`alt+h`) and ignored files (`alt+i`) are silently inoperative when running in `sbx`, regardless of keyboard layout.

## Environment

- **Host OS:** macOS (any version, any keyboard layout)
- **Terminal:** Terminal.app, GhosTTY (but probably others too)
- **Agent:** Running inside Docker sandbox via `sbx run`

## Steps to Reproduce

1. Start the agent on macOS
2. Trigger a file picker prompt
3. Press **Option+H** or **Option+I**
4. Observe that hidden/ignored file visibility does not toggle

## Root Cause

On macOS, the OS intercepts Option+key combinations at the input system level — across **all keyboard layouts** — and substitutes Unicode characters before the terminal emulator ever receives the event. The terminal therefore sends UTF-8 bytes, not ESC+letter sequences.

| Key combo | What a Linux terminal sends | What macOS actually sends |
|-----------|----------------------------|--------------------------|
| `Option+H` | `ESC h` → decoded as `alt+h` | `˙` (U+02D9) |
| `Option+I` | `ESC i` → decoded as `alt+i` | `ˆ` (U+02C6, dead key) |

This is not specific to US International PC — the standard **US layout** produces the same characters. Every macOS keyboard layout uses Option as a compose/AltGr layer. The only exception is users who have explicitly enabled "Use Option as Meta key" in their terminal preferences, which breaks Option-layer character input for those users.

The Kitty keyboard protocol cannot help here: the OS-level substitution happens before the bytes reach the terminal, regardless of protocol.

In `pkg/tui/dialog/file_picker.go`, the current detection:

```go
case msg.String() == "alt+h":
    d.toggleHidden()
case msg.String() == "alt+i":
    d.toggleIgnored()
```

…never matches because `msg.String()` returns `"˙"` / `"ˆ"` instead.

## Why Ctrl and Cmd Don't Fix This

**`ctrl+h` / `ctrl+i`** — not viable. These are hardwired at the ASCII level: Ctrl+H = `0x08` (Backspace) and Ctrl+I = `0x09` (Tab). They cannot be distinguished from the Backspace and Tab keys in a standard terminal, and both are almost certainly already used by the file picker for navigation.

[...]

PR #3333, opened 2026-06-30
fix(tui): accept macOS Option-key chars for file picker toggles

Sayt-0:
## Summary

The file picker `alt+h` / `alt+i` visibility toggles never fire on macOS. The OS substitutes a Unicode character for Option+key at the input-system level before the terminal receives the event, so `msg.String()` returns the substituted character instead of `alt+h` / `alt+i`.

This implements Option B from #2611: keep `alt+h` / `alt+i` as the primary bindings and accept the macOS-produced characters as aliases. Addresses #2611.

## Change

| Area | Before | After |
|---|---|---|
| Binding storage | inline `msg.String() == "alt+h"` checks | `filePickerKeyMap` struct (mirrors `costDialogKeyMap`) |
| Toggle hidden | `alt+h` only | `alt+h` and `˙` (U+02D9, macOS Option+H) |
| Toggle ignored | `alt+i` only | `alt+i` and `ˆ` (U+02C6, macOS Option+I) |
| Matching | string compare | `key.Matches` |

## Issue expectations mapped

| Expectation from #2611 | Status |
|---|---|
| Keep `alt+h` / `alt+i` working (Linux, "Option as Meta" enabled) | yes, primary bindings unchanged |
| Accept the macOS-substituted chars as aliases | yes, U+02D9 / U+02C6 added |
| Follow the existing `key.NewBinding` / keymap style | yes, `filePickerKeyMap` + `defaultFilePickerKeyMap()` |
| Backward compatible, low risk | yes, additive only |

## Known limitations (as noted in the issue's Option B section)

- Covers the standard US and US-International layouts only. Other national layouts emit different Unicode characters for Option+H / Option+I and are not handled.
- `ˆ` (U+02C6) is a dead key on macOS, so Option+I may compose with the next keystroke rather than arrive as a standalone press in some terminals.
- Option A (modifier-free keys such as `.` and `!`) remains the cleaner long-term fix and is independent of this change. Option B is the low-risk patch the issue proposed to apply immediately while Option A is discussed.

## Tests

- `TestFilePickerToggleHiddenViaMacOSOptionChar` and `TestFilePickerToggleIgnoredViaMacOSOptionChar` simulate the macOS `KeyPressMsg` and assert the toggles fire.
- The existing `alt+h` / `alt+i` tests continue to pass.
- The new tests assert against the `KeyPressMsg` the issue documents macOS producing (`Text` set to U+02D9 / U+02C6); they exercise the binding wiring, not a hardware-captured event. Removing the aliases makes the new tests fail while the `alt+h` / `alt+i` tests still pass, confirming they are genuine regression guards.

## Validation

`go build ./...`, `gofmt`, `go vet`, `golangci-lint` (0 issues), and the full `pkg/tui/dialog` test suite pass.

### DIFF
--- a/pkg/tui/dialog/file_picker.go
+++ b/pkg/tui/dialog/file_picker.go
@@ -44,9 +44,39 @@ var filePickerLayout = pickerLayout{
 	ListStartOffset: filePickerListStartY,
 }
 
+// filePickerKeyMap holds the file-picker-specific visibility toggles, on top
+// of the navigation bindings provided by the shared pickerKeyMap.
+//
+// On macOS the OS substitutes a Unicode character for Option+key at the
+// input-system level before the terminal sees the event, so the alt+h / alt+i
+// ESC sequences never reach us. Each toggle therefore also accepts the
+// characters that the standard US and US-International layouts emit (Option+H
+// -> "˙" U+02D9, Option+I -> "ˆ" U+02C6). Option+I is a dead key on macOS, so
+// the first press may arrive as an escaped or combining circumflex. See issue
+// #2611. The visible help labels are rendered separately (and dynamically) by
+// filePickerHelpKeysRows.
+type filePickerKeyMap struct {
+	ToggleHidden  key.Binding
+	ToggleIgnored key.Binding
+}
+
+func defaultFilePickerKeyMap() filePickerKeyMap {
+	return filePickerKeyMap{
+		ToggleHidden: key.NewBinding(
+			key.WithKeys("alt+h", "˙", "alt+˙"),
+			key.WithHelp("alt+h", "toggle hidden"),
+		),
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes look correct and well-structured. The `filePickerKeyMap` struct and `defaultFilePickerKeyMap()` constructor follow the existing keymap pattern. The Unicode code points U+02D9 (`˙`) for Option+H and U+02C6 (`ˆ`) for Option+I are correct for macOS US/US-International layouts, and the switch from direct `msg.String()` comparisons to `key.Matches` is the right approach. Both the production code and the tests are sound — no bugs introduced by this PR.

melmennaoui (approved):

PR #3334, opened 2026-06-30
ci: add concurrency group to pr-review-trigger to prevent duplicate reviews

derekmisler:
Adds a concurrency group keyed by PR number to `pr-review-trigger.yml` to prevent
triplicate reviews when simultaneous `pull_request` events fire for the same fork PR
(e.g., multiple `review_requested` events when several reviewers are added at once).

Also skips Bot-type senders (Dependabot, Renovate) early to save Actions minutes.

### DIFF
--- a/.github/workflows/pr-review-trigger.yml
+++ b/.github/workflows/pr-review-trigger.yml
@@ -7,8 +7,18 @@ on:
 
 permissions: {}
 
+# Deduplicate simultaneous pull_request events for the same fork PR.
+# When reviewers are requested at the same time, GitHub fires multiple
+# review_requested events. Without this group each event triggers a
+# separate review via workflow_run, producing duplicate reviews.
+concurrency:
+  group: pr-review-trigger-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
   save-context:
+    # Only run on fork PRs; skip GitHub App bot accounts (Dependabot, Renovate, etc.) early.
+    if: github.event.pull_request.head.repo.fork && github.event.sender.type != 'Bot'
     runs-on: ubuntu-latest
     steps:
       - name: Save event context


.github/workflows/pr-review-trigger.yml
```diff
@@ -7,8 +7,18 @@ on:
 
 permissions: {}
 
+# Deduplicate simultaneous pull_request events for the same fork PR.
+# When reviewers are requested at the same time, GitHub fires multiple
+# review_requested events. Without this group each event triggers a
+# separate review via workflow_run, producing duplicate reviews.
+concurrency:
+  group: pr-review-trigger-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
```
docker-agent, line 16: **[MEDIUM] `cancel-in-progress: true` can cancel a mid-flight artifact upload, silently dropping a review trigger**

The concurrency group `pr-review-trigger-${{ github.event.pull_request.number }}` with `cancel-in-progress: true` is broader than intended. The goal is to deduplicate *simultaneous* same-type events (e.g., multiple `review_requested` events fired at once), but it also cancels any in-progress run for the same PR number when *any* new event fires — including a `pull_request_review_comment` event arriving while an `opened`/`review_requested` run is mid-flight uploading its artifact.

Since the artifact upload is the final step, a late cancellation mid-upload means the downstream `workflow_run` consumer never receives the artifact and the review is silently dropped, with no error surfaced to the PR author.

**Suggested fix:** Scope the concurrency key to include the event type so that only duplicate events of the same type cancel each other:
```yaml
concurrency:
  group: pr-review-trigger-${{ github.event.pull_request.number }}-${{ github.event_name }}
  cancel-in-progress: true
```
Or use `cancel-in-progress: false` to queue rather than cancel, accepting that rare near-simultaneous duplicates may both run.

docker-agent (commented):

The PR correctly addresses the duplicate-review problem caused by simultaneous `review_requested` events. The concurrency group and bot-skip guard are well-motivated. One issue with the `cancel-in-progress: true` scope warrants attention before merging.

Sayt-0 (approved):

.github/workflows/pr-review-trigger.yml
```diff
@@ -7,8 +7,18 @@ on:
 
 permissions: {}
 
+# Deduplicate simultaneous pull_request events for the same fork PR.
+# When reviewers are requested at the same time, GitHub fires multiple
+# review_requested events. Without this group each event triggers a
+# separate review via workflow_run, producing duplicate reviews.
+concurrency:
+  group: pr-review-trigger-${{ github.event.pull_request.number }}
```
docker-agent, line 15: **[MEDIUM] Concurrency group conflates `pull_request` and `pull_request_review_comment` events — a comment-triggered run may be cancelled by a PR status event**

The new concurrency key `pr-review-trigger-${{ github.event.pull_request.number }}` applies to **all** event types this workflow handles. Both `pull_request` and `pull_request_review_comment` events populate `github.event.pull_request.number`, so they share the same group. With `cancel-in-progress: true`, a `pull_request` event (e.g. `ready_for_review`) arriving while a `pull_request_review_comment` run is in progress will silently cancel it — the downstream review workflow never gets triggered for that comment.

The PR's stated goal is deduplicating simultaneous `review_requested` events, but the current key is too broad. Including the event name in the key avoids cross-event cancellation while still deduplicating same-type events:

```yaml
concurrency:
  group: pr-review-trigger-${{ github.event_name }}-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

This way multiple `review_requested` events still collapse into one run, but a comment-triggered run is never cancelled by an unrelated PR status event.

docker-agent (commented):

One confirmed issue was found in the new concurrency configuration introduced by this PR.

PR #3337, opened 2026-06-30
feat: add OpenRouter provider

dgageot:
OpenRouter is an increasingly popular gateway that exposes hundreds of models through a single OpenAI-compatible endpoint, but users had to configure it manually as a custom provider. This change registers it as a first-class built-in alias so the experience is as smooth as Anthropic or Groq.

When `OPENROUTER_API_KEY` is set, the provider is auto-detected, its base URL (`https://openrouter.ai/api/v1`) is resolved automatically, and it surfaces in default/env/sandbox paths the same way every other built-in provider does. The fake proxy gains a URL-rewrite rule so that paths containing `/api` are preserved correctly — previously the rewrite stripped the `/api` segment, breaking OpenRouter requests in tests. A new example config (`examples/openrouter.yaml`) and a dedicated docs page round out the addition so users can get started without reading source code.

No breaking changes. Existing custom-provider configurations that pointed at `https://openrouter.ai/api/v1` continue to work unchanged.


pkg/sandbox/sandbox.go
```diff
@@ -299,6 +299,7 @@ var proxyManagedEnvVars = []string{
 	"ANTHROPIC_API_KEY",
 	"GOOGLE_API_KEY",
 	"MISTRAL_API_KEY",
+	"OPENROUTER_API_KEY",
```
docker-agent, line 302: **[MEDIUM] OPENROUTER_API_KEY added to proxyManagedEnvVars — verify Docker Desktop proxy support exists**

`proxyManagedEnvVars` is an exclusion list: keys in it are **never forwarded** to the sandbox container because Docker Desktop's built-in proxy is expected to inject them automatically. Adding `OPENROUTER_API_KEY` here silently drops the key from the sandbox environment unless Docker Desktop already handles it.

If Desktop proxy support for OpenRouter is not yet live, every sandbox run that uses OpenRouter will fail with `401 Unauthorized` with no clear indication of the cause — the key just won't be present inside the container.

All the other entries in the list (OpenAI, Anthropic, Google, Mistral, xAI, Nebius) have confirmed Desktop proxy backing. Before merging, confirm that the same support exists for `OPENROUTER_API_KEY`, or consider deferring this addition until Desktop proxy coverage is verified.

docker-agent (commented):

The OpenRouter provider integration is well-structured and consistent with other providers across the codebase. One area to verify before merging: the sandbox env-var handling.

Sayt-0 (approved):

PR #3338, opened 2026-06-30
feat: add per-model bypass_models_gateway option

dgageot:
When most models in a deployment route through a central gateway (`--models-gateway` / `CAGENT_MODELS_GATEWAY`), there are cases where one specific model must connect directly to its provider — for example, a private fine-tune hosted on OpenAI that cannot be accessed through the gateway, or a model whose provider credentials differ from those the gateway holds. Until now, the gateway setting was all-or-nothing per agent invocation.

This adds a `bypass_models_gateway` boolean field to the per-model config in the agent YAML. Setting it to `true` on a model makes that model skip the gateway entirely and connect directly to its provider, using that provider's own credentials (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `token_key`, etc.). The rest of the agent's models continue routing through the gateway as before. Bypass is propagated transparently through router models: a bypass-flagged routing model passes the flag to all of its routed targets, so neither the config author nor the runtime needs to repeat the flag on every leaf. The combination of `bypass_models_gateway: true` with `first_available` is rejected at validation time because it has no coherent semantics.

The security implication of this field is noted in the type definition: on an untrusted config, a malicious `base_url` combined with this flag could be used to exfiltrate provider credentials to an attacker-controlled endpoint. Users should only enable this on configs they control.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `bypass_models_gateway` implementation is correct and well-structured. The drafter attempted analysis of the diff and I performed direct code analysis to verify four areas of concern — all were dismissed.

**Areas reviewed:**

- **`pkg/model/provider/factory.go`** — The functional-options pattern correctly appends `WithGateway("")` after the existing opts so the last writer wins; the gateway-bypass logic fires only when a gateway is actually configured; the routing model propagation and leaf model paths are both handled consistently.
- **`pkg/config/gather.go`** — The `bypassOnly` flag is propagated correctly through the `gatherEnvVarsForModel` call chain. Named routing targets with their own `BypassModelsGateway` flag are handled. Non-recursion into sub-routers is safe because `resolveRoutedModel` explicitly rejects models-with-routing-rules as routing targets at validation time.
- **`pkg/config/latest/validate.go`** — The `validateFirstAvailable` rejection of `bypass_models_gateway` is correctly placed; `first_available` candidates are plain strings so they cannot carry the flag on the parent config entry.
- **`pkg/config/latest/types.go`** — The `isShorthandOnly()` guard correctly includes `!f.BypassModelsGateway` so a model with bypass set serializes as a full config object rather than a shorthand string.

Sayt-0 (approved):

PR #3341, opened 2026-07-01
feat: add Baseten provider support

dgageot:
Baseten is a model deployment platform that exposes an OpenAI-compatible API, but its base URL and model naming conventions differ enough that users previously had to configure everything manually. This PR adds first-class support for Baseten as a named provider.

The `baseten` provider alias is registered alongside the existing OpenAI-compatible aliases, so users can write `provider: baseten` in their agent config and get the correct base URL and default model (`mistral-7b-instruct`) automatically. The factory picks up the `BASETEN_API_KEY` environment variable and the provider is recognized by the auto-detection logic. Message merging (consecutive same-role consolidation) is enabled for Baseten, matching the behavior required by its API. Streaming responses from Baseten include usage data in the final chunk, so the `oaistream` adapter is updated to capture and propagate that usage correctly.

No breaking changes. The `bypass_models_gateway` option works with Baseten the same way it does with other OpenAI-compatible providers.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -182,7 +182,7 @@
       "properties": {
         "provider": {
           "type": "string",
-          "description": "The underlying provider type. Defaults to \"openai\" when not set. Supported values: openai, anthropic, google, amazon-bedrock, dmr, and any built-in alias (requesty, openrouter, azure, xai, ollama, mistral, etc.).",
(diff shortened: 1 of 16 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The Baseten provider integration is well-structured and consistent. All new code, documentation, and examples agree on `deepseek-ai/DeepSeek-V3.1` as the default model. The `oaistream` adapter change correctly handles the case where Baseten attaches usage data to normal content chunks (rather than a final usage-only chunk), preventing premature stop-event synthesis. The `shouldMergeConsecutiveMessages` helper cleanly expresses the intent for message consolidation without touching first-class providers.

aheritier (approved):

PR #3366, opened 2026-07-01
feat: collapse file picker shortcuts to one line on wide dialogs

dgageot:
The `/attach` file picker dialog always rendered its keyboard shortcuts across two rows at the bottom, regardless of how wide the dialog was. On wider terminals this wasted vertical space and made the dialog taller than necessary.

When the dialog is wide enough to fit all shortcuts on a single line, they now render on one row; on narrower dialogs the two-row layout is preserved as a fallback. The dialog height stays stable in both cases, so there is no layout jump when the window is resized across the threshold.

As part of this work, a `helpKeysLine` helper was extracted in `pkg/tui/dialog/base.go` so that `RenderHelpKeys` and `HelpKeysWidth` share the same formatting logic instead of duplicating it.

### DIFF
--- a/pkg/tui/dialog/base.go
+++ b/pkg/tui/dialog/base.go
@@ -136,9 +136,10 @@ func RenderHelp(text string, contentWidth int) string {
 	return styles.DialogHelpStyle.Width(contentWidth).Align(lipgloss.Center).Render(text)
 }
 
-// RenderHelpKeys renders key bindings in the same style as the main TUI's status bar.
-// Each binding is a pair of [key, description] strings.
-func RenderHelpKeys(contentWidth int, bindings ...string) string {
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly extracts a `helpKeysLine` helper and uses it to collapse the file picker shortcuts to a single row on wide dialogs. The width threshold logic, split indices (`all[:8]`/`all[8:]` correctly splitting 5 key-description pairs as 4+1), and height-stability mechanism (the nil `row2` still producing a blank line via a mandatory second `AddHelpKeys` call) are all correct. No high- or medium-severity bugs were found.

melmennaoui (approved):

PR #3367, opened 2026-07-01
fix(tui): remove duplicate MCP-init spinner from sidebar

dgageot:
The sidebar was showing an "Initializing MCP servers…" spinner that duplicated the existing `ToolsetInfoEvent`-driven "Loading tools…" indicator. Worse, it fired on every conversation turn even for agents with no MCP servers — including the default built-in agent whose toolsets are filesystem, shell, and fetch. The runtime emits `MCPInitStartedEvent` and `MCPInitFinishedEvent` around `getTools()` on every loop iteration whenever the agent has any toolset at all, so the misleading message flashed repeatedly for roughly three seconds on some turns.

This removes the sidebar's MCP-init spinner UI only: the `mcpInit` bool field, its event handlers in `Update()`, its role in `needsSpinner()` and the spinner-tick guard, its reset in `StreamCancelledMsg`, and the "Initializing MCP servers…" render lines in both the expanded and collapsed `workingIndicator()` views. The `MCPInitStartedEvent` and `MCPInitFinishedEvent` types are intentionally kept — they are still emitted by the runtime and consumed by the remote client and other API consumers.

### DIFF
--- a/pkg/tui/components/sidebar/sidebar.go
+++ b/pkg/tui/components/sidebar/sidebar.go
@@ -118,7 +118,6 @@ type model struct {
 	layoutCfg          LayoutConfig              // layout configuration for spacing
 	sessionUsage       map[string]*runtime.Usage // sessionID -> latest usage snapshot
 	todoComp           *todotool.SidebarComponent
-	mcpInit            bool
 	ragIndexing        map[string]*ragIndexingState // strategy name -> indexing state
 	spinner            spinner.Spinner
 	spinnerActive      bool // true when spinner is registered with animation coordinator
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The removal of the `mcpInit` spinner is clean and complete. All references to the field have been consistently eliminated:

- Field declaration removed from the `model` struct
- `needsSpinner()` updated correctly — remaining states (`workingAgent`, `toolsLoading`, `titleRegenerating`) are all preserved
- Spinner-tick guard in the `default` branch matches `needsSpinner()` exactly
- `StreamCancelledMsg` reset handler correctly clears all remaining relevant fields
- Both `workingIndicator()` and `workingIndicatorCollapsed()` render paths cleaned up
- Dropped `MCPInitStartedEvent`/`MCPInitFinishedEvent` case branches safely — unhandled messages fall through to the `default` no-op branch with no state inconsistency

melmennaoui (approved):

PR #3372, opened 2026-07-01
fix(acp): intercept /new, /compact, /usage commands agent-side

dgageot:
The embedded Gordon assistant (Docker Sandboxes) advertised `/new`, `/compact`, and `/usage` slash commands via `emitAvailableCommands`, but the ACP agent's `Prompt` handler never intercepted them — it forwarded them to the LLM as plain text. The result was that `/new` only cleared the client's transcript view; the persisted session in the SQLite store kept all messages and reloaded them via `ResumeSession` when the sandbox or Gordon reopened, making the clear feel like it never happened.

This change intercepts all three commands agent-side before they reach the LLM. `/new` now resets the conversation both in memory via a new lock-safe `session.Session.ResetHistory()` and in the store via a new atomic `Store.ClearSessionHistory` method — a SQLite transaction that resets token/cost counters and drops sub-sessions and session items while preserving the session row itself; an in-memory equivalent is included and the remote store stubs it as unsupported. `/compact` mirrors `SessionSummaryEvent` and `TokenUsageEvent` back to the store (the direct `Summarize` path bypasses the persistence observer that `RunStream` wires in) and only reports success when a summary was actually applied. Command interception runs under a cancellable context registered on the session so that ACP `Cancel` and `CloseSession` can interrupt a long `/compact`. The command parser also rejects prompts that mix a slash word with non-text content, so an attachment next to `/new` is not misread as a command.

New store-level and ACP-level tests cover the happy paths and the no-op case for sessions that have never been persisted.

### DIFF
--- a/pkg/acp/agent.go
+++ b/pkg/acp/agent.go
@@ -406,8 +406,28 @@ func (a *Agent) Prompt(ctx context.Context, params acp.PromptRequest) (acp.Promp
 	// These are handled agent-side and never forwarded to the LLM; without
 	// this the client's /new only clears its own transcript while the
 	// persisted session keeps every message and reloads them on resume.
-	if handled, err := a.handleCommand(ctx, acpSess, params.Prompt); handled {
-		return acp.PromptResponse{StopReason: acp.StopReasonEndTurn}, err
+	//
+	// Commands run under a cancellable turn context registered on the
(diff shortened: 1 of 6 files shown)


k33g:
@dgageot @rumpl why wasn't this merged? As a result, usage with editors like Zed, or others featuring ACP integration, doesn't work fully (specifically the "new session" feature).

PR #3373, opened 2026-07-01
feat: enforce lifecycle.startup_timeout for MCP/LSP toolset startup

dgageot:
## What

Enforces `lifecycle.startup_timeout` for MCP/LSP toolset startup. Until now this config field (and its `strict`-profile 30s default) was parsed and validated but **never acted on** — `EffectiveStartupTimeout()` had no callers, so a server that hung mid-`initialize` would block startup indefinitely.

## Why

The MCP `clientConnector.Connect` deliberately detaches its context with `context.WithoutCancel` (the session must outlive the request that triggered it). A consequence is that the initialize handshake has **no deadline of its own**: a server that accepts the connection but never completes `initialize` wedges the startup attempt forever, and `startup_timeout` could not rescue it because nothing consumed the value.

## How

- Add `StartupTimeout` to `lifecycle.Policy` and wire `EffectiveStartupTimeout()` into `PolicyFromConfig`.
- Enforce it in `Supervisor.connect()` by **racing `Connect` against a timer** rather than a `ctx` deadline (a deadline would be stripped by `WithoutCancel`). On expiry, `Start` returns `ErrInitTimeout` and the toolset stays `Stopped` so the runtime retries on the next turn.
- Guarantee **at most one `Connect` is ever in flight**: a timed-out `Connect` goroutine is left running and recorded in `inflightConnect`; the next `Start` *adopts* it and `Stop` *reaps* it. This is required because the MCP transport shares one underlying client across `Connect` calls — two overlapping connects would race on the shared session (a late one could clobber/close a newer one). An `adopted` flag under the mutex ensures the resulting session is adopted-or-closed exactly once.

## Scope

`startup_timeout` bounds only the **initial** `Start`, not background watcher reconnects (which already have their own restart/backoff budget). `strict` → 30s default; other profiles → 0 (no timeout); explicit value always wins.

## Tests

New supervisor tests (all pass under `-race`): timeout returns `ErrInitTimeout` with a single `Connect`; the next `Start` adopts a late-completing connect; `Stop` reaps and closes a late connect. New `PolicyFromConfig` test covers the nil/resilient/strict/explicit cases.

### DIFF
--- a/pkg/tools/lifecycle/supervisor.go
+++ b/pkg/tools/lifecycle/supervisor.go
@@ -178,6 +178,12 @@ type Supervisor struct {
 	// after closing the session so no transport goroutines are left behind.
 	watchDone chan struct{}
 
+	// inflightConnect holds a connector.Connect that a startup timeout left
+	// running (see connect). At most one exists at a time; the next Start
+	// adopts it and Stop reaps it, so overlapping Connects never race on the
+	// shared MCP session.
+	inflightConnect *pendingConnect
+
 	// randFloat is the jitter source; tests may override.
 	randFloat func() float64
 }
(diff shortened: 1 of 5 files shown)


pkg/tools/lifecycle/supervisor.go
```diff
@@ -271,6 +289,102 @@ func (s *Supervisor) Start(ctx context.Context) error {
...
+		return s.connector.Connect(ctx)
+	}
+
+	s.mu.Lock()
+	p := s.inflightConnect
+	if p == nil {
+		p = &pendingConnect{done: make(chan struct{})}
+		s.inflightConnect = p
+		go func() {
+			sess, err := s.connector.Connect(ctx)
```
docker-agent: **[MEDIUM] Background goroutine captures first `Start`'s context; second caller may receive stale `context.Canceled`**

The goroutine launched in `connect()` captures `ctx` from the *first* `Start` call that triggered it. When the timer fires and that first `Start` returns `ErrInitTimeout`, the goroutine keeps running with the original `ctx`. If that `ctx` is subsequently cancelled or expires (common for request-scoped contexts), `s.connector.Connect(ctx)` returns `context.Canceled`. The *next* `Start` (with a fresh, live `ctx`) adopts the in-flight `pendingConnect` and receives this stale error — making a healthy retry look like a cancellation error.

The PR description notes that MCP connectors call `context.WithoutCancel` internally, stripping the deadline, so this is benign for MCP. However, `Supervisor` is also used for LSP and any other connector that respects ctx cancellation — for those the second `Start` would always fail with a misleading error if the first caller's context was ever cancelled. A third `Start` would succeed (because `inflightConnect` is cleared when the error is returned), but the intermediate failure is unexpected and non-obvious.

**Suggested fix**: use `context.WithoutCancel(ctx)` (or a background context) when launching the goroutine, so its lifetime is independent of any single caller's context:
```go
go func() {
    sess, err := s.connector.Connect(context.WithoutCancel(ctx))
    // ...
}()
```
This matches the same pattern already used in `Stop` and `watch`.

pkg/tools/lifecycle/supervisor.go
```diff
@@ -271,6 +289,102 @@ func (s *Supervisor) Start(ctx context.Context) error {
...
+			s.mu.Unlock()
+			return nil, ErrNotStarted
+		}
+		p.adopted = true
+		if s.inflightConnect == p {
+			s.inflightConnect = nil
+		}
+		res := p.res
+		s.mu.Unlock()
+		return res.sess, res.err
```
docker-agent, line 376: **[LOW] Adopted `Connect` result: session silently leaked if both session and error are non-nil**

In the `<-p.done` branch, `connect()` returns `res.sess, res.err` directly. If a connector ever returns both a non-nil session and a non-nil error, `Start` propagates the error and discards the session without closing it:

```go
sess, err := s.connect(ctx)
if err != nil {
    s.tracker.Fail(StateStopped, err)
    return err  // sess discarded, never closed
}
```

Well-behaved Go connectors return either `(sess, nil)` or `(nil, err)`, so this won't trigger in practice today. But as a defensive measure — especially since the adoption path deals with goroutines that have already returned — it would be cleaner to close the session explicitly when an error accompanies it:

```go
case <-p.done:
    // ...
    res := p.res
    s.mu.Unlock()
    if res.err != nil && res.sess != nil {
        _ = res.sess.Close(context.WithoutCancel(ctx))
    }
    return res.sess, res.err
```

docker-agent (commented):

This PR correctly implements the startup-timeout enforcement mechanism for MCP/LSP toolset startup with a well-thought-out single-inflight-connect guarantee. Two findings worth the author's attention are filed as inline comments.

pkg/tools/lifecycle/supervisor.go
dgageot: Good catch — fixed. The in-flight `Connect` goroutine now runs with `context.WithoutCancel(ctx)`, so a cancellation of the first caller's context can no longer surface as a stale `context.Canceled` to a later adopting `Start`. This matches the pattern already used in `Stop` and `watch`. Added `TestSupervisor_StartAdoptsLateConnectAfterFirstCtxCancelled` to cover the ctx-respecting-connector case.

pkg/tools/lifecycle/supervisor.go
dgageot: Done. The adopt path now closes the session defensively when the connector returns both a non-nil session and a non-nil error, so it is no longer discarded/leaked when `Start` propagates the error.

dgageot:
Addressed both review findings, rebased on `origin/main`, and verified locally (`task lint` clean, `go test -race ./pkg/tools/lifecycle/...` and `task build` pass).

- **[MEDIUM]** The timed-out in-flight `Connect` goroutine now runs with `context.WithoutCancel(ctx)`, so a cancellation of the first caller's context can't surface as a stale `context.Canceled` to a later adopting `Start`. Added `TestSupervisor_StartAdoptsLateConnectAfterFirstCtxCancelled`.
- **[LOW]** The adopt path now closes the session defensively when a connector returns both a non-nil session and a non-nil error, rather than discarding it.

pkg/tools/lifecycle/supervisor.go
docker-agent: Thanks for the fix! The defensive close on line 377 looks good — it prevents the session leak if a connector ever returns both values, and the early return ensures Start gets a clean nil session when an error occurs.

Sayt-0 (approved):

PR #3386, opened 2026-07-01
feat(lint): add OTelTracerName cop to enforce scoped tracer names

dgageot:
OpenTelemetry tracer names identify the *instrumentation scope*, not the service. Before this change, several call sites passed the hard-coded string `"cagent"` to `otel.Tracer()`, which attributed spans to a stale internal name and made it impossible to tell which package originated a trace. There was also no enforcement to prevent the pattern from spreading.

This adds a custom `Lint/OTelTracerName` cop to the in-tree linter. The cop rejects any `otel.Tracer()` call whose argument is not either the current package's import path or the shared `AppName` constant. The special case `"cagent"` is called out explicitly with a message directing the author to use `otel.Tracer(AppName)` instead. Existing call sites in `pkg/a2a`, `pkg/acp`, `pkg/chatserver`, `pkg/mcp`, and `pkg/server` are updated to pass `AppName`. The constant itself is consolidated in `pkg/version` so every package can import a single source of truth; the duplicate declaration in `cmd/root` is replaced with a reference to `version.AppName`.

`task lint`, `task test`, and `task build` all pass with these changes.

### DIFF
--- a/lint/otel_tracer_name.go
+++ b/lint/otel_tracer_name.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"go/ast"
+	"go/constant"
+	"go/types"
+	"strings"
+
+	"github.com/dgageot/rubocop-go/cop"
+)
+
+const (
+	modulePath             = "github.com/docker/docker-agent"
+	appTracerName          = "docker-agent"
+	legacySharedTracerName = "cagent"
+)
+
+// OTelTracerName enforces package-scoped OpenTelemetry instrumentation names.
+//
+// OpenTelemetry's Tracer name identifies the instrumentation scope, not the
+// service. Spans created directly by a package should therefore use that
+// package's import path, so traces can be attributed to the code that emitted
+// them. Runtime wiring is the exception: it intentionally passes the shared
+// application tracer into runtime code. Prefer otel.Tracer(AppName) for that
+// path; existing otel.Tracer("cagent") calls are kept as legacy shared-tracer
+// exceptions.
+//
+// Per-line suppression: `//rubocop:disable Lint/OTelTracerName`.
+var OTelTracerName = &cop.Func{
+	Meta: cop.Meta{
+		Name:        "Lint/OTelTracerName",
+		Description: "otel.Tracer names must be AppName, cagent, or the current package import path",
+		Severity:    cop.Error,
+	},
+	Types: true,
+	Run: func(p *cop.Pass) {
+		if p.Info == nil || p.Package == nil {
+			return
+		}
+		expected := packageImportPath(p.Package.Path())
+		p.ForEachCall(func(call *ast.CallExpr) {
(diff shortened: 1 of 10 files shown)


lint/otel_tracer_name.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	return modulePath + "/" + strings.TrimPrefix(pkgPath, "./")
+}
+
+func isAppName(info *types.Info, expr ast.Expr, name string) bool {
+	if name != appTracerName {
+		return false
+	}
+	c, ok := constObject(info, expr)
+	return ok && c.Name() == "AppName"
```
docker-agent: **[MEDIUM] `isAppName()` accepts any constant named `AppName` regardless of defining package**

`isAppName` validates a constant by name (`c.Name() == "AppName"`) and value (`name == "docker-agent"`), but never checks which package declares it (`c.Pkg().Path()`). As a result, any package can bypass the cop by declaring a local `const AppName = "docker-agent"` — a perfectly valid Go constant that looks identical to the real `version.AppName` to this check.

The `*types.Const` object returned by `constObject` via `info.Uses` already exposes `Pkg()`, so the fix is straightforward:

```go
func isAppName(info *types.Info, expr ast.Expr, name string) bool {
    if name != appTracerName {
        return false
    }
    c, ok := constObject(info, expr)
    if !ok || c.Name() != "AppName" {
        return false
    }
    pkg := c.Pkg().Path()
    return pkg == "github.com/docker/docker-agent/pkg/version" ||
           pkg == "github.com/docker/docker-agent/cmd/root"
}
```

lint/otel_tracer_name.go
```diff
@@ -0,0 +1,122 @@
...
+	"go/constant"
+	"go/types"
+	"strings"
+
+	"github.com/dgageot/rubocop-go/cop"
+)
+
+const (
+	modulePath    = "github.com/docker/docker-agent"
+	appTracerName = "docker-agent"
```
docker-agent: **[LOW] Hardcoded `appTracerName = "docker-agent"` duplicates `version.AppName` and will silently diverge**

`appTracerName` is a string literal that mirrors `pkg/version.AppName`. If the canonical constant is ever renamed (e.g., when the application is rebranded), the cop will silently start either rejecting valid `version.AppName` uses or accepting the stale string `"docker-agent"` from unrelated constants — with no compile-time link to catch the drift.

Since `lint/` is in the same Go module, consider driving the check from the constant's package path (see the sibling comment on `isAppName`) rather than comparing against a duplicated string. That way a rename of `version.AppName` would surface as a compile error in the cop itself rather than a silent semantic mismatch.

docker-agent (commented):

The PR cleanly migrates all `otel.Tracer("cagent")` call sites to `version.AppName` and the cop structure is sound. Two issues in the new cop logic are worth addressing before this merges.

Sayt-0 (dismissed):

Sayt-0 (approved):

PR #3390, opened 2026-07-01
feat: add tags field to agent config metadata

dgageot:
Agent configs already carry distribution metadata such as author, license, and version, but there was no way to attach categorization tags. Tags make agents easier to organize and discover, both in a local collection of configs and once they are published to an OCI registry.

This adds an optional `tags` field to the `metadata` block of the latest config schema. Tags are a plain list of strings and, like the other metadata fields, are surfaced when packaging an agent as an OCI artifact: they are joined into an `io.docker.agent.tags` annotation so registry tooling can read them without unpacking the config layer. The JSON schema, an example, and the configuration docs are updated to match.

The change is confined to the latest config version, so older frozen config versions are untouched and continue to migrate through JSON cloning without any special handling.

### DIFF
--- a/pkg/oci/package.go
+++ b/pkg/oci/package.go
@@ -76,6 +76,9 @@ func PackageFileAsOCIToStore(ctx context.Context, agentSource config.Source, art
 	if revision := cfg.Metadata.Version; revision != "" {
 		annotations["org.opencontainers.image.revision"] = revision
 	}
+	if len(cfg.Metadata.Tags) > 0 {
+		annotations["io.docker.agent.tags"] = strings.Join(cfg.Metadata.Tags, ",")
+	}
 
 	layer := static.NewLayer(data, "application/yaml")
 	img, err := mutate.AppendLayers(empty.Image, layer)
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1684,6 +1684,13 @@
         "version": {
           "type": "string",
           "description": "Version of the agent configuration (used for OCI registry publishing)"
+        },
+        "tags": {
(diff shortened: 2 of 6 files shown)


pkg/oci/package.go
```diff
@@ -76,6 +76,9 @@ func PackageFileAsOCIToStore(ctx context.Context, agentSource config.Source, art
 	if revision := cfg.Metadata.Version; revision != "" {
 		annotations["org.opencontainers.image.revision"] = revision
 	}
+	if len(cfg.Metadata.Tags) > 0 {
+		annotations["io.docker.agent.tags"] = strings.Join(cfg.Metadata.Tags, ",")
```
docker-agent, line 80: **[MEDIUM] Comma-joined tags annotation is ambiguous when tag values contain commas**

`strings.Join(cfg.Metadata.Tags, ",")` produces an annotation that cannot be round-tripped when any individual tag contains a comma. For example, `tags: ["foo,bar", "baz"]` encodes as `"foo,bar,baz"`, which is indistinguishable from `tags: ["foo", "bar", "baz"]` — any consumer splitting on `,` will silently misparse the original list.

The OCI Image Spec does not prescribe a delimiter for multi-value annotations, but JSON array serialization (`["foo,bar","baz"]`) is an unambiguous alternative. Another option is to validate/reject tag values that contain commas at config-load time and document that restriction.

The existing test only uses simple single-word tags (`coding`, `review`) and would not catch this edge case.

docker-agent (commented):

Sayt-0 (approved):

PR #3391, opened 2026-07-01
feat: add mouse support to the agent picker (hover, double-click, wheel)

dgageot:
The agent picker previously required keyboard navigation to select an agent. This adds full mouse support so users can point and click their way through the list without touching the keyboard.

Hovering over a card highlights it; a single click moves the selection to that card; a double-click selects the card and immediately starts the session — matching the muscle memory most users have from any file-picker dialog. When the YAML details panel is open, mouse-wheel events are forwarded to its viewport so the content can be scrolled without switching to keyboard mode. The subtitle bar now advertises the double-click gesture so the feature is discoverable.

The implementation avoids re-rendering the entire panel on every mouse-motion event by introducing a `panelSize` helper that computes card dimensions cheaply and sharing a `headerText` helper between it and the full `render` path so the two can't drift. A layout bug was also fixed in the details panel: the description width was using `cardWidth-4` instead of `cardWidth-6`, causing long descriptions to wrap and break the fixed card height that hit-testing relies on. Double-click tracking is cleared whenever the details dialog is opened or closed to prevent phantom clicks from carrying across state transitions. The new behaviour is covered by table-driven tests that verify `panelSize` stays consistent with `render`, that `cardAt` resolves the correct card from rendered text positions, that details-reset clears click state, and that wheel events are ignored when no details dialog is open.

### DIFF
--- a/cmd/root/agent_picker.go
+++ b/cmd/root/agent_picker.go
@@ -196,6 +196,7 @@ func (m *agentPickerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			switch {
 			case key.Matches(msg, agentPickerKeys.Quit), key.Matches(msg, agentPickerKeys.Details):
 				m.showDetails = false
+				m.resetClickTracking()
 				return m, nil
 			}
 			var cmd tea.Cmd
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The mouse-support implementation looks correct and well-tested. Key areas reviewed:

- **Hit-testing math** (): Panel centering, / offsets, and card-height division are internally consistent with  and .
- ** /  consistency**:  is shared between both paths, so they cannot drift. The row formula (`3 + N×4 + 2`) and horizontal chrome (`2×(1+3)`) match the layout constants.
- **Double-click tracking**: First click arms state; second click on the same card within the threshold quits and resets index to −1; a miss also resets. State is cleared on both `openDetails` and the Quit/Details keypress so no phantom double-click can fire across dialog transitions.
- **Wheel forwarding**: Correctly guarded by `m.showDetails`; no-ops otherwise.
- ** fix** (`cardWidth-4` → `cardWidth-6`): Aligns description budget with the header, preventing unexpected line wraps that would break the fixed card height hit-testing relies on.
- **Test coverage**: Table-driven tests for `panelSize` vs `render`, `cardAt` vs rendered text positions, double-click timeout, click-outside reset, details-reset, and wheel-ignored-without-details all look correct and comprehensive.

No bugs found in the introduced changes.

Sayt-0 (approved):

Sayt-0 (approved):

PR #3392, opened 2026-07-01
fix(runtime): suppress spurious empty-response warning for benign post-tool stops

dgageot:
## What

The runtime run loop (`runTurn`) emits a UI warning whenever a model turn comes back empty (no text content and no tool calls):

> Model X returned an empty response (stop reason: stop). This usually means the provider rate-limited the request or the output token limit was reached.

This fires spuriously at the tail of **forked-skill sub-sessions**, where the model's last real action is a tool call (e.g. `commit` / `open-pr`) and it then naturally stops with an empty final turn. That's benign — the model already did its work and had nothing left to say — but the warning wrongly implied a rate-limit / token-cap event.

## How

- Track whether the immediately preceding turn made tool calls (`prevTurnMadeToolCalls` on `loopState`), reset on agent switch so it never leaks across agents.
- When an empty turn follows a **clean stop** (`finish_reason == "stop"`) **and** the previous turn made tool calls, treat it as benign: log at debug, no UI warning.
- All other empty turns still warn as before:
  - reasoning-only thinking-mode turns (Qwen3 et al., see #3145),
  - genuinely empty turns with no prior tool work,
  - **`length` truncations** — an output-token-limit cutoff after a tool call is a real failure and must still be surfaced.
- Extracted the message classification into a pure, unit-testable helper `emptyTurnWarning(...)` that returns the warning string or `""` when benign.

## Why `stop`-only (not just "stopped")

`res.Stopped` is `true` for both `finish_reason: stop` and `finish_reason: length`. Keying the benign case on `res.Stopped` would silently swallow genuine token-limit truncations that happen to follow a tool-call turn. The Anthropic adapter maps `end_turn → stop` and `max_tokens → length`, so keying strictly on `stop` preserves the original reported case (`stop reason: stop`) while keeping truncation warnings intact.

## Tests

- `TestEmptyTrailingTurnAfterToolCallsIsSilent` — full run-loop: benign empty stop after a tool call emits no warning.
- `TestEmptyTurnWarning` — pure helper: benign stop, reasoning-only, benign-precedence, no-prior-tools, `length`-after-tool-calls (still warns), non-stop.
- Existing `TestEmptyTurnEmitsWarning` / `TestReasoningOnlyTurnEmitsWarning` still pass.

## Validation

- `go test ./pkg/runtime/... ./pkg/model/provider/anthropic/...` — green
- `task lint` — 0 issues

## Commits

1. `feat(runtime): suppress empty-response warning for benign post-tool stops`
2. `refactor(runtime): extract emptyTurnWarning classification helper`
3. `fix(runtime): tighten empty-turn benign check after review`

### DIFF
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -572,6 +572,37 @@ type loopState struct {
 	prevTurnMadeToolCalls bool
 }
 
+// emptyTurnWarning classifies an empty assistant turn (no content, no tool
+// calls) and returns the user-facing warning message, or "" when the turn is
+// benign and should stay silent. Known cases:
+//   - benign: a natural stop right after a tool-call turn — the model already
+//     did its work and had nothing left to say (common at the tail of a
+//     fork-skill sequence, e.g. commit / open-pr). The real answer is the
+//     previous assistant message, so a UI warning would be noise.
+//   - reasoning-only: thinking-mode models (e.g. Qwen3 via
+//     openai_chatcompletions) that stream only reasoning tokens and then stop
+//     or hit the output token limit, leaving visible content empty (see #3145).
+//   - otherwise: an empty stream from a rate-limited or token-capped provider.
+//
+// Refusals are handled separately by the caller and never reach here.
+func emptyTurnWarning(res streamResult, prevTurnMadeToolCalls bool, modelID string, reason chat.FinishReason) string {
+	switch {
+	case res.Stopped && prevTurnMadeToolCalls:
+		return ""
+	case strings.TrimSpace(res.ReasoningContent) != "":
+		return fmt.Sprintf(
+			"Model %s produced only reasoning and no reply (stop reason: %s). "+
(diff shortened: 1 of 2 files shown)


Sayt-0 (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is clean and correct. `prevTurnMadeToolCalls` is updated at the right point in the control flow (after `processToolCalls`, before any `turnContinue` returns), the two early `turnExit` paths that skip the update both terminate the outer loop so no subsequent iteration can observe a stale value, and the agent-switch reset is correctly scoped to name changes only. The `emptyTurnWarning` helper correctly keys the benign case on `chat.FinishReasonStop` (not `res.Stopped`) to preserve truncation warnings for `finish_reason: length`. Test coverage addresses the critical paths: benign stop, reasoning-only, length-after-tool-calls, and non-stop empty turns.

PR #3394, opened 2026-07-01
feat: show metadata tags as coloured chips in the agent picker

dgageot:
The agent picker previously showed only the agent's name and description. Now that agents can carry `metadata.tags`, it makes sense to surface them visually so users can glance at an agent's purpose without reading the full description.

Tags are rendered as coloured `#tag` chips on each picker card using a rotating palette, width-clamped so they never overflow the card. The picker itself has been given more breathing room — cards are 70 columns wide, titles have a blank line before the subtitle, and panel padding is roomier. The YAML details dialog grows to 110×36 (still clamped to the terminal) and now truncates long lines instead of wrapping them, which keeps instruction-heavy configs from inflating the view height; horizontal scrolling remains available. The details panel's bottom status bar now matches the picker's muted style. The two built-in agents (`default`, `coder`) have been tagged with `general/assistant` and `coding/agentic` respectively. Mouse hit-testing math (`cardAt`, `panelSize`) has been kept in sync with the layout changes, and new unit tests cover tag rendering and the no-panic path with tagged agents.

No breaking changes; the tags field is optional and the picker degrades cleanly when no tags are present.

### DIFF
--- a/cmd/root/agent_picker.go
+++ b/cmd/root/agent_picker.go
@@ -31,10 +31,11 @@ var errAgentPickerCancelled = errors.New("agent selection cancelled")
 
 // agentChoice is a single entry in the agent picker.
 type agentChoice struct {
-	ref         string // agent reference as passed on the command line
-	description string // one-line description loaded from the agent config
-	yaml        string // raw config YAML, shown in the details dialog
-	err         error  // non-nil when the config could not be loaded
+	ref         string   // agent reference as passed on the command line
+	description string   // one-line description loaded from the agent config
+	tags        []string // metadata tags shown as coloured chips
+	yaml        string   // raw config YAML, shown in the details dialog
+	err         error    // non-nil when the config could not be loaded
 }
 
 // loadAgentChoices resolves and loads metadata for each ref so the picker can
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The tag-chip rendering, layout math, mouse hit-testing, and test coverage all look correct. No bugs were found in the changed code. One minor cosmetic note is left as an inline comment.

cmd/root/agent_picker.go
```diff
@@ -616,16 +643,56 @@ func (m *agentPickerModel) renderCard(choice agentChoice, cardWidth int, selecte
...
+		// Account for the single-space separator between chips.
+		sep := 0
+		if len(chips) > 0 {
+			sep = 1
+		}
+		if used+sep+lipgloss.Width(label) > width {
+			break
+		}
+		used += sep + lipgloss.Width(label)
+		style := tagChipStyles[i%len(tagChipStyles)]
```
docker-agent, line 690: **[LOW] Tag chip palette index counts skipped blank tags, causing minor colour inconsistency**

`renderTags` uses the raw loop index `i` to choose a colour from `tagChipStyles[i%len(tagChipStyles)]`, but blank/whitespace-only tags are `continue`d before any chip is rendered. When a blank tag appears mid-slice the visible chips' colours skip a palette slot, so an agent whose tags include an empty string will show a different colour sequence than an identical agent whose tags list had no blanks. In practice the built-in agents have clean tags, so this is cosmetic only — but if tags come from user-supplied YAML it could produce subtly unexpected colouring. A trivial fix is to increment a separate `chipCount` variable only when a chip is actually rendered and use that for the modulo.

Sayt-0 (approved):

PR #3396, opened 2026-07-01
refactor: extract shared calleeObject helper in lint cops

dgageot:
Both `constructor_command_exec.go` and `constructor_network_io.go` contained nearly identical type-switch blocks to extract the callee object from a call expression — handling both bare `*ast.Ident` and selector-expression `*ast.SelectorExpr` cases, then resolving the corresponding types.Object from type info. The duplication made each cop harder to read and meant any future fix had to be applied twice.

This change extracts that logic into a single `calleeObject` helper in `constructor_command_exec.go` and updates `constructor_network_io.go` to call it, removing around 20 lines of duplicated branching with no change in behaviour.

No breaking changes. Both cops produce identical diagnostics before and after.

### DIFF
--- a/lint/constructor_command_exec.go
+++ b/lint/constructor_command_exec.go
@@ -54,19 +54,27 @@ var ConstructorCommandExec = &cop.Func{
 }
 
 func commandConstructorCall(p *cop.Pass, call *ast.CallExpr) (string, bool) {
-	if p.Info != nil {
-		switch fun := call.Fun.(type) {
-		case *ast.SelectorExpr:
-			if name, ok := osExecFuncName(p.Info.Uses[fun.Sel]); ok {
-				return name, true
-			}
-		case *ast.Ident:
-			if name, ok := osExecFuncName(p.Info.Uses[fun]); ok {
-				return name, true
-			}
-		}
+	if name, ok := osExecFuncName(calleeObject(p.Info, call)); ok {
+		return name, true
 	}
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This refactoring correctly extracts the duplicated callee-resolution type-switch into a single `calleeObject` helper. The behaviour is identical to the original code in both callers:

- **Nil-safety preserved**: `calleeObject` guards `info == nil` exactly as the old `if p.Info != nil` blocks did. Both `osExecFuncName` and `networkIOFuncName` use the safe comma-ok type assertion (`obj.(*types.Func)`), so passing `nil` is well-defined and returns the zero/false pair — no panic risk.
- **Map-lookup semantics unchanged**: `info.Uses[fun.Sel]` and `info.Uses[fun]` return `nil` (the zero value for the `types.Object` interface) on a missing key, which both downstream functions already handle correctly.
- **Fallthrough paths intact**: The `cop.CallTo` fallback in `commandConstructorCall` and the two `cop.CallTo` checks in `networkIOCall` are untouched and still reachable.
- **Type-switch coverage identical**: The helper covers exactly the same `*ast.SelectorExpr` and `*ast.Ident` cases as the original inline blocks, with an explicit `default: return nil` that is semantically equivalent to not entering the switch.

trungutt (approved):

PR #3398, opened 2026-07-01
refactor(provider): deduplicate provider client boilerplate

dgageot:
The five provider clients (openai, anthropic, gemini, bedrock, dmr) each
contained several identical blocks of setup code, making every new provider a
copy-paste exercise and any cross-cutting fix a multi-file hunt. This PR
extracts those repeated patterns into shared helpers with no observable
behavior change.

Six kinds of duplication were collapsed: the option-accumulation loop that all
constructors and factories ran (`options.Apply`); the HTTP transport-wrapping
block repeated in seven places (`options.ModelOptions.WrapTransport`); the
`track_usage` flag check present at seven call sites (`base.Config.TrackUsageEnabled`);
the Docker AI Gateway fail-fast and per-request short-lived token fetch that
appeared in six constructors (`base.VerifyDockerGatewayAuth` and
`base.GatewayAuthToken`); the gateway `httpclient` option list copied three
times (`base.GatewayHTTPOptions`); and the rerank-score JSON parser that was
byte-identical in openai and anthropic (`providerutil.ParseRerankScores`). One
small behavior improvement is folded in: `ParseRerankScores` now returns
"expected N scores, got M" when the JSON is valid but the score count is wrong,
instead of the generic "invalid rerank JSON" error. All other error strings,
log lines, and side-effect timing are preserved exactly — the per-request token
fetch still happens inside its gateway closure.

New unit tests cover every extracted helper directly, and the full test suite,
linter, race detector, and `GOOS=js GOARCH=wasm` build all pass.

### DIFF
--- a/pkg/model/provider/anthropic/beta_client.go
+++ b/pkg/model/provider/anthropic/beta_client.go
@@ -123,7 +123,7 @@ func (c *Client) createBetaStream(
 	}
 
 	stream := client.Beta.Messages.NewStreaming(ctx, params, requestOpts...)
-	trackUsage := c.ModelConfig.TrackUsage == nil || *c.ModelConfig.TrackUsage
+	trackUsage := c.TrackUsageEnabled()
 	ad := c.newBetaStreamAdapter(stream, trackUsage)
 
 	// Set up single retry for context length errors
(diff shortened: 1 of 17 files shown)


pkg/model/provider/anthropic/beta_client.go
```diff
@@ -346,43 +346,6 @@ func extractAnthropicStructuredOutputJSON(msg *anthropic.BetaMessage) (string, e
 	return "", errors.New("no structured JSON text found in Anthropic BetaMessage content")
 }
 
-// parseRerankScoresAnthropic parses a JSON payload of the form {"scores":[...]} and validates length.
-// This helper is local to the Anthropic provider to avoid cyclic dependencies.
-func parseRerankScoresAnthropic(raw string, expected int) ([]float64, error) {
```
rumpl, line 351: This was moved to some util thing, but it's only used by anthropic, why would we move it and make it public?

dgageot: Because it's used by openai too?

trungutt (approved):

PR #3399, opened 2026-07-01
test: make the test suite faster and sleep-free

dgageot:
The test suite contained 44 `time.Sleep` calls spread across 22 test files. These sleeps slowed every run, introduced timing-dependent flakiness on loaded CI machines, and gave no actual guarantee of ordering — a fast machine might still race past them. This change removes all of them, replacing each with a deterministic synchronization primitive appropriate to the situation.

Depending on the test, the replacement is a channel gate, `require.Eventually`, `gotest.tools/poll.WaitOn`, `time.AfterFunc`, a fake clock, or a backdated TTL entry. For the prompt-serialization test in `pkg/tools/builtin/shell`, a small production change was needed: `askpass.go` gained an atomic counter (`promptArrivals`) so the test can wait for a prompt to arrive rather than sleeping. The 429 rate-limit test in `pkg/remote` also dropped go-containerregistry's default retry backoff (which slept ~4.4 s total) in favour of a millisecond-scale backoff with the same retry count. The three remaining `time.Sleep` calls are fake-time advances inside `testing/synctest` bubbles and carry `//nolint` directives explaining why they are intentional. A `forbidigo` lint rule now bans `time.Sleep` in `*_test.go` files so the pattern cannot silently return.

Wall-clock speedups are most visible in the packages that leaned on sleep the most: `pkg/runtime` drops from ~9 s to ~3 s and `pkg/server` from ~5 s to ~2 s in isolation. The full suite was run with `-race` on all touched packages; `golangci-lint` and the new custom cop are clean.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -83,6 +83,8 @@ linters:
           msg: "do not use os.Chdir() in tests, use t.Chdir()"
         - pattern: fmt\.Print.*()
           msg: "do not use fmt.Print() or fmt.Println() in tests"
+        - pattern: time\.Sleep
+          msg: "do not sleep in tests: synchronize on channels/require.Eventually, or use fake time in a synctest bubble (nolint with a reason)"
     depguard:
       rules:
         internal:
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -6,7 +6,6 @@ import (
 	"log/slog"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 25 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly replaces 44 `time.Sleep` calls across 22 test files with deterministic synchronization primitives. No bugs were found in the changed code.

**Key areas reviewed:**
- **`askpass.go` production change** — `promptArrivals atomic.Int32` is correctly placed and used with atomic operations. No race conditions introduced.
- **Channel-based synchronization** — All channels in elicitation, server, and supervisor tests are appropriately buffered or closed; no goroutine leaks or deadlocks detected.
- **`require.Eventually` / `poll.WaitOn`** — Timeout values (1–2 s with ms polling) are reasonable for in-process test conditions.
- **Rate-limit retry backoff** — `Backoff{Duration: time.Millisecond, Factor: 1, Steps: 3}` correctly removes multi-second backoff while preserving retry-path coverage.
- **`forbidigo` lint rule** — Correctly scoped to `*_test.go` files; prevents `time.Sleep` from silently re-entering the test suite.

The three intentional `time.Sleep` calls inside `synctest` bubbles with `//nolint` directives are appropriate and well-documented.

melmennaoui (approved):

PR #3400, opened 2026-07-01
fix: show lean tui cost and context limit

rumpl:


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly wires cost and context-limit data through the lean TUI and sidebar. The session-tracking logic in `update.go` is generally sound — `activeUsage()` covers the common cases (active stack entry → root session → single-entry map). Two low-severity edge cases were noted in the new session-tracking code; see inline comments for details.

pkg/leantui/update.go
```diff
@@ -355,11 +356,99 @@ func (m *model) handleStreamStopped(ctx context.Context) {
...
+		return
+	}
+
+	snapshot := usageSnapshot{
+		contextLength: usage.ContextLength,
+		contextLimit:  usage.ContextLimit,
+		tokens:        usage.InputTokens + usage.OutputTokens,
+		cost:          usage.Cost,
+	}
+	if sessionID == "" {
```
docker-agent, line 393: **[LOW] setTokenUsage silently drops empty-sessionID usage events once session tracking is active**

When `sessionID == ""` and `m.usageBySession` is already populated (i.e., at least one session-keyed event has been received), `setTokenUsage` returns without updating the display at all:

```go
if sessionID == "" {
    if len(m.usageBySession) == 0 {
        m.applyStatusUsage(snapshot, usage.Cost, true)
    }
    return  // silent drop if map is non-empty
}
```

This is a defensible design choice (session-aware tracking takes priority over legacy events), but it means any runtime path that emits `TokenUsageEvent` with an empty `SessionID` after session-keyed events have started will silently stop updating the status bar. Consider adding a comment to document the intent, or log the drop at debug level, to avoid confusing future callers.

pkg/leantui/update.go
```diff
@@ -355,11 +356,99 @@ func (m *model) handleStreamStopped(ctx context.Context) {
...
+	for _, usage := range m.usageBySession {
+		totalCost += usage.cost
+	}
+
+	if usage, ok := m.activeUsage(); ok {
+		m.applyStatusUsage(usage, totalCost, true)
+		return
+	}
+
+	for _, usage := range m.usageBySession {
```
docker-agent: **[LOW] applyUsageSnapshot falls back to an arbitrary map entry when activeUsage() returns false with multiple sessions**

When `activeUsage()` cannot resolve the active session (stack empty, `rootSessionID` unset, and multiple sessions in the map), the fallback iterates the map and picks whichever entry Go's runtime happens to return first:

```go
// Fallback: pick first arbitrary entry from the map
for _, usage := range m.usageBySession {
    m.applyStatusUsage(usage, totalCost, true)
    return
}
```

In practice this requires `rootSessionID` to be unset, which only happens if `trackStreamStarted` was never called before usage events arrived. That ordering is unlikely in normal flow, but theoretically possible under event reordering. If this fallback is ever triggered, the status bar may show context numbers from a sub-session rather than the root session.

Consider using `m.rootSessionID` as an explicit tiebreaker here, or at minimum log a warning when the fallback fires.

dgageot (approved):

PR #3404, opened 2026-07-02
perf(tui): memoize rendering hot paths for long conversations

dgageot:
TUI rendering of long conversations was spending significant time re-computing things that rarely change: scrollbar geometry, visible-line widths, and the per-frame join of the content column with the scrollbar column. On a 500-message history at 120×40, a single render frame cost ~185µs and 353 allocations.

The fix memoizes three independent hot paths. The scrollbar now caches its track and thumb cells keyed on thumb geometry and styled cell content, so drag state and theme changes still invalidate naturally. The scroll view caches `ansi.StringWidth` results per content line, keyed on the string pointer identity of each line, and invalidates the whole cache when `SetContent` replaces the slice. The content+scrollbar column join replaces `lipgloss.JoinHorizontal` (which re-splits and re-measures already-padded strings) with a single-pass string builder that zips the two columns directly. A `ViewWithRestyledLines` fast path was also added so the messages list can benefit from cached widths even after selection or URL-hover restyling; restyled lines fall back to a fresh width measurement rather than trusting the cache, since `runewidth` and `ansi.StringWidth` can disagree on complex grapheme clusters.

After these changes the same benchmark runs at 6.1µs/op with 32 allocations — roughly a 30× throughput improvement. New unit tests assert byte-identical output versus the old `JoinHorizontal`-based composition across all scrollbar and reserved-space modes, and cover width-cache invalidation and the restyled-line re-measure fallback.

### DIFF
--- a/pkg/tui/components/messages/messages.go
+++ b/pkg/tui/components/messages/messages.go
@@ -615,10 +615,12 @@ func (m *model) View() string {
 	visibleLines = m.applyURLUnderline(visibleLines, startLine)
 
 	// Sync scroll state and delegate rendering to scrollview which guarantees
-	// fixed-width padding, pinned scrollbar, and exact height.
+	// fixed-width padding, pinned scrollbar, and exact height. Selection and
+	// URL-hover restyling preserve display width, so the scrollview can reuse
+	// memoized line widths instead of re-measuring every visible line.
 	m.scrollview.SetContent(m.renderedLines, m.totalScrollableHeight())
 	m.scrollview.SetScrollOffset(m.scrollOffset)
-	return m.scrollview.ViewWithLines(visibleLines)
+	return m.scrollview.ViewWithRestyledLines(visibleLines)
 }
 
 // updateScrollState recomputes rendered content, bottom slack and scroll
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The memoization work is well-structured and the performance wins are real. The three independent hot-path caches (scrollbar geometry, per-line width, and the zip builder) each have clear invalidation paths and are exercised by the new unit tests.

**On the two hypotheses raised during review:**

1. **GC-based address reuse in `SetContent`** (`&lines[0] != &m.lines[0]` guard): Go's GC is non-moving, so an object's address is stable for its lifetime. At the point the comparison executes, `m.lines` still holds the old backing array, keeping it alive — so the old and new arrays can never have the same address simultaneously. This is safe.

2. **`baseLine` captured before `syncScrollbar` clamping**: `ViewWithRestyledLines` captures `m.scrollOffset` as `baseLine`, then `viewWithLines` calls `syncScrollbar()`. In the messages.go call sequence, `SetContent` and `SetScrollOffset` are called immediately before `ViewWithRestyledLines`, already propagating the clamped offset to the scrollbar. `syncScrollbar` therefore reads back the same value and makes no change. The `baseLine` is always consistent with the visible-line slice.

No correctness issues were found in the changed code.

aheritier (approved):

PR #3410, opened 2026-07-02
feat(server): resume sessions across agent URL query changes

gtardif:
## Problem

Sessions started via `docker-agent serve api <url>` record the **full URL-encoded agent reference** as their source key, query parameters included. 
When the server is relaunched with a different tag (e.g. different version of an agent yaml), the exact key recorded by the client no longer exists in `Sources`. Resuming a stored session then fails:

```
500  error="failed to run session: agent not found: http..."
```

It works for the *first* message only because a live runtime is cached in `runtimeSessions` (checked before `Sources`); after a restart that cache is empty, so the lookup must rebuild from `Sources` and misses.

Note: **viewing/listing** old sessions was never affected — those are keyed by session ID in the store. This PR fixes **continuing** them.

## Fix

Add a resume-time fallback in the session manager. `loadTeam` / `loadTeamWithConfig` now share a `resolveSource` helper that:

1. **Prefers an exact key match** — all existing behaviour (side-by-side variants, `/api/agents`, directory mode) is unchanged.
2. **On a miss, matches on a stable identity** via a new `config.StableSourceKey`. For URL references the identity is the **path** (`scheme + host + path`); the **entire query string and fragment are treated as volatile**. This is a deliberate rule rather than an enumerated denylist, so any current *or future* query parameter is ignored without further code changes.
3. **Only resolves when unambiguous** — if several live sources share the requested stable identity, it declines (returns `agent not found`) rather than guessing, so side-by-side variants are never silently mis-selected.

`task build`, `task lint` (0 issues), and the affected packages' tests all pass.

### DIFF
--- a/pkg/config/resolve.go
+++ b/pkg/config/resolve.go
@@ -114,6 +114,43 @@ func resolveOne(resolvedPath string, envProvider environment.Provider) (string, 
 	}
 }
 
+// StableSourceKey reduces a Sources map key to its stable identity: the part
+// that persists across variant selectors and caller/environment metadata.
+//
+// For URL references the identity is the URL's path (scheme + host + path); the
+// entire query string and fragment are treated as volatile. This is what lets a
+// session created under one variant (e.g. ?gordonTag=v9-light) resume under
+// another (?gordonTag=v9-dev) after the server is relaunched with a different
+// tag: only the query differs, so both share one identity. Treating the whole
+// query as volatile — rather than an enumerated denylist — keeps this robust as
+// new query parameters are introduced by callers.
+//
+// Keys for URL references are url.QueryEscape'd URLs; this decodes the key
+// before parsing. For keys that are not URL references (local files, OCI refs,
+// built-ins) or that cannot be decoded/parsed, the key is returned unchanged,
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The session-resumption fallback logic is well-designed and correctly implemented.

**What was reviewed:**
- `StableSourceKey` in `pkg/config/resolve.go`: URL decoding + query/fragment stripping is symmetric — both stored map keys and client-supplied keys are `url.QueryEscape`-encoded, so both sides go through the same decode→strip path, producing equal stable keys. Non-URL keys (local files, OCI refs, builtins) fall through unchanged. Error handling on malformed inputs is sound (return key unchanged on any parse failure).
- `resolveSource` in `pkg/server/session_manager.go`: Exact-match-first logic is correct. The ambiguity guard (`matches > 1` → decline) prevents silent mis-selection of side-by-side variants. The `matches == 0` path also correctly returns agent-not-found.
- Integration into `loadTeam` / `loadTeamWithConfig`: straightforward delegation with unchanged error propagation.

Sayt-0 (approved):

PR #3412, opened 2026-07-02
feat(tui): generate a TUI e2e test from a `--record` session

dgageot:
Writing a TUI regression test today means manually instrumenting a session after the fact — guessing at the right `WaitFor` anchors, cross-referencing the cassette, and tweaking until it passes. This PR closes that loop: run the agent once with `--record`, and the session produces a ready-to-edit `*_test.go` alongside the cassette, with synchronization points derived from what was actually on screen.

## How it works

The new `pkg/tui/recorder` package wraps the top-level `tea.Model` transparently. Before each keystroke or left click is forwarded to the real model, the current rendered frame is captured (ANSI-stripped, matching tuitest normalization) and stored alongside the event and its timestamp. When the session ends, `GenerateTest` converts the event log into tuitest DSL: rapid keystrokes coalesce into `Type(...)` bursts, special keys become `Press(...)`/`Enter()`, and clicks resolve to `ClickText(...)` when the cell content is unambiguous, falling back to `Click(x, y)`. Idle pauses where the screen content also changed are turned into `WaitFor(tuitest.Contains(...))` anchors by diffing consecutive frames and picking the longest new stable line (borders and spinners are filtered). Trailing `Ctrl+C`/`Ctrl+D` quit keystrokes are trimmed, and the settled final frame drives the closing `WaitFor`. The result is `gofmt`'d and written next to the cassette with a three-step header comment (`mv`/`cp` the agent YAML, cassette, and test file into `e2e/tui/`).

Hardening in the follow-up commit covers: agent-path fallback for built-in and registry refs, deterministic FNV hash when the cassette name yields no valid Go identifier characters, `O_EXCL` + numbered-suffix so existing recordings are never silently clobbered, and control-character sanitization of user-controlled strings interpolated into the generated header comment.

## Example output

Running `docker-agent run --record myagent.yaml` and asking "What's 2+2?" produces something like:

```go
func TestRecordedCagentRecording01(t *testing.T) {
    tui := newTUI(t, "testdata/myagent.yaml", 220, 50)
    tui.WaitFor(tuitest.Not(tuitest.Contains("Loading")))
    tui.Type("What's 2+2?")
    tui.Enter()
    tui.WaitFor(tuitest.Contains("2 + 2 equals 4."))
}
```

## Tests and known limitations

Unit tests cover the recorder wrapper, burst coalescing, key-to-DSL mapping, `ClickText` extraction, `WaitFor` anchor diffing, quit-key trimming, a Go-injection pin on hostile cassette paths, and unique-file write semantics. `task build`, `task test`, and `task lint` all pass.

`WaitFor` anchors are derived from screen content at recording time, so anything time-stamped, counter-suffixed, or machine-dependent will likely need manual editing before the test is reliable in CI. Agents loaded from the built-in catalog or an OCI registry need their YAML materialized into `testdata/` — the generated header comment calls this out explicitly.


cmd/root/record_tui.go
```diff
@@ -0,0 +1,55 @@
...
+			path = fmt.Sprintf("%s_%d_test.go", base, n)
+		}
+		f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
+		if os.IsExist(err) {
+			continue
+		}
+		if err != nil {
+			return "", err
+		}
+		if _, err := f.Write(content); err != nil {
```
docker-agent, line 50: **[MEDIUM] Orphaned file left on disk when Write() fails in writeUniqueFile**

When `os.OpenFile` with `O_EXCL` succeeds (creating the file) but the subsequent `f.Write(content)` call fails, the code closes the file and returns an error — but never removes it:

```go
if _, err := f.Write(content); err != nil {
    f.Close()
    return "", err  // file still exists on disk, empty/partial
}
```

The newly-created-but-empty file remains on disk. On the next `--record` run, `os.OpenFile` with `O_EXCL` returns `os.IsExist` for that slot, skipping it permanently. After 100 such orphans the loop exhausts all slots and returns an error — even though none of the slots contain valid content.

**Fix**: remove the file before returning the write error:
```go
if _, err := f.Write(content); err != nil {
    f.Close()
    _ = os.Remove(path)
    return "", err
}
```

pkg/tui/recorder/generate.go
```diff
@@ -0,0 +1,395 @@
...
+
+// trimTrailingQuit drops the trailing Ctrl+C/Ctrl+D presses that ended the
+// session; replayed tests are torn down by the harness instead. It returns
+// the frame of the first dropped event — the settled screen the user quit
+// from — for the final WaitFor.
+func trimTrailingQuit(events []event) (kept []event, quitFrame string, trimmed bool) {
+	end := len(events)
+	for end > 0 {
+		e := events[end-1]
+		if e.kind != keyEvent || e.key.Mod != tea.ModCtrl || (e.key.Code != 'c' && e.key.Code != 'd') {
```
docker-agent: **[MEDIUM] trimTrailingQuit may not detect Ctrl+C/D if bubbletea v2 uses dedicated key codes**

The check assumes `Ctrl+C` is represented as `{Mod: tea.ModCtrl, Code: 'c'}`:

```go
if e.kind != keyEvent || e.key.Mod != tea.ModCtrl || (e.key.Code != 'c' && e.key.Code != 'd') {
    break
}
```

In bubbletea v2 (`charm.land/bubbletea/v2`), Ctrl+C/D may arrive as dedicated key constants (e.g., `tea.KeyCtrlC`) rather than as a modifier+letter combination. If the library surfaces them with `Mod == 0` and a dedicated `Code`, neither branch of this condition fires and the quit keystrokes are *not* trimmed — they remain in the generated test chain as `Press(...)` calls that would terminate the replay harness unexpectedly.

The rest of the codebase (e.g., `isTyping` checking `k.Mod != 0`) suggests the authors are working with modifier flags, which is consistent with the v2 design. However, it's worth adding a quick check against `tea.KeyCtrlC`/`tea.KeyCtrlD` constants as a fallback if the library exposes them, or adding a unit test that feeds a `KeyPressMsg` exactly as the real bubbletea input parser would produce it for Ctrl+C to lock in the assumption.

docker-agent (commented):

Two medium-severity issues found in the newly introduced recording infrastructure. The rest of the PR is well-structured: O_EXCL for safe file creation, FNV hash fallback for invalid identifiers, `%q` escaping for all user-supplied strings in generated code, and `commentSafe` for header injection hardening are all correct.

cmd/root/record_tui.go
dgageot: Fixed in bd5cc4812: the file is now removed when either `Write` or `Close` fails, so a failed write can no longer permanently occupy an O_EXCL slot.

pkg/tui/recorder/generate.go
dgageot: Addressed in bd5cc4812. Bubble Tea v2 has no dedicated `KeyCtrlC` constant — ultraviolet's key table maps ETX/EOT to `{Mod: ModCtrl, Code: letter}` on both the legacy and Kitty paths — but the strict `Mod != ModCtrl` check did break under Kitty lock modifiers. Extracted `isQuitKey` which masks Caps/Num/Scroll Lock and prefers `BaseCode` (matching `parseCtrlNumberKey` in pkg/tui), and added `TestIsQuitKey` pinning the parser's exact representations.

docker-agent: Thanks for the correction and the thorough fix! I was wrong about dedicated `KeyCtrlC` constants — you're right that bubbletea v2's ultraviolet parser normalizes ETX/EOT to `{Mod: ModCtrl, Code: letter}` on all input paths. The real issue you caught (Kitty lock modifiers breaking the strict equality check) is more subtle and important.

The `isQuitKey` implementation looks solid: masking `lockModifiers` and preferring `BaseCode` mirrors `parseCtrlNumberKey`'s logic perfectly, which keeps the codebase consistent. The test coverage you added pins down both the legacy and Kitty representations exactly as they flow from the parser, so this won't regress.

Sayt-0 (approved):

PR #3418, opened 2026-07-02
test: fix flaky TestAttachedServer_DeleteEmitsSessionExited

dgageot:
The SSE handler in the attached server flushes response headers before it subscribes to the event log via `StreamEvents`. In the test, the `HasEventSource` guard was effectively a no-op because the test itself registers the event source — so `DeleteSession` could race ahead and remove the event log before the handler had a chance to subscribe. The stream then closed with a clean EOF and no `session_exited` event, causing the test to fail intermittently (observed in CI run https://github.com/docker/docker-agent/actions/runs/28590850454/job/84773735898).

The fix adds a "hello" sentinel event to the fake event source. The test now reads that event from the SSE stream before calling `DeleteSession`, which proves the handler is actively subscribed before the session is torn down. The test then asserts the terminal `session_exited` event as before.

Validated with 100 runs under `-race`, the full `pkg/server` suite, `go vet`, and `golangci-lint` with zero issues.

### DIFF
--- a/pkg/server/server_attached_test.go
+++ b/pkg/server/server_attached_test.go
@@ -587,7 +587,8 @@ func TestAttachedServer_DeleteEmitsSessionExited(t *testing.T) {
 
 	sm := NewSessionManager(ctx, config.Sources{}, store, 0, &config.RuntimeConfig{})
 	sm.AttachRuntime(t.Context(), sess.ID, &fakeRuntime{}, sess)
-	sm.RegisterEventSource(sess.ID, func(ctx context.Context, _ func(any)) {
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The sentinel-event synchronization approach is correct and well-designed. The "hello" event is emitted before `DeleteSession` is called, providing a deterministic proof that the SSE handler is actively subscribed — eliminating the race that caused intermittent failures. No bugs were introduced by the changed lines.

Sayt-0 (approved):

PR #3420, opened 2026-07-02
test: fix flaky tuitest frame sync and TestListen_FD fd double-close

dgageot:
Follow-up to #3418 (flaky `TestAttachedServer_DeleteEmitsSessionExited`). While investigating that failure, two more independent root causes for flaky tests were found and fixed.

`TestListen_FD` in `pkg/server` was intermittently corrupting parallel tests. `Listen("fd://N")` adopts and closes descriptor N, but the test also closed it via `defer file.Close()`. The OS recycles N for a concurrent test's socket between the two closes, so the second close yanked live connections out from under unrelated tests. Symptoms ranged from crossed HTTP responses (`malformed HTTP status code "/ping"`) and dial i/o timeouts to 20-minute package-level hangs. The fix transfers fd ownership to `Listen` and pins the `os.File` so neither an explicit `Close` nor its GC finalizer can re-close the recycled descriptor. Before the fix, `go test -race -count=20 ./pkg/server/` failed 5/6 rounds; after, 8/8 rounds (160 iterations) pass and the suite runs ~7× faster.

The `sendSync` helper in `pkg/tui/tuitest` had a frame-counting assumption that broke under `-race`: it assumed exactly one frame per sent message, but startup and internal messages also produce frames, so it could return before the sent key was processed (`TestDebugLiveFrames` failed 8/50 locally). The debug sink also wrote frames *after* the store push that tests synchronize on, so a test could read a dump file before its content landed. The fix replaces frame counting with a FIFO barrier sentinel acknowledged by `captureModel`, writes sink output before the store push, and unblocks `sendSync` when the sent message causes the program to quit. A regression test (`TestSendSyncWithQuit`) covers that last case.

No behavior changes outside of test helpers; all existing tests continue to pass.

### DIFF
--- a/pkg/server/listen_test.go
+++ b/pkg/server/listen_test.go
@@ -14,6 +14,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// fdOwnershipPin keeps os.Files whose descriptor was handed to
+// Listen("fd://...") reachable for the life of the test process. Listen
+// adopts and closes the descriptor, so closing the os.File as well —
+// explicitly or via its GC cleanup — would close the same fd number a second
+// time after the OS has recycled it for another parallel test's socket,
+// corrupting that test's connection.
+var fdOwnershipPin []*os.File
+
 func TestListen_FD(t *testing.T) {
 	t.Parallel()
 
@@ -24,7 +32,8 @@ func TestListen_FD(t *testing.T) {
 
 	file, err := orig.(*net.TCPListener).File()
 	require.NoError(t, err)
-	defer file.Close()
+	// Ownership of file's fd passes to Listen; see fdOwnershipPin.
+	fdOwnershipPin = append(fdOwnershipPin, file)
 
 	ln, err := Listen(t.Context(), fmt.Sprintf("fd://%d", file.Fd()))
 	require.NoError(t, err)
--- a/pkg/server/listen_test.go
+++ b/pkg/server/listen_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


pkg/server/listen_test.go
```diff
@@ -24,7 +32,8 @@ func TestListen_FD(t *testing.T) {
 
 	file, err := orig.(*net.TCPListener).File()
 	require.NoError(t, err)
-	defer file.Close()
+	// Ownership of file's fd passes to Listen; see fdOwnershipPin.
+	fdOwnershipPin = append(fdOwnershipPin, file)
```
docker-agent: **[HIGH] `fdOwnershipPin` concurrent writes — data race under `go test -race`**

`fdOwnershipPin` is an unsynchronised package-level `[]*os.File`. `TestListen_FD` calls `t.Parallel()`, so when the suite is run with `-count=20 -parallel=N` (as described in the PR) multiple goroutines reach line 36 simultaneously and all execute `fdOwnershipPin = append(fdOwnershipPin, file)`. A concurrent `append` reads the slice header (pointer/len/cap) and may reallocate the backing array — this is a textbook data race that `go test -race` will detect (two goroutines access the same memory address with at least one write and no synchronisation).

The PR was explicitly motivated by fixing races found with `-count=20`, so a new race on the slice itself would be caught in the same CI run.

**Suggested fix** — protect the append with a mutex, or (simplest) use a `sync.Mutex` in the same file:

```go
var (
    fdOwnershipPinMu  sync.Mutex
    fdOwnershipPin    []*os.File
)

// in TestListen_FD:
fdOwnershipPinMu.Lock()
fdOwnershipPin = append(fdOwnershipPin, file)
fdOwnershipPinMu.Unlock()
```

Alternatively, since `TestListen_FD` is the only test that writes the pin and the purpose is purely to keep the `*os.File` alive past GC, a `runtime.KeepAlive(file)` call at the end of `TestListen_FD` (after `ln.Close()`) would achieve the same effect without any global state:

```go
defer runtime.KeepAlive(file)   // prevent GC finalizer from closing recycled fd
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One high-severity confirmed finding in the new test code introduced by this PR.

pkg/server/listen_test.go
dgageot: Fixed in 7cc4680df — the append is now guarded by a mutex via a `pinFDOwnership` helper. Went with the mutex rather than `runtime.KeepAlive`: KeepAlive only pins the file until the end of *this* test, but the recycled fd can belong to a parallel test that outlives it, so the pin must last for the whole process. Verified with 5×`go test -race -count=20 ./pkg/server/` (100 iterations), all passing.

docker-agent: Perfect fix — the mutex-protected helper is clean and the comment explaining the process-lifetime pinning requirement is excellent. You're absolutely right that `runtime.KeepAlive` would be insufficient here: I missed that the recycled fd can be acquired by a *different* parallel test that outlives the original, so pinning only until test end would still leave a window where the finalizer could corrupt the newer test's socket. The mutex approach correctly handles this.

Thanks for the thorough verification with 100 race-detector iterations!

Sayt-0 (approved):

Issue #3356, opened 2026-07-01
Add NVIDIA (build.nvidia.com / NIM) as a supported model provider


aheritier:
## Summary
Add first-class support for **NVIDIA** (NIM / build.nvidia.com) as an OpenAI-compatible model provider via the alias map.

## Why
NVIDIA provides access to Nemotron and many other open models through build.nvidia.com (with a free tier) and supports on-prem/self-hosted deployment via NVIDIA NIM. OpenAI-compatible API makes hosted access a simple alias addition, with a configurable base URL for on-prem NIM.

## Resources
- OpenAI-compatible base URL (hosted): `https://integrate.api.nvidia.com/v1`
- Token env var: `NVIDIA_API_KEY`
- Docs: https://docs.nvidia.com/nim/ and https://build.nvidia.com/
- Implement in: `pkg/model/provider/aliases.go`
- Test: `pkg/model/provider/aliases_test.go`

## Definition of Done
- [ ] Alias entry added: "nvidia": {APIType: "openai", BaseURL: "https://integrate.api.nvidia.com/v1", TokenEnvVar: "NVIDIA_API_KEY"}
- [ ] Unit test added; asserts alias resolves and `IsCatalogProvider("nvidia")` is true
- [ ] Provider appears in model catalog via `CatalogProviders()`
- [ ] On-prem NIM usage via custom `base_url` documented
- [ ] `task test` and `task lint` pass
- [ ] Example YAML config using an NVIDIA model added

Priyanshu-sde:
Hi @aheritier

I'd like to work on this issue. Here's my proposed implementation:

| # | Task | Approach |
|---|------|----------|
| 1 | Alias entry | Add `"nvidia"` in `pkg/model/provider/aliases.go` with `APIType: "openai"`, `BaseURL: "https://integrate.api.nvidia.com/v1"`, `TokenEnvVar: "NVIDIA_API_KEY"` |
| 2 | Unit tests | Extend `aliases_test.go` to cover alias resolution and `IsCatalogProvider("nvidia")` |
| 3 | Catalog integration | Confirmed automatic via non-empty `BaseURL`   no extra plumbing needed |
| 4 | Example config | Add `examples/nvidia.yaml` using `nvidia/llama-3.1-nemotron-70b-instruct` |
| 5 | On-prem NIM docs | Note in PR body that `base_url` override supports self-hosted NIM |
| 6 | Validation | Run `task test && task lint` |

**Notes:**
[...]

Issue #3344, opened 2026-07-01
Qwen 3.6 35B (served via vLLM) gives error HTTP 400 system message must be at the beignning


pkimtani:
### Description

Similar to #2327 

When using Qwen 3.6 35B MLX (served via Ollama), things seem to work fine.
When I switch to my AI box running Qwen 3.6 35B served via vLLM, I get error `HTTP 400: System message must be at the beignning`
I also tried switching to `nvidia/Qwen3.6-35B-A3B-NVFP4`.

I initially suspected that it might be chat_template issue on model side, but when I hooked Copilot CLI, it worked out of the box.

Docker Agent version:
```
docker-agent version
docker-agent version v1.91.0
Commit: e07c94cd22f83d2b10210de86f47374d2c0f94d5
```

### Docker Agent version

v1.91.0

### OS & terminal

macOS and docker sbx

### Error output

```shell

```

### Screenshots

<img width="1402" height="946" alt="Image" src="https://github.com/user-attachments/assets/ebab71c4-b9ff-45b7-ab10-04a24750a9d7" />

Issue #2327, opened 2026-04-05
`Jinja Exception: System message must be at the beginning` with custom provider, docker model runner and qwen 3.5 family models


k33g:
`docker-agent version 1.42.0`

In **theory**, when sending a set of messages with the OpenAI API to a local model, there should be only one message with a `system` role, and it must be the first message in the list.

Until now, inserting multiple messages with a `system` role hadn't really caused any issues (most chat templates are fairly permissive).

However, now, Qwen3.5's Jinja chat template rejects a message list of this type. 

Qwen3.5's official `chat_template.jinja` contains logic like:

```
{%- for message in messages %}
{%- set content = render_content(message.content, true)|trim %}
{%- if message.role == "system" %}
{%- if not loop.first %}
{{- raise_exception('System message must be at the beginning.')
}}
{%- endif %}
```
> https://huggingface.co/unsloth/Qwen3.5-2B/blob/main/chat_template.jinja#L85

This has an impact (error) on certain use cases of `docker-agent` with Docker Model Runner when using a custom provider.

### 1. `docker-agent` + `dmr` provider (it works)

If I use the model `huggingface.co/unsloth/qwen3.5-2b-gguf:Q4_K_M` with the `dmr` provider, I don't have any issues — every system-type message (such as those used for skill detection) is concatenated with the first system message:

**`docker-agent` provider**:

```yaml
models:

  brain:
    provider: dmr
    model: huggingface.co/unsloth/qwen3.5-2b-gguf:Q4_K_M
    temperature: 0.0
    top_p: 0.95
    presence_penalty: 1.5 
    max_tokens: 65536
```

**Request**:

```json
{
  "max_tokens": 65536,
  "messages": [
    {
[...]

PR #3421, opened 2026-07-02
feat: add NVIDIA provider as supported model provider

Priyanshu-sde:
## Summary

Add first-class support for NVIDIA (NIM / build.nvidia.com) as an OpenAI-compatible model provider. Hosted NVIDIA models work via a new alias, while self-hosted NVIDIA NIM is supported by overriding `base_url`.

## Changes

- Added a `nvidia` alias in `pkg/model/provider/aliases.go`:
  - `APIType: "openai"`
  - `BaseURL: "https://integrate.api.nvidia.com/v1"`
  - `TokenEnvVar: "NVIDIA_API_KEY"`
- Added `nvidia` to `TestCatalogAliases` in `pkg/model/provider/aliases_test.go`.
- Added `examples/nvidia.yaml` demonstrating NVIDIA NIM with the `nvidia/nemotron-3-super-120b-a12b` model.

## Notes

- The example model exists in `pkg/modelsdev/snapshot.json`, allowing example validation without special-casing.
- Self-hosted NVIDIA NIM is supported by overriding `base_url` to point to a local NIM endpoint.

## Testing

- ✅ `task test` — all tests pass.
- ✅ `task lint` — no issues.


Priyanshu-sde:
This pr closes #3356

pkg/model/provider/aliases.go
```diff
@@ -52,6 +52,11 @@ var Aliases = map[string]Alias{
 		BaseURL:     "https://api.studio.nebius.com/v1",
 		TokenEnvVar: "NEBIUS_API_KEY",
 	},
+	"nvidia": {
```
Sayt-0, line 55: Registering the alias here also needs a matching entry in `openModelHostProviders` (`pkg/model/provider/openai/client.go`). NVIDIA NIM serves open-weight models with strict chat templates, so the per-source system messages must be coalesced; otherwise a request carrying multiple system messages fails with an empty stream or `System message must be at the beginning` (#3145, #2327, #3344). The example in this PR (agent instruction plus three toolsets) emits multiple system messages and would exercise this path.

pkg/model/provider/aliases_test.go
```diff
@@ -46,6 +46,7 @@ func TestCatalogAliases(t *testing.T) {
 		"together":    {APIType: "openai", BaseURL: "https://api.together.xyz/v1", TokenEnvVar: "TOGETHER_API_KEY"},
 		"huggingface": {APIType: "openai", BaseURL: "https://router.huggingface.co/v1", TokenEnvVar: "HF_TOKEN"},
 		"moonshot":    {APIType: "openai", BaseURL: "https://api.moonshot.ai/v1", TokenEnvVar: "MOONSHOT_API_KEY"},
+		"nvidia":      {APIType: "openai", BaseURL: "https://integrate.api.nvidia.com/v1", TokenEnvVar: "NVIDIA_API_KEY"},
```
Sayt-0, line 49: Every OpenAI-compatible alias also has a row in the `openAIAliasProviders` end-to-end table (`pkg/model/provider/openai_alias_providers_test.go`). Please add one for `nvidia` with `mergesSystemMessages: true` so the system-message coalescing behavior is covered.

examples/nvidia.yaml
```diff
@@ -0,0 +1,15 @@
+models:
```
Sayt-0, line 3: The sibling provider examples (cerebras, deepseek, baseten, ...) start with the schema header. Please add it here for consistency:

```yaml
# yaml-language-server: $schema=../agent-schema.json
```

Sayt-0 (changes requested):
Thanks for adding NVIDIA NIM as a built-in alias. The core wiring is correct (`APIType`, `base_url`, `TokenEnvVar`), `TestCatalogAliases` covers it, and the example model resolves against the models.dev snapshot, so `TestParseExamples` passes without special-casing. A few items should be addressed before this is a complete first-class provider, matching how the other OpenAI-compatible aliases are wired.

## Functional

- Register `nvidia` in `openModelHostProviders` (`pkg/model/provider/openai/client.go`). NVIDIA NIM fronts open-weight models (Nemotron, Llama, Qwen, DeepSeek) whose chat templates often reject more than one system message. Without this entry, the per-source system messages are not coalesced, which reproduces the empty-stream / `System message must be at the beginning` failures (#3145, #2327, #3344). The shipped example already emits multiple system messages (agent instruction plus the `filesystem`, `shell`, and `think` toolsets), so it is likely to hit this path.
- Add a row for `nvidia` to the `openAIAliasProviders` end-to-end table (`pkg/model/provider/openai_alias_providers_test.go`) with `mergesSystemMessages: true`, matching every other alias.

## Docs (each existing alias ships these)

| File | Content |
| --- | --- |
| `docs/providers/overview/index.md` | built-in providers table |
| `docs/concepts/models/index.md` | providers table |
| `docs/configuration/models/index.md` | provider enumerations (lines 20, 51, 403) |
| `docs/providers/nvidia/index.md` | dedicated provider page (missing) |

## Optional (non-blocking)

- `agent-schema.json` provider description ends in `etc.`, so adding `nvidia` there is nice-to-have.
- Absence from `cloudProviders` / `DefaultModels` in `pkg/config/auto.go` is consistent with `nebius` / `xai` / `minimax` (nvidia simply will not participate in auto-selection). `NVIDIA_API_KEY` is still forwarded as a credential via `ProviderAPIKeyEnvVars()` since it iterates the aliases.

See inline comments for the specific anchors.

Priyanshu-sde:
Thanks so much for the detailed and thorough review, @Sayt-0 ,
I really appreciate you taking the time to point out the exact anchors  instead of just flagging the gaps. It made it easy to understand it better.
I've pushed the fixes:
- Registered nvidia in openModelHostProviders so consecutive system messages get coalesced.
- Added a nvidia row to openAIAliasProviders with mergesSystemMessages: true.
- Added the # yaml-language-server: $schema=../agent-schema.json header to examples/nvidia.yaml.


This is genuinely one of the more educational reviews I've gotten. I'll keep digging into the codebase and try to pick up more issues I can meaningfully contribute to. Thanks again for the guidance!

Sayt-0:
Hi @Priyanshu-sde ! You're welcome. Don't hesitate if you want to contribute again !

Sayt-0:
The changes looks good. Approved ! Have a nice day !

Sayt-0 (dismissed):

Sayt-0:
one last thing: your commits must be signed in this repo

Priyanshu-sde:
Hi @Sayt-0 , I have updated to signed commits, also that's a new thing i got to learn about Thank you

Sayt-0:
`canonical-check` (docs-lint) is failing: the new `docs/providers/nvidia/index.md` is missing the `canonical:` front-matter field, which recently became required on every docs page (convention added after this PR was opened).

**Fix:** in the front matter of `docs/providers/nvidia/index.md`, add the `canonical:` line inside the top `---` block (path-derived value, same pattern as the other provider pages like `docs/providers/nebius/index.md`):

```yaml
---
title: "NVIDIA NIM"
description: "Use NVIDIA NIM models with docker-agent."
keywords: docker agent, ai agents, model providers, llm, nvidia, nim, nemotron
weight: 290
canonical: https://docs.docker.com/ai/docker-agent/providers/nvidia/   # <-- add this line
---
```

The URL must match the file path exactly (`providers/nvidia/`) and end with a trailing slash, otherwise the check still fails. After adding it and rebasing on `main`, `canonical-check` goes green.

Don't hesitate to ping me when you're done !

Priyanshu-sde:
@Sayt-0 I have added the mentioned line , please have a look

dgageot (approved):

PR #3423, opened 2026-07-02
fix: bound toolset startup and never block the TUI exit when Docker is wedged

dgageot:
When Docker Desktop is wedged and containers fail to start, running an agent with MCP toolsets produced two interrelated failures. First, the sidebar "N tools available…" spinner would animate forever: `emitStartupInfo` had a timeout for *listing* tools but not for toolset *Start*, and the default resilient lifecycle profile carries no startup timeout while the MCP connector detaches its context with `context.WithoutCancel`, so a wedged `docker`-backed MCP server blocked `emitToolsProgressively` indefinitely and the terminal `ToolsetInfo{Loading:false}` event was never emitted. Second, confirming the Ctrl-C exit dialog did nothing: `cleanupAll` called `cleanupManagedResources` (supervisor shutdown → toolset `Stop`) synchronously inside the Bubble Tea `Update` loop, and `Stop` blocks on the `StartableToolSet` mutex held by the wedged `Start`, so `tea.Quit` was never returned — and the shutdown safety net was only armed after that blocking call.

The fix addresses both independently and then hardens the shutdown path. Toolset `Start` is now raced against a new `toolStartTimeout` (default 30 s, overridable via `WithToolStartTimeout` for tests) inside `startToolsetWithTimeout`, mirroring the existing `listToolsWithTimeout`. A toolset that times out is skipped in the startup sidebar pass with a user-visible warning; the original start attempt continues in the background and the toolset is picked up once it becomes ready. For the exit path, managed-resource cleanup is now launched in a background goroutine from `cleanupAll`, and the shutdown safety net races both `program.Wait()` and that cleanup goroutine against the deadline, force-exiting whichever stalls first. Ctrl-C and confirm now always exits within the shutdown timeout (5 s) even with Docker fully wedged. A final follow-up commit wraps the entire `cleanupAll` sequence in a `sync.Once` so that repeat invocations cannot stack goroutines behind a wedged cleanup or arm parallel safety nets, and rewords the start-timeout warning so it does not over-promise — the abandoned `Start` goroutine holds the toolset's single-flight lock, meaning the next turn waits for the original attempt rather than spawning a fresh retry.

Three regression tests cover the new behaviour: `TestEmitStartupInfo_SkipsToolsetWhoseStartHangs`, `TestCleanupAll_WedgedResourceCleanupForcesExit`, and `TestCleanupAll_RepeatCallsDoNotStackOnWedgedCleanup`. All pass under `-race`. The only pre-existing test failures (`TestLoadExamples` for `dmr.yaml` and `unload_on_switch.yaml`) also fail on the unmodified baseline and require a working local Docker model runner.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -315,6 +315,11 @@ type LocalRuntime struct {
 	// defaultToolListTimeout; overridden via WithToolListTimeout.
 	toolListTimeout time.Duration
 
+	// toolStartTimeout bounds how long EmitStartupInfo waits for a single
+	// toolset to start before skipping it for the startup sidebar pass.
+	// Defaults to defaultToolStartTimeout; overridden via WithToolStartTimeout.
+	toolStartTimeout time.Duration
+
 	// pauseMu guards pauseCh.
 	pauseMu sync.Mutex
 	// pauseCh is non-nil and open while /pause has paused the run loop;
@@ -482,6 +487,19 @@ func WithToolListTimeout(d time.Duration) Opt {
 	}
 }
 
+// WithToolStartTimeout overrides how long EmitStartupInfo waits for a single
+// toolset to start before skipping it. Defaults to defaultToolStartTimeout.
+// A non-positive value is ignored so the default stands. Tests pass a short
+// timeout to exercise the skip path (a toolset whose Start() blocks) without
+// a real-time wait.
+func WithToolStartTimeout(d time.Duration) Opt {
+	return func(r *LocalRuntime) {
+		if d > 0 {
+			r.toolStartTimeout = d
+		}
+	}
+}
+
 // WithRetryOnRateLimit enables automatic retry with backoff for HTTP 429 (rate limit)
 // errors when no fallback models are available. When enabled, the runtime will honor
 // the Retry-After header from the provider's response to determine wait time before
@@ -570,6 +588,7 @@ func NewLocalRuntime(ctx context.Context, agents *team.Team, opts ...Opt) (*Loca
 		providerRegistry:       provider.DefaultRegistry(),
 		maxOverflowCompactions: defaultMaxOverflowCompactions,
 		toolListTimeout:        defaultToolListTimeout,
+		toolStartTimeout:       defaultToolStartTimeout,
 		dmrModelLister:         dmr.ListModels,
 	}
 	r.bgAgents = agenttool.NewHandler(r)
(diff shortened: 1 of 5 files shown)


pkg/runtime/runtime.go
```diff
@@ -1562,8 +1581,19 @@ func (r *LocalRuntime) emitToolsProgressively(ctx context.Context, a *agent.Agen
...
 		if startable, ok := toolset.(*tools.StartableToolSet); ok {
-			if err := startable.Start(ctx); err != nil {
+			if err := startToolsetWithTimeout(ctx, startable, r.toolStartTimeout); err != nil {
 				desc := tools.DescribeToolSet(startable.ToolSet)
+				// A start that outlived its deadline (e.g. an MCP container
+				// stuck behind a wedged Docker daemon) is reported directly:
+				// the abandoned Start goroutine has not returned, so the
+				// once-per-streak guard below has recorded nothing and would
+				// silently swallow the warning.
+				if errors.Is(err, context.DeadlineExceeded) && ctx.Err() == nil {
```
docker-agent, line 1591: **[LOW] `context.Canceled` from parent-ctx shutdown may reach `ShouldReportFailure` and surface a confusing user warning**

When the parent `ctx` is cancelled (e.g., normal application shutdown) before the new `startToolsetWithTimeout` deadline fires, `startCtx.Done()` closes and `startCtx.Err()` returns `context.Canceled`. The caller correctly skips the `context.DeadlineExceeded` branch (`ctx.Err()` is non-nil), but the `context.Canceled` error then falls through to `ShouldReportFailure()`. If that flag is armed, the user sees a toolset warning like *"context canceled"* during what is otherwise a clean exit.

Before this PR, `Start` was called with the plain `ctx` directly and would have been cancelled too, but the call site didn't exist at all — so the new error path is genuinely new. Consider short-circuiting the whole block when the parent context is done:

```go
if ctx.Err() != nil {
    return ctx.Err()  // or just: continue
}
if err := startToolsetWithTimeout(ctx, startable, r.toolStartTimeout); err != nil {
    ...
}
```

pkg/tui/tui.go
```diff
@@ -2596,46 +2602,68 @@ func (m *appModel) cleanupManagedResources() {
...
+			ed.Cleanup()
+		}
+
+		// Shut down managed resources (supervisor, TUI state store) in the
+		// background: supervisor.Shutdown stops every session's toolsets, which
+		// can block indefinitely (e.g. an MCP toolset wedged behind a broken
+		// Docker daemon). Running it synchronously here would freeze the Update
+		// loop before tea.Quit is ever returned, leaving the exit confirmation
+		// apparently ignored.
+		cleanupDone := make(chan struct{})
```
docker-agent, line 2624: **[LOW] Background cleanup goroutine has no watchdog when `program == nil` (test paths)**

The `cleanupManagedResources` goroutine is started unconditionally before the `program == nil` early-return. When there is no program (unit tests, or early exit before `SetProgram`), the safety-net goroutine is never armed, so a wedged `cleanupManagedResources` (e.g., `TestCleanupAll_WedgedResourceCleanupForcesExit` without a real program) runs forever — bounded only by the test binary's overall timeout, not the 5 s shutdown deadline.

In the existing test `TestCleanupAll_RepeatCallsDoNotStackOnWedgedCleanup`, the cleanup function blocks indefinitely; if that test uses a `nil` program, the goroutine will outlive the `cleanupAllOnce` guard and may interfere with subsequent tests that share the same process.

Consider closing `cleanupDone` unconditionally when there is no program, or adding a context-based watchdog in the `program == nil` branch:

```go
if program == nil {
    go func() {
        select {
        case <-cleanupDone:
        case <-time.After(timeout):
            slog.Warn("Cleanup timed out (no program)")
        }
    }()
    return
}
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

The two core fixes — bounding toolset `Start` via `startToolsetWithTimeout` and making managed-resource cleanup async in `cleanupAll` — are sound and correctly address the reported bugs. The `sync.Once` guard on `cleanupAll` is well-placed. The sequential deadline-race loop in the safety-net goroutine is valid (the `time.After` channel fires once and is correctly shared across both selects). Two low-severity observations are left as inline comments.

Sayt-0 (approved):

gtardif (approved):

PR #3428, opened 2026-07-02
fix: route --record proxy through the configured models gateway

dgageot:
When `docker agent run <agent> --record` was used with a models gateway configured, the recording proxy bypassed the gateway entirely and forwarded requests directly to providers' public endpoints. This caused HTTP 401 errors (e.g. Anthropic "invalid x-api-key") because the gateway's managed credentials were never used, and environment API keys that happened to be set were injected instead.

The recording proxy now chains through the configured upstream gateway. `GatewayTargetURL` preserves the gateway's path prefix, query parameters, and percent-encoded paths when rewriting request targets. Cassette entries are stored under the canonical provider URL (via `WithRecordURL`) so existing `--fake` replay continues to work without modification.

The second commit hardens the forwarding path against credential leakage: the Docker Desktop token is never forwarded to non-Docker gateways (credential headers are stripped and provider env keys are re-applied instead); requests with an unknown or missing `X-Cagent-Forward` host are rejected rather than silently misrouted; the gateway URL is validated at startup; and full target URLs are kept out of logs and cassettes so gateway query parameters that may carry secrets are never recorded. New tests in `pkg/fake/proxy_gateway_test.go` cover forwarding, auth handling, cassette contents, and credential-leakage prevention.

### DIFF
--- a/pkg/fake/proxy.go
+++ b/pkg/fake/proxy.go
@@ -13,6 +13,7 @@ import (
 	"maps"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"os"
 	"regexp"
 	"strings"
@@ -21,6 +22,9 @@ import (
 	"github.com/labstack/echo/v4"
 	"gopkg.in/dnaeon/go-vcr.v4/pkg/cassette"
 	"gopkg.in/dnaeon/go-vcr.v4/pkg/recorder"
+
+	"github.com/docker/docker-agent/pkg/desktop"
+	"github.com/docker/docker-agent/pkg/environment"
 )
 
 // ProxyOptions configures the fake proxy behavior.
@@ -30,6 +34,10 @@ type ProxyOptions struct {
 	// StreamChunkDelay is the delay between SSE chunks when SimulateStream is true.
 	// Defaults to 15ms if not set.
 	StreamChunkDelay time.Duration
+	// UpstreamGateway, when set, forwards requests to this models gateway
+	// instead of the provider's public endpoint. Used when recording a
+	// session that normally routes through a models gateway.
+	UpstreamGateway string
 }
 
 // ProxyOption is a function that configures ProxyOptions.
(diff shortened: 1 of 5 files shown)


aheritier (approved):

pkg/fake/proxy.go
```diff
@@ -119,19 +142,59 @@ func StartStreamingRecordingProxy(
...
+	if environment.IsTrustedDockerURL(gateway) {
+		// The Desktop token the client attached is exactly what the gateway
+		// expects (localhost gateways are equally trusted in normal mode).
+		return func(string, *http.Request) {}
+	}
+	// Non-Docker gateways must never see the Desktop token. Strip the
+	// credential headers the client set and re-apply the provider env key
+	// it would have sent to the gateway directly (SDKs default to env keys
+	// when no gateway token is available).
+	return func(host string, req *http.Request) {
```
docker-agent, line 191: **[MEDIUM] gatewayAuthHeaderUpdater silently strips credentials for unrecognized provider hosts**

When a non-Docker third-party gateway is configured and the `X-Cagent-Forward` host is not one of the 5 hard-coded providers (`api.openai.com`, `api.anthropic.com`, `generativelanguage.googleapis.com`, `api.mistral.ai`, `openrouter.ai`), `envAPIKeyForHost` returns `""`, so the `if` guard at line 195 prevents `APIKeyHeaderUpdater` from being called. However, the three `Header.Del` calls above it (lines 192–194) have already executed unconditionally, stripping `Authorization`, `X-Api-Key`, and `X-Goog-Api-Key` from the outgoing request.

The result is that any custom or unrecognized provider endpoint (Azure OpenAI, Cohere, Bedrock-compatible, private OpenAI-compatible deployments, etc.) routed through a non-Docker gateway will have **all credential headers deleted with no replacement**, producing silent 401 errors from the gateway. The comment says "re-apply the provider env key it would have sent to the gateway directly", but the re-apply is skipped for the unknown-host case with no warning or fallback.

**Suggestion**: For unrecognized hosts, preserve the original `Authorization` header instead of deleting it (the goal is stripping the Docker Desktop token specifically, which is only present for Docker-trusted URLs — the early return above already handles that case):
```go
return func(host string, req *http.Request) {
    if envAPIKeyForHost(host) != "" {
        // Known provider: strip Desktop token and apply provider env key
        req.Header.Del("Authorization")
        req.Header.Del("X-Api-Key")
        req.Header.Del("X-Goog-Api-Key")
        APIKeyHeaderUpdater(host, req)
    }
    // Unknown host: leave whatever credentials the client sent intact
}
```

docker-agent (commented):

This PR correctly routes the `--record` proxy through the configured models gateway and hardens credential handling. The URL construction, cassette URL override via context (`WithRecordURL`), gateway URL validation at startup, and Docker-trusted gateway pass-through are all sound. One credential-handling edge case in the newly introduced `gatewayAuthHeaderUpdater` warrants attention.

PR #3431, opened 2026-07-03
fix(remote): recover proxy after cooldown instead of latching to direct

trungutt:
## Motivation

`fallbackTransport` was designed to gracefully degrade when the HTTP proxy socket that docker-agent optionally routes through disappears mid-session. In practice the "graceful degradation" is a one-way trap: the first socket error flips a `proxyDisabled` flag that is never cleared, so every subsequent request in that process goes direct forever — even after the proxy comes back.

For short-lived CLI invocations this is fine. For a long-lived process (any interactive session, server, or agent host that outlives a single proxy outage) it means the proxy stops being used the first time the socket blips, and the only way to get it back is to restart the process.

```
Before                                     After
------                                     -----

t0  proxy up                               t0  proxy up
    request -> proxy: OK                       request -> proxy: OK

t1  proxy stops                            t1  proxy stops
    request -> proxy: error                    request -> proxy: error
    proxyDisabled = true (permanent)           proxy disabled until t1+30s
    fallback to direct                         fallback to direct

t2  proxy back up                          t2  proxy back up (still in cooldown)
    request -> direct forever                  request -> direct

    ...                                    t1+30s+ cooldown expires
                                               next request -> proxy: OK
                                               back to normal
```

## Why a cooldown rather than always retrying

Always re-probing on every request would add a dial round-trip to every hot-path request while the proxy is down. A short cooldown (30s) keeps the request path cheap during an extended outage while still recovering automatically — one dial per cooldown, not per request.

## Why not just remove the fallback altogether

Some traffic through this transport (OCI push/pull, catalog lookups) works fine on a direct connection when the proxy is unavailable. Auth-required traffic doesn't, and callers already own that concern. This PR changes only the recovery behaviour, not the routing decision.

### DIFF
--- a/pkg/remote/transport.go
+++ b/pkg/remote/transport.go
@@ -17,10 +17,9 @@ import (
 
 var memoizer = memoize.New[bool](1 * time.Minute)
 
-// NewTransport returns an HTTP transport that uses Docker Desktop proxy if
-// available. If the proxy becomes unavailable during the session, it falls
-// back to direct connections and re-probes the proxy after a short cooldown,
-// so a long-lived process recovers automatically once the proxy is back.
+// NewTransport returns an HTTP transport that uses the Docker Desktop proxy
+// if available, and falls back to direct connections while re-probing the
+// proxy after a cooldown so long-lived processes recover on their own.
 func NewTransport(ctx context.Context) http.RoundTripper {
 	t, ok := http.DefaultTransport.(*http.Transport)
 	if !ok {
@@ -55,27 +54,17 @@ func NewTransport(ctx context.Context) http.RoundTripper {
 	return transport
 }
 
-// proxyRetryCooldown is how long we skip the proxy after a socket failure
-// before probing it again. Keeps the request path cheap when the proxy is
-// down for an extended period (one dial per cooldown rather than per
-// request) while still recovering automatically when the proxy comes back.
+// Bounded backoff: one probe per cooldown, not per request.
 const proxyRetryCooldown = 30 * time.Second
 
-// fallbackTransport wraps a proxy transport and falls back to a direct transport
-// when the proxy socket becomes unavailable (e.g., Docker Desktop proxy dies).
-//
-// Failures are treated as temporary: after a proxy socket error we skip the
-// proxy for proxyRetryCooldown, then probe it again on the next request. This
-// lets a long-lived agent process recover automatically once the proxy is
-// available again, instead of latching into direct mode for the rest of its
-// lifetime.
+// fallbackTransport tries the proxy first, direct second. A socket error
+// disables the proxy for proxyRetryCooldown, so a stale error can't latch
+// the transport into direct mode for the rest of the process's lifetime.
 type fallbackTransport struct {
 	proxy  *http.Transport
 	direct *http.Transport
 
-	// disabledUntilUnixNano holds the earliest wall-clock time at which the
-	// proxy may be tried again, encoded as time.Time.UnixNano(). Zero means
-	// the proxy is currently enabled.
+	// Zero = enabled. Non-zero = disabled until this UnixNano deadline.
 	disabledUntilUnixNano atomic.Int64
 }
 
@@ -94,9 +83,6 @@ func (f *fallbackTransport) DisableCompression() {
 	f.direct.DisableCompression = true
 }
 
-// proxyEnabled reports whether the proxy should be tried on this request. It
-// also clears an expired cooldown so the next caller in the cooldown-expired
-// state re-enters the "enabled" branch cleanly.
 func (f *fallbackTransport) proxyEnabled() bool {
 	until := f.disabledUntilUnixNano.Load()
 	if until == 0 {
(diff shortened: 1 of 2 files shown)


pkg/remote/transport.go
```diff
@@ -80,10 +94,30 @@ func (f *fallbackTransport) DisableCompression() {
...
+// state re-enters the "enabled" branch cleanly.
+func (f *fallbackTransport) proxyEnabled() bool {
+	until := f.disabledUntilUnixNano.Load()
+	if until == 0 {
+		return true
+	}
+	if time.Now().UnixNano() < until {
+		return false
+	}
+	f.disabledUntilUnixNano.CompareAndSwap(until, 0)
```
docker-agent, line 95: **[LOW] Theoretical CAS/clock-collision race between proxyEnabled() and disableProxy()**

If a goroutine calls `proxyEnabled()` and loads an already-expired `until` value, then a *second* goroutine calls `disableProxy()` and stores a new deadline *before* the first goroutine's `CompareAndSwap(until, 0)` runs, the CAS will only succeed if the new deadline (now+30s as UnixNano) happens to equal the old expired deadline — an astronomically unlikely clock collision. In practice this cannot happen, but the comment above `proxyEnabled()` could note the subtlety for future readers. No code change required.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly replaces a permanent `proxyDisabled` boolean with an `atomic.Int64` cooldown timer. The design is sound: zero means enabled, non-zero means disabled until that Unix-nano timestamp, and `CompareAndSwap` prevents redundant clears. No high or medium severity issues were found — only minor theoretical edge cases noted below as informational.

**Changed files reviewed:** `pkg/remote/transport.go`, `pkg/remote/transport_test.go`

pkg/remote/transport_test.go
```diff
@@ -143,3 +146,134 @@ type testError struct {
...
+	resp, err = ft.RoundTrip(req)
+	require.NoError(t, err)
+	_ = resp.Body.Close()
+	require.Equal(t, int32(1), ft.fakeProxy.calls.Load(), "proxy should be skipped during cooldown")
+	require.Equal(t, int32(2), ft.fakeDirect.calls.Load())
+
+	// Simulate the cooldown expiring by clearing the disabled marker directly
+	// (equivalent to time.Now() advancing past disabledUntilUnixNano). Also
+	// swap the proxy to a healthy variant so the retry succeeds.
+	ft.disabledUntilUnixNano.Store(0)
```
docker-agent: **[LOW] Recovery test bypasses natural cooldown expiry path by directly zeroing disabledUntilUnixNano**

`TestFallbackTransport_ProxyRecoversAfterCooldown` simulates cooldown expiry with `ft.disabledUntilUnixNano.Store(0)` (line 228) rather than storing a past deadline. This bypasses the `CompareAndSwap` clear path inside `proxyEnabled()`. The dedicated unit test `TestFallbackTransport_ProxyEnabledAfterCooldownExpires` does exercise that path in isolation, so there's no uncovered code — but the end-to-end recovery flow (expired timestamp → CAS clear → probe → re-enable) isn't exercised through `fakeFallback.RoundTrip`. Consider storing `time.Now().Add(-time.Second).UnixNano()` instead of `0` to make the integration path complete.

trungutt: Fixed in 4ce7e36cd8a05bd2e74ddcc1a430d826f7a86833: the test now stores a past deadline instead of zero, so the end-to-end path exercises the `CompareAndSwap` clear branch inside `proxyEnabled()`.

pkg/remote/transport.go
trungutt: Leaving as-is. The race you describe is harmless in both directions: the only writer that stores a non-zero value is `disableProxy()` with `time.Now() + 30s` (nanoseconds, monotonically non-repeating in practice), so a failed CAS just means a concurrent `disableProxy()` won — which is the correct outcome. The reader that failed the CAS then returns `true` and its request pays one extra proxy attempt, which is exactly what happens on the first request after any failure anyway. Not worth a comment on top of the `CompareAndSwap` idiom, which already documents itself for readers familiar with atomics.

docker-agent: Fair point — you're right that the failed CAS is harmless because the goroutine that "lost" just returns true and tries the proxy, which is the correct fallback behavior when concurrent disables are racing. The monotonic nanosecond timestamps make the collision scenario purely theoretical. Thanks for the clarification on the intended semantics.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The cooldown-based proxy recovery mechanism is well-structured. The  usage is correct, the CAS-clear idiom is sound, and test coverage is comprehensive. Two minor edge cases are noted in inline comments — neither is a correctness hazard for the stated use case.

pkg/remote/transport.go
docker-agent, line 95: **[LOW] `proxyEnabled` returns `true` even when the CAS fails due to a concurrent `disableProxy` call**

After `Load()` returns an expired timestamp, if a concurrent `disableProxy()` stores a new future deadline between the `Load()` and the `CompareAndSwap(until, 0)`, the CAS will correctly fail to clear the value — but the function still returns `true`, causing this goroutine to attempt the proxy during an active new cooldown window.

This is a one-request race that leaks a single probe through immediately after a new disable event; it is not a safety hazard or data-loss risk. If tighter semantics are desired, re-reading the value after a failed CAS and returning `false` when the new value is in the future would close the gap:

```go
if !f.disabledUntilUnixNano.CompareAndSwap(until, 0) {
    // A concurrent disableProxy() stored a new deadline; honour it.
    return false
}
return true
```

pkg/remote/transport.go
```diff
@@ -80,10 +94,30 @@ func (f *fallbackTransport) DisableCompression() {
...
+	return true
+}
+
+// disableProxy marks the proxy as unavailable for proxyRetryCooldown.
+func (f *fallbackTransport) disableProxy() {
+	f.disabledUntilUnixNano.Store(time.Now().Add(proxyRetryCooldown).UnixNano())
```
docker-agent, line 100: **[LOW] `disableProxy` unconditionally overwrites the deadline, extending the cooldown on concurrent failures**

`disableProxy()` uses `Store(time.Now().Add(proxyRetryCooldown).UnixNano())` unconditionally. If multiple goroutines hit socket errors at the same time, each resets the deadline to a fresh `now + 30s`, potentially sliding the cooldown window forward with each call rather than anchoring it to the first failure.

For most use cases this is inconsequential (the cooldown only extends slightly), but if a stricter "cooldown starts at first failure" guarantee is wanted, a CAS that only writes when the value is currently `0` would prevent subsequent concurrent callers from extending the window:

```go
func (f *fallbackTransport) disableProxy() {
    deadline := time.Now().Add(proxyRetryCooldown).UnixNano()
    f.disabledUntilUnixNano.CompareAndSwap(0, deadline)
}
```

pkg/remote/transport.go
trungutt: Same race as the earlier thread ([here](https://github.com/docker/docker-agent/pull/3431#discussion_r3518412357)), same conclusion. The reviewer's own analysis ("a one-request race that leaks a single probe through immediately after a new disable event; it is not a safety hazard") matches mine: harmless, not worth the code.

pkg/remote/transport.go
trungutt: Leaving as-is. Anchoring the cooldown to the *most recent* observed failure is the intended behaviour, not a bug -- if the proxy is still failing right now, we want to keep skipping it for another 30s from *now*, not expire the window because it happened to start 29s ago from an earlier failure. The proposed CAS-from-zero variant would inhibit that refresh. And, as the review itself notes, the drift is inconsequential either way.

docker-agent: Thanks for clarifying the intent — that makes perfect sense. A sliding cooldown that extends from the *most recent* failure is the right behavior here, ensuring the proxy stays disabled for a full 30s from whenever issues were last observed, not from the first failure in a burst. I misread this as unintended drift rather than the designed behavior.

Appreciate the explanation!

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly replaces the permanent `proxyDisabled bool` latch with an `atomic.Int64` cooldown timestamp, enabling automatic proxy recovery after 30 s. The core logic — CAS-based expiry clear in `proxyEnabled()` and timestamp-refresh in `disableProxy()` — is sound. Test coverage exercises the failure path, cooldown-skip, CAS-clear on expiry, non-socket errors, and state-machine transitions directly.

No high or medium severity issues were found. Three minor observations (not blocking):

1. **Low — concurrent CAS interaction**: When the cooldown expires simultaneously across goroutines, each calls `CAS(until, 0)`. Only one succeeds; the others observe the new value and fall through correctly. The design is safe but the interaction with a concurrent `disableProxy()` store is subtle — a brief comment would aid future maintainers.

2. **Low — thundering-herd at cooldown boundary**: All goroutines racing past an expired cooldown will each probe the proxy in parallel. For Docker Desktop's typical concurrency this is acceptable, but under high concurrency a singleflight/probe-goroutine pattern would be more efficient.

3. **Low — `bodyErr` silently discarded**: When `req.GetBody()` fails, the returned error is the original proxy socket error, not the body-read error. This is a pre-existing pattern and this PR does not make it worse.

dgageot (approved):

PR #3441, opened 2026-07-03
fix: show proper toolset names in /tools dialog for loader-created toolsets

dgageot:
Toolsets created directly by the team loader — `skills`, `transfer_task`, `handoff`, `deferred`, and `code_mode` — bypass the registry's `tools.WithName` wrapper. Because they implemented neither `tools.Named` nor `tools.Describer`, the `/tools` TUI dialog fell back to Go type names such as `*skills.ToolSet`, which made the picker hard to read and navigate.

Each of those toolset types now implements `tools.Named`, returning a stable, human-readable name. The `transfer_task` and `handoff` implementations reuse their existing `ToolName` constants so the name is defined in exactly one place. `RestartToolset` in the runtime was also tightened: it now skips non-restartable name matches rather than aborting on the first hit, since display names are not guaranteed to be unique across toolsets.

No behaviour changes outside the `/tools` dialog display names.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -1030,23 +1030,30 @@ func (r *LocalRuntime) AgentToolsetStatuses(name string) []tools.ToolsetStatus {
 //   - no toolset matches name (matching uses the same logic as the
 //     /tools dialog: the toolset's Name() if any, otherwise its
 //     description),
-//   - the toolset is not supervisor-backed (no Restartable capability),
+//   - no matching toolset is supervisor-backed (no Restartable capability),
 //   - the supervisor itself returned an error (timeout, classified
 //     transport failure, etc.).
+//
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes correctly add `tools.Named` implementations to the five loader-created toolsets (`skills`, `transfer_task`, `handoff`, `deferred`, `code_mode`). Each `Name()` method returns a stable, human-readable string consistent with the project's naming convention. The `transfer_task` and `handoff` implementations reuse existing `ToolName` constants as intended.

The `RestartToolset` logic change in `runtime.go` — switching from an early `return false` on the first non-restartable name match to `continue` with exhaustive iteration — is semantically correct: it now properly handles the case where multiple toolsets share a display name, skipping non-restartable entries rather than aborting on the first hit.

No bugs were found in the introduced code.

aheritier (approved):

PR #3443, opened 2026-07-03
feat(picker): add Lean Mode checkbox to agent picker

dgageot:
The full-screen agent picker (`--agent-picker`) had no way to control the TUI mode at launch time, forcing users to know in advance whether they wanted the lean interface and pass `--lean` on the command line before the picker even appeared.

This adds a "Lean Mode" checkbox to the bottom of the picker panel. It is off by default and can be toggled with the `l` key or by clicking it with the mouse. When ticked, the chosen agent launches in the lean TUI; when unticked, it launches in the full one. The checkbox is seeded with the *effective* lean state derived from `--lean` and the user-config lean default, so it never shows a value that disagrees with what would actually run. The returned value is authoritative: unticking the checkbox overrides a lean default from user config, and ticking it enables lean even if no flag was passed.

The follow-up commit tightens the seeding logic, extracts `panelOrigin` and `cardRows` helpers to remove duplication between the hit-test and render paths, and replaces an open-coded `lastClickIndex = -1` reset with `resetClickTracking`. Tests cover the default state, `l`-key and mouse-click toggling, hit-zone boundaries (including just-inside and just-outside the checkbox row), and layout consistency between the render and hit-test helpers.

### DIFF
--- a/cmd/root/agent_picker.go
+++ b/cmd/root/agent_picker.go
@@ -78,15 +78,16 @@ func loadAgentChoices(ctx context.Context, refs []string, env environment.Provid
 	return choices
 }
 
-// selectAgentRef shows a full-screen picker and returns the chosen agent ref.
-// When only a single ref is supplied there is nothing to choose, so it is
-// returned directly without showing any UI.
-func selectAgentRef(ctx context.Context, refs []string, env environment.Provider) (string, error) {
+// selectAgentRef shows a full-screen picker and returns the chosen agent ref
+// along with whether the user ticked the "Lean Mode" checkbox. When only a
+// single ref is supplied there is nothing to choose, so it is returned
+// directly without showing any UI.
+func selectAgentRef(ctx context.Context, refs []string, env environment.Provider) (ref string, lean bool, err error) {
 	if len(refs) == 0 {
-		return "", errors.New("no agent refs to choose from")
+		return "", false, errors.New("no agent refs to choose from")
 	}
 	if len(refs) == 1 {
-		return refs[0], nil
+		return refs[0], false, nil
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR adds a **Lean Mode checkbox** to the agent picker TUI. The implementation is clean and correct:

- **Checkbox seeding**: `initialLean` is computed from `f.lean || (!f.leanChanged && userconfig.Get().Lean && !f.tour)` — this correctly mirrors `applyUserSettings` logic, so the checkbox always shows the value that would actually run.
- **Hit-test math**: `leanCheckboxAt()` places the checkbox at `originY + agentPickerCardsTop + m.cardRows() + 1`, which correctly accounts for the blank line between the card list and the checkbox row.
- **Panel sizing**: `panelSize()` changes `+2` → `+4` to cover `blank + checkbox + blank + help` — the arithmetic is consistent with `render()`.
- **State threading**: The picker's returned `lean` bool is correctly assigned to `f.lean` and `f.leanChanged = true` in `run.go`, making the checkbox authoritative over both `--lean` flag and user-config defaults as described.
- **`f.leanChanged` placement**: Moving `f.leanChanged = cmd.Flags().Changed("lean")` before the picker block is necessary so `initialLean` is computed with the correct `leanChanged` value.
- **`cardRows()` refactor**: Extracting the helper cleanly deduplicates the hit-test and render paths.
- **Tests**: Cover default state, `l`-key toggle, mouse-click toggle, and boundary conditions for hit-zone detection.

Sayt-0 (approved):

PR #3444, opened 2026-07-03
feat: agent picker discovers ~/.agents configs when no list is given

dgageot:
`--agent-picker` without an explicit list now offers more than just the two built-in agents. On top of `default` and `coder` it scans `~/.agents` and appends any agent config files (`.yaml`, `.yml`, `.hcl`) it finds there directly — no nested directories, no special files. Cards for local configs promote the agent's `description` field to the card title, with the `~`-shortened file path shown as the muted detail line (and as the YAML dialog title). This makes the picker a practical launchpad for anyone who keeps a small library of personal agents alongside the catalogue entries.

The follow-up hardening commit addresses all the edge cases that a real `~/.agents` directory can surface: the card list is now windowed to the terminal height with a scrolling cursor and an "x–y of N" subtitle so a large directory can't overflow the panel; file discovery skips non-regular files (FIFOs, sockets) while keeping symlinks to regular files; the `--agent-picker` `NoOptDefVal` sentinel is replaced with the self-documenting literal `"defaults"`; filenames containing terminal escape sequences are stripped before they reach the dialog title; and whitespace-only or control-character-only descriptions are not promoted to card titles. The windowing arithmetic is integrated with the Lean Mode checkbox added in #3443 so `cardRows` counts only the visible slice and the panel-overhead constant stays accurate.

The `~/.agents` discovery and the card-list windowing both have new unit tests. No existing behaviour changes when `--agent-picker` is passed an explicit comma-separated list.


Sayt-0 (dismissed):

docker-agent (commented):
### Assessment: 🟢 APPROVE

No high or medium severity bugs found. The `~/.agents` discovery, windowed card list, and hardening changes (escape stripping, FIFO skip, sentinel rename, description guard) are all well-implemented. Three low-severity observations are noted as inline comments.

cmd/root/agent_picker.go
```diff
@@ -193,19 +248,47 @@ func (m *agentPickerModel) moveUp() {
...
+// visibleCount returns the number of cards shown at once: all of them when
+// they fit (or before the first WindowSizeMsg), otherwise as many as fit the
+// terminal height, at least one.
+func (m *agentPickerModel) visibleCount() int {
+	n := len(m.choices)
+	if m.height <= 0 {
+		return n
+	}
+	stride := agentPickerCardHeight + agentPickerCardGap
+	fit := (m.height - agentPickerPanelOverhead + agentPickerCardGap) / stride
```
docker-agent, line 275: **[LOW] visibleCount() divides against terminal height, not available panel height**

`visibleCount()` computes `fit = (m.height - agentPickerPanelOverhead + agentPickerCardGap) / stride`, where `m.height` is the full terminal height. The rendered panel is a floating box centred with `lipgloss.Place`; when the terminal is taller than the maximum panel height the formula overestimates `fit`, potentially returning more visible slots than cards actually drawn. The `min(n, max(fit, 1))` guard in line 272 prevents out-of-bounds access, so behaviour is correct today, but the derivation is conceptually wrong and will silently break if the overhead constant ever diverges from the real chrome height.

cmd/root/agent_picker.go
```diff
@@ -576,7 +667,12 @@ func (m *agentPickerModel) panelSize() (w, h int) {
 // render and panelSize so their layout math can't drift apart.
 func (m *agentPickerModel) headerText() (title, subtitle, help string) {
 	title = styles.HighlightWhiteStyle.Render("Choose an agent to run")
-	subtitle = styles.MutedStyle.Render("Pick the agent you want to start a session with, or double-click a card.")
+	subtitleText := "Pick the agent you want to start a session with, or double-click a card."
+	if n := m.visibleCount(); n < len(m.choices) {
```
docker-agent, line 672: **[LOW] Subtitle width change during windowed scroll can shift panel origin for one frame**

`headerText()` returns a longer subtitle string (e.g. `"3–5 of 12"`) once windowing activates. Because `headerText()` is called inside `panelSize()` → `panelOrigin()` → `cardAt()`, every `MouseMotionMsg` re-derives the panel bounding box. When the subtitle grows or shrinks (e.g. going from 9 → 10 agents), the panel width changes by a few runes mid-scroll, shifting `panelOrigin()` and causing mouse-click zones to mis-track the visible cards for one frame.

Consider computing `panelSize()` once per `WindowSizeMsg` / navigation event and caching it rather than recalculating on every mouse motion.

cmd/root/agent_picker.go
```diff
@@ -16,14 +19,61 @@ import (
...
+	if err != nil {
+		return nil
+	}
+	var refs []string
+	for _, entry := range entries {
+		if !isConfigFileName(entry.Name()) {
+			continue
+		}
+		ref := filepath.Join(dir, entry.Name())
+		if info, err := os.Stat(ref); err != nil || !info.Mode().IsRegular() {
```
docker-agent, line 63: **[LOW] TOCTOU window between os.Stat and config.Load in agentRefsInDir**

`os.Stat(ref)` (line 60) follows symlinks and confirms the target is a regular file, but by the time `config.Load` (or equivalent) later reads the file, a symlink target could have been replaced with a special file (FIFO, socket, device). In practice this is negligible for a single-user CLI tool, but a comment here noting the intentional trade-off would align with the project's `AGENTS.md` philosophy of documenting non-obvious edge cases.

cmd/root/agent_picker.go
dgageot: Addressed in 198295245: `panelSize` now derives its height from the shared `agentPickerPanelOverhead` constant (`h = m.cardRows() + agentPickerPanelOverhead`), so `visibleCount` and the rendered chrome can no longer drift apart. Also documented that `fit` may exceed the card count on tall terminals and is clamped.

cmd/root/agent_picker.go
dgageot: Addressed in 198295245: the scroll indices in the subtitle are now padded to the width of the total (`%*d`), so the subtitle — and thus the centred panel geometry used by mouse hit-testing — keeps a constant width while scrolling. Added a regression test asserting `panelSize` is stable across scroll offsets.

cmd/root/agent_picker.go
dgageot: Addressed in 198295245: added a comment documenting the intentional TOCTOU trade-off — the stat guards against stray FIFOs/sockets, not races, which is acceptable for a single-user CLI.

docker-agent: Thanks for adding the comment! The clarification makes the design intention clear — the check defends against accidental special files, not race conditions. This aligns well with the pragmatic approach outlined in AGENTS.md.

Sayt-0 (approved):

PR #3450, opened 2026-07-03
feat(tui): add /context inventory and /drop to remove attached files

### LINKED ISSUE #3435: feat(tui): context inventory — list and remove attached/prompt files (`/drop`)
## Summary
Let users see what files are currently in context and remove them — an Aider-style `/add` / `/drop` inventory. docker-agent has `/attach` and `@file` references but no way to view or remove them.

## Proposed behavior
- Extend the `/context` view (or add a `/drop` command) to list attached files and prompt files, each with a per-item token estimate.
- Allow removing an attachment from the session.

## Implementation notes (starting points)
- Add a removal API to `Session.AttachedFiles` in `pkg/session/session.go` (~718-745)
- `pkg/tui/dialog/context.go`, `pkg/tui/commands/commands.go`
- Keep `@ref` completion consistent in `pkg/tui/components/editor/editor.go`

Complexity: M. (Note: this pairs naturally with the `/context` view but should be shippable on its own against the current attachment model.)

Background: shared plan `context-management-gap-analysis` (Phase 2, item 4).

## Dependencies / Relationships
- **Blocked by:** #3432 (extends the `/context` dialog)

Sayt-0:
Closes #3435

## Summary

Adds an Aider-style context inventory to the TUI: a `/context` dialog listing every file currently entering the session context (attachments and prompt files) with per-item token estimates, plus the ability to drop attachments, interactively or via `/drop <path>`.

## Issue expectations mapping

| Issue expectation | Implemented |
|---|---|
| List attached files and prompt files with a per-item token estimate | `/context` dialog with "Attached files" and "Prompt files (agent config)" sections, `~N tokens` per file (size/4, the same rule of thumb as the session truncation budget), plus a totals line |
| Allow removing an attachment from the session | `d`/`x`/`delete`/`backspace` on the selected entry, or `/drop <path>` directly; `/drop` without argument opens the dialog |
| Removal API on `Session.AttachedFiles` | `Session.RemoveAttachedFile(absPath) bool`, mutex-guarded, next to `AddAttachedFile` |
| Keep `@ref` completion consistent (`editor.go`) | No editor change needed: `@` completion lists workspace files (session-independent); re-mentioning a dropped file re-attaches it through the existing send path |

## Flow

```
/context ──► ShowContextDialogMsg ──► Session.AttachedFilesSnapshot()
                                      + App.CurrentAgentPromptFiles()   (local runtime only)
                                  ──► NewContextDialog(BuildContextFiles(...))

d (dialog) ─┐
/drop path ─┴► DropAttachedFileMsg ──► expand ~/$VAR + filepath.Abs
                                   ──► Session.RemoveAttachedFile()
                                   ──► SessionStore.UpdateSession()
                                   ──► notification toast
```

## Notes

- Prompt files are listed but not droppable: they are re-resolved from the agent config at every turn (`add_prompt_files` hook), so dropping one is refused with an info toast.
- Prompt-file paths are resolved with the same lookup rules as the turn-start hook (`promptfiles.PathsFromEnv`), exposed through an optional runtime capability; remote runtimes simply omit the section.
- Files deleted since attachment render as `missing` (no token estimate).
- Dropped entries are persisted via `SessionStore.UpdateSession` so resumed sessions do not resurrect them.
- Long lists use a sliding window that keeps the selection visible.

## Testing

- Unit tests: `Session.RemoveAttachedFile` (3 cases) and the context dialog (8 cases: entry building, empty state, sections/token rendering, navigation bounds, drop semantics, prompt-file refusal, selection clamping, close).
- `go build ./...` and `golangci-lint run`: clean.
- Full `go test ./...`: green except `TestURLSource_Read_RejectsLocalAddresses` (pkg/config), which fails identically on `main` in this environment (network-dependent, unrelated to this change).

### DIFF
--- a/docs/features/tui/index.md
+++ b/docs/features/tui/index.md
@@ -79,6 +79,8 @@ Type `/` during a session to see available commands, or press <kbd>Ctrl</kbd>+<k
 | `/yolo`            | Toggle automatic tool call approval                                                  |
 | `/title`           | Set or regenerate session title                                                      |
 | `/attach`          | Attach a file to your message                                                        |
+| `/context`         | List attached files and prompt files in context, each with a token estimate; press <kbd>d</kbd> to drop the selected attachment |
+| `/drop`            | Remove an attached file from context (`/drop <path>`, or `/drop` alone to pick from the list) |
 | `/shell`           | Open a shell                                                                         |
 | `/star`            | Star/unstar the current session                                                      |
 | `/cost`            | Show cost breakdown for this session                                                 |
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -267,6 +267,25 @@ func (a *App) CurrentAgentSkills() []skills.Skill {
 	return st.Skills()
 }
 
+// promptFilesProvider is an optional runtime capability: resolving the
+// current agent's add_prompt_files entries to on-disk paths. Only the local
+// runtime implements it; remote runtimes don't, so the /context dialog
+// simply omits the prompt-files section for them.
+type promptFilesProvider interface {
+	CurrentAgentPromptFiles() []string
+}
+
+// CurrentAgentPromptFiles returns the resolved prompt-file paths injected
+// into the current agent's context at each turn, or nil when the runtime
+// cannot resolve them (remote runtime or no prompt files configured).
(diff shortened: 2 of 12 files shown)


Sayt-0:
need to improve /context first

PR #3451, opened 2026-07-03
feat: add docker agent board, a Kanban TUI for orchestrating agents

dgageot:
Working on several agent tasks in parallel means juggling terminals, worktrees, and half-remembered session IDs. This adds `docker agent board`, a full-screen Kanban TUI that turns that workflow into cards on a pipeline: each card launches an agent in a tmux session on an isolated git worktree, and moving a card forward (Dev → Review → Push → Done) delivers the destination column's prompt to its agent. It is a TUI port of the experimental web-based board project, built natively on docker-agent's own primitives (`--worktree`, `--session`, `--listen`).

The engine (`pkg/board`) keeps one watcher goroutine per card tailing the agent's control-plane event stream over a per-card unix socket, mirroring title and running/waiting/paused/failed status into a JSON store, and relaunching the tmux session (resuming the same conversation and worktree) if the agent dies. Prompt delivery, busy-move rejection, and delete/relaunch races are all serialized through the store and a relaunch lock. Projects and column prompts live in the user's global config file (`~/.config/cagent/config.yaml`, new `board:` section) and are editable from the TUI, which follows the main TUI's standards: shared dialog chrome, theme, keymap (remapped quit/suspend bindings are honored), scrollbars, and mouse support. Pressing enter attaches the terminal to a card's agent; `ctrl+q` detaches back to the board.

The board is single-instance (flock on the state file), sanitizes agent-controlled strings before rendering, runs its tmux sessions on a private validated socket, and requires tmux (it fails with a clear message otherwise; Windows builds compile but the command is effectively unix-only). Docs live in `docs/features/board/`.

### DIFF
--- a/cmd/root/board.go
+++ b/cmd/root/board.go
@@ -1,0 +1,36 @@
+package root
+
+import (
+	"github.com/spf13/cobra"
+
+	boardtui "github.com/docker/docker-agent/pkg/board/tui"
+	"github.com/docker/docker-agent/pkg/telemetry"
+)
+
+// newBoardCmd creates the `docker agent board` command: a Kanban TUI that
+// orchestrates one agent per card, each running in a tmux session on an
+// isolated git worktree. Projects and column prompts are configured in the
+// user's global config file, or from the TUI itself.
+func newBoardCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "board",
+		Short: "Orchestrate agents on a Kanban board",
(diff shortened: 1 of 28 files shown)


rumpl:
Didn't look at the code, one note: i'm using the tasks toolset a lot, can you make sure that the board can be reused? I would want to have a board for the tasks at some point

aheritier (approved):

Issue #3442, opened 2026-07-03
feat(cli): guide users from "no API key" to a working model (actionable errors, doctor, setup, tutorial)


Sayt-0:
## Summary
First contact with docker-agent on a machine without provider credentials is confusing.
Typing just `docker agent` can silently start a TUI on a model that is not pulled and fail
with a raw HTTP 404 at the first message, or exit 0 with no output at all in a non-TTY
context. Error messages mention concepts that are never named ("built-in environment
variable providers"), `--env-from-file` failures are silently swallowed, and no command
answers "what is my model/credential state?". Docs cover each topic in isolation but no
tutorial connects run-time failures to fixes.

## Current behavior (reproduced on 1138afae)

Bare `docker agent` (no args, implicit `run` of the default agent, model `auto`:
anthropic -> openai -> google -> mistral -> ... -> dmr/ai/qwen3):

| # | Environment | Result | Problem |
|---|---|---|---|
| 1 | TTY, no keys, DMR not installed | `AutoModelFallbackError`: pull / install / configure one of ~20 env vars | Correct but a wall of text; no interactive next step, no `docker agent models` or docs link |
| 2 | TTY, no keys, DMR endpoint reachable but `docker model` CLI plugin missing/broken | TUI starts silently on `dmr/ai/qwen3` (not pulled); first message fails with raw `HTTP 404: POST .../chat/completions` | "Not installed" detection is an exact string match on `"unknown flag: --json"` (`dmr/client.go:88-94`); any other status-query failure is logged and the pull skipped. The DMR HTTP API had already reported 0 models before the message was sent |
| 3 | non-TTY (pipe, CI) | Exit code 0, zero output | Silent no-op; no "needs a terminal, use `--exec`" error (`getting-started` does check isatty) |

Other scenarios:

| # | Scenario | Result | Problem |
|---|---|---|---|
| 4 | `run agent.yaml` (cloud model), no keys | `RequiredEnvError`: "...Store those secrets using one of the built-in environment variable providers." | Providers never named, no example command, no docs link, no local-model alternative |
| 5 | `run --exec` hitting the scenario-2 state | `❌ model failed: ... 404 ... Error: model failed: ... 404` | Error printed twice (CLI printer + cobra) |
| 6 | `run --env-from-file typo.env` | Same "must be set" error as #4 | Missing/malformed file only logged via `slog` (invisible without `--debug`), run continues; the flag recommended by error #4 itself no-ops on a typo |
[...]

PR #3452, opened 2026-07-03
feat(cli): actionable errors for missing models and credentials (phase 1 of #3442)

Sayt-0:
Part of #3442 (Phase 1: actionable errors). Phases 2-4 (`doctor`, guided `setup`, docs tutorial) are independently shippable and left to follow-up PRs, as laid out in the issue.

## Changes mapped to the issue scenarios

| # | Scenario | Before | After |
|---|---|---|---|
| 3 | Bare `docker agent`, non-TTY, empty stdin (CI) | exit 0, zero output | error naming the next steps (pass a message, pipe one, or use a terminal), exit 1; piped prompts keep working |
| 2 | DMR endpoint reachable, `docker model` CLI broken, model not pulled | TUI starts, first message fails with raw `HTTP 404: POST .../chat/completions` | fails at client creation: "model X is not pulled in Docker Model Runner" + `docker model pull X` |
| 1/2 | "Not installed" detection | exact string match on the full docker usage text | content match on `unknown flag: --json`, shared between `NewClient` and `ListModels` |
| 4 | Missing model credentials (`RequiredEnvError`, `first_available` variant) | "use one of the built-in environment variable providers" (never named) | the secret sources able to hold arbitrary keys are named with a one-line example each (export, `--env-from-file`, `pass`, macOS Keychain), plus the docs URL and the local-model alternative (only when the missing vars are model credentials) |
| 6 | `run --env-from-file typo.env` | missing/malformed file logged via `slog` (invisible without `--debug`), run continues | run aborts: `--env-from-file: open typo.env: no such file or directory`; parse errors name the file and expected format |
| 5 | Exec-mode failure | printed twice (CLI error printer + cobra `Error: ...`) | printed once by cobra, on stderr |

## Sample output (missing model credentials)

```
Error: The following environment variables must be set:
 - ANTHROPIC_API_KEY

Provide them using any of these sources:
 - Shell environment:  export ANTHROPIC_API_KEY=<value>
 - Env file:           docker agent run --env-from-file <file> ...
 - pass:               pass insert ANTHROPIC_API_KEY
 - macOS Keychain:     security add-generic-password -a "$USER" -s ANTHROPIC_API_KEY -w

See https://docs.docker.com/ai/docker-agent/guides/secrets/ for details.

No API key? Run a local model instead: docker agent run --model dmr/ai/qwen3 ...
(the model is pulled on first use; `docker model ls` shows models already pulled)
```

## Message coherence pass

Follow-up to the review comment about message consistency. Every claim made by the new messages is now backed by the code path it describes:

| Message | Problem | Fix |
|---|---|---|
| `ModelNotAvailableError` | "is not available" in the full message but "is not pulled" in the summary shown under `AutoModelFallbackError`; final bullet diverged from `PullFailedError`'s | both say "is not pulled in Docker Model Runner"; final bullet identical to `PullFailedError`'s |
| Secret sources help | suggested Docker Desktop and the credential helper, but `DockerDesktopProvider` and `CredentialHelperProvider` only resolve `DOCKER_EMAIL`/`DOCKER_USERNAME`/`DOCKER_TOKEN` and can never satisfy the reported missing keys | both lines removed; a `NotContains` test locks the exclusion |
| Local-model hint | claimed pulled models are "detected automatically by the default `auto` model" (the built-in agent uses `first_available`, and pinned agentfile models are never re-selected) and pointed at `docker agent models` for local models (that command deliberately skips DMR discovery and lists the catalog) | now: "the model is pulled on first use; `docker model ls` shows models already pulled" |
| Empty-stdin errors | the two sibling messages introduced their next steps with different separators | aligned on ": " |

Pre-existing divergences left out of this PR, candidates for follow-ups:

- `docs/guides/secrets/` presents Docker Desktop and the credential helper as generic secret sources (same conflation as above, docs side)
- "To fix this, you can:" (`auto.go`) vs "To resolve this, you can:" (`pull.go`, `available.go`)
- `ErrNotInstalled` lowercase "docker model runner" branding
- lowercase fallback branch "no first_available candidate has credentials configured (tried: ...)" vs the capitalized main branch

## Implementation notes

```
docker agent run (dmr model, no BaseURL / MODEL_RUNNER_HOST)
  |- docker model status ok        -> inspect + pull prompt (unchanged)
  |- status fails, "unknown flag"  -> ErrNotInstalled (content match, was exact match)
  `- status fails otherwise        -> GET <dmr>/models at client creation
       |- model listed             -> proceed
       |- model absent             -> ModelNotAvailableError ("docker model pull X")
       `- endpoint unqueryable     -> "cannot query Docker Model Runner at <url>" + install link
```

- `ModelNotAvailableError` passes through the teamloader unwrapped (same treatment as `PullFailedError`) and exposes a one-line summary when nested under `AutoModelFallbackError`, so guidance blocks do not stack.
- The non-TTY check lives at input time in `cli.Run` rather than at bare-command dispatch: a dispatch-level `isatty(stdout)` check would break the working `echo "prompt" | docker agent > out.txt` flow. The acceptance criterion (actionable error, exit != 0, never a silent no-op) is met either way; if the dispatch-level check seems preferable, it can be changed.
- `--env-from-file` validation reuses the provider chain built by `RuntimeConfig.EnvProvider()` (new `EnvFilesError()` accessor) and aborts in the shared command pre-run, so every command taking the flag is covered.
- `RequiredEnvError.MissingModelCredentials` is computed in `gatherMissingEnvVars` so the local-model hint is not shown for missing tool secrets (e.g. `GITHUB_PERSONAL_ACCESS_TOKEN`).

## Testing

| Area | Coverage |
|---|---|
| `pkg/environment` | message content for both flag states, sources that cannot hold arbitrary keys are excluded, parse error names the file |
| `pkg/model/provider/dmr` | `errIndicatesNotInstalled`, `modelAvailable` (tag/repo matching), `checkModelAvailable` via httptest (pulled / not pulled / unqueryable), `NewClient` shim test with variant docker usage text |
| `pkg/cli` | empty stdin (bare and `-`) fails, piped stdin still works, `ErrorEvent` returned once and not printed |
| `pkg/config` | `EnvFilesError` (missing, malformed, valid, clone), `MissingModelCredentials` wiring, updated `first_available` assertions |
| `cmd/root` | pre-run aborts on missing and malformed `--env-from-file` |

All scenarios also verified end to end against a built binary (including a live reproduction of scenario 2: reachable DMR, broken `docker model` CLI, not-pulled model). The message coherence pass only changes user-facing strings; the updated unit tests assert the new content. Existing e2e expectations (`environment variables must be set`, `OPENAI_API_KEY` on stderr) are preserved. Remaining suite failures (`TestURLSource_Read_RejectsLocalAddresses`, SSRF tests, `TestLoadExamples` DMR pull) reproduce on `main` and are environment-related.

### DIFF
--- a/cmd/root/flags.go
+++ b/cmd/root/flags.go
@@ -103,6 +103,13 @@ func addGatewayFlags(cmd *cobra.Command, runConfig *config.RuntimeConfig, loadUs
 
 		env := runConfig.EnvProvider()
 
+		// A missing or malformed --env-from-file must abort the run: the flag
+		// is the documented way to supply credentials, so silently continuing
+		// without it leads to confusing "env var must be set" errors later.
+		if err := runConfig.EnvFilesError(); err != nil {
+			return fmt.Errorf("--env-from-file: %w", err)
+		}
+
 		// Precedence: CLI flag > environment variable > user config
 		if runConfig.ModelsGateway == "" {
 			if gateway, _ := env.Get(ctx, envModelsGateway); gateway != "" {
--- a/cmd/root/flags_test.go
+++ b/cmd/root/flags_test.go
@@ -2,6 +2,8 @@ package root
 
 import (
 	"context"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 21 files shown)


dgageot:
@Sayt-0 we need to review some of the messages

dgageot (approved):

PR #3453, opened 2026-07-03
feat(mcp/remote): support unix:// URLs in remote MCP client

maxcleme:
I wanted to expose MCP server running on my host to `docker-agent` running inside a container and realized it was forced to use http port for this.

This PR add support for unix socket so one can run:

```
      - type: mcp
        remote:
          url: unix:///tmp/mcp-notify.sock
          transport_type: streamable
```

If that's reasonable I'd like to delegate the documentation to someone else :)

### DIFF
--- a/pkg/tools/mcp/remote.go
+++ b/pkg/tools/mcp/remote.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
 	"net/http/cookiejar"
 	neturl "net/url"
+	"strings"
 
 	gomcp "github.com/modelcontextprotocol/go-sdk/mcp"
 
@@ -77,6 +79,12 @@ func sanitizeRemoteAddress(rawURL string) string {
 	return u.Host
 }
 
+// unixSocketPath mirrors pkg/server.Listen: it strips the "unix://" prefix so
+// `unix:///tmp/x.sock` maps to `/tmp/x.sock`.
+func unixSocketPath(rawURL string) (string, bool) {
+	return strings.CutPrefix(rawURL, "unix://")
+}
+
 func (c *remoteMCPClient) Initialize(ctx context.Context, _ *gomcp.InitializeRequest) (*gomcp.InitializeResult, error) {
 	// Create HTTP client with OAuth support. We keep a reference to the
 	// oauthTransport so we can enrich Connect errors with the server's own
@@ -89,15 +97,22 @@ func (c *remoteMCPClient) Initialize(ctx context.Context, _ *gomcp.InitializeReq
 
 	var transport gomcp.Transport
 
+	endpoint := c.url
+	if _, ok := unixSocketPath(c.url); ok {
+		// http.Client can't dial the "unix" scheme; the socket is dialed by
+		// the transport, the host here is a placeholder.
+		endpoint = "http://localhost/"
+	}
+
 	switch c.transportType {
 	case "sse":
 		transport = &gomcp.SSEClientTransport{
-			Endpoint:   c.url,
+			Endpoint:   endpoint,
 			HTTPClient: httpClient,
 		}
 	case "streamable", "streamable-http":
 		transport = &gomcp.StreamableClientTransport{
-			Endpoint:             c.url,
+			Endpoint:             endpoint,
 			HTTPClient:           httpClient,
 			DisableStandaloneSSE: true,
 		}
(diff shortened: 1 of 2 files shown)


pkg/tools/mcp/remote.go
```diff
@@ -224,8 +239,16 @@ func (c *remoteMCPClient) createHTTPClient() (*http.Client, *oauthTransport, err
 }
 
 func (c *remoteMCPClient) headerTransport() http.RoundTripper {
+	base := http.DefaultTransport
+	if path, ok := unixSocketPath(c.url); ok {
+		t := http.DefaultTransport.(*http.Transport).Clone()
```
docker-agent, line 244: **[MEDIUM] Unchecked type assertion on `http.DefaultTransport` will panic if replaced**

`http.DefaultTransport.(*http.Transport).Clone()` is a direct (non-checked) type assertion introduced by this PR. It panics at runtime if `http.DefaultTransport` has been replaced with any non-`*http.Transport` implementation — for example, a test shim, an observability wrapper, or any middleware that replaces the default transport.

While `http.DefaultTransport` is `*http.Transport` by default, it is an exported `var` (`var DefaultTransport RoundTripper`) that can be freely swapped. Any code path that sets a unix:// URL **and** runs in an environment with a replaced `DefaultTransport` will crash.

Consider using a comma-ok assertion with a safe fallback:
```go
base := http.DefaultTransport
if path, ok := unixSocketPath(c.url); ok {
    var t *http.Transport
    if dt, ok := http.DefaultTransport.(*http.Transport); ok {
        t = dt.Clone()
    } else {
        t = &http.Transport{}
    }
    t.DialContext = func(ctx context.Context, _, _ string) (net.Conn, error) {
        return (&net.Dialer{}).DialContext(ctx, "unix", path)
    }
    base = t
}
```

pkg/tools/mcp/remote.go
```diff
@@ -77,6 +79,12 @@ func sanitizeRemoteAddress(rawURL string) string {
 	return u.Host
 }
 
+// unixSocketPath mirrors pkg/server.Listen: it strips the "unix://" prefix so
+// `unix:///tmp/x.sock` maps to `/tmp/x.sock`.
+func unixSocketPath(rawURL string) (string, bool) {
+	return strings.CutPrefix(rawURL, "unix://")
```
docker-agent, line 85: **[LOW] `unixSocketPath` silently accepts relative socket paths without validation**

`strings.CutPrefix(rawURL, "unix://")` strips the scheme prefix but does not validate that the resulting path is absolute. For `unix:///tmp/x.sock` the result is `/tmp/x.sock` (correct), but for a malformed URL like `unix://relative/path.sock` the result is `relative/path.sock` — a relative path silently passed to `net.DialContext` as a unix socket address.

Unix domain sockets technically support relative paths (resolved against the process's CWD), so no immediate error is raised at the `DialContext` call — the connection will likely just fail with a confusing "no such file or directory" error rather than a clear validation message.

The comment above the function documents the intended use as `unix:///tmp/x.sock` (absolute path). A simple guard would make the contract explicit:
```go
func unixSocketPath(rawURL string) (string, bool) {
    path, ok := strings.CutPrefix(rawURL, "unix://")
    if !ok || !strings.HasPrefix(path, "/") {
        return "", false
    }
    return path, true
}
```

docker-agent (commented):

pkg/tools/mcp/remote.go
docker-agent, line 244: **[MEDIUM] Unsafe type assertion `http.DefaultTransport.(*http.Transport)` may panic**

The direct (non-comma-ok) type assertion on line 244 will panic at runtime if `http.DefaultTransport` has been replaced with any type other than `*http.Transport`. While Go's standard library initialises `DefaultTransport` as `*http.Transport`, nothing prevents user code, a third-party library, or a test helper from swapping it out — a common pattern in testing (e.g. `http.DefaultTransport = myMockTransport`).

The panic would only fire when a `unix://` URL is used, but it would be a hard crash with no recovery. A safer approach is to use the comma-ok form and fall back to a fresh `&http.Transport{}`:

```go
if path, ok := unixSocketPath(c.url); ok {
    var t *http.Transport
    if dt, ok := http.DefaultTransport.(*http.Transport); ok {
        t = dt.Clone()
    } else {
        t = &http.Transport{}
    }
    t.DialContext = func(ctx context.Context, _, _ string) (net.Conn, error) {
        return (&net.Dialer{}).DialContext(ctx, "unix", path)
    }
    base = t
}
```

docker-agent (commented):

One medium-severity issue found in the new unix socket code path. The rest of the implementation (endpoint substitution, custom dialer, integration test) is sound.

dgageot (approved):

PR #3454, opened 2026-07-03
ci: build and publish docker-agent sandbox templates

dgageot:
Deploy sbx templates for docker-agent on each merge to main and tag.

```
main:
docker/docker-agent-sbx-templates:edge

release:
docker/docker-agent-sbx-templates:v1.x
docker/docker-agent-sbx-templates:latest
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,22 @@ jobs:
           cache-from: type=gha,scope=buildx-${{ matrix.platform }}
           cache-to: type=gha,mode=min,scope=buildx-${{ matrix.platform }}
 
+      # The sandbox template shares the builder-linux stage, so it reuses the
+      # binary compiled by the step above.
+      - name: Build template
+        uses: docker/build-push-action@ee4ca427a2f43b6a16632044ca514c076267da23 # v6.19.0
+        with:
+          context: .
+          target: template
+          platforms: ${{ matrix.platform }}
+          push: false
+          sbom: false
+          provenance: false
+          build-args: |
+            GIT_TAG=pr
+            GIT_COMMIT=dev
+          cache-from: type=gha,scope=buildx-${{ matrix.platform }}
+
   build-and-push-image:
     if: github.event_name != 'pull_request' && !github.event.repository.fork
     needs: [lint, build-and-test, license-check]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -236,3 +285,45 @@ jobs:
...
+      - name: Create manifest list and push
+        working-directory: ${{ runner.temp }}/template-digests
+        run: |
+          if [[ "${GITHUB_REF}" == refs/tags/v* ]]; then
+            tags=(-t "docker/docker-agent-sbx-templates:${GITHUB_REF_NAME#v}" -t "docker/docker-agent-sbx-templates:latest")
+          else
+            tags=(-t "docker/docker-agent-sbx-templates:edge")
+          fi
+          args=()
+          for digest in *; do args+=("docker/docker-agent-sbx-templates@sha256:${digest}"); done
```
docker-agent, line 323: **[MEDIUM] Glob `for digest in *` may pass literal `*` to `imagetools create` if directory is unexpectedly empty**

When bash runs `for digest in *` in a directory and `nullglob` is not set (the default), an empty directory leaves `*` unexpanded — so `args` would contain `docker/docker-agent-sbx-templates@sha256:*` instead of real digests. This would cause `docker buildx imagetools create` to fail with a confusing error rather than a clean CI failure.

The `if-no-files-found: error` guard on the upload step provides strong protection in the happy path, but the download step (`actions/download-artifact`) has no equivalent guard. A defensive fix would add a check after the loop:

```bash
args=()
for digest in *; do args+=("docker/docker-agent-sbx-templates@sha256:${digest}"); done
if [[ ${#args[@]} -eq 0 ]]; then
  echo "ERROR: no template digests found" >&2
  exit 1
fi
docker buildx imagetools create "${tags[@]}" "${args[@]}"
```

Alternatively, `shopt -s nullglob` before the loop would make the glob expand to nothing (empty `args`), and the subsequent length check would catch it cleanly.

docker-agent (commented):

The changes correctly mirror the existing multi-arch image publish pattern and introduce a new CI job (`merge-and-push-template`) to build and push sandbox templates. One notable issue was found: the manifest-list assembly step uses a bare bash glob without a guard against an empty digest directory.

aheritier (approved):

PR #3456, opened 2026-07-03
fix(researcher): raise max_tokens to 16000 and add HTML artifact guidance

derekmisler:
## What

Two changes to `.agents/AP/researcher.yaml`:

1. **`max_tokens: 8000` → `16000`** — the 8000 ceiling was the root cause of repeated `unexpected end of JSON input` failures. HTML investigation reports are 15–25 KB (~6,000–8,000 output tokens); at 8000 tokens the LLM hit the limit mid-generation, producing a truncated JSON tool-call argument. Every `write_file`/`shell` call with a ~20 KB payload failed. Session cb3cd11f shows 10+ identical 8000-token failures before the agent gave up.

2. **Add `## Artifact Format` instruction section** — the YAML was missing the HTML artifact format guidance that the deployed system prompt carries. Added: HTML-first format with Tailwind, `upload_artifact` as the preferred path (filesystem-based, avoids base64), a chunked Python write pattern (open + append in sections) to avoid hitting the token ceiling inside a single large tool call, and explicit guidance to split rather than retry on large-payload failures.

## Why these two together

Raising `max_tokens` is the immediate fix; the prompt section closes the YAML/system-prompt drift and prevents the retry loop even if a future token ceiling is hit.

## Evidence

Investigation report: artifact `94713bfa-062a-41a6-9cc6-1e4996871208` — every failure in the transcript has exactly 8000 output tokens and invalid JSON; every success has low token usage. Correlation is 100%.

## Testing

YAML validated as well-formed. No Go code changes; no build/test required.

### DIFF
--- a/.agents/AP/researcher.yaml
+++ b/.agents/AP/researcher.yaml
@@ -1,0 +1,107 @@
+models:
+  default:
+    provider: anthropic
+    model: claude-sonnet-4-6
+    max_tokens: 16000  # raised from 8000 — HTML artifacts run 6k-8k tokens; 8k caused truncated tool-call JSON
+
+agents:
+  root:
+    model: default
+    description: >
+      Research and investigation specialist. Investigates bugs, traces failure
+      patterns across logs and code, and delivers findings as self-contained
+      HTML reports saved as platform artifacts.
+    instruction: |
+      You are a research and investigation agent. Your job is to thoroughly
+      investigate a problem, gather evidence, and deliver a clear, well-structured
+      report as a platform artifact.
+
+      ## Working Principles
+
+      - Read before you conclude. Use tools to gather real evidence; never
+        assert facts you have not verified this session.
+      - Be systematic: form a hypothesis, find evidence that confirms or
+        refutes it, and state your confidence level.
+      - Prefer primary sources (code, logs, config) over inference.
+      - Call independent tools concurrently when the calls are not dependent.
+      - Use the think tool to reason through complex chains of evidence.
+
+      ## Investigation Workflow
+
+      1. **Scope**: Understand exactly what is being investigated. Read the
+         task carefully; identify the key question(s) to answer.
+      2. **Gather**: Use shell, filesystem, and fetch tools to collect
+         evidence — logs, source code, configs, error messages.
+      3. **Analyse**: Use the think tool to reason through the evidence.
+         Identify root causes, supporting data points, and confidence level.
+      4. **Report**: Write a clear, structured investigation report and save
+         it as a platform artifact (see Artifact Format below).
+
+      ## Artifact Format — Render as HTML, not Markdown
+
+      Always save investigation reports as self-contained HTML files with
+      Tailwind CSS styling. Never save plain Markdown for final reports.
+
+      **Preferred tool: `upload_artifact`** — it reads a file from the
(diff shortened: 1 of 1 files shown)


.agents/AP/researcher.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+      # Step 3: close the document
+      python3 - <<'EOF'
+      with open('/tmp/report.html', 'a') as f:
+          f.write("</body></html>\n")
+      EOF
+
+      # Step 4: upload as a platform artifact
+      # (upload_artifact is injected by the platform when shell/filesystem toolsets are active)
+      upload_artifact(
```
docker-agent, line 82: **[MEDIUM] Ambiguous `upload_artifact` call inside mixed shell/Python code fence**

The new `## Artifact Format` block labels the code fence as `` ```python `` but mixes two different invocation styles: shell heredoc commands (`python3 - <<'EOF'`) for file-write steps, then a bare `upload_artifact(...)` call for the upload step. An LLM agent following these instructions literally may attempt to run `upload_artifact(...)` as a shell command or embed it inside a Python script — both of which fail. The inline comment `# (upload_artifact is injected by the platform …)` does hint at the distinction, but it is easy to miss given the surrounding syntax.

**Suggested fix**: Move `upload_artifact(...)` outside the code fence and present it as an explicit tool-call step, e.g.:

> After writing the file, call the `upload_artifact` tool directly (it is a platform tool, not a shell command):
>
> ```
> upload_artifact(path="/tmp/report.html", name="...", mime_type="text/html")
> ```

.agents/AP/researcher.yaml
```diff
@@ -0,0 +1,107 @@
...
+          path="/tmp/report.html",
+          name="investigation-<topic>-<YYYY-MM-DD>.html",
+          mime_type="text/html"
+      )
+      ```
+
+      If a tool call with large content fails with a JSON parse error, split
+      the content across multiple smaller `python3 -c` or append calls rather
```
docker-agent, line 90: **[LOW] Retry advice references `python3 -c` but the demonstrated pattern uses heredoc**

The fallback guidance says "split the content across multiple smaller `python3 -c` or append calls" (line 90), while the primary chunked pattern throughout this section uses `python3 - <<'EOF'` heredoc syntax. The `-c` flag takes the script as a command-line argument, making it unsuitable for multi-line HTML containing quotes and special characters — exactly the content being discussed. An agent that follows the retry advice literally with `python3 -c` and a large HTML chunk would encounter quoting errors worse than the original truncation problem.

**Suggested fix**: Replace `python3 -c` with `python3 - <<'EOF'` in the retry sentence, e.g.:

> "…split the content across multiple smaller heredoc (`python3 - <<'EOF'`) or append calls rather than retrying with the same payload."

docker-agent (commented):

derekmisler:
Closing — investigation shows this file has no consumer. The docker-agent runtime only reads `.agents/skills/`; `.agents/AP/` has no meaning to any code in docker/docker-agent or docker/agentic-platform. The researcher agent (id d729b3fa-1ad0-4b22-87f7-8ef6eee5aded) is a user-owned agent whose config lives in the platform DB, not in this repo. The real fix is a direct API/DB update to that agent record, not a repo commit.

PR #3457, opened 2026-07-03
refactor: share git-branch detection between the lean and full TUIs

rumpl:
## What

Two related dedup steps for the two TUIs, each a behavior-preserving commit.

### 1. Share git-branch detection (`pkg/gitbranch`)

Both TUIs detected the current git branch independently — the lean TUI read `.git` metadata directly, the full-TUI sidebar shelled out to `git rev-parse --abbrev-ref HEAD`. Consolidated onto a new `pkg/gitbranch.Current(dir)` using the subprocess-free implementation.

- No dependency on a `git` binary; no subprocess spawn; preserves detached-HEAD info (short commit).
- Sidebar drops `os/exec` and the 2s-timeout/`context.Context` plumbing (`formatWorkingDirectory`/`getCurrentWorkingDirectory` no longer take a context).
- **Behavior note:** in the sidebar a detached HEAD now shows a short commit instead of the literal `HEAD`, matching the lean TUI.

### 2. Reuse `pathx.ShortenHome` for home-path display

The sidebar and the session browser each open-coded the "replace `$HOME` prefix with `~`" logic. Both now call the shared `pkg/path.ShortenHome`:

- Sidebar `formatWorkingDirectory`: drops the inline prefix replacement (and its partial-component-match bug, e.g. `/home/bob2`) and the `pkg/paths` dependency.
- Session browser: removes the local `abbreviateHome` helper; its sibling-directory edge case is already covered by `pkg/path`'s `TestIsWithin` / `TestShortenHomeDir`.
- The lean TUI's own `shortenPath` (introduced earlier in this branch) was likewise dropped in favour of `pathx.ShortenHome`.

### Validation

- `go build ./...`, `go vet`, `go test` — pass
- `golangci-lint run` (v2.12.2, repo config) on `pkg/leantui`, `pkg/gitbranch`, `pkg/tui/components/sidebar`, `pkg/tui/dialog` — 0 issues

Rebased onto `main` after #3455 merged.

### DIFF
--- a/pkg/gitbranch/gitbranch.go
+++ b/pkg/gitbranch/gitbranch.go
@@ -1,4 +1,6 @@
-package leantui
+// Package gitbranch resolves the current branch of a git working directory by
+// reading .git metadata directly, without shelling out to the git binary.
+package gitbranch
 
 import (
 	"os"
@@ -6,10 +8,14 @@ import (
 	"strings"
 )
 
-// gitBranch returns the current branch name (or short commit for a detached
+// Current returns the current branch name (or the short commit for a detached
 // HEAD) for the repository containing dir, walking up parent directories until
-// a .git entry is found. It returns "" when dir is not inside a repository.
-func gitBranch(dir string) string {
+// a .git entry is found. It returns "" when dir is empty or not inside a
+// repository.
+func Current(dir string) string {
+	if dir == "" {
+		return ""
+	}
 	d := dir
 	for {
 		gitPath := filepath.Join(d, ".git")
@@ -60,19 +66,4 @@ func readHead(headPath string) string {
 		return line[:7] // detached HEAD
 	}
 	return ""
-}
-
-// shortenPath replaces the user's home directory prefix with "~".
-func shortenPath(dir string) string {
-	home, err := os.UserHomeDir()
-	if err != nil || home == "" {
-		return dir
-	}
-	if dir == home {
-		return "~"
-	}
-	if rest, ok := strings.CutPrefix(dir, home+string(filepath.Separator)); ok {
-		return "~" + string(filepath.Separator) + rest
-	}
-	return dir
 }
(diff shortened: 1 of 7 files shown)


aheritier:

docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, well-structured refactoring PR. The consolidation of git-branch detection into `pkg/gitbranch` and the replacement of bespoke home-path abbreviation functions with `pathx.ShortenHome` both look correct.

**Key observations:**

- **`pkg/gitbranch.Current`** — The new implementation reads `.git` metadata directly without shelling out, which is faster and more reliable. The empty-`dir` guard (`if dir == "" { return "" }`) is a nice improvement over the previous version. ✅
- **Detached-HEAD behavior change** — The old sidebar used `git rev-parse --abbrev-ref HEAD` (which returns the literal string `"HEAD"` for detached HEAD), while the new shared implementation returns a 7-char short SHA (e.g., `a1b2c3d`). This is documented in the PR description as an intentional improvement and matches the lean TUI behavior. ✅
- **`pathx.ShortenHome` replacement** — The partial-path-component guard (preventing `/home/bob2` from matching `/home/bob`) is preserved via `filepath.Rel`-based `IsWithin` logic in `pkg/path`. ✅
- **Context removal** — Dropping the `context.Context` parameter from `formatWorkingDirectory`/`getCurrentWorkingDirectory` is safe since the new code no longer shells out. ✅
- **`pathx.ShortenHome` syscall frequency** — The new code calls `os.UserHomeDir()` on each invocation instead of using the old cached home dir. This is not a hot path, so the impact is negligible. ✅

No bugs were found. The changes are behavior-preserving where claimed, and the single documented behavioral difference (detached-HEAD display) is an improvement.

Sayt-0 (approved):

PR #3461, opened 2026-07-03
feat: add shell background job recall

rumpl:
### Shell background job recall
- add `recall: true` shell toolset config that exposes a `recall` parameter only on the recall-enabled background-job schema
- send background-job completion output back through runtime recall, steering active runs or waking idle sessions through the session manager
- document shell recall and add an example config

### Unified `tools.Runtime` handle
Reviewing the recall plumbing surfaced a design smell: tool→runtime communication went through two context keys (`ToolOutputEmitter`, `ToolRecallEmitter`), each with its own `With*`/`FromContext`/`Emit*` boilerplate — invisible in the handler signature and needing ad-hoc lifetime tricks (`context.WithoutCancel` + closure capture) for background work. Both are replaced with a single capability handle passed explicitly to every handler:

- `tools.ToolHandler` is now `func(ctx, ToolCall, rt Runtime)`; runtime-managed handlers (`toolexec.ToolHandler`) grow the same parameter, unifying the two tool APIs on one capability surface
- `Runtime` methods (`EmitOutput`, `Recall`, `Supports`) take ctx per call and stay valid after the handler returns, so background jobs hold the handle instead of a captured closure — no contexts stored in structs
- `NopRuntime` serves hosts without an agent loop (slash commands, prompt expansion, tests); `Supports()` lets tools fail fast before starting background work
- `NewHandler` keeps its signature for the ~68 tools that don't talk back; `NewRuntimeHandler` is the variant for those that do (shell streaming/recall)
- codemode forwards the handle into script-invoked tools, preserving the nested-capability inheritance the ctx transport gave implicitly

New capabilities (elicitation, progress, sampling) now land as one interface method + one nop, instead of five declarations and a context key.

## Validation
- `task build`
- `task test`
- `task lint`

Note: full `task test` fails on the unrelated live provider test `TestOpenAIAliasProvider_LiveAPI/moonshot` (provider 404, reproduces on clean tree).

### DIFF
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -226,7 +226,13 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 	slog.DebugContext(ctx, "Starting runtime stream", "agent", r.currentAgentName(), "session_id", sess.ID)
 	events := make(chan Event, defaultEventChannelCapacity)
 
-	go r.runStreamLoop(ctx, sess, events)
+	go func() {
+		if !sess.IsSubSession() {
+			r.activeRootStreams.Add(1)
+			defer r.activeRootStreams.Add(-1)
+		}
+		r.runStreamLoop(ctx, sess, events)
+	}()
 	return r.observe(ctx, sess, events)
 }
 
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -225,10 +225,13 @@ func (r *LocalRuntime) finalizeEventChannel(ctx context.Context, sess *session.S
 func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-chan Event {
 	slog.DebugContext(ctx, "Starting runtime stream", "agent", r.currentAgentName(), "session_id", sess.ID)
 	events := make(chan Event, defaultEventChannelCapacity)
+	rootStream := !sess.IsSubSession()
+	if rootStream {
+		r.activeRootStreams.Add(1)
+	}
 
 	go func() {
-		if !sess.IsSubSession() {
-			r.activeRootStreams.Add(1)
+		if rootStream {
 			defer r.activeRootStreams.Add(-1)
 		}
 		r.runStreamLoop(ctx, sess, events)
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -179,6 +179,24 @@ func (sm *SessionManager) RegisterFollowUpInjector(sessionID string, fn FollowUp
 	sm.followUpInjectors.Store(sessionID, fn)
 }
 
+type recallHandlerSetter interface {
+	SetRecallHandler(handler runtime.RecallHandler)
+}
+
+func (sm *SessionManager) registerRecallHandler(sessionID string, rt runtime.Runtime) {
+	setter, ok := rt.(recallHandlerSetter)
+	if !ok {
+		return
+	}
+	setter.SetRecallHandler(func(ctx context.Context, msg runtime.QueuedMessage) bool {
+		if err := sm.recallSession(ctx, sessionID, msg); err != nil {
+			slog.WarnContext(ctx, "Failed to handle tool recall", "session_id", sessionID, "error", err)
+			return false
+		}
+		return true
+	})
+}
+
 // StreamEvents replays and tails the events buffered for sessionID, calling
 // send for each one with its sequence number. When since is non-nil only
 // events newer than *since are replayed before tailing (see [eventLog.stream]
@@ -210,6 +228,7 @@ func (sm *SessionManager) AttachRuntime(ctx context.Context, sessionID string, r
 		cancel:  cancel,
 		session: sess,
 	})
+	sm.registerRecallHandler(sessionID, rt)
 	sm.markReady()
 }
 
@@ -613,6 +632,7 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 			titleGen: titleGen,
 		}
 		sm.runtimeSessions.Store(sessionID, runtimeSession)
+		sm.registerRecallHandler(sessionID, rt)
 		sm.markReady()
 	} else {
 		titleGen = runtimeSession.titleGen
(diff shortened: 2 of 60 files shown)


pkg/server/session_manager.go
```diff
@@ -811,6 +831,56 @@ func (sm *SessionManager) FollowUpSession(ctx context.Context, sessionID string,
...
+		sm.mux.Unlock()
+		rt.streaming.Unlock()
+		return err
+	}
+	rt.session = sess
+	sm.mux.Unlock()
+
+	go func() {
+		defer rt.streaming.Unlock()
+		stream := rt.runtime.RunStream(context.WithoutCancel(ctx), sess)
```
docker-agent: **[MEDIUM] Recall goroutine spawned with non-cancellable context — cannot be stopped by `DeleteSession`**

The goroutine spawned in `recallSession` calls `rt.runtime.RunStream(context.WithoutCancel(ctx), sess)` where `ctx` is already a non-cancellable `recallCtx` from the tool dispatcher. This means the goroutine is not connected to the session-level cancellation context.

When `DeleteSession` fires it calls `sessionRuntime.cancel()`, which cancels the session context — but the recall goroutine's `RunStream` keeps running because its context's `Done()` channel is never closed. The LLM API calls and tool executions continue for a session that has already been logically deleted.

Compare to the `RunSession` path (line 619), which derives `streamCtx` from the HTTP request context via `context.WithCancel(ctx)`, giving it a real cancellation path when the client disconnects or the session is deleted.

**Suggested fix:** derive the goroutine's context from the session-level done channel so `DeleteSession` can interrupt it:
```go
sessCtx, sessCancel := context.WithCancelCause(context.WithoutCancel(ctx))
go func() {
    defer rt.streaming.Unlock()
    defer sessCancel(nil)
    // Use sessCtx, and wire sessCancel to rt.done somewhere reachable
    stream := rt.runtime.RunStream(sessCtx, sess)
    ...
}()
```
Alternatively, store the cancel function on `rt` so `DeleteSession` can call it.

pkg/server/session_manager.go
```diff
@@ -811,6 +831,56 @@ func (sm *SessionManager) FollowUpSession(ctx context.Context, sessionID string,
...
+		for event := range stream {
+			if pe, ok := sm.eventLogs.Load(sessionID); ok {
+				pe.log.append(event)
+			}
+		}
+		if err := sm.sessionStore.UpdateSession(context.WithoutCancel(ctx), sess); err != nil {
```
docker-agent, line 894: **[MEDIUM] Recall goroutine may `UpdateSession` on an already-deleted session, potentially resurrecting it**

After `RunStream` completes, the recall goroutine calls `sm.sessionStore.UpdateSession(context.WithoutCancel(ctx), sess)`. Because the goroutine's context is non-cancellable (see sibling comment), a `DeleteSession` call can run to completion while the recall `RunStream` is still executing. The sequence:

1. `DeleteSession` removes the session from the store (and fires `cancel()`).
2. Recall `RunStream` finishes responding.
3. `UpdateSession(sess)` writes the post-stream session data back to the store.

For session stores with upsert semantics (insert-or-update), step 3 silently recreates the deleted session. The error path only logs a warning and is otherwise ignored.

The `RunSession` goroutine has a smaller exposure window because its context is tied to the HTTP request and is typically cancelled before `DeleteSession` completes, but the recall goroutine has no such coupling.

**Suggested fix:** check whether the session was deleted before persisting, e.g.:
```go
if _, stillExists := sm.runtimeSessions.Load(sessionID); stillExists {
    if err := sm.sessionStore.UpdateSession(...); err != nil {
        slog.WarnContext(...)
    }
}
```
or check `sm.deletedSessions` instead.

pkg/runtime/loop.go
```diff
@@ -226,7 +226,13 @@ func (r *LocalRuntime) RunStream(ctx context.Context, sess *session.Session) <-c
 	slog.DebugContext(ctx, "Starting runtime stream", "agent", r.currentAgentName(), "session_id", sess.ID)
 	events := make(chan Event, defaultEventChannelCapacity)
 
-	go r.runStreamLoop(ctx, sess, events)
+	go func() {
+		if !sess.IsSubSession() {
+			r.activeRootStreams.Add(1)
```
docker-agent: **[LOW] TOCTOU in `activeRootStreams` can trigger a duplicate `RunStream` on the `AttachRuntime`/TUI path**

`activeRootStreams` is incremented *inside* the goroutine body, not before `RunStream` returns the channel. For sessions managed via `RunSession`, the session-manager's `streaming` mutex is locked synchronously *before* the goroutine starts — so even when `activeRootStreams` is transiently 0, `recallSession`'s `TryLock` fails and falls back to `Steer`.

However, the `AttachRuntime` path (used by the TUI/embedded driver) stores the runtime without holding `streaming` at all. If a background job completes during the brief window between the TUI calling `RunStream` and the goroutine inside `RunStream` incrementing `activeRootStreams`, the recall path sees `activeRootStreams == 0` **and** `streaming.TryLock()` succeeds, causing `recallSession` to spawn a second concurrent `RunStream` on the same session.

Two concurrent `RunStream` loops on the same session will produce interleaved events and likely corrupt the session's conversation history.

**Suggested fix:** increment `activeRootStreams` *before* spawning the goroutine (and use the matching decrement in a `defer` set up before the spawn), or move the counter increment to just after the goroutine is created but before the channel is returned. A simpler alternative is to move the responsibility of holding `streaming` into `recallSession` only when it knows the counter is unreliable (i.e., check both guards before concluding the session is idle).

docker-agent (commented):

**PR**: feat: add shell background job recall
**Findings**: 2 MEDIUM (CONFIRMED/LIKELY) + 1 LOW (LIKELY)

The new recall pipeline is well-structured overall, but the session lifecycle when a recalled `RunStream` overlaps with `DeleteSession` has a gap: the recall goroutine uses a non-cancellable context and has no guard against persisting to an already-deleted session.

pkg/server/session_manager.go
rumpl: Fixed in `aea71cd6d`. The recall goroutine now checks `runtimeSessions` under `sm.mux` after the stream drains and returns without persisting if the session was deleted. I also added `TestRecallSession_DeleteCancelsRecalledRunAndDoesNotResurrect` to cover this path.

pkg/runtime/loop.go
rumpl: Fixed in `aea71cd6d`: `RunStream` now increments `activeRootStreams` synchronously before returning the event channel, with the decrement still deferred in the goroutine. Added `TestRunStreamIncrementsActiveRootStreamsBeforeReturning` for the race window.

pkg/server/session_manager.go
rumpl: Thanks, good catch. Fixed in `aea71cd6d`: idle recall streams now use a cancellable `runCtx`, and `rt.cancel` is updated to that cancel func before the goroutine starts so `DeleteSession` interrupts a recall-started `RunStream`.

pkg/runtime/loop.go
docker-agent: Perfect! The synchronous increment at line 230 closes the race window cleanly — `activeRootStreams` now reflects the active stream *before* `RunStream` returns control to the caller, so `recallSession` can no longer observe the transient zero state and mistakenly spawn a duplicate loop.

The test confirms the counter is visible immediately after `RunStream` returns, covering exactly the scenario the original finding described. Nice work!

dgageot (dismissed):

aheritier:
Hi @rumpl 👋

This PR has merge conflicts with `main` and has been converted to draft until they are resolved.

**To fix:**
1. Rebase onto the latest `main`: `git fetch origin && git rebase origin/main`
2. Resolve any conflicts, then `git push --force-with-lease`
3. Mark the PR ready for review once the conflicts are cleared — the `status/needs-rebase` label will be removed automatically on the next triage run.

Let me know if you need help!

dgageot (approved):

PR #3463, opened 2026-07-03
feat(shell): add wait_background_job tool

### LINKED ISSUE #3462: Shell background job tool needs a companion `wait_background_job` tool to wait on completion
## Problem

The shell toolset can start background jobs via `run_background_job` and return a job ID immediately. However, once a job is running there is no way to *synchronously wait* for it to finish and collect its result. The only option is to poll with `view_background_job` in a loop, which is fragile and wastes model turns.

## Gap

- `run_background_job` starts a job → returns job ID
- `list_background_jobs` / `view_background_job` → non-blocking status snapshots
- `stop_background_job` → terminates a running job
- **No tool exists to block until a job completes and retrieve its exit code + output**

This forces agents to poll `view_background_job` repeatedly, which is wasteful and can trip the loop detector.

Relevant source: `pkg/tools/builtin/shell/shell.go`

## Proposed Solution

Add a `wait_background_job` tool that:
- Takes a `job_id` (same identifier returned by `run_background_job`)
- Blocks until the job completes (or an optional `timeout` expires)
- Returns the job's exit code/status and captured stdout/stderr
- Handles edge cases: unknown job ID, job already completed (returns cached result, not an error), job still running when timeout fires

## Acceptance Criteria

- [ ] New `wait_background_job` tool registered in the shell toolset
- [ ] `done chan struct{}` on each `backgroundJob`, closed when the job finishes (regardless of whether it completed, failed, or was stopped)
- [ ] Handler blocks on the done channel and returns full job status + output
- [ ] Optional `timeout` parameter (default 60 s); returns partial output with a timeout notice when exceeded
- [ ] Tests covering completed, failed, already-completed, unknown-id, timeout, and context-cancelled cases
- [ ] Docs updated (currently says "five tools" — should say six)

simonferquel-clanker:

Adds a `wait_background_job` tool to the shell toolset that blocks until a background job finishes and returns its exit code and captured output.

Previously, agents had to poll `view_background_job` in a loop to know when a background job completed. The new tool blocks on a `done chan struct{}` closed by `monitorJob`, eliminating polling entirely. An optional `timeout` parameter (default 60 s) caps the wait; the job keeps running if the timeout fires and the tool returns the current output with a notice.

**Key changes:**
- `done chan struct{}` on `backgroundJob`, closed by `monitorJob` via a deferred close (LIFO: `Unlock` before `close`, so callers see fully written state)
- Extract `renderBackgroundJob()` helper; refactor `ViewBackgroundJob` to use it; fix pre-existing issue: stopped jobs no longer show `Exit Code: 0`
- `WaitBackgroundJob` handler with `ReadOnlyHint: true`; selects on `job.done / timer.C / ctx.Done`
- 7 tests: completed, failed-exit, already-completed, unknown-id, timeout, context-cancelled, stopped-then-wait
- Docs updated: six tools, `wait_background_job` parameters table

Closes #3462

### DIFF
--- a/docs/tools/shell/index.md
+++ b/docs/tools/shell/index.md
@@ -89,7 +89,7 @@ Notes and limitations:
 
 ## Available Tools
 
-The shell toolset exposes five tools:
+The shell toolset exposes six tools:
 
 | Tool Name              | Description                                                                                    |
 | ---------------------- | ---------------------------------------------------------------------------------------------- |
@@ -98,6 +98,7 @@ The shell toolset exposes five tools:
 | `list_background_jobs` | List all background jobs with their status, runtime, and metadata.                             |
 | `view_background_job`  | View the buffered output and status of a specific background job by ID.                        |
 | `stop_background_job`  | Stop a running background job. Child processes are terminated too.                             |
+| `wait_background_job`  | Block until a job finishes and return its exit code and output. Safe on already-finished jobs. |
 
 Background job output is captured up to 10 MB per job. All background jobs are automatically terminated when the agent session ends.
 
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `wait_background_job` implementation is correct and well-structured. Key design points verified:

- **Defer ordering is correct**: `defer close(job.done)` is registered first, `defer job.outputMu.Unlock()` second — LIFO ensures the mutex is released before the channel is closed, so `WaitBackgroundJob` readers always see fully-committed state.
- **Stopped-job path**: `monitorJob` acquires the lock, detects `statusStopped`, returns early; both defers still fire in correct order. `statusStopped` is present in `statusStrings` so the map lookup in `WaitBackgroundJob` is always valid.
- **No double-close risk**: `done` is only closed inside `monitorJob`, which runs exactly once per job.
- **Timeout path**: `renderBackgroundJob` is safely callable while the job is running (uses `outputMu.RLock`).
- **7 tests** cover: completed, failed-exit, already-completed, unknown-id, timeout, context-cancelled, and stopped-then-wait.

No bugs found in the changes introduced by this PR.

Sayt-0 (approved):

PR #3464, opened 2026-07-03
feat(hcl): render file() as a template when given variables

rumpl:
## What

Extends the HCL `file()` function with an optional second argument that renders the file as an HCL template — `templatefile` semantics without adding a new function:

```hcl
agent "reviewer" {
  instruction = file("prompts/reviewer.md", {
    language   = "Go"
    strictness = "high"
  })
}
```

Each key of the vars object becomes a `${key}` variable inside the file. Full HCL template syntax is supported, including `%{ for }` / `%{ if }` directives.

## Behavior

- `file(path)` is unchanged: contents are returned verbatim, so literal `${...}` runtime snippets (e.g. `${shell({cmd: "..."})}`) pass through untouched.
- `file(path, vars)` renders the file as a template; `vars` must be an object/map.
- Referencing a variable not present in `vars` is an error.
- Templates expose no functions, so a rendered file cannot recursively call `file()`.
- Reads stay sandboxed to the config directory via the existing `os.OpenRoot` path.

## Changes

- `pkg/config/hcl/hcl_file.go`: optional `vars` param + template rendering
- `pkg/config/hcl/hcl_file_test.go`: tests for rendering, directives, missing variables, non-object vars, no-functions-in-templates, and raw passthrough
- `docs/configuration/hcl/index.md`: new "Loading Files with `file()`" section; fixed the stale "no docker-agent-specific HCL functions" bullet
- `examples/instructions_from_file.{hcl,md}`: demonstrate the template form

## Validation

- [x] `task build`
- [x] `task test` (only pre-existing live-API failure `TestOpenAIAliasProvider_LiveAPI/moonshot`, unrelated)
- [x] `task lint`

### DIFF
--- a/pkg/config/hcl/hcl_file.go
+++ b/pkg/config/hcl/hcl_file.go
@@ -51,7 +51,7 @@ func fileFunction(baseDir string) function.Function {
 // cannot recursively call file().
 func renderTemplate(data []byte, filename string, vars cty.Value) (cty.Value, error) {
 	t := vars.Type()
-	if vars.IsNull() || (!t.IsObjectType() && !t.IsMapType()) {
+	if vars.IsNull() || !vars.IsKnown() || (!t.IsObjectType() && !t.IsMapType()) {
 		return cty.NilVal, errors.New(`template variables must be an object, e.g. { name = "value" }`)
 	}
 
--- a/pkg/config/hcl/hcl_file.go
+++ b/pkg/config/hcl/hcl_file.go
@@ -6,27 +6,75 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/convert"
 	"github.com/zclconf/go-cty/cty/function"
 )
 
+// fileFunction implements `file(path)` and `file(path, vars)`. With a single
+// argument the file contents are returned verbatim. With a vars object, the
+// file is rendered as an HCL template with each key exposed as a variable.
 func fileFunction(baseDir string) function.Function {
 	return function.New(&function.Spec{
 		Params: []function.Parameter{{
 			Name: "path",
 			Type: cty.String,
 		}},
+		VarParam: &function.Parameter{
+			Name:             "vars",
+			Type:             cty.DynamicPseudoType,
+			AllowDynamicType: true,
+		},
 		Type: function.StaticReturnType(cty.String),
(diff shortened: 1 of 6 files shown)


pkg/config/hcl/hcl_file.go
```diff
@@ -6,29 +6,77 @@ import (
...
+		return cty.NilVal, errors.New(`template variables must be an object, e.g. { name = "value" }`)
+	}
+
+	expr, diags := hclsyntax.ParseTemplate(data, filename, hcl.InitialPos)
+	if diags.HasErrors() {
+		return cty.NilVal, fmt.Errorf("parsing template %q: %s", filename, diags.Error())
+	}
+
+	evalCtx := &hcl.EvalContext{Variables: map[string]cty.Value{}}
+	for it := vars.ElementIterator(); it.Next(); {
```
docker-agent, line 64: **[LIKELY / MEDIUM] Missing `IsKnown()` guard before `ElementIterator()` — may panic on unknown cty values**

`renderTemplate` checks `vars.IsNull()` and the type (`IsObjectType` / `IsMapType`) before calling `vars.ElementIterator()`, but does not check `vars.IsKnown()`.

In go-cty, calling `ElementIterator()` on an *unknown* value with a concrete object or map type causes a runtime panic. The existing type guard handles the common case (an unknown with `cty.DynamicPseudoType` is neither an object nor a map, so the guard fires first), but a typed unknown — e.g. `cty.UnknownVal(cty.Map(cty.String))` — has type `cty.Map`, passes the guard, and reaches `ElementIterator()` where it panics.

**Suggested fix:** add `!vars.IsKnown()` to the guard condition:

```go
if vars.IsNull() || !vars.IsKnown() || (!t.IsObjectType() && !t.IsMapType()) {
    return cty.NilVal, errors.New(`template variables must be an object, e.g. { name = "value" }`)
}
```

This is a one-line defensive addition that eliminates the panic path entirely.

docker-agent (commented):

pkg/config/hcl/hcl_file.go
rumpl: Fixed in fbc34ed6e — added `!vars.IsKnown()` to the guard and a unit test exercising `renderTemplate` with `cty.UnknownVal(cty.Map(cty.String))` (not reachable from literal HCL config syntax, hence the direct call).

docker-agent: Thanks for the quick fix! The guard placement and unit test look perfect — explicitly exercising the typed-unknown case ensures this path stays covered even though it's not reachable from HCL syntax.

Sayt-0 (approved):

PR #3466, opened 2026-07-03
fix: remove Anthropic Files API usage

rumpl:
## Summary
- remove Anthropic Files API upload/delete manager and tests
- stop adding the files-api beta header or routing legacy file parts to Anthropic file IDs
- keep legacy message part structs for stored-session compatibility while documenting that new attachments should use documents

## Tests
- `go test ./pkg/model/provider/anthropic ./pkg/chat`
- `task lint`
- `task build`


pkg/model/provider/anthropic/client.go
```diff
@@ -580,36 +559,25 @@ func (c *Client) convertUserMultiContent(ctx context.Context, parts []chat.Messa
...
 					}))
 				}
 			} else if strings.HasPrefix(part.ImageURL.URL, "http://") || strings.HasPrefix(part.ImageURL.URL, "https://") {
-				// URL-based images
 				contentBlocks = append(contentBlocks, anthropic.NewImageBlock(anthropic.URLImageSourceParam{
 					URL: part.ImageURL.URL,
 				}))
 			}
 
 		case chat.MessagePartTypeFile:
```
docker-agent, line 576: **[LOW] Legacy FileID-only file parts silently dropped with no warning in non-beta path**

When a stored session contains a `MessagePartTypeFile` part where `part.File == nil` (i.e., only a `FileID` string is present, no local file object), this case reaches the `convertUserMultiContent` non-beta path and silently skips the part with no log entry. The `part.File != nil` branch correctly logs a deprecation warning, but the `FileID`-only case goes unlogged.

Consider also warning when `part.File == nil` (e.g. checking `part.FileID != ""`):
```go
case chat.MessagePartTypeFile:
    if part.File != nil {
        slog.WarnContext(ctx, "anthropic: dropping legacy file attachment; use document attachments instead", "path", part.File.Path)
    } else if part.FileID != "" {
        slog.WarnContext(ctx, "anthropic: dropping legacy Files-API file ID; use document attachments instead", "file_id", part.FileID)
    }
```

pkg/model/provider/anthropic/beta_converter.go
```diff
@@ -294,41 +288,7 @@ func (c *Client) convertBetaUserMultiContent(ctx context.Context, parts []chat.M
 			}
 
 		case chat.MessagePartTypeFile:
```
docker-agent, line 290: **[LOW] Beta path silently drops legacy file parts with no log warning (inconsistent with non-beta path)**

The non-beta path (`convertUserMultiContent`) logs a `slog.WarnContext` deprecation message when it encounters a `MessagePartTypeFile` part, but the beta path here uses a bare `continue` with no log output. A session that reaches `createBetaStream` (e.g. because it uses structured output or extended thinking) with legacy file parts will silently lose those attachments with no operator-visible signal.

For consistency, add the same warning here:
```go
case chat.MessagePartTypeFile:
    if part.File != nil {
        slog.WarnContext(ctx, "anthropic: dropping legacy file attachment; use document attachments instead", "path", part.File.Path)
    }
    continue
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly removes the Anthropic Files API (FileManager, upload/delete logic, the files-api beta header, and the routing guard). Legacy `MessagePartTypeFile` structs are retained for stored-session deserialization compatibility — a good design choice. Two minor observability gaps were found in the new `MessagePartTypeFile` handling paths.

Sayt-0 (approved):

PR #3469, opened 2026-07-04
fix: compaction data loss on empty summary + honest compaction outcome reporting

dgageot:
Session compaction had two independent bugs that together made it both silently destructive and untrustworthy to observe.

The first bug was a data-loss path in `compactor.RunLLM`. When the summarization model returned an empty response — due to a rate limit, token cap, or a flaky local model — the code called `GetLastAssistantMessageContent()` on a sub-session that had been seeded with the conversation being summarized. That meant the *conversation's own last assistant message* was used as the summary, silently replacing the compacted history with stale content. The fix counts only assistant content appended after the seed; an empty response now takes the no-op path instead. A regression test reproduces the bug when the fix is reverted. The same treatment is applied to whitespace-only model output, which is trimmed at extraction time so it too becomes a skip rather than a corrupt write.

The second bug was in outcome reporting. `doCompact` emitted an identical "completed" event regardless of whether a summary was applied, the compaction no-opped (e.g. hook-vetoed), or it failed. The TUI announced "Session compacted successfully." even immediately after an error. The `SessionCompactionEvent` now carries an `Outcome` field with shared `CompactionOutcome*` constants. The chat TUI toasts success only for an applied outcome, info for skipped, and stays silent for failed (the `ErrorEvent` already surfaces those). Hook-vetoed compactions synthesized by the app are reported as skipped.

The `Outcome` field is `omitempty` on the wire, so consumers receiving events from an older server see empty and preserve legacy behavior. All consumers were checked: chat TUI, leantui, app, and remote client.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -1150,7 +1150,11 @@ func (a *App) CompactSession(ctx context.Context, cancel context.CancelFunc, add
 			a.sendEvent(ctx, event)
 		}
 		if !completed && ctx.Err() == nil {
-			a.sendEvent(ctx, runtime.SessionCompaction(sess.ID, "completed", agentName))
+			// The compaction never emitted its started/completed pair (e.g. a
+			// pre_compact / before_compaction hook vetoed it). Synthesize the
+			// terminal event so the TUI unsticks, but report it as skipped —
+			// nothing was compacted.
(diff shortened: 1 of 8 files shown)


dgageot:
Superseded by #3470 (same commits, opened from the fork instead of a repo-local worktree branch).

PR #3470, opened 2026-07-04
fix: compaction data loss on empty summary + honest compaction outcome reporting

dgageot:
## What

Four commits from a scenario-driven review of the session-compaction code:

1. **fix: compaction no longer replaces history with a stale reply on empty summary** — `compactor.RunLLM` read the summary via `GetLastAssistantMessageContent()` on a sub-session seeded with the conversation being summarized. When the summarization model returned an empty response (rate limit, token cap, flaky local model), the conversation's own last assistant message was applied as the "summary", silently wiping the compacted history. Fixed by only counting assistant content appended after the seed; an empty response is now a no-op. Regression tests included.

2. **fix: report compaction outcome (applied/skipped/failed) on the completed event** — `doCompact` emitted an identical "completed" event whether a summary was applied, the compaction no-opped, or it failed, so the TUI announced "Session compacted successfully." even right after an error. The completed `SessionCompactionEvent` now carries an `Outcome` field (shared `CompactionOutcome*` constants); the TUI only toasts success for applied (empty outcome = backward compat with older servers), info for skipped, silent for failed (the ErrorEvent already surfaced). Hook-vetoed compactions synthesized by the App are reported as skipped.

3. **test: add compaction edge-case scenario tests** — estimator guards (reasoning > output tokens, output-only usage, compaction between anchors), kept-tail boundary snapping (never starts on a tool result), threshold misconfiguration fallback, and documentation of the lossy prior-summary drop under tight summarization budgets.

4. **fix: treat a whitespace-only compaction summary as a no-op** — trimmed at the extraction point so whitespace-only model output takes the skip path.

## Why

Empty summarizer responses were corrupting session history (data loss), and failed/no-op compactions were indistinguishable from successful ones for users and event consumers.

## Backward compatibility

`Outcome` is `omitempty`-additive on the wire; all consumers verified (chat TUI, leantui, app, remote client). Empty outcome from older servers keeps legacy behavior. Every commit builds and tests green individually.

## Testing

- Regression test reproduces the data-loss bug when the fix is reverted
- `task build`, `task test` (full suite), `task lint` all green; `-race` clean on runtime/session/compaction packages

Supersedes #3469.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -1218,7 +1218,11 @@ func (a *App) CompactSession(ctx context.Context, cancel context.CancelFunc, add
 			a.sendEvent(ctx, event)
 		}
 		if !completed && ctx.Err() == nil {
-			a.sendEvent(ctx, runtime.SessionCompaction(sess.ID, "completed", agentName))
+			// The compaction never emitted its started/completed pair (e.g. a
+			// pre_compact / before_compaction hook vetoed it). Synthesize the
+			// terminal event so the TUI unsticks, but report it as skipped —
+			// nothing was compacted.
+			a.sendEvent(ctx, runtime.SessionCompactionCompleted(sess.ID, runtime.CompactionOutcomeSkipped, agentName))
 		}
 	}()
 }
--- a/pkg/runtime/compactor/compactor.go
+++ b/pkg/runtime/compactor/compactor.go
@@ -348,17 +348,19 @@ func firstKeptSessionIndex(sess *session.Session, sessIndices []int, splitIdx in
 	return sessIndices[splitIdx]
 }
 
-// lastAssistantContentAfter returns the content of the last assistant
-// message appended to sess after the first seedLen items, or "" when
-// the run produced none. sess is the throwaway compaction sub-session,
-// owned exclusively by the caller once RunAgent has returned.
+// lastAssistantContentAfter returns the content of the last non-empty
+// assistant message appended to sess after the first seedLen items, or
+// "" when the run produced none. Whitespace-only replies are skipped
+// so a trailing blank token doesn't hide an earlier valid summary.
+// sess is the throwaway compaction sub-session, owned exclusively by
+// the caller once RunAgent has returned.
 func lastAssistantContentAfter(sess *session.Session, seedLen int) string {
 	for i := len(sess.Messages) - 1; i >= seedLen; i-- {
 		item := sess.Messages[i]
 		if item.IsMessage() && item.Message.Message.Role == chat.MessageRoleAssistant {
-			// Whitespace-only output is no summary; trimming keeps it on
-			// the no-op path.
-			return strings.TrimSpace(item.Message.Message.Content)
+			if trimmed := strings.TrimSpace(item.Message.Message.Content); trimmed != "" {
+				return trimmed
+			}
 		}
 	}
 	return ""
--- a/pkg/runtime/compactor/compactor.go
+++ b/pkg/runtime/compactor/compactor.go
@@ -216,13 +216,23 @@ func RunLLM(ctx context.Context, args LLMArgs) (result *Result, err error) {
 		session.WithTitle("Generating summary"),
 		session.WithMessages(toItems(messages)),
 	)
+	seedLen := len(compactionSession.Messages)
 
 	if err := args.RunAgent(ctx, compactionAgent, compactionSession); err != nil {
 		return nil, fmt.Errorf("run compaction agent: %w", err)
 	}
 
-	summary := compactionSession.GetLastAssistantMessageContent()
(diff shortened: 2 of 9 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly fixes both target bugs: the seed-length guard in `lastAssistantContentAfter` prevents stale conversation history from being used as a compaction summary when the LLM returns empty, and the new `Outcome` field on `SessionCompactionEvent` lets the TUI (and event consumers) differentiate applied/skipped/failed outcomes. Backward compatibility is maintained via `omitempty`. The scenario tests are thorough and directly exercise the fixed behaviors. Two minor low-severity observations are noted inline.

pkg/app/app.go
```diff
@@ -1150,7 +1150,11 @@ func (a *App) CompactSession(ctx context.Context, cancel context.CancelFunc, add
 			a.sendEvent(ctx, event)
 		}
 		if !completed && ctx.Err() == nil {
-			a.sendEvent(ctx, runtime.SessionCompaction(sess.ID, "completed", agentName))
+			// The compaction never emitted its started/completed pair (e.g. a
+			// pre_compact / before_compaction hook vetoed it). Synthesize the
+			// terminal event so the TUI unsticks, but report it as skipped —
+			// nothing was compacted.
+			a.sendEvent(ctx, runtime.SessionCompactionCompleted(sess.ID, runtime.CompactionOutcomeSkipped, agentName))
```
docker-agent, line 1225: **[LOW] Synthesized `completed` event emitted without a prior `started` event when hook vetoes compaction**

When a pre-compaction hook vetoes the operation, `doCompact` never fires, meaning no `SessionCompactionStarted` event is emitted. This synthesized `CompactionOutcomeSkipped` completed event is intentional for the TUI (which guards on `Status == "completed"` to unstick spinners). However, any external API or event-stream consumer replaying the event log will see an unpaired `completed` with no preceding `started`, which may confuse replay logic or event-sourced state machines.

Consider documenting in the `SessionCompactionCompleted` godoc or the event schema that a `completed` event can appear without a preceding `started` in the hook-veto case, so SDK consumers can handle it gracefully.

pkg/runtime/compactor/compactor.go
```diff
@@ -337,6 +348,22 @@ func firstKeptSessionIndex(sess *session.Session, sessIndices []int, splitIdx in
...
+// message appended to sess after the first seedLen items, or "" when
+// the run produced none. sess is the throwaway compaction sub-session,
+// owned exclusively by the caller once RunAgent has returned.
+func lastAssistantContentAfter(sess *session.Session, seedLen int) string {
+	for i := len(sess.Messages) - 1; i >= seedLen; i-- {
+		item := sess.Messages[i]
+		if item.IsMessage() && item.Message.Message.Role == chat.MessageRoleAssistant {
+			// Whitespace-only output is no summary; trimming keeps it on
+			// the no-op path.
+			return strings.TrimSpace(item.Message.Message.Content)
```
docker-agent: **[LOW] `lastAssistantContentAfter` returns early on whitespace-only content, skipping any earlier valid assistant reply**

When iterating backward from the last message, the loop returns `strings.TrimSpace(...)` immediately upon finding the first assistant message — even if that message is whitespace-only. A whitespace-only trailing assistant reply (e.g., a reasoning model emitting `\n` as a final token) causes the function to return `""` and take the no-op path, silently discarding any earlier, valid assistant summary produced in the same sub-session.

In practice the summarizer produces exactly one assistant message, so this edge is unlikely to trigger. But if future compaction flows allow tool-calls followed by a final assistant message, a whitespace-only final reply would drop the real summary. A `continue` on empty content (rather than `return`) would be safer:
```go
trimmed := strings.TrimSpace(item.Message.Message.Content)
if trimmed != "" {
    return trimmed
}
// continue searching backward for a non-empty assistant reply
```

pkg/app/app.go
dgageot: Documented in ba4f411e5: the SessionCompactionCompleted godoc now states that a "completed" event may arrive without a preceding "started" when a pre-compaction hook vetoes the operation.

pkg/runtime/compactor/compactor.go
dgageot: Good catch — fixed in ba4f411e5: the loop now continues backward past whitespace-only assistant replies instead of returning early, with unit tests covering the edge case.

pkg/app/app.go
docker-agent: Perfect, thank you for the documentation! The godoc now explicitly calls out the unpaired-event scenario, which will help anyone consuming the event stream understand this edge case.

Sayt-0 (approved):

PR #3471, opened 2026-07-04
chore: standardize agent_name JSON tag in Message struct

Piyush0049:
### What does this PR do?
This PR addresses a piece of technical debt by resolving an outstanding `TODO` in `pkg/session/session.go`. It standardizes the JSON serialization tag for `Message.AgentName` from camelCase (`agentName`) to snake_case (`agent_name`), aligning it with the repository's JSON schema conventions.

### Changes made:
- Updated the JSON struct tag for `AgentName` in `pkg/session/session.go`.
- Removed the legacy `// TODO: rename to agent_name` comment.
- Updated all legacy evaluation fixtures in `examples/eval/evals/` and `examples/golibrary/tool/` to match the new `agent_name` schema.

### Backwards Compatibility
This is a safe change with zero downtime or backwards-incompatibility for end users:
- **SQLite Database:** The `session.Message` struct itself is not marshaled to the database; it maps directly to a dedicated `agent_name` SQL column, so existing local databases are completely unaffected.
- **Legacy JSON Files:** Go's `encoding/json` package falls back to a case-insensitive field match (`AgentName` -> `"agentName"`) if the struct tag (`"agent_name"`) isn't found in older JSON files. This means legacy exported sessions or evaluation files will still load successfully without dropping the agent name. Future exports will correctly emit `"agent_name"`.

### DIFF
--- a/examples/eval/evals/41b179a2-ed19-4ae2-a45d-95775aaa90f7.json
+++ b/examples/eval/evals/41b179a2-ed19-4ae2-a45d-95775aaa90f7.json
@@ -5,7 +5,7 @@
         {
             "message": {
                 "agentFilename": "./agent.yaml",
-                "agentName": "",
+                "agent_name": "",
                 "message": {
                     "role": "user",
                     "content": "How many files in the local folder?",
@@ -16,7 +16,7 @@
         {
             "message": {
                 "agentFilename": "",
-                "agentName": "root",
+                "agent_name": "root",
                 "message": {
                     "role": "assistant",
                     "content": "",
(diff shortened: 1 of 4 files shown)


Sayt-0 (approved):

Sayt-0:
Hello @Piyush0049 your commits must be signed in this repo

Piyush0049:
Done! I've amended the commit with the signature, and force-pushed.

Sayt-0:
Nice ! Thanks for the contribution !

PR #3473, opened 2026-07-04
Split background jobs into separate toolset

rumpl:
## Summary
- move background job tools and process-management code into a new background_jobs toolset/package
- leave the shell toolset exposing only the synchronous shell tool
- update schema, default configs, examples, and docs for background_jobs

## Validation
- task build
- task --force lint
- env -u MOONSHOT_API_KEY task test

### DIFF
--- a/pkg/tools/builtin/backgroundjobs/backgroundjobs.go
+++ b/pkg/tools/builtin/backgroundjobs/backgroundjobs.go
@@ -1,0 +1,607 @@
+// Package backgroundjobs exposes the built-in background_jobs toolset.
+package backgroundjobs
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/docker/docker-agent/pkg/concurrent"
+	"github.com/docker/docker-agent/pkg/config"
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/environment"
+	"github.com/docker/docker-agent/pkg/shellpath"
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+const (
+	// ToolNameRunBackgroundJob starts a shell command asynchronously.
+	ToolNameRunBackgroundJob = "run_background_job"
+	// ToolNameListBackgroundJobs lists known background jobs.
+	ToolNameListBackgroundJobs = "list_background_jobs"
+	// ToolNameViewBackgroundJob renders one background job's status and output.
+	ToolNameViewBackgroundJob = "view_background_job"
+	// ToolNameStopBackgroundJob stops a running background job.
+	ToolNameStopBackgroundJob = "stop_background_job"
+	// ToolNameWaitBackgroundJob waits for a background job to finish.
+	ToolNameWaitBackgroundJob = "wait_background_job"
+
+	maxBackgroundJobOutputBytes = 10 * 1024 * 1024
+)
+
+// ToolSet manages long-running shell commands.
+type ToolSet struct {
+	handler *backgroundJobsHandler
+}
+
+// Verify interface compliance
+var (
+	_ tools.ToolSet      = (*ToolSet)(nil)
+	_ tools.Startable    = (*ToolSet)(nil)
+	_ tools.Instructable = (*ToolSet)(nil)
+)
+
+type backgroundJobsHandler struct {
+	shell           string
+	shellArgsPrefix []string
+	env             []string
+	workingDir      string
+	jobs            *concurrent.Map[string, *backgroundJob]
+	jobCounter      atomic.Int64
+	recall          bool
+}
+
+const (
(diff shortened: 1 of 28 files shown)


pkg/tools/builtin/backgroundjobs/backgroundjobs.go
```diff
@@ -0,0 +1,607 @@
...
+			AddDescriptionParameter: true,
+		},
+	}, nil
+}
+
+func (t *ToolSet) Start(context.Context) error {
+	return nil
+}
+
+func (t *ToolSet) Stop(context.Context) error {
```
docker-agent, line 599: **[MEDIUM] `Stop()` does not join `monitorJob` goroutines before returning**

`ToolSet.Stop()` CAS-flips each running job to `statusStopped` and calls `kill()`, but it never waits on the `job.done` channels before returning. This means `monitorJob` goroutines can still be executing `cmd.Wait()` (and the lock/status-check block that follows) after `Stop()` has already returned to the caller.

The CAS-before-kill pattern correctly prevents `rt.Recall()` from firing for stopped jobs, so no recall message will be sent. However, the goroutines still hold references to `job.outputMu`, `job.output`, and `job.rt` — fields that live on the heap-allocated `backgroundJob` struct and remain valid as long as the goroutine holds a reference, which Go's GC guarantees. In practice the risk is low in steady-state operation, but if the ToolSet is torn down in a test or a shutdown sequence that quickly re-initialises state, the unsynchronised goroutines can be surprising to reason about and may cause flaky tests.

Consider waiting for all goroutines after killing them:
```go
func (t *ToolSet) Stop(context.Context) error {
    t.handler.jobs.Range(func(_ string, job *backgroundJob) bool {
        if job.status.CompareAndSwap(statusRunning, statusStopped) {
            _ = kill(job.process, job.processGroup)
        }
        return true
    })
    // Wait for all monitorJob goroutines to finish
    t.handler.jobs.Range(func(_ string, job *backgroundJob) bool {
        <-job.done
        return true
    })
    return nil
}
```

pkg/tools/builtin/backgroundjobs/backgroundjobs.go
```diff
@@ -0,0 +1,607 @@
...
+	if params.Timeout > 0 {
+		timeout = time.Duration(params.Timeout) * time.Second
+	}
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	select {
+	case <-job.done:
+		status := statusStrings[job.status.Load()]
```
docker-agent, line 393: **[LOW] Bare map lookup in `WaitBackgroundJob` bypasses the `statusToString()` safety helper**

Line 393 uses `statusStrings[job.status.Load()]` directly (no ok-check). If a future status constant is added without a corresponding entry in the `statusStrings` map, this returns an empty string, producing a confusing message like `"Job job_42 ."` with a blank status.

The `statusToString()` helper (defined just a few lines above) already handles this gracefully by falling back to `"unknown"`. All other callers in this file already use the helper; this one slipped through.

```go
// Current (fragile):
status := statusStrings[job.status.Load()]

// Suggested (consistent with the rest of the file):
status := statusToString(job.status.Load())
```

pkg/tools/builtin/backgroundjobs/backgroundjobs.go
```diff
@@ -0,0 +1,607 @@
...
+
+	go h.monitorJob(context.WithoutCancel(ctx), job, cmd)
+
+	var recallStatus string
+	if params.Recall {
+		recallStatus = "\nRecall: requested; a steering message will be sent when the job finishes."
+	}
+
+	return tools.ResultSuccess(fmt.Sprintf("Background job started with ID: %s\nCommand: %s\nWorking directory: %s%s",
+		jobID, params.Cmd, params.Cwd, recallStatus)), nil
```
docker-agent, line 267: **[LOW] Success message reports the unresolved `params.Cwd` instead of the actual working directory**

The `run_background_job` success message shows the user-supplied `params.Cwd` string verbatim, but the command actually runs in the directory returned by `h.resolveWorkDir(params.Cwd)` (assigned to `cmd.Dir`). When `params.Cwd` is empty or `"."`, the message will print `"Working directory: "` or `"Working directory: ."` rather than the true absolute path, making it harder to debug where the process actually ran.

```go
// Current — reports raw input:
return tools.ResultSuccess(fmt.Sprintf("Background job started with ID: %s\nCommand: %s\nWorking directory: %s%s",
    jobID, params.Cmd, params.Cwd, recallStatus)), nil

// Suggested — capture the resolved path and use it in the message:
resolvedCwd := h.resolveWorkDir(params.Cwd)
cmd.Dir = resolvedCwd
// ...
return tools.ResultSuccess(fmt.Sprintf("Background job started with ID: %s\nCommand: %s\nWorking directory: %s%s",
    jobID, params.Cmd, resolvedCwd, recallStatus)), nil
```

docker-agent (commented):

This PR cleanly extracts background-job tooling into its own package. The refactoring is well-structured, mutex usage is consistent, and the CAS-before-kill pattern in the stop/monitor flow is correct. Three findings in the new `backgroundjobs` package are worth addressing before merge.

dgageot (approved):

PR #3474, opened 2026-07-04
fix: support steering in lean TUI

rumpl:
## Summary
- route busy lean TUI editor submissions through the runtime steering queue
- render pending steering/follow-up messages at the end of the live stream with muted styling
- reconcile runtime user-message confirmations so pending steering messages do not duplicate

## Tests
- go test ./pkg/app ./pkg/leantui ./pkg/tui/page/chat
- task lint
- task --force build
- task test *(fails in pkg/model/provider: TestOpenAIAliasProvider_LiveAPI/moonshot returns HTTP 404 / permission denied for kimi-k2-0905-preview)*

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -817,6 +817,11 @@ func (a *App) Resume(req runtime.ResumeRequest) {
 	a.runtime.Resume(a.ctx(), req)
 }
 
+// Steer queues a user message for mid-turn injection into the running agent.
+func (a *App) Steer(ctx context.Context, msg runtime.QueuedMessage) error {
+	return a.runtime.Steer(ctx, msg)
+}
+
 // TogglePause toggles whether the runtime loop is paused at iteration
 // boundaries. The second return value is false if the underlying runtime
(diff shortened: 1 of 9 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The steering queue routing, pending-message reconciliation, and deduplication logic introduced in this PR are all correct.

**Key paths reviewed:**
- `dispatchUserMessage` correctly routes `busySubmitSteer` via `app.Steer` and records a pending entry with `display`/`content` fields.
- `handleUserMessageEvent` correctly consumes the pending entry by matching on `content` (not `display`), so the user-facing display text (e.g. `/change`) is preserved when echoing while the resolved content is used for matching.
- `samePendingUserContent` correctly handles the trailing `\n` that the runtime appends to emitted messages via `strings.TrimSuffix`, so reconciliation succeeds reliably.
- `finishBusy` correctly dequeues follow-up messages and promotes their pending display text before echoing.
- Error from `app.Steer` is surfaced to the user via a notice; no silent failure.

dgageot (approved):

PR #3483, opened 2026-07-06
fix(session): accept legacy agentName JSON key when unmarshaling Message

dgageot:
PR #3471 renamed the JSON tag on `session.Message.AgentName` from `agentName` to `agent_name`. However, contrary to that PR's description, Go's `encoding/json` does *not* perform case-insensitive matching between `agentName` and a field tagged `agent_name` — the two differ by more than case. As a result, any previously exported session files or evaluation JSON documents that used the old key silently dropped the agent name on load.

This PR adds a compatibility `UnmarshalJSON` on `session.Message` that accepts both `agentName` and `agent_name`. The new canonical `agent_name` key wins when both are present; decoding is done through a pointer-aliased struct to preserve the stock null-handling and field-merging semantics without recursion. Marshaling is unchanged and still emits only `agent_name`. Unit tests cover the legacy key, the new key, both keys present at once, and a realistic end-to-end legacy session document. The evaluation docs example is updated to the canonical key.

No breaking changes. Existing sessions written with `agent_name` are unaffected; sessions written with the old `agentName` key now round-trip correctly.

### DIFF
--- a/docs/features/evaluation/index.md
+++ b/docs/features/evaluation/index.md
@@ -69,7 +69,6 @@ Each eval file is a JSON session that captures a complete conversation. The key 
   "messages": [
     {
       "message": {
-        "agentFilename": "./agent.yaml",
         "message": {
           "role": "user",
           "content": "How many files in the local folder?"
@@ -78,7 +77,7 @@ Each eval file is a JSON session that captures a complete conversation. The key 
     },
     {
       "message": {
-        "agentName": "root",
+        "agent_name": "root",
         "message": {
           "role": "assistant",
           "tool_calls": [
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `UnmarshalJSON` implementation is correct and well-structured. The pointer-aliased type alias technique (`type message Message` + embedded `*message`) properly avoids infinite recursion while populating all `Message` fields in-place during unmarshal. The fallback logic (`if m.AgentName == ""`) correctly handles both absent and explicitly-empty `agent_name` values. No nil-pointer risks are introduced — the `*message` pointer is always initialized before unmarshal. The design decision to treat empty and absent `agent_name` equivalently is intentional and clearly documented in the comment.

No bugs found in the introduced code.

Sayt-0 (approved):

PR #3498, opened 2026-07-07
feat(picker): add Open Board button to the agent picker

dgageot:
The agent picker (`--agent-picker`) had no way to reach the Kanban board without first running an agent. Users who want to review the board before selecting an agent had to exit and run a separate command.

Pressing `b` or clicking the new `[ Open Board ]` button in the picker now launches `docker agent board` directly. A `errAgentPickerStartBoard` sentinel is returned from the picker, and `run.go` handles it by calling `boardtui.Run` with the appropriate board telemetry — keeping all board-start logic in one place. Hit-zone math is factored into a shared `bottomRowY()` helper alongside the existing lean checkbox, `panelSize()` accounts for the extra width of the bottom row, and header lines are truncated on narrow terminals so mouse hit-testing stays correct regardless of terminal width. Tests cover the key press, click hit-zone boundaries, windowed list scrolling, narrow widths, and the details dialog intercepting `b`.


cmd/root/agent_picker.go
```diff
@@ -677,17 +732,25 @@ func (m *agentPickerModel) headerText() (title, subtitle, help string) {
...
+		agentPickerKeys.Lean.Help().Key + " " + agentPickerKeys.Lean.Help().Desc,
+		agentPickerKeys.Board.Help().Key + " " + agentPickerKeys.Board.Help().Desc,
+		agentPickerKeys.Quit.Help().Key + " " + agentPickerKeys.Quit.Help().Desc,
+	}, "   ")
+	// Truncate header lines to the panel's content width so they can't
+	// terminal-wrap on narrow terminals: a wrapped line would shift every row
+	// below it and break the row-based mouse hit-testing.
+	if m.width > 0 {
+		maxWidth := max(m.width-2*(1+4), agentPickerMinCardWidth)
+		subtitleText = toolcommon.TruncateText(subtitleText, maxWidth)
```
docker-agent, line 749: **[MEDIUM] Title line not truncated on narrow terminals, potentially breaking hit-zone math**

The `headerText()` function truncates `subtitleText` and `helpText` to `maxWidth = max(m.width-2*(1+4), agentPickerMinCardWidth)` on narrow terminals — but the `title` ("Choose an agent to run", 22 visible chars) is rendered without truncation. The comment right above the truncation block explicitly states the reason: *"a wrapped line would shift every row below it and break the row-based mouse hit-testing."* The title is subject to the same risk.

`panelSize()` does include `lipgloss.Width(title)` as a candidate for `contentWidth`, so the panel is computed wide enough to hold the title. However, if the terminal is narrower than the computed panel width (i.e. < 32 columns), lipgloss renders the panel at its full computed width, which may exceed the terminal width. Whether the title wraps or clips depends on the renderer, but the omission is inconsistent with the intent of the new truncation block — subtitle and help are explicitly protected while title is not.

**Suggested fix** — apply the same guard inside the `if m.width > 0` block:
```go
title = toolcommon.TruncateText(
    styles.HighlightWhiteStyle.Render("Choose an agent to run"),
    maxWidth,
)
```
Or truncate the raw string before rendering, matching the subtitle/help pattern.

docker-agent (commented):

One medium-severity finding: the new narrow-terminal truncation guard protects subtitle and help but leaves the title untruncated.

Sayt-0 (approved):

PR #3503, opened 2026-07-07
fix(cli): resolve session DB default against the data dir

dgageot:
The `--session-db` default for `run`, `serve a2a`, and `serve acp` was computed at flag-registration time as `~/.cagent/session.db` (using `paths.GetHomeDir()`), before the root `PersistentPreRunE` applied the `--data-dir` override via `paths.SetDataDir`. As a result, `--data-dir` never relocated the session database: sessions were always read from and written to `~/.cagent/session.db`, making past sessions stored under a custom data dir invisible.

The flag default is now empty. A new `sessionDBPath()` helper in `cmd/root/flags.go` lazily resolves it to `<data-dir>/session.db` at command run time, after any directory overrides have been applied. An explicit `--session-db` value still wins and `~` expansion is preserved. `serve api` keeps its separate CWD-relative `session.db` default, unchanged.

New regression tests in `cmd/root/session_db_test.go` cover helper resolution, explicit-flag precedence, and wiring through `createSessionRequest`. Docs in `docs/features/cli/index.md` and `docs/features/acp/index.md` are updated to reflect the corrected defaults and the previously missing `--session-db` row in the `serve a2a` table.

### DIFF
--- a/cmd/root/a2a.go
+++ b/cmd/root/a2a.go
@@ -1,14 +1,11 @@
 package root
 
 import (
-	"path/filepath"
-
 	"github.com/spf13/cobra"
 
 	"github.com/docker/docker-agent/pkg/a2a"
 	"github.com/docker/docker-agent/pkg/cli"
 	"github.com/docker/docker-agent/pkg/config"
-	"github.com/docker/docker-agent/pkg/paths"
 	"github.com/docker/docker-agent/pkg/telemetry"
 )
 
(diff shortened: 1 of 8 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The lazy `sessionDBPath()` helper correctly defers path resolution until after `PersistentPreRunE` has applied any `--data-dir` override. All three call-sites (`run.go` via `payload.go`, `a2a.go`, `acp.go`) route through the helper. Explicit `--session-db` values still take priority, `~` expansion is preserved, and unused imports are properly cleaned up. No logic errors, nil dereferences, race conditions, or security issues were found in the added code.

Sayt-0 (approved):

PR #3508, opened 2026-07-07
fix(board): surface agent startup failures instead of silently doing nothing

dgageot:
On the Kanban board (`docker agent board`), an agent that crashed at startup — bad ref, missing API key, `--listen` bind failure — left its card stuck in "starting" forever. Relaunch errors were silently discarded, the controller relaunched the crashing agent every 500ms (destroying the dead pane's error output each time), a first-launch failure before the worktree existed caused every subsequent relaunch to fail on a nonexistent working directory, and the user could not even attach to investigate because attach is gated on the control plane answering.

The fix caps consecutive startup deaths at three before giving up: the card transitions to an error state, the dead tmux pane is preserved so its output can be read, and polling slows to 5 s instead of 500 ms. `watch()` now surfaces relaunch failures as `StatusError` rather than dropping them. `relaunch()` falls back to launching from the repo root with `--worktree`/`--worktree-base` when the card's worktree was never created. `AttachCommand` allows attaching to an errored card even when the control plane is down, and returns a clear error when the session is gone entirely (via a new `sessionManager.Exists` check). The crash-loop counter is per-card controller state; a user-initiated relaunch (one that carries a prompt) resets it so a red card gets fresh attempts when moved forward.

No breaking changes. The new `sessionManager.Exists` method wraps `tmux has-session` and is additive.

### DIFF
--- a/pkg/board/app.go
+++ b/pkg/board/app.go
@@ -370,14 +370,20 @@ var ErrAgentStarting = errors.New("the agent is still starting")
 // agent's control plane answers, so the user never lands on a bare launch
 // command. An errored card is attachable regardless: its agent may have
 // died at startup, and the dead pane holds the error output the user needs
-// to read. Before attaching, the session's header row is refreshed with the
-// card's current title and project.
+// to read — unless the session is gone entirely (its recreation failed), in
+// which case a clear error beats tmux's raw "can't find session". Before
+// attaching, the session's header row is refreshed with the card's current
+// title and project.
 func (a *App) AttachCommand(cardID string) (*exec.Cmd, error) {
 	card, err := a.store.GetCard(cardID)
 	if err != nil {
 		return nil, err
 	}
-	if card.Status != StatusError && !a.controller.Ready(card) {
+	if card.Status == StatusError {
+		if exists, err := a.sessions.Exists(card.Session); err == nil && !exists {
+			return nil, errors.New("the agent's session is gone; move the card forward to relaunch it, or delete it")
+		}
+	} else if !a.controller.Ready(card) {
 		return nil, ErrAgentStarting
 	}
 	socket, err := TmuxSocketPath()
--- a/pkg/board/app.go
+++ b/pkg/board/app.go
@@ -368,14 +368,16 @@ var ErrAgentStarting = errors.New("the agent is still starting")
 // AttachCommand returns the command that attaches the caller's terminal to
 // the card's agent session. It fails with [ErrAgentStarting] until the
 // agent's control plane answers, so the user never lands on a bare launch
-// command. Before attaching, the session's header row is refreshed with the
+// command. An errored card is attachable regardless: its agent may have
+// died at startup, and the dead pane holds the error output the user needs
+// to read. Before attaching, the session's header row is refreshed with the
 // card's current title and project.
 func (a *App) AttachCommand(cardID string) (*exec.Cmd, error) {
 	card, err := a.store.GetCard(cardID)
 	if err != nil {
 		return nil, err
 	}
-	if !a.controller.Ready(card) {
+	if card.Status != StatusError && !a.controller.Ready(card) {
 		return nil, ErrAgentStarting
 	}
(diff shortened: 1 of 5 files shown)


melmennaoui (approved):

pkg/board/app.go
```diff
@@ -368,14 +368,22 @@ var ErrAgentStarting = errors.New("the agent is still starting")
...
+// attaching, the session's header row is refreshed with the card's current
+// title and project.
 func (a *App) AttachCommand(cardID string) (*exec.Cmd, error) {
 	card, err := a.store.GetCard(cardID)
 	if err != nil {
 		return nil, err
 	}
-	if !a.controller.Ready(card) {
+	if card.Status == StatusError {
+		if exists, err := a.sessions.Exists(card.Session); err == nil && !exists {
```
docker-agent, line 383: **[LOW] `Exists()` error silently ignored for `StatusError` cards in `AttachCommand`**

When `card.Status == StatusError`, the check is:
```go
if exists, err := a.sessions.Exists(card.Session); err == nil && !exists {
    return nil, errors.New("the agent's session is gone; ...")
}
```
If `Exists()` returns `(false, someErr)` — for example, an unexpected tmux error string not covered by `isNoSuchSession` — the condition `err == nil && !exists` is false, so the code falls through and attempts to attach anyway. The user then receives tmux's raw error instead of the friendly message the PR intends to provide.

`isNoSuchSession` already handles the common "no server running" case (returning `(false, nil)`), so this only bites on unanticipated tmux error output. Consider returning the error explicitly when `Exists()` fails, e.g.:
```go
exists, err := a.sessions.Exists(card.Session)
if err != nil || !exists {
    return nil, errors.New("the agent's session is gone; move the card forward to relaunch it, or delete it")
}
```

pkg/board/controller.go
```diff
@@ -160,15 +197,25 @@ func (c *controller) watch(ctx context.Context, cardID string) {
...
 			// gone, relaunch to resume; otherwise it is still starting, so
-			// just wait and retry.
+			// just wait and retry. An agent that keeps dying before ever
+			// answering is crashing at startup: surface the failure instead
+			// of silently relaunching forever, and stop killing the dead
+			// pane so the user can attach and read the agent's last output.
+			delay := retryDelay
 			if alive, aerr := c.sessions.Alive(card.Session); aerr == nil && !alive {
-				_ = c.relaunch(card, "")
+				if c.launchFailed(cardID) >= maxLaunchFailures {
```
docker-agent, line 206: **[LOW] Crash-loop counter grows unboundedly while card remains in `StatusError`**

Once `launchFailed(cardID) >= maxLaunchFailures` and the card transitions to `StatusError`, the watcher continues to loop every `cappedRetryDelay` (5 s). On each iteration the pane is still dead (`alive == false`), so `launchFailed()` is called again — incrementing the counter to 4, 5, 6, … indefinitely.

The practical impact is minimal (`int` won't overflow in any realistic timeframe, and `setStatus(StatusError)` is idempotent). However, after the user triggers a prompt-bearing relaunch, `resetLaunchFailures` correctly clears the counter, so recovery still works. The simplest fix is to skip the `launchFailed()` increment when the card is already in `StatusError`:
```go
if card.Status != StatusError {
    if c.launchFailed(cardID) >= maxLaunchFailures {
        c.setStatus(cardID, StatusError)
        delay = cappedRetryDelay
    } else {
        c.resume(card)
    }
} else {
    delay = cappedRetryDelay
}
```
Or alternatively read the card's status after the store fetch (it's already available) and bail out of the `Alive`/`launchFailed` branch early.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly addresses a real UX gap — startup-crashing agents leaving cards stuck in "starting" forever. The overall design (crash-loop cap, dead-pane preservation, worktree fallback, errored-card attach) is sound and well-tested. Two low-severity observations on newly added code are left as inline comments.

PR #3514, opened 2026-07-07
chore(board): rename $BOARD_EDITOR to $DOCKER_AGENT_BOARD_EDITOR

dgageot:
The board feature lets users configure a preferred editor via an environment variable, but the original name `BOARD_EDITOR` is too generic and doesn't follow the `DOCKER_AGENT_` prefix convention used by the project's other env vars.

This renames the variable to `DOCKER_AGENT_BOARD_EDITOR` across the codebase and documentation. To avoid silently breaking existing setups, the old `BOARD_EDITOR` name is kept as a fallback for one release — both in the Go env lookup and in the tmux shell expansion (`${DOCKER_AGENT_BOARD_EDITOR:-${BOARD_EDITOR:-code}}`), matching the pattern already used for other renamed env vars in this project.

Users relying on `BOARD_EDITOR` will continue to see their editor launched without any change; they can migrate to `DOCKER_AGENT_BOARD_EDITOR` at their own pace before the fallback is removed.

### DIFF
--- a/docs/features/board/index.md
+++ b/docs/features/board/index.md
@@ -46,7 +46,7 @@ Requirements: `tmux` and `git` must be installed.
 | `n`           | Create a card (project + prompt)                    |
 | `enter`       | Attach to the card's agent (`ctrl+q` detaches)      |
 | `d`           | View the card's worktree diff                       |
-| `o`           | Open the card's worktree in `$BOARD_EDITOR` (`code`) |
+| `o`           | Open the card's worktree in `$DOCKER_AGENT_BOARD_EDITOR` (`code`) |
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, well-structured rename of `BOARD_EDITOR` → `DOCKER_AGENT_BOARD_EDITOR`. The backward-compatibility fallback chain is correctly implemented in both the Go env lookup and the tmux shell expansion, and documentation is updated consistently. No bugs or regressions introduced.

gtardif (approved):

PR #3515, opened 2026-07-07
fix(board): run agent control-plane sockets outside the data dir so cards start in docker sandboxes

dgageot:
When the board runs inside a Docker sandbox that bind-mounts `~/.cagent` from the host, unix sockets cannot be bound on that shared mount. Every per-card `--listen` socket lived under `~/.cagent/run`, so every agent died at startup and cards never left "starting". The fix moves those sockets into the board's per-user socket dir under the system temp dir, alongside the private tmux socket, which is always sandbox-local.

Two smaller companion fixes land in the same batch. When a card's session could not even be recreated on relaunch, `resume()` was silently dropping the error: the card went red with no pane to attach to and `attach` only said "the agent's session is gone". The controller now records the last relaunch failure per card and `AttachCommand` surfaces it. The other fix tightens the teardown path: per-card controller state is now stopped before waiting for the watcher, forgotten after the wait, and forgotten again under the relaunch lock in `Teardown`, closing a race window where a concurrent relaunch could re-insert state for a card being deleted.

All three fixes are covered by new regression tests (`TestSocketPathOutsideDataDir`, `TestControllerFailedRelaunchGoesRed`, `TestAttachCommandExplainsFailedRelaunch`, `TestTeardownForgetsControllerState`), and `task test` / `task lint` are clean. One migration caveat: cards whose agents are still alive on the old `~/.cagent/run` socket stay unreachable until that agent exits, then self-heal on relaunch.

### DIFF
--- a/pkg/board/controller_test.go
+++ b/pkg/board/controller_test.go
@@ -491,3 +491,22 @@ func TestRelaunchResumesFromExistingWorktree(t *testing.T) {
 	assert.Equal(t, wt, sessions.workDir)
 	assert.Empty(t, sessions.worktreeName)
 }
+
+func TestTeardownForgetsControllerState(t *testing.T) {
+	t.Parallel()
+
+	// A SendPrompt relaunch runs outside the watcher, so Stop cannot wait
+	// for it: Teardown must drop the per-card state it may have re-added.
+	store := testStore(t)
+	require.NoError(t, store.InsertCard(&Card{ID: "c1", Session: "s", Worktree: t.TempDir()}))
+
+	c := newController(t.Context(), store, &crashingSessions{newErr: errors.New("boom")}, func() {})
+	card, err := store.GetCard("c1")
+	require.NoError(t, err)
+	c.resume(card)
+	require.Error(t, c.LaunchError("c1"))
+
+	c.Teardown(card)
+	require.NoError(t, c.LaunchError("c1"))
+	assert.Equal(t, 1, c.launchFailed("c1"), "failure count should have been dropped")
+}
--- a/pkg/board/controller_test.go
+++ b/pkg/board/controller_test.go
@@ -295,7 +295,7 @@ func (s *crashingSessions) Exists(string) (bool, error) { return false, nil }
 
 // watchCrashingCard spins up a watcher for a card whose agent never answers
 // and whose tmux pane is dead, simulating a startup crash.
-func watchCrashingCard(t *testing.T, sessions *crashingSessions) *Store {
+func watchCrashingCard(t *testing.T, sessions *crashingSessions) (*Store, *controller) {
 	t.Helper()
 	store := testStore(t)
 	require.NoError(t, store.InsertCard(&Card{ID: "c1", Column: "dev", Status: StatusStarting, Session: "s", Worktree: t.TempDir()}))
@@ -309,7 +309,7 @@ func watchCrashingCard(t *testing.T, sessions *crashingSessions) *Store {
 	require.NoError(t, err)
 	c.Start(card)
 	t.Cleanup(func() { c.Stop("c1") })
-	return store
+	return store, c
 }
 
 func TestControllerStartupCrashLoopGoesRed(t *testing.T) {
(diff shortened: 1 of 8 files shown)


melmennaoui (approved):

pkg/board/controller_test.go
```diff
@@ -486,3 +491,22 @@ func TestRelaunchResumesFromExistingWorktree(t *testing.T) {
...
+
+	c := newController(t.Context(), store, &crashingSessions{newErr: errors.New("boom")}, func() {})
+	card, err := store.GetCard("c1")
+	require.NoError(t, err)
+	c.resume(card)
+	require.Error(t, c.LaunchError("c1"))
+
+	c.Teardown(card)
+	require.NoError(t, c.LaunchError("c1"))
+	assert.Equal(t, 1, c.launchFailed("c1"), "failure count should have been dropped")
```
docker-agent, line 511: **[MEDIUM] Vacuous assertion: `launchFailed()` mutates state, so the post-`Teardown` check proves nothing**

`launchFailed()` is not a read-only accessor — it increments the counter and returns the new value:

```go
func (c *controller) launchFailed(cardID string) int {
    c.mu.Lock()
    defer c.mu.Unlock()
    c.launchFailures[cardID]++
    return c.launchFailures[cardID]
}
```

After `c.Teardown(card)` calls `forget()` → `delete(c.launchFailures, "c1")`, the map entry is gone (value is effectively 0). Then `c.launchFailed("c1")` increments the now-absent key (0 → 1) and returns 1. The assertion `assert.Equal(t, 1, ...)` therefore **always passes**, regardless of whether `Teardown` actually cleared the state — the call itself re-creates the entry.

The test comment says *"failure count should have been dropped"* but it confirms the opposite: it does nothing to validate `Teardown`'s cleanup of `launchFailures`. A regression that broke `forget()` would not be caught.

**Suggested fix**: Check the map directly (or add a read-only getter) to assert the count is 0 after `Teardown`, rather than calling the mutating `launchFailed()`:

```go
c.Teardown(card)
require.NoError(t, c.LaunchError("c1"))
// Directly verify the failure count was zeroed, without re-incrementing it:
c.mu.Lock()
count := c.launchFailures["c1"]
c.mu.Unlock()
assert.Equal(t, 0, count, "failure count should have been dropped by Teardown")
```

docker-agent (commented):

The PR's core changes are sound: moving control-plane sockets to the system temp dir is the correct fix, the `launchErrors` map is properly locked, and the double-`forget()` teardown pattern closes the race window it targets.

One **medium** finding and one **low** finding were confirmed in newly added code.

| Severity | File | Finding |
|----------|------|---------|
| 🟡 MEDIUM | `pkg/board/controller_test.go:511` | Vacuous assertion — `launchFailed()` mutates state, so the post-`Teardown` check always passes and proves nothing |
| 🔵 LOW | `pkg/board/model.go:122` | Error-fallback in `socketPath()` drops the `"board-"` prefix and per-user isolation |

pkg/board/model.go
```diff
@@ -108,10 +109,19 @@ func newSessionName() string {
...
+// dir, whose bind mount into a docker sandbox cannot host unix sockets: the
+// agent would die at startup failing to bind --listen. Kept short to stay
+// under the ~104-byte unix sun_path limit.
 func socketPath(agentSession string) string {
-	return filepath.Join(paths.GetDataDir(), "run", "board-"+agentSession+".sock")
+	dir, err := socketDir()
+	if err != nil {
+		// The board cannot run at all when the socket dir is unusable (its
+		// tmux socket lives there too); keep the path deterministic anyway.
+		dir = os.TempDir()
```
docker-agent, line 122: **[LOW] `socketPath()` error fallback drops the `"board-"` prefix and per-user isolation**

In the happy path the socket lands in the per-user `cagent-board-<uid>` subdirectory (good). In the error fallback (`socketDir()` fails), `os.TempDir()` is used instead — a flat, world-accessible directory — and the filename is the raw `agentSession` UUID with no `"board-"` prefix or per-user isolation:

```go
func socketPath(agentSession string) string {
    dir, err := socketDir()
    if err != nil {
        dir = os.TempDir()   // e.g. /tmp — no per-user subdirectory
    }
    return filepath.Join(dir, agentSession+".sock")   // no "board-" prefix
}
```

The comment rightly notes the board is already broken when `socketDir()` fails (the tmux socket is also missing), so this path is rarely hit in practice. That said, since `socketDir` is `sync.OnceValues`, a single transient failure permanently routes **all** cards to the flat temp dir for the process lifetime, where other users could potentially observe or pre-create the socket names.

A minimal safeguard would be to retain the prefix so the names are at least recognisable:

```go
return filepath.Join(dir, "board-"+agentSession+".sock")
```

PR #3517, opened 2026-07-07
feat(board): add `s` shortcut to open a shell in the card's worktree

dgageot:
The `docker agent board` TUI had no way to drop into a shell inside a card's worktree — you had to leave the board, find the directory, and open a terminal manually. This adds an `s` shortcut that opens an interactive shell directly in the selected card's worktree, bringing board parity with the main TUI's `/shell` slash command.

When `s` is pressed the board validates that the worktree exists on disk, then execs an interactive shell in that directory using the same logic that powers `/shell`. To avoid duplicating that logic, `InteractiveShellCmd` has been extracted from `pkg/tui` into the new exported `pkg/shellpath` function, so both TUIs share a single implementation. The exit message is passed via the `DOCKER_AGENT_SHELL_EXIT_MSG` environment variable rather than being interpolated into the shell command string, closing a shell-injection vector and making the Windows (`cmd.exe` with `/V:ON`) and PowerShell paths cleaner. POSIX `printf` replaces the non-portable `echo -e`. The help dialog, footer hints, and `docs/features/board/index.md` are all updated.

If the worktree directory no longer exists the board shows a brief flash message instead of crashing, so the error surface is friendly. A test locks in the Unix argv/env contract for `InteractiveShellCmd`.

### DIFF
--- a/docs/features/board/index.md
+++ b/docs/features/board/index.md
@@ -47,6 +47,7 @@ Requirements: `tmux` and `git` must be installed.
 | `enter`       | Attach to the card's agent (`ctrl+q` detaches)      |
 | `d`           | View the card's worktree diff                       |
 | `o`           | Open the card's worktree in `$DOCKER_AGENT_BOARD_EDITOR` (`code`) |
+| `s`           | Open an interactive shell in the card's worktree    |
 | `[` / `]`     | Move the card back / forward                        |
 | `x`           | Delete the card, its session, worktree, and branch  |
 | `p`           | Manage projects (add, edit, reorder, remove)        |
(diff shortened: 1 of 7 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly extracts `InteractiveShellCmd` into a shared `pkg/shellpath` package, adds the `s` shortcut to the board TUI, and eliminates the shell-injection vector by passing the exit message through `DOCKER_AGENT_SHELL_EXIT_MSG` instead of interpolating it into the command string. The worktree-existence guard handles the empty-path and missing-directory cases correctly on all supported platforms. No bugs introduced by this PR were confirmed.

Sayt-0 (approved):

PR #3518, opened 2026-07-07
fix(board): forward directory overrides to spawned agents

dgageot:
`docker agent board` spawns each card's agent in a tmux pane as `docker-agent run <agent> …`. The board's `--config-dir`, `--data-dir`, and `--cache-dir` overrides are process-local flags, not inherited environment, so spawned agents ignored them. In environments where `$HOME` differs from the board's directories — for example, running the board inside a Docker sandbox with the host's `~/.config/cagent` and `~/.cagent` bind-mounted at their host paths — spawned agents read an empty config (alias refs like `gopher` fail to resolve and are mistaken for OCI image references, so the agent dies at startup and the card errors) and create their worktrees and session databases under a data dir the board never watches, breaking status milestones, diffs, and cleanup.

`agentCommand` in `pkg/board/tmux.go` now forwards the board's own directories to every spawned agent via `--config-dir`, `--data-dir`, and `--cache-dir`, using `paths.GetConfigDir()`, `paths.GetDataDir()`, and `paths.GetCacheDir()`, all shell-quoted. On hosts without overrides this forwards the defaults — a behavioral no-op. Flags are used deliberately rather than environment variables: the board's private tmux server survives board restarts, so panes could inherit stale env; flags take precedence over env in the child's resolution order.

The existing `TestAgentCommand` is extended to assert the forwarded flags, and a new `TestAgentCommandForwardsDirOverrides` test sets real `paths.Set{Config,Data,Cache}Dir` overrides and asserts they appear correctly quoted in the spawned command.

### DIFF
--- a/pkg/board/tmux.go
+++ b/pkg/board/tmux.go
@@ -173,14 +173,15 @@ func applyServerDefaults(ctx context.Context) {
 // owns sessionID and passes it via --session: the first run creates that
 // session, later runs resume it.
 //
-// --config-dir and --data-dir forward the board's own directories so the
-// agent resolves the same aliases and creates its worktree and session under
-// the data dir the board watches. Directory overrides are process-local
-// flags, not inherited env, so without forwarding an agent spawned in an
-// environment whose $HOME differs from the board's directories (e.g. a
-// docker sandbox with the host dirs bind-mounted) would silently use its
-// own empty config and a data dir the board never looks at. On a host with
-// no overrides this forwards the defaults — a no-op.
+// --config-dir, --data-dir, and --cache-dir forward the board's own
+// directories so the agent resolves the same aliases, creates its worktree
+// and session under the data dir the board watches, and shares the board's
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The drafter was unable to complete the full analysis of this PR (returned `review_complete: false`). No issues were surfaced.

Based on the diff, the change is straightforward: `agentCommand` in `pkg/board/tmux.go` now forwards `--config-dir`, `--data-dir`, and `--cache-dir` to each spawned agent using `paths.GetConfigDir()`, `paths.GetDataDir()`, and `paths.GetCacheDir()`, all run through the existing `shQuote` helper. The approach is consistent with the existing pattern, the flags are correctly ordered in the format string, and the test coverage extension follows the project's style.

No bugs identified in the changed code.

Sayt-0 (approved):

PR #3520, opened 2026-07-08
fix(board): clarify dirty-worktree removal prompt with explicit consequences

dgageot:
When a card's worktree has uncommitted changes or unpushed commits and the user
tries to close the card, docker-agent asks whether to delete it. The old prompt
crammed the question and the explanation into a single line, making it unclear
what "y" and "N" would actually do. Users had to parse a single run-on sentence
to understand that "y" meant permanent deletion.

The prompt now prints the two options on separate indented lines before asking
the question, so the consequences of each choice are unambiguous at a glance.
The question itself is moved to the last line using `Print` instead of
`Println`, which leaves the cursor right after the `(y/N)` marker — consistent
with how every other interactive prompt in the CLI behaves.

No behavioural change; the logic that reads and acts on the response is
untouched.

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -674,7 +674,9 @@ func promptRemoveDirtyWorktree(ctx context.Context, out *cli.Printer, wt *worktr
 	}
 
 	out.Println("\nThe git worktree " + wt.Dir + " (branch " + wt.Branch + ") still has " + strings.Join(held, ", ") + ".")
-	out.Println("Remove it and discard this work? Keeping preserves the directory and branch so you can return later. (y/N):")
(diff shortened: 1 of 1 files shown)


melmennaoui (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

Clean, focused prompt reformatting. The switch from `Println` to `Print` for the question line is correct — the trailing space keeps the cursor inline for the `ReadLine` call, consistent with other interactive prompts in the CLI. The two new indented option lines make the consequences of each choice unambiguous. No behavioural change, no bugs introduced.

PR #3522, opened 2026-07-08
feat(board): move cards to any column with digit keys and drag & drop

dgageot:
The Kanban board previously only let you step a card one column at a time with `[` and `]`. For boards with many columns this is tedious, and there was no way to reposition a card at all without a keyboard.

This PR adds two complementary shortcuts. Pressing a digit key 1–9 moves the selected card directly to that column number (`MoveTo` binding), while `[`/`]` continue to do single-step moves. Moving a card forward still triggers the destination column's prompt via the existing `App.MoveCard` engine path, so the workflow is unchanged. For mouse users, pressing on a card and dragging it over another column highlights that column's border in green and shows a "Release to move the card to \<column\>" hint in the footer; releasing the button completes the move. Drops on the origin column or outside the board are no-ops, and motion jitter within the pressed cell is treated as a click rather than a drag.

A second commit hardens the edge cases surfaced during review: drag state is cleared whenever a dialog opens or a key is pressed, and it is re-validated on every board refresh so a swallowed mouse-release cannot leave stale state that fires on a later click. The drop resolves the target by card ID rather than position, so a refresh mid-drag cannot move the wrong card. Non-left-button releases are ignored, and `renderFooter` bounds-checks the drop-target column index before accessing it.

### DIFF
--- a/docs/features/board/index.md
+++ b/docs/features/board/index.md
@@ -49,11 +49,12 @@ Requirements: `tmux` and `git` must be installed.
 | `o`           | Open the card's worktree in `$DOCKER_AGENT_BOARD_EDITOR` (`code`) |
 | `s`           | Open an interactive shell in the card's worktree    |
 | `[` / `]`     | Move the card back / forward                        |
+| `1`-`9`       | Move the card to column N                           |
 | `x`           | Delete the card, its session, worktree, and branch  |
 | `p`           | Manage projects (add, edit, reorder, remove)        |
 | `e`           | Edit the selected column's prompt                   |
 | `←↓↑→` `hjkl` | Navigate                                            |
-| mouse         | Click selects, double-click attaches, wheel scrolls |
+| mouse         | Click selects, double-click attaches, drag moves, wheel scrolls |
 | `?`           | Help                                                |
 | `q`           | Quit (agents keep running)                          |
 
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed the drag-and-drop + digit-key card move implementation across `pkg/board/tui/tui.go`, `pkg/board/tui/view.go`, `pkg/board/tui/dialogs.go`, and `pkg/board/tui/tui_test.go`.

**Summary**: The implementation is sound. Key design decisions were validated:

- **`moveCard` bounds-checks `dst`** (`dst < 0 || dst >= len(m.columns)`) — digit 1–9 and drag releases outside columns are safe no-ops.
- **Drag state invariant holds**: `m.dragging = true` is only set in `handleMotion` when `m.dragCardID != ""` (guarded at line 745), so `handleRelease` can never call `moveCard` with an empty card ID.
- **`msg.String()[0]` in `MoveTo`** is safe: bubbletea guarantees a single-character string for single-character key bindings, and the path is only reached after `key.Matches` succeeds for one of "1"–"9".
- **Drag state cleared correctly** on dialog open, key press, and board refresh — the re-validation in `reload()` catches any swallowed mouse-release.
- **Drop resolves by card ID**, not position — a refresh mid-drag cannot redirect to the wrong card.
- **`renderFooter` bounds-checks** `m.dragCol` before indexing `m.columns`.
- **Non-left-button releases ignored** in `handleRelease` (line 767).

No bugs introduced by this PR were found.

Sayt-0 (approved):

PR #3524, opened 2026-07-08
feat(board): fade dragged card immediately and fix stale drag state

dgageot:
When dragging a card on the board, there was no visual feedback during the drag itself — the card looked identical whether it was being moved or stationary, making it hard to tell that a drag was in progress. Additionally, the drag only started after the pointer had moved outside the card's bounds, which felt sluggish and meant a fast move could pick up the wrong card.

The card now begins fading toward the theme background on the first mouse motion while the button is pressed, not when the pointer crosses a boundary. New helpers in `pkg/board/tui/view.go` blend the card's title, project badge, status line, and border color toward the background, giving a clear "lifted" appearance. The click-vs-drag decision moved to release time: releasing the button over the card that was originally pressed counts as a plain click (so double-click-to-attach still works), while releasing anywhere else finalizes the move.

A follow-up fix resets drag state that was lost when `tea.ExecProcess` (the attach flow) consumed the mouse-release event, preventing a later plain click from silently moving a card. Wheel events are also ignored while a card is held down to avoid accidental column scrolls mid-drag.

### DIFF
--- a/pkg/board/tui/tui.go
+++ b/pkg/board/tui/tui.go
@@ -413,6 +413,9 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m, cmd
 
 	case attachReadyMsg:
+		// tea.ExecProcess leaves mouse mode: a drag in flight loses its
+		// release and must not leak into the gestures after detach.
+		m.resetDrag()
 		return m, tea.ExecProcess(msg.cmd, func(err error) tea.Msg { return attachDoneMsg{err: err} })
 
 	case attachFailedMsg:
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The drag-fade implementation is well-structured. The visual feedback helpers in `view.go` correctly blend card colors toward the background using explicit foreground values that are always set. The click-vs-drag decision at release time is logically sound — bubbletea's single-threaded update loop means there is no race between `cardAt` bounds-checking and the card-ID lookup on the same line. The wheel-suppression guard (`dragCardID != ""`) is intentional per the PR description and avoids accidental column scrolls during a card press. No bugs found in the introduced changes.

Sayt-0 (approved):

PR #3526, opened 2026-07-08
docs: add desktop notifications with hooks tip

dgageot:
Long-running agents shouldn't require staring at the terminal. This adds a new tip to the guides showing how to wire up desktop notifications using global hooks in your user config, so every agent can alert you when it needs input or finishes a task — without any per-agent config changes.

The section covers macOS (`osascript`) and Linux (`notify-send`) for the `on_user_input` and `stop` events, a more advanced example using the `notification` event that reads a structured message from the stdin JSON payload via `jq`, and a pointer to the built-in `settings.sound` option for cases where an audio cue is sufficient.

### DIFF
--- a/docs/guides/tips/index.md
+++ b/docs/guides/tips/index.md
@@ -377,6 +377,48 @@ settings:
 
 This model is used when you run `docker agent run` without a config file.
 
+### Get Desktop Notifications with Hooks
+
+Long-running agents shouldn't require staring at the terminal. Add [global hooks](../../configuration/hooks/index.md#global-user-level-hooks) to your user config so every agent notifies you when it needs attention or finishes:
+
+```yaml
+# ~/.config/cagent/config.yaml
+settings:
+  hooks:
+    # Agent is waiting for your input (question, approval prompt, ...)
+    on_user_input:
+      - type: command
+        command: osascript -e 'display notification "Agent needs your input" with title "docker-agent"'
+
+    # Agent finished responding
+    stop:
+      - type: command
+        command: osascript -e 'display notification "Task finished" with title "docker-agent"'
+```
+
+On Linux, replace `osascript` with `notify-send`:
+
+```yaml
+command: notify-send "docker-agent" "Agent needs your input"
+```
+
+To also get alerted on errors and warnings, hook the `notification` event and read the message from the JSON payload on stdin:
+
+```yaml
+settings:
+  hooks:
+    notification:
+      - type: command
+        timeout: 10
+        command: |
+          MESSAGE=$(cat | jq -r '.notification_message // "Agent error"')
+          osascript -e "display notification \"$MESSAGE\" with title \"docker-agent\""
+```
+
+If a sound is enough, set `settings: { sound: true }` instead — docker-agent plays a notification sound when a task that ran longer than `sound_threshold` seconds (default 10) succeeds or fails.
+
+See the [Hooks documentation](../../configuration/hooks/index.md) for the full list of events and their payloads.
+
 ### GitHub PR Reviewer Example
 
 Use docker-agent as a GitHub Actions PR reviewer:


Sayt-0 (approved):

docs/guides/tips/index.md
```diff
@@ -377,6 +377,48 @@ settings:
...
+
+```yaml
+settings:
+  hooks:
+    notification:
+      - type: command
+        timeout: 10
+        command: |
+          MESSAGE=$(cat | jq -r '.notification_message // "Agent error"')
+          osascript -e "display notification \"$MESSAGE\" with title \"docker-agent\""
```
docker-agent, line 415: **[HIGH] Shell injection: unsanitized `$MESSAGE` interpolated into `osascript -e` double-quoted string**

The `notification` hook example interpolates LLM/agent-generated content directly into a double-quoted shell string passed to `osascript -e`:

```sh
MESSAGE=$(cat | jq -r '.notification_message // "Agent error"')
osascript -e "display notification \"$MESSAGE\" with title \"docker-agent\""
```

If `$MESSAGE` contains a double-quote character, the `-e` argument is misparsed by the shell. If it contains backticks or `$(...)` sequences, the shell evaluates them as command substitution **before** osascript ever sees the string — a command injection vector since message content comes from the agent/LLM via stdin.

Suggested fix — pass the message via an AppleScript variable to avoid shell interpolation entirely:

```sh
MESSAGE=$(cat | jq -r '.notification_message // "Agent error"')
osascript -e "set msg to \"$MESSAGE\"" -e 'display notification msg with title "docker-agent"'
```

Or more robustly, use `printf` and process substitution / `osascript` stdin:

```sh
MESSAGE=$(cat | jq -r '.notification_message // "Agent error"')
osascript <<OSASCRIPT
display notification "$MESSAGE" with title "docker-agent"
OSASCRIPT
```

Even with the heredoc approach, a message containing a literal `"` could still break the AppleScript string — the safest approach is to sanitize or escape `$MESSAGE` before embedding it (e.g. `MESSAGE="${MESSAGE//\"/\\'}"`)

docs/guides/tips/index.md
```diff
@@ -377,6 +377,48 @@ settings:
...
+    # Agent finished responding
+    stop:
+      - type: command
+        command: osascript -e 'display notification "Task finished" with title "docker-agent"'
+```
+
+On Linux, replace `osascript` with `notify-send`:
+
+```yaml
+command: notify-send "docker-agent" "Agent needs your input"
```
docker-agent, line 402: **[MEDIUM] Linux `notify-send` example is structurally incomplete — missing `- type: command` wrapper**

The Linux snippet shows only the `command:` field in isolation:

```yaml
command: notify-send "docker-agent" "Agent needs your input"
```

But every complete hook entry in this same section (and in the macOS examples above) requires both `- type: command` and `command:` as sibling fields within the hook's list entry. Without `type: command`, this YAML is not a valid hook definition. A user copying it verbatim would get a broken configuration.

The snippet should match the structure of the macOS examples:

```yaml
on_user_input:
  - type: command
    command: notify-send "docker-agent" "Agent needs your input"
```

docs/guides/tips/index.md
```diff
@@ -377,6 +377,48 @@ settings:
...
+To also get alerted on errors and warnings, hook the `notification` event and read the message from the JSON payload on stdin:
+
+```yaml
+settings:
+  hooks:
+    notification:
+      - type: command
+        timeout: 10
+        command: |
+          MESSAGE=$(cat | jq -r '.notification_message // "Agent error"')
```
docker-agent, line 414: **[MEDIUM] Assumed JSON payload field `.notification_message` — verify against actual event schema**

The example reads:

```sh
MESSAGE=$(cat | jq -r '.notification_message // "Agent error"')
```

The field name `.notification_message` is not verified anywhere in the diff against the actual `notification` event payload schema. If the hook's stdin JSON uses a different key (e.g. `.message`, `.text`, or `.body`), the `// "Agent error"` fallback always fires silently and users will never see real notification content — making the example non-functional while appearing to work.

Please cross-check this field name against the actual event payload definition (e.g. in the hooks schema or `agent-schema.json`) before publishing, and ideally link to the payload reference in the [Hooks documentation](../../configuration/hooks/index.md) so readers can verify it themselves.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

The new notification hook example introduces a shell injection vulnerability where LLM/agent-generated message content is interpolated into a double-quoted shell string passed to `osascript -e`. Additionally, the Linux `notify-send` snippet is missing its required `- type: command` wrapper, and the JSON payload field name `.notification_message` should be verified against the actual event schema.

PR #3527, opened 2026-07-08
fix(docs): switch .toc-aside from position:fixed to sticky in flex .main row

dgageot:
The right-sidebar table of contents (`.toc-aside`) was using `position:fixed` and a full viewport height, which caused it to overlap the site footer when a reader scrolled to the bottom of a long page. Since fixed positioning takes the element out of flow, there was no natural boundary to stop it from sitting on top of the footer.

The fix replaces `position:fixed` with `position:sticky` on `.toc-aside`, pins it just below the header with `top`, and gives it a bounded `max-height` plus `overflow-y:auto` so it scrolls independently when the page content is taller than the viewport. To make content and TOC proper flow siblings, `.main` is set to `display:flex` with `flex-direction:row` at the ≥1280px breakpoint. The now-redundant `max-width:1280px` visibility-hide rule for the TOC is removed since visibility is still controlled by the default hidden state and the breakpoint show rule.

No template or JS changes are needed. Verified with a Hugo build via the docs Dockerfile; `task lint` and `task test` are clean.

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -435,13 +435,17 @@ details.sidebar-section[open] > summary.sidebar-heading::after {
 
 /* ===== Table of Contents (right sidebar) =====
  * Layout, font sizes and colors mirror docs.docker.com's right-column
- * \"Edit this page\" + \"Table of contents\" pattern. */
+ * \"Edit this page\" + \"Table of contents\" pattern.
+ * Sticky (not fixed) so it stays in the .main flex row and can never
+ * overlap the footer below; shown only at the ≥1280px breakpoint. */
 .toc-aside {
-  position: fixed;
+  display: none;
+  position: sticky;
   top: var(--header-height);
-  right: 0;
+  align-self: flex-start;
+  flex-shrink: 0;
   width: var(--toc-width);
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The CSS change correctly replaces `position:fixed` with `position:sticky` on `.toc-aside` and introduces a flex row layout on `.main` at the ≥1280px breakpoint. No real bugs were found in the introduced changes.

**Analysis summary**: Two potential concerns around `position:sticky` were investigated and dismissed: (1) the ancestor chain (html → body → .main → .toc-aside) has no `overflow:hidden/auto/scroll` that would break sticky; (2) the flex setup is textbook-correct — `align-items:flex-start` on `.main` plus `align-self:flex-start` on `.toc-aside` ensure the item doesn't stretch to full height, and `top:var(--header-height)` is present. The breakpoint semantics (TOC shown at ≥1280px) are preserved correctly with the default `display:none` + `@media(min-width:1280px) { display:block }` pattern.

Sayt-0 (approved):

PR #3532, opened 2026-07-08
feat(board): preview the dragged card at its insertion point

dgageot:
When dragging a card in the board TUI, there was no visual indication of where the card would land. Drop targets highlight the destination column with a green border, but the exact insertion point — always the tail of the column — was invisible.

While a drag is in progress and a column is highlighted as the drop target, the column now renders a faded "ghost" clone of the dragged card at the tail position, exactly where the card will be appended on drop. The column's scroll window shifts temporarily to reveal that tail slot, giving the user a clear preview of the outcome. If the drag is cancelled the scroll is restored, leaving the column untouched. The ghost uses faded title and project text with a border in faded success green, matching the existing green column-highlight treatment.

A second commit hardens the feature against edge cases: terminals too small to show even one real card get a `… N more` line instead of a ghost-only view, the ghost is suppressed when the dragged card is already in the target column to handle mid-drag refresh races, and stale drag state is cleared at the top of `handleClick` to recover from lost release events. Tests cover tail-scroll behaviour, the tiny-terminal fallback, and duplicate-card suppression.

### DIFF
--- a/pkg/board/tui/view.go
+++ b/pkg/board/tui/view.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"image/color"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -244,7 +245,11 @@ func (m *model) renderBoard() string {
 
 func (m *model) renderColumn(idx int, col board.Column, colWidth, boardHeight int) string {
 	selected := idx == m.selCol
-	dropTarget := m.dragging && idx == m.dragCol && idx != m.selCol
+	cards := m.cards[col.ID]
+	// A column already holding the dragged card is not a drop target: the
+	// move would be a no-op. This also covers a refresh moving the card
+	// under the drag — the card must never render twice in one column.
+	dropTarget := m.dragging && idx == m.dragCol && !containsCard(cards, m.dragCardID)
 	// The border box's Width/Height are outer sizes; the content area is two
 	// cells narrower (and shorter).
 	innerWidth := colWidth - 2
@@ -256,7 +261,6 @@ func (m *model) renderColumn(idx int, col board.Column, colWidth, boardHeight in
 	if selected {
 		nameStyle = styles.HighlightWhiteStyle
 	}
-	cards := m.cards[col.ID]
 	header := " " + columnLabel(col, nameStyle)
 	// Right-aligned: a clickable + to create a card on the first column
 	// (see plusButtonAt), then the card count rightmost.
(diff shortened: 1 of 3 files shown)


pkg/board/tui/view.go
```diff
@@ -277,11 +282,32 @@ func (m *model) renderColumn(idx int, col board.Column, colWidth, boardHeight in
...
 
+	// The drop target previews the dragged card at its insertion point: a
+	// faded ghost takes the last slot and the window slides to the column's
+	// tail, where a dropped card lands. The forced scroll is not persisted,
+	// so a cancelled drag leaves the column's window untouched.
+	ghost := ""
+	if dropTarget {
+		if card := m.cardByID(m.dragCardID); card != nil {
+			ghost = m.renderGhost(card, innerWidth)
+			slots--
```
docker-agent, line 293: **[LOW] Ghost preview hides all real cards when terminal fits exactly one card slot**

When `visibleSlots()` returns `1` (the minimum, enforced by the `max(..., 1)` guard), `slots--` drives it to `0`. The subsequent computation then becomes:

```go
scroll = max(len(cards)-0, 0)  // = len(cards)
end    = min(len(cards)+0, len(cards))  // = len(cards)
```

The render loop (`for i := scroll; i < end`) is empty, so every real card disappears and the column shows only `header + rule + "… N more" + ghost`. On a tall-enough terminal this never triggers (slots ≥ 2), but on a constrained terminal where only one card row fits, a drag-over causes the entire card list to vanish except the ghost.

A simple guard avoids the underflow:
```go
if slots > 1 {
    slots--
    scroll = max(len(cards)-slots, 0)
}
```
That way a one-slot terminal still renders its single real card and skips the ghost, matching the PR description's intent ("terminals too small to show even one real card get a `… N more` line").

pkg/board/tui/view.go
```diff
@@ -293,7 +319,7 @@ func (m *model) renderColumn(idx int, col board.Column, colWidth, boardHeight in
 	// dimmed secondary border.
 	borderColor := darken(styles.BorderSecondary, 0.35)
 	switch {
-	case m.dragging && idx == m.dragCol && idx != m.selCol:
+	case dropTarget:
```
docker-agent, line 322: **[LOW] Dropping `idx != m.selCol` lets `dropTarget` override the selected column's white border with green**

The old condition was:
```go
case m.dragging && idx == m.dragCol && idx != m.selCol:
```

The new `dropTarget` is:
```go
m.dragging && idx == m.dragCol && !containsCard(cards, m.dragCardID)
```

There is no longer a guard preventing `dropTarget` from being true when the destination column is also the selected column (`m.dragCol == m.selCol`). Because `case dropTarget:` appears first in the border-color switch, a drag toward a column the user previously selected will show a green border instead of the white one they'd normally see.

This may well be intentional — the green ghost+border is more informative during a drag — but since the removal is silent and the old guard was explicit, it's worth a quick confirmation that this is the desired UX before merging.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The ghost-preview feature is well-structured: the `containsCard` guard prevents double-rendering, the scroll-to-tail logic is correct for the common case, and the `resetDrag` refactor in `tui.go` is a clean improvement. Two minor edge cases in `view.go` are worth a quick look before merging.

Sayt-0 (approved):

PR #3539, opened 2026-07-08
fix(tui): drop the redundant copy toast when the inline "copied" flash shows

dgageot:
Clicking a copy button in the TUI — the "⎘ copy" label on an assistant message or the copy icon on a code block — was showing two pieces of feedback at once. The inline "copied" flash (introduced in a recent commit) would appear, and simultaneously the pre-existing "Text copied to clipboard." toast would fire. Because toasts stack and linger for 10 seconds, repeated clicks could pile them up noticeably.

The fix splits `copyTextToClipboard` in `clipboard.go` into two functions: `copyTextToClipboardSilent` (which writes to the clipboard and emits `tea.SetClipboard` only) and a thin wrapper that adds the toast on top. The two click handlers in `messages.go` that already have an inline flash are switched to the silent variant. The drag-selection and keyboard "c" copy paths keep the toast since they have no other visual confirmation.

### DIFF
--- a/pkg/tui/components/messages/clipboard.go
+++ b/pkg/tui/components/messages/clipboard.go
@@ -330,15 +330,25 @@ func (m *model) copySelectedMessageToClipboard() tea.Cmd {
 	return copyTextToClipboard(content)
 }
 
-// copyTextToClipboard copies text to the system clipboard
+// copyTextToClipboard copies text to the system clipboard and confirms with
+// a toast. Copy buttons that flash an inline "copied" label must use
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change cleanly splits `copyTextToClipboard` into a silent variant (no toast) and a thin wrapper that preserves the toast, then routes the two inline-flash click handlers through the silent path. The logic is correct and the pre-existing drag-selection / keyboard copy paths are unaffected.

gtardif (approved):

PR #3542, opened 2026-07-08
fix: correct Decide precedence and remove sub-session scope escalation

Piyush0049:
**What this PR does / why we need it**:
This PR resolves permission override scoping issues between parent and child sessions and fixes the tool-approval chain precedence. 

Previously, a successful sub-session would back-propagate its `ToolsApproved` boolean and `Permissions` to the parent session. This created a scope escalation bug where a sub-agent could elevate the parent's permissions for the remainder of the parent's turn. Additionally, the tool execution pipeline didn't correctly prioritize explicit `Deny` rules over the `--yolo` flag.

This PR fixes these issues by:
1. Removing the upward permission back-propagation (`SetToolsApproved` and `SetPermissions`) when a child session completes, ensuring permissions only flow downwards.
2. Correcting the `Decide` precedence logic so that explicit `Deny` rules correctly override `--yolo`, while keeping `--yolo` above `ForceAsk`.
3. Integrating the `PermissionsConfig.Clone()` refactor to safely isolate parent and child session permissions without data races.

**Special notes for your reviewer**:
- Added `TestRunAgent_EndToEndPermissions` to explicitly verify that the Allow, Ask, and Deny rules are correctly enforced along the real dispatch path.
- Restored the exhaustive `Decide` test matrix by adding tests for YOLO paths without checker matches and YOLO overriding `ForceAsk`.
- Refreshed relevant documentation in `docs/` to correctly indicate that explicit `Deny` rules block tools even in YOLO mode.


Piyush0049:
**Note:** Learning from previous PRs, I also added a unit test to verify that this behavior works correctly and stays fixed in the future. You can run the test locally using this command:
`go test -v -run TestSubSessionInheritsPermissions ./pkg/runtime/...`

pkg/runtime/agent_delegation.go
```diff
@@ -184,6 +184,9 @@ func newSubSession(parent *session.Session, cfg SubSessionConfig, childAgent *ag
 	if cfg.PinAgent {
 		opts = append(opts, session.WithAgentName(cfg.AgentName))
 	}
+	if parent.Permissions != nil {
+		opts = append(opts, session.WithPermissions(parent.Permissions))
```
Sayt-0: This shares the parent's `*PermissionsConfig` (and its underlying slices) with the child instead of cloning it. Every other call site clones through `clonePermissionsConfig` (`session.Clone`, `copySessionMetadata` in branch.go, store.go). The aliasing is not benign: on the interactive `transfer_task` path, `handleResume`'s `ResumeTypeApproveTool` case appends to `sess.Permissions.Allow`, so a child approving a tool would mutate the parent's permissions (and append to a shared slice). Suggest cloning here, for example by exposing the existing `clonePermissionsConfig` helper and passing `session.WithPermissions(clone)`.

pkg/runtime/agent_delegation.go
```diff
@@ -482,10 +485,6 @@ func (r *LocalRuntime) CurrentAgentSubAgentNames() []string {
 // authorises all tool calls made by the sub-agent when they approve
 // run_background_agent. Callers should be aware that prompt injection in
 // the sub-agent's context could exploit this gate-bypass.
```
Sayt-0: Removing this TODO looks premature. `RunAgent` (background agents) sets `ToolsApproved: true`, and `Decide` (pkg/runtime/toolexec/permissions.go) returns Allow on the yolo flag before evaluating any checker. So the newly inherited parent Deny/Ask rules are never consulted for a background sub-session, which is precisely the bypass this TODO called out ("rather than a single shared ToolsApproved flag"). The warning just above ("could exploit this gate-bypass") still holds, so it now contradicts the TODO removal. Options: keep the TODO, or make Deny/ForceAsk win over the yolo flag in Decide (at least for an inherited session-level Deny) and cover it with a test.

pkg/runtime/agent_delegation_test.go
```diff
@@ -266,3 +266,28 @@ func TestSubSessionWithoutAttachedFilesOmitsBlock(t *testing.T) {
...
+		AgentName: "worker",
+		Title:     "Refactor",
+	}
+
+	s := newSubSession(parent, cfg, childAgent)
+
+	require.NotNil(t, s.Permissions)
+	assert.Equal(t, perms.Allow, s.Permissions.Allow)
+	assert.Equal(t, perms.Deny, s.Permissions.Deny)
+	assert.Equal(t, perms.Ask, s.Permissions.Ask)
```
Sayt-0, line 300: This checks field copying but not that inheritance changes dispatch behaviour. Consider driving the approval path instead: a parent Deny of `write_*` should still deny in the child, and in particular the background-agent case (`ToolsApproved: true`) where the inherited Deny currently has no effect. That case would document the gap flagged in agent_delegation.go.

Sayt-0 (commented):
Thanks for tackling this. Inheriting the parent permissions in newSubSession is the right insertion point. Two points look blocking before this can land, plus a test gap; details are inline.

| area | concern |
| --- | --- |
| correctness | background agents run with ToolsApproved: true, and Decide short-circuits on the yolo flag before any checker runs, so the inherited Allow/Ask/Deny rules are never consulted on that path (the exact case the removed TODO described) |
| isolation | the parent PermissionsConfig pointer is shared with the child instead of cloned, unlike every other call site |
| tests | the new test asserts field copying only, not dispatch behaviour |

Piyush0049:
@Sayt-0 Thank you for the review, I read it and have pushed an update addressing all three of your points:

1. **Isolation / Aliasing:** Added a public `.Clone()` method to `PermissionsConfig` and updated `newSubSession` to use it, ensuring child permissions don't alias the parent. (Also cleaned up the old unexported helper in `pkg/session`).
2. **Correctness (YOLO Flag):** Re-ordered the logic in `Decide()` so that explicit `Deny` and `ForceAsk` rules now properly win over the `ToolsApproved: true` (yolo) flag for background agents.
3. **Dispatch Test:** Updated `TestSubSessionInheritsPermissions` to explicitly drive the approval path, proving that a background agent (with the yolo flag on) correctly returns `OutcomeDeny` when inheriting a parent's denial.

All tests are passing locally. Do let me know if any change is required please.

Piyush0049:
I just pushed a quick follow-up commit to fix the CI failures.

Sayt-0:
HI @Piyush0049 can you resolve the conflicts ? ping me when its done

Piyush0049:
[Sayt-0](https://github.com/Sayt-0) Sure

Piyush0049:
[Sayt-0](https://github.com/Sayt-0) I have resolved the merge conflicts. Please do let me know if any changes are required.

Piyush0049:
I've pushed a commit to fix the CI timeout in `build-and-test`. 

The `TestForceAskOverridesYoloMode` test was hanging because it routed to `askUser()`, which blocked indefinitely waiting for UI input. Adding `sess.NonInteractive = true` allows it to correctly fail fast without hanging.

**Side Note:** While testing locally on Windows, I noticed a few pre-existing issues in `main` that break the local test suite (e.g. `syscall.Mkfifo` breaks compilation, hardcoded `/abs/foo.go` paths fail `filepath.IsAbs`, and `sh`/`jq` commands in hooks). Since these pass on Linux CI, I kept them out of this PR, but let me know if you'd like a separate PR later to make the tests Windows compatible

Piyush0049:
The `build-image` CI job probably failed due to a transient `apt-get` mirror sync error on `download.docker.com`. Could a maintainer please re-run the failed jobs when they get a chance?

Piyush0049:
Thanks for re-running the jobs. Please do let me know if any other changes are required.

pkg/runtime/toolexec/permissions.go
```diff
@@ -92,6 +88,10 @@ func Decide(
 		}
 	}
 
+	if yoloApproved {
```
Sayt-0, line 94: This reorders the approval pipeline for every session, not just sub-sessions with inherited permissions.

`Deny` winning over `--yolo` matches the contract `PermissionsConfig` has documented since it was introduced (`pkg/config/latest/types.go`, and every frozen version v3 to v11): "Deny: Tools matching these patterns are always rejected, even with --yolo".

`ForceAsk` winning over `--yolo` reverses commit `7351e4fa0` ("--yolo must accept any tool call"), which added `TestYoloMode_OverridesForceAsk` (renamed here to `TestForceAskOverridesYoloMode`, assertion flipped) specifically to lock in the opposite behavior, and documented it in `pkg/permissions/permissions.go`'s `CheckWithArgs` comment: "Note that --yolo mode takes precedence over ForceAsk" (still present on main, now stale).

The doc comment on this function, line 61, still lists yolo as step 1 ahead of the checkers; it needs updating either way this lands.

Question: should this only reorder the `Deny` case and leave `ForceAsk` behind `--yolo` (matching the two contracts above), or is overriding `ForceAsk` too an intentional posture change? If the latter, `pkg/permissions/permissions.go` and `docs/tools/background-agents/index.md:33` ("via YOLO mode or explicit allow rules") need updating in this PR.

pkg/runtime/toolexec/permissions_test.go
```diff
@@ -18,13 +18,13 @@ func newChecker(t *testing.T, allow, ask, deny []string) *permissions.Checker {
...
 
-func TestDecide_YoloShortCircuits(t *testing.T) {
+func TestDecide_DenyOverridesYolo(t *testing.T) {
 	t.Parallel()
 	d := Decide(true, []NamedChecker{
 		{Checker: newChecker(t, nil, nil, []string{"shell"}), Source: "team"},
 	}, "shell", nil, false)
 
-	assert.Equal(t, PermissionDecision{Outcome: OutcomeAllow, Reason: ReasonYolo}, d)
+	assert.Equal(t, PermissionDecision{Outcome: OutcomeDeny, Reason: ReasonChecker, Source: "team"}, d)
```
Sayt-0, line 27: Matches the documented `PermissionsConfig` contract for `Deny`, no objection to this one specifically. See the comment on `pkg/runtime/toolexec/permissions.go` about the related `ForceAsk` case that still needs a decision.

pkg/runtime/runtime_test.go
```diff
@@ -2561,11 +2562,12 @@ func TestYoloMode_OverridesPermissionsDeny(t *testing.T) {
...
 
-	// With --yolo, the tool should execute despite deny permission
-	require.True(t, executed, "expected tool to be executed in --yolo mode despite deny permission")
+	// With --yolo and Deny/ForceAsk precedence, the tool should NOT execute.
+	require.False(t, executed, "expected tool to NOT be executed in --yolo mode because Deny wins")
 }
 
-func TestYoloMode_OverridesForceAsk(t *testing.T) {
+// TestForceAskOverridesYoloMode verifies that ForceAsk permissions take precedence over the yolo flag.
+func TestForceAskOverridesYoloMode(t *testing.T) {
```
Sayt-0: This test (renamed from `TestYoloMode_OverridesForceAsk`) now expects the tool NOT to execute, same for `TestSessionDenyOverridesYoloMode` below. This is the `ForceAsk`-over-`--yolo` change flagged on `pkg/runtime/toolexec/permissions.go`, holding these pending that decision.

Separately, `sess.NonInteractive = true` on line 2602 fixing the CI hang (`askUser` blocking on a resume channel with no listener) is correct and worth keeping regardless of the outcome above.

pkg/runtime/agent_delegation.go
```diff
@@ -535,7 +535,8 @@ func (r *LocalRuntime) handleTaskTransfer(ctx context.Context, sess *session.Ses
 		delegationAttrs = append(delegationAttrs, attribute.Int("cagent.delegation.task_length", len(params.Task)))
 	}
 	if genai.EmitLegacyAttributes() {
-		delegationAttrs = append(delegationAttrs,
+		delegationAttrs = append(
```
Sayt-0: This hunk (and the `applyForceHandoff` closing-paren move further down) is pure reformatting unrelated to the fix. Confirmed with `golangci-lint fmt` that neither the old nor the new layout is enforced by the formatter, so this is incidental noise. Suggest reverting both to keep the diff focused.

pkg/runtime/agent_delegation_test.go
```diff
@@ -82,7 +85,8 @@ func TestNewSubSession(t *testing.T) {
 	t.Parallel()
 
 	parent := session.New(session.WithUserMessage("hello"))
-	childAgent := agent.New("worker", "a worker agent",
+	childAgent := agent.New(
```
Sayt-0: Same as the note on `agent_delegation.go`: the `agent.New(...)` reflow here (and 3 more occurrences: lines 194, 414, 471) is unrelated to the fix and adds review noise. Suggest reverting.

pkg/runtime/agent_delegation_test.go
```diff
@@ -270,6 +275,48 @@ func TestSubSessionWithoutAttachedFilesOmitsBlock(t *testing.T) {
...
+	checker := permissions.NewChecker(&latest.PermissionsConfig{
+		Allow: s.Permissions.Allow,
+		Ask:   s.Permissions.Ask,
+		Deny:  s.Permissions.Deny,
+	})
+	namedCheckers := []toolexec.NamedChecker{
+		{Checker: checker, Source: "session permissions"},
+	}
+
+	decision := toolexec.Decide(s.ToolsApproved, namedCheckers, "write_file", map[string]any{"path": "foo"}, false)
```
Sayt-0, line 314: Nit, not blocking: this builds a `permissions.Checker` by hand from `s.Permissions` and calls `toolexec.Decide` directly. That's a real improvement over asserting field copies only, but it still bypasses the actual wiring in `permissionCheckers` / `rt.processToolCalls`. Consider complementing it with an end-to-end case through `RunAgent`/`runCollecting` with a real background sub-session, similar to `TestDenyOverridesYoloMode`, so the real dispatcher path is covered too.

Sayt-0 (changes requested):
Thanks for the follow-up and for iterating on the earlier feedback. The isolation fix (`PermissionsConfig.Clone()`) is clean, and the race-prone double-cloning issue from the earlier round looks closed (verified locally with `go test -race ./pkg/runtime/... ./pkg/session/...`, no race, all green, matching CI).

One design point needs a decision before this can merge, plus a few smaller items, all left as inline comments.

| Area | Concern | Blocking |
| --- | --- | --- |
| Scope / design | `Decide()` reordering makes `Deny` and `ForceAsk` win over `--yolo` globally, not just for inherited sub-sessions | yes, needs a decision |
| Docs consistency | Several comments and one public doc page still describe the old ordering | yes |
| Diff hygiene | A few hunks are pure reformatting unrelated to the fix | no, nit |
| Test depth | New test exercises `Decide()` directly rather than the full dispatch path | no, nit |

**What's good:**
- `PermissionsConfig.Clone()`: exported, nil-safe, doc-commented, consistent with `Session.ClonePermissions()`.
- `TestSubSessionInheritsPermissions` is a real improvement over asserting field copies only.
- No new allocations or performance concerns; `slices.Clone` is equivalent to the helper it replaces.

pkg/runtime/toolexec/permissions.go
Piyush0049: It was not an intentional posture change! I have updated the logic to only reorder the `Deny` case, leaving `ForceAsk` behind `--yolo` as originally intended. 

Because this preserves the original contract, the external docs (`pkg/permissions/permissions.go` and `index.md`) remain accurate and didn't need updating. (I've also fixed the doc comment on `Decide()` to reflect this correctly).

Piyush0049:
Thanks for the review. I read it and have addressed all the points in the table:
1. **Scope/design**: Decided to revert the `ForceAsk` change. It now stays behind `--yolo` as originally intended, while `Deny` correctly continues to override it.
2. **Docs consistency**: Updated the `Decide()` block comment. Since the original contract for `ForceAsk` was restored, the external docs are still accurate and were left untouched.
3. **Diff hygiene**: Reverted the formatting noise in `agent_delegation.go` (both the `EmitLegacyAttributes` hunk and the `applyForceHandoff` closing paren) and the `agent.New()` reflows in the test file. 
4. **Test depth**: Fixed the expectation on `TestYoloMode_OverridesForceAsk` (kept the CI hang fix) and added a brand new `TestRunAgent_EndToEndPermissions` to verify the `permissions.Checker` end-to-end through the real `RunAgent` dispatcher path.

Please do let me know if any other changes are required.

pkg/runtime/runtime_test.go
```diff
@@ -2517,7 +2517,8 @@ func TestTransferTaskPersistsSubSessionOnError(t *testing.T) {
 		"SubSessionCompletedEvent must fire on the error path so observers persist the sub-session")
 }
 
-func TestYoloMode_OverridesPermissionsDeny(t *testing.T) {
+// TestDenyOverridesYoloMode verifies that Deny permissions take precedence over the yolo flag.
+func TestDenyOverridesYoloMode(t *testing.T) {
 	t.Parallel()
 
 	// Test that --yolo flag takes precedence over deny permissions
```
Sayt-0: Stale: this contradicts the renamed test and its assertion (Deny now wins over yolo). The doc comment above the function already states the behavior, so this line can go.

```suggestion
```

pkg/runtime/runtime_test.go
```diff
@@ -2609,11 +2612,13 @@ func TestYoloMode_OverridesForceAsk(t *testing.T) {
...
+	// and the tool executes automatically.
+	require.True(t, executed, "expected tool to be executed in --yolo mode because YOLO wins over ForceAsk")
 }
 
-func TestYoloMode_OverridesSessionDeny(t *testing.T) {
+// TestSessionDenyOverridesYoloMode verifies that session-level Deny permissions take precedence over the yolo flag.
+func TestSessionDenyOverridesYoloMode(t *testing.T) {
 	t.Parallel()
 
 	// Test that --yolo flag takes precedence over session-level deny
```
Sayt-0: Same as above: stale comment stating the pre-PR behavior under a test that now asserts the opposite.

```suggestion
```

pkg/runtime/agent_delegation_test.go
```diff
@@ -409,6 +456,60 @@ func TestRunAgent_InheritsParentPermissions(t *testing.T) {
...
+		session.WithUserMessage("Test"),
+		session.WithToolsApproved(true),
+		session.WithPermissions(parentPerms),
+	)
+
+	rt.RunAgent(t.Context(), agenttool.RunParams{
+		AgentName:     "worker",
+		Task:          "do something",
+		ParentSession: parentSession,
+	})
```
Sayt-0: The result is discarded, so `require.False(executed)` would pass vacuously if `RunAgent` failed before dispatch (wrong agent name, model error, ...). Asserting the result and the sub-session keeps the test honest. Verified locally, this exact block passes on this branch:

```suggestion
	result := rt.RunAgent(t.Context(), agenttool.RunParams{
		AgentName:     "worker",
		Task:          "do something",
		ParentSession: parentSession,
	})
	require.Empty(t, result.ErrMsg, "RunAgent should succeed")

	var childSession *session.Session
	for _, item := range parentSession.Messages {
		if item.SubSession != nil {
			childSession = item.SubSession
			break
		}
	}
	require.NotNil(t, childSession, "parent must have a sub-session")
```

pkg/runtime/agent_delegation_test.go
```diff
@@ -270,6 +274,48 @@ func TestSubSessionWithoutAttachedFilesOmitsBlock(t *testing.T) {
...
+
+	s := newSubSession(parent, cfg, childAgent)
+
+	require.NotNil(t, s.Permissions)
+	assert.Equal(t, perms.Allow, s.Permissions.Allow)
+	assert.Equal(t, perms.Deny, s.Permissions.Deny)
+	assert.Equal(t, perms.Ask, s.Permissions.Ask)
+
+	// Verify the gap flagged in PR review: even if ToolsApproved is true (yolo flag),
+	// the inherited Deny should correctly override the yolo flag during dispatch.
```
Sayt-0: Nit: "the gap flagged in PR review" will not mean much to future readers; a self-contained phrasing ages better.

```suggestion
	// Even with ToolsApproved set (yolo), an inherited Deny must win during dispatch.
```

pkg/runtime/toolexec/permissions_test.go
```diff
@@ -18,13 +18,13 @@ func newChecker(t *testing.T, allow, ask, deny []string) *permissions.Checker {
 	})
 }
 
-func TestDecide_YoloShortCircuits(t *testing.T) {
+func TestDecide_DenyOverridesYolo(t *testing.T) {
```
Sayt-0, line 21: Nit: after the rename, two paths of the new ordering have no `Decide`-level case left: plain yolo-allow (no checker match) and yolo-over-ForceAsk. Two one-liners would restore the exhaustive matrix promised by `Decide`'s doc comment (both verified locally):

```go
func TestDecide_YoloAllowsWhenNoCheckerMatches(t *testing.T) {
	t.Parallel()
	d := Decide(true, nil, "shell", nil, false)
	assert.Equal(t, PermissionDecision{Outcome: OutcomeAllow, Reason: ReasonYolo}, d)
}

func TestDecide_YoloOverridesForceAsk(t *testing.T) {
	t.Parallel()
	d := Decide(true, []NamedChecker{
		{Checker: newChecker(t, nil, []string{"shell"}, nil), Source: "team"},
	}, "shell", nil, false)
	assert.Equal(t, PermissionDecision{Outcome: OutcomeAllow, Reason: ReasonYolo}, d)
}
```

pkg/runtime/runtime_test.go
```diff
@@ -2597,6 +2599,7 @@ func TestYoloMode_OverridesForceAsk(t *testing.T) {
 	require.NoError(t, err)
 
 	sess := session.New(session.WithUserMessage("Test"), session.WithToolsApproved(true))
+	sess.NonInteractive = true
```
Sayt-0: Nit: with yolo winning over ForceAsk, `askUser` is never reached, so this line is purely defensive (fail fast instead of hanging CI if the behavior regresses). Worth one short comment, matching the style of the `// mirror --exec mode` note at line 2466.

```suggestion
	sess.NonInteractive = true // fail fast instead of hanging on askUser if this regresses
```

pkg/runtime/agent_delegation_test.go
```diff
@@ -364,7 +410,8 @@ func TestRunAgent_InheritsParentPermissions(t *testing.T) {
 	agent.WithSubAgents(worker)(root)
 
 	tm := team.New(team.WithAgents(root, worker))
-	rt, err := NewLocalRuntime(t.Context(), tm,
+	rt, err := NewLocalRuntime(
+		t.Context(), tm,
```
Sayt-0: Nit: reformat-only hunk, no semantic change. Reverting keeps the diff focused on the fix (same below at line 524).

```suggestion
	rt, err := NewLocalRuntime(t.Context(), tm,
```

pkg/runtime/agent_delegation_test.go
```diff
@@ -420,7 +521,8 @@ func TestTransferTask_PropagatesPermissions(t *testing.T) {
 	agent.WithSubAgents(librarian)(root)
 
 	tm := team.New(team.WithAgents(root, librarian))
-	rt, err := NewLocalRuntime(t.Context(), tm,
+	rt, err := NewLocalRuntime(
+		t.Context(), tm,
```
Sayt-0: Same reformat-only hunk as above.

```suggestion
	rt, err := NewLocalRuntime(t.Context(), tm,
```

Sayt-0 (changes requested):
Thanks for iterating again. The design question from the previous round is now settled: making explicit `Deny` win over `--yolo` globally is accepted. It matches the documented "Deny patterns take priority" philosophy, and background agents inheriting `ToolsApproved: true` would otherwise carry dead-letter Deny rules. Keeping YOLO above `ForceAsk` restores the original contract, good call.

Verified locally on the branch: `go build ./...` and `go test -race ./pkg/runtime/... ./pkg/session/...` are green.

Remaining items before merge, all small; details inline.

| # | Area | Concern | Blocking |
| --- | --- | --- | --- |
| 1 | Stale comments | Two inline comments in `runtime_test.go` still state the old behavior ("--yolo takes precedence over deny") under tests that now assert the opposite | yes |
| 2 | Docs consistency | Three doc lines still describe the old chain order (table below) | yes |
| 3 | Test robustness | `TestRunAgent_EndToEndPermissions` discards the `RunAgent` result, so it would pass vacuously if the run failed before any tool call | yes |
| 4 | PR title/description | The `newSubSession` inheritance described in the PR body already landed on `main` via fd89580b; the merge commit would tell the wrong story. Refresh both to describe the actual change (`Decide` precedence + `Clone()` refactor) | yes |
| 5 | Diff hygiene | Two `NewLocalRuntime(` reflow-only hunks remain in `agent_delegation_test.go` | no, nit |
| 6 | Test coverage | After the rename, no `Decide`-level case covers plain yolo-allow or yolo-over-ForceAsk | no, nit |

Docs lines to update (files not in the diff, need a commit):

| File | Line | Current | Problem |
| --- | --- | --- | --- |
| `docs/configuration/hooks/index.md` | 60 | "approval chain (yolo / permissions / readonly / ask)" | Deny permissions now precede yolo |
| `docs/configuration/hooks/index.md` | 716 | "tool-approval chain (yolo / permissions / readonly / ...)" | same |
| `docs/features/cli/index.md` | 31 | "`--yolo`: Auto-approve all tool calls" | no longer "all": explicit deny rules still block |

Non-blocking observation, for the record: with first-checker-wins semantics, a session-level `ForceAsk` consulted before a team-level `Deny` still yields Allow under yolo. This is pre-existing layer-precedence behavior, unchanged by this PR.

What's good:
- The `Decide()` reordering is minimal and keeps the function pure; returning `ReasonYolo` on the ForceAsk-under-yolo path keeps `approval_source` audit labels consistent for hooks.
- `PermissionsConfig.Clone()` is nil-safe and removes the private helper cleanly across all six call sites.
- `TestRunAgent_EndToEndPermissions` finally exercises the real dispatch path, closing the test-depth gap from round one.

Piyush0049:
Thank you for the review. I read it and have pushed an update addressing your feedback:

- Fixed the stale test comments and updated the docs to correctly reflect that `Deny` rules precede `--yolo`.
- Removed the upward permission back-propagation bug in `agent_delegation.go`.
- Added the missing YOLO paths to the `Decide` test matrix.
- Fixed the `TestRunAgent_EndToEndPermissions` assertions and reverted the formatting noise.
- The PR title and description have also been updated to reflect the `Decide` precedence and bug fixes.

Piyush0049:
I have also updated the PR title and description according to the review. Please do let me know if any other change is required.

pkg/runtime/agent_delegation.go
```diff
@@ -322,12 +322,6 @@ func (r *LocalRuntime) runForwarding(ctx context.Context, parent *session.Sessio
 		return nil, subSessionErr
 	}
 
-	// Only propagate ToolsApproved and Permissions on success. A failed sub-session
-	// must not silently escalate the parent's tool-approval gate: the user approved
-	// tools within a sub-session scope that ended in error, and that approval
-	// should not carry over to the parent's remaining turns.
-	parent.SetToolsApproved(s.IsToolsApproved())
-	parent.SetPermissions(s.ClonePermissions())
 	span.SetStatus(codes.Ok, "sub-session completed")
```
Sayt-0, line 325: This removal is a real behavior change, not a cleanup, and it was not among the previous round's asks. The success-path back-propagation was introduced deliberately in 838f7ddd ("Update the tools approved of the parent session once the sub-session ended") and recently narrowed to the success path on main (fd89580b) with an explicit rationale comment, the one deleted here.

What changes for interactive users:

| Action inside a `transfer_task` / `run_skill` child | before | after |
|---|---|---|
| "Approve all for session" | carries over to the parent's remaining turns | scoped to the child; the parent re-prompts |
| "Always allow `<tool>`" | appended to the parent's Allow list | scoped to the child |
| Failed child session | no propagation | no propagation (unchanged) |

The security argument is sound: it matches the "permissions only flow downwards" scoping this PR establishes, and it closes the ratchet where a child scope elevates the parent for the rest of the turn. The cost is UX: users who click "approve all" inside a delegation will be prompted again in the parent.

Accepting the removal is reasonable, under two conditions (tracked in the summary table):
1. the `runForwarding` doc comment (line 251, "propagating tool-approval state back to the parent on completion") must state the new contract instead, e.g. "The child's approval state stays scoped to the sub-session and never flows back to the parent";
2. a test must lock the invariant, otherwise the next reader may take this for an accidental drop and restore it; see the proposal on `agent_delegation_test.go`.

If keeping the "approve all persists to the parent" UX is preferred instead, this hunk can be reverted and moved to its own PR so the `Decide` precedence fix is not held up.

pkg/runtime/agent_delegation_test.go
```diff
@@ -409,6 +454,70 @@ func TestRunAgent_InheritsParentPermissions(t *testing.T) {
 		"parent permissions must be isolated from child mutations")
 }
 
+func TestRunAgent_EndToEndPermissions(t *testing.T) {
```
Sayt-0, line 510: The new no-back-propagation invariant has no locking test: no existing test fails if `parent.SetToolsApproved(...)` / `parent.SetPermissions(...)` come back. The following drives the real `runForwarding` path with a child scope broader than the parent's (as if the user had clicked "approve all" inside the sub-session) and asserts the parent is untouched. Verified: passes on d12ae3fc, fails on the parent commit 38c27a8c (pre-removal code) on both assertions.

```go
// TestRunForwarding_DoesNotBackPropagateApprovals locks the "permissions only
// flow downwards" invariant: approvals granted within a sub-session scope must
// not escalate the parent's ToolsApproved gate or permission rules.
func TestRunForwarding_DoesNotBackPropagateApprovals(t *testing.T) {
	t.Parallel()

	childStream := newStreamBuilder().AddContent("done").AddStopWithUsage(10, 5).Build()
	prov := &mockProvider{id: "test/mock-model", stream: childStream}

	librarian := agent.New("librarian", "Library agent", agent.WithModel(prov))
	root := agent.New("root", "Root agent", agent.WithModel(prov))
	agent.WithSubAgents(librarian)(root)

	tm := team.New(team.WithAgents(root, librarian))
	rt, err := NewLocalRuntime(t.Context(), tm,
		WithSessionCompaction(false),
		WithModelStore(mockModelStore{}),
	)
	require.NoError(t, err)

	parent := session.New(
		session.WithUserMessage("Test"),
		session.WithPermissions(&session.PermissionsConfig{Deny: []string{"dangerous_tool"}}),
	)
	require.False(t, parent.IsToolsApproved())

	evts := make(chan Event, 128)
	// Child scope broader than the parent's, as if the user had clicked
	// "approve all" / "always allow" inside the sub-session.
	_, err = rt.runForwarding(t.Context(), parent, NewChannelSink(evts), delegationRequest{
		SubSessionConfig: SubSessionConfig{
			Task:          "find a book",
			AgentName:     "librarian",
			Title:         "Transferred task",
			ToolsApproved: true,
			Permissions: &session.PermissionsConfig{
				Allow: []string{"exploit_tool"},
				Deny:  []string{"dangerous_tool"},
			},
		},
		SwitchCurrentAgent: true,
	})
	require.NoError(t, err)

	assert.False(t, parent.IsToolsApproved(),
		"a sub-session must not escalate the parent's ToolsApproved gate")
	parentPerms := parent.ClonePermissions()
	require.NotNil(t, parentPerms)
	assert.Empty(t, parentPerms.Allow,
		"child-scope approvals must not leak into the parent's Allow list")
	assert.Equal(t, []string{"dangerous_tool"}, parentPerms.Deny)
}
```

pkg/runtime/agent_delegation_test.go
```diff
@@ -409,6 +454,70 @@ func TestRunAgent_InheritsParentPermissions(t *testing.T) {
...
+	var childSession *session.Session
+	for _, item := range parentSession.Messages {
+		if item.SubSession != nil {
+			childSession = item.SubSession
+			break
+		}
+	}
+	require.NotNil(t, childSession, "parent must have a sub-session")
+
+	require.False(t, executed, "expected dangerous_tool to NOT be executed because it is denied by inherited permissions")
```
Sayt-0, line 573: Nit: `childSession` is fetched but only nil-checked. Asserting the inherited Deny actually reached the child makes the `executed == false` assertion self-explanatory (denied because inherited, not because of an unrelated failure). Verified locally, the suggestion passes on this branch.

```suggestion
	require.NotNil(t, childSession, "parent must have a sub-session")
	require.NotNil(t, childSession.Permissions)
	assert.Equal(t, []string{"dangerous_tool"}, childSession.Permissions.Deny,
		"child must inherit the parent's Deny rules")

	require.False(t, executed, "expected dangerous_tool to NOT be executed because it is denied by inherited permissions")
```

Sayt-0 (changes requested):
Thanks for the follow-up. All six items from the previous round are addressed: the stale `runtime_test.go` comments are gone, the three doc lines now describe the `permissions / yolo` order, `TestRunAgent_EndToEndPermissions` asserts the run result, the reflow-only hunks are reverted, the `Decide` matrix is exhaustive again, and the PR title/body were refreshed.

Verified locally on the branch head (d12ae3fc): `go build ./...` and `go test -race ./pkg/runtime/... ./pkg/session/...` are green.

The last push, however, adds a behavior change that was not among the previous round's asks: the removal of the success-path back-propagation in `runForwarding`. It needs an explicit decision plus two follow-ups; details inline.

| # | Area | Concern | Blocking |
|---|---|---|---|
| 1 | Scope / design | Removing the `runForwarding` back-propagation reverses a deliberate behavior (838f7ddd); the direction fits the "permissions flow downwards" philosophy, but the UX change must be decided explicitly rather than land silently in a follow-up push | yes, needs a decision |
| 2 | Stale doc comment | `runForwarding`'s doc comment still says "propagating tool-approval state back to the parent on completion" | yes |
| 3 | Test coverage | No test locks the new no-back-propagation invariant; a ready-made test is proposed inline (passes on this branch, fails on the pre-removal code) | yes |
| 4 | Test depth | `TestRunAgent_EndToEndPermissions` nil-checks `childSession` but never inspects it | no, nit |

What's good:
- `TestDecide_YoloAllowsWhenNoCheckerMatches` and `TestDecide_YoloOverridesForceAsk` restore the exhaustive matrix promised by `Decide`'s doc comment, asserting exact `PermissionDecision` values.
- The doc updates land exactly where round 3 asked (hooks page twice, CLI page twice, including the alias bullet).
- `TestRunAgent_EndToEndPermissions` now asserts `result.ErrMsg` and the sub-session presence, closing the vacuous-pass gap.
- The diff is free of reformat-only hunks.

Piyush0049:
@Sayt-0 Thank you for the review, I read it and have pushed a commit addressing :

* Updated the runForwarding doc comment to explicitly state that the approval state stays scoped to the sub-session and never flows back up.
* Added the TestRunForwarding_DoesNotBackPropagateApprovals test to lock the no-back-propagation invariant.
* Added the explicit Deny list assertion to TestRunAgent_EndToEndPermissions.

Please do let me know if any other change is required please.

Sayt-0 (dismissed):
Looks good now ! Thank you !

Sayt-0:
Can you sign your commits ?

Piyush0049:
@Sayt-0  Thank you for the approval.
I have just rebased the branch to properly sign off the remaining commits.

Sayt-0:
Hi @Piyush0049 ! all commits must be signed. Can you take a look and ping me when it's done ?

Piyush0049:
@Sayt-0 Now all the commits are verified. I actually missed one commit last time. I hope everything is fine now.
Please do let me know if any other changes are required.

Sayt-0 (approved):

Sayt-0:
good now ! thank you !

PR #3543, opened 2026-07-08
style(tui): polish agent picker layout, cards, and status bars

dgageot:
The agent picker had a handful of rough edges: the status bars used a hand-rolled key rendering path while every other dialog had already moved to `dialog.RenderHelpKeys`, cards were taller than necessary, the panel was wider than it needed to be, and the selected card had no visual accent to make it stand out.

This change brings the picker in line with the rest of the TUI. The main help line and the YAML details dialog both now go through `dialog.RenderHelpKeys`, which removes duplicated logic and gives them the same look as other dialogs. A `fitHelpPairs` helper was extracted so that on narrow terminals trailing bindings are dropped cleanly rather than soft-wrapping and distorting the layout. Card height was reduced from 7 to 5 rows by removing vertical padding, and the panel was narrowed from ~99 to ~83 columns with its title, subtitle, and status bar centered. The redundant "double-click select" binding was dropped. The selected card now gets a thick accent border, the Lean Mode label is rendered brighter when ticked, and the Open Board entry is rendered as a button-like control.

Mouse hit-testing invariants (`panelSize == render geometry`) are preserved throughout; existing tests were updated and new regression tests for `fitHelpPairs` and hit-testing were added.

### DIFF
--- a/cmd/root/agent_picker.go
+++ b/cmd/root/agent_picker.go
@@ -757,12 +757,21 @@ func (m *agentPickerModel) headerText() (title, subtitle string, helpPairs []str
 	if m.width > 0 {
 		maxWidth := max(m.width-2*(1+4), agentPickerMinCardWidth)
 		subtitleText = toolcommon.TruncateText(subtitleText, maxWidth)
-		for len(helpPairs) > 2 && dialog.HelpKeysWidth(helpPairs...) > maxWidth {
-			helpPairs = helpPairs[:len(helpPairs)-2]
-		}
+		helpPairs = fitHelpPairs(helpPairs, maxWidth)
 	}
 	subtitle = styles.MutedStyle.Render(subtitleText)
 	return title, subtitle, helpPairs
+}
+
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The diff was reviewed for bugs introduced by the changed `+` lines. No issues were found.

**Scope reviewed:**
- `fitHelpPairs` helper logic (narrow-terminal truncation)
- Card height invariants (7 → 5 row change) and panel width (~99 → ~83)
- Mouse hit-testing math consistency with render geometry
- `dialog.RenderHelpKeys` integration in status bars and YAML details dialog
- Accent border rendering for selected card
- Lean Mode / Open Board button rendering
- New regression tests for `fitHelpPairs` and hit-testing

The refactor cleanly removes duplicated key-rendering logic, preserves hit-testing invariants, and the new tests provide good regression coverage.

aheritier (approved):

PR #3549, opened 2026-07-09
feat(teamloader): add WithModelOptions passthrough for provider-agnostic HTTP transport wrapping

simonferquel-clanker:

## Problem

`teamloader.Load`/`LoadWithConfig` has no way for a caller to inject a custom `options.Opt` into the model clients it constructs internally. In particular, `options.WithHTTPTransportWrapper` (#3089) already lets a caller wrap the HTTP transport used by anthropic/openai/gemini provider clients — e.g. to inject a bearer token provider-agnostically, decoupled from `environment.IsTrustedDockerURL` — but that only works today for callers who build a provider client directly (e.g. `agentic-platform`'s `internal/gordon`). Any caller going through `teamloader.Load` (which resolves providers dynamically from a cagent.yml config, so it can't build clients directly) has no equivalent hook.

This surfaced while investigating a credential-injection gap in `agentic-platform`'s autonomous sandbox-runner sessions (tracked there as APT-632): the runner uses `teamloader.Load`, so it can't currently apply a transport wrapper for gateway auth across arbitrary configured providers.

## Change

- Adds `WithModelOptions(opts ...options.Opt) Opt`, appended after teamloader's own built-in opts (`options.WithGateway`, `options.WithStructuredOutput`, etc.) at all 4 internal provider-construction sites (`getModelsForAgent`, `getFallbackModelsForAgent`, `getTitleModelForAgent`, `getCompactionModelForAgent`), so caller-supplied opts take precedence for anything both sides set.
- Threads it into `loadExternalAgent` so external (OCI/URL-referenced) sub-agents inherit it too.
- Adds an end-to-end regression test (`TestWithModelOptions_TransportWrapperReachesAgentModel`): loads a team from an inline cagent config via `WithModelOptions(options.WithHTTPTransportWrapper(...))`, drives the resulting agent's model through a real `httptest` streaming call, and asserts the wrapper's `RoundTrip` was invoked — proving the opt reaches the actual HTTP client, not just that it compiles. (Confirmed by reverting only `teamloader.go`: the test then fails to compile with `undefined: WithModelOptions`.)

## Testing

- `go build ./pkg/teamloader/...`, `go vet ./pkg/teamloader/...`, `go test ./pkg/teamloader/...` — all pass.
- `go mod tidy --diff` — empty.
- `go run ./lint .` — no offenses (1491 files).
- `golangci-lint v2.12.2 run ./pkg/teamloader/...` — 0 issues.
- Independently reviewed by a sub-agent; verdict: approve. One non-blocking note: the new test only exercises the primary-model site end-to-end (the other 3 sites apply the identical one-line append, verified by code reading); happy to extend coverage if maintainers want it before merge.

### DIFF
--- a/pkg/teamloader/teamloader.go
+++ b/pkg/teamloader/teamloader.go
@@ -46,6 +46,7 @@ type loadOptions struct {
 	promptFiles      []string
 	toolsetRegistry  ToolsetRegistry
 	providerRegistry *provider.Registry
+	modelOpts        []options.Opt
 }
 
 type Opt func(*loadOptions) error
@@ -80,6 +81,23 @@ func WithProviderRegistry(registry *provider.Registry) Opt {
 		if registry != nil {
 			opts.providerRegistry = registry
 		}
+		return nil
+	}
+}
+
+// WithModelOptions appends caller-supplied [options.Opt] values to every model
+// client teamloader constructs for this load: primary, fallback, title, and
+// compaction models, as well as models built while loading external
+// (OCI/URL-referenced) sub-agents. Use this to thread cross-cutting model
+// configuration — most notably options.WithHTTPTransportWrapper, which lets an
+// embedder authenticate every outbound LLM request (regardless of provider)
+// without depending on provider-specific environment variables or
+// environment.IsTrustedDockerURL. The opts are appended after teamloader's own
+// built-in opts (options.WithGateway, options.WithStructuredOutput, etc.), so
+// they take precedence for any option that both sides set.
+func WithModelOptions(opts ...options.Opt) Opt {
+	return func(o *loadOptions) error {
+		o.modelOpts = append(o.modelOpts, opts...)
 		return nil
 	}
 }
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `WithModelOptions` passthrough is mechanically correct. The new `modelOpts []options.Opt` field is safely threaded through all four provider-construction sites (`getModelsForAgent`, `getFallbackModelsForAgent`, `getTitleModelForAgent`, `getCompactionModelForAgent`) and into `loadExternalAgent`. Option ordering (caller opts appended after built-ins) is intentional and correctly documented. Nil/empty-slice edge cases are safe — variadic `append(opts, modelOpts...)` on an empty/nil slice is a no-op in Go.

The regression test (`TestWithModelOptions_TransportWrapperReachesAgentModel`) correctly validates end-to-end that the wrapper reaches the actual HTTP client: the `countingTransport` closure is invoked once at provider construction time (the documented semantics of `WithHTTPTransportWrapper`), so `counter.base` is always populated before any `RoundTrip` call. The `httptest`-driven streaming path and assertion on `counter.calls.Load()` are sound.

No bugs were found in the changes introduced by this PR.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The `WithModelOptions` passthrough is correctly implemented: caller-supplied opts are appended after teamloader's built-in opts at all four provider-construction sites (`getModelsForAgent`, `getFallbackModelsForAgent`, `getTitleModelForAgent`, `getCompactionModelForAgent`) and correctly threaded through `loadExternalAgent` for external sub-agents. The `append` semantics are sound — no aliasing hazards. The end-to-end regression test exercises the full `Load → agent → provider → HTTP` path and correctly asserts that the transport wrapper is invoked. No bugs introduced by this PR were confirmed.

dgageot (approved):

PR #3550, opened 2026-07-09
feat(config): hint when a config key requires a newer schema version

dgageot:
When a config file fails strict YAML parsing, the error message today gives no guidance on *why* — it just says the key is unknown. This is particularly confusing when someone copies a snippet from the docs that uses a feature introduced in a newer schema version, because the fix is simple (bump the top-level `version` field) but invisible from the error alone.

This change adds a hint in that failure path. After a strict parse fails, the code probes each parser for schema versions above the one declared in the file, finds the smallest version that successfully parses the document, and appends a message like:

```text
hint: this key is supported by config version 11; update the top-level 'version' field (currently 10)
```

The probing is cheap — it only runs when strict parsing has already failed — and is gated on finding a strictly higher version that actually accepts the input, so it stays silent when the error is a genuine typo or unsupported key. A follow-up commit adds a numeric version ordering test and expands the `newerVersionHint` doc comment.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 
 	"github.com/goccy/go-yaml"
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `newerVersionHint` implementation is clean and correct. The logic correctly:
- Gates on `*yaml.UnknownFieldError` (so type mismatches and genuine typos stay silent)
- Compares version numbers numerically (not lexicographically), preventing `"9" > "10"`-style ordering bugs
- Picks the smallest accepting version, giving users the most precise upgrade target
- Only runs on the already-failed error path — no performance impact on the happy path

The doc comment honestly acknowledges the best-effort nature of the hint (a newer parser accepting the doc for unrelated reasons). Test coverage via `TestLoadUnknownKeyFromNewerVersionHint`, `TestLoadUnknownKeyNoNewerVersionNoHint`, and `TestLoadNewerVersionHintNumericOrdering` is solid for the key scenarios.

No bugs or significant concerns found in the changed code.

Sayt-0 (approved):

PR #3551, opened 2026-07-09
fix(sqlite): avoid WAL journal mode inside docker sandboxes

dgageot:
When `~/.cagent` is bind-mounted into a Docker sandbox VM (virtiofs), SQLite's WAL journal mode breaks down. WAL relies on a shared-memory index (the `-shm` file, mmap'd by every connection) that is not coherent across the VM boundary — SQLite itself documents WAL as unsafe on such filesystems — so any database opened in WAL mode inside a sandbox is prone to corruption.

`sqliteutil.OpenDB` now picks the journal mode at open time via a `journalMode()` helper: WAL on a regular host, DELETE when running inside a Docker sandbox (`SANDBOX_VM_ID` set — the same signal `environment.InSandbox()` uses, read directly to keep `sqliteutil` dependency-free). DELETE mode relies only on ordinary file locks that virtiofs forwards correctly. Opening an existing WAL database in DELETE mode automatically folds any leftover `-wal` file back into the main database. Pragma ordering keeps `busy_timeout` first so the WAL→DELETE demotion benefits from the 5s timeout; `PRAGMA wal_checkpoint` on a DELETE-mode database is a harmless no-op, so `CheckpointAndClose` is unaffected.

New tests verify WAL on the host, DELETE inside a simulated sandbox, and clean conversion of a host-created WAL database (data intact, mode flipped, no `-wal` file left behind). The existing in-memory WAL test now pins `SANDBOX_VM_ID=""` so it passes when the suite itself runs inside a sandbox. A cross-reference comment on `sandboxVMIDEnv` in `pkg/environment` prevents the duplicated env-var literal from drifting.

### DIFF
--- a/pkg/environment/sandbox.go
+++ b/pkg/environment/sandbox.go
@@ -2,6 +2,8 @@ package environment
 
 import "os"
 
+// sandboxVMIDEnv is also read directly by pkg/sqliteutil (journalMode) to
+// avoid a heavy import; update both sites if the detection signal changes.
 const sandboxVMIDEnv = "SANDBOX_VM_ID"
 
 // InSandbox reports whether docker agent is running inside a Docker sandbox.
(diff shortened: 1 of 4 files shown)


dgageot:
Superseded by #3552 (same change, opened from fork).

PR #3552, opened 2026-07-09
fix(sqlite): avoid WAL journal mode inside docker sandboxes

dgageot:
## Problem

docker-agent opens all its SQLite databases (`session.db`, memory, RAG, TUI state) through `sqliteutil.OpenDB` with `journal_mode(WAL)`. When `~/.cagent` is bind-mounted into a Docker sandbox VM (virtiofs), WAL's shared-memory index (the `-shm` file, mmap'd by every connection) is **not coherent across the VM boundary** — SQLite documents WAL as unsafe on such filesystems — so mounted databases frequently end up corrupted.

## Fix

`OpenDB` now picks the journal mode via a `journalMode()` helper:

- **WAL** normally (unchanged behavior on the host)
- **DELETE** when running inside a Docker sandbox (`SANDBOX_VM_ID` set — the same signal `environment.InSandbox()` uses, read directly to keep `sqliteutil` dependency-free)

DELETE mode relies only on ordinary file locks that the bind mount forwards correctly, and opening an existing WAL database in DELETE mode automatically folds any leftover `-wal` file back into the main database.

Notes:

- Pragma ordering keeps `busy_timeout` first so the WAL→DELETE demotion benefits from the 5s busy timeout.
- `PRAGMA wal_checkpoint` on a DELETE-mode DB is a harmless no-op, so `CheckpointAndClose` is unaffected.

## Tests

- New tests verify WAL on host, DELETE inside a sandbox, and clean conversion of a host-created WAL database (data intact, persistent mode flipped, no `-wal` file).
- The existing memory-db WAL test now pins `SANDBOX_VM_ID=""` so it also passes when the suite runs inside a sandbox.
- Added a cross-reference comment on `sandboxVMIDEnv` in `pkg/environment` so the duplicated env-var literal doesn't drift.

## Validation

`task lint`, `task build`, and tests for all sqlite consumers (`sqliteutil`, `session`, `memory`, `tuistate`, `rag`) pass.

### DIFF
--- a/pkg/environment/sandbox.go
+++ b/pkg/environment/sandbox.go
@@ -2,6 +2,8 @@ package environment
 
 import "os"
 
+// sandboxVMIDEnv is also read directly by pkg/sqliteutil (journalMode) to
+// avoid a heavy import; update both sites if the detection signal changes.
 const sandboxVMIDEnv = "SANDBOX_VM_ID"
 
 // InSandbox reports whether docker agent is running inside a Docker sandbox.
--- a/pkg/memory/database/sqlite/sqlite_test.go
+++ b/pkg/memory/database/sqlite/sqlite_test.go
@@ -344,6 +344,10 @@ func TestDatabaseOperationsWithCanceledContext(t *testing.T) {
 }
 
 func TestMemoryDatabaseUsesWALAndBusyTimeout(t *testing.T) {
+	// WAL is only expected outside a Docker sandbox (sqliteutil falls back
+	// to DELETE inside one), so pin the detection env var.
+	t.Setenv("SANDBOX_VM_ID", "")
+
 	db := setupTestDB(t)
 	memDB := db.(*MemoryDatabase)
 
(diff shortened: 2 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is well-scoped and correctly addresses the WAL corruption issue in Docker sandbox (virtiofs) environments.

**Changes reviewed:**
- `pkg/environment/sandbox.go` — cross-reference comment on `sandboxVMIDEnv` so the duplicated env-var literal is discoverable
- `pkg/sqliteutil/sqlite.go` — `journalMode()` helper reads `SANDBOX_VM_ID` directly (no heavy import), DSN construction is correct, pragma ordering (`busy_timeout` first) is correct
- `pkg/sqliteutil/sqlite_test.go` / `pkg/memory/database/sqlite/sqlite_test.go` — tests cover WAL on host, DELETE inside sandbox, WAL→DELETE conversion with data integrity, and the existing memory-db WAL test pins `SANDBOX_VM_ID=""` for sandbox-host parity

No bugs or regressions were identified in the introduced code.

Sayt-0 (approved):

PR #3556, opened 2026-07-09
feat(board): let the user configure their own columns

dgageot:
The `docker agent board` Kanban view previously used a fixed set of columns
that could not be changed without editing the raw config file. This adds a
first-class, in-TUI columns manager so teams can name, reorder, and tune each
column's LLM prompt without ever leaving the board.

Pressing `c` opens the columns dialog. From there you can add a column,
edit its name or emoji inline, reorder it with shift+↑/↓, open the full
prompt editor with `p`, or remove it (guarded by a confirmation step that
also blocks deletion when cards remain or only one column exists). Renames
preserve the column's stable id, so existing cards stay attached. Restoring
the default pipeline removes the custom columns section from the config entirely.
All changes are persisted to the global user config immediately. Config loading
is also hardened: ids are derived from names when absent, duplicates are dropped,
and defaults are cloned before mutation — fixing a latent bug where editing a
default column's prompt would silently mutate the package-level `DefaultColumns`
slice for the lifetime of the process.

Two follow-up fixes are included: `MoveCard` now resolves the destination column
from a single pipeline snapshot so a concurrent column removal or reorder can
no longer panic or land a card in the wrong column; and column names/emoji are
normalized to a single line at the source so newlines can't break the header
layout or mouse hit-boxes, closing a terminal-escape injection path in the
prompt editor render path.

### DIFF
--- a/pkg/board/app.go
+++ b/pkg/board/app.go
@@ -115,6 +115,117 @@ func (a *App) SetColumnPrompt(colID, prompt string) error {
 	return a.saveConfigLocked()
 }
 
+// AddColumn validates and appends a column to the pipeline, persisting it
+// to the user's global config file. The column's id is derived from its
+// name and kept unique; the saved column (with its id) is returned so the
+// UI can select it.
+func (a *App) AddColumn(col Column) (Column, error) {
+	col, err := normalizeColumn(col)
+	if err != nil {
+		return Column{}, err
+	}
+
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	col.ID = a.uniqueColumnIDLocked(col.Name)
+	a.columns = append(a.columns, col)
+	if err := a.saveConfigLocked(); err != nil {
+		return Column{}, err
+	}
+	return col, nil
+}
+
+// UpdateColumn replaces the column with the given id, applying the same
+// validation as AddColumn and persisting the change. The id never changes
+// on an update, so existing cards stay attached to the column across a
+// rename.
+func (a *App) UpdateColumn(id string, col Column) (Column, error) {
+	col, err := normalizeColumn(col)
+	if err != nil {
+		return Column{}, err
+	}
+
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	idx := slices.IndexFunc(a.columns, func(c Column) bool { return c.ID == id })
+	if idx < 0 {
+		return Column{}, fmt.Errorf("unknown column %q", id)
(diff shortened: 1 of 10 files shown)


pkg/board/app.go
```diff
@@ -115,6 +108,121 @@ func (a *App) SetColumnPrompt(colID, prompt string) error {
...
+func (a *App) AddColumn(col Column) (Column, error) {
+	col, err := normalizeColumn(col)
+	if err != nil {
+		return Column{}, err
+	}
+
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	col.ID = a.uniqueColumnIDLocked(col.Name)
+	a.columns = append(a.columns, col)
```
docker-agent, line 124: **[MEDIUM] AddColumn / UpdateColumn / MoveColumn / RemoveColumn mutate `a.columns` before `saveConfigLocked()` with no rollback on failure**

All four new column-mutation methods follow this pattern (shown for `AddColumn`, lines 124-126):

```go
a.columns = append(a.columns, col)       // mutation committed to memory
if err := a.saveConfigLocked(); err != nil {
    return Column{}, err                  // error returned — but a.columns is not rolled back
}
```

If `saveConfigLocked()` fails (e.g. disk full, permission error, config path temporarily unavailable), the in-memory `a.columns` has already been mutated but the on-disk config has not. From that point on:
- `Columns()` returns the ghost/corrupted column list.
- `uniqueColumnIDLocked` considers the ghost column when allocating new IDs.
- `slices.Equal(a.columns, DefaultColumns)` in `saveConfigLocked` may now return `false` permanently, so even a restore-defaults call would write a non-default config to disk.
- The only recovery is restarting the process, which reloads from disk.

The fix is to snapshot the old value before mutating and restore it on error, e.g. for `AddColumn`:
```go
before := a.columns
a.columns = append(a.columns, col)
if err := a.saveConfigLocked(); err != nil {
    a.columns = before
    return Column{}, err
}
```
Same pattern applies in `UpdateColumn` (save `old := a.columns[idx]` then restore), `MoveColumn` (save the old slice), and `RemoveColumn`.

docker-agent (commented):

One medium-severity correctness issue found: all four new column-mutation methods (, , , ) mutate the in-memory  slice before calling , with no rollback if the save fails. This leaves in-memory and on-disk state diverged until the process restarts.

melmennaoui (approved):

PR #3559, opened 2026-07-09
feat(tui): on-hover copy button on all messages and on-hover edit button on user messages

dgageot:
Previously, user messages had a persistent "✎" edit icon that was always visible, and the copy affordance was only available on assistant messages. This was visually noisy on the user side and inconsistent across message types.

This PR makes both actions hover-only for user messages. When the cursor enters a user message, a right-aligned action row appears showing "⎘ copy" and "✎ edit" side by side. The row replaces the top padding when hovered (blank spacer otherwise), so message layout never shifts. Clicking copy flashes an inline "copied" confirmation; clicking edit opens the existing inline editor. The copy button is also now surfaced on user messages for the first time, matching the behaviour already present on assistant messages. The `AssistantMessageCopyLabel` type constant was renamed to `MessageCopyLabel` to reflect this broader scope.

To prevent phantom clicks, the edit and copy hit-tests are gated on the hover/selection state and pinned to the action row's first line, so body text that happens to match the label string is never treated as a button. The action row is also prevented from wrapping at very narrow widths — labels are dropped or truncated rather than reflowing — so hovering can never change the message height. A shared `labelHit` helper centralises click detection for edit, copy, and retry. The render cache key now includes the `editable` bit so stale frames are never served after an edit session ends.


Sayt-0 (approved):

pkg/tui/components/messages/messages.go
```diff
@@ -1939,65 +1957,34 @@ func (m *model) codeBlockAt(msgIdx, localLine, col int) (string, bool) {
...
-	before, _, ok := strings.Cut(plainLine, types.AssistantMessageCopyLabel)
-	if !ok {
+	switch m.messages[msgIdx].Type {
+	case types.MessageTypeUser:
+		// The action row is the first line of the user bubble; pinning the
+		// hit test there rules out collisions with message content.
+		if localLine != 0 {
+			return false
+		}
+	case types.MessageTypeAssistant:
```
docker-agent, line 1975: **[MEDIUM] Assistant copy-label hit-test not pinned to the action-row line**

The new `isCopyLabelClick` switch correctly guards the user-message branch with `if localLine != 0 { return false }`, but the `MessageTypeAssistant` branch is empty — no equivalent line-pin is applied.

Because `labelHit` searches for `⎘ copy` in whatever `item.lines[localLine]` it receives, a click on *any* line of an assistant message whose plain text happens to contain `⎘ copy` (e.g. a reply that says "press **⎘ copy** to copy") will silently trigger a whole-message copy instead of propagating to text-selection.

The assistant action row is always rendered at line 0 of the styled bubble (same layout as user messages). Add the same guard:

```go
case types.MessageTypeAssistant:
    if localLine != 0 {
        return false
    }
```

docker-agent (commented):

This PR introduces a well-structured hover-only action row for copy/edit on TUI messages, with good pinning logic for user messages. One asymmetry was found in the new `isCopyLabelClick` implementation where the assistant branch lacks the `localLine == 0` guard that the user branch correctly applies.

PR #3560, opened 2026-07-09
fix(shell): return clear error when working directory does not exist

dgageot:
When Go forks a child process with `SysProcAttr{Setpgid: true}` (which the shell toolset sets to keep child processes in their own process group), the kernel's fork+exec path performs the working-directory `chdir` inside the child. If that directory no longer exists, the error surfaces as "fork/exec /bin/zsh: no such file or directory", making it look like the shell binary itself is missing rather than the cwd.

This is a real problem for TUI sessions that restore a shell tab tied to a git worktree that has since been removed. Every tool call in that tab fails with a cryptic, misleading error.

The fix adds a `checkWorkDir()` pre-flight in the shell tool that runs before the process is spawned. It detects the three actionable cases — path does not exist, path exists but is not a directory, and a stat error — and returns a clear tool-level error message for each. An empty cwd is left alone (meaning "inherit the process cwd"), matching the existing behaviour. The check runs after the OTel span is opened so failed calls remain traceable. A `TestCheckWorkDir` table-driven test covers all cases, and a `TestShellTool_MissingWorkingDir` regression test exercises the full tool path.

### DIFF
--- a/pkg/tools/builtin/shell/shell.go
+++ b/pkg/tools/builtin/shell/shell.go
@@ -144,6 +144,9 @@ func (h *shellHandler) RunShell(ctx context.Context, params RunShellArgs, rt too
 	defer cancel()
 
 	cwd := h.resolveWorkDir(params.Cwd)
+	if msg := checkWorkDir(cwd); msg != "" {
+		return tools.ResultError(msg), nil
+	}
 
 	// Stamp the call shape (cmd, cwd, timeout) onto the active span.
 	// Cmd ships unconditionally — it's the main signal of what the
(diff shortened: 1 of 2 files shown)


Sayt-0 (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The pre-flight `checkWorkDir()` function correctly handles all three actionable cases (non-existent path, path exists but is not a directory, and stat errors) before spawning the child process. The empty-cwd fast-path correctly preserves the existing behaviour of inheriting the process working directory. The table-driven `TestCheckWorkDir` covers the relevant branches, and `TestShellTool_MissingWorkingDir` provides a regression test for the full tool path.

No bugs or issues were found in the changes introduced by this PR.

PR #3561, opened 2026-07-09
fix(mcp): prefer OAuth resource_metadata challenge param

rumpl:
## Summary
- Prefer the WWW-Authenticate resource_metadata parameter when discovering OAuth protected resource metadata
- Keep the legacy resource parameter fallback for existing MCP servers
- Add regression coverage for both header shapes

## Tests
- go test ./pkg/tools/mcp

### DIFF
--- a/pkg/tools/mcp/oauth.go
+++ b/pkg/tools/mcp/oauth.go
@@ -29,8 +29,10 @@ import (
 	"github.com/docker/docker-agent/pkg/tools"
 )
 
-// resourceMetadataFromWWWAuth extracts resource metadata URL from WWW-Authenticate header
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly refactors `resourceMetadataFromWWWAuth` to prefer the `resource_metadata` WWW-Authenticate challenge parameter over the legacy `resource` parameter, with a proper fallback. The new regex patterns handle both quoted and unquoted parameter values, the helper `wwwAuthParam` correctly reads the right capture group, and both new test cases exercise the preference and fallback paths.

No high or medium severity issues were found in the changed code.

Sayt-0 (approved):

PR #3563, opened 2026-07-09
fix(board): reliability fixes for SSE lifecycle, watchdog, transport, and rendering

dgageot:
The board's SSE event stream had several reliability gaps that caused missed events, stalled connections, and occasional rendering glitches under real-world network conditions.

This change tightens the full SSE lifecycle: the watchdog timer now resets on every received line rather than only on data frames, so comment and keep-alive lines properly defer the timeout. Keep-alive probes are disabled on the Unix domain socket transport, which doesn't need them and was generating spurious traffic. The heartbeat goroutine is now drained before the function returns, preventing a subtle goroutine leak on reconnect. The title-truncation helper now trims both halves of a split title consistently. A `defer` replaces the manual unlock in `SetSamplingWithToolsHandler`, fixing the mutex lint warning. Regression tests cover the watchdog reset behaviour and the title truncation, and timing margins in the heartbeat test were widened to reduce flakiness on slower CI runners.

All existing tests pass (`task test`), the linter is clean (`task lint`), and the binary builds (`task build`). The watchdog and heartbeat paths were also verified with targeted `-race` runs.

### DIFF
--- a/pkg/board/client.go
+++ b/pkg/board/client.go
@@ -87,6 +87,7 @@ type client struct {
 // unix socket and targets the given session id.
 func newClient(socket, session string) *client {
 	transport := &http.Transport{
+		DisableKeepAlives: true,
 		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
 			var d net.Dialer
 			return d.DialContext(ctx, "unix", socket)
@@ -149,8 +150,7 @@ func (c *client) Followup(ctx context.Context, idempotencyKey, message string) e
 	if resp.StatusCode != http.StatusAccepted {
 		return fmt.Errorf("followup: %s", resp.Status)
 	}
-	// Drain so the connection can be reused; the body only reports whether
-	// the delivery was a duplicate, which the board does not care about.
+	// Drain the small response body before closing it.
 	_, _ = io.Copy(io.Discard, resp.Body)
 	return nil
 }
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are well-reasoned and correctly address the stated reliability gaps in the SSE lifecycle. Here is a brief summary of the review findings:

- **Watchdog timer reset** (`client.go`): Moving `resetWatchdog()` before the heartbeat-comment branch so every received SSE line (including `:` ping comments) resets the watchdog is logically correct and matches the advertised fix.
- **`DisableKeepAlives: true` on Unix socket transport** (`client.go`): Correct — keep-alive probes are meaningless on Unix domain sockets and could generate spurious traffic; disabling them is appropriate.
- **Heartbeat goroutine drain** (`server.go`): The `heartbeatDone` channel approach correctly prevents the goroutine leak on reconnect; the deferred closure stops the heartbeat and waits for `heartbeatDone`, and the goroutine closes it on exit.
- **`splitTitle` truncation** (`tui/view.go`): Truncating `line1` symmetrically with the remainder is the correct fix for the rendering inconsistency.
- **`defer c.mu.Unlock()`** (`session_client.go`): Replacing manual `Unlock()` with `defer` is safer (panic-proof) and fixes the mutex lint warning as intended.
- **Tests**: Timing margin widening and new regression tests for watchdog reset and title truncation are appropriate.

No bugs were found in the newly introduced code (`+` lines). All changes appear correct and improve reliability.

trungutt (approved):

PR #3565, opened 2026-07-10
fix(tui): wire MCP prompts into slash command dispatch

EronWright:
## What

MCP prompt palette items were registered without `SlashCommand`/`Immediate`, so
typing an MCP prompt as a slash command (e.g. `/review`, `/summarize`) fell
through to the model as plain chat text instead of invoking the prompt — only
clicking the item in the command palette worked.

This wires MCP prompt items into the slash-command parser, mirroring the
existing agent-command and skill items:

- Set `SlashCommand` + `Immediate` so `Parser.Parse` matches them.
- Map a non-empty argument to the prompt's first declared argument (so
  `/summarize <text>` works from the keyboard), falling back to the input dialog
  when a required argument is missing, and to an empty-args run otherwise.

Companion to the skill-dispatch fix in #3054 — this is the MCP-prompt analog.
Unlike skills it does **not** need `SkipImmediateParse`: its `Execute` emits
`MCPPromptMsg` directly rather than re-emitting a `SendMsg` that would be
re-parsed.

## Refactor + tests

To make the item builders unit-testable without standing up an `App`, the
per-item construction is factored into small helpers — `newMCPPromptItem` and,
for symmetry, `newAgentCommandItem` — each with unit tests that drive the real
item through `Parser.Parse` (slash-command wiring + argument mapping/forwarding).

### DIFF
--- a/pkg/tui/commands/commands.go
+++ b/pkg/tui/commands/commands.go
@@ -9,6 +9,7 @@ import (
 	tea "charm.land/bubbletea/v2"
 
 	"github.com/docker/docker-agent/pkg/app"
+	"github.com/docker/docker-agent/pkg/config/types"
 	"github.com/docker/docker-agent/pkg/feedback"
 	mcptools "github.com/docker/docker-agent/pkg/tools/mcp"
 	"github.com/docker/docker-agent/pkg/tui/components/toolcommon"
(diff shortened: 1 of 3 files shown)


Sayt-0 (approved):

Sayt-0:
Hi @EronWright ! your commits must be signed in this repo

rumpl:
This is a good first step, next we would need to handle optional parameters for mcp prompts

PR #3568, opened 2026-07-10
Fix openai 1password api key resolution

krissetto:
avoids errors like this when using the 1password cli to resolve the openai api key

```
model failed: error receiving from stream: HTTP 401: Incorrect API key provided: op://******************. You can find your API key at https://platform.openai.com/account/api-keys.
```


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the 1Password `op://` reference resolution bug. When no `token_key` is configured, the new `default` branch now resolves `OPENAI_API_KEY` through the environment provider chain (which handles `op://` references, keychain lookups, Docker Desktop, etc.) before falling back to the SDK's raw `os.Getenv`. This prevents the unresolved `op://...` reference from being sent as the bearer token.

**Changes reviewed:**
- `pkg/model/provider/openai/client.go` — New `default` case resolves the API key via the env provider chain; clean and correct.
- `pkg/model/provider/openai/client_auth_test.go` — Test coverage for both the env-provider resolution path and the SDK fallback path.

No high or medium severity issues found. The implementation is focused and correct.

melmennaoui (approved):

PR #3580, opened 2026-07-10
Improve TUI rendering

dgageot:


docker-agent (commented):
### Assessment: 🟢 APPROVE

The two new ENV variables ( and ) are correct and well-suited for their purpose. Setting  signals true-color terminal support to applications running in the container, enabling richer TUI rendering. Setting  ensures proper UTF-8 locale handling, which is important for TUI libraries that render Unicode box-drawing characters and other glyphs. No bugs, security issues, or regressions introduced.

Sayt-0 (approved):

PR #3583, opened 2026-07-10
feat: implement dynamic MCP HTTP headers via header factory

Piyush0049:
### Description

This PR refactors `NewHeaderTransport` in the upstream transport layer to use a `headerFactory` (`func(context.Context) map[string]string`) instead of a static map for HTTP headers. 

Currently, the MCP remote client initializes with a static `map[string]string`. This means whatever headers are present at startup are permanently hardcoded into the HTTP client for its lifespan. By migrating to a header factory, this PR enables dynamic, context-aware resolution of HTTP headers per-request. 

This is a critical architectural improvement that allows the agent to securely fetch fresh, short-lived tokens (like OAuth tokens) on every single request and pass dynamic execution metadata during MCP sessions.

### Additional Fixes

* **Windows Test Flakes**: Fixed pre-existing Windows-specific test flakes in `pkg/tools/mcp/keyringstore/tokenstore_test.go`. The test assertions were updated to expect `0666` file permissions on Windows instead of Unix-style `0600`. Go's `os.Stat` always reports ACLs as `0666` or `0444` natively on Windows, regardless of the mode passed to `os.WriteFile`. This keeps the test suite completely green for Windows contributors without weakening application security.

### Testing

- [x] Tested locally with `task test`
- [x] Verified Windows file permission test failures are resolved
- [x] Verified `go build` and `go vet` are completely clean


Sayt-0:
The underlying TODO has since been resolved on `main` by ef6d626a (`chore(mcp): resolve OAuth TODOs and expand headers per request`), which covers per-request header expansion with a backward-compatible `HeaderResolver` design, so the usefulness of this PR now seems doubtful.

Piyush0049:
I see that ef6d626a covers MCP but it missed A2A toolsets. In main pkg/tools/a2a/a2a.go still evaluates headers statically at connection startup. My PR fixes this so A2A also gets per request expansion. It also includes a Windows file permission fix for the tokenstore tests. Should I update this PR to apply the new HeaderResolver pattern to a2a.go and keep the Windows fix?

Sayt-0:
Yes to both: apply `NewHeaderTransportWithResolver` to `pkg/tools/a2a/a2a.go` (it still resolves headers once in `Start()`), and keep the Windows fix but use `t.Skip` on Windows instead of asserting `0o666`, matching `atomicfile_test.go`/`runregistry_test.go`.

dgageot (approved):

PR #3586, opened 2026-07-11
fix(eval): reap orphaned eval containers on context cancellation

### LINKED ISSUE #3585: eval: cancelled/interrupted runs orphan --rm eval containers (never reaped)
### Summary
When a `docker agent eval` batch's context is cancelled (timeout, Ctrl-C, or a per-eval failure that cancels the parent context), the per-eval containers spawned with `--rm` are **not reaped** — they keep running indefinitely after the eval process has printed its summary and exited.

### Observed impact
A full eval run of a multi-agent config reported its final summary at the ~1h30m mark, but left **two `docker-agent-eval-*` containers running for 40+ minutes afterward**, each pegged doing a runaway `search_files_content(path:"/")` filesystem walk inside the container. They had to be killed manually with `docker kill`. Because the containers were launched with `--rm`, one would expect them to be removed on stop — but since they never receive a stop signal, `--rm` never fires.

### Root cause (from source)
In `pkg/evaluation/eval.go`, `runDockerAgentInContainer` launches each eval via:

```go
cmd := exec.CommandContext(ctx, "docker", args...)   // args include: run --name docker-agent-eval-<id> --rm --privileged --init ...
```

`exec.CommandContext` only kills the **`docker` CLI client process** when `ctx` is cancelled — it sends no `docker stop`/`docker kill` to the container the client started. The container keeps running detached from the (now-dead) client, and `--rm` (which only removes the container after it *stops*) never triggers. There is no deferred `docker kill <containerName>` / `docker rm -f <containerName>` on the cancellation/cleanup path.

### Suggested fix
Register cleanup keyed on the deterministic `containerName` (`docker-agent-eval-<uuid>`) that runs even when `ctx` is cancelled — e.g. a `defer`/cleanup that issues `docker rm -f <containerName>` (or `docker kill`) using a fresh, non-cancelled context, so an interrupted/timed-out batch reaps its children instead of orphaning them.

### Environment
- Local `docker-agent` binary (built from a recent `main`, newer than the Docker Desktop plugin).
- macOS, arm64.
- Reproduced via a Tier-2 eval suite run that hit a per-eval failure cascade / context cancellation.

aheritier:
## Summary

Fixes #3585. When a `docker agent eval` run is cancelled (timeout, Ctrl-C, or a per-eval failure cancelling the parent context), the per-eval containers launched with `--rm` were orphaned rather than reaped. `exec.CommandContext` kills the docker CLI process on cancellation but never signals the container it started, so `--rm` cleanup never fires — leaving long-lived containers (observed 40+ min, runaway filesystem walks).

## Changes

- **Core fix:** added `reapContainer`, deferred immediately after `cmd.Start()` in `runDockerAgentInContainer`, which force-removes the container via `docker rm -f <name>` on a **fresh, detached context** (so it runs even when the eval context is already cancelled). Respects `--keep-containers` and treats `No such container` as benign (the normal `--rm` happy path).
- **Stderr race fix:** replaced `StderrPipe` + background `io.ReadAll` goroutine with `cmd.Stderr = &bytes.Buffer{}`, eliminating a pre-existing data race and the incorrect Wait-before-drain ordering by construction. Added `cmd.WaitDelay = 10s` hardening.
- **Tests:** 5 daemon-free integration tests through `runDockerAgentInContainer` (fake `docker` on PATH distinguishing `run`/`rm`) covering caller-context cancellation, non-zero exit, scanner token-too-long, normal success, and `--keep-containers` skipping cleanup. Verified the regression is caught — removing the `defer` fails 4 of 5.
- **Docs:** documented cleanup-on-cancellation behavior in `docs/features/evaluation/index.md` and `docs/features/cli/index.md`.

## Validation

- `task build` / `task test` / `task lint` — all clean
- `go test -race ./pkg/evaluation` — clean

Fixes #3585

### DIFF
--- a/docs/features/cli/index.md
+++ b/docs/features/cli/index.md
@@ -423,7 +423,7 @@ $ docker agent eval <agent-file>|<registry-ref> [<eval-dir>|./evals] [flags]
 | `--output <dir>`    | `<eval-dir>/results`                 | Directory for results, logs, and session databases                         |
 | `--only <pattern>`  | (all)                                | Only run evals with file names matching these patterns (repeatable)        |
 | `--base-image`      | (default)                            | Custom base Docker image for eval containers                               |
-| `--keep-containers` | `false`                              | Keep containers after evaluation (don't remove with `--rm`)                |
+| `--keep-containers` | `false`                              | Keep containers after evaluation, including cancelled/interrupted runs (don't remove with `--rm`) |
 | `-e, --env`         | (none)                               | Environment variables to pass to container (`KEY` or `KEY=VALUE`, repeatable) |
 | `--repeat <n>`      | `1`                                  | Number of times to repeat each evaluation (useful for computing baselines) |
 
--- a/docs/features/evaluation/index.md
+++ b/docs/features/evaluation/index.md
@@ -165,6 +165,11 @@ $ docker agent eval <agent-file>|<registry-ref> [<eval-dir>|./evals]
 | `-e, --env`         | (none)                      | Environment variables to pass to container (`KEY` or `KEY=VALUE`) |
 | `--repeat`          | `1`                         | Number of times to repeat each evaluation (useful for computing baselines) |
 
+> [!NOTE]
+> **Container cleanup on cancellation**
+>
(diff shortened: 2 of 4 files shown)


dgageot (approved):

dgageot (changes requested):
I don't think we need this

dgageot:
Closing in favour of https://github.com/docker/docker-agent/pull/3615

Issue #3584, opened 2026-07-10
Concurrent background-job elicitations are silently dropped — nothing displayed to the user


aheritier:
## Summary
When user input (an elicitation request) is triggered from multiple concurrent background jobs, nothing is displayed to the user at all — the request is silently dropped and the requesting handler blocks forever. This is a functional failure experienced frequently in practice, not just a theoretical race.

## Root cause
The elicitation path assumes a single, synchronous, LIFO-nested request at a time, which breaks under concurrency (e.g. `run_background_agent` jobs running `RunStream` on detached concurrent goroutines):

- The `elicitationBridge` is a **single global channel slot** with a LIFO-nesting assumption (`pkg/runtime/elicitation.go:59`, `pkg/runtime/loop.go:307`/`217`). Concurrent background jobs cause the swaps/restores to interleave arbitrarily.
- **`runCollecting` silently discards `ElicitationRequestEvent`s** — it only handles content/tokens/errors (`pkg/runtime/agent_delegation.go:366-383`). So a background job's elicitation event lands in a collector that drops it while the handler parks forever.
- Interleaved teardown leaves the channel slot **nil or closed** (foreground turn ends first → `errNoElicitationChannel` / recover path) → silent failure; foreground elicitations can also bleed into a background collector.
- Display side: the tab supervisor's `PendingEvent` is a single overwritable slot (`pkg/tui/.../supervisor.go:28`/`202`).
- There is also a TOCTOU: `ResumeElicitation`'s `default:` branch returns "no elicitation request in progress" if the response lands in the gap before the handler parks. `ElicitationID` exists on the request event but is unused for routing.

## Proposed fix
- Per-request response channels keyed by `ElicitationID` (register-before-emit, eliminating the TOCTOU `default:` branch).
- A stable `OnElicitationRequest` UI sink mirroring the existing `OnBackgroundEvent` sink so background-job elicitations are surfaced to the foreground UI.
- Additive `elicitation_id` field on `api.ResumeElicitationRequest` (it does not carry one today).
- Headless/non-interactive fast-decline with a model-readable note.
- A supervisor pending-event queue instead of a single overwritable slot.

## Acceptance criteria
- Elicitation requests raised from concurrent background jobs are displayed to the user and can be answered, with responses routed back to the correct waiter.
- No silent drops when multiple elicitations are pending simultaneously.
- Regression tests covering concurrent/background-job elicitation (with `-race`).

PR #3587, opened 2026-07-11
fix(runtime): reliable, correlation-safe, non-blocking elicitation delivery (#3584)

aheritier:
## Summary

**Previously this PR was the full A6 implementation; it has been split into a reviewable stack** and now scopes to the **runtime core** only. Fixes the heart of #3584 — elicitation requests raised from concurrent background jobs were silently dropped and the requesting handler blocked forever.

### #3584 stack
1. ✅ #3618 — fix(tui): FIFO pending-attention queue (merged)
2. **this PR (#3587)** — runtime core: correlation, delivery, API field
3. #3624 — feat(tui,app): surface & correlate background-job elicitations
4. #3625 — feat(server): session-scoped elicitation sink

## What changed

- **Per-request correlation:** waiters keyed by an internal ID, registered *before* the request event is emitted (eliminates the TOCTOU branch).
- **Reliable delivery:** the `OnElicitationRequest` sink is called synchronously; the best-effort bridge send is ctx-bounded on a detached goroutine and can never block delivery or response processing.
- **Atomic resolve-vs-cancel:** a terminal-state guard ensures a delivered response is never discarded by a racing `ctx.Done()`.
- **Backward-compatible API:** `ResumeElicitation`'s new `elicitationID` is variadic (pre-#3584 callers still compile); additive `elicitation_id` wire field; server pass-through wires it end-to-end.
- **Headless fast-decline** with a model-readable note when no sink is registered.

The `Runtime` interface change and its repo-wide implementor/mock updates are one commit for Go compile atomicity.

## Testing

- `pkg/runtime/elicitation_concurrency_test.go`: duplicate wire IDs, cancel-vs-resolve race (300-iter `-race`), wedged bridge, real-handler TOCTOU, exactly-once delivery, concurrent background-job routing.
- Full stack `task dev` + `-race` — green.

Bottom of the remaining stack (targets `main` now that #3618 is merged).

### DIFF
--- a/docs/guides/go-sdk/index.md
+++ b/docs/guides/go-sdk/index.md
@@ -129,6 +129,23 @@ if err := chat.Restart(); err != nil {
 
 For advanced use (custom elicitation, raw event inspection), call `chat.Runtime()` to access the underlying `runtime.Runtime` directly.
 
+> [!WARNING]
+> **Breaking change: `Runtime.ResumeElicitation` (#3584)**
+>
+> `Runtime.ResumeElicitation` gained an `elicitationID` parameter so responses can
+> be correlated with a specific concurrent elicitation request (needed once
+> multiple background jobs can be eliciting input at the same time). It is
+> declared **variadic** (`elicitationID ...string`) specifically so existing
+> *callers* of the 3-argument form keep compiling unchanged — `rt.ResumeElicitation(ctx, action, content)`
+> still works and falls back to resolving the sole pending request.
+>
+> If you implement your own `runtime.Runtime` (rather than embedding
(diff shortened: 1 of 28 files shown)


docs/guides/go-sdk/index.md
```diff
@@ -129,6 +129,23 @@ if err := chat.Restart(); err != nil {
 
 For advanced use (custom elicitation, raw event inspection), call `chat.Runtime()` to access the underlying `runtime.Runtime` directly.
 
+> [!WARNING]
```
aheritier, line 132: can it impact AP @simonferquel ?

dgageot (approved):

PR #3605, opened 2026-07-11
fix(permissions): match glob wildcards across path separators

### LINKED ISSUE #3604: Permission `deny` rules fail open for argument values containing `/` (`matchGlob` uses `filepath.Match`)
### Description

A configured `deny` permission whose argument value contains a path separator (`/`, or `\` on Windows)
can silently fail to match, so a tool call the operator explicitly tried to block is not denied.

The matcher `matchGlob` in `pkg/permissions/permissions.go` has two paths:

1. A trailing-`*` fast path: if the pattern ends in `*` and the prefix contains no other glob
   metacharacters, it does a plain `strings.HasPrefix`. This is why `shell:cmd=rm*` and `shell:cmd=sudo*`
   correctly block `rm -rf /`.
2. Every other pattern falls through to `filepath.Match(pattern, value)`. `filepath.Match`'s `*` matches
   any run of **non-separator** characters, so it stops at `/`.

As a result, a `deny` pattern with a **non-trailing** wildcard - e.g. `shell:cmd=*rm -rf*` - matches
`rm -rf tmp` but not `rm -rf /`. The docstring on `matchGlob` says "we want `*` to match any characters
including spaces," but that only holds for the fast path; the `filepath.Match` fallback silently reintroduces
separator-boundary behavior. Because this is a `deny` rule, the failure direction is fail-**open**.

### Expected Behavior

A `deny` pattern with a wildcard should match the same command regardless of whether the argument value
happens to contain a `/`. `shell:cmd=*rm -rf*` should deny both `rm -rf tmp` and `rm -rf /`.

### Actual Behavior

`shell:cmd=*rm -rf*` denies `rm -rf tmp` but returns `Ask` for `rm -rf /` - the command is not blocked.
The decision varies with argument content and with OS (separator is `/` on Unix, `\` on Windows).

### Steps to Reproduce

Config-level: configure a deny rule with a non-trailing wildcard and invoke the shell tool with an
argument containing `/`:

```yaml
permissions:
  deny:
    - "shell:cmd=*rm -rf*"
```

- `shell` with `cmd = "rm -rf tmp"` → denied ✅
- `shell` with `cmd = "rm -rf /"`   → **not** denied (falls through to Ask) ❌

Isolated unit reproduction against `pkg/permissions` (the checker's public API + the matcher):

```go
func TestDenyGlobDoesNotCrossPathSeparator(t *testing.T) {
	checker := NewCheckerFromRules(nil, nil, []string{"shell:cmd=*rm -rf*"})
	assert.Equal(t, Deny, checker.CheckWithArgs("shell", map[string]any{"cmd": "rm -rf tmp"})) // passes
	assert.Equal(t, Deny, checker.CheckWithArgs("shell", map[string]any{"cmd": "rm -rf /"}))   // FAILS: got Ask
}

func TestMatchGlobCrossesPathSeparator(t *testing.T) {
	assert.True(t, matchGlob("*rm -rf*", "rm -rf /"))       // FAILS
	assert.True(t, matchGlob("*/etc/*", "cat /etc/passwd")) // FAILS
}
```

`go test ./pkg/permissions/ -run 'TestDenyGlob...|TestMatchGlob...' -v`

### Docker Agent version

Docker version 29.5.3, build d1c06ef

### OS & terminal

Reproduced on Linux (in a `golang:1.26.4` container). The separator-boundary behavior is inherent to `filepath.Match` and also affects Windows (where the boundary is `\`), so the result is platform-dependent.

### Model used

N/A - this is in the permission matcher; it's independent of the model.

### Error output

```shell
--- FAIL: TestDenyGlobDoesNotCrossPathSeparator (0.00s)
    Not equal: expected: 2 (Deny)   actual: 0 (Ask)
    Messages: deny rule must also block 'rm -rf /' (the most dangerous case)
--- FAIL: TestMatchGlobCrossesPathSeparator (0.00s)
    Should be true
    Messages: "*rm -rf*" should match "rm -rf /"
FAIL    github.com/docker/docker-agent/pkg/permissions


(Decision enum: `Ask=0, Allow=1, Deny=2`.)
```

### Screenshots

_No response_

### Additional context

- **Impact:** in a plain interactive session `Ask` still prompts, so this is a degraded control rather than
  immediate silent execution - but it becomes a real bypass under `--yolo`, non-interactive / auto-approval
  runs, or when an overlapping `allow` rule is present. The same `filepath.Match` path also breaks `allow`
  rules, but those fail closed.
- **Trigger:** only patterns with a **non-trailing** wildcard are affected; trailing-only patterns (`rm*`,
  `sudo*`) hit the fast path and work, which tends to hide this in casual testing.

parveshsaini:
## What

`matchGlob` in `pkg/permissions` falls back to `filepath.Match` for any pattern that isn't a plain trailing
wildcard. `filepath.Match`'s `*` and `?` match only **non-separator** characters, so they stop at `/` (and at
`\` on Windows). This translates the glob to an anchored regexp whose `*`/`?` span **any** character instead.

## Why

Permission patterns match argument *values* — shell commands, file paths, URLs — which routinely contain `/`.
Because of the separator boundary, a rule with a **non-trailing** wildcard silently fails to match once the
value contains a `/`:

```yaml
permissions:
  deny:
    - "shell:cmd=*rm -rf*"
```

- `shell` with `cmd = "rm -rf tmp"` → denied ✅
- `shell` with `cmd = "rm -rf /"`   → **not** denied (falls through to `Ask`) ❌

For a `deny` rule this is the unsafe direction — it **fails open**. In an interactive session the fall-through
to `Ask` still prompts, so the practical exposure is a degraded control; it becomes a real bypass under
`--yolo`, non-interactive / auto-approval runs, or when an overlapping `allow` rule is present. The same
boundary also silently breaks `allow` rules, though those fail closed.

The existing docstring already stated the intent — *"for argument matching we want `*` to match any characters
including spaces"* — but that only held for the trailing-wildcard fast path (`sudo*`), which is why casual
testing with trailing-only patterns never surfaced this.

## How

- Replace the `filepath.Match` fallback in `matchGlob` with `globToRegexp` + `regexp.MatchString`.
- `globToRegexp` translates a `filepath.Match`-style glob into an anchored regexp: `*` → `.*`, `?` → `.`,
  character classes (`[...]`) copied through (bracket syntax is shared with regexp), everything else matched
  literally via `regexp.QuoteMeta`.
- The trailing-`*` prefix-match fast path and the case-insensitive lowercasing are unchanged.
- No new dependencies (`regexp` is stdlib; `path/filepath` import dropped).

This keeps all existing `matchGlob` semantics (`*`, `?`, `[...]`, case-insensitivity) — only the separator
boundary changes.

## Tests

- New `TestDenyGlobCrossesPathSeparator`: end-to-end via `CheckWithArgs`, asserting a `*rm -rf*` deny rule
  blocks both `rm -rf tmp` and `rm -rf /`.
- Extended the `TestMatchGlob` table with separator-crossing cases (`*rm -rf*`, `*/etc/*`, `a?c` vs `a/c`).

These fail on `main` and pass with the fix. Full `pkg/permissions` suite, `go vet`, `gofmt`, and
`golangci-lint` are all green (Go 1.26.4).

```
--- PASS: TestDenyGlobCrossesPathSeparator (0.00s)
--- PASS: TestMatchGlob (0.00s)
ok  	github.com/docker/docker-agent/pkg/permissions
```

### DIFF
--- a/pkg/permissions/permissions.go
+++ b/pkg/permissions/permissions.go
@@ -4,7 +4,7 @@ package permissions
 
 import (
 	"fmt"
-	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/docker/docker-agent/pkg/config/latest"
@@ -239,17 +239,18 @@ func argToString(v any) string {
 }
 
 // matchGlob checks if a value matches a glob pattern.
-// Supports glob-style patterns using filepath.Match semantics:
-// - "*" matches any sequence of characters within a path segment
+// Supports glob-style patterns:
+// - "*" matches any sequence of characters (including path separators and spaces)
 // - "?" matches any single character
 // - "[...]" matches character classes
 //
 // Matching is case-insensitive.
 //
-// Note: filepath.Match's "*" stops at path separators, but for argument
-// matching we want "*" to match any characters including spaces.
-// We handle trailing wildcards specially to support patterns like "sudo*"
-// matching "sudo rm -rf /".
+// Note: filepath.Match's "*"/"?" stop at path separators ("/", or "\" on
+// Windows), but argument values (shell commands, file paths, URLs) routinely
+// contain "/", so matching must not depend on it. We therefore translate the
+// glob to an anchored regexp instead of using filepath.Match. Trailing
+// wildcards keep a plain prefix-match fast path (e.g. "sudo*").
 func matchGlob(pattern, value string) bool {
 	// Normalize both to lowercase for case-insensitive matching
 	pattern = strings.ToLower(pattern)
@@ -266,7 +267,70 @@ func matchGlob(pattern, value string) bool {
 		}
 	}
 
-	// Try glob pattern match (also handles exact matches)
-	matched, err := filepath.Match(pattern, value)
-	return err == nil && matched
+	// Full glob match (also handles exact matches). Unlike filepath.Match, the
+	// translated "*"/"?" span any character, so patterns match values that
+	// contain path separators.
+	re, err := globToRegexp(pattern)
+	if err != nil {
+		return false
+	}
+	return re.MatchString(value)
+}
+
+// globToRegexp translates a filepath.Match-style glob into an anchored regular
+// expression. Unlike filepath.Match, the resulting "*" and "?" match any
+// character, including path separators — argument values such as shell commands,
+// file paths and URLs routinely contain "/", and permission patterns must match
+// them regardless. Character classes ("[...]"), whose bracket syntax glob and
+// regexp share, are copied through verbatim; every other character is matched literally.
+func globToRegexp(pattern string) (*regexp.Regexp, error) {
+	var b strings.Builder
+	b.WriteString(`\A`)
+	for i := 0; i < len(pattern); {
+		switch c := pattern[i]; c {
+		case '*':
+			b.WriteString(`.*`)
+			i++
+		case '?':
+			b.WriteByte('.')
+			i++
+		case '[':
+			if end := classEnd(pattern, i); end > 0 {
+				b.WriteString(pattern[i : end+1])
+				i = end + 1
+				continue
+			}
+			// Unterminated class: match "[" literally.
+			b.WriteString(regexp.QuoteMeta("["))
+			i++
+		default:
+			b.WriteString(regexp.QuoteMeta(string(c)))
+			i++
+		}
+	}
+	b.WriteString(`\z`)
+	return regexp.Compile(b.String())
+}
+
+// classEnd returns the index of the "]" closing the character class opened at
+// start, or -1 if the class is unterminated. A "]" in the first position
+// (optionally after a leading "^" negation) is treated as a literal member,
+// matching glob semantics.
+func classEnd(pattern string, start int) int {
+	i := start + 1
+	if i < len(pattern) && pattern[i] == '^' {
+		i++
+	}
+	if i < len(pattern) && pattern[i] == ']' {
+		i++
+	}
+	for ; i < len(pattern); i++ {
+		switch pattern[i] {
+		case '\\':
+			i++
+		case ']':
+			return i
+		}
+	}
+	return -1
 }
--- a/pkg/permissions/permissions.go
+++ b/pkg/permissions/permissions.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/docker/docker-agent/pkg/config/latest"
 )
@@ -240,17 +241,23 @@ func argToString(v any) string {
 
 // matchGlob checks if a value matches a glob pattern.
 // Supports glob-style patterns:
-// - "*" matches any sequence of characters (including path separators and spaces)
-// - "?" matches any single character
+// - "*" matches any sequence of characters, path separators and newlines included
+// - "?" matches any single character, path separator or newline included
 // - "[...]" matches character classes
+// - "\x" matches the literal character x
 //
 // Matching is case-insensitive.
 //
 // Note: filepath.Match's "*"/"?" stop at path separators ("/", or "\" on
-// Windows), but argument values (shell commands, file paths, URLs) routinely
-// contain "/", so matching must not depend on it. We therefore translate the
-// glob to an anchored regexp instead of using filepath.Match. Trailing
-// wildcards keep a plain prefix-match fast path (e.g. "sudo*").
+// Windows), but argument values are shell commands, file paths and URLs: they
+// routinely contain "/" and are often multi-line (heredocs, "&&" chains), so
+// matching must not depend on either. We therefore translate the glob to an
+// anchored regexp rather than using filepath.Match. Trailing wildcards keep a
+// plain prefix-match fast path (e.g. "sudo*").
+//
+// One deliberate divergence from filepath.Match: "\" escapes the next character
+// on every platform (filepath.Match treats it as a literal on Windows), and it
+// does so both inside and outside character classes.
 func matchGlob(pattern, value string) bool {
 	// Normalize both to lowercase for case-insensitive matching
 	pattern = strings.ToLower(pattern)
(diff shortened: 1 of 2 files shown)


pkg/permissions/permissions.go
```diff
@@ -266,7 +267,70 @@ func matchGlob(pattern, value string) bool {
 		}
 	}
 
-	// Try glob pattern match (also handles exact matches)
-	matched, err := filepath.Match(pattern, value)
-	return err == nil && matched
+	// Full glob match (also handles exact matches). Unlike filepath.Match, the
+	// translated "*"/"?" span any character, so patterns match values that
+	// contain path separators.
+	re, err := globToRegexp(pattern)
```
Sayt-0: `regexp.Compile` runs on every `matchGlob` call, i.e. per pattern per tool call (tool name plus each argument). Patterns are static for a `Checker`'s lifetime, so regexps could be compiled once at construction, or memoized in a package-level `sync.Map` keyed by the lowered pattern. Fine as a follow-up; not blocking.

pkg/permissions/permissions.go
```diff
@@ -266,7 +267,70 @@ func matchGlob(pattern, value string) bool {
...
+
+// globToRegexp translates a filepath.Match-style glob into an anchored regular
+// expression. Unlike filepath.Match, the resulting "*" and "?" match any
+// character, including path separators — argument values such as shell commands,
+// file paths and URLs routinely contain "/", and permission patterns must match
+// them regardless. Character classes ("[...]"), whose bracket syntax glob and
+// regexp share, are copied through verbatim; every other character is matched literally.
+func globToRegexp(pattern string) (*regexp.Regexp, error) {
+	var b strings.Builder
+	b.WriteString(`\A`)
```
Sayt-0: In Go regexp, `.` does not match `\n` unless the `s` flag is set, so the translated wildcards stop at newlines exactly the way `filepath.Match` stopped at `/`. Shell commands are routinely multi-line (heredocs, `&&` chains), so a non-trailing-wildcard deny rule still fails open for them, and some values that matched on `main` stop matching:

| pattern | value | `main` | this branch |
|---|---|---|---|
| `*rm -rf*` | `"echo hi\nrm -rf /"` | no match | no match (still fails open) |
| `*b*` | `"a\nb"` | match | no match (regression) |
| `a?b` | `"a\nb"` | match | no match (regression) |

(`filepath.Match` wildcards stop only at path separators; they do cross `\n`, hence the regression rows.)

Fix: ``b.WriteString(`(?s)\A`)``. Included in the full rewrite suggested in the comment below on the `default:` branch.

pkg/permissions/permissions.go
```diff
@@ -266,7 +267,70 @@ func matchGlob(pattern, value string) bool {
...
+		switch c := pattern[i]; c {
+		case '*':
+			b.WriteString(`.*`)
+			i++
+		case '?':
+			b.WriteByte('.')
+			i++
+		case '[':
+			if end := classEnd(pattern, i); end > 0 {
+				b.WriteString(pattern[i : end+1])
```
Sayt-0: Two semantic notes on copying class bodies verbatim:

- RE2 accepts more inside `[...]` than glob: `[\d]` meant literal `d` under `filepath.Match` but is the digit class in RE2 (`d` no longer matches, `5` does). Perl/POSIX classes silently leak into glob syntax. Worth either escaping `\` inside copied classes or documenting the extension.
- The leading-`]`-is-literal rule (`[]a]`, `[^]]`) is POSIX glob behavior, not `filepath.Match` behavior (`main` returned `ErrBadPattern`, so no match). Arguably an improvement, but the PR description's "keeps all existing matchGlob semantics" does not hold for these; a line in the doc comment would help.

pkg/permissions/permissions.go
```diff
@@ -266,7 +267,70 @@ func matchGlob(pattern, value string) bool {
...
+				i = end + 1
+				continue
+			}
+			// Unterminated class: match "[" literally.
+			b.WriteString(regexp.QuoteMeta("["))
+			i++
+		default:
+			b.WriteString(regexp.QuoteMeta(string(c)))
```
Sayt-0: `c` is a single byte. For a multi-byte UTF-8 rune, `string(c)` converts each byte to its own (wrong) rune, so the compiled regexp contains mojibake and literal non-ASCII patterns stop matching:

| pattern | value | `main` | this branch |
|---|---|---|---|
| `café` | `café` | match | no match (emitted regexp: `\AcafÃ©\z`) |
| `*café*` | `xx café yy` | match | no match |

Byte-wise scanning itself is safe (all glob metacharacters are ASCII, UTF-8 continuation bytes are >= 0x80); the fix is to accumulate literal runs and quote whole substrings instead of one byte at a time. Suggested replacement (also includes the `(?s)` fix):

```go
func globToRegexp(pattern string) (*regexp.Regexp, error) {
	var b strings.Builder
	// (?s): argument values are often multi-line, "." must match "\n" too.
	b.WriteString(`(?s)\A`)
	lit := 0
	flush := func(end int) {
		if lit < end {
			b.WriteString(regexp.QuoteMeta(pattern[lit:end]))
		}
	}
	for i := 0; i < len(pattern); {
		switch pattern[i] {
		case '*':
			flush(i)
			b.WriteString(`.*`)
			i++
			lit = i
		case '?':
			flush(i)
			b.WriteByte('.')
			i++
			lit = i
		case '[':
			if end := classEnd(pattern, i); end > 0 {
				flush(i)
				b.WriteString(pattern[i : end+1])
				i = end + 1
				lit = i
				continue
			}
			// Unterminated class: "[" stays in the literal run.
			i++
		default:
			i++
		}
	}
	flush(len(pattern))
	b.WriteString(`\z`)
	return regexp.Compile(b.String())
}
```

This variant passes the existing `TestMatchGlob` table, the new separator-crossing cases, and the newline/UTF-8 cases from this review.

pkg/permissions/permissions.go
```diff
@@ -266,7 +267,70 @@ func matchGlob(pattern, value string) bool {
...
+	i := start + 1
+	if i < len(pattern) && pattern[i] == '^' {
+		i++
+	}
+	if i < len(pattern) && pattern[i] == ']' {
+		i++
+	}
+	for ; i < len(pattern); i++ {
+		switch pattern[i] {
+		case '\\':
```
Sayt-0, line 410: `classEnd` honors `\]` escapes inside a class, but outside classes `\` is now always a literal backslash. On `main`, `filepath.Match` treated `\` as an escape on non-Windows (`foo\*` matched literal `foo*`, and the fast-path comment mentions exactly that case) and as a literal on Windows. This branch makes it a literal everywhere, which is a defensible unification but is a behavior change on non-Windows. Options:

1. Handle `\x` outside classes as literal `x` (one extra `case '\\':` in the translator), keeping non-Windows `main` behavior and making escapes work consistently inside and outside classes.
2. Keep backslash-as-literal and document the divergence in the `matchGlob` doc comment.

Either works; the current state (escape inside classes, literal outside) is the one combination that is hard to explain.

pkg/permissions/permissions_test.go
```diff
@@ -544,6 +544,12 @@ func TestMatchGlob(t *testing.T) {
 		{"mcp:*", "mcp:github:create_issue", true}, // * matches everything including :
 		{"mcp:github:*", "mcp:github:create_issue", true},
 		{"mcp:github:create_*", "mcp:github:create_issue", true},
+
+		// "*" and "?" span path separators. Argument values (shell commands,
+		// paths, URLs) routinely contain "/", so wildcards must not stop at it.
+		{"*rm -rf*", "rm -rf /", true},
+		{"*/etc/*", "cat /etc/passwd", true},
+		{"a?c", "a/c", true},
```
Sayt-0, line 552: The new rows cover the `/` boundary but not the two other boundaries the rewrite is sensitive to. Suggested additions:

```go
// "*" and "?" must also span newlines: shell commands are often multi-line.
{"*rm -rf*", "echo hi\nrm -rf /", true},
{"a?b", "a\nb", true},
// Multi-byte UTF-8 in patterns must keep matching literally.
{"café", "café", true},
{"*café*", "xx café yy", true},
```

All four rows fail on this branch as-is and pass with the fixes from the other comments.

Sayt-0 (changes requested):
The diagnosis in #3604 is correct and the glob-to-regexp translation is the right direction: RE2 is linear-time (no ReDoS), the `\A`/`\z` anchoring is right, and the fast path is preserved. Two correctness issues in `globToRegexp` block the merge; both have small fixes, detailed in the inline comments.

| # | Finding | Severity | Effect |
|---|---------|----------|--------|
| 1 | Translated regexp lacks the `(?s)` flag, so `.` and `.*` stop at `\n` | blocker | multi-line commands still bypass non-trailing-wildcard deny rules; strict regression vs `main` for newline-containing values |
| 2 | Byte-wise `QuoteMeta(string(c))` corrupts multi-byte UTF-8 runes | blocker | patterns containing non-ASCII stop matching (`café` no longer matches `café`) |
| 3 | `\` escape semantics drift outside character classes | minor | `foo\*` matched literal `foo*` on `main` (non-Windows), now matches `foo\` + anything |
| 4 | Classes copied verbatim expose RE2 Perl classes | minor | `[\d]` meant literal `d` under `filepath.Match`, now means any digit |
| 5 | Regexp compiled on every `matchGlob` call | nit | avoidable, patterns are static per `Checker` |

Finding 1 matters most: it recreates the exact failure class this PR sets out to fix, with the boundary moved from `/` to `\n`.

Pre-existing and out of scope, noted for a possible follow-up: a pattern that fails to compile silently disables its rule (`return false`, same direction as `ErrBadPattern` on `main`), so a typo in a deny rule fails open; validating patterns at config load would surface this.

pkg/permissions/permissions.go
parveshsaini: Did it rather than defer, it was small 😅 
Package-level `sync.Map` keyed by the lowered pattern, so a rule compiles once. Cache is bounded by the number of configured rules since patterns come from config, not from tool input.

pkg/permissions/permissions.go
parveshsaini: Added your regression rows: `{"*b*", "a\nb", true}` and `{"a?b", "a\nb", true}` (both were regressions vs `main`), 
plus `*rm -rf*` against `"echo hi\nrm -rf /"`, and a multi-line case in the deny test since that's the whole point of the PR.

pkg/permissions/permissions.go
parveshsaini: Both fixed rather than documented. Class bodies are translated member-by-member now: glob `\x` escapes resolve
to the literal, and the other members get escaped, so `[\d]` is the literal `d` and `[[:alpha:]]` can't smuggle
in a POSIX class. And I dropped my leading-`]`-is-literal rule , you're right it was POSIX, not
`filepath.Match`, so `[]a]` is back to matching nothing like it did under `ErrBadPattern`. Table has `[\d]`
against both `d` and `5`.

pkg/permissions/permissions.go
parveshsaini: Good catch! `string(c)` on a byte was just wrong. Took your rewrite , accumulating literal runs and quoting
whole substrings. Kept the byte-wise scan for the reasons you gave (metacharacters are all ASCII, continuation
bytes are >= 0x80) and noted that in the doc comment so the next reader doesn't "fix" it back. `café` and
`*café*` are in the table now.

pkg/permissions/permissions.go
parveshsaini: Went with option 1. `\x` outside a class is now a literal `x`, so escapes behave the same on both sides of a
bracket and `foo\*` matches literal `foo*` again like it did on `main` (non-Windows). You were right that the
old split , escape inside, literal outside , was the one combination nobody could explain. The remaining
Windows divergence (`\` was a literal there) is deliberate and now in the doc comment.

pkg/permissions/permissions_test.go
parveshsaini: All four added, plus `{"*b*", "a\nb", true}` and the escape/class rows (`foo\*` vs `foo*`/`foobar`, `[\d]` vs
`d`/`5`). Also extended `TestDenyGlobCrossesPathSeparator` with the multi-line command. 

Confirmed they fail on the previous commit and pass now 👍

parveshsaini:
Hey @Sayt-0! I've addressed the requested changes here. Would appreciate a quick review whenever you get the chance! 

Many Thanks 🚀

Sayt-0 (approved):
LGTM ! Thank you for the contribution ! One more thing : can you sign your commits ?

parveshsaini:
> LGTM ! Thank you for the contribution ! One more thing : can you sign your commits ?

Many Thanks @Sayt-0! I just signed them 🚀

Sayt-0:
CI is currently red, so this cannot be merged as is. The failure is not caused by this PR's code, but it still needs to be resolved on this branch.

| Item | Detail |
|---|---|
| Failing job | `build-and-test` ([run 29446416662](https://github.com/docker/docker-agent/actions/runs/29446416662/job/87552352197)) |
| Failing test | `TestDocYAMLSnippetsAreValid/../../docs/tools/scheduler/index.md:74` (`pkg/config`) |
| Related to this diff | no; the PR only touches `pkg/permissions`, whose suite is green |
| Cause | the branch is ~200 commits behind `main`; the CI merge snapshot hit a docs/schema mismatch on `main` that has since been fixed |
| Verified | with current `main` merged into this branch, `go test ./pkg/config/` and `go test ./pkg/permissions/` both pass |

Action needed before merge:

- Rebase the branch onto current `main` (or merge `main` into it) and push, so CI re-runs against a fresh merge snapshot.
- Keep the commits signed when rebasing (a rebase rewrites them).

PR #3633, opened 2026-07-13
Fix leak connections when fetching JWT token

### LINKED ISSUE #3629: File descriptor / unix-socket leak: RawClient builds a new http.Transport per call to Docker Desktop backend.sock
## Summary
A long-running `docker-agent run` process leaks unix-domain socket file descriptors steadily. A 9.5-hour session was observed holding **1,028 open unix-domain sockets** (FD numbers contiguously spanning 6→1071), climbing toward the process FD ceiling.

## Root cause
`pkg/desktop/raw_client.go`: `RawClient.client` is a `func() *http.Client` that constructs a **brand-new `http.Client` + bare `http.Transport` on every call**. Each call dials Docker Desktop's `backend.sock`. With HTTP keep-alives enabled and **no `IdleConnTimeout` set**, the connection is parked in the throwaway transport's idle pool and pinned indefinitely by its `persistConn` goroutines → **1 leaked unix FD (+2 goroutines) per call**.

## Primary trigger — one leak per LLM request (Gateway mode)
In Docker AI Gateway mode, per-request auth-token closures call `base.GatewayAuthToken` on every request, then unmemoized `desktop.GetToken` and `ClientBackend.Get`.

Secondary drips include 1-minute `/ping` probes, unmemoized `IsDockerDesktopRunning`, and 5-minute `GetVersion`.

MCP client pooling and stdio child reaping were verified correct and are not the cause.

## Impact
Long-lived sessions accumulate unix-socket FDs and goroutines without bound, risking `EMFILE` / “too many open files”.

## Proposed fix
Build the `RawClient` HTTP client/transport once and reuse it, with a bounded idle pool (for example `MaxIdleConns: 2`, `IdleConnTimeout: 90s`). Optionally memoize `desktop.GetToken` and cache the gateway auth client per provider. Add a regression test asserting open unix-socket connections stay flat across many `Get` calls.

## Evidence
A 48-second sampling window showed FD count flat at 1028 while mid-turn; the leak is tied to discrete LLM requests. FD breakdown: 1,028 unix sockets of 1,088 total FDs, zero established TCP; RSS was ~450MB.

dgageot:
Fix #3629

### DIFF
--- a/pkg/desktop/raw_client.go
+++ b/pkg/desktop/raw_client.go
@@ -17,17 +17,20 @@ type RawClient struct {
 }
 
 func newRawClient(dialer func(ctx context.Context) (net.Conn, error)) *RawClient {
-	return &RawClient{
-		client: func() *http.Client {
-			return &http.Client{
-				Transport: &http.Transport{
-					DialContext: func(ctx context.Context, _, _ string) (conn net.Conn, err error) {
-						return dialer(ctx)
-					},
-				},
-			}
+	client := &http.Client{
+		Transport: &http.Transport{
+			DisableKeepAlives: true,
+			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				return dialer(ctx)
+			},
 		},
+	}
+
+	return &RawClient{
 		timeout: 10 * time.Second,
+		client: func() *http.Client {
+			return client
+		},
 	}
 }
 


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly fixes the connection leak by moving to a single shared `*http.Client` instance with `DisableKeepAlives: true` on the transport.

**Analysis summary**: The old code recreated a new `*http.Client` (and therefore a new `*http.Transport`) on every call through the `client()` closure. Each transport maintained its own idle-connection pool, so connections were never reused and leaked after each JWT fetch. The fix:

1. Creates one `*http.Client` at construction time — safe, since `http.Client` is documented as goroutine-safe.
2. Adds `DisableKeepAlives: true` — the transport will not pool idle connections, so each connection is closed immediately after the response body is read. This directly addresses the leak.
3. The `func() *http.Client` indirection now returns a constant, which is a minor leftover abstraction but not a correctness concern.

No bugs were introduced by this change.

Sayt-0 (approved):

PR #3650, opened 2026-07-15
fix(browser): reap child process to avoid zombie after opening URL

### LINKED ISSUE #3630: Zombie child processes: browser.go calls Start() without Wait(), leaking defunct processes per URL open
## Summary
Opening a URL/browser leaves a permanent zombie (`<defunct>`) child process. A long-running session was observed with 5 zombie children parented to the `docker-agent` process.

## Root cause
`pkg/browser/browser.go` calls `exec.CommandContext(...).Start()` but never calls `Wait()`, so the child is never reaped after it exits and remains a zombie.

## Triggers
Every browser/URL open path, including MCP OAuth handshakes, TUI open-link, and the `open_url` tool. Five zombies correlated with five browser opens.

## Proposed fix
Reap the child asynchronously, following the existing pattern in `pkg/board/app.go`:

```go
go func() { _ = cmd.Wait() }()
```

## Impact
Accumulating zombies consume PIDs and indicate missing process lifecycle management; over long sessions this can contribute to process-table pressure.

oh-summy:
## Summary
`pkg/browser/browser.go` called `exec.CommandContext(...).Start()` without ever calling `Wait()`, leaving a permanent defunct (`<defunct>`) child process parented to `docker-agent` after every URL/browser open. Long-running sessions accumulated zombies — one observed session had 5 zombies from 5 browser opens (MCP OAuth handshake, TUI open-link, `open_url` tool).

## Root cause
The result of `exec.CommandContext(...).Start()` was used only for its error return; the `*exec.Cmd` was discarded, so the child was never reaped.

## Fix
Hold the `*exec.Cmd` in a variable and reap it asynchronously with a background `Wait()`, mirroring the established pattern in `pkg/board/app.go` (`OpenEditor`).

## Verification
- `go build ./pkg/browser/` ✅
- `go vet ./pkg/browser/` ✅
- `go test ./pkg/browser/...` ✅ (existing tests pass)

Fixes #3630

### DIFF
--- a/pkg/browser/browser.go
+++ b/pkg/browser/browser.go
@@ -32,10 +32,12 @@ func Open(ctx context.Context, urlToOpen string) error {
 		return fmt.Errorf("unsupported platform: %s", runtime.GOOS)
 	}
 
-	err := exec.CommandContext(ctx, cmd, args...).Start()
-	if err != nil {
+	command := exec.CommandContext(ctx, cmd, args...)
+	if err := command.Start(); err != nil {
 		return fmt.Errorf("failed to open browser: %w", err)
 	}
+	// Reap the child in the background so it does not linger as a zombie.
+	go func() { _ = command.Wait() }()
 
 	return nil
 }


Sayt-0 (approved):
lgtm ! Thank you for the contribution !

PR #3667, opened 2026-07-15
feat: custom base_url implies bypass_models_gateway

dgageot:
When a model is configured with a custom `base_url` — either set directly on the model or inherited from a named entry in the `providers:` section — it is clearly meant to reach a specific endpoint that is not the default for that provider. Routing such a model through a configured models gateway serves no purpose and often breaks things, yet until now callers had to remember to also set `bypass_models_gateway: true` explicitly.

This change makes the implication automatic. A custom (non-default) `base_url` now sets `bypass_models_gateway: true` at config resolution time, so the model dials its endpoint directly. Built-in provider alias defaults are unaffected. The predicate that decides whether a URL is "custom" is consolidated into `latest.HasCustomBaseURL` so the logic lives in one place. The env-var preflight now correctly requires direct credentials for these models even when a gateway is configured, and also expands `${env.X}` refs that come from an inherited provider-level `base_url`. `first_available` candidates that bypass the gateway — whether via the explicit flag or the new implied path — are credential-checked against the direct endpoint rather than being selected blindly.

No breaking change to existing configs: anyone who already sets `bypass_models_gateway: true` alongside `base_url` will see identical behavior.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1515,7 +1515,7 @@
         },
         "bypass_models_gateway": {
           "type": "boolean",
-          "description": "When true, this model connects directly to its provider, ignoring any configured models gateway (--models-gateway / CAGENT_MODELS_GATEWAY). The model then authenticates with the provider's own credentials (e.g. OPENAI_API_KEY, ANTHROPIC_API_KEY, or token_key) instead of routing through the gateway."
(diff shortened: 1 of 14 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

All findings from the drafter were dismissed upon verification. The logic in `HasCustomBaseURL`, the `routerFallbackLeaf` helper, and the `first_available` credential-checking path are all correct. The behavior change in `gather.go` (stripping router-level fields before passing the fallback leaf to env-var collection) is intentional and mirrors the runtime's actual behavior, as explicitly documented by the code comments.

**Summary**: The PR cleanly implements the implied-bypass invariant: a custom `base_url` on a model or inherited from a provider definition now automatically implies `bypass_models_gateway: true`. No regressions found.

aheritier (approved):

PR #3673, opened 2026-07-15
perf: speed up model switcher credential discovery

rumpl:
Opening the model picker checked every provider credential sequentially. Missing environment variables fall through the secret-provider chain and can spawn a macOS Keychain lookup for each key, causing a multi-second delay before the model picker opened.

Deduplicate credential names and resolve the independent lookups concurrently before filtering catalog providers. This preserves the existing provider availability checks while substantially reducing model-picker latency.

On my machine this went from ~3 second for the model switcher to open to about <1s, it's still not great but not sure what else we can do

### DIFF
--- a/pkg/runtime/model_switcher.go
+++ b/pkg/runtime/model_switcher.go
@@ -7,9 +7,11 @@ import (
 	"log/slog"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/docker/docker-agent/pkg/agent"
+	"github.com/docker/docker-agent/pkg/concurrent"
 	"github.com/docker/docker-agent/pkg/config/latest"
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The concurrent credential pre-fetch is implemented correctly. The `lookupEnvironmentValues` helper properly deduplicates names before spawning goroutines, uses a thread-safe `concurrent.Map` for writes, and drains all goroutines via `wg.Wait()` before the result map is built. No data races, goroutine leaks, or correctness issues were found in the new code.

**Reviewed**: `pkg/runtime/model_switcher.go` — concurrent credential discovery refactor

dgageot (approved):

PR #3674, opened 2026-07-15
Add model switching to the lean TUI

rumpl:
Add `/model` support to the lean TUI and share fuzzy model search between the lean and full TUIs.

### DIFF
--- a/pkg/tui/dialog/model_picker.go
+++ b/pkg/tui/dialog/model_picker.go
@@ -14,6 +14,7 @@ import (
 	tea "charm.land/bubbletea/v2"
 	"charm.land/lipgloss/v2"
 
+	"github.com/docker/docker-agent/pkg/modelpicker"
 	"github.com/docker/docker-agent/pkg/runtime"
 	"github.com/docker/docker-agent/pkg/tui/components/toolcommon"
 	"github.com/docker/docker-agent/pkg/tui/core"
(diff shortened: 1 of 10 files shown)


pkg/tui/dialog/model_picker.go
```diff
@@ -324,26 +325,17 @@ func (d *modelPickerDialog) filterModels() {
...
 
-	// If query looks like a custom model spec and we have no exact match, show it as an option
-	if isCustomQuery && len(d.filtered) == 0 {
-		d.filtered = append(d.filtered, runtime.ModelChoice{
+	// A provider/model query is an explicit reference rather than a fuzzy
+	// search when it does not identify an available choice.
+	if isCustomQuery && !slices.ContainsFunc(d.models, func(model runtime.ModelChoice) bool {
+		return strings.EqualFold(model.Ref, query) || strings.EqualFold(model.Provider+"/"+model.Model, query)
+	}) {
+		d.filtered = []runtime.ModelChoice{{
```
docker-agent: **[MEDIUM] Custom-model override replaces all fuzzy matches when query contains "/" but isn't an exact match**

In `filterModels()`, the comment says *"show Custom option as well as matches"* but the code **replaces** `d.filtered` entirely instead of appending to it:

```go
// A provider/model query is an explicit reference rather than a fuzzy
// search when it does not identify an available choice.
if isCustomQuery && !slices.ContainsFunc(d.models, func(model runtime.ModelChoice) bool {
    return strings.EqualFold(model.Ref, query) || strings.EqualFold(model.Provider+"/"+model.Model, query)
}) {
    d.filtered = []runtime.ModelChoice{{   // ← replaces, does not append
        Name: "Custom: " + query,
        Ref:  query,
    }}
}
```

The condition checks `d.models` (all models, for an **exact** match) — **not** `d.filtered` (fuzzy results already computed). So for a partial query like `openai/gpt`, `modelpicker.Filter` may have returned several relevant matches, but since `"openai/gpt"` is not an exact `EqualFold` match for any model's `Ref` or `Provider+"/"+Model`, those fuzzy results are discarded and only `"Custom: openai/gpt"` is shown.

**Suggested fix**: append the custom entry to the existing fuzzy results rather than replacing them, and only show the custom entry when fuzzy search returns nothing:

```go
if isCustomQuery && !slices.ContainsFunc(d.models, func(model runtime.ModelChoice) bool {
    return strings.EqualFold(model.Ref, query) || strings.EqualFold(model.Provider+"/"+model.Model, query)
}) {
    customEntry := runtime.ModelChoice{
        Name: "Custom: " + query,
        Ref:  query,
    }
    if len(d.filtered) == 0 {
        d.filtered = []runtime.ModelChoice{customEntry}
    } else {
        d.filtered = append(d.filtered, customEntry)
    }
}
```

docker-agent (commented):

One medium-severity issue found in the model picker filtering logic introduced by this PR.

pkg/tui/dialog/model_picker.go
rumpl: Good catch. This was a real issue: partial provider/model queries could produce useful fuzzy matches, but the custom entry replaced them. The picker now appends the custom entry after the fuzzy results, and the test covers retaining a fuzzy match alongside the custom option. Fixed in fe27014d3.

docker-agent: Thanks for the quick fix! The refactoring looks solid — extracting the fuzzy search into a shared package is a nice improvement, and appending the custom entry after fuzzy matches solves the replacement issue perfectly. The updated test also correctly validates that both fuzzy matches and the custom option are retained.

dgageot (approved):

PR #3625, opened 2026-07-13
feat(server): session-scoped elicitation sink for API/server runtimes (#3584)

aheritier:
feat(server): session-scoped elicitation sink for API/server runtimes (#3584)

## Summary

Lets elicitation (user-input) requests raised by API/server-created runtimes — including background jobs — reach a connected client, instead of being misclassified as headless and silently fast-declined. Part 4 (top) of the #3584 stack.

**Stacked on the TUI/app PR** — review the [diff against that branch](https://github.com/docker/docker-agent/compare/feat/tui-background-elicitations...feat/server-session-elicitation-sink).

## What changed

- **Session-scoped sink:** `runtimeForSession` registers a per-session `OnElicitationRequest` sink that appends elicitation events to a lazily created session event log, replayable via `GET /api/sessions/:id/events`.
- **Event-log lifecycle:** the lazy log is closed on session delete (single + batch), so a connected `/events` stream receives `session_exited` and terminates rather than blocking forever.
- **Docs:** API-server docs updated for the `/events` out-of-band elicitation flow on API-created sessions, and the answer-vs-raise wording on the POST endpoint.

## Testing

- Session-scoped sink wiring; delete (single + batch) closes a live `/events` stream; server-side background elicitations surface instead of fast-declining.
- Full stack `task dev` + `-race` — green.

PR #3678, opened 2026-07-16
fix(server): surface background elicitations over API

### LINKED ISSUE #3584: Concurrent background-job elicitations are silently dropped — nothing displayed to the user

Sayt-0:
## Summary

Complete the server/API portion of #3584 so elicitation requests raised by concurrent background jobs are replayable and answerable instead of being auto-declined when no per-request stream is active.

This supersedes #3625 with the same core session-scoped sink plus the lifecycle and concurrency hardening found during review.

## Issue expectations

|---|---|
| Display concurrent background elicitations | API-created runtimes append each request to a session-scoped `/events` log |
| Route responses to the correct job | `elicitation_id` is preserved through `POST /api/sessions/:id/elicitation` |
| Avoid silent drops | Event logs are created on demand and retain multiple requests in order |
| Avoid duplicate requests | Mirrored `RunStream` copies are skipped when the reliable sink already delivered the event |
| Handle session lifecycle safely | Single and batch deletion close lazy logs, clear server-owned sinks, and tombstone deleted IDs |
| Cover concurrency under the race detector | Regression tests exercise concurrent delivery, deletion, source registration, replay, and response routing |

## Implementation

- Register a per-session `OnElicitationRequest` sink for API/server-created runtimes.
- Create replayable event logs lazily for out-of-band events.
- Reuse existing attached logs rather than replacing buffered events or listeners.
- Serialize creation, source registration, and deletion to prevent stale callbacks from resurrecting leaked logs.
- Preserve exactly-once delivery when local runtimes mirror elicitation events onto `RunStream`.
- Document API event-log coverage and correlated responses.

## Validation

- `go test ./...`
- `go test -race ./pkg/server ./pkg/runtime/...`
- `go build ./...`
- `go vet ./pkg/server/... ./pkg/runtime/...`
- `golangci-lint run ./pkg/server/... ./pkg/runtime/...`
- `markdownlint-cli2 docs/features/api-server/index.md`

All checks pass. The full test suite was run with workspace proxy variables unset because the injected proxy intercepts the private-address SSRF tests.

Closes #3584

### DIFF
--- a/docs/features/api-server/index.md
+++ b/docs/features/api-server/index.md
@@ -46,7 +46,7 @@ All endpoints are under the `/api` prefix.
 | `GET`    | `/api/sessions/:id`                 | Get a session by ID (messages, tokens, permissions)     |
 | `GET`    | `/api/sessions/:id/status`          | Lightweight runtime state (streaming, title, agent, tokens). Requires an attached runtime. |
 | `GET`    | `/api/sessions/:id/snapshot`        | Full state in one call (stored fields + runtime state + `last_event_seq`) for gapless resync — see [Reconnecting without gaps](#reconnecting-without-gaps). |
-| `GET`    | `/api/sessions/:id/events`          | Live session event stream (SSE) with sequence numbers and replay. Available for a run attached via [`--listen`](#listen). |
+| `GET`    | `/api/sessions/:id/events`          | Live session event stream (SSE) with sequence numbers and replay. Available for a run attached via [`--listen`](#listen), or once a session has raised at least one out-of-band event (e.g. a background job's elicitation, answered via `POST .../elicitation`), which creates a session-scoped event log on demand carrying such out-of-band events — see [Session event stream](#session-event-stream-and-reconnection) for what each kind of log contains. |
 | `DELETE` | `/api/sessions/:id`                 | Delete a session                                        |
 | `PATCH`  | `/api/sessions/:id/title`           | Update session title                                    |
 | `PATCH`  | `/api/sessions/:id/permissions`     | Update session permissions                              |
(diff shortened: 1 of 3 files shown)


aheritier:

dgageot (approved):

PR #3679, opened 2026-07-16
Render Mermaid diagrams inline

rumpl:
Screenshot 2026-07-16 at 12 16 55

### DIFF
--- a/pkg/tui/components/mermaid/render_primitives.go
+++ b/pkg/tui/components/mermaid/render_primitives.go
@@ -1,0 +1,132 @@
+package mermaid
+
+import (
+	"strings"
+	"unicode/utf8"
+
+	"github.com/mattn/go-runewidth"
+
+	mermaidparser "github.com/docker/docker-agent/pkg/mermaid"
+)
+
+func mermaidCanvas(width int) []string {
+	canvas := make([]string, width)
+	for i := range canvas {
+		canvas[i] = " "
+	}
+	return canvas
+}
+
+func writeMermaidCanvas(canvas []string, start int, value string) {
+	for _, r := range value {
+		runeWidth := runewidth.RuneWidth(r)
+		if runeWidth == 0 {
+			for i := min(start-1, len(canvas)-1); i >= 0; i-- {
+				if canvas[i] != "" {
+					canvas[i] += string(r)
+					break
+				}
+			}
+			continue
+		}
+		if start >= 0 && start < len(canvas) {
+			canvas[start] = string(r)
+		}
+		for i := 1; i < runeWidth && start+i >= 0 && start+i < len(canvas); i++ {
+			canvas[start+i] = "" // Continuation cells must not add width when joined.
+		}
+		start += runeWidth
+	}
+}
+
+func mermaidCanvasText(canvas []string) string {
+	return strings.TrimRight(strings.Join(canvas, ""), " ")
+}
+
+func mermaidLifeline(order []string, centers map[string]int, width int) string {
+	canvas := mermaidCanvas(width)
+	for _, id := range order {
+		canvas[centers[id]] = "│"
+	}
+	return mermaidCanvasText(canvas)
+}
+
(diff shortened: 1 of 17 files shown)


docker-agent (commented):
### Assessment: 🔴 CRITICAL

**1 high-severity finding** in new Mermaid rendering code.

- [low] pkg/mermaid/parser.go:33 — Sequence diagrams with only notes (no messages) incorrectly rejected as ErrInvalidDiagram (confidence: weak 35/100)
- [low] pkg/mermaid/parser.go:69 — Escape check in splitMermaidStatements uses simple look-behind, mishandling double-escaped backslashes (confidence: negligible 25/100 — dropped)
- [low] pkg/mermaid/parser.go:144 — readDelimited byte-level escape check can misfire on multi-byte UTF-8 content (confidence: negligible 25/100 — dropped)
- [low] pkg/tui/components/mermaid/render_sequence.go:115 — Self-message label may overflow canvas width silently (confidence: negligible 20/100 — dropped)
- [low] pkg/tui/components/mermaid/render_graph.go:14 — append(roots, order...) may mutate roots backing array if it has spare capacity (confidence: negligible 20/100 — dropped)

pkg/tui/components/mermaid/render_primitives.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func writeMermaidCanvas(canvas []rune, start int, value string) {
+	for _, r := range value {
+		if start >= 0 && start < len(canvas) {
+			canvas[start] = r
+		}
+		start++
+	}
+}
```
docker-agent: **[high] `writeMermaidCanvas` skips a column for every wide Unicode character, corrupting diagram layout**

The loop advances `start` by 1 for every rune, but double-width characters (CJK, emoji) occupy **2 terminal columns**. After writing a wide rune to `canvas[start]`, the next character is placed at `start+1` — the second visual cell of the wide character — causing characters to visually overlap and misalign.

The rest of the file correctly uses `runewidth.StringWidth()` and `runewidth.RuneWidth()` for width accounting; this function is the odd one out. The fix is to advance by the rune's display width and, for the two-column case, also blank out the adjacent cell so it is not filled with stale content:

```suggestion
func writeMermaidCanvas(canvas []rune, start int, value string) {
	for _, r := range value {
		if start >= 0 && start < len(canvas) {
			canvas[start] = r
		}
		rw := runewidth.RuneWidth(r)
		if rw == 2 && start+1 >= 0 && start+1 < len(canvas) {
			canvas[start+1] = ' '
		}
		start += rw
	}
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

rumpl: Fixed in aba4fe758. I changed the canvas from `[]rune` to display-cell-aware `[]string` storage: wide runes reserve continuation cells that serialize to zero additional width, and zero-width combining runes are appended to the preceding cell. This avoids the extra terminal column that a literal space in the continuation cell would introduce. Added regression coverage for CJK and combining characters in `render_primitives_test.go`.

rumpl:
Review follow-up in `aba4fe758`:

- Fixed display-cell accounting for wide and combining Unicode characters and added regression tests.
- Allowed sequence diagrams containing notes without messages.
- Made quote escape detection handle runs of backslashes correctly.
- Avoided mutating the roots slice backing array during graph layout.

Validation passes with `task build`, `task test`, and `task lint`.

pkg/tui/components/mermaid/render_primitives.go
docker-agent: Excellent fix! The cell-aware storage approach is much cleaner than my suggestion. Storing wide-rune continuation cells as empty strings and appending zero-width combining characters to the preceding cell correctly handles both categories of non-trivial Unicode without introducing phantom width. The test coverage for both CJK and combining characters confirms it works as designed.

Thank you for the thorough implementation!

Sayt-0 (approved):

PR #3680, opened 2026-07-16
fix(teamloader): preserve deferred toolset instructions

Piyush0049:
### Description
This pull request fixes a logic bug in the teamloader where setting `defer_all: true` on a toolset would cause the entire toolset to be dropped from the agent's context.

Dropping the toolset had the unintended side effect of permanently deleting the toolset's custom `Instructions`, leaving the LLM completely unaware of when or how to ask for the deferred tools to be loaded.

### Changes Made
* Implemented `WithNoToolsFilter` in `pkg/teamloader/filter.go` to wrap toolsets.
* This new filter intercepts all tool requests and returns an empty slice, guaranteeing the agent sees zero tools.
* The filter safely delegates to the wrapped toolset for `Instructions`, ensuring the instructions are seamlessly preserved in the agent's system prompt.
* Replaced the destructive `continue` block in `pkg/teamloader/teamloader.go` with the new filter.
* Added comprehensive unit tests in `pkg/teamloader/filter_test.go`.

### Verification
* Ran `go test` and verified all unit tests pass.
* Verified no formatting or linting regressions with `go fmt` and `go vet`.

### DIFF
--- a/pkg/teamloader/filter.go
+++ b/pkg/teamloader/filter.go
@@ -81,3 +81,33 @@ func (f *filterTools) Tools(ctx context.Context) ([]tools.Tool, error) {
 
 	return filtered, nil
 }
+
+// WithNoToolsFilter creates a toolset that exposes no tools but preserves all other
+// capabilities (like Instructions) of the inner toolset.
+func WithNoToolsFilter(inner tools.ToolSet) tools.ToolSet {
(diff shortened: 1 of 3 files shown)


rumpl (approved):
nice

dgageot (approved):

PR #3681, opened 2026-07-16
feat(run): autodiscover project agent config

HajimohammadiNet:
Refs #2435

## Summary
- autodiscover `docker-agent.yaml` / `docker-agent.yml` / `docker-agent.hcl` for no-argument local `docker agent run`
- normalize discovered project configs before sandbox selection so `runtime.sandbox: true` is honored the same as explicit config paths
- keep explicit config arguments and `--remote` runs unchanged
- update user-facing docs for the new project-file-before-built-in-default behavior

## Why
Repositories can now check in a conventional project-level agent config and let developers run `docker agent run` without remembering a file path. This keeps the first slice focused on autodiscovery and leaves native `docker agent new` scaffolding for a follow-up.

## Testing
- `gofmt`
- `git diff --check`
- `go test -p=1 ./cmd/root -run TestDiscoverRunAgentArgs -count=1` in `golang:1.26.5`
- `go test -p=1 ./cmd/root -count=1` in `golang:1.26.5`
- `./scripts/build.sh` in `golang:1.26.5`
- `go mod tidy --diff` in `golang:1.26.5`
- `go run ./lint .` in `golang:1.26.5`
- `golangci-lint run --concurrency=2` with v2.12.2 in `golang:1.26.5`

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -48,6 +48,8 @@ import (
 // cobra's optional-value flags the two are indistinguishable by design.
 const worktreeAutoName = "auto"
 
+var projectDefaultAgentFiles = []string{"docker-agent.yaml", "docker-agent.yml"}
+
 type runExecFlags struct {
 	agentName         string
 	autoApprove       bool
@@ -327,10 +329,7 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		}
 	}()
 
-	var agentFileName string
-	if len(args) > 0 {
-		agentFileName = args[0]
-	}
+	agentFileName := f.resolveRunAgentFileName(args)
 
 	// Apply global user settings first (lowest priority)
 	// User settings only apply if the flag wasn't explicitly set by the user
@@ -528,6 +527,32 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		f.cleanupWorktree(context.WithoutCancel(ctx), out, createdWorktree)
 	}
 	return nil
+}
+
+func (f *runExecFlags) resolveRunAgentFileName(args []string) string {
+	if len(args) > 0 {
+		return args[0]
+	}
+	if f.remoteAddress != "" {
+		return ""
+	}
+	if agentFileName, ok := discoverProjectDefaultAgentFile(); ok {
+		return agentFileName
+	}
+	return ""
+}
+
+func discoverProjectDefaultAgentFile() (string, bool) {
+	for _, name := range projectDefaultAgentFiles {
+		info, err := os.Stat(name)
+		if err == nil && !info.IsDir() {
+			return name, true
+		}
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return name, true
+		}
+	}
+	return "", false
 }
 
 // loadTeamInWorktree creates the requested worktree (if any) and then loads
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -121,7 +121,7 @@ func newRunCmd() *cobra.Command {
 		Long:  "Run an agent with the specified configuration and prompt",
 		Example: `  docker-agent run ./agent.yaml
   docker-agent run ./team.yaml --agent root
-  docker-agent run # built-in default agent
+  docker-agent run # project config or built-in default agent
   docker-agent run coder # built-in coding agent
   docker-agent run ./echo.yaml "INSTRUCTIONS"
   docker-agent run ./echo.yaml "First question" "Follow-up question"
@@ -277,6 +277,9 @@ func (f *runExecFlags) runRunCommand(cmd *cobra.Command, args []string) (command
 		args = prependAgentRef(chosen, args)
 	}
 
+	var discoveredProjectConfig bool
+	args, discoveredProjectConfig = f.discoverRunAgentArgs(args)
+
 	// Resolve alias / runtime-declared sandbox opt-in before dispatch.
 	// An explicit --sandbox=<bool> on the CLI always wins, so we only
 	// consult the lower-priority sources when the flag wasn't set.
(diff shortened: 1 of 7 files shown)


rumpl:
and .hcl please

cmd/root/run.go
```diff
@@ -530,6 +529,32 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 	return nil
 }
 
+func (f *runExecFlags) resolveRunAgentFileName(args []string) string {
```
Sayt-0, line 538: Discovery at this layer runs after `runRunCommand` has already made the sandbox decision from `args[0]`:

- `resolveSandboxDefault(ctx, agentRef, ...)` receives `agentRef == ""` on a no-arg run, so `runtime.sandbox: true` in a discovered `docker-agent.yaml` is silently ignored. The same file passed explicitly runs sandboxed.
- On a `--sandbox` no-arg run, `runInSandbox` receives an empty `agentRef`: the kit build is skipped and `sandbox.EnvForAgent` forwards nothing. The inner run only works because it re-discovers the file inside the workspace, in a degraded mode compared to an explicit path.

The `--agent-picker` block already solves this shape of problem by normalizing `args` before dispatch. Discovery can reuse that pattern in `runRunCommand`, right after the picker block and before `resolveSandboxDefault`:

```go
if len(args) == 0 && f.remoteAddress == "" {
	if name, ok := discoverProjectDefaultAgentFile(); ok {
		args = prependAgentRef(name, args)
	}
}
```

`resolveRunAgentFileName` then reduces to the previous `args[0]` read, and explicit, discovered, sandboxed, and remote runs all agree on which config is in play.

cmd/root/run.go
```diff
@@ -327,10 +329,7 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
 		}
 	}()
 
-	var agentFileName string
-	if len(args) > 0 {
-		agentFileName = args[0]
-	}
+	agentFileName := f.resolveRunAgentFileName(args)
```
Sayt-0, line 338: Discovery is silent: nothing in the output distinguishes a run that picked `./docker-agent.yaml` from one using the built-in default agent. Recording mode and worktrees both announce themselves via `out.Println`; a similar one-liner here keeps surprises down, and matters most for `--exec` where no TUI sidebar hints at the loaded agent:

```go
out.Println("Using project config: " + agentFileName)
```

Printed only when the name came from discovery, not from an explicit argument.

cmd/root/run.go
```diff
@@ -530,6 +529,32 @@ func (f *runExecFlags) runOrExec(ctx context.Context, out *cli.Printer, args []s
...
+	return ""
+}
+
+func discoverProjectDefaultAgentFile() (string, bool) {
+	for _, name := range projectDefaultAgentFiles {
+		info, err := os.Stat(name)
+		if err == nil && !info.IsDir() {
+			return name, true
+		}
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
```
Sayt-0, line 561: This branch reports a candidate as discovered even though `os.Stat` failed (e.g. permission error), so the downstream load surfaces a real error instead of silently running the built-in default next to an unreadable `docker-agent.yaml`. Failing loudly seems right, but two things are missing:

- a comment: without one the code reads like a bug, and the branch also means an unreadable `docker-agent.yaml` shadows a readable `docker-agent.yml` later in the list, which is a deliberate choice worth stating
- optionally a test locking the behavior; a self-referential symlink named `docker-agent.yaml` triggers a non-`ErrNotExist` stat error portably on unix (skip on Windows)

docs/features/cli/index.md
```diff
@@ -13,7 +13,7 @@ _Complete reference for all docker-agent command-line commands and flags._
 > [!TIP]
 > **No config needed**
 >
-> Running `docker agent run` without a config file uses a built-in default agent. Perfect for quick experimentation.
+> Running `docker agent run` without a config argument uses `docker-agent.yaml`, `docker-agent.yml`, or `docker-agent.hcl` from the current directory when present. Otherwise, it uses a built-in default agent that is perfect for quick experimentation.
```
Sayt-0, line 16: Three more spots still describe the pre-discovery behavior and now contradict this paragraph:

| Location | Current text |
|---|---|
| `docs/features/cli/index.md:630` | Agent References table, row "Default": "(no argument) uses built-in default agent" |
| `docs/index.md:277` | "Zero Config" section: "# Run the built-in default agent" |
| `cmd/root/run.go:124` | help example: `docker-agent run # built-in default agent` |

Sayt-0 (changes requested):
Autodiscovery matches #2435 (including the `.hcl` request), keeps explicit arguments and `--remote` untouched, and ships focused tests plus doc updates. One inconsistency blocks approval: discovery runs after the sandbox decision in `runRunCommand`, so a discovered config declaring `runtime.sandbox: true` silently runs unsandboxed. Details in the inline comments.

| # | Severity | Finding |
|---|----------|---------|
| 1 | blocking | `runtime.sandbox: true` in a discovered config is ignored; fix by normalizing `args` early, like `--agent-picker` does |
| 2 | should | Discovery is silent; print which project config was picked |
| 3 | nit | Stat-error branch needs a comment (and ideally a test) |
| 4 | nit | Three doc/help mentions still describe the old behavior |
| 5 | discussion | Trust boundary: a repo-provided config runs MCP `command` toolsets and `session_start` hooks with no consent gate; a direnv-style consent prompt could be a follow-up |

Verified on PR head: `go test -run TestResolveRunAgentFileName ./cmd/root` pass, `gofmt` clean, no conflicts with `main`.

HajimohammadiNet:
Thanks for the review. I pushed `636a73f0e` to address the actionable items:

- normalized discovered project configs into `args[0]` before sandbox resolution, so `runtime.sandbox: true` is honored the same way as an explicit config path
- print `Using project config: ...` only when a project config was autodiscovered
- documented the non-`ErrNotExist` stat path and added coverage for the shadowing behavior
- updated the stale help/docs mentions that still described the old built-in-default-only behavior

For the trust-boundary point around repo-provided configs running command toolsets/hooks, I left that as follow-up scope since it was marked as discussion and would need a broader consent model than this autodiscovery slice.

Sayt-0:
looks good now ! thanks for the contribution !

Sayt-0 (approved):

Sayt-0:
Can you sign your commits ?

HajimohammadiNet:
Done, thanks for the reminder. I rewrote the PR commits with SSH signatures and added the signing key to GitHub; the commits now show valid signatures.

PR #3647, opened 2026-07-15
feat(safer_shell): add safe-auto safety policy

trungutt:
### Why

The session-level safety choice today is binary: prompt for every non-read-only tool call, or `--yolo` and prompt for nothing. On the prompt-per-call side users get tired of confirming trivially safe reads (`ls`, `docker ps`, `git status`, `read_file`, …) and flip to `--yolo`. The next destructive call then runs unprompted.

This adds two middle-ground policies: `safe-auto` and `safer`. Together with the existing `strict` (default) and `unsafe`, they form a graduated escalation ladder the caller can walk one click at a time.

### Behavior matrix

| policy    | safe        | destructive | unknown     |
|-----------|-------------|-------------|-------------|
| unsafe    | allow       | allow       | allow       |
| **safer**     | **allow**       | ask         | **allow**       |
| **safe-auto** | **allow**       | ask         | ask         |
| strict    | ask         | ask         | ask         |

"safe / destructive / unknown" comes from the confirmation event's `blast_radius` metadata (see below). `safer` is a strict superset of `safe-auto` — it also auto-approves calls the classifier can't recognise.

### Where `blast_radius` comes from

Three layers, lowest priority first:

```
  tool annotations                   ← ReadOnlyHint → safe, DestructiveHint → high
       │
       ▼
  permission_request hook metadata
       │
       ▼
  preempt-yolo pre_tool_use hook     ← safer_shell classifier for shell
```

Every layer can override the one above it. `safer_shell` still owns per-command shell verdicts; MCP annotations cover everything else.

### API surface

- `SessionResponse` / `SessionSnapshotResponse` include `safety_policy`.
- `PATCH /api/sessions/:id/safety-policy` sets the policy mid-session.
- `POST /api/sessions/:id/resume` accepts two new `confirmation` values:
  - `approve-safe` → flips session to `safe-auto` and approves the pending call.
  - `approve-safer` → flips session to `safer` and approves the pending call.

Symmetric with the existing `approve-session` (which flips to `unsafe`).

PR #3685, opened 2026-07-16
fix(session): persist SafetyPolicy across turns

trungutt:
### Why

The per-session `SafetyPolicy` field (added in #3647) is supposed to let a caller opt into `safe-auto` or `safer` mid-turn — either by `PATCH`ing the field, or by resuming with `approve-safe` / `approve-safer` / `approve-session` — and have that choice survive to the next turn.

Today it doesn't. The choice is remembered in-memory for the rest of the current turn, then quietly forgotten when the next turn starts. The user gets prompted again for the exact tool calls their opt-in was meant to auto-approve.

```
Turn 1  RunSession loads session      safety_policy=""
        User resumes with approve-safe  (dispatcher mutates struct in memory)
        rest of turn works              (auto-approves safe tools)
        turn ends                       (mutation was never persisted)

Turn 2  RunSession loads session      safety_policy=""   ← forgot the opt-in
        Asks for approve-safe again
```

The value falls into a gap between three separate pieces of the persistence path:

1. The sessions table has no column for `safety_policy`, so every write drops it and every read returns `""`.
2. Even with a column, both `UpdateSession` snapshot builders (SQLite + in-memory) rebuild the outgoing session field-by-field and neither of them copied `safety_policy`.
3. Even with the snapshots fixed, the dispatcher mutates `safety_policy` mid-turn from `handleResume`, but nothing calls `UpdateSession` at that point — the observer only writes on `OnRunStart`.

Each of the three gaps has to close for the round-trip to actually work; fixing any subset in isolation still loses the value.

### What this PR changes

- A new migration adds the `safety_policy TEXT DEFAULT ''` column and threads it through `sessionSelectColumns`, `scanSession`, and every INSERT / ON CONFLICT UPDATE in the store.
- Both `UpdateSession` snapshot builders now copy `SafetyPolicy` alongside the neighbouring `ToolsApproved` — the same symmetric persistence pattern that column already uses.
- `SessionManager.ResumeSession` calls `UpdateSession` synchronously for the three mid-turn transitions (`approve-safe`, `approve-safer`, `approve-session`) before forwarding the resume signal, so the mutation is durable before the dispatcher goroutine gets a chance to run.
- `SetSessionSafetyPolicy` mirrors the value onto the live runtime session when one is attached, so an explicit `PATCH` takes effect on the very next tool call instead of only after the next `RunSession`.

### Notes for the reviewer

- No behaviour change for callers that never set `SafetyPolicy` — the field still defaults to `''` (equivalent to `strict`).
- The `ResumeSession` mutation is intentionally redundant with the dispatcher's own `SetSafetyPolicy` / `SetToolsApproved` call. They operate on the same pointer, so the second call is a no-op — the point is to trigger `UpdateSession` synchronously.

### DIFF
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -919,11 +919,8 @@ func (sm *SessionManager) ResumeSession(ctx context.Context, sessionID, confirma
 		return errors.New("session not found")
 	}
 
-	// approve-safe / approve-safer / approve-session mutate the session
-	// mid-turn from the dispatcher, but PersistenceObserver only persists
-	// on OnRunStart — the mutation would otherwise not survive to the
-	// next turn. Apply and persist synchronously here so the state is
-	// durable regardless of when the dispatcher processes the resume.
+	// Mirror + persist mid-turn session mutations synchronously —
+	// PersistenceObserver only persists on OnRunStart.
 	if rt.session != nil {
 		mutated := false
 		switch runtime.ResumeType(confirmation) {
@@ -936,6 +933,13 @@ func (sm *SessionManager) ResumeSession(ctx context.Context, sessionID, confirma
 		case runtime.ResumeTypeApproveSession:
 			rt.session.SetToolsApproved(true)
 			mutated = true
+		case runtime.ResumeTypeApproveTool:
+			// Skip when toolName is empty — the dispatcher's own
+			// fallback (pending tool call name) isn't reachable here.
+			if toolName != "" {
+				rt.session.AppendPermissionAllow(toolName)
+				mutated = true
+			}
 		}
 		if mutated {
 			if err := sm.sessionStore.UpdateSession(ctx, rt.session); err != nil {
@@ -1156,9 +1160,8 @@ func (sm *SessionManager) SetSessionSafetyPolicy(ctx context.Context, sessionID 
 	sm.mux.Lock()
 	defer sm.mux.Unlock()
 
-	// Mirror onto the live runtime session (if any) so the running
-	// dispatcher picks up the new policy on the very next tool call,
-	// then persist so it survives to the next turn.
+	// Mirror onto the live runtime session so the dispatcher picks up
+	// the new policy on the next tool call, not just the next turn.
 	if rt, ok := sm.runtimeSessions.Load(sessionID); ok && rt.session != nil {
 		rt.session.SetSafetyPolicy(policy)
 		return sm.sessionStore.UpdateSession(ctx, rt.session)
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -914,10 +914,35 @@ func (sm *SessionManager) ResumeSession(ctx context.Context, sessionID, confirma
 	sm.mux.Lock()
 	defer sm.mux.Unlock()
 
-	// Ensure the session runtime exists
 	rt, exists := sm.runtimeSessions.Load(sessionID)
 	if !exists {
 		return errors.New("session not found")
+	}
+
+	// approve-safe / approve-safer / approve-session mutate the session
+	// mid-turn from the dispatcher, but PersistenceObserver only persists
+	// on OnRunStart — the mutation would otherwise not survive to the
+	// next turn. Apply and persist synchronously here so the state is
+	// durable regardless of when the dispatcher processes the resume.
+	if rt.session != nil {
+		mutated := false
+		switch runtime.ResumeType(confirmation) {
+		case runtime.ResumeTypeApproveSafe:
+			rt.session.SetSafetyPolicy(session.SafetyPolicySafeAuto)
+			mutated = true
+		case runtime.ResumeTypeApproveSafer:
+			rt.session.SetSafetyPolicy(session.SafetyPolicySafer)
+			mutated = true
+		case runtime.ResumeTypeApproveSession:
+			rt.session.SetToolsApproved(true)
+			mutated = true
+		}
+		if mutated {
+			if err := sm.sessionStore.UpdateSession(ctx, rt.session); err != nil {
+				slog.WarnContext(ctx, "failed to persist mid-turn session state",
+					"session_id", sessionID, "confirmation", confirmation, "err", err)
+			}
+		}
 	}
 
 	rt.runtime.Resume(ctx, runtime.ResumeRequest{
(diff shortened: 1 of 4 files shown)


pkg/server/session_manager.go
```diff
@@ -914,12 +914,37 @@ func (sm *SessionManager) ResumeSession(ctx context.Context, sessionID, confirma
...
+		case runtime.ResumeTypeApproveSafe:
+			rt.session.SetSafetyPolicy(session.SafetyPolicySafeAuto)
+			mutated = true
+		case runtime.ResumeTypeApproveSafer:
+			rt.session.SetSafetyPolicy(session.SafetyPolicySafer)
+			mutated = true
+		case runtime.ResumeTypeApproveSession:
+			rt.session.SetToolsApproved(true)
+			mutated = true
+		}
```
docker-agent, line 943: **[medium] `ResumeTypeApproveTool` not handled — mid-turn per-tool permission grant silently not persisted**

The new switch covers `approve-safe`, `approve-safer`, and `approve-session`, but omits `approve-tool`. The dispatcher's `handleResume` calls `c.sess.AppendPermissionAllow(toolName)` for `ResumeTypeApproveTool`, which mutates `session.Permissions.Allow`. Without a matching case here, the `UpdateSession` call that makes the other three types durable is never reached for tool-level approvals.

The comment directly above the switch says *"Apply and persist synchronously here so the state is durable regardless of when the dispatcher processes the resume."* That guarantee is not met for `approve-tool`: a per-tool approval granted mid-turn will still be lost if the process restarts before `PersistenceObserver.OnRunStart` fires at the start of the next turn.

```suggestion
		switch runtime.ResumeType(confirmation) {
		case runtime.ResumeTypeApproveSafe:
			rt.session.SetSafetyPolicy(session.SafetyPolicySafeAuto)
			mutated = true
		case runtime.ResumeTypeApproveSafer:
			rt.session.SetSafetyPolicy(session.SafetyPolicySafer)
			mutated = true
		case runtime.ResumeTypeApproveSession:
			rt.session.SetToolsApproved(true)
			mutated = true
		case runtime.ResumeTypeApproveTool:
			toolToApprove := toolName
			if toolToApprove == "" {
				toolToApprove = confirmation // fallback; dispatcher uses req.ToolName with same fallback
			}
			rt.session.AppendPermissionAllow(toolToApprove)
			mutated = true
		}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

This PR correctly fixes the three identified persistence holes for `SafetyPolicy`: the new SQL migration, both snapshot builders, and the synchronous pre-persist in `ResumeSession` and `SetSessionSafetyPolicy`. One medium-confidence finding: the new mid-turn persistence switch omits `ResumeTypeApproveTool`, leaving per-tool approvals with the same durability gap the PR fixes for the other resume types.

pkg/server/session_manager.go
trungutt: Good catch — `approve-tool` has the same "mid-turn mutation, no durable write" shape and needed the same treatment. Fixed in f52743111.

One deviation from the suggestion: when `toolName` is empty, the fix skips the persist rather than falling back to `confirmation`. `confirmation` at that point is the literal string `"approve-tool"`, not a tool name — using it as a fallback would store a nonsense allow-rule. The dispatcher can fall back to `c.tc.Function.Name` (the pending call) because it has the call in scope; here at the SessionManager layer we do not. A caller that omits `toolName` still gets the correct in-runtime behaviour via the dispatcher fallback, they just lose durability for that specific call — which shouldn't happen in practice since every client that sends `approve-tool` also sends `toolName`.

docker-agent: You're absolutely right about the fallback — I missed that `confirmation` holds the resume type constant, not a tool name. Using it as a fallback would have stored `"approve-tool"` itself as an allow-rule, which is nonsense. The skip-when-empty approach is correct: callers that send `approve-tool` also send `toolName` in practice, so the durability gap only exists for malformed requests, and those still get the correct in-runtime behavior via the dispatcher's fallback to the pending call.

Thanks for catching that and for the detailed explanation of why the fallback doesn't translate across layers!

docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation correctly addresses all three persistence holes described in the PR:

1. **SQL migration** — Migration 24 adds `safety_policy TEXT DEFAULT ''` to the sessions table. The `DEFAULT ''` aligns with the Go zero value (`SafetyPolicy("") = strict`), so pre-existing rows are treated correctly on read.

2. **Snapshot builders** — Both `InMemorySessionStore.UpdateSession` and `SQLiteSessionStore.UpdateSession` now copy `SafetyPolicy` in the field-by-field snapshot, making the persistence path symmetric with `ToolsApproved`.

3. **ResumeSession mid-turn persistence** — The new `switch` block mirrors the dispatcher's mutation onto `rt.session` and synchronously calls `UpdateSession` before forwarding the resume signal. The warn-and-continue approach on `UpdateSession` failure is a deliberate, defensible design choice: returning an error here would silently drop the user's approval and leave the session hanging, which is a worse outcome than a transient DB write failure.

4. **SetSessionSafetyPolicy live mirroring** — The new early-return path correctly updates the live runtime session pointer under `sm.mux`, ensuring an explicit PATCH takes effect on the next tool call.

5. **Pinned migrations test** — `wantDigest` in `migrations_pinned_test.go` was correctly updated from the previous value to reflect migration 24 (confirmed via git diff against merge base).

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly closes all three gaps in the SafetyPolicy persistence round-trip:

1. **Migration** —  adds  and is correctly sequenced after migration 23.
2. **Store layer** — , , both  INSERT paths, and the  upsert are all updated symmetrically. The  snapshot builder is also updated.
3. **Manager layer** —  now calls  synchronously for all four mid-turn resume types before forwarding the resume signal, and  mirrors the policy onto the live runtime session immediately.

No bugs introduced by this PR were found.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly closes all three gaps in the SafetyPolicy persistence round-trip:

1. **Migration** — `024_add_safety_policy_column` adds `safety_policy TEXT DEFAULT ''` and is correctly sequenced after migration 23.
2. **Store layer** — `sessionSelectColumns`, `scanSession`, both `AddSession` INSERT paths, and the `UpdateSession` upsert are all updated symmetrically. The `InMemorySessionStore` snapshot builder is also updated.
3. **Manager layer** — `ResumeSession` now calls `UpdateSession` synchronously for all four mid-turn resume types before forwarding the resume signal, and `SetSessionSafetyPolicy` mirrors the policy onto the live runtime session immediately.

No bugs introduced by this PR were found.

melmennaoui (approved):

dgageot (approved):

PR #3686, opened 2026-07-16
Centralize Anthropic cache control

rumpl:
The centralized policy marks:

 1. The final system block
 2. The immediate/deferred tool boundary, when applicable
 3. The final two conversation messages

### DIFF
--- a/pkg/model/provider/anthropic/cache_control.go
+++ b/pkg/model/provider/anthropic/cache_control.go
@@ -1,0 +1,162 @@
+package anthropic
+
+import (
+	"slices"
+	"strings"
+
+	anthropic "github.com/anthropics/anthropic-sdk-go"
+
+	"github.com/docker/docker-agent/pkg/chat"
+	"github.com/docker/docker-agent/pkg/tools"
+)
+
+const maxCacheControlBreakpoints = 4
+
+// applyPromptCacheControl is the single policy point for cache breakpoints in
+// standard Anthropic requests. Conversion functions deliberately leave cache
+// control unset so a complete request cannot accumulate markers implicitly.
+func applyPromptCacheControl(source []chat.Message, requestTools []tools.Tool, params *anthropic.MessageNewParams) {
+	remaining := maxCacheControlBreakpoints
+	for _, index := range systemCacheBreakpoints(source) {
+		if remaining == 0 {
+			return
+		}
+		if index < len(params.System) {
+			params.System[index].CacheControl = anthropic.NewCacheControlEphemeralParam()
+			remaining--
+		}
+	}
+
+	if index, ok := toolCacheBreakpoint(requestTools); ok && remaining > 0 && index < len(params.Tools) && params.Tools[index].OfTool != nil {
+		params.Tools[index].OfTool.CacheControl = anthropic.NewCacheControlEphemeralParam()
+		remaining--
+	}
+
+	for i, marked := len(params.Messages)-1, 0; i >= 0 && remaining > 0 && marked < 2; i-- {
+		if setMessageCacheControl(&params.Messages[i]) {
+			remaining--
+			marked++
+		}
+	}
+}
+
+// applyBetaPromptCacheControl mirrors applyPromptCacheControl for the SDK's
+// beta request types. Keep all beta marker translation here rather than in the
+// individual system, message, and tool converters.
+func applyBetaPromptCacheControl(source []chat.Message, requestTools []tools.Tool, params *anthropic.BetaMessageNewParams) {
+	remaining := maxCacheControlBreakpoints
+	for _, index := range systemCacheBreakpoints(source) {
+		if remaining == 0 {
+			return
+		}
+		if index < len(params.System) {
+			params.System[index].CacheControl = anthropic.NewBetaCacheControlEphemeralParam()
+			remaining--
+		}
+	}
+
+	if index, ok := toolCacheBreakpoint(requestTools); ok && remaining > 0 && index < len(params.Tools) && params.Tools[index].OfTool != nil {
+		params.Tools[index].OfTool.CacheControl = anthropic.NewBetaCacheControlEphemeralParam()
+		remaining--
+	}
+
+	for i, marked := len(params.Messages)-1, 0; i >= 0 && remaining > 0 && marked < 2; i-- {
+		if setBetaMessageCacheControl(&params.Messages[i]) {
+			remaining--
+			marked++
+		}
+	}
+}
+
+func systemCacheBreakpoints(messages []chat.Message) []int {
+	lastBySection := make(map[chat.PromptSection]int)
+	blockCount := 0
+	for _, msg := range messages {
+		if msg.Role != chat.MessageRoleSystem {
+			continue
+		}
+
+		before := blockCount
+		if len(msg.MultiContent) > 0 {
+			for _, part := range msg.MultiContent {
+				if part.Type == chat.MessagePartTypeText && strings.TrimSpace(part.Text) != "" {
+					blockCount++
+				}
+			}
+		} else if strings.TrimSpace(msg.Content) != "" {
+			blockCount++
+		}
+		if msg.PromptSection != "" && blockCount > before {
+			lastBySection[msg.PromptSection] = blockCount - 1
+		}
+	}
+
+	indices := make([]int, 0, len(lastBySection))
+	for _, index := range lastBySection {
+		indices = append(indices, index)
+	}
+	slices.Sort(indices)
+	return indices
+}
+
+func toolCacheBreakpoint(requestTools []tools.Tool) (int, bool) {
+	if !containsDeferredTool(requestTools) {
+		return 0, false
+	}
+	immediateCount := 0
+	for _, tool := range requestTools {
+		if !tool.Deferred {
+			immediateCount++
+		}
+	}
+	if immediateCount == 0 {
+		return len(requestTools) - 1, len(requestTools) > 0
+	}
+	return immediateCount - 1, true
+}
+
+func setMessageCacheControl(msg *anthropic.MessageParam) bool {
+	if len(msg.Content) == 0 {
+		return false
+	}
+	block := &msg.Content[len(msg.Content)-1]
+	cacheControl := anthropic.NewCacheControlEphemeralParam()
+	switch {
+	case block.OfText != nil:
+		block.OfText.CacheControl = cacheControl
+	case block.OfToolUse != nil:
+		block.OfToolUse.CacheControl = cacheControl
+	case block.OfToolResult != nil:
+		block.OfToolResult.CacheControl = cacheControl
+	case block.OfImage != nil:
+		block.OfImage.CacheControl = cacheControl
+	case block.OfDocument != nil:
+		block.OfDocument.CacheControl = cacheControl
+	default:
+		return false
+	}
+	return true
+}
+
+func setBetaMessageCacheControl(msg *anthropic.BetaMessageParam) bool {
+	if len(msg.Content) == 0 {
+		return false
+	}
+	block := &msg.Content[len(msg.Content)-1]
+	cacheControl := anthropic.NewBetaCacheControlEphemeralParam()
+	switch {
+	case block.OfText != nil:
+		block.OfText.CacheControl = cacheControl
+	case block.OfToolUse != nil:
+		block.OfToolUse.CacheControl = cacheControl
+	case block.OfToolResult != nil:
+		block.OfToolResult.CacheControl = cacheControl
+	case block.OfImage != nil:
+		block.OfImage.CacheControl = cacheControl
+	case block.OfDocument != nil:
+		block.OfDocument.CacheControl = cacheControl
+	default:
+		return false
+	}
+	return true
+}
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -1516,12 +1516,6 @@ func New(opts ...Opt) *Session {
 	return s
 }
 
-func markLastMessageAsCacheControl(messages []chat.Message) {
-	if len(messages) > 0 {
-		messages[len(messages)-1].CacheControl = true
-	}
-}
-
 // buildInvariantSystemMessages builds system messages that are identical
 // for all users of a given agent configuration. These messages can be
 // cached efficiently as they don't change between sessions, users, or projects.
@@ -1686,8 +1680,8 @@ func (s *Session) buildSessionSummaryMessages(items []Item) ([]chat.Message, int
 //
 // This method intentionally bypasses GetMessages's agent-level
 // transformations — invariant system prompts, NumHistoryItems
-// trimming, old-tool-content truncation, whitespace normalization,
-// orphan-tool-call sanitization, and cache_control marking. None of
+// trimming, old-tool-content truncation, whitespace normalization, and
+// orphan-tool-call sanitization. None of
 // those belong in compaction input: the compactor needs the full,
 // untrimmed history (so the LLM can summarize what trimming would
 // have hidden), supplies its own system/user prompt, and runs through
@@ -1785,6 +1779,14 @@ func (s *Session) instructionMessages() ([]chat.Message, []InstructionUpdate) {
 	return initial, slices.Clone(s.InstructionContext.Updates)
 }
 
+func setPromptSection(messages []chat.Message, section chat.PromptSection) {
+	for i := range messages {
+		if messages[i].Role == chat.MessageRoleSystem {
+			messages[i].PromptSection = section
+		}
+	}
+}
+
 func (s *Session) GetMessages(a *agent.Agent, extraSystemMessages ...chat.Message) []chat.Message {
 	return s.getMessages(a, true, extraSystemMessages...)
 }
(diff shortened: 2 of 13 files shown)


melmennaoui (dismissed):

pkg/model/provider/anthropic/cache_control.go
```diff
@@ -0,0 +1,97 @@
...
+	if !containsDeferredTool(requestTools) {
+		return 0, false
+	}
+	immediateCount := 0
+	for _, tool := range requestTools {
+		if !tool.Deferred {
+			immediateCount++
+		}
+	}
+	if immediateCount == 0 {
```
docker-agent, line 112: **[low] `toolCacheBreakpoint` places a cache breakpoint at the last tool when all tools are deferred, despite no real immediate/deferred boundary existing**

When all tools are deferred, `containsDeferredTool` returns `true`, so `toolCacheBreakpoint` proceeds. It counts `immediateCount == 0` and falls through to:

```go
return len(requestTools) - 1, len(requestTools) > 0
```

This returns the index of the very last tool. But `orderToolsForDeferredLoading` handles the all-deferred case by **clearing every tool's `Deferred` flag** and treating them all as immediate — there is no actual immediate/deferred boundary in `params.Tools`. The cache breakpoint is placed at the last tool with no semantic meaning (it does not mark a "boundary between immediate and deferred" as intended).

The more consistent behaviour in this edge case would be to return `(0, false)` (no breakpoint), matching the "no deferred tools" path. Alternatively, add a comment explaining the intentional fallback semantics so future maintainers understand why the last tool receives a marker when none are truly immediate.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

pkg/session/session.go
```diff
@@ -1817,20 +1810,9 @@ func (s *Session) getMessages(a *agent.Agent, includeInstructionContext bool, ex
...
-	// Volatile extras (the daily date) live behind the same marker, which
-	// is acceptable: the cache simply rotates when the date rolls over,
-	// matching the behavior of the previous inline
-	// buildContextSpecificSystemMessages path.
-	if len(extraSystemMessages) > 0 {
-		messages = append(messages, extraSystemMessages...)
-		markLastMessageAsCacheControl(messages[len(messages)-len(extraSystemMessages):])
-	}
+	// (e.g. turn_start hook output) inserted before the conversation.
+	messages = append(messages, extraSystemMessages...)
```
docker-agent, line 1826: **[low] Removing `markLastMessageAsCacheControl` on `extraSystemMessages` reduces cache stability when volatile system messages are appended**

Before this PR, `getMessages` called `markLastMessageAsCacheControl` at three points:

1. After `invariantMessages` — stable anchor on the invariant block
2. After the full `messages` slice (including `instructionInitial`) — second stable anchor
3. After appending `extraSystemMessages` — volatile anchor on the transient block

The new code removes all three calls and relies entirely on `applyPromptCacheControl` placing a single `cache_control` marker on `params.System[last]`. When volatile extras are present (e.g. date/time, environment info, `turn_start` hook output), `params.System[last]` now points to a volatile block, so the cache breakpoint rotates on every turn (or every day for date-based extras). The previously-stable invariant anchor is gone.

**Impact**: Increased cache misses for sessions that always append `extraSystemMessages`, because the system-block breakpoint no longer sits at a stable position. This matches the PR's stated intent ("The final system block") and is a deliberate trade-off, but may be surprising if callers relied on the stable invariant cache for cost/latency efficiency.

Consider documenting this behaviour change in a comment inside `applyPromptCacheControl`, or re-introducing a stable intermediate breakpoint after the invariant messages if cache efficiency matters here.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The centralized cache-control policy is well-structured and correctly ordered (system extraction before marker application in both client.go and beta_client.go). Two minor findings noted inline.

pkg/session/session.go
dgageot: @rumpl this one was important, I think

docker-agent: Thanks for flagging this, David. The current implementation in `applyPromptCacheControl` uses `systemCacheBreakpoints` to place markers at the last block of each `PromptSection`, which means when volatile extras (date/time, environment, turn_start hooks) are present in the context section, the cache breakpoint moves with them on every turn.

The stable anchor after invariant messages is now replaced by a potentially volatile one. For sessions that frequently append dynamic system messages, this could increase cache misses compared to the old three-point approach. However, I notice the new `systemCacheBreakpoints` function is section-aware and places multiple breakpoints — one per section — which is actually an improvement over the old single invariant marker if the sections are used correctly to separate stable from volatile content.

The key question is whether the `PromptSection` tagging is being applied consistently (invariant vs context sections) throughout the prompt assembly in `getMessages`. If it is, then this is likely a net improvement despite my initial concern.

rumpl: @dgageot the bot was reviewing the old version.

it also wasn't that important and there are still some inaccuracies in the code, followup

rumpl: This also becomes unimportant if you use the new way of handling context changes

docker-agent: Thanks for the clarification, @rumpl. I'll close out this thread since it was based on stale code. Apologies for the noise — I should have caught the version mismatch.

Looking forward to seeing the follow-up changes that address the remaining issues.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The centralization of Anthropic cache control looks correct. The new `applyPromptCacheControl` / `applyBetaPromptCacheControl` functions cleanly consolidate cache marker placement into a single policy point.

Key observations:
- `systemCacheBreakpoints` correctly maintains a global `blockCount` across all non-empty system messages (regardless of `PromptSection`), so the indices it returns always align with the positions in `params.System` built by `extractSystemBlocks`.
- The `blockCount > before` guard correctly prevents recording a breakpoint when a system message contributes no blocks, avoiding stale indices.
- `toolCacheBreakpoint` handles edge cases (all-deferred, no-deferred, empty) safely.
- `setMessageCacheControl` correctly skips thinking/redacted-thinking blocks.
- The removal of the `CacheControl bool` field from `chat.Message` eliminates a stored-state risk cleanly.
- `setPromptSection` in `session.go` correctly guards on `Role == MessageRoleSystem`.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly centralizes Anthropic cache-control marking into a single policy point (`cache_control.go`), replacing scattered `CacheControl bool` flags with a semantically richer `PromptSection` field.

The new `systemCacheBreakpoints` logic correctly tracks the last block index per section, consistent with how `extractSystemBlocks` builds `params.System`. The `toolCacheBreakpoint` correctly handles both mixed and all-deferred tool cases. The `setPromptSection` helper is correctly scoped to system-role messages, and the `PromptSection json:"-"` tag is appropriate since the field is recomputed per-request.

No bugs introduced by the changed code.

pkg/model/provider/anthropic/cache_control.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	for i, marked := len(params.Messages)-1, 0; i >= 0 && remaining > 0 && marked < 2; i-- {
+		if setBetaMessageCacheControl(&params.Messages[i]) {
+			remaining--
+			marked++
+		}
+	}
+}
+
```
docker-agent, line 70: **[medium] Instruction-context cache breakpoint is lost when `extraSystemMessages` is also non-empty**

`systemCacheBreakpoints` tracks the last block index per `PromptSection` value in a plain `map[chat.PromptSection]int`. Both `instructionInitial` and `extraSystemMessages` are tagged with the same value — `PromptSectionContext` — so every iteration over an `extraSystemMessages` system block **overwrites** the entry set by `instructionInitial`. The result: `systemCacheBreakpoints` emits at most one cache breakpoint for both groups combined (the end of extras), never the separate breakpoint at the end of the instruction context.

**Before this PR** the old `getMessages` code called `markLastMessageAsCacheControl` three times independently:
1. After `invariantMessages` → breakpoint at end-of-invariant
2. After appending `instructionInitial` → breakpoint at end-of-instruction-context
3. After appending `extraSystemMessages` → breakpoint at end-of-extras

**After this PR** only two breakpoints are emitted: end-of-invariant (`PromptSectionInvariant`) and end-of-extras (`PromptSectionContext`). The instruction-context breakpoint is gone.

**Trigger:** any request where `InstructionContext` is populated *and* the caller passes at least one `extraSystemMessages` (e.g. `turn_start` hook output, `AddPromptFiles`, daily date injection).

**Impact:** The instruction context (potentially large RAG content, tool-set documentation) is no longer independently cached. Any change in `extraSystemMessages` — including a daily date rotation — invalidates the cache beyond the invariant section, forcing a full re-tokenisation of the instruction context on every new day and whenever ephemeral extras change. This increases latency and cost.

**Fix:** Give `instructionInitial` its own distinct `PromptSection` constant (e.g. `PromptSectionInstructionContext`) so the map stores two separate entries and both breakpoints survive:

```go
// pkg/chat/chat.go — add a third constant:
PromptSectionInstructionContext PromptSection = "instruction_context"
```

```go
// pkg/session/session.go — tag instructionInitial with the new constant:
setPromptSection(instructionInitial, chat.PromptSectionInstructionContext)
// keep: setPromptSection(extraSystemMessages, chat.PromptSectionContext)
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

This PR cleanly centralizes Anthropic cache-control marking into a single policy point (`cache_control.go`), removing scattered per-converter cache logic. The new `applyPromptCacheControl` / `applyBetaPromptCacheControl` functions correctly handle the final system block, tool boundary, and final two conversation messages. One behavioral regression was found: both `instructionInitial` and `extraSystemMessages` share the same `PromptSectionContext` enum value, so `systemCacheBreakpoints` can only emit one cache breakpoint for the combined group, losing the independent breakpoint that previously existed at the end of the instruction context.

pkg/model/provider/anthropic/cache_control.go
rumpl: This combination is possible through the `Session.GetMessages` API, but it is not reachable through the runtime path. `messagesWithDynamicContext` keeps the two modes mutually exclusive: when cache-stable prompts are disabled it clears `InstructionContext` and calls `GetMessagesWithoutInstructionContext(a, legacyExtras...)`; when enabled it populates `InstructionContext` and calls `GetMessages(a)` without extras. Date, git status, prompt files, etc. are therefore either legacy extras or instruction-context sources, never both.

Adding a third section would also consume another slot from Anthropic’s four-breakpoint budget and could displace a recent-message checkpoint. I would prefer to keep the two sections unless we decide that populated instruction context plus explicit extras is a supported `Session.GetMessages` use case.

docker-agent: You're absolutely right — I apologize for the false positive. I missed that `messagesWithDynamicContext` enforces mutual exclusivity between the two modes: when cache-stable prompts are enabled it calls `GetMessages(a)` *without* extras, and when disabled it clears `InstructionContext` and passes extras to `GetMessagesWithoutInstructionContext`. The overlapping `PromptSectionContext` tag is harmless since only one code path executes at runtime.

Your point about the breakpoint budget is also well taken — adding a third section would consume a slot unnecessarily when the API design already prevents the collision. Thank you for the detailed explanation and for catching this!

docker-agent (commented):

One medium-severity finding: the centralized cache-control policy shares a single `PromptSectionContext` tag across both `instructionInitial` and `extraSystemMessages`, collapsing what were previously two system cache checkpoints into one. See the inline comment for details.

pkg/session/session.go
```diff
@@ -1809,6 +1811,7 @@ func (s *Session) getMessages(a *agent.Agent, includeInstructionContext bool, ex
 	var instructionUpdates []InstructionUpdate
 	if includeInstructionContext {
 		instructionInitial, instructionUpdates = s.instructionMessages()
+		setPromptSection(instructionInitial, chat.PromptSectionContext)
```
docker-agent, line 1814: **[medium] Shared `PromptSectionContext` for instruction and extra system messages collapses two cache checkpoints into one**

Prior to this PR, `GetMessages` placed **two** system-prompt cache checkpoints when both `instructionInitial` and `extraSystemMessages` were present:
1. After the last instruction block (end of the stable per-session context)
2. After the last extra-system block (end of volatile per-turn extras like today's date)

After this refactor, both `instructionInitial` and `extraSystemMessages` share the same `PromptSectionContext` tag. In `systemCacheBreakpoints`, `lastBySection[PromptSectionContext]` is overwritten for each successive system message in that section, so only the final block — belonging to `extraSystemMessages` — survives as a checkpoint. The checkpoint at the end of `instructionInitial` is silently lost.

**Impact**: When a session has instruction context (e.g. `InstructionContext` is set) AND the caller supplies `extraSystemMessages`, the Anthropic prompt cache can no longer reuse the cached representation of the instruction section independently from the extras. Every volatile extra-system change (e.g. a daily-date message) now evicts the entire instruction+extra prefix from cache rather than only the extra suffix. This increases cache-miss re-processing costs proportionally to the size of the instruction section.

**Reproduction trigger**: any call path where `getMessages` is invoked with `includeInstructionContext=true` AND at least one non-empty `extraSystemMessages` (e.g. `turn_start` hook output, `AddEnvironmentInfo`, `AddPromptFiles`).

**Possible fix**: Assign `instructionInitial` its own distinct `PromptSection` value (e.g. `PromptSectionInstruction`) so `systemCacheBreakpoints` records it as a separate breakpoint:

```go
const (
    PromptSectionInvariant   PromptSection = "invariant"
    PromptSectionInstruction PromptSection = "instruction"  // new
    PromptSectionContext      PromptSection = "context"
)
```

and in `getMessages`:
```go
setPromptSection(instructionInitial, chat.PromptSectionInstruction)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

rumpl: This is the same mutually-exclusive-path scenario discussed in https://github.com/docker/docker-agent/pull/3686#discussion_r3596388356. The runtime never supplies both groups together: cache-stable mode calls `GetMessages(a)` without extras, while legacy mode clears `InstructionContext` and calls `GetMessagesWithoutInstructionContext(a, legacyExtras...)`. Consequently date, environment, and prompt-file context is represented by one mechanism or the other, not both, and the shared section does not collapse runtime checkpoints. A direct caller can construct the combination, but it is not a supported runtime path; adding another section would unnecessarily consume one of Anthropic’s four breakpoint slots.

docker-agent: I appreciate the clarification — I misread the runtime call paths. You're right that the two modes are mutually exclusive in practice:

- **Cache-stable mode**: uses `GetMessages` with `InstructionContext` set, no `extraSystemMessages`
- **Legacy mode**: clears `InstructionContext` and passes extras to `GetMessagesWithoutInstructionContext`

Because the runtime never combines both in a single call, there's no actual checkpoint collapse in production. The shared `PromptSectionContext` label doesn't create a problem, and reserving a separate `PromptSection` would indeed waste one of the four Anthropic breakpoint slots for a code path that isn't exercised.

I should have traced the call sites more carefully before flagging this. Thanks for the pointer to the other discussion — that context makes the design clear.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly centralizes Anthropic prompt-cache-control marking into a single `cache_control.go` policy module and replaces the persisted `CacheControl bool` field with a runtime-only `PromptSection` metadata tag. The refactoring is correct and well-structured.

**Key areas reviewed:**

- **`systemCacheBreakpoints`** — Traverses messages with the same `if/else` branching and text-block filtering as `extractSystemBlocks`, so the produced block indices exactly match the `params.System` array layout. The `index < len(params.System)` bounds guard provides a safety net.
- **`toolCacheBreakpoint`** — Computes `immediateCount` from the original `requestTools`; `convertTools` places immediate tools first via `orderToolsForDeferredLoading`, so `params.Tools[immediateCount-1]` correctly identifies the last immediate tool. The all-deferred edge case (returns `len(requestTools)-1`) is handled consistently.
- **`setPromptSection` + `slices.Clone`** — The clone before labeling `extraSystemMessages` correctly prevents mutation of the caller's backing array when a slice is spread as a variadic argument.
- **Removed `CacheControl bool`** — The old field had `json:"cache_control,omitempty"` so it was persisted; the new `PromptSection` field is `json:"-"` and computed at request time, which is the right approach. Existing sessions with stale `cache_control: true` JSON silently ignore the unknown field on unmarshal — no data loss.
- **Compactor / benchmark cleanups** — Removing `CacheControl = false` resets are correct since `PromptSection` is never stored and `CompactionInput()` returns messages without that field set.

docker-agent (commented):
### Assessment: 🟢 APPROVE

Sayt-0 (approved):

dgageot (changes requested):
wait

dgageot:
```
Two issues:

- Regression — system prompt caching silently disabled by hooks. The new chat.Message.PromptSection field is tagged json:"-", but before_llm_call hooks
  round-trip messages through JSON (pkg/hooks/executor.go → pkg/hooks/handler.go → pkg/runtime/loop.go). When a hook rewrites messages — notably the default-on
  redact_secrets hook — the section metadata is wiped, so Anthropic system-prompt caching is lost for that request (and every subsequent one if a secret stays in
  history). This was reproduced on the PR head; the old CacheControl field survived that path because it had a JSON tag.
- Source-breaking API change. Removing the exported chat.Message.CacheControl field breaks external Go consumers (all in-tree callers were updated, though).
```

PR #3689, opened 2026-07-16
fix: limit anthropic cache control markers

rumpl:
Fix Anthropic requests exceeding the provider’s limit of four cache-control breakpoints.

 Cache-control checkpoint selection now originates from the session, while the Anthropic provider translates those candidates into API markers and enforces Anthropic’s limit.

 What changed

 - Mark the two most recent conversation messages as cache-control candidates when assembling session messages.
 - Remove the Anthropic provider’s unconditional caching of the last two converted messages.
 - Apply Anthropic cache markers only when the source chat.Message has CacheControl enabled.
 - Propagate cache control correctly when consecutive tool-result messages are merged.
 - Limit each Anthropic request to four cache-control markers.
 - Reserve one marker slot when deferred tools require a cache breakpoint.
 - Preserve the stable prefix checkpoint and prioritize recent context when there are too many candidates.
 - Apply the same behavior to both standard and beta Anthropic APIs.
 - Refresh affected Anthropic title-generation cassettes.

### DIFF
--- a/e2e/testdata/cassettes/TestDebug_Title/Anthropic.yaml
+++ b/e2e/testdata/cassettes/TestDebug_Title/Anthropic.yaml
@@ -8,7 +8,7 @@ interactions:
         proto_minor: 1
         content_length: 0
         host: api.anthropic.com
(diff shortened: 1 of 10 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The cache-control limit enforcement is correct. The `limitCacheControlMarkers` function accurately preserves the prefix checkpoint and fills remaining slots with the most recent candidates — the slice arithmetic is sound. The session-layer marking (`markRecentConversationCacheControl`) correctly preserves system-message checkpoints while refreshing the two most recent conversation markers. The provider-layer limit acts as the authoritative enforcer, and both the standard and beta clients call it consistently. Cache propagation through merged tool-result batches is handled symmetrically in both converters.

PR #3690, opened 2026-07-16
fix(tui): refocus the editor after external editing so Enter sends the content

dgageot:
When `Ctrl+G` opens the external editor, the edited content is written back into the input box once the editor exits. The bug was that focus was not restored to the input panel at that point — so if the user had tabbed away or clicked the transcript while watching a running turn, pressing Enter would route to the transcript rather than the input box, silently discarding the composed message. The same problem prevented the message from being queued when the agent was busy and `busy_send_mode: queue` was configured.

The fix extracts the `tea.ExecProcess` callback into a package-level `externalEditorCallback` function in `pkg/tui/tui.go`. On a successful edit it now returns a `messages.RequestFocusMsg{Target: messages.PanelEditor}` alongside the content update, so the input editor always regains focus after an external edit. Error paths (editor exit error, temp-file read failure) are unchanged and continue to surface a notification without touching the input content.

Regression tests in `pkg/tui/external_editor_test.go` cover the happy path (content written back with trailing newline trimmed, refocus message returned), blank content clearing the editor, editor-exit errors surfacing a notification, temp-file cleanup on both success and error, and the temp-file read-failure path.

### DIFF
--- a/pkg/tui/external_editor_test.go
+++ b/pkg/tui/external_editor_test.go
@@ -75,3 +75,15 @@ func TestExternalEditorCallback_EditorError(t *testing.T) {
 	_, err := os.Stat(path)
 	assert.True(t, os.IsNotExist(err), "the temp file must be removed on error")
 }
+
+func TestExternalEditorCallback_ReadError(t *testing.T) {
+	ed := &valueRecordingEditor{value: "untouched"}
+	path := filepath.Join(t.TempDir(), "missing.md")
+
+	msg := externalEditorCallback(ed, path)(nil)
+
+	show, ok := msg.(notification.ShowMsg)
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is clean and correct. The extracted `externalEditorCallback` function properly returns `messages.RequestFocusMsg{Target: messages.PanelEditor}` on success, ensuring the input editor regains focus after external editing. Temp-file cleanup is present in all code paths. The new regression tests cover the four main cases (happy path, blank content, editor exit error, read failure) and are structurally sound.

No medium or high severity issues found in the changed code.

Sayt-0 (approved):

PR #3692, opened 2026-07-16
fix(anthropic): keep cache breakpoints within Anthropic's limit of 4 with deferred tools

dgageot:
Commit 83956fcb6 added a `cache_control` breakpoint on the tool list when deferred tools are present. Combined with the existing breakpoints applied to system blocks (up to 2) and the last conversation messages (up to 2), requests could exceed Anthropic's hard limit of 4 `cache_control` blocks and return API errors at runtime.

`applyMessageCacheControl` and `applyBetaMessageCacheControl` now accept a breakpoint budget. When deferred tools are in the request, only the last message receives a breakpoint instead of the last two, keeping the total at or below 4. The non-deferred path is unchanged: 2 system + 2 messages = 4. With deferred tools the worst case is 2 system + 1 tools + 1 message = 4. Cache efficiency is preserved because Anthropic's caching layer automatically looks back roughly 20 blocks for prior hits, so a single message breakpoint still gets cache hits on subsequent turns.

New tests assert that a worst-case request — two system breakpoints, deferred tools, and a multi-turn conversation — stays within the 4-block limit for both the standard and beta converters, and that the message breakpoint count drops from two to one when deferred tools are present.

### DIFF
--- a/pkg/model/provider/anthropic/beta_converter.go
+++ b/pkg/model/provider/anthropic/beta_converter.go
@@ -137,8 +137,14 @@ func (c *Client) convertBetaMessagesWithDeferred(ctx context.Context, messages [
 		}
 	}
 
-	// Add ephemeral cache to last 2 messages' last content block
-	applyBetaMessageCacheControl(betaMessages)
+	// Anthropic allows at most 4 cache_control breakpoints per request.
+	// Deferred tools consume one on the tool list, so keep a single message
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

rumpl (approved):

PR #3702, opened 2026-07-17
feat(budget): cap what an agent may spend in money, tokens, or working time

### LINKED ISSUE #3701: Add budgets to cap what an agent may spend in money, tokens, or working time
### Overview

There is no way to bound what an agent consumes. `max_iterations` caps how many turns a run takes, but a turn is not a unit of cost: ten turns against `gpt-4o` with large contexts cost far more than a hundred against `gpt-4o-mini`. An agent that fans out to sub-agents multiplies that further, and nothing stops a runaway or badly-prompted session from spending until someone notices and hits Ctrl+C.

This matters most exactly where supervision is thinnest - unattended jobs, CI, background agents, and demos handed to someone else.

### Motivation

The runtime **already computes everything needed** to enforce this. It just has no way to act on it:

- `computeMessageCost` (`pkg/runtime/loop.go`) prices every response from `usage` and the model's pricing table.
- Model pricing comes from the models.dev catalogue, or an explicit model-level `cost:` block (`CostConfig`).
- `after_llm_call` hooks receive that exact per-call `cost`, and the docs describe using it for "a sidecar cost ledger".

But a ledger can only **record**. `executeAfterLLMCallHooks` (`pkg/runtime/hooks.go`) calls `dispatchHook` and **discards the returned `*hooks.Result`** - both callsites (`loop.go`, `harness.go`) invoke it as a bare statement, and `pkg/runtime/after_llm_call_test.go` locks that in. `Continue: false` is only honored in `pkg/hooks/executor.go`, which the event never reaches.

So `after_llm_call` is observational by design: a hook can watch spend accumulate and can do nothing about it. Enforcement has to live in the runtime.

### Use cases

**1. A budget name is one shared pot.** When several agents reference `shell-work`, they draw from the *same* ceiling, not a copy each. This is the whole point: per-agent copies would let a run spend `max_cost` × N by fanning out to N sub-agents, and the ceiling would mean nothing for exactly the workloads that most need one. Distinct names give independent pots. The same applies to sub-sessions, which share the root's wallets - they run on the same `LocalRuntime` (`loop.go` already distinguishes `rootStream := !sess.IsSubSession()`), so a tracker installed there is reached by every child.

Note this **intentionally differs from `max_iterations`**, where per-child caps (`agent_delegation.go`) are the *right* semantics. Money doesn't behave like iterations, so that precedent must not be copied.

**2. A budget spans the session, not a message.** `RunStream` is invoked once per user message. Resetting there would silently redefine `max_cost` as "per message" - a session could spend the whole ceiling again on every turn, which is not a budget at all.

**3. `max_time` must measure working time, not wall-clock.** Follows from (2): a session sits idle while the user reads and types. Wall-clock would let a budget expire during a coffee break - leave the TUI open ten minutes and the next message instantly trips `max_time: 2m` without the agent having done anything. It has to accumulate turn durations.

**4. Tokens need their own accumulator.** `max_tokens` cannot read `session.Usage()`: `ApplyCompaction` resets those counters because they measure *context length*, not spend. Reading them would make `max_tokens` silently stop working on exactly the long runs it exists for.

**5. Unpriced models can't be enforced against.** `computeMessageCost` returns `nil` for a model with no pricing table (unknown ID, or a custom endpoint). There is no honest number to add, so such spend cannot count towards `max_cost`. Failing the run closed would break local/custom-endpoint users; counting it as `$0` would be a lie that reads reassuringly low precisely because it is incomplete. Warn once, mark the reading, and document `cost:` as the fix.

**6. Terminal, not resumable.** Unlike `max_iterations`, no "continue?" prompt. A budget is a ceiling set deliberately; offering to raise it defeats the point. Raising it means editing the config. This also avoids touching `resumeChan` and the headless consumers.

**7. Boundary-checked.** Checked between turns, like `max_iterations`, so a run overshoots by at most the turn in flight and `max_time` won't interrupt an in-flight call. A hard `context.WithTimeout` would be tighter but changes error semantics for in-flight tools; boundary checking is the conservative first cut and should be documented as such.

### Proposed solution

Two ways to declare a budget, composing:

```yaml
# One ceiling for the whole session, charged for every agent.
budget:
  max_cost: 0.50      # USD
  max_tokens: 100000  # cumulative input+output
  max_time: 10m       # time agents spend working

# Named budgets an agent opts into by name.
budgets:
  shell-work:
    max_cost: 0.03
  research:
    max_time: 1m

agents:
  root:
    budgets: [shell-work]
  researcher:
    budgets: [shell-work, research]   # shares shell-work with root
```

Every limit optional; unset means unlimited; declaring nothing leaves runs unbudgeted (the default, so existing configs are unaffected). Crossing any ceiling stops the run with a message naming the exact YAML path to raise, and the TUI tracks every budget live, broken down per agent.

The named form follows the existing top-level `mcps` / `rag` / `toolsets` reference-by-name convention.

### Alternatives

_No response_

### Related issues

_No response_

### Additional context

- `BudgetConfig`, `Config.Budget`, `Config.Budgets`, `AgentConfig.Budgets`; validation (negative limits, unknown budget names); `agent-schema.json` + an HCL block rule for the `budgets` map.
- A `budgetTracker` / `budgetSet` in `pkg/runtime`, installed once per runtime, recorded from the existing `computeMessageCost` value so the ceiling can never disagree with what the session bills.
- Enforcement beside `enforceMaxIterations`; a `budget_exceeded` stream end reason.
- `budget_usage` / `budget_exceeded` events; sidebar tracking with per-agent breakdown; warning on stop.
- Docs page, runnable `examples/budget.yaml`, tests.

`loopState`'s own doc comment already anticipates this feature by name - *"trivial to add new per-stream tracking (cost ceiling, token budget, turn timing)"*.

dwin-gharibi:
Adds budgets that stop an agent once it crosses a configured ceiling, and tracks every budget live in the TUI, broken down per agent.

```yaml
# One ceiling for the whole session, charged for every agent.
budget:
  max_cost: 0.50      # USD
  max_tokens: 100000  # cumulative input+output
  max_time: 10m       # time agents spend working

# Named budgets an agent opts into by name.
budgets:
  shell-work:
    max_cost: 0.03
  research:
    max_time: 1m

agents:
  root:
    budgets: [shell-work]
  researcher:
    budgets: [shell-work, research]   # shares shell-work with root
```

Every limit is optional, unset means unlimited, and declaring nothing leaves runs unbudgeted. **Existing configs are unaffected** - the feature is inert unless opted into.

docker agent budget

The named form follows the existing top-level `mcps` / `rag` / `toolsets` reference-by-name convention. An agent may list several budgets; all apply, and the first exhausted stops the run.

Closes #3701.

## Why this needs runtime code

The runtime already prices every response (`computeMessageCost`) and hands that exact figure to `after_llm_call` hooks, which the docs suggest using for "a sidecar cost ledger".

But a ledger can only record. `executeAfterLLMCallHooks` (`hooks.go`) **discards** the hook `Result` - both callsites invoke it as a bare statement, and `after_llm_call_test.go` locks that in. `Continue: false` is only honored in `executor.go`, which the event never reaches. So spend can be watched but not acted on; enforcement has to live in the loop.

This PR adds no new pricing math. It reuses the value the session already bills, so the ceiling can never disagree with the invoice.

## Design decisions

Each of these is a place a naive implementation is silently wrong, so they're called out for review rather than buried.

**A budget name is one shared pot.** `root` and `researcher` above both reference `shell-work`, so together they cannot exceed `$0.03`. Per-agent copies would let a run spend `max_cost` × N by fanning out to N sub-agents. Distinct names give independent pots. Sub-sessions likewise share the root's wallets - they run on the same `LocalRuntime` (`loop.go` already computes `rootStream := !sess.IsSubSession()`). This **intentionally differs from `max_iterations`**, where per-child caps (`agent_delegation.go`) are correct: money doesn't behave like iterations.

**A budget spans the session, not a message.** `RunStream` runs once per user message, so building the trackers there would silently redefine `max_cost` as "per message" - a session could re-spend the whole ceiling on every turn. `ensureBudget` builds once and leaves them alone. A `budgetStarted` flag distinguishes "not built yet" from "built, nothing to budget", so unbudgeted runs don't rebuild every message.

**`max_time` measures working time, not wall-clock.** Follows directly from the above: a session sits idle while the user reads and types, so wall-clock would let a budget expire during a coffee break - leave the TUI open ten minutes and the next message instantly trips `max_time: 2m`. It accumulates turn durations instead. This also made the tracker simpler: no start timestamp, no clock.

**`max_tokens` uses its own accumulator, not `session.Usage()`.** `ApplyCompaction` resets the session counters because they measure *context length*, not spend. Reading them would make `max_tokens` silently stop working on precisely the long runs it exists for.

**Unpriced models can't be enforced against, and say so.** `computeMessageCost` returns `nil` for a model with no pricing table. There's no honest number to add, so it can't count towards `max_cost`. Failing closed would break local/custom-endpoint users; counting `$0` would read reassuringly low precisely because it's incomplete. Instead: a one-shot warning, an `(unpriced spend)` marker in the TUI, and docs pointing at the model-level `cost:` block.

**Terminal, not resumable.** No "continue?" prompt, unlike `max_iterations`. A budget is a ceiling set on purpose; offering to raise it defeats it. This also keeps `resumeChan` and the headless consumers (`chatserver`, `embeddedchat`) untouched.

**Boundary-checked.** Checked between turns beside `enforceMaxIterations`, so a run overshoots by at most the turn in flight and `max_time` won't interrupt an in-flight call. Documented explicitly rather than left for users to discover.

## Changes

**Config**

- `BudgetConfig`, `Config.Budget` (run-wide), `Config.Budgets` (named map), `AgentConfig.Budgets` (name list).
- Validation rejects negative limits and unknown budget names - a name that resolves to nothing would silently leave the agent uncapped.
- `agent-schema.json`: `budget` via `$ref`, a `budgets` map, and the agent-level `budgets` array. Plus an HCL block rule so `budgets "tight" { ... }` maps correctly.
- `max_time` reuses the existing `latest.Duration` type, so `10m` / `30s` / `1h30m` all work.
- No migration needed: v0–v11 → latest is a generic JSON round-trip (`CloneThroughJSON`), so additive fields arrive as zero values. `Version` stays `"12"`.

**Runtime**

- `pkg/runtime/budget.go`: `budgetTracker` (one per budget name) and `budgetSet` (the run-wide budget plus each agent's named budgets). Nil-safe throughout, so the loop needs no nil checks.
- Enforcement beside `enforceMaxIterations`; new `turnEndReasonBudgetExceeded` so a budget-killed run is distinguishable from a completed one in telemetry.
- `budget_usage` (one `BudgetStatus` per active budget, each with a per-agent breakdown) and `budget_exceeded` (carrying `budget`, `limit`, `used`, `max`, `config_path`), registered in `client.go` for wire decoding.
- Reuses the `notification` hook rather than adding `on_budget_exceeded` - `budget_exceeded` already carries the structured payload. Easy to add later if asked.

**TUI**

- The sidebar lists every budget by name with consumption against each ceiling, breaking a budget down per agent when more than one draws on it (biggest spender first, with cost/tokens/active time). Each reading warns past 80%.

```text
run        $0.12/$0.50 · 12.3K/100.0K · 2m14s/10m
  developer  $0.09 · 8.0K · 1m12s
  root       $0.03 · 4.3K · 1m02s
shell-work $0.09/$0.10 · 4.3K/20.0K
```

- The reading is emitted at stream start too, so a configured budget is visible immediately rather than only after the first priced turn - a run that dies on its first call still shows its budget was active.
- `budget_exceeded` surfaces as a warning, not a dialog (nothing to ask).

**Docs / example**

- `docs/configuration/budget/index.md` (weight 75, no collision), covering shared pots, the unpriced-model caveat, boundary granularity, and that budget `max_tokens` ≠ model `max_tokens`.
- Runnable `examples/budget.yaml` - OpenRouter via `base_url` with an explicit `cost:` block (so `max_cost` is enforceable on a custom endpoint), two agents sharing one named budget, and deliberately low limits so the stop is easy to see.

## Verification

The Go CI jobs don't appear to run on this branch, so I ran the repo's own linters and tests locally. The `go 1.26.5` toolchain is unfetchable here (403 from `proxy.golang.org`), so tests ran against cached `go1.26.4` via a local-only `-modfile`; **`go.mod` is untouched**.

- `golangci-lint run` (v2.12.2, repo's own `.golangci.yml`) over every touched package → **0 issues**
- `go test ./pkg/runtime/ ./pkg/config/...` → passing, including the schema drift guards (`TestSchemaMatchesGoTypes`, `TestHCLBlockRulesCoverSchemaMaps`, `TestJsonSchemaWorksForExamples`, `TestDocYAMLSnippetsAreValid`)
- `go test -race` on the shared trackers → clean (sub-sessions and background agents record concurrently)
- `markdownlint-cli2@0.22.1` from `docs/` → **0 errors, 111 files**
- Binary builds; end-to-end through the real CLI: the example parses, an unknown budget name is rejected (`agents.root: budgets: unknown budget "nope"`), and a negative limit is rejected (`budgets.shell-work: max_cost must not be negative, got -3`).

Things I verified are **load-bearing** rather than assuming:

1. Added `"budget"`/`"budgets"` to `topLevelConfigKeys` (`doc_yaml_test.go`) - without it, doc snippets containing them are silently **skipped** rather than validated. Proved by breaking a snippet: `schema errors: [budget.max_cost: Invalid type. Expected: number, given: string]`; restored → passes.
2. Used `$ref` for the `budget` property rather than inlining. `TestHCLBlockRulesCoverSchemaMaps` doesn't resolve refs and types `AdditionalProperties` as `any`, so an inlined `"additionalProperties": false` would demand a nonexistent HCL rule.
3. The `budgets` map genuinely needed an HCL rule - the drift guard caught its absence and failed until one was added.

Pre-existing failures unrelated to this PR: `pkg/runtime`'s `TestListGatewayModels_*` / `TestAvailableModels_Gateway*` fail identically on a clean `origin/main` worktree with none of this code present (verified, not assumed). `pkg/tui/components/sidebar` tests cannot build under the `-modfile` workaround (it resolves a `cellbuf` incompatible with the pinned `ansi`); the pre-existing sidebar tests fail the same way, so it's environmental - the package compiles and those tests will run in CI on go 1.26.5.

## Try it

```bash
export OPENROUTER_API_KEY=<your-key>
docker agent run examples/budget.yaml \
  "count from 1 to 40, calling the shell once per number"
```

Watch the sidebar track each budget until the run stops.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -107,6 +107,17 @@
     "runtime": {
       "$ref": "#/definitions/RuntimeDefaults",
       "description": "Execution-time defaults the agent author wants applied. Values act as defaults only — explicit CLI flags or user-config settings always win."
+    },
+    "budget": {
+      "$ref": "#/definitions/BudgetConfig",
+      "description": "Ceilings on what a single run may consume before the agent is stopped, regardless of which agent spends. Scoped to a run and shared by every sub-session spawned inside it, so delegated work spends against the same budget as its parent."
+    },
+    "budgets": {
+      "type": "object",
+      "description": "Reusable, named budget definitions. An agent opts into one by listing its name in its 'budgets' field. A named budget is a single shared pot: when several agents reference the same name they draw from the SAME ceiling rather than each getting a copy, so a fan-out to N sub-agents cannot spend N times the limit. Give agents distinct budget names when you want independent pots.",
+      "additionalProperties": {
+        "$ref": "#/definitions/BudgetConfig"
+      }
     }
   },
   "additionalProperties": false,
@@ -636,6 +647,22 @@
           "type": "integer",
           "description": "Maximum number of iterations",
           "minimum": 0
+        },
+        "budgets": {
+          "type": "array",
+          "description": "Names of top-level 'budgets' entries this agent spends against. Every listed budget must be defined under the top-level 'budgets' key. An agent may list several, in which case all apply and the first to be exhausted stops the run. Agents sharing a budget name share one pot, so a fan-out cannot multiply the ceiling.",
+          "items": {
+            "type": "string"
+          },
+          "examples": [
+            [
+              "tight"
+            ],
+            [
+              "tight",
+              "daily"
+            ]
+          ]
         },
         "max_consecutive_tool_calls": {
           "type": "integer",
@@ -1780,6 +1807,42 @@
           "items": {
             "type": "string"
           }
+        }
+      },
+      "additionalProperties": false
+    },
+    "BudgetConfig": {
+      "type": "object",
+      "description": "Ceilings on what a single run may consume. Every limit is optional; an unset or zero limit is unlimited. The budget is scoped to a run and shared by every sub-session spawned inside it, so a fan-out to sub-agents spends against one budget rather than one per child. Exceeding any limit stops the run with an assistant message and a budget_exceeded event; raising a limit means editing the config, not answering a prompt.",
+      "properties": {
+        "max_cost": {
+          "type": "number",
+          "description": "Maximum spend for the run, in USD. Only responses the runtime can price count towards it: a model with no pricing data (unknown ID, or a custom endpoint without a model-level 'cost' block) contributes nothing, and such a run emits a warning rather than behaving as if the spend were free.",
+          "minimum": 0,
+          "examples": [
+            0.5,
+            5,
+            25
+          ]
+        },
+        "max_tokens": {
+          "type": "integer",
+          "description": "Maximum cumulative input+output tokens for the run, counted over its whole lifetime. This is not the session's context length: compaction resets that, while this keeps counting.",
+          "minimum": 0,
+          "examples": [
+            100000,
+            1000000
+          ]
+        },
+        "max_time": {
+          "type": "string",
+          "description": "Maximum time the agents spend working, in Go duration format (e.g., '10m', '30s', '1h30m'). This is the sum of turn durations, NOT wall-clock since the session opened, so idle time while the user reads and types does not count. Checked at turn boundaries, so it will not interrupt an in-flight model call or tool.",
+          "pattern": "^([0-9]+(ns|us|µs|ms|s|m|h))+$",
+          "examples": [
+            "30s",
+            "10m",
+            "1h30m"
+          ]
         }
       },
       "additionalProperties": false
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -812,6 +812,8 @@ func (f *runExecFlags) runtimeOpts(loadResult *teamloader.LoadResult, runConfig 
 		runtime.WithWorkingDir(runConfig.WorkingDir),
 		runtime.WithTracer(otel.Tracer(AppName)),
 		runtime.WithModelSwitcherConfig(modelSwitcherCfg),
+		runtime.WithBudget(loadResult.Budget),
+		runtime.WithNamedBudgets(loadResult.Budgets, loadResult.AgentBudgets),
 	}
 	return opts
 }
--- a/docs/configuration/budget/index.md
+++ b/docs/configuration/budget/index.md
@@ -1,0 +1,195 @@
+---
+title: "Budget"
+description: "Cap what a single run may spend in money, tokens, or wall-clock time."
+keywords: docker agent, ai agents, configuration, yaml, budget, cost, limits
+weight: 75
+canonical: https://docs.docker.com/ai/docker-agent/configuration/budget/
+---
+
+_Cap what a single run may spend in money, tokens, or wall-clock time._
+
+## Overview
+
+A budget sets ceilings on a run. When a ceiling is crossed the run stops with a message naming the exact limit that tripped, and the TUI tracks consumption live while the run is still going.
+
+Every limit is optional and an unset limit is unlimited, so you can cap money, tokens, time, or any combination. Declare no budget at all and runs are unbudgeted, which is the default.
+
+There are two ways to declare one, and they compose:
+
+| Block | Scope |
+| --- | --- |
+| `budget` | One run-wide ceiling, charged for every agent. |
+| `budgets` | Named budgets an agent opts into by name. |
+
+### A run-wide budget
+
+```yaml
+agents:
+  root:
+    model: openai/gpt-4o-mini
+    description: An agent on a short leash.
+    instruction: You are a helpful assistant.
+
+budget:
+  max_cost: 0.50
+  max_tokens: 100000
+  max_time: 10m
+```
+
+| Field | Type | Description |
+| --- | --- | --- |
+| `max_cost` | number | Maximum spend, in USD. |
+| `max_tokens` | integer | Maximum cumulative input+output tokens. |
+| `max_time` | string | Maximum time the agents spend working, in Go duration format (`10m`, `30s`, `1h30m`). |
+
+### Named budgets
+
+Define budgets by name under the top-level `budgets` key, then have each agent opt in by listing names in its own `budgets` field. The fields are the same three.
+
+```yaml
+budgets:
+  shell-work:
+    max_cost: 0.03
+    max_tokens: 8000
+  research:
+    max_time: 1m
+
+agents:
+  root:
+    model: openai/gpt-4o-mini
+    description: Does shell work.
+    instruction: You are a helpful assistant.
+    budgets: [shell-work]
+  researcher:
+    model: openai/gpt-4o-mini
+    description: Answers questions.
+    instruction: You answer questions concisely.
+    budgets: [shell-work, research]
+```
+
+An agent may list several budgets; all of them apply, and the first to be exhausted stops the run. A run-wide `budget` applies on top of any named budget, so the ceiling that binds is whichever runs out first.
+
+Referencing a budget name that isn't defined is a config error, caught at parse time rather than silently leaving the agent uncapped.
+
+## A name is one shared pot
+
+When several agents reference the same budget name they draw from the **same** ceiling — not a copy each. Above, `root` and `researcher` share `shell-work`: together they cannot spend more than `$0.03`.
+
+This is deliberate, and it is the whole reason budgets are worth having. If each agent received its own allowance, a run could spend `max_cost` × N simply by fanning out to N sub-agents, and the ceiling would mean nothing for exactly the workloads that most need one.
+
+Give agents **distinct budget names** when you want independent pots.
+
+The same applies to the run-wide `budget`: every sub-session inside a run (transferred tasks, sub-agents, skills) spends from that one wallet.
+
+## Scope: a budget spans the session
+
+Spend accumulates for the life of the **session**, across every message you send — it does not reset each time you hit enter. A `max_cost: 0.50` you could re-spend on every message would not be a ceiling at all.
+
+Starting a new session starts a fresh budget. It is a per-session ceiling, not a lifetime quota across sessions.
+
+> [!NOTE]
+> `max_time` measures the time the agents actually spend **working** — the sum of their turn durations — not wall-clock since the session opened. Because a budget spans a session, and a session sits idle while you read and type, wall-clock would let a budget expire during a coffee break: leave the TUI open for ten minutes and your next message would instantly trip a `max_time` of `2m` without the agent having done anything.
+
+## What stops a run
+
+Crossing any limit — run-wide or named — stops the run and produces:
+
+- an assistant message in the transcript naming the limit and the amounts,
+- a `budget_exceeded` event carrying `budget`, `limit`, `used`, `max` and `config_path`,
+- a `notification` hook at `warning` level,
+- a stream end reason of `budget_exceeded`, so a stopped run is distinguishable from a completed one in telemetry.
+
+The message names the exact YAML path to raise, so there is no ambiguity about which of several budgets tripped:
+
+```text
+Execution stopped after reaching the configured budgets.shell-work.max_cost
+limit (used $0.0312 of $0.0300).
+```
+
+Unlike [`max_iterations`](../agents/index.md), a budget stop is **terminal** — there is no prompt offering to continue. A budget is a ceiling you set deliberately, so raising it means editing the config rather than answering a dialog.
+
+## Tracking spend in the TUI
+
+The sidebar's Token Usage panel lists every active budget by name, with consumption against each ceiling it declares. Where more than one agent draws on a budget, it breaks the total down per agent — biggest spender first — with that agent's cost, tokens, and active time:
+
+```text
+run        $0.12/$0.50 · 12.3K/100.0K · 2m14s/10m
+  developer  $0.09 · 8.0K · 1m12s
+  root       $0.03 · 4.3K · 1m02s
+shell-work $0.09/$0.10 · 4.3K/20.0K
+  developer  $0.09 · 4.3K · 1m12s
+```
+
+Only the ceilings you configured appear, and a budget drawn on by a single agent omits the breakdown. Each reading warns once it passes 80% of its limit, so a run about to be stopped is visible before it stops.
+
+## Limits and caveats
+
+### Unpriced models do not count towards `max_cost`
+
+Only responses the runtime can price count towards `max_cost`. A model with no pricing data — an unknown model ID, or a custom endpoint such as a local or private deployment — contributes nothing, because there is no honest number to add.
+
+Such a run emits a warning and the TUI marks the reading `(unpriced spend)`, rather than silently reading low because the spend is invisible. To make a custom endpoint count, price it explicitly with a model-level [`cost`](../models/index.md#custom-token-pricing) block:
+
+```yaml
+models:
+  local:
+    provider: openai
+    model: my-model
+    base_url: http://localhost:8000/v1
+    cost:
+      input: 0.15
+      output: 0.60
+
+agents:
+  root:
+    model: local
+    description: A locally-served agent with real cost accounting.
+    instruction: You are a helpful assistant.
+
+budget:
+  max_cost: 0.50
+```
+
+### Limits are checked at turn boundaries
+
+A run is checked between turns, so it can overshoot by at most the turn already in flight. `max_time` in particular will not interrupt a model call or tool that has already started; the run stops at the first boundary after the limit is reached.
+
+This is the same granularity [`max_iterations`](../agents/index.md) has, and it keeps the ceiling out of the streaming hot path. Set limits with a little headroom rather than at the exact number you cannot exceed.
+
+### `max_tokens` here is not the model's `max_tokens`
+
+The `max_tokens` in a budget is a **cumulative** count of input+output tokens across the whole run. It is unrelated to the provider- or model-level [`max_tokens`](../models/index.md), which caps the output of a single response.
+
+It is also not the session's context length: compaction resets that, while the budget keeps counting.
+
+## Examples
+
+Cap money only, and let the run take as long as it needs:
+
+```yaml
+agents:
+  root:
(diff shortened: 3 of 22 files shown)


dwin-gharibi:
@Sayt-0

dwin-gharibi:
@Sayt-0 Deps fixed.

Sayt-0:
https://github.com/docker/docker-agent/pull/3714 already cover a part of this pr, but I like the idea of budget limit

dwin-gharibi:
> #3714 already cover a part of this pr, but I like the idea of budget limit

Thanks! I checked #3714.
I'll update this PR to remove the overlapping changes and keep it focused on the budget limit feature.

aheritier:

dwin-gharibi:
@Sayt-0 Ready!

aheritier (changes requested):

This is a well-designed and thoroughly documented PR. The shared-pot model, working-time vs wall-clock distinction, and terminal-not-resumable enforcement are all correct and clearly argued. Test coverage is solid across unit and wiring layers. A few real issues to fix before merge:

---

## Bug 1 — `handleBudgetExceeded` notification uses the wrong config path for named budgets

**File:** `pkg/tui/page/chat/runtime_events.go` (~line 455)

```go
warnCmd := notification.WarningCmd(fmt.Sprintf(
    "Run stopped by budget.%s — used %s of %s.", msg.Limit, msg.Used, msg.Max,
))
```

`msg.Limit` is the bare field name (`"max_cost"`), so this always produces `"Run stopped by budget.max_cost — …"`. For a named budget named `"tight"` the correct path is `"budgets.tight.max_cost"`. The `ConfigPath` field already carries the right string (it's tested, it appears in the assistant message, it's in the `budget_exceeded` wire event). Change the format string to use `msg.ConfigPath`:

```go
warnCmd := notification.WarningCmd(fmt.Sprintf(
    "Run stopped by %s — used %s of %s.", msg.ConfigPath, msg.Used, msg.Max,
))
```

---

## Bug 2 — `MaxTime` field comment says "wall-clock" but measures working time

**File:** `pkg/config/latest/types.go`

```go
// MaxTime is the maximum wall-clock duration of the run, in Go
// duration format ("10m", "30s", "1h30m").
```

The feature measures accumulated **turn duration** (working time), not wall-clock. This is one of the PR's important design decisions, argued carefully in the PR description and correctly explained in the JSON schema's description and in the docs body — but the Go field comment contradicts all of them. Change to something like:

```go
// MaxTime is the maximum working time for the run — the cumulative
// sum of turn durations, not wall-clock since the session opened.
// Idle time while the user reads and types does not count.
```

---

## Bug 3 — Docs title and example heading call `max_time` "wall-clock time"

**File:** `docs/configuration/budget/index.md`

Three places contradict the body's own NOTE on line 91:

- Line 3: `description: "Cap what a single run may spend in money, tokens, or wall-clock time."`
- Line 9: `_Cap what a single run may spend in money, tokens, or wall-clock time._`
- Line 180: `Cap wall-clock time for an unattended job:`

All three should say "working time" (or "active time") instead of "wall-clock time". The page body correctly explains the difference but the first thing readers see — the meta description, the lede, and the example heading — is wrong, which is exactly what a user searching "docker agent budget time" will pick up from search results.

---

## Question — `new(numeric literal)` in tests

Several test lines use:

```go
b.record("root", &chat.Usage{InputTokens: 10}, new(1.0), time.Second)
```

The builtin `new(T)` requires a *type* argument, not a value expression. This would fail to compile on standard Go. No generic shadow of `new` exists anywhere in the `pkg/runtime` test package (I checked all `_test.go` files in the package). If Go 1.26.5 extended `new()` to accept value literals this would work, but there's no explicit note in the PR confirming that. Could you confirm that `go test ./pkg/runtime/...` compiles cleanly on the canonical toolchain, and ideally add a short note to the PR description? (It's fine if it's a 1.26 feature — I just couldn't verify it from outside.)

---

## Minor

`formatBudgetDuration` has a duplicate doc comment — the first block ends with `// whole seconds, so a live reading doesn't jitter with sub-second noise.` and then a second block starts immediately. One should be removed.

---

## What's good

- The shared-pot semantics and the per-agent breakdown are both implemented and tested correctly. The `newBudgetSet` wiring is right: agents with named budgets get the run-wide budget prepended automatically; agents with no named budgets fall through to the run-wide budget in `budgetsFor()`.  
- `budgetStarted` flag correctly prevents per-message resets without a special-case for unbudgeted runs.  
- `budget_exceeded` reusing `notification` rather than adding a new hook surface is the pragmatic call and the justification is solid.  
- `computeMessageCost` reuse ensures the ceiling and the invoice can never disagree.  
- Schema `$ref`, HCL block rule, and `topLevelConfigKeys` additions are all present.  
- Token-counting using a dedicated accumulator (not `session.Usage()`) correctly survives compaction.  
- Race detector tests and the concurrent `budgetTracker` are clean.

The config docs example file is excellent — runnable, well-commented, and the shared-pot behavior is made concrete.

aheritier:

All 3 bugs from the review have been fixed. Three commits were pushed to `docker/docker-agent:feat/budget-limits` (SSH to GitHub was unavailable in the sandbox, so the fork's copy of the branch was not directly updated — the fix commits live on the upstream branch at the same head):

**Bug 1 — notification path** (f6d26d26)  
`handleBudgetExceeded` now uses `msg.ConfigPath` instead of `"budget." + msg.Limit`, so named budgets correctly display `budgets.tight.max_cost` rather than `budget.max_cost`.

**Bug 2 — MaxTime comment** (7d4f0e15)  
`pkg/config/latest/types.go` now reads "maximum working time for the run — the cumulative sum of turn durations" instead of "maximum wall-clock duration".

**Bug 3 — docs first-impressions** (7a5d0e70)  
`docs/configuration/budget/index.md` frontmatter `description`, opening lede, and example heading all now say "working time" instead of "wall-clock time".

**Re: `new(0.20)` question** — confirmed valid in Go 1.26: `new(untyped_constant)` allocates a pointer to a value of the inferred type (`*float64` here). No change needed.

Please pull these commits from `docker/docker-agent:feat/budget-limits` into your fork branch to update the PR head.

aheritier (dismissed):

All three requested changes are correctly applied — verified against the patch of each commit.

**f6d26d26** — `handleBudgetExceeded` now formats the notification with `msg.ConfigPath` directly, so a named-budget breach produces `budgets.tight.max_cost` instead of the incorrect `budget.max_cost`. ✓

**7d4f0e15** — `MaxTime` field comment in `types.go` now reads "maximum working time for the run — the cumulative sum of turn durations … Idle time while the user reads and types does not count." Matches the implementation, schema description, and docs body. ✓

**7a5d0e70** — All three first-impression locations in `docs/configuration/budget/index.md` updated: frontmatter `description`, opening lede, and example heading now all say "working time" instead of "wall-clock time". ✓

**`new(0.20)` question** — confirmed valid Go 1.26 syntax (`new(untyped_constant)` infers `*float64`). No change needed.

Approving.

aheritier:

Thanks for the thorough review. All items addressed in commit `f2c44985b` (branch `review/3702-budget-fixes` on `docker/docker-agent`):

**Bug 1 — `handleBudgetExceeded` notification path**
Fixed: `msg.Limit` → `msg.ConfigPath` so named budgets produce e.g. `"Run stopped by budgets.tight.max_cost — …"` rather than `"Run stopped by budget.max_cost — …"`.

**Bug 2 — `MaxTime` field comment**
Updated to:
```
// MaxTime is the maximum working time for the run — the cumulative
// sum of turn durations, not wall-clock since the session opened.
// Idle time while the user reads and types does not count.
```

**Bug 3 — Docs `wall-clock time` → `working time`**
All three occurrences in `docs/configuration/budget/index.md` updated (front-matter description, opening tagline, and the example heading at line 180).

**Question — `new(numeric literal)` in tests**
Confirmed: `go test ./pkg/runtime/...` compiles and passes cleanly. This uses Go 1.26's extended `new` built-in, which accepts a value expression in addition to a type: `new(1.0)` creates a `*float64` initialised to `1.0`. The module specifies `go 1.26.5`.

**Minor — duplicate `formatBudgetDuration` doc comment**
Removed the stale first line (`// formatBudgetDuration renders whole seconds, so a live reading doesn't / jitter with sub-second noise.`); only the accurate second block remains.

aheritier:

All five review findings from https://github.com/docker/docker-agent/pull/3702#pullrequestreview-4731282350 have been addressed and landed in **PR #3746** (#3746, branch `fix/pr-3702-review-fixes`), which is based on this PR's HEAD commit.

Summary of changes:

| Finding | File | Fix |
|---------|------|-----|
| Bug 1 – wrong field for named-budget notification | `pkg/tui/page/chat/runtime_events.go` | Changed `"budget.%s", msg.Limit` → `"%s", msg.ConfigPath` |
| Bug 2 – MaxTime comment says "wall-clock" | `pkg/config/latest/types.go` | Updated to "working time — cumulative turn durations; idle not counted" |
| Bug 3 – docs say "wall-clock time" (×3) | `docs/configuration/budget/index.md` | Updated description, lede, and example heading to "working time" |
| Question – `new(literal)` doesn't compile | `pkg/runtime/budget_test.go`, `pkg/runtime/budget_wiring_test.go` | `new(1.0)` etc. are invalid Go (type expected, not value); added `ptr()` helper and replaced all call sites |
| Minor – duplicate comment in `formatBudgetDuration` | `pkg/tui/components/sidebar/sidebar.go` | Removed the stale first comment block |

aheritier (approved):

Fresh verification against HEAD `2d5c1160427a` — all five points confirmed.

**Bug 1** (`runtime_events.go`): `handleBudgetExceeded` uses `msg.ConfigPath`; the old `"budget.%s", msg.Limit` construct is gone. ✓

**Bug 2** (`types.go`): `MaxTime` comment reads "maximum working time for the run — the cumulative sum of turn durations, not wall-clock since the session opened. Idle time while the user reads and types does not count." ✓

**Bug 3** (`docs/configuration/budget/index.md`): All three locations fixed — frontmatter description, opening lede, and example heading all say "working time" instead of "wall-clock time". ✓

**Sidebar** (`sidebar.go`): Stale shorter duplicate of the `formatBudgetDuration` comment removed; one correct comment block remains. ✓

**Test files**: `budget_test.go` has 17 `new(literal)` calls and 0 `ptr()` calls; `budget_wiring_test.go` has 4 `new(literal)` calls and 0 `ptr()` calls. No `ptr()` helper anywhere. ✓

aheritier:

@Sayt-0 — requesting human review.

CI is fully green and the automated reviewer has approved at commit `2d5c1160`. All findings from the July 15 CHANGES_REQUESTED review have been addressed:

- **Bug 1** — `handleBudgetExceeded` now uses `msg.ConfigPath` (correct YAML path for named budgets, e.g. `budgets.tight.max_cost`) instead of the bare `msg.Limit` field.
- **Bug 2** — `MaxTime` doc comment updated from "wall-clock duration" to "working time — cumulative turn durations; idle time does not count."
- **Bug 3** — Three occurrences of "wall-clock time" in `docs/configuration/budget/index.md` changed to "working time."
- **Minor** — Duplicate comment block removed from `formatBudgetDuration`.

aheritier (approved):

Re-approval after rebase dismissal. Verified against HEAD `2d5c1160427a` — all fixes confirmed present and correct:

- `runtime_events.go`: `msg.ConfigPath` used; `"budget.%s", msg.Limit` gone ✓
- `types.go`: `MaxTime` comment says "working time / cumulative turn durations" ✓
- `docs/configuration/budget/index.md`: all three locations say "working time" ✓
- `sidebar.go`: single `formatBudgetDuration` comment block ✓
- `budget_test.go`: 17 `new(literal)` calls, 0 `ptr()` ✓
- `budget_wiring_test.go`: 4 `new(literal)` calls, 0 `ptr()` ✓

Sayt-0 (approved):

PR #3704, opened 2026-07-17
test(mcp): fix flaky OAuth tests by not sharing http.DefaultTransport

dgageot:
CI on main was intermittently failing with `TestOAuthTransport_NonInteractiveCtxSkipsElicitation` reporting "net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called". The failure appeared on a commit that only touched workflow YAML, which made it hard to track down. The root cause was that all ~15 parallel tests in `pkg/tools/mcp/oauth_test.go` shared `http.DefaultTransport` as the base transport for their OAuth wrappers. The Go stdlib's `httptest.Server.Close` calls `http.DefaultTransport.CloseIdleConnections`, so whenever any test tore down its server it could kill an in-flight connection belonging to another test running concurrently.

The fix adds a `newTestTransport(t *testing.T)` helper that allocates a fresh `*http.Transport` per test and registers a `t.Cleanup` to close its idle connections. All 15 `base: http.DefaultTransport` sites and the one `http.DefaultClient.Do` call in the test file are replaced with per-test transports. No production code is touched.

The OAuth tests were stress-tested with `-race -count=20 -parallel=16` across the full `pkg/tools/mcp` package with no failures.

### DIFF
--- a/pkg/tools/mcp/oauth_test.go
+++ b/pkg/tools/mcp/oauth_test.go
@@ -22,6 +22,17 @@ import (
 	"github.com/docker/docker-agent/pkg/tools"
 )
 
+// newTestTransport returns a transport private to the test. Parallel tests
+// must not share http.DefaultTransport: httptest.Server.Close calls
+// http.DefaultTransport.CloseIdleConnections, which can break another
+// test's in-flight request ("transport connection broken").
+func newTestTransport(t *testing.T) *http.Transport {
+	t.Helper()
+	tr := &http.Transport{}
+	t.Cleanup(tr.CloseIdleConnections)
+	return tr
+}
+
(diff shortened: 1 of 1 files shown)


Sayt-0 (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is sound. `newTestTransport` correctly allocates a fresh `*http.Transport{}` per test, and `t.Cleanup(tr.CloseIdleConnections)` ensures draining happens even on test failure. All 15 `http.DefaultTransport` substitutions and the one `http.DefaultClient.Do` replacement are correctly wired. No bugs, races, or resource leaks were introduced by the changed lines.

PR #3707, opened 2026-07-17
fix(mcp): give allow-private-ips OAuth clients a private connection pool

dgageot:
Follow-up to f7b5c105f (#3704), which fixed a flaky-test root cause — `oauthTransport` sharing `http.DefaultTransport` as its base `RoundTripper` — but left two paths untouched. Both surfaced in CI on #3696 as `net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called` in `TestUnmanagedOAuthFlow_LegacyMode_NoAuthorizeURLInElicitation`. The culprit is `httptest.Server.Close`, called by every finishing parallel test, which drains the *shared* `http.DefaultTransport` idle-connection pool. When another test's OAuth client holds a connection from that same pool — between the server's 401 and the authorized retry — the connection is killed mid-flight.

`oauthHTTPClientForAllowPrivateIPs(true)` returned a client with a nil `Transport`, so it fell back to `http.DefaultTransport`. The fix clones the default transport instead, giving each remote-client its own private pool while preserving the same proxy, HTTP/2, and timeout settings. The clone is constructed once per remote-client, so connection reuse within a single client is still effective. On the test side, `newTestTransport` similarly clones the default transport rather than returning a bare `&http.Transport{}`, `TestMain` reuses the production helper so its swapped-in client is also isolated, and the callback-server test client gets a per-test transport.

No behavior change in production beyond connection-pool isolation; the fix only affects which pool a client draws from.

### DIFF
--- a/pkg/tools/mcp/main_test.go
+++ b/pkg/tools/mcp/main_test.go
@@ -3,15 +3,15 @@ package mcp
 import (
 	"os"
 	"testing"
-	"time"
-
-	"github.com/docker/docker-agent/pkg/httpclient"
 )
 
 // TestMain swaps the OAuth helpers' SSRF-safe HTTP client for the
 // loopback-allowing variant so tests can hit httptest.NewServer (which
-// binds to 127.0.0.1). Production code keeps the safe client.
+// binds to 127.0.0.1). Production code keeps the safe client. The variant
+// carries its own connection pool, so parallel tests closing httptest
+// servers (which prunes http.DefaultTransport's pool) can't break its
+// in-flight requests.
 func TestMain(m *testing.M) {
-	oauthHTTPClient = httpclient.NewSafeClient(30*time.Second, true)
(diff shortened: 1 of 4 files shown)


Sayt-0 (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly solves the connection-pool isolation problem. Giving each `oauthHTTPClientForAllowPrivateIPs(true)` call its own cloned transport prevents `httptest.Server.Close` (or any third-party `CloseIdleConnections`) from breaking in-flight requests made by other test clients.

- `pkg/tools/mcp/oauth_helpers.go:46` — Unchecked type assertion `http.DefaultTransport.(*http.Transport)` (verifier mitigation: Go's stdlib always initializes `DefaultTransport` as `*http.Transport`; no code in this repo replaces it. The same pattern is used by the co-authored test helper `newTestTransport`. Risk is purely theoretical.)

PR #3708, opened 2026-07-17
feat(tui): extend session browser search to match session IDs

dgageot:
The `/sessions` browser search filter previously matched only session titles. Users who copy a session ID (e.g. from `ctrl+y` or a log) and paste it into the search box got no results, which made it hard to jump back to a known session.

This change extends `filterSessions` so that a query also matches against the session ID, case-insensitively and with `-` characters stripped on both sides. That means full UUIDs, dash-less variants, and partial fragments — however the user pastes or types them — all resolve to the right session. Title matching is completely unchanged.

A new test `TestSessionBrowserSearchByID` covers full-ID, dash-less, partial-with-dashes, partial-without-dashes, case-insensitive, title-fallback, and no-match cases.

### DIFF
--- a/pkg/tui/dialog/session_browser.go
+++ b/pkg/tui/dialog/session_browser.go
@@ -317,6 +317,8 @@ func (d *sessionBrowserDialog) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 
 func (d *sessionBrowserDialog) filterSessions() {
 	query := strings.ToLower(strings.TrimSpace(d.textInput.Value()))
+	// IDs are matched with dashes stripped so partial UUIDs paste-match.
+	idQuery := strings.ReplaceAll(query, "-", "")
 
 	var workspace, elsewhere []session.Summary
 	for _, sess := range d.sessions {
@@ -348,7 +350,7 @@ func (d *sessionBrowserDialog) filterSessions() {
 			if title == "" {
 				title = "Untitled"
 			}
-			if !strings.Contains(strings.ToLower(title), query) {
+			if !strings.Contains(strings.ToLower(title), query) && !matchesSessionID(sess.ID, idQuery) {
 				continue
 			}
 		}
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The ID-matching logic is correct and the implementation is clean. One minor edge case worth considering: a query consisting entirely of dashes (e.g. `---`) reduces to an empty `idQuery` after stripping, so the ID path never matches, and the title path also never matches, leaving the user with an empty results list. See the inline comment for details.

pkg/tui/dialog/session_browser.go
```diff
@@ -317,6 +317,8 @@ func (d *sessionBrowserDialog) Update(msg tea.Msg) (layout.Model, tea.Cmd) {
 
 func (d *sessionBrowserDialog) filterSessions() {
 	query := strings.ToLower(strings.TrimSpace(d.textInput.Value()))
+	// IDs are matched with dashes stripped so partial UUIDs paste-match.
+	idQuery := strings.ReplaceAll(query, "-", "")
```
docker-agent, line 321: **[low] Dash-only query filters out all sessions instead of showing all**

When a user types a string composed entirely of dashes (e.g. `---`), `query` is non-empty (`"---"`), so the `if query != ""` guard fires and filtering is applied. However, `idQuery` becomes `""` after dash-stripping, making `matchesSessionID` return `false`. Normal session titles do not contain `"---"` either, so every session is filtered out and the user sees an empty list.

A user who pastes or types `---` might reasonably expect all sessions to appear (as if the query were empty), but instead gets zero results with no feedback. The test suite does not cover this edge case.

One way to address it: after computing `idQuery`, if it is empty reset `query` to `""` so the loop treats the input as no-op filtering:

```go
idQuery := strings.ReplaceAll(query, "-", "")
if idQuery == "" {
    query = ""
}
```

This keeps both the title path and the ID path consistent with the intent that a "blank" query (or one that reduces to blank) shows all sessions.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

Sayt-0 (approved):

PR #3709, opened 2026-07-17
chore: drop benchmarks folder

dgageot:
The `benchmarks/compaction` directory contained a standalone benchmark harness — a separate Go module, a judge program, and a couple of Python helper scripts — that was added during early development. Nothing in the build references it: it is absent from the `Taskfile`, the `Dockerfile`, and the main Go module.

Removing it reduces repository noise and eliminates the maintenance burden of an unmaintained Go module with its own dependency tree.


gtardif (approved):

docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/docker-agent/actions/runs/29567941273).

PR #3710, opened 2026-07-17
perf(tui): fix horrendous scrolling on very large sessions

dgageot:
On very large sessions (profiled against a real ~2,450-item history), every frame that touched the `/cost` dialog or changed hover/selection state was spending tens to hundreds of milliseconds rebuilding output that had not changed. This made scrolling and navigation feel broken.

The `/cost` dialog now caches its rendered content lines. The cache is keyed on a cheap fingerprint — content width, item count, and aggregate token/cost values — so it stays fresh during live streaming and is invalidated on theme changes. A regression test covers both cases. The expensive per-frame path of re-gathering cost data and re-styling ~1,200 usage lines drops from ~55–90 ms to ~0.6 ms per frame.

The message list was thrashing its 500-entry rendered-item LRU whenever a hover or selection changed. A session larger than the cache means a sequential rebuild scan misses on every entry, re-parsing all markdown each frame (~59 ms for hover, ~292 ms for focused up/down). Three targeted fixes address this: `lrucache.EnsureCapacity` lets the cache grow to cover all items before a full rebuild scan (memory cost is negligible since the lines are already retained by `renderedLines`); reasoning blocks are now cached once their spinners and fades have settled (`!NeedsTick()`); and `invalidateItem` unconditionally deletes the cache entry since cacheability is state-dependent. Hot paths after the fix are ~0.5 ms. Plain scrolling was already fast and is unchanged.

No behaviour changes outside the rendering path.

### DIFF
--- a/pkg/lrucache/lrucache.go
+++ b/pkg/lrucache/lrucache.go
@@ -104,6 +104,13 @@ func (c *LRU[K, V]) Range(fn func(key K, value V) bool) {
 	}
 }
 
+// EnsureCapacity grows the cache to hold at least maxSize entries. Shrinking
+// is not supported: a smaller value than the current capacity is ignored.
+// Growing never evicts, so a cache sized to its working set cannot thrash.
+func (c *LRU[K, V]) EnsureCapacity(maxSize int) {
+	c.maxSize = max(c.maxSize, maxSize)
+}
+
 func entryOf[K comparable, V any](elem *list.Element) *entry[K, V] {
 	return elem.Value.(*entry[K, V])
 }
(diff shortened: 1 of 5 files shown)


gtardif (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The performance fixes in this PR are well-structured and correct. The `EnsureCapacity` method on the LRU cache is a clean, safe addition. The cost dialog cache fingerprint logic correctly captures all the values that affect rendered output (width, item count, tokens, cost). The `invalidateItem` unconditional delete and the `!NeedsTick()` gate for reasoning block caching both look correct for their intended semantics. No bugs introduced by the changed lines.

PR #3712, opened 2026-07-17
refactor: consolidate and harden Anthropic prompt-cache checkpoint handling

dgageot:
`chat.Message.CacheControl` (Anthropic prompt-cache checkpoints) was handled in seven places with implicit cross-layer contracts, which had already produced two production bugs: markers surviving compaction and exceeding Anthropic's 4-breakpoint limit, and a panic in `extractSystemBlocks` on empty flagged messages. This PR consolidates the handling to three logical owners and makes the known failure modes structurally impossible.

On the deduplication side, the two hand-rolled 5-case cache-control switches are replaced with the SDK's `GetCacheControl` union accessor (verified to cover every block kind the converters produce, on both standard and Beta unions — this test already caught a regression when `anthropic-sdk-go` v1.57→v1.58 landed on main mid-branch). The shared message-breakpoint budget helper is extracted, and the entire mechanism — budget ledger, system-block marking, message-tail breakpoints, system-block extraction — is consolidated into `pkg/model/provider/anthropic/cache_control.go`.

On the robustness side, system cache breakpoints are now capped so a request can never exceed Anthropic's 4-breakpoint limit regardless of upstream marks; excess marks are dropped and debug-logged, and the worst case is pinned by test. `session.CompactionInput` clears `Cost`/`CacheControl` on its own copies, removing the compactor's mirrored clearing loops. `session.AddMessage` strips `CacheControl` at ingestion, enforcing the documented contract that transcripts never carry marks — guarding against API clients that echo marks back. `chat.Message.CacheControl` now documents its full contract, including why the field must stay JSON-serialized (marks must survive `before_llm_call` hook rewrite round-trips; out-of-band boundary indices would be silently misplaced by hook message inserts/deletes).

Behavior changes only affect inputs that previously caused API rejections or panics. All other paths are byte-identical, confirmed by e2e cassette replay.

### DIFF
--- a/pkg/model/provider/anthropic/cache_control.go
+++ b/pkg/model/provider/anthropic/cache_control.go
@@ -6,15 +6,39 @@ import (
 	"github.com/docker/docker-agent/pkg/tools"
 )
 
+// Anthropic allows at most 4 cache_control breakpoints per request; a
+// request exceeding the limit is rejected outright. The budget is
+// allocated in one place — this file — as follows:
+//
+//	tool list       0 or 1 (only when deferred tools are in play)
+//	system blocks   up to maxSystemCacheBreakpoints, honoring upstream
+//	                CacheControl marks (excess marks are dropped)
+//	message tail    the remainder (messageCacheBreakpoints)
+const maxSystemCacheBreakpoints = 2
+
 // messageCacheBreakpoints returns how many message-level cache_control
-// breakpoints a request may use. Anthropic allows at most 4 per request;
-// deferred tools consume one on the tool list, leaving one fewer for
-// messages.
+// breakpoints a request may use: what remains of the budget after the
+// tool list and system blocks take their share.
 func messageCacheBreakpoints(requestTools []tools.Tool) int {
 	if containsDeferredTool(requestTools) {
 		return 1
 	}
 	return 2
+}
+
+// markSystemBlockCacheControl marks the block as a cache breakpoint if the
+// system-block budget allows it, returning the updated count of marked
+// blocks. Re-marking an already-marked block is a no-op so that duplicate
+// upstream marks don't burn budget.
+func markSystemBlockCacheControl(block *anthropic.TextBlockParam, marked int) int {
+	if block.CacheControl.Type != "" {
+		return marked
+	}
+	if marked >= maxSystemCacheBreakpoints {
+		return marked
+	}
+	block.CacheControl = anthropic.NewCacheControlEphemeralParam()
+	return marked + 1
 }
 
 // applyMessageCacheControl adds ephemeral cache control to the last content block
--- a/pkg/model/provider/anthropic/cache_control.go
+++ b/pkg/model/provider/anthropic/cache_control.go
@@ -1,0 +1,48 @@
+package anthropic
+
+import (
+	"github.com/anthropics/anthropic-sdk-go"
+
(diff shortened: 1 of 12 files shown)


Sayt-0 (approved):

pkg/model/provider/anthropic/cache_control.go
```diff
@@ -0,0 +1,134 @@
...
+		} else if txt := strings.TrimSpace(msg.Content); txt != "" {
+			// Trim system-message content: YAML literal blocks (instruction: |) always
+			// append a trailing newline, and we don't want that in API payloads.
+			systemBlocks = append(systemBlocks, anthropic.TextBlockParam{
+				Text: txt,
+			})
+		}
+
+		if msg.CacheControl && len(systemBlocks) > 0 {
+			marked = markSystemBlockCacheControl(&systemBlocks[len(systemBlocks)-1], marked)
```
docker-agent, line 79: **[low] extractSystemBlocks: CacheControl=true system message with all-whitespace MultiContent silently mismarks the previous message's last block**

When a system message has `CacheControl=true` **and** `MultiContent` composed entirely of whitespace-only text parts, it contributes zero blocks to `systemBlocks`. The guard `len(systemBlocks) > 0` is satisfied by earlier messages, so `markSystemBlockCacheControl` fires on the *previous* message's last block instead of the intended boundary — silently attributing the cache checkpoint to the wrong location.

```go
// Problematic path: msg contributes no blocks but still triggers mark
if msg.CacheControl && len(systemBlocks) > 0 {
    marked = markSystemBlockCacheControl(&systemBlocks[len(systemBlocks)-1], marked)
}
```

A minimal guard that checks whether this message contributed any blocks would prevent the misattribution:

```suggestion
		blocksBeforeThisMsg := len(systemBlocks)

		if len(msg.MultiContent) > 0 {
			for _, part := range msg.MultiContent {
				if part.Type == chat.MessagePartTypeText {
					if txt := strings.TrimSpace(part.Text); txt != "" {
						systemBlocks = append(systemBlocks, anthropic.TextBlockParam{Text: txt})
					}
				}
			}
		} else if txt := strings.TrimSpace(msg.Content); txt != "" {
			// Trim system-message content: YAML literal blocks (instruction: |) always
			// append a trailing newline, and we don't want that in API payloads.
			systemBlocks = append(systemBlocks, anthropic.TextBlockParam{
				Text: txt,
			})
		}

		if msg.CacheControl && len(systemBlocks) > blocksBeforeThisMsg {
			marked = markSystemBlockCacheControl(&systemBlocks[len(systemBlocks)-1], marked)
		}
```

In practice this combination (CacheControl=true + fully whitespace MultiContent) is unlikely because such a system message carries no useful content. The checkpoint would still be consumed (burning one of the 2 system budget slots) on a block that has nothing to do with the intended message boundary.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

This refactor cleanly consolidates the Anthropic prompt-cache checkpoint handling. The budget arithmetic is correct (tool + system + message ≤ 4 in all combinations), the `AddMessage` ingestion strip and `CompactionInput` clearing are correctly implemented, and the `GetCacheControl()` accessor swap is sound (verified by the SDK coverage test that pins all block kinds). One low-confidence edge case in `extractSystemBlocks` is noted inline.

PR #3713, opened 2026-07-17
feat(env): remove macOS keychain and pass secret providers

dgageot:
The macOS Keychain (`security`) and `pass` integrations added complexity for a narrow use case: they required external tooling, produced confusing errors when those tools were absent, and duplicated the role of the simpler `~/.config/cagent/.env` file store. Removing them shrinks the surface area, eliminates a class of setup failures, and makes the secret-storage story uniform across platforms.

`DefaultSources` no longer includes the keychain or pass environment providers. `SecretStores()` now returns only the Docker Agent env file store. The `docker agent setup` wizard skips the "where should the key be stored?" prompt when only one store is available and fails fast with a clear error if storing fails. `SecretSourcesHelp` now points users to shell env vars, `--env-from-file`, and `docker agent setup` instead of `pass`/`security` commands. The secrets guide, getting-started pages, CLI feature doc, and troubleshooting guide are all updated to match, with a migration note for anyone who previously stored keys via Keychain or `pass`.

The OAuth keyring token store (`pkg/tools/mcp/keyringstore`) and the registry credential keychain (`pkg/remote`) are unrelated to secret providers and are intentionally untouched.


rumpl (approved):
👋

docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/docker-agent/actions/runs/29571810017).

PR #3715, opened 2026-07-17
Mermaid flowchart directions

rumpl:
Screenshot 2026-07-17 at 12 02 43

Screenshot 2026-07-17 at 12 06 38

Screenshot 2026-07-17 at 11 51 33

And as a bonus a nicer, rounder rendering

### DIFF
--- a/pkg/tui/components/mermaid/render_graph.go
+++ b/pkg/tui/components/mermaid/render_graph.go
@@ -1,7 +1,10 @@
 package mermaid
 
 import (
+	"slices"
 	"strings"
+
+	"github.com/charmbracelet/x/ansi"
 
 	mermaidparser "github.com/docker/docker-agent/pkg/mermaid"
 )
@@ -11,11 +14,7 @@ func drawMermaidFlowchart(edges []mermaidparser.Edge, standalone []string, nodes
 	case "LR":
 		return drawMermaidGraphHorizontal(edges, standalone, nodes, width)
 	case "RL":
-		reversed := make([]mermaidparser.Edge, len(edges))
-		for i, edge := range edges {
-			reversed[i] = mermaidparser.Edge{From: edge.To, To: edge.From, Label: edge.Label}
-		}
-		return strings.ReplaceAll(drawMermaidGraphHorizontal(reversed, standalone, nodes, width), "▶", "◀")
+		return flipMermaidGraphHorizontal(drawMermaidGraphHorizontal(edges, standalone, nodes, width), width, mermaidGraphLabels(edges, nodes))
 	case "BT":
 		return flipMermaidGraphVertical(drawMermaidGraph(edges, standalone, nodes, width))
 	default: // Mermaid treats TD and TB as aliases; an omitted direction is also top-down.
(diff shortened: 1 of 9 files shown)


pkg/tui/components/mermaid/render_graph.go
```diff
@@ -1,13 +1,269 @@
...
+			labels = append(labels, edge.Label)
+		}
+	}
+	slices.SortFunc(labels, func(a, b string) int {
+		return len(b) - len(a)
+	})
+	return labels
+}
+
+func flipMermaidGraphHorizontal(diagram string, width int, labels []string) string {
```
docker-agent, line 183: **[low] Labels containing box-drawing characters are double-processed in `flipMermaidGraphHorizontal`**

`flipMermaidGraphHorizontal` has two phases:
1. **Label reversal** — `strings.ReplaceAll(diagram, label, reverseMermaidGraphemes(label))` replaces each label with its grapheme-reversed form. For single graphemes (e.g. `╭`), `reverseMermaidGraphemes` is a no-op; the character stays `╭`.
2. **Per-cluster flip** — the function then iterates every grapheme cluster and applies `flipMermaidHorizontalCluster`, which maps `╭→╮`, `╮→╭`, `╰→╯`, `╯→╰`, `├→┤`, `┤→├`, `▶→◀`.

Because phase 2 is applied globally — including to characters that are inside user-supplied node/edge labels — any label containing one of those box-drawing or arrow characters (e.g. a node named `A ▶ B` or `╭foo`) will have that character flipped as if it were a structural border, producing corrupted label output for RL diagrams.

**Trigger**: A Mermaid flowchart with a node or edge label that contains `╭`, `╮`, `╰`, `╯`, `├`, `┤`, or `▶` rendered with `RL` direction.

**Fix**: Skip `flipMermaidHorizontalCluster` for grapheme clusters that fall within a (already-reversed) label region, or alternatively perform the cluster flip first (on the original diagram) and then do the label text reversal as a separate pass that is exempt from `flipMermaidHorizontalCluster`.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly adds horizontal (LR/RL) and vertical (BT) flowchart direction support to the Mermaid TUI renderer, replacing `go-runewidth` with `github.com/charmbracelet/x/ansi` for proper grapheme-cluster-aware width calculations, and polishing box-drawing characters with rounded corners. The implementation is consistent and well-structured. One minor rendering edge case was identified in the new RL flip path.

pkg/tui/components/mermaid/render_graph.go
rumpl: Fixed in f46b5b046. RL mirroring now treats node and edge labels as opaque segments, so box-drawing and arrow glyphs inside user-supplied labels are preserved while only structural diagram glyphs are mirrored. Added a regression test covering these characters in both node and edge labels.

docker-agent: Perfect fix! The new approach in `flipMermaidGraphHorizontal` treats labels as atomic segments via `mermaidLabelPrefix`, so they're reversed whole without applying `flipMermaidHorizontalCluster` to their interior characters. The regression test in `TestRenderRightToLeftFlowchartPreservesStructuralCharactersInLabels` confirms that node labels like `╭ A ▶ B ┤` and edge labels like `╰ edge ▶ ╮` remain intact in RL diagrams.

Thanks for the quick turnaround and thorough test coverage!

Sayt-0 (approved):

PR #3717, opened 2026-07-17
fix: address Mermaid subgraph review findings

rumpl:
Addresses review comments from https://github.com/docker/docker-agent/pull/3716

### DIFF
--- a/pkg/mermaid/parser_flowchart.go
+++ b/pkg/mermaid/parser_flowchart.go
@@ -1,9 +1,6 @@
 package mermaid
 
-import (
-	"slices"
-	"strings"
-)
+import "strings"
 
 func parseFlowchart(doc *Document, statements []string) {
 	var subgraphStack []int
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The two changes in this PR are both correct:

1. **parser_flowchart.go** — Removing the global dedup guard is intentional and correct. The old guard prevented a node from being assigned to a second subgraph even when the Mermaid source explicitly placed it there (a valid pattern). The stack-based logic already ensures a node is only added to the *current* innermost subgraph; `appendUniqueMermaid` correctly deduplicates within that subgraph's slice.

2. **render_subgraph.go** — The `right` bound arithmetic is correct. With `rightBorder = " │"` the rendered column layout is: `before`(W cols) + `│`(1) + ` `(1) + `text`(T) + ` `(1) + `│`(1). Setting `x = W+2` (start of text) and `right = x + T + 1 = W+T+3` lands exactly on the right border character `│`, matching what `drawMermaidSubgraphBox` expects.

Sayt-0 (approved):

PR #3720, opened 2026-07-17
Render markdown images in agent responses

rumpl:
Screenshot 2026-07-17 at 16 10 45

### DIFF
--- a/pkg/tui/components/message/message.go
+++ b/pkg/tui/components/message/message.go
@@ -1,7 +1,9 @@
 package message
 
 import (
+	"context"
 	"fmt"
+	"maps"
 	"strings"
 
 	tea "charm.land/bubbletea/v2"
@@ -10,6 +12,7 @@ import (
 	"github.com/docker/docker-agent/pkg/tui/components/markdown"
 	"github.com/docker/docker-agent/pkg/tui/components/spinner"
 	"github.com/docker/docker-agent/pkg/tui/core/layout"
+	tuiimage "github.com/docker/docker-agent/pkg/tui/image"
 	"github.com/docker/docker-agent/pkg/tui/styles"
 	"github.com/docker/docker-agent/pkg/tui/types"
 )
@@ -23,7 +26,7 @@ const (
 type Model interface {
 	layout.Model
 	layout.Sizeable
-	SetMessage(msg *types.Message)
+	SetMessage(msg *types.Message) tea.Cmd
 	SetSelected(selected bool)
 	SetHovered(hovered bool)
 	CodeBlocks() []markdown.CodeBlock
@@ -77,6 +80,22 @@ type messageModel struct {
 	// dominates a long session, and they are not worth keeping for messages
 	// that are unlikely to be re-rendered hot.
 	finalized bool
+
+	markdownImages  map[string]tuiimage.Inline
+	loadingImages   map[string]bool
+	markdownImageID int
+}
+
+type markdownImagesLoadedMsg struct {
+	target *messageModel
+	images map[string]tuiimage.Inline
+}
+
+type markdownImageRenderedMsg struct{}
+
+type markdownImagePlaceholder struct {
+	token string
+	lines []string
 }
 
 // renderCache stores the most recent Render result keyed by the inputs that
@@ -93,6 +112,7 @@ type renderCache struct {
 	editable  bool
 	sameAgent bool
 	result    string
+	imageID   int
 }
 
 // New creates a new message view
@@ -111,13 +131,17 @@ func New(msg, previous *types.Message) *messageModel {
 
 // Init initializes the message view
 func (mv *messageModel) Init() tea.Cmd {
+	var cmds []tea.Cmd
 	if mv.message.Type == types.MessageTypeSpinner || mv.message.Type == types.MessageTypeLoading {
-		return mv.spinner.Init()
+		cmds = append(cmds, mv.spinner.Init())
 	}
-	return nil
+	if cmd := mv.loadMarkdownImages(mv.message); cmd != nil {
+		cmds = append(cmds, cmd)
+	}
+	return tea.Batch(cmds...)
 }
 
-func (mv *messageModel) SetMessage(msg *types.Message) {
+func (mv *messageModel) SetMessage(msg *types.Message) tea.Cmd {
 	// Un-finalize when the underlying message is changed (e.g. streaming
 	// resumes into this view). Finalize is meant for views that have
 	// permanently lost their actively-streaming status; mutating the message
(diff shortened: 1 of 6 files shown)


Sayt-0 (approved):
cool

pkg/tui/image/image.go
```diff
@@ -84,6 +100,152 @@ func FromToolResult(result *tools.ToolCallResult) []Inline {
...
+	} else {
+		path := ref.Source
+		switch {
+		case parsed == nil || parsed.Scheme == "":
+		case parsed.Scheme == "file", parsed.Scheme == "sandbox":
+			path = parsed.Path
+		default:
+			return Inline{}, false
+		}
+		file, err := os.Open(path)
```
docker-agent, line 235: **[HIGH — security] Arbitrary local file read via `file://` or `sandbox://` image URL in LLM-generated markdown**

`LoadMarkdownReference` accepts `file://` and `sandbox://` URIs in its `else` branch and opens the resolved path with `os.Open` unconditionally. Because this function is called with URLs extracted from assistant message content — i.e., text produced by the LLM — an adversarially crafted or prompt-injected response containing `img` (or any other sensitive path) will cause the agent to read up to 20 MiB of that file into memory.

`FromBytes` will reject non-image MIME types, so the data won't be *rendered*, but:
1. The bytes are still **fully read into process memory** before the check.
2. Any error-path logging, telemetry, or side channel (e.g., timing, memory pressure) that leaks byte counts or partial content could enable exfiltration.

This is a [confused-deputy](https://en.wikipedia.org/wiki/Confused_deputy_problem) attack: the agent reads local files on behalf of LLM-supplied URLs. The `sandbox://` alias doesn't restrict the path at all — it's treated identically to `file://` with the raw `parsed.Path`.

**Suggested fix**: reject `file://` and `sandbox://` URIs in the markdown image loader. Since these schemes are only appropriate for tool-result images (loaded via a separate, trusted path), they should never be accepted when the source is assistant text content:

```go
// In LoadMarkdownReference, replace the sandbox/file case with an explicit rejection:
case parsed.Scheme == "file", parsed.Scheme == "sandbox":
    return Inline{}, false   // never allow local paths from untrusted markdown
```

If local-path support is intentional for a specific future use-case, add an allowlisted directory check (e.g., `strings.HasPrefix(resolved, allowedDir)`).

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/tui/components/message/message.go
```diff
@@ -133,6 +157,38 @@ func (mv *messageModel) SetMessage(msg *types.Message) {
...
+	refs := tuiimage.MarkdownReferences(msg.Content)
+	pending := make([]tuiimage.MarkdownReference, 0, len(refs))
+	if mv.loadingImages == nil {
+		mv.loadingImages = make(map[string]bool)
+	}
+	for _, ref := range refs {
+		if _, loaded := mv.markdownImages[ref.Source]; loaded || mv.loadingImages[ref.Source] {
+			continue
+		}
+		mv.loadingImages[ref.Source] = true
```
docker-agent, line 176: **[MEDIUM — correctness] Failed image URLs are permanently blocked from retrying**

`loadMarkdownImages` marks each URL as in-flight by setting `mv.loadingImages[ref.Source] = true` before the fetch begins. When `markdownImagesLoadedMsg` is processed in `Update`, only the successfully fetched images are copied into `mv.markdownImages`; failed URLs are never removed from `loadingImages`.

The guard check `mv.loadingImages[ref.Source]` then permanently suppresses any retry for those URLs on every subsequent `SetMessage` call. A transient network error, DNS hiccup, or server-side 5xx during streaming will silently leave the image slot blank for the entire session with no recovery path.

If "try once" semantics are intentional (e.g., to avoid hammering a broken server), that should be documented. Otherwise, remove failed URLs from `loadingImages` when processing `markdownImagesLoadedMsg`:

```suggestion
	if loaded, ok := msg.(markdownImagesLoadedMsg); ok && loaded.target == mv {
		for src := range mv.loadingImages {
			if _, success := loaded.images[src]; !success {
				delete(mv.loadingImages, src)
			}
		}
		if len(loaded.images) > 0 {
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/tui/components/message/message.go
```diff
@@ -396,6 +467,80 @@ func (mv *messageModel) render(width int) string {
...
+		if !ok {
+			result = append(result, line)
+			continue
+		}
+		result = append(result, replacement...)
+		shifts = append(shifts, lineShift{line: lineIndex, delta: len(replacement) - 1})
+	}
+	for i := range codeBlocks {
+		for _, shift := range shifts {
+			if shift.line < codeBlocks[i].Line {
```
docker-agent, line 536: **[MEDIUM — correctness] Code-block line-shift miscalculation when an image placeholder appears after the code block's original position**

In `replaceMarkdownImagePlaceholders`, code-block line numbers are adjusted by iterating over all image-placeholder `shifts`. The comparison is:

```go
if shift.line < codeBlocks[i].Line {
    codeBlocks[i].Line += shift.delta
}
```

The problem is that `codeBlocks[i].Line` is **mutated in-place** during the inner loop, so subsequent comparisons use the already-shifted value rather than the original pre-replacement line number.

**Concrete failure scenario:**
- Code block at original rendered line **10**
- Image placeholder A at original line **3** (expands by 5 lines → delta = 5)
- Image placeholder B at original line **12** (expands by 2 lines → delta = 2)

Expected: B is after the code block (original line 12 > original line 10), so B should NOT shift the code block. Correct adjusted line = 10 + 5 = **15**.

Actual with the current code:
1. Shift A: `3 < 10` → `Line = 10 + 5 = 15` ✓
2. Shift B: `12 < 15` (true! compares against mutated value) → `Line = 15 + 2 = 17` ✗

The code block ends up shifted by an extra 2 lines, causing click-to-copy to target the wrong content.

**Fix**: compare against the original line number by saving it before the inner loop:

```suggestion
	for i := range codeBlocks {
		orig := codeBlocks[i].Line
		for _, shift := range shifts {
			if shift.line < orig {
				codeBlocks[i].Line += shift.delta
			}
		}
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR adds markdown image rendering to the TUI. Three bugs were found in the new code — one security vulnerability and two correctness issues.

#### Findings summary
1. **[HIGH — security]** `pkg/tui/image/image.go:235` — LLM-generated `file://` or `sandbox://` image URLs cause the agent to read arbitrary local files (🟢 97/100)
2. **[MEDIUM — correctness]** `pkg/tui/components/message/message.go:176` — Failed image fetches are permanently suppressed; no retry path exists (🟢 100/100)
3. **[MEDIUM — correctness]** `pkg/tui/components/message/message.go:536` — Code-block line offsets are miscalculated when an image placeholder appears after a code block (🟢 100/100)

dgageot (approved):

PR #3721, opened 2026-07-17
fix(tui): address PR #3720 review feedback on markdown image rendering

dgageot:
Three code-review findings left on PR #3720 (https://github.com/docker/docker-agent/pull/3720) were not addressed before merge. This commit fixes all three.

`LoadMarkdownReference` in `pkg/tui/image/image.go` now rejects `file://` and `sandbox://` URI schemes before attempting a fetch. LLM-generated markdown could otherwise craft an image reference that causes the agent to read arbitrary local files on behalf of the user — a confused-deputy attack. Bare relative paths (used for agent-generated images) are unaffected.

In `pkg/tui/components/message/message.go`, failed image fetches are now removed from `loadingImages` rather than left in place. Because the loaded message carries only the refs that succeeded, leaving failures in the set caused them to be treated as already-in-flight on the next `SetMessage` call, permanently suppressing any retry. The fix lets transient failures recover on a subsequent render cycle. The same file also fixes `replaceMarkdownImagePlaceholders`, which was computing code-block line shifts against the post-replacement line number instead of the original one; any code block appearing after an image placeholder ended up shifted by the wrong amount, breaking click-to-copy on those blocks.

Each fix is covered by a new unit test (`TestLoadMarkdownReferenceRejectsLocalSchemes`, `TestFailedMarkdownImageLoadCanRetry`, `TestReplaceMarkdownImagePlaceholdersShiftsCodeBlocksByOriginalLine`). Build, tests, and linter all pass.

### DIFF
--- a/pkg/tui/components/message/message.go
+++ b/pkg/tui/components/message/message.go
@@ -87,8 +87,9 @@ type messageModel struct {
 }
 
 type markdownImagesLoadedMsg struct {
-	target *messageModel
-	images map[string]tuiimage.Inline
+	target    *messageModel
+	requested []tuiimage.MarkdownReference
+	images    map[string]tuiimage.Inline
 }
 
 type markdownImageRenderedMsg struct{}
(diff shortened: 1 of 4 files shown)


dgageot:
cc @rumpl @Sayt-0 wdyt?

rumpl (approved):

PR #3724, opened 2026-07-17
fix(tui): hide no-op slash commands (/settings) in lean mode

aheritier:
Follows #3723.

In LEAN TUI mode, slash commands that are no-ops should not be offered in autocompletion, the command palette, or the slash-command parser. This excludes /settings in lean mode while leaving classic mode unaffected.

A single source of truth now drives both runtime message dropping and disabled slash-command derivation, with a guard test preventing drift.

Validation: task build, task test, and task lint pass.

### DIFF
--- a/pkg/tui/tui.go
+++ b/pkg/tui/tui.go
@@ -8,7 +8,9 @@ import (
 	"log/slog"
 	"os"
 	"os/exec"
+	"reflect"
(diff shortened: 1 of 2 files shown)


Sayt-0 (approved):

rumpl:
Did /settings really show?

PR #3729, opened 2026-07-17
revert: pkg/tui lean-mode machinery from #3724

### LINKED ISSUE #3725: follow-up(tui): #3724 targets inactive lean-mode path and risks hiding valid custom /settings command
## Context

PR #3724 (merged as `eba94215f075f1586b65045356233951eb89750b`) added machinery to `pkg/tui` to hide no-op slash commands such as `/settings` in lean mode. On review, the change appears to target a code path that is no longer user-facing, and it introduces a latent correctness risk.

## Problems

1. **Wrong code path.** The live `--lean` flow branches in `cmd/root/run.go` into `runLeanTUI`, which uses the separate `pkg/leantui` frontend. It does **not** construct `pkg/tui` with `tui.WithLeanMode()`. `pkg/leantui` also already omits `/settings` from its built-in command list, so the PR does not fix the currently active `docker agent run --lean` experience it claims to address.

2. **Correctness risk / command-name collision.** Filtering is done globally on slash-command text. An agent-defined custom command named `settings` (which produces an `AgentCommandMsg` and is not a no-op) would also be removed.

3. **Test gap.** The added tests instantiate the legacy `pkg/tui` model directly and do not exercise CLI routing or the real `pkg/leantui` command list.

4. **Disproportionate complexity.** Reflection and parallel registries add substantial machinery to exclude a single command on an apparently inactive path.

## Suggested follow-up

Either:
- Revert the unnecessary `pkg/tui` lean-mode machinery introduced in #3724, **or**
- First establish a real caller / reproduction for `tui.WithLeanMode`, and add coverage ensuring custom command-name collisions (e.g. a custom `/settings`) remain functional.

## References
- Follows #3724
- Related: #3723

aheritier:
Reverts #3724

Closes #3725

The live `--lean` flow uses `pkg/leantui`, not `pkg/tui` with `tui.WithLeanMode()`. The `pkg/leantui` frontend already omits `/settings` from its command list, so the filtering in `pkg/tui` was applied to an inactive code path. Additionally, the global slash-command text filter would also suppress any agent-defined custom command named `settings`.


rumpl (approved):
sure

PR #3736, opened 2026-07-18
fix(tui): don't panic when the terminal reports a degenerate size

dgageot:
When a terminal reports a degenerate size (0×0 or 1×1), the resize handle in `pkg/tui/tui.go` computed a negative repeat count for `strings.Repeat`, crashing the TUI with `panic: strings: negative Repeat count`. The existing `width <= 0` guard did not account for the horizontal padding subtracted afterwards, so any width narrower than `appPaddingHorizontal` slipped through and produced a negative inner width.

The fix adds a guard on the padded inner width itself. A follow-up commit also handles the status suffix: `lipgloss` ignores non-positive `MaxWidth`, so a paused/working suffix wider than the terminal could overflow the row on very narrow terminals. The suffix logic is extracted into a `lineWithSuffix` helper that truncates the suffix when the terminal is too narrow to accommodate it.

Both fixes are covered by a unit test over tiny widths (`pkg/tui/tui_resize_handle_test.go`), a full-TUI e2e resize cycle through degenerate sizes (`e2e/tui/degenerate_resize_test.go`), and a board-TUI render sweep across degenerate sizes × all dialogs (`pkg/board/tui/degenerate_size_test.go`). All three tests were verified to panic on main without the fix.

### DIFF
--- a/e2e/tui/degenerate_resize_test.go
+++ b/e2e/tui/degenerate_resize_test.go
@@ -17,8 +17,9 @@ func TestChat_DegenerateResizeDoesNotPanic(t *testing.T) {
 
 	d.WaitFor(tuitest.Contains("Type your message here"))
 
-	// Stream an answer while cycling through degenerate sizes so the busy
-	// chrome (working indicator, resize handle suffix) renders too.
+	// Submit a prompt first so the busy chrome (working indicator, resize
+	// handle suffix) can render during the sweep when replay is still
+	// streaming; the sizes themselves are what must never panic.
 	d.Type("What's 2+2?").Enter()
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The degenerate-size panic fix is correct and well-tested. The `innerWidth <= 0` guard properly catches all cases where `width - appPaddingHorizontal` would be non-positive, preventing the negative `strings.Repeat` count that caused the panic. The extracted `lineWithSuffix` helper handles both the overflow and normal paths correctly. The three test layers (unit, e2e, board-TUI sweep) provide solid coverage.

aheritier (approved):

PR #3737, opened 2026-07-18
feat: group session browser by git repo root (worktree-aware)

dgageot:
The past-sessions dialog groups sessions under a "This workspace" header by comparing the current working directory against each session's recorded directory. This breaks with git worktrees: a session started from a worktree, a subdirectory, or the main checkout of the same repository all appear as separate workspaces, even though they share the same codebase.

The fix introduces `pkg/gitroot`, a small package that resolves any directory to its canonical git repository root without shelling out. It walks up the directory tree looking for a `.git` entry. When `.git` is a file (a linked worktree), it reads the `gitdir:` pointer and then follows the `commondir` file to arrive at the main repository root. Bare-repo worktrees resolve to the bare repo directory; submodules are treated as independent roots. The session browser's `workspaceMatcher` now normalizes both the current directory and each stored session directory to their git root (with symlinks resolved first, and results cached), so all worktrees of the same repository fold under one header. The header itself displays the repository root path, consistent with the grouping key.

A follow-up hardening pass bounds `.git` pointer and `commondir` file reads to 4 KiB, cleans the input path before walking, and adds a symlink-resolution fallback so the displayed header stays consistent even when `os.Getwd` returns a symlinked path. The package has thorough tests covering plain repos, subdirectories, linked worktrees with both relative and absolute `commondir` paths, bare-repo worktrees, submodules, dangling and oversized `.git` pointers, and symlinked workspaces.

### DIFF
--- a/pkg/gitroot/gitroot.go
+++ b/pkg/gitroot/gitroot.go
@@ -4,10 +4,16 @@
 package gitroot
 
 import (
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
 )
+
+// maxPointerFileSize bounds reads of .git pointer and commondir files. Valid
+// ones hold a single path, so anything larger is treated as malformed rather
+// than read into memory.
+const maxPointerFileSize = 4096
 
 // Root returns the top-level directory of the repository containing dir,
 // walking up parent directories until a .git entry is found. Linked worktrees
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The new `gitroot` package and its integration into `session_browser.go` are well-implemented. The worktree-aware grouping logic is correct: `.git` pointer files are parsed, `commondir` is resolved for linked worktrees, file reads are bounded to 4 KiB, and symlinks are resolved before comparison. Test coverage is thorough across plain repos, subdirectories, worktrees, bare repos, submodules, and symlinked paths.

No bugs were found in the code introduced by this PR.

aheritier (approved):

PR #3738, opened 2026-07-18
feat(deferred): add fuzzy search matching for tool discovery

Piyush0049:
## Summary

This PR addresses the `TODO: fuzzy search` in the deferred toolset by replacing the rigid substring search with a lightweight fuzzy matching algorithm.

When an agent uses `search_tool` to discover deferred tools, it can now successfully match acronyms or partial words (e.g. searching for "crfil" will correctly match the "create_file" tool).

## Changes Made

* **Fuzzy Match Algorithm**: Added a dependency-free `fuzzyMatch` helper function to `pkg/tools/builtin/deferred/deferred.go` that verifies if all characters in the query string appear in the target string in relative order.
* **Search Update**: Updated `handleSearchTool` to use `fuzzyMatch` against both the tool name and tool description, and removed the inline TODO comment.
* **Test Coverage**: Added explicit `fuzzy search by name` and `fuzzy search by description` test cases in `deferred_test.go` to ensure correctness.

## Validation

* The full `go test ./pkg/tools/builtin/deferred/...` suite passes successfully.
* The files have been formatted cleanly using `gofumpt` to comply with repository standards.

### DIFF
--- a/pkg/tools/builtin/deferred/deferred.go
+++ b/pkg/tools/builtin/deferred/deferred.go
@@ -5,9 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
+	"sort"
 	"strings"
 	"sync"
 
+	"github.com/junegunn/fzf/src/algo"
+	"github.com/junegunn/fzf/src/util"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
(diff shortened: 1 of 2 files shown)


rumpl:
We are using fzf in this project for other things, let's use fzf also for this search, WDYT?

Piyush0049:
> We are using fzf in this project for other things, let's use fzf also for this search, WDYT?

Hmm, I think that would be better. I just removed the custom matcher and swapped it over to use fzf. It's actually way better this way since we can use the match score to rank the results.
I have pushed the changes. Please do let me know your POV and any other changes required.

rumpl (approved):
<3

rumpl:
Amazing, thanks!

PR #3743, opened 2026-07-19
feat(filesystem): add `.agentsignore` to keep files out of the agent's reach

### LINKED ISSUE #3742: Add `.agentsignore` so a project can keep files out of the agent's reach
### Overview

There is no way to tell an agent "never touch these files".

A repository usually contains things the agent has no business reading: `.env`, `*.pem`, `~/.aws`-style credential directories, customer fixtures, licensed data. Today the only options are:

- **`.gitignore` (via `ignore_vcs`, on by default).** This is a *display filter only*. Gitignored paths are omitted from `list_directory`, `directory_tree` and `search_files_content` — but `read_file` opens them without complaint. Verified: `shouldIgnorePath` (`pkg/tools/builtin/filesystem/filesystem.go:805`) has exactly three callsites, all listing/search; `handleReadFile` (`:978`), `handleWriteFile` (`:1372`) and `handleEditFile` (`:897`) never consult it. So `read_file secrets.env` succeeds on a gitignored file.
- **`allow_list` / `deny_list`.** These genuinely block every filesystem operation, with kernel-level `*os.Root` enforcement — but they take **literal directory roots only**. No globs: `vendor/**` is treated as a directory literally named `**`. You cannot say `*.pem`, and you cannot ignore a single file.
- **`permissions`.** Glob-capable, but rules match the argument string *exactly as the model wrote it* — `pkg/runtime/toolexec/permissions.go` passes the raw arguments through with no path resolution. A rule for `secrets.env` is defeated by `./secrets.env`.

So: the mechanism with the right *syntax* can't enforce, and the mechanism that enforces has the wrong syntax.

Meanwhile every adjacent tool has solved this with one file — `.gitignore`, `.dockerignore`, `.npmignore`, `.eslintignore`. Users already know the syntax and reasonably expect it to exist.

### Motivation

**Syntax must be gitignore, not an approximation.** Parse with the same library git uses (`go-git`'s `gitignore.ParsePattern`), so `!` negation, `**`, trailing-slash directories, leading-slash anchoring and escaping all behave identically. Users should never have to learn a dialect.

**No git repository required.** The existing `fsx.VCSMatcher` returns nothing outside a repo (`pkg/fsx/vcs.go:33`), which would silently disable the feature for non-git projects — exactly the throwaway scratch directories where secrets are most likely to sit unprotected. `.agentsignore` must stand alone.

**Not cached process-globally.** The VCS matcher caches by repo root forever with no invalidation (`pkg/fsx/vcs.go:22-27`), so editing `.gitignore` mid-session has no effect. A per-toolset matcher avoids inheriting that.

**Enforce at `resolveAndCheckPath`.** Every path-taking handler in the filesystem toolset already funnels through it (`filesystem.go:635`), so one check covers read/write/edit/mkdir/rmdir. Listing suppression needs a second touch, in `shouldIgnorePath`.

**Independent of `ignore_vcs`.** Setting `ignore_vcs: false` disables `.gitignore` filtering; it must not silently un-hide `.agentsignore` entries, which are a different intent.

**Sync with permissions.** Derive deny rules from the patterns and merge them into the team checker, so `/permissions` shows that a project ignores files and an obvious direct attempt is refused early. This is defence in depth, explicitly *not* the boundary — see below.

### Use cases

_No response_

### Proposed solution

A `.agentsignore` file at the project root, using **gitignore syntax**:

```text
.env
.env.*
*.pem
*.key
!public.key
.aws/
build/
```

Present ⇒ active. No configuration.

Semantics — matched paths are:

| | |
| --- | --- |
| absent from | `list_directory`, `directory_tree`, `search_files_content` |
| refused for | `read_file`, `read_multiple_files`, `write_file`, `edit_file`, `create_directory`, `remove_directory` |

Paths are resolved (symlinks, `./`, `..`) *before* matching, so an ignored file cannot be reached by spelling it differently — the thing the permissions layer cannot do.

The file hides itself: it names exactly what is being kept back, so serving it to the agent would be a map of what to look for.

### Alternatives

_No response_

### Related issues

_No response_

### Additional context

`.agentsignore` governs the filesystem toolset. An agent with the `shell` toolset can still `cat secrets.env`: `shell.go:192` execs the command without inspecting it, and no filesystem guard is consulted. The `lsp` toolset reads *and writes* files directly too (`lsp.go:1405`, `:1426`).

Closing those would mean either command inspection (brittle, trivially evaded by `base64`/`xxd`/a heredoc) or real sandboxing — which the project already offers separately. The right framing is: `.agentsignore` is a strong default that keeps sensitive files out of the agent's view and context, and pairs with `permissions` or sandbox mode when a genuine boundary is needed.

dwin-gharibi:
Adds a `.agentsignore` file — same syntax as `.gitignore` and `.dockerignore` — whose entries the agent cannot list, read, or write.

```text
# .agentsignore
.env
.env.*
*.pem
*.key
!public.key
.aws/
build/
```

Drop it in the project. There is nothing to configure: the file's presence is the opt-in.

Closes #3742.

## Why this doesn't already exist

Three mechanisms are adjacent and none of them does it:

| | Syntax | Blocks read/write? | Why it isn't enough |
| --- | --- | --- | --- |
| `.gitignore` (`ignore_vcs`) | gitignore | **No** | Display filter only — hides from listings while `read_file` still opens the file |
| `allow_list` / `deny_list` | literal dir roots | Yes | No globs at all: `vendor/**` means a directory literally named `**`. Can't express `*.pem` |
| `permissions` | globs | Advisory | Matches the raw argument string with no path resolution, so `./secrets.env` defeats a rule for `secrets.env` |

The one with the right syntax can't enforce; the one that enforces has the wrong syntax.

The `.gitignore` gap is worth stating precisely, because it is the gap this closes: `shouldIgnorePath` (`filesystem.go:805`) has exactly three callsites — `handleListDirectory`, `handleDirectoryTree`, `handleSearchFilesContent`. `handleReadFile`, `handleWriteFile` and `handleEditFile` never call it. **`read_file secrets.env` succeeds today on a gitignored file.** There's a regression test for that now.

## What it does

| Behaviour | Effect |
| --- | --- |
| `list_directory`, `directory_tree` | matched entries omitted |
| `search_files_content` | matched files skipped |
| `read_file`, `read_multiple_files` | refused |
| `write_file`, `edit_file` | refused — including paths that don't exist yet |
| `create_directory`, `remove_directory` | refused |
| `/permissions` | matching deny rules derived automatically |

Paths are resolved before matching — symlinks, `./`, `..` — so an ignored file can't be reached by spelling it differently. The `.agentsignore` file hides itself: it names exactly what's being kept back.

## Implementation

**`pkg/fsx/agentsignore.go`** — `AgentsIgnoreMatcher`, built with go-git's `gitignore.ParsePattern`/`NewMatcher`, so `!` negation, `**`, trailing-slash directories and anchoring behave identically to `.gitignore`. Two deliberate differences from the existing `VCSMatcher`:

- **No git repo required.** `VCSMatcher` returns nothing outside a repository, which would silently disable the feature in exactly the scratch directories where secrets sit unprotected.
- **Not cached process-globally.** `VCSMatcher`'s cache has no invalidation, so an edited `.gitignore` never takes effect. This matcher lives with its toolset.

**`pkg/tools/builtin/filesystem/filesystem.go`** — two touchpoints:

- `resolveAndCheckPath` (the chokepoint every path-taking handler already funnels through) → blocks read/write/edit/mkdir/rmdir in one place.
- `shouldIgnorePath` → suppresses listings and searches. Checked **before** the `ignore_vcs` guard, so `ignore_vcs: false` disables `.gitignore` filtering without silently un-hiding `.agentsignore` entries.

**`pkg/permissions/agentsignore.go`** — derives deny rules from the patterns; `cmd/root/run.go` merges them into the team checker so the ignore list shows up in `/permissions`. Negations are skipped deliberately: a permission list can't express "except this", and emitting `keep.key` without its `!` would deny the very path the author re-included.

A malformed or unreadable `.agentsignore` is a **hard error** at toolset construction, not a warning. Continuing past it would leave the agent reading files the author believed were hidden.

## Limits

`.agentsignore` governs the filesystem toolset. An agent with `shell` can still run `cat secrets.env` — `shell.go:192` execs the command without inspecting it. `lsp` reads *and writes* files directly (`lsp.go:1405`, `:1426`).

Closing those means command inspection (brittle; `base64`, `xxd` and heredocs evade it) or real sandboxing, which the project already offers separately. The framing in the docs is: a strong default that keeps sensitive files out of the agent's view and context, to be paired with `permissions` or sandbox mode when a genuine boundary is needed.

The derived permission rules are best-effort for the same reason — they match unresolved argument strings. The filesystem toolset's check is the part that actually enforces.

I'd rather ship this honestly scoped than imply a guarantee it can't keep. Happy to reshape if you'd prefer it wired elsewhere.

## Verification

Run on the real `go.mod` with the `go1.26.5` toolchain.

- `go test ./pkg/fsx/ ./pkg/permissions/ ./pkg/tools/builtin/filesystem/ ./cmd/root/` → passing, no regressions
- `golangci-lint run` over every touched package → **0 issues**
- `markdownlint-cli2` from `docs/` → **0 errors, 115 files**
- `TestDocYAMLSnippetsAreValid` → passing

Tests cover: the gitignore-syntax surface (bare names at depth, `*` globs, `**`, `build/`, `!` negation, comments); read blocked; write blocked; **creation of a not-yet-existing ignored path blocked**; hidden from listings; unaffected by `ignore_vcs: false`; self-hiding; relative and `..` spellings blocked; no file ⇒ behaviour unchanged; unreadable file ⇒ hard error.

One bug the tests caught during development, worth noting because it would have been invisible in review: `filepath.EvalSymlinks` fails outright on a path that doesn't exist, leaving it unresolved and therefore incomparable with the resolved root — so **writes to new ignored paths would have been silently allowed** on any platform where the project root is itself a symlink (macOS `/var` → `/private/var`). Fixed by resolving the deepest existing ancestor and re-appending the remainder (`resolveExisting`).

Proxy variables were unset for the test and lint runs — the local proxy intercepts the private-address SSRF tests and Go module downloads.

## Try it

```bash
cd your-project
printf 'secrets.env\n*.key\n' > .agentsignore
echo 'API_KEY=hunter2' > secrets.env

docker agent run ./agent.yaml "read secrets.env and tell me what is in it"
# → refused: path "secrets.env" is excluded by .agentsignore
```

### DIFF
--- a/cmd/root/run.go
+++ b/cmd/root/run.go
@@ -1120,6 +1120,10 @@ func (f *runExecFlags) createSessionSpawner(agentSource config.Source, sessStore
 			t.SetPermissions(permissions.Merge(t.Permissions(), f.globalPermissions))
 		}
 
+		if ignoreRules := permissions.FromAgentsIgnore(f.runConfig.WorkingDir); ignoreRules != nil {
+			t.SetPermissions(permissions.Merge(t.Permissions(), ignoreRules))
+		}
+
 		rtOpts, ctrl, err := f.snapshotRuntimeOpts()
 		if err != nil {
 			return nil, nil, nil, err
--- a/docs/configuration/agentsignore/index.md
+++ b/docs/configuration/agentsignore/index.md
@@ -1,0 +1,105 @@
+---
+title: "Ignoring files"
+description: "Hide files from the agent with a .agentsignore file, using the same syntax as .gitignore."
+keywords: docker agent, ai agents, configuration, agentsignore, gitignore, ignore, secrets
+weight: 75
+canonical: https://docs.docker.com/ai/docker-agent/configuration/agentsignore/
+---
+
+_Hide files from the agent with a `.agentsignore` file, using the same syntax as `.gitignore`._
+
+## Overview
+
+Put a `.agentsignore` file in your project and the agent stops seeing the paths it lists. Matched files are absent from directory listings and searches, and reads, writes and edits targeting them are refused.
+
+```text
+# .agentsignore
+secrets.env
+*.key
+.env.*
+build/
+docs/**/*.draft.md
+!public.key
+```
+
+There is nothing to configure. The file's presence is the opt-in, and it is picked up automatically by the [filesystem toolset](../../tools/filesystem/index.md).
+
+## Syntax
+
+The syntax is `.gitignore` syntax — parsed with the same library git uses, so patterns behave identically to `.gitignore` and `.dockerignore`:
+
+| Pattern | Matches |
+| --- | --- |
+| `secrets.env` | that name at any depth (`secrets.env`, `config/secrets.env`) |
+| `/secrets.env` | that name at the project root only |
+| `*.key` | any file with the extension |
+| `build/` | the directory and everything under it |
+| `docs/**/*.draft.md` | nested matches via `**` |
+| `!public.key` | re-includes a path an earlier pattern excluded |
+| `# comment` | ignored, as are blank lines |
+
+## Where the file is found
+
+The nearest `.agentsignore` at or above the working directory is used, so starting a run in a subdirectory still honours the project's file. Patterns are anchored to the directory containing the file, exactly as git anchors to the directory containing `.gitignore`.
+
+Unlike `.gitignore`, **a git repository is not required** — `.agentsignore` works in any directory.
+
+## What it affects
+
+| Behaviour | Effect |
+| --- | --- |
+| `list_directory`, `directory_tree` | matched entries are omitted |
+| `search_files_content` | matched files are skipped |
+| `read_file`, `read_multiple_files` | refused |
+| `write_file`, `edit_file` | refused, including for files that do not exist yet |
+| `create_directory`, `remove_directory` | refused |
+| `permissions` | matching deny rules are derived automatically, so `/permissions` shows them |
+
+Paths are resolved before matching — symlinks, `./` prefixes and `..` segments are all normalised — so an ignored file cannot be reached by spelling it differently.
+
+The `.agentsignore` file is itself always hidden: it names the very things being kept back, so handing it to the agent would be a map of what to look for.
+
+> [!NOTE]
+> `.agentsignore` is independent of the filesystem toolset's [`ignore_vcs`](../tools/index.md) option. Setting `ignore_vcs: false` turns off `.gitignore` filtering but does **not** un-hide `.agentsignore` entries.
+
+## Relationship to `.gitignore`
+
+They are separate, and they do different amounts of work.
+
+`.gitignore` is respected by default (`ignore_vcs`), but only as a **display filter**: gitignored files are hidden from listings and searches while `read_file` still opens them. That is reasonable for build output, which is noise rather than secret.
+
+`.agentsignore` is for content the agent should not have at all, so it blocks reads and writes as well. Use `.gitignore` for noise, `.agentsignore` for secrets.
+
+## Limits
+
+> [!WARNING]
+> `.agentsignore` governs the filesystem toolset. It is **not** a sandbox.
+>
+> An agent with the [`shell`](../../tools/shell/index.md) toolset can still run `cat secrets.env`, because the shell runs commands the toolset never inspects. The same applies to any toolset that reaches the filesystem on its own, such as [`lsp`](../../tools/lsp/index.md).
+>
+> Treat `.agentsignore` as a strong default that keeps sensitive files out of the agent's view and context — not as a boundary against an agent actively trying to reach them. When you need a real boundary, combine it with [permissions](../permissions/index.md) that restrict `shell`, or run in [sandbox mode](../sandbox/index.md).
+
+The derived permission rules are best-effort for the same reason: permission patterns match the argument string as the model wrote it, without resolving it first, so `./secrets.env` can slip past a rule written for `secrets.env`. The filesystem toolset's own check resolves paths first and is the part that actually enforces.
+
+## Example
+
+```text
+# .agentsignore
+
+# Secrets
+.env
+.env.*
+secrets.env
+*.pem
+*.key
+!public.key          # this one is safe to read
+
+# Credentials directories
+.aws/
+.ssh/
+
+# Large build output the agent doesn't need
+build/
+dist/
+node_modules/
+```
--- a/docs/data/nav.yml
+++ b/docs/data/nav.yml
@@ -40,6 +40,8 @@
       url: /configuration/hooks/
     - title: Permissions
       url: /configuration/permissions/
+    - title: Ignoring Files
+      url: /configuration/agentsignore/
     - title: Sandbox Mode
       url: /configuration/sandbox/
     - title: Structured Output
--- a/pkg/fsx/agentsignore.go
+++ b/pkg/fsx/agentsignore.go
@@ -1,0 +1,157 @@
+package fsx
+
+import (
+	"bufio"
+	"errors"
+	"io/fs"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/go-git/go-git/v5/plumbing/format/gitignore"
+)
+
+const AgentsIgnoreFile = ".agentsignore"
+
+type AgentsIgnoreMatcher struct {
+	root    string
+	matcher gitignore.Matcher
+}
+
+func FindAgentsIgnore(startDir string) string {
+	dir, err := filepath.Abs(startDir)
+	if err != nil {
+		return ""
+	}
+	for {
+		candidate := filepath.Join(dir, AgentsIgnoreFile)
+		if info, err := os.Stat(candidate); err == nil && !info.IsDir() {
+			return candidate
+		}
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			return ""
+		}
+		dir = parent
+	}
+}
+
+func NewAgentsIgnoreMatcher(startDir string) (*AgentsIgnoreMatcher, error) {
+	path := FindAgentsIgnore(startDir)
+	if path == "" {
+		return nil, nil
+	}
+	patterns, err := readAgentsIgnorePatterns(path)
+	if err != nil {
+		return nil, err
+	}
+	if len(patterns) == 0 {
+		return nil, nil
+	}
+	root := filepath.Dir(path)
+	if resolved, err := filepath.EvalSymlinks(root); err == nil {
+		root = resolved
+	}
+	slog.Debug("Loaded .agentsignore", "file", path, "patterns", len(patterns))
+	return &AgentsIgnoreMatcher{root: root, matcher: gitignore.NewMatcher(patterns)}, nil
+}
+
+func readAgentsIgnorePatterns(path string) ([]gitignore.Pattern, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil, nil
+		}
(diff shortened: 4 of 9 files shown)


dwin-gharibi:
@Sayt-0

aheritier (commented):

---

## Review of `feat/agentsignore` — `.agentsignore` filesystem feature

**Overall**: Well-engineered PR with a clear security model, honest documentation of limitations, and solid test coverage for the real enforcement path. The architecture is sound: a path-resolved filesystem-layer check (the actual security gate) plus a best-effort advisory permission layer for visibility in `/permissions`. A few things worth flagging below.

---

### Medium — `read_multiple_files:paths` advisory deny rule can never fire

**`pkg/permissions/agentsignore.go` — `pathToolArgs` / `permissionGlobsFor`**

`read_multiple_files` takes a `paths` **array**, not a scalar string. The permission evaluator's `argToString` handles unknown types via `fmt.Sprintf("%v", v)`, which formats an array as `[elem1 elem2]`. The generated deny rule `read_multiple_files:paths=secrets.env` therefore never matches because the glob `secrets.env` is compared against `[secrets.env README.md]` (or similar). The real per-element enforcement via `resolveAndCheckPath` in `handleReadMultipleFiles` is correct and unaffected — but the advisory rule gives a false impression of coverage in `/permissions` output.

Options: drop `read_multiple_files` from `pathToolArgs` (it cannot be represented as a scalar glob), or add an explicit comment documenting why it never fires.

---

### Low — Negation creates a permissions-display / enforcement mismatch

**`pkg/permissions/agentsignore.go` — `permissionGlobsFor`**

When `.agentsignore` contains:
```
*.key
!public.key
```

The filesystem layer correctly re-includes `public.key` (gitignore negation semantics). The advisory permissions layer, however, emits a `deny` entry covering `*.key` (which also covers `public.key`) and silently drops `!public.key`. The `/permissions` display will therefore show `public.key` operations as denied even though the agent can actually read it.

The PR author notes this is intentional ("a permission list can't express 'except this'"), and the Limits section acknowledges best-effort. A `// NOTE:` comment in `permissionGlobsFor` explaining the consequence would prevent future readers from assuming the display is authoritative when negations are present.

---

### Low — Bare-name patterns generate only 1-level nesting in the advisory layer

**`pkg/permissions/agentsignore.go` — `permissionGlobsFor`**

`secrets.env` expands to `{secrets.env, secrets.env/*, */secrets.env, */secrets.env/*}`. A file at `a/b/secrets.env` is correctly blocked by the filesystem layer (gitignore bare names match at any depth) but does not match any of the four advisory globs. The enforcement gap is zero; the `/permissions` display just underreports coverage. This is acknowledged in the docs.

---

### Info — `resolveExisting` adds EvalSymlinks overhead per path operation when `.agentsignore` is present

**`pkg/fsx/agentsignore.go` — `Match` / `resolveExisting`**

Each `resolveAndCheckPath` call now invokes `agentsIgnore.Match(resolved)`, which calls `filepath.Abs` + `resolveExisting`. For a path with N non-existent ancestors, `resolveExisting` makes O(N) `EvalSymlinks` syscalls. When `.agentsignore` is absent, `agentsIgnore` is nil and `Match` returns immediately — zero cost. When present there is new syscall overhead on every path-taking tool call. Acceptable for normal workloads, but worth noting for latency-sensitive scenarios.

---

### Positive observations

- The dual-layer design (real enforcement via `resolveAndCheckPath` + `shouldIgnorePath`, advisory layer via the permissions checker) is architecturally clean and the split is well-documented.
- `resolveExisting` correctly handles the macOS `/var → /private/var` symlink edge case for non-existent paths — would have silently allowed writes on symlinked project roots without it.
- Hard-failing on an unreadable `.agentsignore` (rather than warn-and-continue) is the right security default.
- `.agentsignore` hiding itself is a good design touch.
- `ignore_vcs: false` independence is tested and correct: agentsIgnore is checked before the `ignoreVCS` guard in `shouldIgnorePath`.
- Test coverage is comprehensive across all three packages, including the non-existent-path case.
- Docs honestly scope the shell/lsp bypass limitation.

aheritier:

Addressed in commit a5a83a82f1d616c116ad6f0e3f5e860d6e5b106f.

**Medium — `read_multiple_files` removed from `pathToolArgs`**: Dropped the entry and replaced it with an explanatory comment documenting why (`paths` is an array; the evaluator formats it as `[elem1 elem2]` so a scalar glob can never match). Real enforcement via `resolveAndCheckPath` in `handleReadMultipleFiles` is unaffected.

**Low — negation mismatch**: Added a `// NOTE:` comment in `permissionGlobsFor` at the `!`-prefix early-return, explaining that negation patterns are intentionally dropped (a deny-list cannot express negation), and that the advisory `/permissions` display may therefore show false denials for negated paths while actual filesystem access is correct.

**Low (bare-name nesting) / Info**: No code change — acknowledged limitations already covered in the docs.

aheritier (approved):

**Re-review of commit a5a83a82f1d616c116ad6f0e3f5e860d6e5b106f — APPROVE**

Both previously-flagged items are cleanly resolved:

- ✅ **Medium — `read_multiple_files` in `pathToolArgs`:** entry removed; a 4-line comment explains why (the `paths` arg is always an array, scalar glob matching can never fire) and confirms per-path enforcement remains intact via `resolveAndCheckPath`.
- ✅ **Low — negation-pattern display/enforcement mismatch:** `permissionGlobsFor` now carries a `// NOTE:` block with the exact example (`*.key` + `!public.key`), names the advisory `/permissions` display as the only affected surface, and labels it an accepted known limitation.

Three additional informational observations (non-blocking, same accepted-limitation class):
- Anchored patterns (e.g. `/secrets.env`) are over-broadened in the advisory display layer (stripped `/` → no `/` present → `*/secrets.env` wildcard generated); enforcement in the filesystem layer is correct. Covered by the existing "best-effort" doc disclaimer.
- `FromAgentsIgnore` silently returns `nil` on read errors; safe in practice because `CreateToolSet` propagates the same error loudly and the session never starts.
- `TestPermissionGlobsFor` has no anchored-pattern test case — a `/secrets.env` row would document the known over-broadening.

None of these are blocking. Implementation is correct, nil-safe, well-tested across three packages, and documentation is accurate with appropriate caveats.

aheritier:

@Sayt-0 CI is all green and the automated review is approved — this PR is ready for a human review when you have a moment.

aheritier (approved):

**Fresh verification pass — APPROVE**

This review re-approves at HEAD `a5a83a82f1d616c116ad6f0e3f5e860d6e5b106f`, identical to the commit reviewed and approved on 2026-07-19. The prior approval was auto-dismissed when #3739 merged into `main` and triggered new CI runs.

**Verification checks:**
- ✅ PR HEAD is unchanged — no new commits since the previous APPROVE
- ✅ No file overlap with #3739 (which touched `cmd/root/agent_picker*`, `pkg/tools/builtin/api/`) — zero semantic conflict
- ✅ PR is mergeable (no conflicts with updated `main`)
- ✅ All CI checks pass: lint, build-and-test, pa11y, markdownlint, validate-upstream, canonical-check, license-check, link-check, build-image (amd64 + arm64), llms-txt-check

The substantive review findings from the previous pass stand:
- Both previously-flagged items (dropped `read_multiple_files` from advisory deny list with explanatory comment; `// NOTE:` block in `permissionGlobsFor` for negation-pattern mismatch) remain correctly addressed.
- The three informational observations (anchored-pattern over-broadening, `FromAgentsIgnore` nil-return, missing anchored-pattern test case) are non-blocking and unchanged.

Sayt-0 (approved):

PR #3752, opened 2026-07-20
feat: allow configuring compaction_model on agents

dgageot:
Previously, `compaction_model` could only be set inside a model config block. This was limiting because agents are often defined with a named model reference rather than an inline config, and there was no way to override just the compaction model without inlining the whole model definition. This change lets you set `compaction_model` directly on an agent, as a natural sibling to the existing `compaction_threshold` field.

The resolution order mirrors `compaction_threshold`: a `compaction_model` set on the agent's model config takes precedence; the agent-level value is the fallback; when neither is set, compaction reuses the agent's own model. The agent-level `compaction_model` accepts the same format as the model-level one — either a named model reference or an inline `provider`/`model` spec. Harness agents reject `compaction_model` at validation time since the harness manages its own context compaction and the field would otherwise be silently ignored.

New teamloader tests cover the agent-level named ref, inline ref, an unknown ref that should fail loading, and model-level precedence over agent-level. A working example is in `examples/compaction_model.yaml`.


melmennaoui (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The compaction_model agent-level fallback is implemented cleanly. The resolution order (model-level → agent-level → nil) correctly mirrors the existing compaction_threshold pattern. The harness validation guard is placed correctly inside validateHarness() and will fire before the harness type switch. The schema addition matches the Go struct semantics. No bugs introduced by this PR.

PR #3754, opened 2026-07-20
feat(tui): open the /cost dialog when clicking the sidebar cost reading

dgageot:
The sidebar already displays token usage and estimated cost, but there was no way to click into the detail view — users had to know about the `/cost` command to see the breakdown. This makes the cost reading interactive, consistent with how other sidebar elements behave.

Clicking the token usage/cost reading in the sidebar now opens the `/cost` dialog, the same dialog triggered by typing `/cost` in the input. A new `ClickUsage` hit-test result was added to the sidebar's hit-testing logic. The vertical Token Usage section records its line range while rendering, and the collapsed band correctly hit-tests the usage reading in both layouts: when it shares the working-directory row (right-aligned) and when it occupies its own row. The chat page maps the new `TargetSidebarUsage` to `ShowCostDialogMsg` on left-click.

A follow-up fix ensures clicks on the sidebar scrollbar column and gap are never swallowed by the content click zones (usage, agents, title), so scrollbar-driven scrolling continues to work correctly. The fix also refreshes stale doc comments on `ClickResult` variants. New tests cover vertical mode, collapsed shared-line, collapsed own-line, hidden-usage, and the scrollbar regression.

### DIFF
--- a/pkg/tui/components/sidebar/sidebar.go
+++ b/pkg/tui/components/sidebar/sidebar.go
@@ -337,6 +337,11 @@ type model struct {
 
 	// Agent click zones: maps content line index to agent name for click detection
 	agentClickZones map[int]string // content line -> agent name
+	// Token Usage click zone: half-open content-line range [start, end) of the
+	// vertical Token Usage section, recorded while rendering. Empty when the
+	// section is hidden.
+	usageZoneStart int
+	usageZoneEnd   int
 	// agentLineOwners records, per rendered agent-section body line, which agent
 	// emitted it (empty for blank separators). It is produced during agentInfo
 	// rendering so click zones can be registered explicitly rather than inferred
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR introduces `ClickUsage` hit-testing to the sidebar so clicking the token usage/cost reading opens the `/cost` dialog. The implementation is correct and the new tests provide good coverage across vertical mode, collapsed shared-line, collapsed own-line, hidden-usage, and the scrollbar regression.

The drafter found no high- or medium-severity issues. All observations were low-severity edge cases with negligible practical impact:

- `appendSection` correctly sets `usageZoneStart` to the index *after* gap lines (not including them), so gap-line clicks are not falsely attributed to `ClickUsage`.
- The scrollbar exclusion guard (`cachedNeedsScrollbar && adjustedX >= contentWidth`) correctly relies on a prior `View()` call, which always precedes click processing in practice.
- The collapsed shared-line usage check has no explicit right-bound column guard, but out-of-bounds clicks don't reach the sidebar handler in normal operation.

No changes requested.

gtardif (approved):

PR #3761, opened 2026-07-20
Enhance the agents API

trungutt:
Enhance the agents API.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -16,9 +16,10 @@ type Message struct {
 
 // Agent represents an agent in the API
 type Agent struct {
-	Name        string `json:"name"`
-	Description string `json:"description"`
-	Multi       bool   `json:"multi"`
+	Name        string   `json:"name"`
+	Description string   `json:"description"`
+	Multi       bool     `json:"multi"`
+	Commands    []string `json:"commands,omitempty"`
 }
 
 // ErrCodeUnknownSession is the [ErrorResponse.Code] for a request that names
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are correct and well-implemented:

- **`Commands []string` with `omitempty`** — older clients silently ignore the new field; newer clients pointing at an older server see an absent list. Backward-compatible by design, no coordinated release required.
- **Map iteration + `slices.Sort`** — Go map iteration is non-deterministic; collecting keys then sorting produces a deterministic, stable API response. Correct approach.
- **Nil-vs-empty handling** — when the root agent has no commands the `commands` variable stays `nil`, which `omitempty` omits from JSON output. No spurious empty arrays emitted to clients.
- **`cfg.Agents.First()` as root** — correct choice for the root agent whose slash commands are the ones clients enumerate.
- **Test coverage** — `TestServer_ListAgents` exercises both the empty-commands (single-agent) and populated-commands (multi-agent) cases with a sorted assertion matching the sorted output.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly exposes root-agent command names in `/api/agents`. The implementation correctly:
- Extracts map keys from `root.Commands` and sorts them deterministically with `slices.Sort`
- Guards against the empty-commands case with `if len(root.Commands) > 0`, leaving `commands` as `nil` so the `omitempty` tag suppresses the field when there are no commands
- Avoids a double call to `cfg.Agents.First()` by caching the result in `root`
- Adds test coverage for the happy path (multi-agent config with commands, single-agent configs without)

No confirmed or likely bugs found.

Sayt-0 (approved):

PR #3764, opened 2026-07-20
feat: support Mermaid state diagram directions

rumpl:


docker-agent (commented):
### Assessment: 🟢 APPROVE

No bugs introduced by this PR. The direction-parsing logic and render path change are correct:

- The `direction` keyword branch in `parseState` always executes `continue` — the only potentially affected syntax (`state direction as Foo`) uses an unquoted label, which the existing `state` branch would reject via `readQuoted()` anyway, so no valid statement is dropped.
- Switching `DiagramState` from `drawMermaidGraph` to `drawMermaidFlowchart` with an empty direction string safely falls through to the same `drawMermaidGraph` call with identical arguments — no behavioural regression for undirected state diagrams.

dgageot (approved):

PR #3767, opened 2026-07-21
fix(permissions): keep colons in argument values from truncating patterns

### LINKED ISSUE #3766: Permission rules fail open when an argument value contains `:` (`parsePattern` truncates the value)
### Description

`parsePattern` in `pkg/permissions/permissions.go` splits the whole pattern on `":"` and then walks the
segments. Once it has started collecting argument conditions, any later segment **without** an `=` is
silently discarded. So any rule whose argument value legitimately contains a colon - a URL, a Windows
drive path, a `key:value` fragment - is truncated at the first colon, and the rule matches something
the operator never wrote.

`fetch:url=https://example.com/*` becomes `argPatterns["url"] = "https"`. The `//example.com/*` segment is
dropped. `"https"` never matches a real URL, so a **`deny`** rule written this way never fires and the
call is allowed through.

This is the same fail-open class as #3604, in the same package, one function earlier in the path. #3605
fixed `matchGlob` so wildcards cross separators - but with a colon in the value the pattern never reaches
`matchGlob` intact, so the hole is still open for URLs. URLs are probably the single most common thing
an operator would want to write a `deny` rule against, which is what makes this worth a separate fix.

### Expected Behavior

The whole argument value survives parsing. `fetch:url=https://example.com/*` should deny
`https://example.com/home`.

### Actual Behavior

The value is truncated at the first colon and the rule goes inert:

| Pattern | Value the operator wrote | Value actually parsed |
|---|---|---|
| `fetch:url=https://example.com/*` | `https://example.com/*` | `https` |
| `shell:cwd=C:\work\*` | `C:\work\*` | `C` |
| `shell:cmd=echo a:b=c` | `echo a:b=c` | `echo a` + a phantom second condition `b="c"` |

The third row is a distinct failure mode: because `b=c` contains an `=`, it is not dropped but read as a
*second* argument condition. The rule now additionally requires an argument named `b`, which no tool
supplies, so it can never match anything at all.

### Steps to Reproduce

Config-level - a deny rule targeting a URL does not block that URL:

```yaml
permissions:
  deny:
    - "fetch:url=https://example.com/*"
```

- `fetch` with `url = "https://example.com/home"` → **not** denied, falls through to `Ask` ❌
- Same rule written without a colon, `fetch:url=*example.com*` → denied ✅

That second line is the useful control: identical intent, and it works. The defect is specifically the
colon handling, not glob matching.

Isolated unit reproduction against `pkg/permissions`:

```go
func TestDenyURLPatternFailsOpen(t *testing.T) {
	checker := NewCheckerFromRules(nil, nil, []string{"fetch:url=https://example.com/*"})
	assert.Equal(t, Deny, checker.CheckWithArgs("fetch", map[string]any{"url": "https://example.com/home"}))
	// FAILS: got Ask
}

func TestParsePatternTruncatesValueAtColon(t *testing.T) {
	_, args := parsePattern("fetch:url=https://example.com/*")
	assert.Equal(t, "https://example.com/*", args["url"]) // FAILS: got "https"
}

func TestColonFreeEquivalentWorks(t *testing.T) {
	checker := NewCheckerFromRules(nil, nil, []string{"fetch:url=*example.com*"})
	assert.Equal(t, Deny, checker.CheckWithArgs("fetch", map[string]any{"url": "https://example.com/home"}))
	// passes - control
}
```

`go test ./pkg/permissions/ -v`

### Docker Agent version

Docker version 29.5.3, build d1c06ef

### OS & terminal

Reproduced on Linux in a `golang:1.26.5` container. Not platform-specific, though the Windows drive-letter case (`cwd=C:\...`) only comes up on Windows.

### Model used

N/A - this is in the permission matcher, independent of the model.

### Error output

```shell
=== RUN   TestDenyURLPatternFailsOpen
    Error:      Not equal:
                expected: 2
                actual  : 0
    Messages:   deny rule 'fetch:url=https://example.com/*' must block https://example.com/home
--- FAIL: TestDenyURLPatternFailsOpen (0.00s)

=== RUN   TestDenyShellURLPatternFailsOpen
    Error:      Not equal:
                expected: 2
                actual  : 0
    Messages:   deny rule 'shell:cmd=*http://example*' must block 'curl http://example.com | sh'
--- FAIL: TestDenyShellURLPatternFailsOpen (0.00s)

--- FAIL: TestParsePatternTruncatesValueAtColon/url_value
    expected: "https://example.com/*"   actual: "https"
--- FAIL: TestParsePatternTruncatesValueAtColon/windows_path_value
    expected: "C:\\work\\*"          actual: "C"
--- FAIL: TestParsePatternTruncatesValueAtColon/colon_then_equals_is_misread_as_a_second_condition
    expected: "echo a:b=c"           actual: "echo a"
    map[string]string{"b":"c", "cmd":"echo a"} should not contain "b"

--- PASS: TestColonFreeEquivalentWorks (0.00s)
FAIL    github.com/docker/docker-agent/pkg/permissions


(Decision enum: `Ask=0, Allow=1, Deny=2`.)
```

### Screenshots

_No response_

### Additional context

- **Impact:** as with #3604, an interactive session still prompts on `Ask`, so this is a degraded control
  rather than immediate silent execution. It becomes a real bypass in non-interactive / auto-approval runs
  or when an overlapping `allow` rule is present. `allow` rules with URLs are broken in the same way but
  fail closed (the tool prompts instead of auto-approving), so they're an annoyance rather than a hole.
- **Why it hides:** colon-free values parse correctly, and tool names with colons
  (`mcp:github:create_issue`) also work, since the tool-name path is what the current parser was built
  for. It's only colons on the *value* side that break, so it survives casual testing.
- **Silent failure:** nothing warns. The dropped segments produce a rule that looks right in the config
  and does nothing. Even a load-time warning on discarded segments would have surfaced this.

parveshsaini:
## What

`parsePattern` split a permission pattern on every `:`, so any argument value
containing a colon was truncated. A `deny` rule written against a URL —
`fetch:url=https://evil.com/*` — parsed as `url = "https"`, which never matches a
real URL, so the rule silently never fired (fail-open). Windows drive paths
(`shell:cwd=C:\work\*` → `C`) were corrupted the same way.

This parses the tool name up to the first `:key=` argument condition, then takes
each value up to the *next* `:key=` boundary rather than the next colon, so
colons inside a value are preserved. Tool names with colons
(`mcp:github:create_issue`) are unaffected — they were already handled by
splitting at the first `:key=`, and that behaviour is unchanged.

## Why

This is the sibling of #3605. That fix made `matchGlob`'s wildcards cross
path separators, but with a colon in the value the pattern never reaches
`matchGlob` intact — so the same fail-open direction was still reachable for URLs
and Windows paths, which are exactly the kind of value an operator writes a
`deny` rule against. In non-interactive / auto-approval runs, or alongside an
overlapping `allow`, a corrupted `deny` is a real bypass. `allow` rules with
colon values were broken too, but those fail closed.

## Tests

Added to `pkg/permissions/permissions_test.go`:
- `TestParsePattern` rows for URL, Windows-drive, and mixed value-colon +
  second-condition patterns.
- `TestDenyRuleWithColonInValue` — end-to-end via `CheckWithArgs`: a `deny` rule
  with a URL value now returns `Deny`, plus a shell-command-with-URL case and a
  colon-free control.

Each new assertion fails against `main` and passes with this change; the existing
`TestParsePattern` rows (including `mcp:github:create_issue`) are untouched.

### Verification (in `golang:1.26.5` + `golangci-lint:v2.12.2`, matching CI)

```
$ go test ./pkg/permissions/...
ok  	github.com/docker/docker-agent/pkg/permissions

$ golangci-lint run ./pkg/permissions/...
0 issues.

$ gofmt -l pkg/permissions/    # (no output)
```

Confirming the tests are meaningful — reverting only `permissions.go` to `main`
while keeping the new tests:

```
--- FAIL: TestParsePattern/fetch:url=https://example.com/*
        expected: map[string]string{"url":"https://example.com/*"}
        actual  : map[string]string{"url":"https"}
--- FAIL: TestParsePattern/shell:cwd=C:\work\*
        expected: map[string]string{"cwd":"C:\\work\\*"}
        actual  : map[string]string{"cwd":"C"}
--- FAIL: TestDenyRuleWithColonInValue
        expected: 2 (Deny)   actual: 0 (Ask)
        Messages: deny rule with a URL value must block the matching URL
```

---

### DIFF
--- a/pkg/permissions/permissions.go
+++ b/pkg/permissions/permissions.go
@@ -162,61 +162,32 @@ func (c *Checker) DenyPatterns() []string {
 	return c.denyPatterns
 }
 
+// argCondition matches the ":key=" boundary that starts an argument condition.
+// Tool names may contain colons ("mcp:github:create_issue") and so may values
+// (URLs, Windows drive paths), so a bare ":" is not a separator: only a ":"
+// immediately followed by "key=" is.
+var argCondition = regexp.MustCompile(`:(\w+)=`)
+
 // parsePattern parses a permission pattern into tool name pattern and argument conditions.
 // Pattern format: "toolname" or "toolname:arg1=val1:arg2=val2"
-// Returns the tool pattern and a map of argument patterns.
-//
-// The tool name runs up to the first ":key=" argument condition, so tool names
-// with colons (like "mcp:github:create_issue") are preserved. Argument values
-// may themselves contain colons (URLs like "https://…", Windows paths like
-// "C:\…"): a value extends up to the next ":key=" boundary rather than the next
-// colon, so such values survive intact.
+// The tool name runs up to the first ":key=" boundary and each value up to the
+// next one, so colons inside tool names and values are preserved.
 func parsePattern(pattern string) (toolPattern string, argPatterns map[string]string) {
 	argPatterns = make(map[string]string)
 
-	// The tool name runs up to the first ":key=" condition; each condition then
-	// runs up to the next one, so colons inside a value are kept.
-	toolPattern, rest := cutAtArgCondition(pattern)
-	for rest != "" {
-		condition, next := cutAtArgCondition(rest)
-		key, value, _ := strings.Cut(condition, "=")
-		argPatterns[key] = value
-		rest = next
+	matches := argCondition.FindAllStringSubmatchIndex(pattern, -1)
+	if len(matches) == 0 {
+		return pattern, argPatterns
 	}
 
-	return toolPattern, argPatterns
-}
-
-// cutAtArgCondition cuts s around the first ":" that begins an argument
-// condition — a ":" immediately followed by "key=" — returning the text before
-// and after it. If there is no such boundary, before is all of s and after is
-// empty. This tells a genuine ":key=value" boundary apart from a colon inside a
-// tool name or an argument value (a URL scheme, a Windows drive letter).
-func cutAtArgCondition(s string) (before, after string) {
-	for i := range len(s) {
-		if s[i] != ':' {
-			continue
+	for i, m := range matches {
+		valueEnd := len(pattern)
+		if i+1 < len(matches) {
+			valueEnd = matches[i+1][0]
 		}
-		if key, _, found := strings.Cut(s[i+1:], "="); found && isArgKey(key) {
-			return s[:i], s[i+1:]
-		}
+		argPatterns[pattern[m[2]:m[3]]] = pattern[m[1]:valueEnd]
 	}
-	return s, ""
-}
-
-// isArgKey reports whether s is a valid argument key: a non-empty run of
-// letters, digits or "_".
-func isArgKey(s string) bool {
-	for i := range len(s) {
-		if !isArgKeyByte(s[i]) {
-			return false
-		}
-	}
-	return s != ""
-}
-
-func isArgKeyByte(c byte) bool {
-	return c == '_' || ('a' <= c && c <= 'z') || ('A' <= c && c <= 'Z') || ('0' <= c && c <= '9')
+	return pattern[:matches[0][0]], argPatterns
 }
 
 // matchToolPattern checks if a tool name and its arguments match a pattern.
--- a/pkg/permissions/permissions.go
+++ b/pkg/permissions/permissions.go
@@ -166,30 +166,70 @@ func (c *Checker) DenyPatterns() []string {
 // Pattern format: "toolname" or "toolname:arg1=val1:arg2=val2"
 // Returns the tool pattern and a map of argument patterns.
 //
-// The parser looks for the first `:key=value` segment to split tool name from arguments.
-// This allows tool names with colons (like "mcp:github:create_issue") to work correctly.
+// The tool name runs up to the first ":key=" argument condition, so tool names
+// with colons (like "mcp:github:create_issue") are preserved. Argument values
+// may themselves contain colons (URLs like "https://…", Windows paths like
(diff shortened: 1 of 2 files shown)


pkg/permissions/permissions.go
```diff
@@ -166,32 +166,72 @@ func (c *Checker) DenyPatterns() []string {
...
+// with colons (like "mcp:github:create_issue") are preserved. Argument values
+// may themselves contain colons (URLs like "https://…", Windows paths like
+// "C:\…"): a value extends up to the next ":key=" boundary rather than the next
+// colon, so such values survive intact.
 func parsePattern(pattern string) (toolPattern string, argPatterns map[string]string) {
 	argPatterns = make(map[string]string)
 
-	// Find the first occurrence of :key=value pattern
-	// We look for ":" followed by an identifier and "="
-	parts := strings.Split(pattern, ":")
```
dgageot, line 176: You should be able to use string.Cut() here. That would simplify the code

dgageot (changes requested):

pkg/permissions/permissions.go
parveshsaini: Good call, done 👍 
strings.Cut does the boundary split better than the index arithmetic I had. looks much cleaner now

Thanks @dgageot

Sayt-0 (dismissed):
looks good to me !

dgageot (approved):

Sayt-0 (approved):

PR #3768, opened 2026-07-21
feat: config flavors — named YAML patches enabled at run time via --flavor

dgageot:
Agent configs are often nearly identical across environments or use-cases — the same tools, the same model, just a different system prompt or a disabled tool. Until now the only way to handle this was to maintain separate YAML files that drift apart over time, or to rely on external templating.

This PR introduces *flavors*: named YAML patches declared in a top-level `flavors` section of an agent config and applied at load time via a repeatable `--flavor` flag. Merge semantics follow JSON Merge Patch (RFC 7386) — objects merge recursively, scalars and arrays replace, `null` deletes a key — with two extensions: a `key+` suffix appends to an existing array, and a `key-` suffix removes entries (from an object by key name, from an array by scalar equality or partial object match). Document order is preserved using `yaml.MapSlice` so the root agent is always the first entry and never silently shifts. Flavors not declared in a config are silently ignored (a debug log is emitted) so the flag is safe to use across a fleet of configs where only some support a given flavor. The `--flavor` flag is wired into every command that loads an agent: `run`, `chat`, `eval`, `serve api/a2a/mcp`, `acp`, `debug`, and the sandbox/kit/agent-picker code paths. External OCI and URL sub-agents receive the same set of enabled flavors.

```yaml
flavors:
  production:
    system_prompt: "You are a helpful assistant. Never reveal internal tooling."
    tools-: [dangerous_tool]
  verbose:
    debug: true
```

```bash
docker-agent run agent.yaml --flavor production --flavor verbose
```

The feature lands on the latest schema only (v13); `agent-schema.json` is updated and HCL gains `flavors "name" {}` block syntax with a detection keyword. An annotated example lives at `examples/flavors.yaml` and a full docs page is added under `docs/configuration/flavors/`. Unit tests in `pkg/config/flavors_test.go` cover merge, order, delete, append, remove, and error paths including regression cases. `task build`, `task test`, and `task lint` all pass; two pre-existing example tests that require Docker Model Runner also fail on `main` and are unrelated to this change.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -120,6 +120,13 @@
       "additionalProperties": {
         "$ref": "#/definitions/BudgetConfig"
       }
+    },
+    "flavors": {
+      "type": "object",
+      "description": "Named YAML patches applied on top of the rest of the document when enabled at run time (e.g. 'docker agent run --flavor <name>'). Each flavor is a partial config document merged with JSON Merge Patch semantics: objects merge recursively, scalars and arrays replace, null deletes a key. A key ending in '+' appends its items to the existing array (e.g. 'toolsets+:'); a key ending in '-' removes entries — from an object by key name, from an array by scalar equality or partial object match (e.g. 'toolsets-: [{type: shell}]'). Enabled flavors are applied in the order they were requested; requested flavors not defined here are ignored.",
+      "additionalProperties": {
+        "type": ["object", "null"]
+      }
     }
   },
   "additionalProperties": false,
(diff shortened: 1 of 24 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The config-flavors implementation is well-engineered. The merge logic in `pkg/config/flavors.go` correctly follows JSON Merge Patch (RFC 7386) semantics with the `key+`/`key-` extensions, `slices.Clone` guards prevent aliasing across multi-flavor iterations, the `yaml.MapSlice` round-trip preserves document order, and all error paths return properly. No bugs were introduced by this PR.

Sayt-0 (approved):

PR #3770, opened 2026-07-21
Make standalone SSE usage in mcp tools configurable

krissetto:
Allow library users to configure if they want standalone sse enabled or not

Defaults to false to not introduce a potentially breaking change

### DIFF
--- a/pkg/tools/mcp/mcp.go
+++ b/pkg/tools/mcp/mcp.go
@@ -424,6 +424,18 @@ func (ts *Toolset) Restart(ctx context.Context) error {
 	return ts.supervisor.RestartAndWait(ctx, sessionMissingRetryTimeout)
 }
 
+// SetStandaloneSSE opts a remote streamable toolset in to the transport's
+// standalone SSE GET stream. Required for servers that send server-initiated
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly introduces a configurable `standaloneSSE` flag on `remoteMCPClient` and exposes it through `Toolset.SetStandaloneSSE`. The implementation is correct:

- `c.mu` is `sync.RWMutex` (embedded from `sessionClient`), so `c.mu.RLock()` / `c.mu.RUnlock()` in `Initialize` compiles correctly and is consistent with existing usage throughout `session_client.go`.
- The type assertion on `ts.mcpClient` in `Toolset.SetStandaloneSSE` without holding `ts.mu` is safe — `mcpClient` is set in the constructor and never reassigned, matching the same pattern already used in `Kind()` and `WorkingDir()`.
- `SetStandaloneSSE` uses a write lock (`c.mu.Lock()`) while `Initialize` uses a read lock to snapshot the field — correct reader/writer separation.
- Defaulting to `false` (standalone SSE disabled) preserves backward compatibility.

rumpl (approved):

PR #3771, opened 2026-07-21
refactor: cut heavy transitive deps (goja, expr, openai-go, go-git, sqlite) from the code-built embedder surface

dgageot:
Go developers who build agents and teams in code (rather than YAML) were pulling in 322 → 1,054 linked packages — a +37.6 MB binary — because `pkg/runtime`, `pkg/agent`, `pkg/session`, and the anthropic provider all hard-imported heavy runtimes (goja, expr, openai-go SDK, full go-git, modernc/sqlite) through their transitive dependency chains. None of that weight is needed just to wire up an agent.

This change cuts the code-built embedder surface from 1,054 down to 719 linked packages by surgically breaking each heavyweight import chain. DMR endpoint resolution and `/models` listing move to a new `pkg/model/provider/dmr/dmrmodels` leaf that uses only `net/http`, so `pkg/runtime` no longer links the openai-go SDK. The full `go-git` import in `pkg/fsx` is replaced by a lightweight `.git` stat/parse that preserves identical no-upward-search semantics, keeping only the `go-billy` and `plumbing/format/gitignore` plumbing. The file-backed SQLite session store moves to `pkg/session/sqlitestore` so callers that only need the in-memory store don't pay for `modernc/sqlite`. The goja JS command evaluator becomes pluggable via `pkg/runtime/jscommands.Register()`; YAML/CLI paths call it automatically through `teamloader`, `pkg/cli.Run`, `cmd/root`, and `embeddedchat/defaults`, so there is no behavior change there. Code-built embedders that need `${...}` expansion opt in with one call. Finally, `pkg/tools/mcp`'s interactive-prompt gating, OAuth sentinel errors, and the authorization-code flow are extracted into `pkg/tools` and a new `pkg/tools/mcp/oauthflow` leaf so neither `pkg/runtime` nor `pkg/agent` need to import the MCP toolset package (which drags in expr via `toolinstall` and goja via `pkg/js`).

The one intentional breaking change is `session.NewSQLiteSessionStore` → `sqlitestore.New`. All other public APIs are preserved via forwarders or aliases. An e2e dependency-budget test (`e2e/dependencies_test.go`) now asserts via `go list -deps` that goja, expr, openai-go, full go-git, sqlite, `pkg/js`, `pkg/tools/mcp`, and `toolinstall` remain absent from the embedder surface, so the budget cannot silently regress.

One item from the original feedback — defining `tools.ToolAnnotations` locally — was not done because the type is also referenced by the runtime's elicitation/sampling handler signatures and `pkg/tools`' own handler types; removing the MCP go-sdk import there would require breaking those public handler types, which is out of scope here.


pkg/runtime/remote_runtime.go
```diff
@@ -658,19 +658,19 @@ func (r *RemoteRuntime) UpdateSessionTitle(ctx context.Context, sess *session.Se
...
-		return make(map[string]mcp.PromptInfo)
+		return make(map[string]tools.PromptInfo)
 	}
-	// Convert map[string]any to map[string]mcp.PromptInfo
-	result := make(map[string]mcp.PromptInfo)
+	// Convert map[string]any to map[string]tools.PromptInfo
+	result := make(map[string]tools.PromptInfo)
 	for k, v := range prompts {
-		if info, ok := v.(mcp.PromptInfo); ok {
+		if info, ok := v.(tools.PromptInfo); ok {
```
docker-agent: **[high] `v.(tools.PromptInfo)` always fails — `CurrentMCPPrompts` always returns empty**

`GetSessionMCPPrompts` deserialises the server response via `json.Unmarshal` into a `map[string]any`. Go's JSON decoder never produces concrete struct types; every JSON object becomes a `map[string]interface{}`. The type assertion `v.(tools.PromptInfo)` therefore always has `ok == false`, so `result` is always returned empty, silently discarding every MCP prompt for `RemoteRuntime` users.

Fix: round-trip each value through the JSON marshaller so the nested map is decoded into the target struct:

```suggestion
	// Convert map[string]any to map[string]tools.PromptInfo via JSON round-trip.
	result := make(map[string]tools.PromptInfo)
	for k, v := range prompts {
		b, err := json.Marshal(v)
		if err != nil {
			continue
		}
		var info tools.PromptInfo
		if err := json.Unmarshal(b, &info); err == nil {
			result[k] = info
		}
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/session/sqlitestore/sqlitestore.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+		return nil, err
+	}
+
+	return store, nil
+}
+
+// backupDatabase moves the database file (and related WAL files) to a backup
+func backupDatabase(path string) error {
+	backupPath := path + ".bak"
```
docker-agent, line 94: **[medium] `backupDatabase` silently overwrites a prior `.bak`, destroying the user's previous backup**

`os.Rename(path, backupPath)` (POSIX) atomically replaces the destination if it already exists. If the user previously experienced a migration failure (so `.bak` already exists from an earlier run), and the freshly created database also fails migration on restart, the existing `.bak` is overwritten with the new (also corrupt) file. The user permanently loses the last known-good backup.

Add an existence check before renaming — if `.bak` is already present, either skip the backup (the existing one is likely more useful) or name it with a timestamp:

```suggestion
	backupPath := path + ".bak"
	if _, statErr := os.Stat(backupPath); statErr == nil {
		// A prior backup already exists — preserve it by using a timestamped name.
		backupPath = fmt.Sprintf("%s.bak.%d", path, time.Now().Unix())
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/tools/mcp/oauthflow/oauthflow.go
```diff
@@ -0,0 +1,401 @@
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to exchange code for token: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("token exchange failed with status %d: %s", resp.StatusCode, string(body))
```
docker-agent, line 172: **[low] Raw token endpoint error body returned verbatim in error — may expose sensitive data**

`io.ReadAll(resp.Body)` on a non-200 response is included verbatim in the returned error:

```go
return nil, fmt.Errorf("token exchange failed with status %d: %s", resp.StatusCode, string(body))
```

The same pattern appears for client registration (`RegisterClientWithClient`) and token refresh (`RefreshAccessTokenWithClient`). OAuth error responses can include `error_description` fields that carry PII or partial credential hints from non-standard servers. Errors propagated to callers often end up in log output.

Consider truncating the body or extracting only the structured `error`/`error_description` fields:

```go
body, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
return nil, fmt.Errorf("token exchange failed with status %d: %s", resp.StatusCode, string(body))
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

pkg/tools/mcp/oauthflow/oauthflow.go
```diff
@@ -0,0 +1,401 @@
...
+// network).
+func HTTPClientForAllowPrivateIPs(allowPrivateIPs bool) *http.Client {
+	if allowPrivateIPs {
+		// Clone keeps the default proxy/HTTP2/timeout behavior but gives the
+		// client its own connection pool: a nil Transport would share
+		// http.DefaultTransport's pool, which third parties may prune via
+		// CloseIdleConnections (httptest.Server.Close does).
+		return &http.Client{
+			Timeout:   30 * time.Second,
+			Transport: http.DefaultTransport.(*http.Transport).Clone(),
```
docker-agent: **[low] Unchecked type assertion on `http.DefaultTransport` — panics if transport is replaced**

```go
Transport: http.DefaultTransport.(*http.Transport).Clone(),
```

If `http.DefaultTransport` has been replaced with a non-`*http.Transport` (e.g. a test helper, an HTTP/3 transport, or a corporate proxy shim), this panics at runtime. Use the comma-ok form to handle the case gracefully:

```suggestion
		transport, ok := http.DefaultTransport.(*http.Transport)
		if !ok {
			transport = &http.Transport{}
		}
		return &http.Client{
			Timeout:   30 * time.Second,
			Transport: transport.Clone(),
		}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One confirmed high-severity bug (MCP prompts always empty for remote runtime) plus two confirmed/likely medium/low findings.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a well-structured dependency-reduction refactor. All seven hypotheses raised during review were dismissed after verification:

- **`pkg/fsx/vcs.go` gitfile parsing**: The new lightweight `.git` stat/parse correctly handles Windows CRLF line endings via `strings.TrimSpace`, and the no-upward-search semantics are preserved and tested.
- **`pkg/session/sqlitestore` db lifecycle**: The `open()` helper correctly closes `db` exactly once on failure; no leak or double-close.
- **`commandEvaluatorFactory` test**: Standard Go test pattern for swapping package-global atomic state; not parallel, so no data race.
- **`ListModelsAt` context use after timeout**: The `slog.DebugContext` call runs before `defer cancel()` fires; context is valid. Even a cancelled context doesn't suppress `slog` output.
- **`HTTPClientForAllowPrivateIPs` type assertion**: Pre-existing pattern moved verbatim; not introduced by this PR.
- **Cache TOCTOU in `NewVCSMatcher`**: The TOCTOU window between read and write lock yields only harmless idempotent writes.
- **`SetHTTPClientForTesting` restore ignored in `TestMain`**: Intentional — the process exits immediately via `os.Exit`, so no restore is needed.

The dependency-budget e2e test (`e2e/dependencies_test.go`) that locks the embedder surface against regression is a solid safeguard for the stated goals.

pkg/runtime/remote_runtime.go
dgageot: Fixed in cdf04aa1a — values are now converted via a JSON round-trip, with a regression test (`TestRemoteRuntime_CurrentMCPPrompts`) pinning the wire shape. Note the bug predates this PR (the old `v.(mcp.PromptInfo)` assertion failed identically since `PromptInfo` is a type alias), but it was worth fixing here.

pkg/session/sqlitestore/sqlitestore.go
dgageot: Fixed in 7fd72d74a — when `.bak` already exists, the broken database is moved aside under a timestamped `.bak.<unix>` name so the last known-good backup is preserved. Covered by a new subtest in `TestBackupDatabase`.

pkg/tools/mcp/oauthflow/oauthflow.go
dgageot: Fixed in b83bd3fbb — all three error paths (token exchange, registration, refresh) now read at most 512 bytes of the response body via `io.LimitReader` before embedding it in the error.

pkg/tools/mcp/oauthflow/oauthflow.go
dgageot: Fixed in b83bd3fbb — comma-ok assertion with a fallback to a fresh `&http.Transport{}` when `http.DefaultTransport` has been replaced.

docker-agent: Thanks for the quick fix! The comma-ok assertion with the fallback to `&http.Transport{}` handles the edge case cleanly — the code now degrades gracefully if `http.DefaultTransport` has been swapped out in tests or unusual runtime environments.

trungutt (approved):

aheritier (approved):

PR #3772, opened 2026-07-21
feat(anthropic): fail fast when thinking_display: display is set on adaptive-thinking models

dgageot:
Anthropic's adaptive-thinking models (Claude Opus/Sonnet 4.6+, Sonnet 5, Fable 5, Mythos 5) only accept `thinking_display` values of `"summarized"` or `"omitted"`. Sending `"display"` causes an HTTP 400 at inference time. Previously docker-agent forwarded the value as-is and the error surfaced only on the first request, with no indication of where to fix it.

This change adds a `validateThinkingDisplay` function in the Anthropic provider that is called during client construction — both in `NewClient` and `NewVertexClient`. A misconfigured `thinking_display: display` on an adaptive-thinking model now fails at parse time with a clear error message rather than silently at request time. The predicate `modelinfo.SupportsFullThinkingDisplay` encodes the model boundary, delegating to `!SupportsAdaptiveThinking` so the two constraints stay in sync. Server-side fallback models (`provider_opts.fallbacks`) go through the same validation, since they receive the identical request shape. Invalid non-enum values keep their existing warn-and-ignore behavior; no coercion is added.

The schema and docs are updated to reflect that `thinking_display` is scoped to the Anthropic provider. `examples/thinking_budget.yaml` is corrected to use a valid value.


trungutt (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

The validation logic is correct and well-structured. `validateThinkingDisplay` is called at client construction in both `NewClient` and `NewVertexClient`, the `SupportsFullThinkingDisplay` predicate correctly delegates to `!SupportsAdaptiveThinking` (keeping the two constraints in sync), and fallback models are validated using the same `append([]string{cfg.Model}, fallbacks...)` idiom used elsewhere. No issues found.

PR #3773, opened 2026-07-21
fix(desktop): force token refresh when Docker Desktop returns an expired JWT

dgageot:
Docker Desktop's auth v2 stack (behind the `SIGN_IN_V2` feature flag) never refreshes the JWT it holds on a `GET /registry/token` request — it returns whatever its in-memory token source currently contains. When Desktop's background refresher stalls on a permanent 4xx (e.g. a PAT/hub token error), it keeps serving the same expired JWT indefinitely, and retrying the endpoint from docker-agent's side never helps.

`desktop.GetToken` now parses the JWT `exp` claim (with 30s clock-skew leeway). When the token is expired, it forces a Desktop-side refresh by `POST`ing `/registry/credstore-updated` — the same hook the Docker CLI triggers after `docker login`, which makes Desktop reload its session from the OS credential store and issue a fresh token — then polls `GET /registry/token` until a non-expired token appears. A single detached goroutine performs the refresh under a hard 10s budget; concurrent callers share its result via singleflight and can bail on their own context cancellation, falling back to the stale token (preserving the previous behavior). After a successful refresh a 30s cooldown applies; after failure, a 2-minute backoff. Non-JWT tokens, empty tokens, and tokens without an `exp` claim pass through unchanged. `raw_client.go` gains a `Post` method to support the new endpoint.

Tests use an in-memory fake Desktop backend over `net.Pipe` and cover refresh success and failure, backoff, cached-result reuse, 8-way concurrent refresh sharing, caller cancellation, and non-JWT/empty token pass-through.

### DIFF
--- a/pkg/desktop/raw_client.go
+++ b/pkg/desktop/raw_client.go
@@ -3,6 +3,7 @@ package desktop
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -56,6 +57,28 @@ func (c *RawClient) Get(ctx context.Context, endpoint string, v any) error {
 
 	if err := json.Unmarshal(buf, &v); err != nil {
 		return err
+	}
+	return nil
+}
+
+func (c *RawClient) Post(ctx context.Context, endpoint string) error {
+	ctx, cancel := context.WithTimeout(ctx, c.timeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, "http://localhost"+endpoint, http.NoBody)
+	if err != nil {
+		return err
+	}
+
+	response, err := c.client().Do(req)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+	_, _ = io.Copy(io.Discard, response.Body)
+
+	if response.StatusCode >= http.StatusBadRequest {
+		return fmt.Errorf("POST %s: %s", endpoint, response.Status)
 	}
 	return nil
 }
--- a/pkg/desktop/login.go
+++ b/pkg/desktop/login.go
@@ -117,6 +117,7 @@ func forceTokenRefresh(ctx context.Context) string {
 		token := runTokenRefresh(ctx)
 
 		refreshState.Lock()
+		defer refreshState.Unlock()
 		refreshState.result = token
 		backoff := refreshCooldown
 		if token == "" {
@@ -124,7 +125,6 @@ func forceTokenRefresh(ctx context.Context) string {
 		}
 		refreshState.nextAttempt = time.Now().Add(backoff)
 		refreshState.inflight = nil
-		refreshState.Unlock()
 		close(done)
 	}()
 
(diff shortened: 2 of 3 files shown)


docker-agent (commented):

**1 medium-severity finding** in the new forced-refresh path.

- [low] pkg/desktop/login.go:120 — goroutine closes `done` channel after releasing lock (brief window where new caller observes inflight==nil before channel is closed) (confidence: weak 43/100)
- [low] pkg/desktop/login.go:150 — poll loop fetches token immediately after POST before async Desktop refresh could occur (minor inefficiency) (confidence: weak 30/100)
- [low] pkg/desktop/login.go:52 — expiry leeway is applied as a 30s grace period after expiry rather than before — may serve tokens expired up to 30s ago (confidence: weak 30/100)
- [low] pkg/desktop/login_test.go:205 — cleanup ordering may expose detached goroutine to restored global values (potential flakiness) (confidence: weak 30/100)

pkg/desktop/raw_client.go
```diff
@@ -59,3 +60,25 @@ func (c *RawClient) Get(ctx context.Context, endpoint string, v any) error {
 	}
 	return nil
 }
+
+func (c *RawClient) Post(ctx context.Context, endpoint string) error {
+	ctx, cancel := context.WithTimeout(ctx, c.timeout)
```
docker-agent, line 65: **[medium] `RawClient.Post` adds a redundant 10s timeout that can consume the entire `refreshBudget`, leaving zero time for the polling loop**

`runTokenRefresh` is called with a context already constrained by `refreshBudget` (10 s, set in the detached goroutine at line 114 of `login.go`). `RawClient.Post` then calls `context.WithTimeout(ctx, c.timeout)` where `c.timeout` is also 10 s — but `context.WithTimeout` uses `min(parent_deadline_remaining, new_duration)`, so the effective POST timeout is the full 10 s budget. If the POST to `/registry/credstore-updated` takes the complete 10 s (e.g., Desktop is under load), the context is already past its deadline when the polling loop begins, and the loop immediately exits via `ctx.Done()` — returning `""` even though the refresh was successfully triggered. The polling loop is specifically there to wait for Desktop's async refresh, so burning the entire budget in the POST silently defeats the whole mechanism.

A simple fix is to give `Post` a shorter dedicated timeout so the polling loop always has a meaningful window:

```suggestion
func (c *RawClient) Post(ctx context.Context, endpoint string) error {
	postTimeout := c.timeout / 3
	ctx, cancel := context.WithTimeout(ctx, postTimeout)
```

Alternatively, the `Post` method could honour the caller's deadline as-is (drop the `WithTimeout` call entirely) and let `runTokenRefresh` split the budget explicitly between the POST and the poll phase.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

dgageot: Good catch — fixed in 8e564b6a7. Kept `RawClient.Post` generic (its 10s cap is right for other callers) and instead capped the nudge in the refresh path: a new `postRefreshNudge` helper limits the POST to `refreshBudget/3`, so the polling loop always keeps a meaningful window. Also moved `close(done)` under the lock to remove the brief `inflight==nil`-before-close window noted in the low-confidence findings. The other low-confidence items are intentional: the immediate first poll handles Desktop refreshing synchronously (cheap GET on a local socket), the leeway direction avoids refresh loops when the local clock runs ahead, and the test cleanup drains the detached goroutine first thanks to LIFO `t.Cleanup` ordering.

docker-agent: Perfect solution — creating the dedicated postRefreshNudge helper is much cleaner than modifying RawClient.Post, which as you point out serves other callers that genuinely need the full 10s cap. Wrapping the nudge with refreshBudget/3 guarantees the polling loop always has a meaningful window regardless of Desktop load, while keeping the generic client method unchanged. The close(done) fix under the lock is a nice bonus.

Thanks for the detailed rationale on the other items — all make sense given the local-socket latency and the clock-skew scenario.

trungutt (approved):

PR #3777, opened 2026-07-21
feat: add X-Cagent-Compacting header for session-compaction LLM calls

dgageot:
Session-compaction (summary) LLM calls are currently indistinguishable from ordinary chat completions at the AI gateway layer. This means gateway-side policies — rate limiting, routing, cost attribution — cannot treat them differently from user-facing conversation turns. Adding an explicit marker gives the gateway a reliable signal it can act on, mirroring the already-established `X-Cagent-GeneratingTitle` pattern.

The change introduces an `X-Cagent-Compacting: 1` HTTP header on gateway-bound requests that originate from the compactor. A new `WithCompacting()` option and `Compacting()` getter are added to `pkg/model/provider/options`, and `FromModelOptions` ensures the flag survives provider cloning. `GatewayHTTPOptions` in the base gateway is refactored to accept `*options.ModelOptions` instead of a bare `generatingTitle bool`, which lets it emit both the title and compacting headers from a single options struct and treat a nil pointer as zero options. The anthropic, openai, and gemini gateway clients are updated accordingly, and the compactor clones its summary model with `options.WithCompacting()`. The DMR client's skip-reconfiguration guard is extended to cover compaction clones, which also override `max_tokens`.

No breaking changes. The new header is purely additive; gateways that do not inspect it are unaffected.

### DIFF
--- a/pkg/model/provider/anthropic/client.go
+++ b/pkg/model/provider/anthropic/client.go
@@ -115,7 +115,7 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 			baseURL := fmt.Sprintf("%s://%s%s/", url.Scheme, url.Host, url.Path)
 
 			// Configure a custom HTTP client to inject headers and query params used by the Gateway.
-			httpOptions := base.GatewayHTTPOptions(url, "https://api.anthropic.com/", cfg, globalOptions.GeneratingTitle())
+			httpOptions := base.GatewayHTTPOptions(url, "https://api.anthropic.com/", cfg, &globalOptions)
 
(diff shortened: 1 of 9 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is clean and correct. The `X-Cagent-Compacting` header follows the established `X-Cagent-GeneratingTitle` pattern faithfully across all five touch-points: the new `WithCompacting()`/`Compacting()` option pair, `FromModelOptions` propagation, the refactored `GatewayHTTPOptions` (nil-safe, emits both headers), the DMR skip-reconfiguration guard extension, and the compactor clone site. No bugs found in the changed code.

aheritier (approved):

PR #3778, opened 2026-07-21
refactor: trim transitive dependencies for library consumers

dgageot:
Downstream Go projects importing cagent packages (`pkg/session`, `pkg/agent`, `pkg/model/provider`, `pkg/model/provider/anthropic`) were pulling heavy transitive dependencies they didn't need — the full OCI registry stack and the entire Google Cloud auth stack — simply because those packages imported `pkg/remote` and `pkg/model/provider/anthropic` unconditionally.

The first commit moves the Docker Desktop proxy HTTP transport from `pkg/remote` into a new leaf package `pkg/desktop/transport`. The new `transport.New` constructor carries no OCI registry baggage; `remote.NewTransport` is kept as a deprecated forwarder so existing callers keep compiling without changes. Internal packages (`pkg/httpclient`, `pkg/modelsdev`, `pkg/gateway`, `pkg/remote`) are updated to import the new location. The second commit moves Vertex AI support out of `pkg/model/provider/anthropic` into `pkg/model/provider/anthropic/vertex`, wired through a new `anthropic.NewClientFromFactory` extension point. Config validation still runs before GCP credential discovery, credentials are still resolved eagerly, and the default providers registry retains full Vertex support through the `vertexai` provider — YAML configs are unaffected. Direct callers of the removed `anthropic.NewVertexClient` get a compile-time error rather than a silent behavior change.

| Package | Before | After |
|---|---|---|
| `pkg/model/provider` | 31 modules | 24 modules |
| `pkg/model/provider/anthropic` | 57 modules | 35 modules |

`task build`, `task test`, and `task lint` all pass. The `pkg/teamloader` example tests require a local Docker Model Runner and fail identically on `main` — pre-existing, unrelated.


pkg/desktop/transport/transport.go
```diff
@@ -0,0 +1,170 @@
...
+	// Retry direct only when the body is safe to replay; otherwise the
+	// proxy may have already consumed it.
+	if req.Body != nil && req.GetBody == nil {
+		return nil, err
+	}
+	retryReq := req.Clone(req.Context())
+	if req.GetBody != nil {
+		body, bodyErr := req.GetBody()
+		if bodyErr != nil {
+			return nil, err
```
docker-agent, line 135: **[medium] Wrong error returned when `GetBody` fails — proxy error masks body-reconstruction failure**

When the proxy is unavailable and `RoundTrip` falls back to a direct retry, it calls `req.GetBody()` to reconstruct the request body. If `GetBody()` itself returns an error (`bodyErr`), the code discards that error and returns the original proxy socket error (`err`) instead:

```go
body, bodyErr := req.GetBody()
if bodyErr != nil {
    return nil, err   // bodyErr is silently dropped
}
```

The caller receives a misleading "proxy unavailable" or "dial unix …" error when the real failure is body reconstruction. Although `GetBody` closures rarely fail in practice (they typically wrap an already-buffered body), the error contract is still broken — if they do fail the wrong error is surfaced, making the root cause invisible.

```suggestion
			return nil, bodyErr
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

aheritier (approved):

PR #3782, opened 2026-07-22
feat(server): resolve agent-switch slash commands in RunSession

trungutt:
Host applications driving the runtime over HTTP had no way to trigger a mid-session agent switch: only the interactive CLI/TUI runners resolved slash commands (via `PrepareUserMessage`) before appending them to the session. Anyone else — REST clients, Electron hosts, custom orchestrators — could only invoke the target agent via the `POST /api/sessions/:id/agent/:agent/:agent_name` two-segment route, whose semantics only actually take effect on the first turn of a session (the cached runtime silently ignores it on turn 2+).

This PR closes that gap on the HTTP side. `SessionManager.RunSession` now runs each incoming user message through `LookupCommand` / `ResolveCommand` before appending, mirroring what the CLI already does:

```
POST /api/sessions/:id/agent/root
body: {"messages":[{"role":"user","content":"/plan design a login flow"}]}
         │
         ▼
    RunSession
         │
         ▼
    applyAgentSwitchCommands
         │
         ├─ LookupCommand("/plan design a login flow")  →  {Agent: "planner"}, rest: "design a login flow"
         ├─ current agent is "root", target is "planner" ➜ SetCurrentAgent("planner")
         └─ replace message content with "design a login flow"
         │
         ▼
    session stores clean history:
      user: "design a login flow"       ◄── no routing prefix leaks in
    runtime runs turn as `planner`
```

**Scope is intentionally narrow: only commands whose `Agent` field is set are resolved here.** Instruction-expansion commands (those declaring an `Instruction`) pass through unchanged, so an existing caller that happens to send text starting with `/` is not silently reinterpreted server-side. The blast radius of the change is limited to configs that already declared `commands.<x>.agent: <name>`, which today are the tightly-controlled agent-switching commands.

Idempotence is a design choice: when the target agent is already current, `SetCurrentAgent` is skipped. That lets host applications re-assert the desired agent on every turn (e.g. by always prefixing the outgoing message with `/plan` or `/build` depending on a UI toggle) without paying for a no-op switch or tracking per-session applied state client-side.

### DIFF
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -1534,6 +1534,7 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
 // expanding an instruction-only command doesn't silently rewrite text
 // existing HTTP callers send as literal user input.
 func (sm *SessionManager) applyAgentSwitchCommands(ctx context.Context, rt runtime.Runtime, messages []api.Message) error {
+	originalAgent := rt.CurrentAgentName(ctx)
 	for i := range messages {
 		if messages[i].Role != chat.MessageRoleUser {
 			continue
@@ -1542,12 +1543,20 @@ func (sm *SessionManager) applyAgentSwitchCommands(ctx context.Context, rt runti
 		if !ok || cmd.Agent == "" {
 			continue
 		}
+		// Resolve against the agent that owns the command definition:
+		// the target sub-agent typically doesn't re-declare the routing
+		// command in its own table.
+		resolved := runtime.ResolveCommand(ctx, rt, messages[i].Content)
 		if cmd.Agent != rt.CurrentAgentName(ctx) {
 			if err := rt.SetCurrentAgent(ctx, cmd.Agent); err != nil {
+				// Undo any earlier successful switch in this batch so
+				// the session isn't stuck on a mid-batch agent after
+				// the caller aborts the turn.
+				_ = rt.SetCurrentAgent(ctx, originalAgent)
 				return fmt.Errorf("switch agent to %q: %w", cmd.Agent, err)
 			}
 		}
-		messages[i].Content = runtime.ResolveCommand(ctx, rt, messages[i].Content)
+		messages[i].Content = resolved
 	}
 	return nil
 }
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -894,6 +894,13 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 		return nil, err
 	}
 
+	if err := sm.applyAgentSwitchCommands(ctx, runtimeSession.runtime, messages); err != nil {
+		undoModelOverride(ctx, prevOverride, hadPrevOverride)
+		runtimeSession.streaming.Unlock()
+		cancel()
+		return nil, err
+	}
+
 	// Now that we hold the streaming lock, it is safe to mutate the session.
 	// Collect user messages for potential title generation
 	var userMessages []string
@@ -1520,6 +1527,29 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
 		}
 	}
 	return prevOverride, hadPrev, undo, nil
+}
+
+// applyAgentSwitchCommands is the HTTP analogue of
+// pkg/cli/runner.PrepareUserMessage. Scoped to agent-switch commands so
+// expanding an instruction-only command doesn't silently rewrite text
+// existing HTTP callers send as literal user input.
+func (sm *SessionManager) applyAgentSwitchCommands(ctx context.Context, rt runtime.Runtime, messages []api.Message) error {
+	for i := range messages {
+		if messages[i].Role != chat.MessageRoleUser {
+			continue
+		}
+		cmd, _, ok := runtime.LookupCommand(ctx, rt, messages[i].Content)
+		if !ok || cmd.Agent == "" {
+			continue
+		}
+		if cmd.Agent != rt.CurrentAgentName(ctx) {
+			if err := rt.SetCurrentAgent(ctx, cmd.Agent); err != nil {
+				return fmt.Errorf("switch agent to %q: %w", cmd.Agent, err)
+			}
+		}
+		messages[i].Content = runtime.ResolveCommand(ctx, rt, messages[i].Content)
+	}
+	return nil
 }
 
 // applyStoredOverrides applies the persisted per-agent model overrides on
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -1533,8 +1533,20 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
 // pkg/cli/runner.PrepareUserMessage. Scoped to agent-switch commands so
 // expanding an instruction-only command doesn't silently rewrite text
 // existing HTTP callers send as literal user input.
+//
+// Two-pass: resolve every message against the pre-batch agent first,
+// then apply switches. This keeps each message's interpretation
+// consistent regardless of how earlier messages in the same batch might
+// have mutated the runtime.
 func (sm *SessionManager) applyAgentSwitchCommands(ctx context.Context, rt runtime.Runtime, messages []api.Message) error {
 	originalAgent := rt.CurrentAgentName(ctx)
+
+	type pending struct {
+		idx     int
+		target  string
+		content string
+	}
+	var switches []pending
 	for i := range messages {
 		if messages[i].Role != chat.MessageRoleUser {
 			continue
@@ -1543,20 +1555,24 @@ func (sm *SessionManager) applyAgentSwitchCommands(ctx context.Context, rt runti
 		if !ok || cmd.Agent == "" {
 			continue
 		}
-		// Resolve against the agent that owns the command definition:
-		// the target sub-agent typically doesn't re-declare the routing
-		// command in its own table.
-		resolved := runtime.ResolveCommand(ctx, rt, messages[i].Content)
-		if cmd.Agent != rt.CurrentAgentName(ctx) {
-			if err := rt.SetCurrentAgent(ctx, cmd.Agent); err != nil {
-				// Undo any earlier successful switch in this batch so
-				// the session isn't stuck on a mid-batch agent after
-				// the caller aborts the turn.
-				_ = rt.SetCurrentAgent(ctx, originalAgent)
-				return fmt.Errorf("switch agent to %q: %w", cmd.Agent, err)
+		switches = append(switches, pending{
+			idx:     i,
+			target:  cmd.Agent,
+			content: runtime.ResolveCommand(ctx, rt, messages[i].Content),
+		})
+	}
+
+	for _, s := range switches {
+		if s.target != rt.CurrentAgentName(ctx) {
+			if err := rt.SetCurrentAgent(ctx, s.target); err != nil {
+				if rbErr := rt.SetCurrentAgent(ctx, originalAgent); rbErr != nil {
+					slog.WarnContext(ctx, "failed to restore agent after switch error; session may be on wrong agent",
+						"original_agent", originalAgent, "stuck_on", s.target, "err", rbErr)
+				}
+				return fmt.Errorf("switch agent to %q: %w", s.target, err)
 			}
 		}
-		messages[i].Content = resolved
+		messages[s.idx].Content = s.content
 	}
 	return nil
 }
--- a/pkg/server/session_manager_test.go
+++ b/pkg/server/session_manager_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"path/filepath"
@@ -23,6 +24,7 @@ import (
 	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/concurrent"
 	"github.com/docker/docker-agent/pkg/config"
+	"github.com/docker/docker-agent/pkg/config/types"
 	"github.com/docker/docker-agent/pkg/runtime"
 	"github.com/docker/docker-agent/pkg/session"
 	"github.com/docker/docker-agent/pkg/session/sqlitestore"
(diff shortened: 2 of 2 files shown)


pkg/server/session_manager.go
```diff
@@ -1522,6 +1529,40 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
...
+			}
+		} else {
+			slog.DebugContext(ctx, "applyAgentSwitchCommands: target agent already current, skipping switch", "agent", current, "message_index", i)
+		}
+		resolved := runtime.ResolveCommand(ctx, rt, messages[i].Content)
+		slog.DebugContext(ctx, "applyAgentSwitchCommands: stripped routing prefix", "original_len", len(messages[i].Content), "resolved_len", len(resolved), "message_index", i)
+		messages[i].Content = resolved
+	}
+	return nil
+}
```
docker-agent, line 1578: **[medium] Partial agent switch not rolled back on error in multi-message batch**

When a batch contains multiple user messages with agent-switch commands, `applyAgentSwitchCommands` processes them sequentially. If message `i`'s `SetCurrentAgent` succeeds and message `i+1`'s `SetCurrentAgent` then fails, the function returns an error — but the runtime's current agent has already been permanently changed to the agent selected at step `i`. The caller in `RunSession` only calls `undoModelOverride` and aborts the turn; there is no corresponding rollback for the agent switch. Future requests to this session will arrive on the wrong agent even though the current turn was rejected.

A simple fix is to capture the agent name before the loop and restore it on error:

```suggestion
func (sm *SessionManager) applyAgentSwitchCommands(ctx context.Context, rt runtime.Runtime, messages []api.Message) error {
	originalAgent := rt.CurrentAgentName(ctx)
	for i := range messages {
		if messages[i].Role != chat.MessageRoleUser {
			continue
		}
		cmd, _, ok := runtime.LookupCommand(ctx, rt, messages[i].Content)
		if !ok || cmd.Agent == "" {
			continue
		}
		current := rt.CurrentAgentName(ctx)
		if cmd.Agent != current {
			slog.DebugContext(ctx, "applyAgentSwitchCommands: switching agent", "from", current, "to", cmd.Agent, "message_index", i)
			if err := rt.SetCurrentAgent(ctx, cmd.Agent); err != nil {
				_ = rt.SetCurrentAgent(ctx, originalAgent) // best-effort rollback
				return fmt.Errorf("switch agent to %q: %w", cmd.Agent, err)
			}
		} else {
			slog.DebugContext(ctx, "applyAgentSwitchCommands: target agent already current, skipping switch", "agent", current, "message_index", i)
		}
		resolved := runtime.ResolveCommand(ctx, rt, messages[i].Content)
		slog.DebugContext(ctx, "applyAgentSwitchCommands: stripped routing prefix", "original_len", len(messages[i].Content), "resolved_len", len(resolved), "message_index", i)
		messages[i].Content = resolved
	}
	return nil
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

Two medium-severity bugs were found in the newly introduced `applyAgentSwitchCommands` function:

1. **Partial agent switch not rolled back on error** — in a multi-message batch, a failed `SetCurrentAgent` for a later message leaves the runtime pointing at the agent selected by an earlier successful switch, even though the turn is aborted.
2. **`ResolveCommand` re-runs `LookupCommand` on the wrong (post-switch) agent** — the routing slash command is typically not defined on the target sub-agent's command table, so `ResolveCommand` falls back to returning the raw `/command args` prefix verbatim, leaking the routing artifact into session history. This is masked by the test mock's single shared command map.

pkg/server/session_manager.go
```diff
@@ -1522,6 +1529,40 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
...
+		current := rt.CurrentAgentName(ctx)
+		if cmd.Agent != current {
+			slog.DebugContext(ctx, "applyAgentSwitchCommands: switching agent", "from", current, "to", cmd.Agent, "message_index", i)
+			if err := rt.SetCurrentAgent(ctx, cmd.Agent); err != nil {
+				return fmt.Errorf("switch agent to %q: %w", cmd.Agent, err)
+			}
+		} else {
+			slog.DebugContext(ctx, "applyAgentSwitchCommands: target agent already current, skipping switch", "agent", current, "message_index", i)
+		}
+		resolved := runtime.ResolveCommand(ctx, rt, messages[i].Content)
```
docker-agent: **[medium] `ResolveCommand` uses the post-switch agent context and may silently fail to strip the slash-command prefix**

After `SetCurrentAgent` is called (line 1553), the runtime's active agent has changed. The subsequent `runtime.ResolveCommand(ctx, rt, messages[i].Content)` call (line 1559) internally calls `LookupCommand`, which reads `rt.CurrentAgentInfo(ctx).Commands` — **the new agent's command table**, not the original one. In a real deployment, agent command tables are per-agent. A routing command like `/plan` is typically defined only on the *parent/root* agent to route to `planner`; the `planner` agent itself does not re-declare `/plan` in its own command table.

Consequence: after switching to `planner`, `LookupCommand` cannot find `/plan` in the planner's commands → `ok` is false → `ResolveCommand` returns `userInput` verbatim (the full `/plan explain X` string). The message appended to the session and forwarded to the sub-agent still contains the raw slash command prefix, leaking the routing artifact into the conversation history.

The existing unit tests do **not** catch this because `commandFakeRuntime` uses a single shared `commands` map for all agent states; after `SetCurrentAgent("planner")`, `CurrentAgentInfo` still returns the same map (containing `"plan": {Agent: "planner"}`), masking the real-runtime behavior.

**Fix**: resolve the command before switching the agent (while the original agent's command table is still active):

```suggestion
		cmd, rest, ok := runtime.LookupCommand(ctx, rt, messages[i].Content)
		if !ok || cmd.Agent == "" {
			continue
		}
		current := rt.CurrentAgentName(ctx)
		if cmd.Agent != current {
			slog.DebugContext(ctx, "applyAgentSwitchCommands: switching agent", "from", current, "to", cmd.Agent, "message_index", i)
			if err := rt.SetCurrentAgent(ctx, cmd.Agent); err != nil {
				return fmt.Errorf("switch agent to %q: %w", cmd.Agent, err)
			}
		} else {
			slog.DebugContext(ctx, "applyAgentSwitchCommands: target agent already current, skipping switch", "agent", current, "message_index", i)
		}
		// Use `rest` captured above (pre-switch) instead of re-resolving after the switch.
		// ResolveCommand internally re-runs LookupCommand on the now-switched agent, which
		// may not define the routing command, causing it to return the full /cmd prefix verbatim.
		messages[i].Content = rest
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/server/session_manager.go
trungutt: Fixed in 8aab8dc6a. Captured `originalAgent` at the top of the function; on any `SetCurrentAgent` failure we best-effort restore it, undoing any earlier switch in the batch. Added `TestApplyAgentSwitchCommands_RollsBackOnBatchFailure` — the fake `SetCurrentAgent` now supports per-target-agent errors so we can drive the "msg[0] switches OK, msg[1] fails" scenario deterministically.

pkg/server/session_manager.go
trungutt: Fixed in 8aab8dc6a. Reordered the loop so `ResolveCommand` runs against the pre-switch agent (matches the ordering `pkg/cli/runner.PrepareUserMessage` uses). Kept the `ResolveCommand` call rather than reusing `rest` from the initial `LookupCommand` so commands that combine `agent:` + `instruction:` still get their instruction expanded (against the source agent, correctly). Added `TestApplyAgentSwitchCommands_ResolvesUsingPreSwitchAgent` — the fake now supports per-agent command tables so the destination-has-no-routing-command case is exercised deterministically. Good catch on the shared-map masking, that was the root of why my original tests passed.

pkg/server/session_manager.go
docker-agent: Perfect! The fix correctly captures `originalAgent` before the loop and restores it on any `SetCurrentAgent` failure, preventing the session from being stuck mid-batch. The test drives the exact scenario (message 0 switches successfully to "planner", message 1's switch to "nonexistent" fails) and verifies both the error propagation and the rollback to "root".

Nice touch using `map[string]error` in the fake to make the failure deterministic — much cleaner than the stochastic approach.

pkg/server/session_manager.go
docker-agent: Excellent fix — capturing `resolved` before the switch correctly matches `pkg/cli/runner.PrepareUserMessage` ordering and handles instruction-only commands without re-running lookup against the wrong table. The rollback on batch failure is a great catch I missed; leaving the runtime mid-batch on an intermediate agent after an error would definitely break subsequent turns.

The updated test with per-agent command tables is exactly what was needed. Thanks for tracking down the shared-map root cause — that's a useful pattern to watch for when testing context-switching logic in general.

pkg/server/session_manager.go
```diff
@@ -1522,6 +1529,38 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
...
+// pkg/cli/runner.PrepareUserMessage. Scoped to agent-switch commands so
+// expanding an instruction-only command doesn't silently rewrite text
+// existing HTTP callers send as literal user input.
+func (sm *SessionManager) applyAgentSwitchCommands(ctx context.Context, rt runtime.Runtime, messages []api.Message) error {
+	originalAgent := rt.CurrentAgentName(ctx)
+	for i := range messages {
+		if messages[i].Role != chat.MessageRoleUser {
+			continue
+		}
+		cmd, _, ok := runtime.LookupCommand(ctx, rt, messages[i].Content)
```
docker-agent, line 1554: **[medium] LookupCommand uses mutated runtime state mid-batch — second command resolved against wrong agent table**

In `applyAgentSwitchCommands`, `LookupCommand` on line 1542 is called with the live `rt` runtime, whose `CurrentAgentName` may have already been mutated by an earlier iteration's `SetCurrentAgent`. When a batch of messages contains two agent-switch commands, the second `LookupCommand` call consults the command table of the *newly-switched* agent (set by the first message), not the original agent.

Concretely: if a caller sends `["/plan …", "/build …"]` in one batch and `/plan` switches to `planner` while `/build` is only declared on the root agent, the second `LookupCommand` fires against `planner`'s command table, fails to find `build`, and silently passes the message through verbatim — different behavior than sending the same two messages individually. No error is returned, so the caller receives no indication of the discrepancy.

The `originalAgent` snapshot on line 1537 is used only for rollback. Consider evaluating all lookups before applying any switches (two-pass: lookup-all then switch-and-replace), or having `LookupCommand` accept an explicit agent name so later iterations are not affected by earlier mutations.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/server/session_manager.go
```diff
@@ -1522,6 +1529,38 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
...
+		// Resolve against the agent that owns the command definition:
+		// the target sub-agent typically doesn't re-declare the routing
+		// command in its own table.
+		resolved := runtime.ResolveCommand(ctx, rt, messages[i].Content)
+		if cmd.Agent != rt.CurrentAgentName(ctx) {
+			if err := rt.SetCurrentAgent(ctx, cmd.Agent); err != nil {
+				// Undo any earlier successful switch in this batch so
+				// the session isn't stuck on a mid-batch agent after
+				// the caller aborts the turn.
+				_ = rt.SetCurrentAgent(ctx, originalAgent)
```
docker-agent: **[medium] Rollback `SetCurrentAgent` error silently discarded — session may be left on wrong agent after failed batch switch**

On line 1555, the rollback call `_ = rt.SetCurrentAgent(ctx, originalAgent)` explicitly discards its error. If `SetCurrentAgent` can fail on the forward path (which is why it is checked), it can also fail on the rollback path — e.g. if the runtime is in a degraded state after the forward switch failed. In that case, the session's current agent remains `cmd.Agent` (an intermediate state), and the function still returns an error to the caller. The caller (`RunSession`) correctly aborts the turn and returns an error to the HTTP client — but the runtime is now pointing at the wrong agent for *all future turns* of this session, with no diagnostic emitted.

A minimal, low-overhead improvement is to log the rollback failure:

```suggestion
				if rbErr := rt.SetCurrentAgent(ctx, originalAgent); rbErr != nil {
					slog.WarnContext(ctx, "failed to restore agent after switch error; session may be on wrong agent",
						"original_agent", originalAgent, "stuck_on", cmd.Agent, "err", rbErr)
				}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):

Two findings were identified in the new `applyAgentSwitchCommands` function — both in the added code, both LIKELY bugs.

None — all surviving findings exceeded the inline threshold (T=55).

None.

pkg/server/session_manager.go
trungutt: Fixed in c187b8433 via the two-pass approach you suggested. Pass 1 loops over every user message and runs `LookupCommand` + `ResolveCommand` against the pre-batch agent, collecting the target agent and resolved content per message. Pass 2 applies the switches and writes the content back. That way `[/plan …, /build …]` gets `/build` resolved against `root`s table even though `/plan` will move the runtime to `planner` first — each message stays independent of what earlier ones did. Added `TestApplyAgentSwitchCommands_MultiBatchLookupUsesPreBatchAgent` with a `commandsByAgent` map where `planner` has no `/build` at all, which is exactly the scenario that would have silently dropped the prefix under the old single-pass code.

pkg/server/session_manager.go
trungutt: Fixed in c187b8433 — adopted your suggestion verbatim. On rollback failure we now `slog.WarnContext` with `original_agent`, `stuck_on`, and the rollback error, so an operator can spot a session that got parked on an intermediate agent. No new test dedicated to log output (there is no slog test-capture pattern in this package), but `TestApplyAgentSwitchCommands_PropagatesSetCurrentAgentError` organically exercises the log path — its `setErr` fails every `SetCurrentAgent` call including the rollback — and the WARN is emitted, visible in the `-v` test run output.

pkg/server/session_manager.go
docker-agent: Excellent fix. The two-pass separation is exactly right: collecting all lookups/resolutions against the pre-batch agent in pass 1, then applying the switches and substituting the resolved content in pass 2, ensures each message's interpretation is independent. The `TestApplyAgentSwitchCommands_MultiBatchLookupUsesPreBatchAgent` test case nails the scenario — `planner` with an empty command table means `/build` in the second slot would have been silently dropped under the old code, but now it resolves correctly against `root`'s table.

The bonus rollback-failure logging is a nice touch; operators now get a WARN with the stuck agent name instead of silent corruption on the error path.

pkg/server/session_manager.go
docker-agent: Thank you for adopting the suggestion and adding test coverage! I'm glad `TestApplyAgentSwitchCommands_PropagatesSetCurrentAgentError` organically exercises the rollback-failure path — that's a pragmatic approach, and having the warning appear in verbose test output provides both a smoke test and documentation of the behavior.

One quick clarification: when the test fails *every* `SetCurrentAgent` call (forward and rollback), does the test assertion verify that the function still returns the *forward* error to the caller, not the rollback error? That's the critical invariant — the rollback failure is logged but shouldn't mask the original cause. If that's already covered, this is in great shape.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is clean and correct. `applyAgentSwitchCommands` correctly mirrors what the CLI's `PrepareUserMessage` does:

- **Two-pass resolution is sound**: all `LookupCommand` / `ResolveCommand` calls happen against the pre-batch agent before any `SetCurrentAgent` call, so every message in a multi-message batch is interpreted against the session's starting agent — not against whatever an earlier message switched to.
- **Rollback on error is correct**: `originalAgent` is captured before any mutations; if `SetCurrentAgent` fails mid-batch, the runtime is restored, and the WARN log on rollback failure provides observability without crashing.
- **Message content is mutated only after a successful switch**: the `messages[s.idx].Content = s.content` line executes after `SetCurrentAgent` succeeds (or is skipped as a no-op), so history is never partially rewritten on failure.
- **Scope is intentionally narrow**: only commands with `Agent` set are resolved; instruction-expansion commands pass through unchanged, preserving backward compatibility for existing HTTP callers.
- **Test suite is thorough**: all critical paths are covered — happy path, idempotent re-assertion, instruction-only commands, unknown commands, non-user roles, error propagation, two-pass lookup independence, and mid-batch rollback.

pkg/server/session_manager.go
```diff
@@ -1522,6 +1529,54 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
...
+			content: runtime.ResolveCommand(ctx, rt, messages[i].Content),
+		})
+	}
+
+	for _, s := range switches {
+		if s.target != rt.CurrentAgentName(ctx) {
+			if err := rt.SetCurrentAgent(ctx, s.target); err != nil {
+				if rbErr := rt.SetCurrentAgent(ctx, originalAgent); rbErr != nil {
+					slog.WarnContext(ctx, "failed to restore agent after switch error; session may be on wrong agent",
+						"original_agent", originalAgent, "stuck_on", s.target, "err", rbErr)
```
docker-agent, line 1570: **[low] `stuck_on` log field names the attempted target, not the agent the runtime is actually on**

When both `SetCurrentAgent(s.target)` and the rollback `SetCurrentAgent(originalAgent)` fail, the warning log records `"stuck_on", s.target`. But `s.target` is the agent the code *tried* to switch to — and that call failed, so the runtime almost certainly did **not** move to `s.target`. The session is more likely stuck on whatever agent it held before the attempt (could be a prior iteration's result or the original), not the one named in the log. An operator reading the warning to diagnose a stuck session would be misled into looking at the wrong agent.

A one-line fix is to query the runtime's actual current state:

```suggestion
						"original_agent", originalAgent, "stuck_on", rt.CurrentAgentName(ctx), "err", rbErr)
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The two-pass design is sound: all `LookupCommand` calls happen before any `SetCurrentAgent` call, correctly scoping each lookup to the pre-batch agent. Rollback on failure and the idempotent skip of same-agent switches are both handled correctly. One low-severity logging accuracy issue was found (inline).

pkg/server/session_manager.go
docker-agent, line 1570: **[low] `stuck_on` log field names the wrong agent when rollback also fails**

When `SetCurrentAgent(ctx, s.target)` fails on line 1567, the runtime was never placed on `s.target` — the call errored out. If a prior iteration in the same batch had already successfully switched to an intermediate agent `T_prev`, the runtime is currently on `T_prev` at the point the rollback attempt is made. If the rollback (`SetCurrentAgent(ctx, originalAgent)`) also fails, the session is stuck on `T_prev` — but the log records `stuck_on: s.target`, which is the agent the *failed* switch was targeting, not the agent the runtime is actually on.

An operator diagnosing a stuck session from this log would search for the wrong agent. The correct value to surface is the runtime's actual current agent at the time of rollback failure, i.e. the result of a `rt.CurrentAgentName(ctx)` call at that point.

This only manifests when a batch contains multiple agent-switch commands **and** both the Nth switch and the rollback fail — an uncommon but valid scenario with multi-message batches.

```suggestion
				slog.WarnContext(ctx, "failed to restore agent after switch error; session may be on wrong agent",
					"original_agent", originalAgent, "stuck_on", rt.CurrentAgentName(ctx), "err", rbErr)
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/server/session_manager.go
```diff
@@ -1522,6 +1529,54 @@ func (sm *SessionManager) applyRunModelOverride(ctx context.Context, rs *activeR
...
+				}
+				return fmt.Errorf("switch agent to %q: %w", s.target, err)
+			}
+		}
+		messages[s.idx].Content = s.content
```
docker-agent, line 1575: **[low] Partial message content mutation if a multi-message batch switch fails mid-way**

The second-pass loop rewrites `messages[s.idx].Content = s.content` immediately after each successful `SetCurrentAgent`, before the loop finishes. If the loop contains multiple switch entries (a batch of messages each starting with a different slash command) and a *later* entry's `SetCurrentAgent` fails, the earlier messages will already have had their slash-command prefix stripped while the later ones retain it. The `messages` slice is passed by reference, so the caller sees a partially-rewritten slice.

In the current `RunSession` call-site, this is harmless — the caller discards the slice on error (returns `nil, err`) and never appends anything to the session. But if retry logic is ever added that re-uses the same `messages` slice, it will encounter inconsistently-modified content (some prefixes stripped, others intact) with no indication that a partial mutation occurred.

A defensive fix is to stage the content rewrites and apply them only after all `SetCurrentAgent` calls in the loop have succeeded:

```suggestion
	for _, s := range switches {
		if s.target != rt.CurrentAgentName(ctx) {
			if err := rt.SetCurrentAgent(ctx, s.target); err != nil {
				if rbErr := rt.SetCurrentAgent(ctx, originalAgent); rbErr != nil {
					slog.WarnContext(ctx, "failed to restore agent after switch error; session may be on wrong agent",
						"original_agent", originalAgent, "stuck_on", s.target, "err", rbErr)
				}
				return fmt.Errorf("switch agent to %q: %w", s.target, err)
			}
		}
	}
	for _, s := range switches {
		messages[s.idx].Content = s.content
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The new `applyAgentSwitchCommands` function is well-structured with a clear two-pass design that correctly separates lookup from apply. The idempotence check and rollback logic are present and tested. Two minor low-severity observations are noted inline on newly introduced code.

dgageot (approved):

PR #3783, opened 2026-07-22
feat(config): add description attribute to models

dgageot:
Agent configs can reference multiple models, but there has been no way to annotate what each model is for — why you chose it, what it is good at, or what role it plays in the agent. This PR adds an optional `description` field to `ModelConfig` so authors can attach a short, human-readable summary to any model entry.

The field is purely informational: it is stored in the config, surfaced in tooling, and never sent to the model provider. It works with every model form — inline shorthand strings, full `ModelConfig` objects, and `first_available` selectors. When a `first_available` selector resolves to a concrete model, the selector's own description is preserved on the resolved value so callers always see it. The shorthand marshaller is fixed to carry the description through the round-trip rather than silently dropping it.

Schema, docs, and a worked example (`examples/model_descriptions.yaml`) are all updated. New tests cover the round-trip and the `first_available` resolution path.

### DIFF
--- a/pkg/config/first_available.go
+++ b/pkg/config/first_available.go
@@ -60,6 +60,11 @@ func resolveFirstAvailableModel(ctx context.Context, cfg *latest.Config, name st
 
 	slog.DebugContext(ctx, "Resolved first_available model",
 		"model_name", name, "selected", ref, "provider", chosen.Provider, "model", chosen.Model)
+	// The selector's description documents the selection itself; keep it over
+	// the chosen candidate's own description.
+	if m.Description != "" {
+		chosen.Description = m.Description
+	}
 	cfg.Models[name] = chosen
 	return nil
 }
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1510,6 +1510,10 @@
           "type": "string",
           "description": "Model name"
         },
+        "description": {
+          "type": "string",
+          "description": "Optional human-readable summary of the model's purpose or strengths (e.g. 'fast and cheap, good for summaries')"
+        },
         "temperature": {
           "type": "number",
           "description": "Sampling temperature",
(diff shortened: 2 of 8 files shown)


pkg/config/first_available.go
```diff
@@ -60,6 +60,11 @@ func resolveFirstAvailableModel(ctx context.Context, cfg *latest.Config, name st
 
 	slog.DebugContext(ctx, "Resolved first_available model",
 		"model_name", name, "selected", ref, "provider", chosen.Provider, "model", chosen.Model)
+	// The selector's description documents the selection itself; keep it over
+	// the chosen candidate's own description.
+	if m.Description != "" {
```
docker-agent, line 65: **[medium] Candidate description leaks through when the selector has no description**

The guard `if m.Description != ""` copies the selector's description onto the resolved candidate only when the selector carries one. When `m.Description` is empty the candidate's own description is left untouched, so callers see the chosen candidate's description instead of the selector's.

The comment directly above says _"keep it [the selector's description] over the chosen candidate's own description"_, which implies the selector should always win — meaning the candidate's description should be cleared when the selector has none. If the intent is purely a preference/fallback (use the selector description when present, fall back to the candidate's otherwise), the comment is misleading and should be updated to say so.

If clearing is indeed the intent, add an `else` branch:

```suggestion
	if m.Description != "" {
		chosen.Description = m.Description
	} else {
		chosen.Description = ""
	}
```

If fallback behaviour is intentional, update the comment to reflect that — e.g. _"Prefer the selector's description; fall back to the candidate's own."_

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):

gtardif (approved):

PR #3789, opened 2026-07-22
fix(tui): never render dead code-block copy buttons

dgageot:
The ⎘ copy affordance rendered by `FastRenderer` on every fenced code block was only wired up for hit-testing in assistant message views. In reasoning blocks, elicitation dialogs, leantui, shell output, and welcome messages the icon appeared but clicking it did nothing.

The fix adds a `HideCopyIcon` option to `FastRenderer` (exposed as `markdown.NewRendererWithoutCopyIcon`). When hidden, the icon renders as plain padding and records no `CodeBlock` hit target. Reasoning blocks, elicitation dialogs, and leantui — which never participate in code-block click handling — now use this renderer. Shell-output and welcome message views gain functional copy icons by populating their `codeBlocks` slice (accounting for the welcome envelope's line offset) so the existing `codeBlockAt` click path handles them. A new test, `TestCodeBlockCopyIconsAreNeverDead`, asserts that every rendered copy icon is clickable and that surfaces incapable of handling clicks render no icon at all.

### DIFF
--- a/pkg/leantui/ui/render.go
+++ b/pkg/leantui/ui/render.go
@@ -56,7 +56,8 @@ func RenderAssistantLines(text string, width int) []string {
 		return nil
 	}
 
-	rendered, err := markdown.NewRenderer(width).Render(text)
+	// No copy icon: leantui does not hit-test clicks on code blocks.
+	rendered, err := markdown.NewRendererWithoutCopyIcon(width).Render(text)
 	if err != nil {
 		return WrapANSI(text, width)
 	}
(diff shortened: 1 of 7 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is clean and correct. The `HideCopyIcon` option integrates properly with the parser pool lifecycle, the line-offset calculations for welcome messages (border + padding = 2) and shell output (no envelope) are accurate, and the error paths correctly clear `codeBlocks` to avoid stale state. The new `TestCodeBlockCopyIconsAreNeverDead` test provides good coverage of the invariant being enforced.

gtardif (approved):

PR #3791, opened 2026-07-22
sandbox: rely on sbx-login injection for gateway auth, modern sandbox CLI surface

dgageot:
The sandbox integration previously relied on a host-side background process writing fresh Docker tokens into a `sandbox-tokens.json` file that the in-sandbox environment provider would read and expose as `DOCKER_TOKEN`. This approach was fragile: it depended on a polling writer, file-based coordination across host and sandbox, and a bespoke token-forwarding path that duplicated credential logic.

This PR replaces that mechanism with `sbx-login` injection, which landed in the sandbox runtime via docker/sandboxes commit 57f83f158. A generated mixin kit (`pkg/sandbox/loginkit.go`) declares a reserved `sbx-login` credential service; the sandbox proxy picks this up and injects the user's current Docker login JWT directly into outbound HTTPS requests to the docker.com gateway host, while also exporting `DOCKER_TOKEN` inside the sandbox as a proxy-managed sentinel. The host-side token writer (`pkg/environment/sandbox_tokens.go`) and its in-sandbox counterpart are deleted entirely.

Alongside the auth change, the PR modernises the sandbox CLI surface: both the `docker sandbox` and `sbx` backends now use the current `policy allow network --sandbox NAME host1,host2` spelling (old spellings were removed from recent CLI releases), the deprecated `vms` ls fallback is dropped, `rm` is always called with `--force`, and new sandboxes are created under the canonical `docker-agent` name rather than the old `cagent` alias. The default template is now `docker/docker-agent-sbx-templates:latest`, the repo-published image built by this repo's CI from the latest release.

This requires a recent version of `sbx` / `docker sandbox` that includes the `sbx-login` credential service support. Sandboxes holding a stale login kit (written by an older version of this tool) are rejected at creation time rather than silently reused.


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly replaces the fragile host-side token-polling mechanism (`sandbox_tokens.go`) with sbx-login injection via a generated mixin kit (`loginkit.go`). The implementation is well-structured:

- **loginkit.go**: Hostname validation is correct (ToLower before regex, *.docker.com scope check). YAML generation uses `%%s` correctly to produce a literal `%s` template placeholder for the sbx proxy. Atomic file writes prevent partial-read races on concurrent sandbox creates. The `staleLoginKit` comparison logic correctly strips `:ro` suffixes before comparing paths.
- **sandbox.go**: The loginKit path is appended to extras before `cleanExtras` (so it gets deduplicated and normalised), and it's separately passed as `--kit` to `create`. Stale-kit detection prevents reusing a sandbox that would keep authenticating against the wrong gateway host.
- **backend.go**: Unifying both backends onto the modern CLI surface (`policy allow network --sandbox` and `rm --force`) is an intentional breaking change explicitly documented in the PR description and enforced by the new `TestAllowHosts_ArgvSpelling` test.
- **Removal of sandbox_tokens.go**: Clean deletion with no orphaned references; the host-side polling goroutine and token file path are gone.

aheritier (approved):

PR #3796, opened 2026-07-23
feat(config): make oauth clientId optional for remote MCP toolsets

dgageot:
Some OAuth providers support Dynamic Client Registration (RFC 7591) or an interactive credential prompt, so hard-requiring `clientId` in the remote MCP `oauth` block prevents users from using those flows even though the runtime already handles them. It also blocked a common pattern where you want to pin the callback port or supply a custom redirect URL without pre-registering a client.

This change drops `clientId` from the `required` list in both the JSON schema and the `latest` config package. When `clientId` is absent, the runtime falls back to dynamic registration or prompts interactively; `callbackPort`, `scopes`, and `callbackRedirectURL` continue to be honored as before. As a guard-rail, specifying `clientSecret` without a `clientId` is now rejected with a clear validation error, since a secret is meaningless without a corresponding client identity. Frozen older config versions (`v0`, `v1`, …) are untouched.

The docs and the `atlassian-expert` example have been updated to reflect the new optionality.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -3089,11 +3089,11 @@
     },
     "RemoteOAuthConfig": {
       "type": "object",
-      "description": "OAuth configuration for remote MCP servers that do not support Dynamic Client Registration (RFC 7591)",
+      "description": "OAuth configuration for remote MCP servers. Set clientId for servers that do not support Dynamic Client Registration (RFC 7591); other fields tune the flow and also apply when the client is registered dynamically",
       "properties": {
         "clientId": {
           "type": "string",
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The validation logic is correct and well-tested. Making `clientId` optional is a clean, backward-compatible change: the guard-rail (rejecting `clientSecret` without `clientId`) is logically sound, the JSON schema update aligns with the Go validation, and the three new test cases cover the meaningful new branches.

aheritier (approved):

PR #3797, opened 2026-07-23
fix(tui): keep copy label background under "copied" flash

dgageot:
When a user clicks a copy button, the label briefly swaps to a \"copied\" confirmation. For code blocks, the background of that flash was already forced to match the code block's band. For user-message copy labels, which sit on the `BackgroundAlt` band, no such fix existed, so the transient label rendered with a transparent background and punched a visible hole in the colored band.

The fix introduces `backgroundAt(line, col)`, which walks the ANSI SGR escape sequences of a rendered line and returns the background color in effect at the given column — handling truecolor, 256-indexed, basic, and bright colors, plus resets. The \"copied\" flash now calls this helper for all copy labels, replacing the code-block-only special case with a single, general path.

New regression tests cover the user-message band case, the code-block band case, and a table-driven suite for `backgroundAt` itself.

### DIFF
--- a/pkg/tui/components/messages/copyfeedback.go
+++ b/pkg/tui/components/messages/copyfeedback.go
@@ -1,11 +1,11 @@
 package messages
 
 import (
+	"image/color"
 	"strings"
 	"time"
 
 	tea "charm.land/bubbletea/v2"
-	"charm.land/lipgloss/v2"
 	"github.com/charmbracelet/x/ansi"
 
 	"github.com/docker/docker-agent/pkg/tui/components/markdown"
@@ -80,11 +80,10 @@ func (m *model) applyCopiedFlash(lines []string, viewportStartLine int) []string
 	end := start + ansi.StringWidth(label)
 
 	style := styles.SuccessStyle.Bold(true)
-	if f.codeBlock {
-		// Keep the code block's background band intact behind the swap.
-		if bg := styles.MarkdownStyle().CodeBlock.BackgroundColor; bg != nil {
-			style = style.Background(lipgloss.Color(*bg))
-		}
+	// Keep whatever band the copy label sat on (user message background,
+	// code block background) intact behind the swap.
+	if bg := backgroundAt(line, start); bg != nil {
+		style = style.Background(bg)
 	}
 
 	result := make([]string, len(lines))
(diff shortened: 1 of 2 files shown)


pkg/tui/components/messages/copyfeedback.go
```diff
@@ -94,3 +93,50 @@ func (m *model) applyCopiedFlash(lines []string, viewportStartLine int) []string
...
+// backgroundAt returns the SGR background color in effect at the given cell
+// column of a rendered line, or nil when none is set.
+func backgroundAt(line string, col int) color.Color {
+	p := ansi.GetParser()
+	defer ansi.PutParser(p)
+
+	var bg color.Color
+	var state byte
+	for w := 0; line != "" && w <= col; {
+		seq, width, n, newState := ansi.DecodeSequence(line, state, p)
```
docker-agent, line 106: **[high] `backgroundAt` can spin in an infinite loop when `ansi.DecodeSequence` returns `n == 0`**

`ansi.DecodeSequence` can return `n == 0` when the first byte of the remaining input is an invalid UTF-8 sequence (e.g., an isolated continuation byte `0x80`–`0xBF`). In that case:
- `line = line[n:]` is a no-op — `line` does not shrink
- `line != ""` remains `true`
- `w` never advances (the invalid byte has visual `width == 0`)
- The loop condition `w <= col` stays satisfied

This causes a busy infinite loop, freezing the TUI.

From `charmbracelet/x/ansi@v0.11.7/parser_decode.go`, the "Invalid UTF-8 sequence" return path is:
```go
// Invalid UTF-8 sequence
return b[:i], 0, i, NormalState  // i == 0 on first byte → n == 0
```

This path is reachable whenever a rendered line contains an isolated continuation byte or other invalid UTF-8. While charmbracelet's own rendering is unlikely to emit such bytes, `backgroundAt` is called on arbitrary terminal lines, so robustness matters.

**Fix**: add an early-out guard immediately after the `DecodeSequence` call:

```suggestion
		seq, width, n, newState := ansi.DecodeSequence(line, state, p)
		if n == 0 {
			break
		}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

1 high-severity finding in new code.

aheritier (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

PR #3804, opened 2026-07-23
chore: remove every trace of the discontinued agent catalog

dgageot:
The `agentcatalog` Docker Hub namespace was discontinued, so every reference to it — in code, CLI help texts, tests, documentation, and examples — pointed users toward something that no longer exists. This PR removes all of those references cleanly.

All `agentcatalog/*` examples are replaced with the neutral placeholder `myorg/agent:tag` (or `myorg/agent` where an untagged form is semantically required). The "Agent Catalog" documentation section is removed, along with all hub.docker.com/u/agentcatalog links in the docs index, quickstart, multi-agent guide, and footer partial. The CLI reference table row previously labeled "Agent catalog" is renamed "Hub shorthand". The example file `examples/sub-agents-from-catalog.yaml` is renamed to `examples/sub-agents-from-registry.yaml` with updated wording to reflect that sub-agents can be pulled from any OCI registry.

`pkg/teamloader` now skips that renamed example in `TestLoadExamples` since it references a placeholder repo that cannot actually be pulled. The `docker agent share push`/`pull` commands and OCI artifact hosting remain fully supported — only references to the catalog namespace are gone.


rumpl (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly removes all references to the discontinued `agentcatalog` Docker Hub namespace. All example strings are consistently replaced with the neutral placeholder `myorg/agent:tag`, the renamed example file `sub-agents-from-registry.yaml` is correctly added to `skipExamples` in the teamloader test (so the placeholder OCI reference is never resolved in tests), and no logic errors, resource leaks, or broken test coverage were introduced. The change is safe to merge.

PR #3806, opened 2026-07-23
feat: three-mode safety policy with decoupled classifier

trungutt:
## Why

Shell command approval today is a coarse two-mode toggle:

- **safer** on a shell toolset: the classifier asks about anything
  it doesn't positively recognise as safe.
- **safer** off (or `--yolo`): everything auto-approves.

That's it. There's no way to say "I trust you on read-only calls,
but ask me before you destroy something". A user who finds strict
mode annoying has only one escape hatch — full auto-approve — and
loses the safety net for genuinely destructive commands.

The classifier itself is also entangled with the approval verdict.
It reads the session's safety policy, decides whether to ask, and
emits its verdict directly. That makes it impossible to reuse the
classifier's judgement anywhere else (UI badges, telemetry, other
gating logic) without duplicating its rules, and adding a new
approval mode means editing the classifier.

## What this PR does

Introduce **three explicit safety modes** and split the classifier
into a pure labeller so the mode is what actually gates the call.

```
                     ┌─────────┬─────────────┬─────────┐
                     │  safe   │ destructive │ unknown │
        ┌────────────┼─────────┼─────────────┼─────────┤
        │ Strict     │   ASK   │     ASK     │   ASK   │
        │ Balanced   │  ALLOW  │     ASK     │   ASK   │
        │ Autonomous │  ALLOW  │    ALLOW    │  ALLOW  │
        └────────────┴─────────┴─────────────┴─────────┘
```

- **Strict** — today's default. Prompt on every tool call.
- **Balanced** — new middle tier. The read-only path (ls, git status,
  docker ps, …) flows through silently; destructive and unrecognised
  calls still prompt.
- **Autonomous** — legacy yolo. Auto-approve everything.

Custom Allow/Ask/Deny rules layer on top and always win over the
mode. A distinction the runtime enforces: a rule the user added at
runtime (session-scoped) beats even Autonomous, while a rule the
agent author baked into the YAML is advisory and yields to
Autonomous.

The classifier that used to make the verdict now only labels the
call. Downstream (the runtime, hosting applications, telemetry) all
see the same three values — `safe` / `destructive` / `unknown` — and
each layer decides what to do with them independently.

```
                      ┌────────────────────┐        ┌────────────────────┐
   shell call    ───▶ │  safer_shell hook  │ ─────▶ │  mode × label      │ ─▶ allow / ask
                      │  (label only)      │ label  │  decision table    │
                      └────────────────────┘        └────────────────────┘
                        safe / destructive           reads session mode
                        / unknown
```

## Notes for the reviewer

- The `safer: true` opt-in on shell toolsets is removed from the
  schema. Any agent that declares a shell toolset gets the
  classifier automatically — it never blocks, only labels.
- Resume verbs the CLI/API can send back grow two entries so the
  confirmation UI can offer "approve + switch to Balanced" and
  "approve + switch to Autonomous" alongside the existing single-call
  approve.
- No behavioural change for existing `--yolo` users: yolo maps to
  Autonomous and every call still auto-approves.
- Strict is unchanged from today's no-yolo default — every call
  prompts. The only new behaviour is Balanced.


pkg/tui/components/toolconfirm/toolconfirm.go
```diff
@@ -36,9 +36,10 @@ const (
 	// ApproveTool runs the call and always allows the tool, scoped by the
 	// permission pattern from BuildPermissionPattern.
 	ApproveTool
-	// ApproveSession runs the call and approves all tools for the rest of
-	// the session.
-	ApproveSession
+	// ApproveAutonomous runs the call and flips the session to
+	// SafetyPolicyAutonomous (auto-approve everything for the rest of
+	// the session).
+	ApproveAutonomous
```
docker-agent, line 46: **[medium] `ApproveBalanced` decision missing from TUI — 'approve + switch to Balanced' unreachable from the interactive confirmation dialog**

The PR description states the confirmation UI should offer _"approve + switch to Balanced"_ and _"approve + switch to Autonomous"_ as distinct choices. `ResumeTypeApproveBalanced` is wired into the protocol and the server-side `handleResume`, but the TUI confirmation component has no key binding, no `Decision` constant, and no `OptionsHelp` entry for it.

The only new shortcut added to `ActionKeys = "YNTA"` is `A` → `ApproveAutonomous`. A user who wants to opt into Balanced mode from the interactive prompt has no key to press — `ApproveBalanced` is reachable only through the API, not from the TUI dialog.

**Impact**: The stated feature goal is partially unimplemented. Any user interacting via the terminal cannot opt into Balanced mode from a confirmation prompt; they must use the API or set the mode before starting the session. If a future key label says "approve balanced", pressing it will invoke the wrong resume verb.

**Suggested fix**: Add `ApproveBalanced` to the `Decision` enum, a key binding (e.g. `B`), an entry in `ActionKeys`, `DecisionForAction`, `OptionsHelp`, and the `Resume()` switch — mirroring the `ApproveAutonomous` pattern already present.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/server/session_manager.go
```diff
@@ -949,6 +949,9 @@ func (sm *SessionManager) ResumeSession(ctx context.Context, sessionID, confirma
 			}
 		}
 		if mutated {
+			slog.DebugContext(ctx, "ResumeSession mutated session",
+				"session_id", sessionID, "safety_policy", string(rt.session.SafetyPolicy),
```
docker-agent: **[medium] Data race: `rt.session.SafetyPolicy` and `rt.session.ToolsApproved` read without holding `s.mu` in debug log**

After `rt.session.SetSafetyPolicy(...)` completes and releases `s.mu`, the `slog.DebugContext` call on lines 952–954 reads `rt.session.SafetyPolicy` and `rt.session.ToolsApproved` as bare field accesses — no lock is held at the point of the read.

```go
slog.DebugContext(ctx, "ResumeSession mutated session",
    "session_id", sessionID, "safety_policy", string(rt.session.SafetyPolicy),
    "tools_approved", rt.session.ToolsApproved)
```

**Trigger**: The dispatcher's `handleResume` (which runs concurrently) also calls `c.sess.SetSafetyPolicy(...)` on the same `*Session` object after receiving the Resume channel message. The session_manager holds `sm.mux`, not `s.mu`, so there is no barrier preventing the dispatcher's write from racing with this log read. Go's race detector would flag this as an unsynchronized concurrent read and write on a `string` (`SafetyPolicy`) and a `bool` (`ToolsApproved`).

**Impact**: Undefined behaviour under the Go memory model; the race detector will report this. The practical consequence is a stale or torn log value, but the race itself is real.

**Fix**: Read the fields inside a helper that holds `s.mu`, or snapshot them immediately after the `SetSafetyPolicy` call (before the lock is released) and log the snapshots. For example:

```suggestion
			policy, approved := rt.session.SafetyPolicySnapshot()
			slog.DebugContext(ctx, "ResumeSession mutated session",
				"session_id", sessionID, "safety_policy", string(policy),
				"tools_approved", approved)
```

(Alternatively, inline the read inside a `SetSafetyPolicy`-style helper that returns the new values, or use `s.mu.RLock()` around the log reads.)

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):

Two medium-severity findings (both CONFIRMED) introduced by this PR:

1. **Missing TUI binding for ApproveBalanced** — The new `ResumeTypeApproveBalanced` resume verb exists in the protocol but has no corresponding key binding, `Decision` constant, or `OptionsHelp` entry in the confirmation dialog. 'Approve + switch to Balanced' is unreachable from the TUI.

2. **Data race on session fields in debug log** — After `SetSafetyPolicy` releases `s.mu`, the `slog.DebugContext` call reads `rt.session.SafetyPolicy` and `rt.session.ToolsApproved` without holding the lock. The concurrent `handleResume` path in the dispatcher also writes these fields, creating a real data race detectable by Go's race detector.

docker-agent (commented):

This PR is well-structured and internally consistent. The three-mode safety policy (Strict / Balanced / Autonomous) and the decoupled classifier are cleanly implemented. The mode × label decision table is correct, the resume verb renaming is complete, and the new tests cover the key new behaviors. One propagation gap surfaced below.

---

- [low] `pkg/server/session_manager.go`:949 — `SetSafetyPolicy` called twice on the same session for ApproveBalanced/ApproveAutonomous resumes (once in `ResumeSession` under `sm.mux`, once in the dispatcher goroutine); both writes set the same value and session-level locking makes this safe, but the intermediate `UpdateSession` persist fires before the dispatcher's own write, so the store always sees the correct final state (confidence: weak 35/100)

pkg/session/session.go
```diff
@@ -1462,20 +1456,20 @@ func (s *Session) SetToolsApproved(approved bool) {
...
 // so legacy branches on ToolsApproved keep working. Runtime callers
 // use this to persist a user's mid-session mode change (e.g. opting
-// into safe-auto from a confirmation prompt).
+// into balanced from a confirmation prompt).
 func (s *Session) SetSafetyPolicy(policy SafetyPolicy) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.SafetyPolicy = policy
-	if policy == SafetyPolicyUnsafe {
+	if policy == SafetyPolicyAutonomous {
```
docker-agent, line 1478: **[medium] `SafetyPolicyBalanced` is not propagated to sub-sessions — child sessions silently downgrade to Strict**

When a user responds "approve + switch to Balanced" at a confirmation prompt, the dispatcher calls `sess.SetSafetyPolicy(SafetyPolicyBalanced)`. This correctly records the mode on the active session, but `SetSafetyPolicy(SafetyPolicyBalanced)` does **not** flip `ToolsApproved = true` (only `SafetyPolicyAutonomous` does that, at this line). So `sess.IsToolsApproved()` returns `false` for Balanced.

The sub-session propagation paths in this codebase use only `ToolsApproved` as their carry signal:

```go
// pkg/runtime/agent_delegation.go:534  (background agents)
ToolsApproved: params.ParentSession.IsToolsApproved(), // false for Balanced

// pkg/runtime/agent_delegation.go:594  (task delegation)
ToolsApproved: sess.IsToolsApproved(), // false for Balanced

// pkg/runtime/skill_runner.go:112
ToolsApproved: sess.IsToolsApproved(), // false for Balanced
```

`SubSessionConfig` has no `SafetyPolicy` field, so any Balanced-mode parent creates a child session with `WithToolsApproved(false)` → no explicit SafetyPolicy → the child runs under **Strict** (prompts on every tool call). Any sub-session or skill invoked while the user is in Balanced mode will silently revert to pestering the user for confirmation on every shell read-only command.

**Fix**: add a `SafetyPolicy session.SafetyPolicy` field to `SubSessionConfig` (in `pkg/runtime/agent_delegation.go`), populate it from `sess.SafetyPolicy` at each call site, and propagate it via `session.WithSafetyPolicy(cfg.SafetyPolicy)` in `runCollecting`/`createChildSession`.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/runtime/toolexec/dispatcher.go
```diff
@@ -434,51 +422,77 @@ func (c *call) approveAndRun(ctx context.Context, runTool func() CallOutcome) Ca
...
+func (c *call) permissionDecision() PermissionDecision {
 	var checkers []NamedChecker
 	if c.d.Permissions != nil {
 		checkers = c.d.Permissions(c.sess)
 	}
-	return Decide(
-		c.sess.IsToolsApproved(),
+	label := c.classifierLabel()
+	d := Decide(
+		c.sess.SafetyPolicy,
```
docker-agent: **[high] Data race: `sess.SafetyPolicy` read without mutex in `permissionDecision`**

The `permissionDecision` function reads `c.sess.SafetyPolicy` directly at lines 451 and 460 with no lock held:

```go
d := Decide(
    c.sess.SafetyPolicy,   // ← bare field read
    label,
    ...
)
slog.Debug("permissionDecision",
    "safety_policy", string(c.sess.SafetyPolicy),  // ← bare field read
    ...
)
```

`SetSafetyPolicy` (called by `handleResume` for `ResumeTypeApproveBalanced` / `ResumeTypeApproveAutonomous`) writes the same field under `s.mu.Lock()`. There is no corresponding `GetSafetyPolicy()` accessor that holds `s.mu.RLock()` for reads.

When a user approves a call with "approve-balanced" or "approve-autonomous" while sibling tool calls are concurrently executing `permissionDecision`, Go's race detector will flag a data race between the write in `SetSafetyPolicy` and these unguarded reads. This is the exact same pattern the project already guards correctly in `IsToolsApproved()` (which uses `mu.RLock()`).

**Fix**: Add a `GetSafetyPolicy()` accessor to `Session` that reads under `mu.RLock`, mirroring `IsToolsApproved()`:

```suggestion
	d := Decide(
		c.sess.GetSafetyPolicy(),
		label,
		checkers,
		c.tc.Function.Name,
		ParseToolInput(c.tc.Function.Arguments),
	)
```

And add to `pkg/session/session.go`:
```go
func (s *Session) GetSafetyPolicy() SafetyPolicy {
    s.mu.RLock()
    defer s.mu.RUnlock()
    return s.SafetyPolicy
}
```

Apply the same fix to all other bare `c.sess.SafetyPolicy` reads in the dispatcher.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One high-severity data race was found and confirmed in the new three-mode safety policy implementation.

- [medium] pkg/runtime/toolexec/permissions.go:101 — Team-level `ForceAsk` now yields to Autonomous mode — agent YAML `ask:` rules are advisory under Autonomous (DISMISSED by verifier: explicitly documented as intentional design)

pkg/runtime/toolexec/dispatcher.go
docker-agent: **[medium] `c.sess.SafetyPolicy` read without holding `sess.mu`, racing against concurrent `SetSafetyPolicy()` writes**

The PR replaces the previously lock-guarded `c.sess.IsToolsApproved()` call (which used `s.mu.RLock()`) with a bare field read of `c.sess.SafetyPolicy`, while the concurrent writer `SetSafetyPolicy()` still holds `s.mu.Lock()`. When a batch of parallel tool calls is in flight and one goroutine's `handleResume()` calls `SetSafetyPolicy(SafetyPolicyBalanced)`, sibling goroutines executing `permissionDecision()` concurrently read `c.sess.SafetyPolicy` without any lock — a data race per the Go memory model.

The same unprotected read is repeated on line 460 in the `slog.Debug` call.

**Fix**: add a locked `GetSafetyPolicy()` accessor on `*Session` (mirroring `IsToolsApproved()`), and call it here instead of the direct field access:

```suggestion
		c.sess.GetSafetyPolicy(),
```

Where `GetSafetyPolicy` is:
```go
func (s *Session) GetSafetyPolicy() SafetyPolicy {
    s.mu.RLock()
    defer s.mu.RUnlock()
    return s.SafetyPolicy
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

This PR cleanly replaces the two-mode (yolo / strict) approval toggle with a well-structured three-mode table (Strict / Balanced / Autonomous) and decouples the `safer_shell` classifier from the approval verdict. The architecture is sound and the new tests cover the key transitions.

One medium-severity data race was identified in the critical approval path.

pkg/tui/components/toolconfirm/toolconfirm.go
trungutt: Done in 41c682d0f — added `ApproveBalanced` to the shared toolconfirm surface (Decision constant, `B` key binding, `OptionsHelp` entry, `Resume` mapping) and wired it into the docker-agent TUI dialog. Scoped to the TUI-facing package as flagged; leaving the CLI text prompt out of this PR.

pkg/server/session_manager.go
trungutt: Done in dbda8ff89 — dropped the debug log entirely. It was left over from local development; the field reads it did are gone with it, so the race is closed at the source.

pkg/runtime/toolexec/dispatcher.go
trungutt: Done in dbda8ff89 — added a locked `GetSafetyPolicy()` accessor to `*Session` (mirroring `IsToolsApproved()`) and routed the `Decide()` call plus every other bare `sess.SafetyPolicy` read in the runtime through it. The `permissionDecision` debug log that read the field a second time is dropped.

trungutt: Done in dbda8ff89 — same fix as the parallel finding. `Session.GetSafetyPolicy()` (RLocked accessor) plus removal of the debug log that carried the second bare read.

pkg/session/session.go
trungutt: Done in 8105dcb00 — added `SafetyPolicy` to `SubSessionConfig`, populated it from the parent at every construction site (`RunAgent`, `handleTaskTransfer`, `handleRunSkill`), and propagated it via `session.WithSafetyPolicy` in `newSubSession`. Balanced parents no longer silently downgrade their children to Strict.

pkg/tui/components/toolconfirm/toolconfirm.go
docker-agent: Perfect — verified the implementation.  now has full parity with  across the shared surface (Decision enum, key binding, help text, resume mapping) and the TUI dialog dispatches it correctly. The  shortcut makes 'approve + switch to Balanced' reachable from the interactive prompt as designed.

Scoping this to the TUI-facing package makes sense; the CLI text prompt is a separate flow. Thanks for the quick turnaround!

pkg/session/session.go
docker-agent: Perfect — the fix covers all three propagation paths (delegated agents, task transfers, and skill runs) and correctly pipes GetSafetyPolicy() through SubSessionConfig to WithSafetyPolicy at construction. Balanced-mode parents will now spawn Balanced-mode children instead of silently downgrading them to Strict. Thanks for the quick turnaround!

pkg/leantui/update.go
```diff
@@ -443,7 +443,7 @@ func (m *model) handleConfirmKey(k ui.Key) {
 	case 'a', 'A':
 		m.resolveConfirm(runtime.ResumeApproveTool(m.screen.Confirm.Tool))
 	case 's', 'S':
-		m.resolveConfirm(runtime.ResumeApproveSession())
+		m.resolveConfirm(runtime.ResumeApproveAutonomous())
```
docker-agent, line 448: **[medium] leantui `handleConfirmKey` wires `s`/`S` to Autonomous but provides no key for Balanced mode**

The PR introduces `ResumeApproveBalanced()` as a new resume verb and wires it in the full TUI (`pkg/tui/dialog/tool_confirmation.go`), but `pkg/leantui/update.go` only maps `s`/`S` to `ResumeApproveAutonomous()`. There is no `b`/`B` binding for `ResumeApproveBalanced()`.

Users of the leantui interface (the lightweight/no-TUI variant) are offered three options from a confirmation prompt: approve once (`y`), approve this tool forever (`a`), or go fully Autonomous (`s`). The new Balanced mode — the PR's main new feature — is inaccessible from leantui. A user in leantui who wants "auto-approve safe calls only" has no way to opt into it interactively and must either approve call-by-call or jump straight to fully Autonomous.

The fix is to add a `b`/`B` case that calls `m.resolveConfirm(runtime.ResumeApproveBalanced())`.

```suggestion
	case 'b', 'B':
		m.resolveConfirm(runtime.ResumeApproveBalanced())
	case 's', 'S':
		m.resolveConfirm(runtime.ResumeApproveAutonomous())
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):

This PR cleanly implements the three-mode safety policy (Strict/Balanced/Autonomous) with a well-designed decoupled classifier. The core logic in `pkg/runtime/toolexec/permissions.go` is correct: the mode × label decision table, the session-vs-team ForceAsk precedence, and the legacy `ToolsApproved` sync are all sound.

One confirmed gap: the new **Balanced** mode is wired in the full TUI but is missing from the leantui keyboard handler — see the inline comment below.

pkg/cli/printer.go
```diff
@@ -21,10 +21,10 @@ import (
 type ConfirmationResult string
 
 const (
-	ConfirmationApprove        ConfirmationResult = "approve"
-	ConfirmationApproveSession ConfirmationResult = "approve_session"
-	ConfirmationReject         ConfirmationResult = "reject"
-	ConfirmationAbort          ConfirmationResult = "abort"
+	ConfirmationApprove           ConfirmationResult = "approve"
+	ConfirmationApproveAutonomous ConfirmationResult = "approve_autonomous"
```
docker-agent, line 25: **[medium] CLI printer exposes no way to select Balanced mode — no `[b]alanced` key in the interactive prompt**

`PrintToolCallWithConfirmation` advertises `([y]es/[a]ll/[n]o)` and handles only three choices: `y`/`Y` → Approve, `a`/`A` → Autonomous, `n`/`N` → Reject. The PR introduces Balanced as the "new middle tier" and correctly wires it into the TUI (`toolconfirm.go` adds `B` → `ApproveBalanced` → `ResumeTypeApproveBalanced`), but the non-TUI CLI path is missing the equivalent:

- No `ConfirmationApproveBalanced` constant added alongside `ConfirmationApproveAutonomous`.
- No `'b'`/`'B'` case in the raw-mode key loop (around line 108–122).
- No `"b"` case in the line-based fallback scanner (around line 135–141).
- No `case ConfirmationApproveBalanced:` in `pkg/cli/runner.go`.

Users running in non-TUI mode (`--exec`, piped stdin, the lean TUI's `s`-key Autonomous path, or any SDK integration calling `PrintToolCallWithConfirmation`) cannot select Balanced mode from the interactive confirmation prompt.

**Suggested additions** to make the CLI consistent with the TUI:
```go
// pkg/cli/printer.go — add alongside ConfirmationApproveAutonomous
ConfirmationApproveBalanced  ConfirmationResult = "approve_balanced"

// prompt text
p.Printf("\n%s", bold("Can I run this tool? ([y]es/[b]alanced/[a]ll/[n]o): "))

// raw-mode key loop — insert before the 'a', 'A' case
case 'b', 'B':
    p.Print(bold("Balanced mode 🟡"))
    return ConfirmationApproveBalanced

// line-based fallback
case "b":
    return ConfirmationApproveBalanced
```
```go
// pkg/cli/runner.go — add before the ConfirmationApproveAutonomous case
case ConfirmationApproveBalanced:
    sess.SetSafetyPolicy(session.SafetyPolicyBalanced)
    rt.Resume(ctx, runtime.ResumeApproveBalanced())
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

This PR is well-structured: the three-mode safety architecture (Strict / Balanced / Autonomous), the decoupled classifier labeller, and the custom rule precedence model (session ForceAsk beats mode; team ForceAsk is advisory and yields to Autonomous) are all implemented correctly. One functional gap was found in the CLI interaction path.

pkg/leantui/update.go
trungutt: Done in 7e1ca2a1d — added the `[b]/[B]` binding in `handleConfirmKey` and updated the leantui prompt legend to advertise it.

docker-agent: Thanks for the quick fix! Verified that 7e1ca2a adds both the / key binding in  and updates the prompt legend in  to advertise it. Balanced mode is now fully accessible from leantui. ✓

pkg/cli/printer.go
trungutt: Deferred — CLI printer / runner is out of scope for this PR. The three-mode + labeller architecture is complete without touching this surface (Balanced is reachable from the API and both TUIs). Extending the non-TUI CLI prompt is a UX call that stands on its own and will land in a follow-up so this PR stays focused on the runtime + shared TUI surface the reviewer explicitly asked me to focus on.

docker-agent: Fair call on scope — confirmed that Balanced is fully wired through the runtime (ResumeTypeApproveBalanced + SafetyPolicyBalanced) and reachable via the TUI's 'B' key in toolconfirm.go. The non-TUI CLI printer/runner (y/a/n prompt) is indeed a distinct interaction surface, and deferring it keeps this PR focused on the core three-mode architecture.

When the follow-up lands, the CLI additions will be straightforward: mirror the TUI's B-key handling (ConfirmationApproveBalanced constant + case in both raw-mode and line-based paths, plus the runner case that calls SetSafetyPolicy + Resume).

PR #3817, opened 2026-07-24
refactor(board): simplify app and controller internals

dgageot:
The `pkg/board` package had several repeated patterns that made the code harder to follow and introduced subtle maintenance hazards. Column lookup was duplicated across four methods using identical `slices.IndexFunc` calls, stored-project access was guarded by the same `if a.config.Board != nil` block in four places, and the per-card controller state was split across three parallel maps under the same mutex, which could drift out of sync.

This PR introduces two small private helpers in `app.go` — `columnIndexLocked` and `projectsLocked` — to centralize those repeated patterns, and consolidates the three per-card maps in `controller.go` (`expectTurn`, `launchFailures`, `launchErrors`) into a single `map[string]*cardState`. The `forget()` method now drops one entry instead of three, eliminating the map-drift risk entirely.

All changes are strictly behavior-preserving refactors with no functional modifications.

### DIFF
--- a/pkg/board/app.go
+++ b/pkg/board/app.go
@@ -257,6 +257,15 @@ func (a *App) saveConfigLocked() error {
 	})
 }
 
+// projectsLocked returns the stored projects, nil when the board section
+// does not exist yet. Callers must hold a.mu.
+func (a *App) projectsLocked() []userconfig.BoardProject {
+	if a.config.Board == nil {
+		return nil
+	}
+	return a.config.Board.Projects
+}
+
 // Projects returns the configured projects.
 func (a *App) Projects() []Project {
 	a.mu.Lock()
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a clean, behavior-preserving refactor. The two new helpers (`columnIndexLocked` and `projectsLocked`) correctly centralize repeated patterns without altering semantics. The consolidation of three per-card maps into a single `map[string]*cardState` simplifies `forget()` and eliminates map-drift risk without introducing any correctness issues.

Key equivalences verified:
- `AddProject` reorder: when `Board == nil`, `projectsLocked()` returns nil; the duplicate loop over nil is a no-op — same outcome as iterating an empty `Projects` slice.
- `setExpectTurn(false)`: now creates a `*cardState` entry instead of deleting; `turnExpected` correctly returns `false` via the `s != nil && s.expectTurn` check.
- `resetLaunchFailures`: setting to 0 vs deleting is functionally identical since callers only care whether the count is non-zero.
- `setLaunchError(nil)`: entry persists with a nil error; `LaunchError()` correctly returns nil either way.

trungutt (approved):

PR #3829, opened 2026-07-24
feat: three-mode safety policy (strict / balanced / autonomous) with native shell classification

trungutt:
## Why

Tool approval today is a coarse two-option choice. When a call needs confirmation, the user can approve it once — or approve *everything for the rest of the session*. Prompt fatigue pushes people to the blanket option without them realising it also covers destructive commands (`rm -rf`, `docker volume rm`, …). There is no middle ground, and no obvious way back.

The shell classifier (`safer: true`) was also entangled with the approval verdict: it read the session policy and decided whether to ask, which made its judgement impossible to reuse and every new approval behavior an edit to the classifier.

## What this PR does

Introduce **three explicit safety modes** and split the classifier into a pure labeller so the mode is what actually gates the call:

```
               safe     destructive   unknown
  strict       ask      ask           ask
  balanced     ALLOW    ask           ask
  autonomous   ALLOW    ALLOW         ALLOW
```

- **strict** — ask before every tool call, read-only ones included.
- **balanced** — the new middle tier: positively-recognised safe calls (safe-listed shell commands, read-only-annotated tools) run silently; destructive and unrecognised calls still ask.
- **autonomous** — the legacy `--yolo`: everything runs.

Every call gets a `safe` / `destructive` / `unknown` label from a new dependency-free `pkg/safety` package (shell commands via the pattern taxonomy, other tools via their MCP annotations), and the pipeline becomes:

```
preempt hooks  →  custom rules (deny/allow/ask)  →  mode × label  →  default hooks  →  ask user
```

Custom rules always win over the mode — a `deny:` rule blocks even autonomous, an `allow:` rule silences even strict. One deliberate asymmetry: session-scoped `ask:` rules (direct user intent) always prompt, while agent-author `ask:` rules yield to a user-chosen auto-approving mode.

Escalation is guided and reversible: confirmation prompts offer the next rung ([B]alanced on safe calls, approve-all on risky ones), a new `--safety` flag and the existing `PATCH /sessions/:id/safety-policy` endpoint set or undo the mode at any time, and downgrading genuinely revokes the blanket approval.

## Compatibility

- Legacy policy values (`unsafe` / `safer` / `safe-auto`) and resume verbs (`approve-session`, `approve-safe`, `approve-safer`) are accepted and normalized, so existing API consumers keep working. One deliberate nuance: `safer` maps to `balanced`, which is *narrower* — old `safer` waved unrecognised commands through silently, `balanced` prompts for them. The mapping errs toward asking; it never widens.
- Sessions that never choose a mode keep the historical default: read-only tools auto-approve, everything else asks. Explicit `strict` is stricter than that on purpose.
- The shell toolset's `safer:` flag stays parseable but is ignored — classification now always runs. YAMLs pinning the `safer_shell` builtin keep working; it degrades to a metadata-only labeller.
- Confirmation events and the `on_tool_approval_decision` hook now carry `safety_label`, so host applications can render risk tiers without re-implementing the taxonomy.
- **Deliberate behavior change:** the safety mode is session-scoped state, so resuming a session stored as `autonomous` keeps it autonomous without re-passing `--yolo`. On main, blanket approval had to be re-requested on every run. An explicit `--safety` on resume still overrides the stored mode.

Supersedes #3806 (same goal; this version keeps the decision table authoritative — no hidden read-only fast path under explicit strict — moves the classifier out of the hook lane into a runtime component, and preserves wire compatibility for configs, resume verbs, and persisted sessions).


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR introduces a well-structured three-mode safety policy (strict / balanced / autonomous) with a clean separation between the classifier and the approval decision gate. The new `pkg/safety` package, the session-scoped policy persistence, and the legacy normalization paths all look correct.

**Review scope**: 65 files, 2751 diff lines across all four chunks were analyzed. No CONFIRMED or LIKELY bugs were found in the code introduced by this PR.

**Key areas reviewed**:
- `pkg/safety/classify.go` — shell classifier, pattern matching, metacharacter guards
- `pkg/runtime/toolexec/dispatcher.go` — `handleResume`, `confirmationMetadata`, decision pipeline
- `pkg/runtime/toolexec/permissions.go` — `Decide`, mode × label table, checker tiers
- `pkg/server/session_manager.go` — `ResumeSession`, `ToggleToolApproval`, mid-turn persistence
- `pkg/session/session.go` — `SetSafetyPolicy`, `ToggleYolo`, `ToolsApproved` sync

**Notable design correctness confirmed**:
- Legacy `approve-session` → `ResumeTypeApproveAutonomous` → `SetSafetyPolicy(Autonomous)` correctly syncs `ToolsApproved=true` via the new path — backward compat maintained.
- `ToggleToolApproval` is symmetric for live vs. inactive sessions (both call `ToggleYolo()`), with correct rollback on store failure.
- `confirmationMetadata` intentionally always returns a non-nil map with `safety_label` — this is the designed behavior for surfacing risk tiers to host applications.
- The `handleResume` silent default is pre-existing and guarded upstream by `IsValidResumeType`.

pkg/tui/handlers.go
```diff
@@ -57,13 +57,16 @@ func (m *appModel) handleBranchFromEdit(msg messages.BranchFromEditMsg) (tea.Mod
...
-		return m, notification.ErrorCmd(fmt.Sprintf("Failed to save branched session: %v", err))
-	}
-
+	// Apply live-session state before the store write so mid-session
+	// changes the store copy lacks (e.g. a mode downgrade) are persisted
+	// with the branch, not just patched in memory.
 	if current := m.application.Session(); current != nil {
 		newSess.HideToolResults = current.HideToolResults
-		newSess.ToolsApproved = current.ToolsApproved
+		newSess.SetSafetyPolicy(current.GetSafetyPolicy())
```
docker-agent, line 65: **[medium] `SetSafetyPolicy` in `handleBranchFromEdit` silently clears `PriorSafetyPolicy`, breaking toggle-back on the branch**

`BranchSession` → `copySessionMetadata` correctly copies both `SafetyPolicy` and `PriorSafetyPolicy` from the parent. But the call to `newSess.SetSafetyPolicy(current.GetSafetyPolicy())` unconditionally resets `PriorSafetyPolicy = ""` (per the implementation in `pkg/session/session.go`: "An explicit choice invalidates the yolo-toggle memory").

**Trigger path:** User is in `balanced` mode, escalates mid-session to `autonomous` (which saves `PriorSafetyPolicy = "balanced"`), then branches the session. The branch receives `PriorSafetyPolicy = "balanced"` from `copySessionMetadata`, but `SetSafetyPolicy("autonomous")` immediately wipes it to `""`. The branch's `ToggleYolo` now drops to the legacy default (`""`) instead of restoring `balanced`.

**Impact:** A user who branched while in an escalated mode permanently loses the toggle-back memory on the branch. Every other code path that sets `PriorSafetyPolicy` is deliberate (either the user escalated, or `ToggleYolo` recorded the prior mode). This path erases it without user intent.

**Fix:** Copy the live session's safety fields directly without going through `SetSafetyPolicy`, so `PriorSafetyPolicy` is preserved:
```go
if current := m.application.Session(); current != nil {
    newSess.HideToolResults = current.HideToolResults
    // Copy both fields directly to preserve PriorSafetyPolicy (the
    // toggle-restore memory). SetSafetyPolicy would clear it.
    newSess.SafetyPolicy = current.GetSafetyPolicy()
    newSess.ToolsApproved = current.IsToolsApproved()
    newSess.PriorSafetyPolicy = current.GetPriorSafetyPolicy() // if such a getter exists
}
```
Alternatively, add a `CopySafetyFrom(*Session)` method on `Session` that copies both fields atomically.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/hooks/builtins/safer_shell.go
```diff
@@ -1,458 +1,59 @@
...
-			case policySafeAuto:
-				return allowWithMetadata(radiusSafe, match.Category, reason), nil
-			default:
-				return askWithMetadata(radiusSafe, match.Category, reason), nil
-			}
-		}
-	}
-	// Unknown command: safer defers, strict / safe-auto ask.
-	if policy == policySafer {
+	warnSaferShellOnce()
```
docker-agent, line 63: **[low] `warnSaferShellOnce()` fires for any `pre_tool_use` event, not only shell-tool invocations**

The deprecation warning is emitted at line 46, before the `in.ToolName != safety.ShellToolName` guard at line 47. This means the warning fires the first time `safer_shell` handles *any* tool's `pre_tool_use` event — not just `shell`. If an operator pinned `safer_shell` explicitly to a non-shell tool in their YAML, the warning fires on that tool's first invocation (which may be before any shell call). Because `sync.OnceFunc` ensures it fires exactly once per process, there is no repeated noise, but the semantic message ("safer_shell hook is deprecated") is attributed to whichever tool triggered the first `pre_tool_use` event.

If the intent is to warn only when a shell command is actually classified, move the call after the `ToolName` guard. If the intent is to warn as soon as the hook is active regardless of tool (arguably correct for a deprecation), the placement is intentional and this is not a bug.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):

#### Findings summary

- **[medium] ** —  clears  on branch, breaking toggle-back (confidence: 🟢 strong 100/100)
- **[low] ** —  placement fires before tool-name guard (confidence: 🟡 moderate 70/100)

pkg/tui/handlers.go
docker-agent, line 65: **[medium] `SetSafetyPolicy` in `handleBranchFromEdit` silently clears `PriorSafetyPolicy`, breaking toggle-back on the branch**

`BranchSession` → `copySessionMetadata` correctly copies both `SafetyPolicy` and `PriorSafetyPolicy` from the parent. But the call to `newSess.SetSafetyPolicy(current.GetSafetyPolicy())` unconditionally resets `PriorSafetyPolicy = ""` (per the implementation in `pkg/session/session.go`: "An explicit choice invalidates the yolo-toggle memory").

**Trigger path:** User is in `balanced` mode, escalates mid-session to `autonomous` (which saves `PriorSafetyPolicy = "balanced"`), then branches the session. The branch receives `PriorSafetyPolicy = "balanced"` from `copySessionMetadata`, but `SetSafetyPolicy("autonomous")` immediately wipes it to `""`. The branch's `ToggleYolo` now drops to the legacy default (`""`) instead of restoring `balanced`.

**Impact:** A user who branched while in an escalated mode permanently loses the toggle-back memory on the branch. Every other code path that sets `PriorSafetyPolicy` is deliberate (either the user escalated, or `ToggleYolo` recorded the prior mode). This path erases it without user intent.

**Fix:** Copy the live session's safety fields directly without going through `SetSafetyPolicy`, so `PriorSafetyPolicy` is preserved:
```go
if current := m.application.Session(); current != nil {
    newSess.HideToolResults = current.HideToolResults
    // Copy both fields directly to preserve PriorSafetyPolicy (the
    // toggle-restore memory). SetSafetyPolicy would clear it.
    newSess.SafetyPolicy = current.GetSafetyPolicy()
    newSess.ToolsApproved = current.IsToolsApproved()
    newSess.PriorSafetyPolicy = current.GetPriorSafetyPolicy() // if such a getter exists
}
```
Alternatively, add a `CopySafetyFrom(*Session)` method on `Session` that copies both fields atomically.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/hooks/builtins/safer_shell.go
docker-agent, line 63: **[low] `warnSaferShellOnce()` fires for any `pre_tool_use` event, not only shell-tool invocations**

The deprecation warning is emitted at line 46, before the `in.ToolName != safety.ShellToolName` guard at line 47. This means the warning fires the first time `safer_shell` handles *any* tool's `pre_tool_use` event — not just `shell`. If an operator pinned `safer_shell` explicitly to a non-shell tool in their YAML, the warning fires on that tool's first invocation (which may be before any shell call). Because `sync.OnceFunc` ensures it fires exactly once per process, there is no repeated noise, but the semantic message ("safer_shell hook is deprecated") is attributed to whichever tool triggered the first `pre_tool_use` event.

If the intent is to warn only when a shell command is actually classified, move the call after the `ToolName` guard. If the intent is to warn as soon as the hook is active regardless of tool (arguably correct for a deprecation), the placement is intentional and this is not a bug.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):

#### Findings summary

- **[medium] `pkg/tui/handlers.go:65`** — `SetSafetyPolicy` clears `PriorSafetyPolicy` on branch, breaking toggle-back (confidence: 🟢 strong 100/100)
- **[low] `pkg/hooks/builtins/safer_shell.go:46`** — `warnSaferShellOnce()` placement fires before tool-name guard (confidence: 🟡 moderate 70/100)

pkg/tui/handlers.go
trungutt: Done in `b8de977df` — went with the surgical variant: `SetSafetyPolicy` still applies the live mode (keeping `ToolsApproved` in sync), then the live toggle memory is restored via a new `GetPriorSafetyPolicy()` getter (`newSess` is not yet shared, so the direct field write is safe). Added a branch test pinning the escalate-then-branch toggle-back scenario.

trungutt: Duplicate of the sibling thread on this line — fixed in `b8de977df`, see the reply there.

pkg/hooks/builtins/safer_shell.go
trungutt: Intentional — the deprecation concerns the configuration being present (the builtin no longer enforces anything, whatever it is matched against), not any particular shell call, so warning at the hook's first firing is the earliest useful signal. Made that explicit in the comment in `c28266fbc`.

trungutt: Duplicate of the sibling thread on this line — see the reply there (intentional, clarified in `c28266fbc`).

docker-agent: Got it — thanks for the clarification. The updated comment in `c28266fbc` makes the intent clear. Closing this as intentional.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR introduces a well-structured three-mode safety policy (strict/balanced/autonomous) with clean separation between classification (pkg/safety) and gating (pkg/runtime/toolexec/permissions.go). The decision table, legacy compatibility normalization, and concurrent session-state management are all handled correctly.

**Reviewed findings summary:**
- ApprovalSourceYolo reuse for Autonomous mode: intentional, documented, preserves hook-contract stability (DISMISSED)
- legacyReadOnlyAutoApprove OriginAnnotation check: intentional per comments, no regression (DISMISSED)  
- SetSafetyPolicy/GetSafetyPolicy concurrency: properly guarded by RWMutex (DISMISSED)
- confirmationMetadata nil-guard removal: behavioral change but expected as part of the safety-label feature (low, inline)

pkg/runtime/toolexec/dispatcher.go
```diff
@@ -842,70 +877,49 @@ func (c *call) runPermissionRequestHook(ctx context.Context, runTool func() Call
...
+		preemptMeta = c.preYoloResult.Metadata
 	}
-	annotationMeta := blastRadiusFromAnnotations(c.tool.Annotations)
-	if len(annotationMeta) == 0 && len(c.tool.Metadata) == 0 && len(permissionMeta) == 0 && len(safetyMeta) == 0 {
-		return nil
-	}
-	merged := make(map[string]string, len(annotationMeta)+len(c.tool.Metadata)+len(permissionMeta)+len(safetyMeta))
-	maps.Copy(merged, annotationMeta)
+	labelMeta := c.safetyLabel().Metadata()
+	merged := make(map[string]string, len(c.tool.Metadata)+len(permissionMeta)+len(labelMeta)+len(preemptMeta))
```
docker-agent, line 897: **[low] `confirmationMetadata` no longer returns `nil` when sources are empty**

The old implementation had an explicit nil-guard that returned `nil` when no annotation metadata, tool metadata, permission metadata, or preempt-hook metadata was available:

```go
// OLD:
if len(annotationMeta) == 0 && len(c.tool.Metadata) == 0 && len(permissionMeta) == 0 && len(safetyMeta) == 0 {
    return nil
}
```

The new code removes this guard and always allocates a map, because `c.safetyLabel().Metadata()` now always contributes at least `{"safety_label": "unknown"|"safe"|"destructive"}`. This means `EmitToolCallConfirmation` is now always called with a non-nil metadata map (previously it could receive `nil`).

Any downstream consumer that tested `if metadata == nil` to skip processing will now always receive metadata with at least the `safety_label` key. This is intentional — the safety label is new feature data — but it is a **behavioral change** worth confirming that all callers of `EmitToolCallConfirmation` (e.g. hook dispatchers, event emitters, API serializers) handle a non-nil populated map correctly where they previously may have relied on `nil` as a "no metadata" sentinel.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 65/100 |

pkg/hooks/builtins/safer_shell.go
```diff
@@ -1,458 +1,61 @@
...
-
 // saferShell is the [hooks.BuiltinFunc] registered under [SaferShell].
-// See the file-level comment for the policy-aware logic. Taxonomy
-// load failure asks with blast_radius=unknown regardless of policy
-// (fail-closed), except under unsafe which stays silent.
-func saferShell(_ context.Context, in *hooks.Input, args []string) (*hooks.Output, error) {
+// Pure labeller: never blocks, never asks — it returns classification
+// metadata with no permission decision so the approval pipeline is
+// unaffected.
+func saferShell(_ context.Context, in *hooks.Input, _ []string) (*hooks.Output, error) {
```
docker-agent, line 59: **[HIGH] safer_shell builtin silently drops enforcement — security regression for operators relying on `preempt_yolo: true`**

Before this PR, `safer_shell` was a decision-emitting hook that could block destructive shell commands even when `--yolo` was active (via `preempt_yolo: true`). This PR converts it into a pure metadata labeller that **never** emits a permission decision. Any existing YAML that relied on `safer_shell` with `preempt_yolo: true` to enforce a hard block on destructive commands will silently stop enforcing — the only signal is a one-time `slog.Warn` emitted at the first tool invocation.

The risk: operators who configured `safer_shell` as a security control (e.g., in team or shared agent YAMLs) receive no error, no startup warning, and no indication at deploy time that their enforcement has been removed. The warn fires at runtime, once per process, and only when the shell tool is actually called — it will be missed in logs for long-running services or when log level is above WARN.

**Suggested mitigation in the deprecation warning**: Surface the warning at YAML load time (when the builtin is registered) rather than at first invocation, so operators catch it during deployment/startup rather than after a destructive command has already run unblocked.

Alternatively, consider keeping the enforcement semantics when `preempt_yolo: true` is present in the YAML entry, and only silently labelling when it is absent.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):

This PR introduces a well-structured three-mode safety policy (strict / balanced / autonomous) with a clean decision table in `pkg/runtime/toolexec/permissions.go`. The legacy policy normalization, session state management, and approval pipeline refactor all look correct.

One finding warrants attention: the `safer_shell` builtin's behavior change from an enforcement hook to a metadata-only labeller is intentional and documented, but the deprecation notification path (a one-time `slog.Warn` at first shell-tool invocation) may be insufficient for operators relying on it as a security control.

pkg/runtime/toolexec/dispatcher.go
```diff
@@ -842,70 +877,49 @@ func (c *call) runPermissionRequestHook(ctx context.Context, runTool func() Call
...
-		safetyMeta = c.preYoloResult.Metadata
+		preemptMeta = c.preYoloResult.Metadata
 	}
-	annotationMeta := blastRadiusFromAnnotations(c.tool.Annotations)
-	if len(annotationMeta) == 0 && len(c.tool.Metadata) == 0 && len(permissionMeta) == 0 && len(safetyMeta) == 0 {
-		return nil
-	}
-	merged := make(map[string]string, len(annotationMeta)+len(c.tool.Metadata)+len(permissionMeta)+len(safetyMeta))
-	maps.Copy(merged, annotationMeta)
+	labelMeta := c.safetyLabel().Metadata()
```
docker-agent, line 896: **[medium] `safety_label` key missing from `safetyConventionKeys` — leaks as raw metadata in the confirmation dialog**

`confirmationMetadata` now always calls `c.safetyLabel().Metadata()`, which always includes `safety_label` (the `MetaSafetyLabel` constant) as a key. The confirmation dialog's `renderMetadata` function (`pkg/tui/dialog/tool_confirmation.go`) uses `safetyConventionKeys` to suppress internal classifier keys from the raw key/value section — but `safetyConventionKeys` only lists `blast_radius`, `category`, and `reason`. The new `safety_label` key is absent from that set.

**Trigger:** Any tool call that requires a confirmation prompt. `labelMeta` now always contains `{"safety_label": "safe" | "unknown" | "destructive"}`. When `blast_radius` is also present (all shell commands, and annotation-hinted MCP tools), `renderMetadata` enters the suppression branch but does not skip `safety_label`, so it renders as a plain `safety_label: safe` (or `unknown`) line in the dialog.

**Impact:** Every tool confirmation dialog shows a spurious raw `safety_label: <value>` row in the Metadata section alongside the polished warning block. This clutters the UX with programmatic classifier state that is intended for API consumers (the PR description notes this explicitly), not end users.

**Fix:** Add `"safety_label": {}` to `safetyConventionKeys` in `pkg/tui/dialog/tool_confirmation.go`:
```go
var safetyConventionKeys = map[string]struct{}{
    "blast_radius": {},
    "category":     {},
    "reason":       {},
    "safety_label": {},   // add this
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/safety/classify.go
```diff
@@ -0,0 +1,395 @@
...
+				BlastRadius: "safe",
+				Category:    match.Category,
+				Reason:      "Command matches safe read-only pattern: " + match.Pattern,
+			}
+		}
+	}
+	return Label{
+		Class:       ClassUnknown,
+		Origin:      OriginClassifier,
+		BlastRadius: "unknown",
```
docker-agent, line 60: **[medium] `ClassUnknown` shell commands render as "⚠ Destructive command" — misleading for unclassified commands**

The classifier returns `BlastRadius: "unknown"` for any shell command it cannot positively identify as safe or destructive. This value is emitted into the confirmation event metadata via `confirmationMetadata`. The existing `renderSafetyWarning` function in `pkg/tui/dialog/tool_confirmation.go` only checks `if radius == "safe"` for a green badge; any other value (including `"unknown"`) falls into the `else` branch: `⚠ Destructive command — unknown blast radius`.

**Trigger:** The user is prompted to approve any unrecognized shell command (e.g. `npm test`, `make build`, custom scripts, or any tool not in the safe patterns list). Before this PR, `blastRadiusFromAnnotations` only set `blast_radius` for tools with explicit annotations, so unrecognized commands had no blast_radius and `renderSafetyWarning` returned `""`. Now `ClassifyCommand` always returns a label with `BlastRadius` set, so the warning block always appears.

**Impact:** Every unrecognized shell command prompts the user with "⚠ Destructive command — unknown blast radius" even when it is entirely benign. This misrepresents the command's risk level, erodes trust in the safety warnings, and could cause users to ignore or dismiss all warnings as false alarms — the opposite of the feature's intent.

**Fix:** Add a dedicated branch in `renderSafetyWarning` for `"unknown"` that renders a neutral "unrecognised" badge rather than reusing the destructive warning copy, e.g.:
```go
case radius == "safe":
    // green read-only message
case radius == "unknown":
    heading = styles.BaseStyle.Bold(true).Foreground(styles.Warning).
        Render("⚠  Command not recognised — unknown safety classification")
default:
    // orange destructive message
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/runtime/toolexec/dispatcher.go
```diff
@@ -472,13 +528,18 @@ func (c *call) autoApprovalAfterConfirmationWait() (PermissionDecision, bool) {
...
 		}
 		return PermissionDecision{}, false
 	}
-	decision := c.permissionDecision(c.tool.Annotations.ReadOnlyHint)
+	// A mode escalation or allow grant may have landed while this call
+	// was queued behind another confirmation. The legacy read-only
+	// fast path is deliberately NOT rechecked here: it is static, so a
+	// call that reached askUser despite it was explicitly asked about
+	// (a hook or ask rule), and that prompt must be honored.
+	decision := c.permissionDecision()
```
docker-agent, line 542: **[medium] `autoApprovalAfterConfirmationWait` re-evaluates mode×label after confirmation wait, potentially auto-approving Balanced+ClassSafe calls that were originally prompted by a hook**

When a tool call reaches `askUser` because a pre-tool-use hook (not the preempt-yolo hook) returned `DecisionAsk`, the call is placed in the confirmation queue. If while blocked on the queue, the session mode is escalated (e.g. user chooses [B]alanced in response to a different prompt), `autoApprovalAfterConfirmationWait` calls `permissionDecision()` and auto-approves if `OutcomeAllow` is returned.

**Trigger:**
1. Session is in default/strict mode. A `permission_request` hook is configured for the tool and returns `DecisionAsk` (not `DecisionForceAsk`).
2. The call reaches `askUser`. A second, concurrent call's confirmation prompt offers `[B]alanced` to the user; the user accepts.
3. Session mode is now `SafetyPolicyBalanced`. The original call re-evaluates via `autoApprovalAfterConfirmationWait`.
4. `permissionDecision()` → `Decide()` → no checker returns `ForceAsk` (the hook returned `Ask`, not `ForceAsk`, which only adds advisory input) → `applyMode(Balanced, ClassSafe)` → `OutcomeAllow`.
5. The call runs without the user ever seeing the hook's prompt.

**The preempt-yolo guard** (`c.preYoloResult != nil && c.preYoloResult.Decision == hooks.DecisionAsk`) only protects against preempt-yolo lane escalations — regular `permission_request` hooks that returned `Ask` via the standard hook chain have no corresponding guard here.

**Severity note:** This requires a concurrent mode escalation to occur while a hook-prompted call is queued, which is a specific race. However, the class of call affected (`permission_request` hook + `ClassSafe` label + `Balanced` mode) is not exotic.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

Two medium-severity bugs were found in the classifier-to-UI rendering path, plus one moderate-confidence security concern in the approval pipeline.

**Summary of inline findings:**
1. `safety_label` key not added to `safetyConventionKeys` — leaks as raw metadata in every tool confirmation dialog (see comment on `dispatcher.go:896`)
2. `ClassUnknown` shell commands display "⚠ Destructive command" — misleading for merely unclassified commands (see comment on `classify.go:60`)
3. `autoApprovalAfterConfirmationWait` may auto-approve Balanced+ClassSafe calls originally prompted by a hook under a concurrent mode escalation (see comment on `dispatcher.go:542`)

pkg/session/session.go
```diff
@@ -1454,30 +1508,72 @@ func (s *Session) SetPermissions(perms *PermissionsConfig) {
...
+	return s.PriorSafetyPolicy
+}
+
+// ToggleYolo flips the session between Autonomous and the mode that was
+// active before the escalation, so an explicit Balanced/Strict choice
+// survives a toggle round-trip. A session that was never escalated from
+// a named mode drops back to the legacy default (""). The method is its
+// own inverse, which callers use to roll back a toggle whose
+// persistence failed.
+func (s *Session) ToggleYolo() {
```
docker-agent, line 1562: **[medium] ToggleYolo toggle-off when PriorSafetyPolicy is empty silently demotes session to empty policy**

`ToggleYolo` is designed to be its own inverse: toggle on escalates to `Autonomous`, toggle off restores the prior mode. However, when a session starts as `Autonomous` without going through the escalation path (e.g. `--yolo` flag, `SetSafetyPolicy(Autonomous)` directly, or resuming a stored autonomous session), `PriorSafetyPolicy` is `""`. On toggle-off:

```go
s.SafetyPolicy = s.PriorSafetyPolicy.Normalize()  // "" → ""
s.PriorSafetyPolicy = ""
// ...
s.ToolsApproved = s.SafetyPolicy == SafetyPolicyAutonomous  // false
```

The session ends up with `SafetyPolicy = ""` and `ToolsApproved = false`. This is the legacy default (no explicit mode), which is functionally less permissive than the original `Autonomous` baseline — it is a silent one-way demotion. A user who starts with `--yolo`, hits the toggle off once, and hits it again will be back to `Autonomous`, but the intermediate state is a regression to the default mode with no warning.

**Suggested fix:** guard against restoring to a meaningless empty mode. Before assigning the restored policy, default it to something safe:

```suggestion
	if current == SafetyPolicyAutonomous {
		restored := s.PriorSafetyPolicy.Normalize()
		if restored == "" {
			restored = SafetyPolicyBalanced
		}
		s.SafetyPolicy = restored
		s.PriorSafetyPolicy = ""
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/tui/dialog/tool_confirmation.go
```diff
@@ -340,29 +346,61 @@ func (d *toolConfirmationDialog) handleMouseClick(msg tea.MouseClickMsg) (layout
...
+	// action key. Labels can contain uppercase letters (the always-allow
+	// label echoes the command pattern), so scanning the clicked character
+	// itself could fire the wrong action. Separator gaps are dead zones:
+	// attributing them to either neighbour would fire some action on a
+	// near-miss, and no attribution is uniformly the safer one (left
+	// makes the Y/N gap approve, right makes the B/A gap go autonomous).
+	start := 0
+	for start < plainLen {
+		textEnd := plainLen
+		if sep := strings.Index(optionsPlain[start:], "  "); sep >= 0 {
```
docker-agent, line 397: **[medium] Mouse click hit-test splits on first double-space in label, risking mis-attribution when permission pattern contains `"  "`**

The segment walker finds boundaries using `strings.Index(optionsPlain[start:], "  ")` — a hard-coded two-space delimiter. The rendered line looks like:

```
Y yes  N no  T always allow ls*  B balanced  A all tools
```

Segments are joined by `"  "` (two spaces) while each segment's key-and-label uses a single space. This works when no label itself contains two consecutive spaces, but `AlwaysAllowLabel` echoes the raw permission pattern — e.g. `"always allow docker  run"`. In that case `strings.Index` finds the double-space *inside* the T segment's label before the T→B boundary, splits the T segment early, and the right-half click on `"run"` dispatches the wrong action key (`r` → no-op), while a click on `"balanced"` is attributed to the T segment rather than B, silently triggering `ApproveTool` instead of `ApproveBalanced`.

The defect is latent under typical patterns but is directly triggered by any permission pattern that includes two consecutive spaces (e.g. a regex-style whitespace run, or a user-entered pattern with accidental double-spacing).

**Consider** using a fixed-width approach or building the segment offset table once from `helpKeysLine`'s rendering pass rather than re-scanning the plain string.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):

- [medium] pkg/session/session.go:1256 — WithSafetyPolicy then WithToolsApproved(true) leaves ToolsApproved=true inconsistent with non-Autonomous SafetyPolicy (confidence: weak 52/100): When options are applied in order WithSafetyPolicy(balanced) then WithToolsApproved(true), the raw ToolsApproved field is set to true while SafetyPolicy remains balanced — a contradictory state visible to legacy readers of ToolsApproved. In practice, the PR's own call sites treat the two paths as mutually exclusive, so this is a theoretical ordering hazard rather than a triggered bug. No realistic caller currently hits this combination.

pkg/hooks/builtins/safer_shell.go
trungutt: Done in `d7f705aad` — the composed hook config is scanned when executors are built, so the warning now fires at startup; the invocation-time warning stays as a backstop for executor-construction paths that skip the scan. I did not take the alternative (keeping enforcement when `preempt_yolo: true`): the PR's core design is that verdicts live in the deterministic pipeline (mode × label + permission rules), and a builtin that sometimes enforces and sometimes labels depending on a config flag would reintroduce exactly the entanglement this PR removes. Operators who want a non-bypassable gate have a first-class replacement: a custom `pre_tool_use` hook with an explicit decision, which the warning points at.

pkg/runtime/toolexec/dispatcher.go
trungutt: Done in `f4f3dfd66` — `safety_label` added to the suppressed convention keys, with a test pinning that it never renders as a raw pair.

pkg/safety/classify.go
trungutt: Done in `f4f3dfd66` — `unknown` now gets its own neutral heading ("Unrecognised command — not classified as safe") instead of the destructive copy, with a test asserting the destructive heading never appears for unknown.

pkg/tui/dialog/tool_confirmation.go
trungutt: Done in `f49ac5550` — took the cheaper equivalent of the offset-table suggestion: `AlwaysAllowLabel` now collapses whitespace runs (display only; the granted pattern is untouched), so the rendered row can never contain an interior `"  "` and the walker's separator stays unambiguous. This also covers patterns with newlines, which would have wrapped the row and broken row-based hit-testing outright.

pkg/runtime/toolexec/dispatcher.go
trungutt: Confirmed intentional, and audited the consumers: the TUI dialog guards with `len(metadata) == 0` (not nil), the API/event path just serializes the map, and no caller uses `nil` as a "no metadata" sentinel (the only `Metadata == nil` checks in the tree are on unrelated types). `safety_label` being always present is the feature contract — hosts render risk tiers from it without re-implementing the taxonomy.

pkg/runtime/toolexec/dispatcher.go
trungutt: By design, and consistent with the pipeline's own ordering rather than a gap in it. Under Balanced, a safe-labeled call never reaches the default hook lane in the first place — stage 2 is only consulted when the mode said Ask (`approveAndRun`). So the queued call being auto-approved after the user escalates is exactly what would have happened had it arrived a moment later; re-evaluating `Decide()` after the wait converges on that, instead of making the outcome depend on queueing accidents. The guarantees are tiered deliberately: default-lane hook asks are advisory under a user-chosen auto-approving mode (same rule as team-tier `ask:` rules), while hooks that must not be bypassable have the preempt-yolo lane — which is precisely the case that keeps its guard here.

pkg/session/session.go
trungutt: Deliberate, and pinned by tests (`TestToggleYolo_LegacyRoundTrip`, `TestToggleToolApproval_RoundTripsThroughSafetyMode`). Toggle-off means "revoke the blanket approval"; for a session that was never escalated from a named mode (plain `--yolo`), the pre-toggle state IS the legacy default with approval on, and dropping to the legacy default is the exact historical contract (`ToolsApproved=false` behaved the same before modes existed). Defaulting the restore to Balanced would silently *auto-approve* classifier-safe commands — wider than what toggle-off has ever meant, and a surprising mode the user never chose. Calling the intermediate state a "regression to default" inverts the intent: less permissive than Autonomous is what revoking is for, and the round-trip back to Autonomous is symmetric.

pkg/runtime/toolexec/dispatcher.go
docker-agent: Thanks for confirming! Sounds good — the len(metadata) == 0 guard in the TUI is the right pattern, and having safety_label always present makes it a reliable contract for hosts that render risk tiers.

The behavioral shift from nil-gated to always-present-with-label is exactly what you'd expect when introducing a feature like this. Appreciate the audit.

trungutt:
Superseded by #3835 — same change squashed into a single commit (tree verified identical, then rebased onto current main). All review threads here were addressed and resolved.

PR #3832, opened 2026-07-25
fix(server): prevent send-on-closed-channel panic in session title generation

Piyush0049:
### Description
This PR fixes a critical runtime concurrency bug in `SessionManager.RunSession` where asynchronous session title generation could attempt to send events on a closed channel, causing the server process to crash.

### Root Cause
When a session turn required title generation, `sm.generateTitle` was spawned asynchronously as a background goroutine writing title events to `streamChan`. If the conversation stream (`RunStream`) terminated early due to cancellation, completion, or error, the function returned immediately and triggered `defer close(streamChan)`. When the background title goroutine subsequently attempted to emit `runtime.SessionTitle` to `streamChan`, sending on the closed channel caused a runtime panic.

### Key Changes
* **WaitGroup Lifecycle Coordination:** Added a `sync.WaitGroup` around the asynchronous `generateTitle` goroutine and placed `wg.Wait()` before the return paths in `RunSession`. This guarantees that `close(streamChan)` cannot execute until title generation completes or safely aborts on context cancellation.
* **Graceful Stream Consumer Loop:** Replaced a direct `return` with `break` inside the `RunStream` consumer loop when encountering a context error. This ensures execution always falls through to `wg.Wait()`, guaranteeing background synchronization even during early client disconnects or stream aborts.

### Verification
* Ran `task lint` and `golangci-lint run` cleanly with zero issues reported.
* Verified that context cancellation in `generateTitle` unblocks immediately via `select` on `ctx.Done()`, preventing hangs or deadlocks.

### DIFF
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -1039,9 +1039,13 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 		defer cancel()
 		defer runtimeSession.streaming.Unlock()
 
-		// Start title generation in parallel if needed
+		// Start title generation in parallel if needed, coordinating via WaitGroup
+		// so close(streamChan) does not fire while generateTitle is still sending.
+		var wg sync.WaitGroup
 		if needsTitle {
-			go sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
+			wg.Go(func() {
+				sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
+			})
 		} else if titleToEmit != "" {
 			// Re-emit the existing title so late-joining SSE consumers
 			// and boards can pick it up without an extra API call.
@@ -1051,9 +1055,16 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 		stream := runtimeSession.runtime.RunStream(streamCtx, sess)
 		for event := range stream {
 			if streamCtx.Err() != nil {
-				return
+				break
 			}
 			streamChan <- event
+		}
+
+		// Ensure title generation finishes before defers run close(streamChan).
+		wg.Wait()
+
+		if streamCtx.Err() != nil {
+			return
 		}
 
 		if err := sm.sessionStore.UpdateSession(ctx, sess); err != nil {
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -1043,6 +1043,11 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 		// so close(streamChan) does not fire while generateTitle is still sending.
 		var wg sync.WaitGroup
 		if needsTitle {
+			// Note: generateTitle runs on the request ctx, so wg.Wait() only unblocks
+			// quickly when that context is cancelled. In the DeleteSession path
+			// (where the stream context is cancelled but the request context stays alive),
+			// the wait blocks until the title LLM call completes, delaying stream teardown.
+			// This bounded delay (first turn only) prioritizes persisting the title.
 			wg.Go(func() {
 				sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
 			})
@@ -1060,7 +1065,6 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 			streamChan <- event
 		}
 
-		// Ensure title generation finishes before defers run close(streamChan).
 		wg.Wait()
 
 		if streamCtx.Err() != nil {
--- a/pkg/environment/credential_helper_test.go
+++ b/pkg/environment/credential_helper_test.go
@@ -1,6 +1,7 @@
 package environment
 
 import (
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


aheritier:
Same here @Piyush0049 

Thanks

Piyush0049:
@aheritier This fixes a runtime panic in `docker agent serve api` when a client disconnects during a streaming session.

### Root Cause

`SessionManager.RunSession()` closes `streamChan` when the main stream goroutine exits. However, `generateTitle()` is still running in a separate goroutine and may try to send `runtime.SessionTitle` after the channel has already been closed, resulting in:

```text
panic: send on closed channel
```

### Fix

I synchronized the background `generateTitle()` goroutine using a `sync.WaitGroup` and wait for it to finish before closing `streamChan`.

All unit tests in pkg/server pass cleanly

This guarantees the channel isn't closed while it's still being written to, preventing the panic.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The concurrency fix is correctly implemented. The `sync.WaitGroup` coordinates the async `generateTitle` goroutine with the `close(streamChan)` defer, eliminating the send-on-closed-channel panic. Key points verified:

- **`wg.Go()` availability**: Go 1.26.5 (per go.mod) includes `sync.WaitGroup.Go` (introduced in Go 1.25) — compiles cleanly.
- **Defer ordering**: `wg.Wait()` is called inline (not deferred), so it blocks the goroutine before any of the LIFO defers (`streaming.Unlock`, `cancel`, `close(streamChan)`) fire — correct.
- **`return` → `break`**: The change correctly allows execution to fall through to `wg.Wait()` on context cancellation, preserving the early-exit semantics while ensuring the WaitGroup is honoured.
- **All return paths covered**: Whether the stream completes normally or context is cancelled, `wg.Wait()` is reached before the deferred `close(streamChan)`.
- **One drafter hypothesis (stream producer goroutine-leak on `break`)** was investigated and dismissed: the `break` has identical effect to the previous `return` — both abandon the range loop at the same point — so no new risk is introduced.

pkg/server/session_manager.go
```diff
@@ -891,9 +891,13 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
...
 		defer runtimeSession.streaming.Unlock()
 
-		// Start title generation in parallel if needed
+		// Start title generation in parallel if needed, coordinating via WaitGroup
+		// so close(streamChan) does not fire while generateTitle is still sending.
+		var wg sync.WaitGroup
 		if needsTitle {
-			go sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
+			wg.Go(func() {
+				sm.generateTitle(ctx, sess, titleGen, userMessages, streamChan)
```
Sayt-0, line 1052: `generateTitle` runs on the request `ctx`, so the `wg.Wait()` below only unblocks quickly when that context is cancelled. In the `DeleteSession` path (`runtimeSession.cancel` cancels `streamCtx` while the request context stays alive), the wait now blocks until the title LLM call completes, which delays `streaming.Unlock()` and `close(streamChan)`; the session answers `ErrSessionBusy` in the meantime. The delay is bounded (a single short LLM call, first turn only) and arguably acceptable, but it deserves a comment here or a note in the PR description. If fast teardown matters more than persisting the title on cancellation, passing `streamCtx` instead is the alternative.

pkg/server/session_manager.go
```diff
@@ -903,11 +907,18 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
...
 		for event := range stream {
 			if streamCtx.Err() != nil {
-				return
+				break
 			}
 			streamChan <- event
 		}
 
+		// Ensure title generation finishes before defers run close(streamChan).
+		wg.Wait()
```
Sayt-0, line 1068: A regression test is needed: this kind of goroutine/close coordination regresses silently. A test with a stubbed title generator that blocks past stream completion, combined with an agent stream that ends immediately, panics on main and passes with this patch when run under `-race`. `pkg/server/session_manager_test.go` already contains WaitGroup-based concurrency tests to model after.

pkg/server/session_manager.go
```diff
@@ -903,11 +907,18 @@ func (sm *SessionManager) RunSession(ctx context.Context, sessionID, agentFilena
 		stream := runtimeSession.runtime.RunStream(streamCtx, sess)
 		for event := range stream {
 			if streamCtx.Err() != nil {
-				return
+				break
 			}
 			streamChan <- event
 		}
 
+		// Ensure title generation finishes before defers run close(streamChan).
```
Sayt-0: Nit: this comment and the one above `var wg sync.WaitGroup` (lines 894-895) explain the same invariant; keeping only one of the two is enough.

Sayt-0 (changes requested):
The fix addresses a real panic vector, confirmed by tracing the contexts involved:

| Path | Behaviour on main |
|---|---|
| Run finishes before title generation | `close(streamChan)` runs; `generateTitle` receives the request `ctx` (not `streamCtx`) and the deferred `cancel()` only cancels `streamCtx`, so the `ctx.Done()` fallback in the emit select never fires and the send hits a closed channel (deterministic panic) |
| Client disconnect mid-run | request `ctx` cancelled, both select cases ready, the send on the closed channel can still be chosen (probabilistic panic) |

The WaitGroup coordination itself is correct:

- `wg.Wait()` runs inline, before the LIFO defers (`Unlock`, `cancel`, `close(streamChan)`).
- The `return` to `break` change makes `wg.Wait()` reachable on cancellation.
- The SSE consumer (`runAgent`) keeps draining until the channel closes, so no deadlock is introduced.
- Side benefit: on normal completion the `SessionTitle` event is now guaranteed to be delivered instead of being lost or panicking.

Two changes requested before merge:

1. A regression test reproducing the panic (details inline at `wg.Wait()`).
2. A short note on the new teardown trade-off in the `DeleteSession` path (details inline at `wg.Go`).

aheritier:

Piyush0049:
@Sayt-0 I have rebased the PR. Please do let me know if any other changes are required.

aheritier:

PR #3835, opened 2026-07-27
feat: three-mode safety policy (strict / balanced / autonomous) with native shell classification

trungutt:
## Why

Tool approval today is a coarse two-option choice. When a call needs confirmation, the user can approve it once — or approve *everything for the rest of the session*. Prompt fatigue pushes people to the blanket option without them realising it also covers destructive commands (`rm -rf`, `docker volume rm`, …). There is no middle ground, and no obvious way back.

The shell classifier (`safer: true`) was also entangled with the approval verdict: it read the session policy and decided whether to ask, which made its judgement impossible to reuse and every new approval behavior an edit to the classifier.

## What this PR does

Introduce **three explicit safety modes** and split the classifier into a pure labeller so the mode is what actually gates the call:

```
               safe     destructive   unknown
  strict       ask      ask           ask
  balanced     ALLOW    ask           ask
  autonomous   ALLOW    ALLOW         ALLOW
```

- **strict** — ask before every tool call, read-only ones included.
- **balanced** — the new middle tier: positively-recognised safe calls (safe-listed shell commands, read-only-annotated tools) run silently; destructive and unrecognised calls still ask.
- **autonomous** — the legacy `--yolo`: everything runs.

Every call gets a `safe` / `destructive` / `unknown` label from a new dependency-free `pkg/safety` package (shell commands via the pattern taxonomy, other tools via their MCP annotations), and the pipeline becomes:

```
preempt hooks  →  custom rules (deny/allow/ask)  →  mode × label  →  default hooks  →  ask user
```

Custom rules always win over the mode — a `deny:` rule blocks even autonomous, an `allow:` rule silences even strict. One deliberate asymmetry: session-scoped `ask:` rules (direct user intent) always prompt, while agent-author `ask:` rules yield to a user-chosen auto-approving mode.

Escalation is guided and reversible: confirmation prompts offer the next rung ([B]alanced on safe calls, approve-all on risky ones), a new `--safety` flag and the existing `PATCH /sessions/:id/safety-policy` endpoint set or undo the mode at any time, and downgrading genuinely revokes the blanket approval.

## Compatibility

- Legacy policy values (`unsafe` / `safer` / `safe-auto`) and resume verbs (`approve-session`, `approve-safe`, `approve-safer`) are accepted and normalized, so existing API consumers keep working. One deliberate nuance: `safer` maps to `balanced`, which is *narrower* — old `safer` waved unrecognised commands through silently, `balanced` prompts for them. The mapping errs toward asking; it never widens.
- Sessions that never choose a mode keep the historical default: read-only tools auto-approve, everything else asks. Explicit `strict` is stricter than that on purpose.
- The shell toolset's `safer:` flag stays parseable but is ignored — classification now always runs. YAMLs pinning the `safer_shell` builtin keep working; it degrades to a metadata-only labeller.
- Confirmation events and the `on_tool_approval_decision` hook now carry `safety_label`, so host applications can render risk tiers without re-implementing the taxonomy.
- **Deliberate behavior change:** the safety mode is session-scoped state, so resuming a session stored as `autonomous` keeps it autonomous without re-passing `--yolo`. On main, blanket approval had to be re-requested on every run. An explicit `--safety` on resume still overrides the stored mode.

Supersedes #3806 (same goal; this version keeps the decision table authoritative — no hidden read-only fast path under explicit strict — moves the classifier out of the hook lane into a runtime component, and preserves wire compatibility for configs, resume verbs, and persisted sessions).


docker-agent (commented):
### Assessment: 🟢 APPROVE

The three-mode safety policy implementation (strict / balanced / autonomous) is well-structured. The new `pkg/safety` package provides a clean, dependency-free classifier with thread-safe pattern compilation via `sync.OnceValues`. Legacy resume verbs (`approve-session`, `approve-safe`, `approve-safer`) are properly normalized through `NormalizeResumeType` before the policy switch, so backward compatibility is maintained. The `safer_shell` deprecation path is intentional per its documentation. No bugs introduced by this PR were confirmed.

pkg/cli/printer.go
```diff
@@ -134,6 +142,8 @@ func (p *Printer) PrintToolCallWithConfirmation(ctx context.Context, toolCall to
 	switch text {
 	case "y":
 		return ConfirmationApprove
+	case "b":
```
docker-agent, line 145: **[low] Uppercase `"B"` in the non-TTY fallback text-input path silently rejects instead of switching to balanced mode**

The raw-terminal branch correctly handles both `'b'` and `'B'`:
```go
case 'b', 'B':
    return ConfirmationApproveBalanced
```

But the Enter-based fallback only matches the lowercase string:
```go
case "b":
    return ConfirmationApproveBalanced
// "B" falls through to: default: return ConfirmationReject
```

A user who types `B` (uppercase) when prompted via the fallback path gets a silent rejection instead of the balanced-mode upgrade they requested. This inconsistency was introduced by this PR — the other pre-existing cases (`"y"`, `"a"`, `"n"`) have the same issue, but `"b"`/`"B"` is new here.

```suggestion
	case "b", "B":
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/tui/dialog/tool_confirmation.go
```diff
@@ -281,6 +289,11 @@ func (d *toolConfirmationDialog) executeAction(decision toolconfirm.Decision) (l
 			core.CmdHandler(CloseDialogMsg{}),
 			core.CmdHandler(RuntimeResumeMsg{Request: toolconfirm.ApproveTool.Resume(d.permissionPattern, "")}),
 		)
+	case toolconfirm.ApproveBalanced:
+		return d, tea.Sequence(
+			core.CmdHandler(CloseDialogMsg{}),
+			core.CmdHandler(RuntimeResumeMsg{Request: toolconfirm.ApproveBalanced.Resume("", "")}),
+		)
```
docker-agent, line 296: **[medium] `ApproveBalanced` handler does not update the TUI yolo indicator — can remain stale when downgrading from autonomous**

`ApproveSession` correctly calls `d.sessionState.SetYoloMode(true)` to sync the local TUI indicator:
```go
case toolconfirm.ApproveSession:
    d.sessionState.SetYoloMode(true)  // keeps TUI header in sync
    return d, tea.Sequence(...)
```

`ApproveBalanced` omits the symmetric call:
```go
case toolconfirm.ApproveBalanced:
    return d, tea.Sequence(
        core.CmdHandler(CloseDialogMsg{}),
        core.CmdHandler(RuntimeResumeMsg{Request: toolconfirm.ApproveBalanced.Resume("", "")}),
    )
    // ← no SetYoloMode(false) here
```

`SessionState.yoloMode` is only initialised from `IsToolsApproved()` at load time and mutated via explicit `SetYoloMode` calls — there is no reactive refresh wired to the resume flow. If a session is currently in autonomous mode (`yoloMode = true`) and the user presses `B` to downgrade to balanced, the server-side session is correctly updated, but the TUI yolo indicator stays `true` until the user restarts, making the UI mislead them into thinking they are still in autonomous mode.

```suggestion
	case toolconfirm.ApproveBalanced:
		d.sessionState.SetYoloMode(false)
		return d, tea.Sequence(
			core.CmdHandler(CloseDialogMsg{}),
			core.CmdHandler(RuntimeResumeMsg{Request: toolconfirm.ApproveBalanced.Resume("", "")}),
		)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

- pkg/runtime/toolexec/permissions.go:97 — Team-tier `ask:` rules silently bypassed by Balanced/Autonomous modes (verifier mitigation: documented intentional design — user-chosen auto-approve mode is explicitly higher priority than agent-author advisory asks; use `deny:` for hard enforcement)

pkg/safety/classify.go
```diff
@@ -0,0 +1,395 @@
...
+// or substitute (backticks, `$(`) commands — with or without
+// surrounding whitespace, so `grep foo|rm -rf /` is caught just like
+// `grep foo | rm -rf /`. The safe list must never vouch for such a
+// string: a safe-looking prefix says nothing about what the rest does,
+// and trailing-wildcard patterns (`grep ...`) would otherwise cover the
+// injected tail. Erring toward "not safe" only costs a confirmation
+// prompt. Deliberately the same strictness as the runtime's
+// session-grant check for shell commands.
+func containsShellMetacharacter(command string) bool {
+	return strings.ContainsAny(command, ";&|<>`\n") || strings.Contains(command, "$(")
```
docker-agent, line 218: **[medium] `containsShellMetacharacter` rejects quoted pipe characters, misclassifying safe grep/awk/sed patterns as unknown**

The new `containsShellMetacharacter` checks for a bare `|` anywhere in the raw command string via `strings.ContainsAny(command, ";&|<>` + "`" + `\n")`. The previous `containsShellSeparator` (in `safer_shell.go`) required the pipe to be **surrounded by spaces** — it checked `strings.Contains(" "+normalized+" ", " | ")`.

This means `grep 'foo|bar' file.txt` (a regex alternation inside a single-quoted argument) is now rejected because the `|` character appears in the string, even though it cannot act as a shell pipe in this position. The old check would not have flagged it.

**Trigger:** Any shell command whose arguments contain `|`, `<`, `>`, or `` ` `` — common in regex patterns passed to `grep`, `awk`, `sed`, and `find`.

**Impact under Balanced mode:** These commands receive `ClassUnknown` instead of `ClassSafe`. `applyMode` maps `Balanced × Unknown → ask`, so a confirmation prompt appears for every `grep 'foo|bar'` invocation. The PR description promises "positively-recognised safe calls (safe-listed shell commands…) run silently" under Balanced — but this class of common commands is silently removed from that guarantee, reintroducing the prompt fatigue the PR aims to reduce.

**Suggested fix:** Check for shell metacharacters only when they appear as bare tokens (surrounded by whitespace, start/end of string, or outside quoted regions). A minimal improvement would be to also accept the old whitespace-surrounding heuristic:
```suggestion
func containsShellMetacharacter(command string) bool {
	// Bare metacharacters that chain, pipe, or redirect commands: check
	// for the character surrounded by word boundaries so a `|` inside a
	// quoted regex argument (`grep 'foo|bar'`) doesn't veto the safe list.
	// For ; and &, any occurrence is a concern (they rarely appear as
	// literal argument text without quoting that the shell would strip).
	if strings.ContainsAny(command, ";&`\n") || strings.Contains(command, "$(") {
		return true
	}
	// Pipe and redirects: only flag as metacharacters when space-separated
	// (matching the old containsShellSeparator heuristic) so quoted regex
	// patterns like `grep 'foo|bar'` don't get classified as unknown.
	for _, sep := range []string{" | ", " < ", " > "} {
		if strings.Contains(command, sep) {
			return true
		}
	}
	return false
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

This PR is well-structured: the three-mode safety policy (strict/balanced/autonomous), the new `pkg/safety` classifier, the legacy normalization, the tier-based checker logic, and the session state management are all architecturally sound. One behavioral regression was found in the shell metacharacter classifier.

#### Findings summary
- 1 medium-severity confirmed finding (inline comment)

docker-agent (commented):
### Assessment: 🟢 APPROVE

The three-mode safety policy (strict / balanced / autonomous) is well-implemented. The (mode × label) decision table in `pkg/runtime/toolexec/permissions.go` is correct and exhaustively tested. The classifier in `pkg/safety/` is conservative by design (unknown = ClassUnknown, compound shell commands blocked, deny-flag list). Session state transitions are mutex-guarded. Legacy wire compatibility (`unsafe`/`safer`/`safe-auto` normalization) is correctly implemented throughout the pipeline.

The permission pipeline change — where team-tier `ForceAsk` rules yield to a user-chosen Autonomous mode while session-tier `ForceAsk` always prompts — is consistent with the pre-PR behavior (old yolo bypassed all ForceAsk), and the PR actually improves it by preserving session-tier ForceAsk even under Autonomous. This design decision is correctly documented in code comments.

gtardif (approved):

Sayt-0:
Don't merge yet there are still a few bugs. I am currently correcting it

PR #3843, opened 2026-07-27
TUI - Program-scoped animation runtime

krissetto:
First of many TUI improvement proposals :) 

This PR sets some animation-related foundations for enabling a better overall TUI UX where we can:
- centralize spinners, durations, transitions, effects, etc for easier re-use
- make time based animations and effects, so frame rate changes or system load doesn't affect how long animations run
- start introducing more advanced effects more easily and robustly, e.g. fades, transitions, etc
- improve TUI re-rendering logic to reduce cpu usage
- facilitate integration of docker agent components into other pre-existing TUIs

### Why

The current global coordinator advances animations by frame count and does not own queued ticks. Under event-loop delays, animations slow down; stale or duplicate tick chains also cannot be rejected cleanly.

### What changed

- Adds one program-owned animation runtime with a single shared tick lease.
- Rejects stale, duplicate, and foreign ticks; stopping invalidates queued work.
- Drives transitions and spinners from elapsed duration, so missed frames do not change animation speed.
- Preserves the package facade during partial migration: delivered legacy ticks re-arm without `Runtime.Accept`, and `IsCurrentGen` remains a side-effect-free generation check. This matches some consumer's direct use of `TickMsg`, `StartTick*`, `HasActive`, and `IsCurrentGen`.

Component adoption of the new infra is for later PRs. The legacy facade can be removed after components and consumers have migrated.

```mermaid
flowchart LR
    Root["TUI root"] --> AnimationRuntime["Program animation runtime"]
    AnimationRuntime -->|shared tick| Components
    Components -->|subscribe while active| AnimationRuntime
```

```mermaid
flowchart LR
    Delivery["Tick delivery"] --> Clock["Elapsed time"]
    Clock --> Transition["progress = elapsed / duration"]
    Clock --> Spinner["frame = elapsed / cadence"]
```

### DIFF
--- a/pkg/tui/animation/architecture_test.go
+++ b/pkg/tui/animation/architecture_test.go
@@ -1,0 +1,40 @@
+package animation
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+func TestSingleGlobalTimingSourceArchitecture(t *testing.T) {
+	_, file, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("locate animation package")
+	}
+	root := filepath.Dir(file)
+	forbidden := []string{"SubscribeEvery", "NewSubscriptionEvery", "startEvery", "registerEvery", "unregisterEvery", "ScheduledFor"}
+	err := filepath.WalkDir(root, func(path string, entry os.DirEntry, err error) error {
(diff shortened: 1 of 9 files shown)


aheritier:

Sayt-0 (approved):

PR #3851, opened 2026-07-27
TUI - enable segmented markdown rendering

krissetto:
This change lets streaming consumers receive Markdown output as two sections: completed top-level blocks that can be retained, and the trailing block that may still change as new text arrives.  

The existing renderer already avoided reparsing completed blocks. This change will enable us to avoid repeatedly joining, scanning, splitting, and styling the full rendered response.

The renderer remains conservative around unfinished or context-sensitive Markdown, and tests compare every intermediate result with the result of rendering the complete accumulated response from scratch.

This PR enables a subsequent change, where transcript and viewport code retain the completed rendered lines and replace only the changing tail. It will not bring any noticeable performance improvements on its own

### DIFF
--- a/pkg/tui/components/markdown/incremental.go
+++ b/pkg/tui/components/markdown/incremental.go
@@ -37,7 +37,19 @@ type IncrementalRenderer struct {
 	// fallback is used for the actual rendering work; it is reused across calls
 	// so its parser pool (and chroma caches) stay warm.
 	fallback *FastRenderer
+
+	stats IncrementalStats
 }
+
+// IncrementalStats reports deterministic parser work for scaling tests.
+type IncrementalStats struct {
+	Calls       uint64
+	FullRenders uint64
+	ParsedBytes uint64
+}
+
+// Stats returns cumulative work counters.
+func (r *IncrementalRenderer) Stats() IncrementalStats { return r.stats }
 
 // NewIncrementalRenderer creates a new incremental renderer with the given
 // terminal width.
@@ -56,22 +68,49 @@ func (r *IncrementalRenderer) Render(input string) (string, error) {
 	return out, err
 }
 
+// RenderedParts exposes the reusable rendered block prefix separately from the
+// still-mutable trailing block. Consumers that retain lines can splice only
+// MutableTail while StablePrefix grows monotonically during streaming.
+type RenderedParts struct {
+	StablePrefix string
+	MutableTail  string
+	CodeBlocks   []CodeBlock
+}
+
+// RenderParts renders input without concatenating the stable rendered prefix
+// to the mutable tail. This avoids rescanning/copying an ever-growing ANSI
+// string in viewport-oriented consumers.
+func (r *IncrementalRenderer) RenderParts(input string) (RenderedParts, error) {
+	stable, tail, blocks, err := r.renderParts(input)
+	return RenderedParts{StablePrefix: stable, MutableTail: tail, CodeBlocks: blocks}, err
+}
+
 // RenderWithCodeBlocks behaves like Render but additionally returns the list
 // of fenced code blocks in the rendered output. Each entry's Line is the
 // 0-indexed line within the returned string where the block's copy label is
 // drawn.
 func (r *IncrementalRenderer) RenderWithCodeBlocks(input string) (string, []CodeBlock, error) {
+	stable, tail, blocks, err := r.renderParts(input)
+	if err != nil {
+		return "", nil, err
+	}
+	return r.joinPrefixAndTail(stable, tail), blocks, nil
+}
+
+func (r *IncrementalRenderer) renderParts(input string) (string, string, []CodeBlock, error) {
+	r.stats.Calls++
 	if input == "" {
 		r.inputPrefix = ""
 		r.outputPrefix = ""
 		r.codeBlocksPrefix = nil
-		return "", nil, nil
+		return "", "", nil, nil
 	}
 
 	// If the new input no longer starts with our cached prefix, the user (or a
 	// retry) replaced earlier content; fall back to a full render.
 	if r.inputPrefix == "" || !strings.HasPrefix(input, r.inputPrefix) {
-		return r.fullRender(input)
+		stable, tail, blocks, err := r.fullRenderParts(input)
+		return stable, tail, blocks, err
 	}
 
 	// Locate a fresh stable boundary anywhere up to the end of the current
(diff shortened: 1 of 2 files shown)


pkg/tui/components/markdown/incremental.go
```diff
@@ -145,39 +205,119 @@ func (r *IncrementalRenderer) Reset() {
...
+			break
+		}
+		lineEnd := cut + space
+		line := strings.TrimSpace(remaining[cut:lineEnd])
+		if line == "" {
+			break
+		}
+		renderedLine, err := r.fallback.Render(line)
+		if err != nil {
+			stable, tail, blocks, fallbackErr := r.fullRenderParts(input)
```
docker-agent: **[medium] `renderPlainParagraph` mid-loop fallback leaves `plainOutputPrefix` partially mutated**

When the inner `r.fallback.Render(line)` call succeeds for one or more iterations, `r.plainOutputPrefix` is updated (lines 303–306). If a subsequent iteration's `Render` call returns an error, the code falls back to `r.fullRenderParts(input)` and returns early. `fullRenderParts` resets the block-level state (`inputPrefix`, `outputPrefix`, `codeBlocksPrefix`) but does **not** touch `r.plainInputPrefix` or `r.plainOutputPrefix`.

On the next call to `renderParts`, if the input is still plain ASCII, `isPlainStreamingParagraph` returns `true` again and `renderPlainParagraph` is re-entered. The guard at line 278 (`strings.HasPrefix(input, r.plainInputPrefix)`) passes — because `r.plainInputPrefix` was not advanced inside the loop, so it still correctly describes a prefix of the new input. However, `r.plainOutputPrefix` already contains rendered lines from the **previous** partial run. The function then appends more rendered lines on top, producing duplicated or mismatched output.

**Trigger path**: input is a long plain paragraph (> 2× width), `r.fallback.Render` succeeds on at least one iteration then fails on a later one. Rare in practice (render errors are exceptional), but the state machine is left inconsistent.

**Fix**: before calling `r.fullRenderParts` in the mid-loop error path (and the post-loop path at line 317), reset the plain prefix state:
```suggestion
			r.plainOutputPrefix = ""
			r.plainInputPrefix = ""
			stable, tail, blocks, fallbackErr := r.fullRenderParts(input)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

pkg/tui/components/markdown/incremental.go
```diff
@@ -145,39 +205,119 @@ func (r *IncrementalRenderer) Reset() {
...
+func (r *IncrementalRenderer) joinRenderedParts(prefix, tail string, continuation bool) string {
+	if !continuation {
+		return r.joinPrefixAndTail(prefix, tail)
+	}
+	if prefix == "" {
+		return tail
+	}
+	if tail == "" {
+		return prefix
+	}
```
docker-agent: **[medium] `joinRenderedParts` continuation join uses `"\n"` — may diverge from `FastRenderer.Render` of the same text**

When `continuation=true` (returned only from `renderPlainParagraph` when `plainOutputPrefix != ""`), `joinRenderedParts` joins prefix and tail with a single `"\n"`. The plain-paragraph path renders each completed wrapped line by calling `r.fallback.Render(line)` independently. Because `FastRenderer.renderParagraph` writes `wrapped + "\n\n"` and `finalizeOutput` pads every line to `width` with trailing spaces, a line rendered in isolation (e.g., `"word1 word2"`) receives its own width-padding. The tail is rendered similarly.

The join `plainOutputPrefix + "\n" + renderedTail` therefore concatenates separately-padded lines. By contrast, `FastRenderer.Render(fullInput)` renders the whole paragraph as one block, with `wrapText` choosing the wrapping boundaries over the full text and `finalizeOutput` padding each resulting line. If `renderPlainParagraph`'s greedy word-split boundaries happen to differ from `wrapText`'s — for example, because `wrapText` skips multiple spaces at once while `renderPlainParagraph` advances `cut` past each space individually — the output will diverge from a full render.

The test `TestIncrementalPlainParagraphScalingBound` is designed to catch this via `require.Equal(t, expected, got)` at every step. If the test actually passes in CI, the two algorithms happen to agree for the test's `chunk` string. Consider verifying coverage with inputs whose word-boundary decisions differ between the two algorithms (e.g., text with multiple consecutive spaces, or words whose lengths happen to straddle the width boundary differently depending on context).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

- [low] pkg/tui/components/markdown/incremental.go:317 — renderPlainParagraph tail-render fallback leaves plainInputPrefix advanced while block-level state reflects full render (confidence: weak 52/100)

Sayt-0 (dismissed):

pkg/tui/components/markdown/incremental.go
```diff
@@ -80,16 +119,17 @@ func (r *IncrementalRenderer) RenderWithCodeBlocks(input string) (string, []Code
 	// tail may have been an unfinished paragraph that should now be part of a
 	// completed block.
 	tail := input[len(r.inputPrefix):]
+	r.stats.ParsedBytes += uint64(len(tail))
```
docker-agent: **[low] Double-counting of `ParsedBytes` on error-fallback path**

When `renderParts` takes the incremental path (a valid cached prefix exists), it unconditionally increments `r.stats.ParsedBytes += uint64(len(tail))` at line 122. However, if any of the three subsequent `r.fallback.RenderWithCodeBlocks(...)` calls returns an error, the code falls back to `fullRenderParts(input)`, which at line 184 increments `r.stats.ParsedBytes += uint64(len(input))` again.

Since `len(input) = len(r.inputPrefix) + len(tail)`, the error path counts `len(tail) + len(input)` bytes instead of just `len(input)`, inflating `ParsedBytes` by `len(tail) + len(r.inputPrefix)` per fallback occurrence.

This only corrupts the stats counter — rendered output is unaffected. But `TestIncrementalRendererScalingBound` asserts `float64(stats.ParsedBytes)/float64(input.Len()) <= 3.5`; if a render error triggers the fallback repeatedly the ratio could breach the bound, causing a test failure that is hard to diagnose.

**Fix**: move the `ParsedBytes` increment inside the success branches (after the `RenderWithCodeBlocks` call succeeds), or subtract `len(tail)` before delegating to `fullRenderParts` in each error branch:

```suggestion
	tail := input[len(r.inputPrefix):]
	boundary := stableBoundary(tail)
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The segmented rendering design is sound and well-tested. One minor stats-accounting issue was found in the error-fallback path.

dgageot (approved):

PR #3853, opened 2026-07-27
feat: add first-class host UX for plan management (#3844)

### LINKED ISSUE #3844: feat: add first-class host UX for plan management
## Problem / context

docker-agent now has plan primitives, but plan management is still primarily an agent/tool concern. A human using the host TUI or CLI cannot reliably discover, inspect, switch, approve, resume, or export plans as a first-class workflow. This makes plans hard to use across turns and frontends, and leaves host applications to reconstruct state from tool events or files.

This is a **host-UX enhancement**, not a replacement for the existing plan toolsets or their storage work. The goal is to expose the same plan lifecycle through stable host surfaces while preserving the existing agent-facing collaboration model.

Relevant history and prior art:

- [#2788](https://github.com/docker/docker-agent/issues/2788) proposed `/plan`, approval UX, and plan/execution separation; it was closed after implementation direction moved to readonly planner sub-agents and dedicated plan tools.
- [#3227](https://github.com/docker/docker-agent/pull/3227) / [#3237](https://github.com/docker/docker-agent/issues/3237) introduced the shared `plan` toolset and its pluggable storage direction.
- [#3239](https://github.com/docker/docker-agent/pull/3239) added the storage interface.
- [#3263](https://github.com/docker/docker-agent/issues/3263) covered file-based revisions, export, free-form status, and optimistic locking.
- [#3274](https://github.com/docker/docker-agent/pull/3274) implemented revisions/status/locking work.
- [#3292](https://github.com/docker/docker-agent/pull/3292) and [#3305](https://github.com/docker/docker-agent/pull/3305) established the per-session markdown plan and `exit_plan_mode` marker alongside the shared plan toolset.
- [#3140](https://github.com/docker/docker-agent/pull/3140) and [#3168](https://github.com/docker/docker-agent/pull/3168) are related plan-mode/persona history.

Those changes are complementary. This issue addresses the missing host-facing management layer and must not duplicate or regress the shared, cross-session plan capability.

## Goals

- Make plans discoverable and manageable from the TUI without requiring knowledge of tool names or storage paths.
- Provide a scriptable CLI surface with stable, machine-readable output.
- Reuse the existing plan/session-plan abstractions and storage/concurrency semantics rather than creating a second plan database.
- Make status, revision, and conflict information visible to users.
- Keep host UX usable for both shared named plans and the current session plan where supported.
- Define one host contract that future API/frontends can consume.

## Non-goals

- Replacing `plan` or `session_plan` toolsets, their storage backends, or agent prompts.
- Automatically approving plans or turning every tool call into an approval prompt.
- Designing a new project/task tracker, dependency graph, or GitHub integration.
- Changing plan content semantics, free-form status semantics, or optimistic-locking rules already established by the plan implementation.
- Requiring a TUI to use the CLI internally, or making headless execution interactive.

## User stories

- As a TUI user, I can list plans, see status/revision/updated time, and open one without leaving the session.
- As a TUI user, I can create, rename/select, edit, export, archive/delete, and resume a plan with clear confirmation for destructive actions.
- As an agent operator, I can see when a plan is awaiting review, in progress, blocked, done, or conflicted.
- As a CLI/script user, I can list and inspect plans as JSON, update status/content with an expected revision, and receive a non-zero conflict result when stale.
- As a headless caller, I can use the same operations without TUI-only assumptions or ANSI output.
- As a team using shared plans, I can safely observe and update plans across sessions without silent last-write-wins overwrites.

### TUI

Add a discoverable plan browser (for example `/plans`) and a plan detail/editor flow. Exact key bindings are a decision below, but the interaction should support list, filter/open, create, edit, status update, export, and delete/archive. The active plan and status should be visible without obscuring the conversation.

Illustrative rendering:

```
╭─ Plans ──────────────────────────────────────────────────────────────╮
│ > release-prep       in-progress   v7   updated 2m ago              │
│   api-migration      blocked       v3   updated yesterday            │
│   session plan       awaiting-review                                  │
│                                                                     │
│ [Enter] Open  [n] New  [e] Edit  [s] Status  [x] Export  [d] Delete │
╰─────────────────────────────────────────────────────────────────────╯
```

```
╭─ Plan: release-prep · in-progress · v7 ─────────────────────────────╮
│  1. Validate the release configuration                               │
│  2. Run the compatibility test matrix                               │
│  3. Publish the signed artifacts                                     │
│                                                                     │
│ Last updated by session abc123.  Optimistic-lock revision: 7         │
│ [Edit] [Set status] [Export] [Refresh] [Back]                       │
╰─────────────────────────────────────────────────────────────────────╯
```

For a stale update, show an actionable conflict (refresh/review/ retry) and never silently overwrite the newer revision. Destructive operations require confirmation and should identify the plan name.

### CLI

Introduce a command group under the existing CLI convention (exact command name is a decision):

```
docker agent plans list [--json]
docker agent plans get <name> [--json]
docker agent plans create <name> [--file <path>]
docker agent plans update <name> [--file <path>] [--expected-version <n>]
docker agent plans status <name> <status> [--expected-version <n>]
docker agent plans export <name> --output <path>
docker agent plans delete <name> [--force]
```

Human output is concise and readable; `--json` has stable fields (name/id, content or content reference, status, version, timestamps, and storage/session identity as applicable). Errors, including optimistic-lock conflicts, are machine-detectable and include the current version when safe.

## Architecture and constraints

- Define a host-facing service/interface around the existing shared-plan and session-plan implementations; TUI and CLI must call that interface rather than duplicate filesystem or storage logic.
- Preserve pluggable storage. Do not assume local files in the host layer; storage errors must remain distinguishable from not-found and conflict errors.
- Preserve atomic writes, validation, revisions, and optimistic locking. Every mutating operation that can race accepts an expected version or has an explicit safe replacement policy.
- Keep shared named plans and ephemeral per-session plans distinct in the model and UX; never expose a session plan as if it were globally shareable.
- Keep TUI rendering/event handling separate from plan persistence. Refresh after writes and surface update events where available; do not rely on polling as the only consistency mechanism.
- CLI output must not depend on TUI state, terminal capabilities, or an active interactive session.
- Follow existing config/schema and command-registration patterns; no new config is required for basic management.
- Define behavior for corrupt/unreadable plans, stale session plans, missing storage, concurrent deletion, and unsupported operations before implementation.

## Phased delivery

1. **Host contract and read-only discovery:** model/errors, list/get, status/version metadata, TUI browser, and CLI list/get with JSON.
2. **Safe mutations:** create/update/status/export/delete/archive, TUI actions, CLI flags, expected-version conflicts, confirmations, and end-to-end storage tests.
3. **Session-plan integration:** show the active session plan, connect `exit_plan_mode`/plan-updated events to host state, and make shared versus session scope explicit.
4. **Polish and adoption:** accessibility/keyboard review, documentation/examples, telemetry or diagnostics if justified, and rollout behind an opt-in flag only if compatibility findings require it.

## Acceptance criteria

- [ ] A user can discover and open existing shared plans from the TUI without inspecting the data directory or invoking an agent tool.
- [ ] TUI list/detail views show plan identity, scope, status, version/revision, and last-update information; long content is safely truncated or paginated.
- [ ] TUI supports the agreed create/edit/status/export/delete-or-archive actions with confirmation for destructive actions.
- [ ] The CLI provides list/get and all delivered mutating operations with documented human output and stable `--json` output.
- [ ] Shared and session plans are clearly distinguished, and unsupported cross-scope operations fail with an actionable error.
- [ ] Concurrent updates using a stale expected version fail deterministically, preserve the newer content, and report the current version; no silent overwrite occurs.
- [ ] Not-found, corrupt/unreadable, storage, validation, and conflict errors are distinct in the host contract and covered by tests.
- [ ] TUI and CLI tests cover empty lists, long content, invalid names/statuses, delete confirmation, refresh-after-write, and headless/non-TTY execution.
- [ ] Existing plan toolset behavior and existing storage backends remain backward compatible; historical shared-plan and session-plan tests continue to pass.
- [ ] Documentation includes examples for interactive and scripted workflows and explicitly links this feature to the historical work above.

## Risks and rollout

- **Duplicate abstractions:** a host-only store could diverge from agent tools. Mitigate by reusing the existing interfaces and adding contract tests shared by TUI/CLI.
- **Accidental destructive edits:** require confirmations, show scope/version, and default to safe non-destructive actions.
- **Concurrent writers:** enforce expected-version checks and provide refresh/retry UX.
- **Scope confusion:** label shared versus session plans in every relevant view and command.
- **Storage/backend differences:** capability detection and precise errors are preferable to pretending every backend supports every operation.
- **Breaking scripts:** introduce the CLI incrementally, document JSON schema/versioning, and avoid changing existing tool output. Roll out read-only discovery first, then mutations, with an opt-in flag only if needed.

## Decisions and open questions before implementation

1. What exact CLI namespace and command names fit the existing `docker agent` command taxonomy?
2. Is `/plans` the right TUI entry point, and how should it coexist with any existing plan/session slash commands?
3. Should delete be permanent, or should the host UX use archive/restore while retaining the existing delete tool semantics?
4. What is the canonical identity for a session plan, and which session-plan operations are safe to expose after the session ends?
5. Which storage capabilities are mandatory for phase 1, and how should remote/custom backends advertise unsupported operations?
6. Is status validation intentionally completely free-form, or should the host provide suggestions without enforcing an enum?
7. What JSON schema/versioning and exit-code contract should scripts rely on?
8. Which plan-updated/conflict events already exist, and what event payload is needed for a live TUI refresh?
9. Should plan content editing be an embedded TUI editor, an external `$EDITOR`, or both?
10. Are any operations permitted while a plan is actively being written by an agent, and how is that state represented?
11. Which accessibility, color, and keyboard-navigation requirements are mandatory for the browser/editor?
12. Do we need a feature flag for the first release, or can read-only host management ship directly while mutations follow in later phases?

Sayt-0:
## Summary

Closes #3844.

Adds a host-facing plan management contract used by a scriptable CLI and the TUI while preserving the existing `plan` and `session_plan` toolsets.

### Shared plan storage and notifications

- Reuses the existing pluggable `plan.Storage` backend through `plan.SharedStorage()`.
- Adds portable cross-process serialization for filesystem mutations:
  - `flock` on Unix;
  - `LockFileEx` on Windows;
  - documented in-process fallback where no OS lock primitive is available.
- Keeps reads lock-free and safe through atomic rename publication.
- Makes guarded cross-process races deterministic: one writer succeeds, stale writers receive the current revision.
- Bounds plan content at 10 MiB and safely rejects oversized, corrupt, non-regular, FIFO, and device-backed files.
- Replaces the former single callback slot with a race-safe multi-subscriber `ChangeNotifier`, registered once per runtime stream and released before its event channel closes.

### `pkg/plans` host contract

A shared service interface for CLI and TUI:

- explicit `shared` and `session` scopes;
- list, get, create, update, status, export, and delete operations;
- typed not-found, validation, corrupt, storage, conflict, and unsupported errors;
- create-only and update-only semantics;
- expected-version optimistic locking;
- read-only session-plan integration with bounded reads;
- stable host JSON fields using `snake_case`.

### `docker agent plans`

```text
plans list [--json] [--session <id>]
plans get [<name>] [--scope shared|session] [--session <id>]
plans create <name> --file <path> [--title|--author|--status] [--json]
plans update <name> --file <path> [--expected-version <n>|--force] [--json]
plans status <name> <status> [--expected-version <n>|--force] [--json]
plans export [<name>] --output <path> [--json]
plans delete [<name>] [--expected-version <n>|--force] [--json]
```

- schema-versioned JSON success and error documents;
- conflict exit code `3`, generic failure exit code `1`;
- bounded regular-file and stdin input;
- no TTY or TUI dependency;
- machine-readable Cobra validation errors when `--json` has been parsed.

### TUI preview

The TUI is rendered with the active theme; the schematic below shows the information hierarchy and controls without depending on terminal colors.

**Plan browser**

```text
╭─ Plans ──────────────────────────────────────────────────────────────────────╮
│Filter: /release                                                              │
│                                                                              │
│  scope    plan                 status         ver   updated   title          │
│▸ shared   release-prep         in-progress    v7    2m ago    Release        │
│  shared   api-migration        blocked        v3    1d ago    API move       │
│  session  a1b2c3d4…            awaiting       -     5m ago    -              │
│                                                                              │
│↑/↓ navigate  Enter open  n new  e edit  s status  x export  d delete         │
╰──────────────────────────────────────────────────────────────────────────────╯
```

**Plan detail**

```text
╭─ Plan: release-prep · shared ────────────────────────────────────────────────╮
│Scope: shared — collaborative, versioned                                      │
│Status: in-progress    Version: v7    Updated: 2m ago                         │
│Author: release-agent                                                         │
│                                                                              │
│# Release preparation                                                         │
│                                                                              │
│1. Validate the release configuration                                         │
│2. Run the compatibility test matrix                                          │
│3. Publish signed artifacts                                                   │
│                                                                              │
│↑/↓ scroll  r refresh  e edit  s status  x export  d delete  Esc back         │
╰──────────────────────────────────────────────────────────────────────────────╯
```

Session plans use the same browser/detail flow but are explicitly labeled read-only and unversioned. Destructive shared-plan actions name the plan and require confirmation.

### TUI `/plans`

- Discoverable plan browser and detail view.
- Scope, identity, status, revision, updated time, title, and scrollable content.
- Filter, keyboard and mouse navigation, create, external-editor update, status, export, and confirmed delete.
- Guarded mutations with actionable stale-version handling and preserved editor drafts.
- Asynchronous, bounded storage operations so slow or wedged storage does not freeze Bubble Tea's update loop.
- Live refresh across sessions and background tabs, including dialogs buried in the stack.
- Refresh coalescing, stale-session result rejection, duplicate-open guards, targeted idempotent detail closes, and duplicate-export protection.

### Runtime event

`PlanChangedEvent` carries scope, name, action, and version without content. Every active runtime stream subscribed to the shared toolset receives the event; subscriptions are deduplicated and released with stream lifecycle.

## Issue expectations

| Issue expectation | Implementation |
| --- | --- |
| Discover and inspect shared plans from the TUI | `/plans` browser and detail dialog |
| Show scope, identity, status, version, and updated time | Browser and detail metadata |
| Create/edit/status/export/delete with destructive confirmation | TUI actions and CLI subcommands |
| Stable headless JSON contract | `schema_version: "1"`, snake_case fields, structured errors |
| Distinguish shared and session plans | Explicit scopes; session mutations return `unsupported` |
| Prevent stale silent overwrites | expected-version checks plus cross-process mutation lock |
| Distinct host errors | Six typed service errors, covered by tests |
| Preserve existing plan tools/backends | Shared storage interface and backward-compatible tool JSON |
| Document workflows and historical work | CLI/TUI/tool docs with links to prior issues and PRs |

## Validation

| Check | Result |
| --- | --- |
| `task build` | pass |
| `task test` | pass |
| `task test-binary` | pass |
| `golangci-lint run` and repository lint | pass |
| Markdown/docs lint | pass |
| Linux and Windows image builds in CI | pass |
| Plan cross-compilation: Linux, Windows, js/wasm, wasip1, plan9 | pass |
| Race tests for plan storage, runtime fan-out, CLI, and plan TUI paths | pass |
| Real subprocess optimistic-lock race | exactly one winner and one conflict |
| Real binary CLI lifecycle and exit-code smoke tests | pass |

## Review history

The dedicated adversarial review team was run repeatedly against published PR heads. Its confirmed findings were corrected in follow-up commits, including cross-process races, notifier fan-out, bounded I/O, TUI lock waits, stale async results, duplicate dialogs/exports, and create-only revision-zero handling.

One unrelated pre-existing race remains in `TestLoadSessionThenClickEditLabel` (`App.ReplaceSession` versus session startup); the plan-focused race suites pass.

## Known limitation

The TUI's default no-overwrite export remains a `stat` plus atomic write sequence. Duplicate in-process exports are prevented, but a separate external process can still create the destination during the narrow race window.


Sayt-0:
not finished yet don't approve or merge

aheritier (commented):
## Review — first-class host UX for plan management

**Head:** `352d124a` · **CI:** green (21 check-runs, all completed — 15 success incl. `lint`, `build-and-test`, `license-check`, CodeQL ×3, `build-image` linux/amd64+arm64 and the docs checks; 6 skipped push/merge jobs) · no commit statuses, nothing pending or failing · `MERGEABLE`, `mergeStateStatus: BLOCKED`.

**Recommendation: do not merge or approve yet** — the author states ["not finished yet don't approve or merge"](https://github.com/docker/docker-agent/pull/3853#issuecomment-5094245333). This is a COMMENT-only review; the findings below are for the author's next pass, not gates I'm claiming beyond that.

The shape of this is strong: the file-lock build-tag split (`unix` / `windows` / `!unix && !windows`) with an honest doc comment about where cross-process exclusion genuinely degrades, the never-deleted sentinel inode rationale, atomic temp+rename writes, optimistic locking mirrored consistently across tools/CLI/TUI, the non-blocking event sink with an explicit release path, the pointer-identity dedup and its panic-safety justification, and a genuinely dense test suite (subprocess lock contention, TUI scroll/selection preservation, JSON error contract). Comments explain *why*, per AGENTS.md.

Four points are inline. The rest:

**1. Windows locking is compile-verified only (moderate).** `pkg/tools/builtin/plan/lock_windows.go` implements the `LockFileEx`/`UnlockFileEx` path, and `plan_lock_test.go` is — correctly — untagged, so it *would* exercise cross-process locking on Windows. But no CI job runs on a Windows runner: `.github/workflows/ci.yml` has `lint`, `build-and-test`, `license-check` on `ubuntu-latest` and `build-image` on linux/amd64 + linux/arm64 only. `task check-plan-cross` proves the Windows path compiles, not that it locks — and locking correctness is the crux of this feature. A minimal `windows-latest` job running `go test ./pkg/tools/builtin/plan/ ./pkg/plans/` would close the gap cheaply.

**2. `check-plan-cross` covers the build tags but not the new callers (minor).** All three lock variants are covered for `./pkg/tools/builtin/plan/` and `./pkg/plans/` (linux for `unix`, windows, and js/wasip1/plan9 for `!unix && !windows`) — good. Not covered: the new host code that also branches on GOOS, i.e. `pkg/tui/plans.go` (`notepad` vs `vi`) and `cmd/root/plans.go`; both packages ship `//go:build !windows` test files, so a Windows-only build break there would pass CI silently. Adding `./cmd/root/ ./pkg/tui/` to at least the `GOOS=windows` line would catch it.

**3. `plan_changed` isn't documented on the API/A2A surfaces (nit).** The event is registered in the runtime event registry (`pkg/runtime/event.go`) and decodes off the SSE client, so it reaches API and A2A consumers — but the docs updated here are TUI/CLI/tool pages only, and `docs/features/api-server/index.md`'s "Event types include" list isn't extended. That list is already non-exhaustive, so this is discretionary; a new externally visible event type still seems like a good candidate for it.

**4. Telemetry carries raw positional args (nit / privacy minimisation).** `runPlans` builds `trackArgs := append([]string{sub}, args...)`, so plan names and `--session` values reach `TrackCommand`/`TrackCommandError`, and the error variant also ships `err.Error()` (which can embed `--output`/`--file` paths). This matches the existing convention in `cmd/root/{models,alias,pull,a2a,debug}.go`, so it is not a regression introduced here. But plan names are user-authored content; tracking only `sub` — as `cmd/root/sandbox_cmd.go` already does — would be strictly better for a brand-new command group.

**5. `plan_detail.go` has no dedicated test file (nit).** Coverage is genuinely there — seven `TestPlanDetail*` tests plus the `newTestPlanDetail` helper — but they live in `pkg/tui/dialog/plan_browser_test.go`. AGENTS.md asks for tests alongside source files; splitting them into `plan_detail_test.go` would make the ~300-line dialog's coverage discoverable without changing a single assertion.

pkg/tui/plans.go
```diff
@@ -0,0 +1,969 @@
...
+		}
+	}
+	args := append(parts[1:], result.path)
+	// The editor process is owned by tea.ExecProcess, so exec.Command is intentional.
+	cmd := exec.Command(parts[0], args...) //nolint:noctx // owned by tea.ExecProcess
+	return tea.ExecProcess(cmd, func(err error) tea.Msg {
+		result.err = err
+		return result
+	})
+}
```
aheritier, line 723: **Minor (refactor):** this editor-resolution + launch block duplicates `pkg/tui/tui.go:3141-3153` almost line for line — `cmp.Or(os.Getenv("VISUAL"), os.Getenv("EDITOR"))` → `strings.Fields` → `notepad`/`vi` GOOS fallback → `append(parts[1:], path)` → `exec.Command` with the same `//nolint:noctx // owned by tea.ExecProcess` rationale.

There's already a `pkg/tui/internal/editorname` package that centralises the *display name* half of exactly this resolution (and duplicates the same Windows/Vi fallback decision a third time), so the launch half is the odd one out. Extracting one helper — e.g. `editorname.Command(path) *exec.Cmd`, or a small `externalEditorCmd(path)` in `pkg/tui` — would keep the sites from drifting; the Windows fallback and any future argument/flag handling are the likely divergence points.

cmd/root/plans.go
```diff
@@ -0,0 +1,892 @@
...
+			return plansValidationError(cmd, inner(cmd, args))
+		}
+	}
+	cmd.PreRunE = func(cmd *cobra.Command, _ []string) error {
+		if err := cmd.ValidateRequiredFlags(); err != nil {
+			return plansValidationError(cmd, err)
+		}
+		return plansValidationError(cmd, cmd.ValidateFlagGroups())
+	}
+}
```
aheritier, line 206: **Minor (latent footgun):** `cmd.Args` is chained defensively just above (`if inner := cmd.Args; inner != nil`), but `cmd.PreRunE` is assigned unconditionally — so any `PreRunE` already set on a plans subcommand is silently dropped. No subcommand in this file sets one today, so this isn't a live bug, and the root command's `PersistentPreRunE` is a separate hook that stays intact.

The concern is the asymmetry: the next person adding a `PreRunE` to a plans subcommand gets a silently skipped hook, not a compile error. `cmd/root/root.go:244-246` already uses the capture-then-chain pattern for this exact reason:

```go
inner := cmd.PreRunE
cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
	if inner != nil {
		if err := inner(cmd, args); err != nil {
			return plansValidationError(cmd, err)
		}
	}
	if err := cmd.ValidateRequiredFlags(); err != nil {
		return plansValidationError(cmd, err)
	}
	return plansValidationError(cmd, cmd.ValidateFlagGroups())
}
```

pkg/plans/service.go
```diff
@@ -0,0 +1,409 @@
...
+	if info, err := os.Stat(clean); err == nil && info.IsDir() {
+		return &ValidationError{Message: fmt.Sprintf("path %q is a directory, not a file", path)}
+	}
+	if err := os.MkdirAll(filepath.Dir(clean), 0o700); err != nil {
+		return &StorageError{Scope: scope, Op: "export", Err: err}
+	}
+	if err := atomicfile.Write(clean, strings.NewReader(content), 0o600); err != nil {
+		return &StorageError{Scope: scope, Op: "export", Err: err}
+	}
+	return nil
```
aheritier, line 408: **Moderate (behaviour / docs mismatch):** `writeExportFile` only rejects a *directory* target; an existing **file** is silently overwritten by `atomicfile.Write`. The TUI export path deliberately does the opposite — it pre-checks and refuses, and its own error message points the user at this command as the escape hatch:

> `<path> already exists — move it away, or export to a custom path with 'docker agent plans export'.`

and `docs/tools/plan/index.md` / `docs/features/tui/index.md` both document "An existing file is never overwritten" / "refusing to overwrite an existing file" for that path.

So the divergence looks deliberate, but it is undocumented where it matters: `newPlansExportCmd`'s `Long` promises only that "parent directories are created and the write is atomic, so a reader never observes a partial export", and `--output`'s help is just "Destination file for the plan content; required". A reader coming from the TUI's no-overwrite guarantee will reasonably assume the CLI is equally safe. Either state the overwrite explicitly in `Long`/`--output`, or apply the same guard here and add `--force` to opt out — which would also match the `--expected-version`/`--force` idiom the rest of the group already uses.

pkg/runtime/loop.go
```diff
@@ -1313,6 +1322,85 @@ func (r *LocalRuntime) configureToolsetHandlers(a *agent.Agent, events EventSink
...
+// The emitted event carries no agent attribution: the shared toolset
+// executes mutations for whichever session's agent is calling, so this
+// subscriber cannot know the mutator (see plan.Change). An empty agent name
+// follows the AgentContext convention for events without a meaningful agent;
+// consumers read the plan's Author field for collaborative attribution.
+//
+// Non-blocking sink like the RAG forwarder: the callback fires from
+// another session's tool call, and a blocking send into this stream's
+// events channel could hang that session's tool call.
+func (r *LocalRuntime) subscribePlanChanges(sess *session.Session, events EventSink) (release func()) {
```
aheritier, line 1344: **Question (design):** the fan-out here is per-stream across *every* agent's toolsets plus `sess.ExtraToolSets`, deduped by pointer identity on what the doc comment describes as a process-wide `*plan.ToolSet` singleton. The consequence is that a mutation performed by session A's tool call is delivered to every open stream in the process, with a hardcoded `"shared"` scope and empty agent attribution.

For the TUI (one process, one user, one `/plans` browser to refresh) that is exactly the desired behaviour, and the comment justifies the dedup well. For a long-lived `serve api` / `serve mcp` / `serve a2a` process serving many concurrent sessions, though, every session's SSE stream also observes every other session's plan mutations — plan names included. Is that intended on the server surfaces, or should the fan-out be scoped there (opt-in per stream, or filtered by scope/session)? Either way it'd be worth one line in this doc comment: it currently explains the dedup and the non-blocking sink, but not the cross-session visibility that follows from a shared notifier.

dgageot (approved):

aheritier:
See my comments for followups @Sayt-0

PR #3857, opened 2026-07-27
docs: explain GITHUB_TOKEN forwarding for eval containers

### LINKED ISSUE #3841: Docs: Explain `GITHUB_TOKEN` forwarding for eval
### Overview

Docker Agent does not automatically forward `GITHUB_TOKEN` or `GH_TOKEN` into eval containers, even when defined in the shell or `~/.config/cagent/.env`. This is intentional because GitHub tokens are broad credentials, but the current docs do not make it clear.

Add a warning to the evaluation and GitHub Copilot provider docs:

 Eval cases run in isolated containers. For security reasons, GITHUB_TOKEN and GH_TOKEN are not forwarded automatically. Explicitly pass the required token by name:

```
   docker agent eval agent.yaml ./evals -e GITHUB_TOKEN
```

 When using a custom env file, both flags are required:

```
   docker agent eval agent.yaml ./evals \
     --env-from-file /path/to/secrets.env \
     -e GITHUB_TOKEN
```

Also clarify that host-side judge validation may succeed while the evaluated agent fails if the token is not forwarded.

### Motivation

Env vars for other providers are forwarded to eval containers automatically and it is confusing that this doesn't happen for github-copilot.

### Use cases

_No response_

### Proposed solution

_No response_

### Alternatives

_No response_

### Related issues

_No response_

### Additional context

_No response_

priyanka25aug:
Fixes #3841

Eval containers deliberately do not receive GITHUB_TOKEN / GH_TOKEN, even though dedicated provider API keys are forwarded automatically (nonForwardableTokenEnvVars in pkg/config/auto.go excludes GITHUB_TOKEN). This is intentional but undocumented, and the asymmetry with other providers is confusing.

This PR adds a warning to the evaluation docs and the GitHub Copilot provider docs:
- How to pass the token explicitly with -e GITHUB_TOKEN
- That --env-from-file alone is not sufficient
- That the LLM judge runs on the host, so judge scoring can succeed while the evaluated agent fails to authenticate

### DIFF
--- a/docs/features/evaluation/index.md
+++ b/docs/features/evaluation/index.md
@@ -165,6 +165,38 @@ $ docker agent eval <agent-file>|<registry-ref> [<eval-dir>|./evals]
 | `-e, --env`         | (none)                      | Environment variables to pass to container (`KEY` or `KEY=VALUE`) |
 | `--repeat`          | `1`                         | Number of times to repeat each evaluation (useful for computing baselines) |
 
+### Provider Credentials
+
+Eval containers are isolated from your host environment. Dedicated model
+provider API keys (for example `ANTHROPIC_API_KEY` or `OPENAI_API_KEY`) are
+forwarded into eval containers automatically, so most provider setups work
+without extra flags.
+
+> [!WARNING]
+> **`GITHUB_TOKEN` and `GH_TOKEN` are not forwarded automatically**
+>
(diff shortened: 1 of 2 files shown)


Sayt-0 (approved):
looks good thank you for the contribution !

Sayt-0:
One more thing : can you sign your commits ? It's mandatory in this repo

dgageot (approved):

PR #3865, opened 2026-07-29
fix(plans): address review follow-ups from #3853

Sayt-0:
## Summary

Follow-up to #3853 that addresses review feedback around export safety, telemetry privacy, validation hooks, shared plan events, editor resolution, test placement, and Windows coverage.

## Review feedback

| Review point | Resolution |
| --- | --- |
| Windows locking was compile-only | Added a native `windows-latest` plan test job and made it gate image publication |
| Windows caller coverage was incomplete | Added `cmd/root` and `pkg/tui` to the Windows cross-build |
| Plans telemetry included user values | Limited telemetry to the subcommand and stable error codes |
| Existing `PreRunE` hooks could be replaced | Chained the existing hook and added regression tests |
| CLI export overwrite behavior was ambiguous | Existing destinations are refused by default; `--force` explicitly enables atomic replacement |
| Editor launch resolution was duplicated | Centralized command construction in `pkg/tui/internal/editorname` |
| Plan detail tests were hard to discover | Moved them alongside `plan_detail.go` |
| `plan_changed` was undocumented | Documented its payload and process-global visibility |

## Validation

| Check | Result |
| --- | --- |
| `task build` | pass |
| `task test` | pass |
| `task test-binary` | pass |
| `task lint` | pass |
| `task check-plan-cross` | pass |
| `go test -race -count=1 ./pkg/plans` | pass |
| Workflow and documentation lint | pass |
| Native Windows plan tests | pending PR CI |

## Design notes

- Non-force export uses atomic hard-link publication after fully writing, syncing, closing, and securing a temporary file. Filesystems without hard-link support return a typed storage error rather than weakening the no-overwrite guarantee.
- `plan_changed` remains process-global because the shared notifier does not carry mutating-session identity and every open plans view must refresh. The API documentation now makes that visibility explicit.


.github/workflows/ci.yml
```diff
@@ -73,6 +73,23 @@ jobs:
           task test
           task test-binary
 
+  # Plan storage relies on OS-specific file locking and path semantics, so its
```
aheritier, line 76: it's interesting @Sayt-0 as we said that we wanted to expand a bit our tests on windows to avoid some issues which were impacting only this OS. To see if we want to run only a few tests or all of them in the future

aheritier (approved):

Issue #3871, opened 2026-07-30
Sandbox runs broken by version skew: bundled v1.88.1 self-update 404s and ignores --models-gateway


aboullaite:
## Summary

Running `docker agent run <agent> --sandbox` with a Docker Model Runner model fails
end to end because of a version mismatch between the host CLI and the docker-agent
binary bundled in the sandbox template, and the bundled version cannot self-update.

## Environment

- Host: macOS 15 (M1 Max), Docker Desktop 4.83.0, Docker CLI 29.6.2
- Host docker-agent: v1.119.0
- Sandbox template `docker/sandbox-templates:docker-agent`: bundles docker-agent v1.88.1
- sbx CLI installed from docker/tap (July 2026)

## Steps to reproduce

1. `docker model pull ai/gemma4` (Model Runner enabled and serving on the host)
2. `sbx policy init balanced`
3. In a project directory:
   `docker agent run coder --sandbox --yolo --exec --model dmr/ai/gemma4 --models-gateway http://host.docker.internal:12434/engines "any task"`

## What happens

Host side, the CLI dies after sandbox creation:

```
Models gateway: http://host.docker.internal:12434/engines (allowlisting host.docker.internal:12434 in the sandbox proxy)
ERROR: inspect exec: context deadline exceeded
```

Meanwhile inside the sandbox the agent process is still alive (orphaned), so failed
runs leave zombie docker-agent processes in the VM.

Running docker-agent manually inside the sandbox surfaces the underlying failures:

```
docker-agent: updating from v1.88.1 to v1.119.0...
WARN Self-update skipped; continuing with current binary error="fetching checksums.txt: HTTP 404"
docker-agent: self-update failed (fetching checksums.txt: HTTP 404); continuing with current version
Error: docker model runner is not available
please install it and try again (https://docs.docker.com/ai/model-runner/get-started/)
```

Three distinct problems:

1. **Self-update 404**: the in-sandbox binary tries to update to the host version and
   fails fetching `checksums.txt` from the release assets (the sandbox proxy allows
   github.com/release-assets, so this looks like a missing artifact, not a policy block).
2. **v1.88.1 ignores `--models-gateway`** for the DMR provider availability check: it
   probes for a VM-local Model Runner and hard-fails, even though the gateway endpoint
   is reachable through the sandbox proxy (verified with curl from inside the VM).
3. **Host CLI times out** (`inspect exec: context deadline exceeded`) instead of
[...]

PR #3875, opened 2026-07-31
fix(dmr): honor the models gateway when creating the client

aboullaite:
Fixes the `--models-gateway` part of #3871.

Every other provider (openai, anthropic, gemini) routes through the configured models gateway, but the DMR provider ignored it: `NewClient` always probed for a local Model Runner via `docker model status` and failed with `ErrNotInstalled` in environments where the gateway is the only reachable path. The concrete case from the issue is a Docker Sandbox, where `docker agent run --sandbox --models-gateway ...` has already allowlisted the gateway in the sandbox proxy, yet the agent inside the sandbox refuses to start.

## What changed

When a gateway is configured, `NewClient` now dials it directly as the OpenAI compatible endpoint (`<gateway>/v1/`), skipping endpoint discovery, auto pull and backend configuration, which all assume a local installation. The gateway branch follows the same if/else shape the sibling providers use and shares the existing construction tail. Gateway requests carry the standard request contract (`X-Cagent-Forward` set to the default DMR host URL, provider and model identity headers, gateway query parameters). DMR requires no auth, so the gateway is dialed with an empty API key, matching the direct path. `dmrmodels.defaultHostURL` is now exported as it doubles as the forward target.

One consideration deliberately left out: threading `environment.Provider` through to call `VerifyDockerGatewayAuth`/`GatewayAuthToken`. Those helpers no op for gateways outside the trusted Docker domains, and DMR is a local, credential free provider, so there is no real configuration today where they would fire. Happy to add the plumbing if you want parity with the other providers.

## Validation

1. `go test ./pkg/model/provider/dmr/...` passes, including a request contract test that sends a real SDK backed operation (`CreateBatchEmbedding`) through an httptest gateway and asserts the path, query parameters and `X-Cagent-*` headers, plus a broken docker CLI shim test proving no local discovery happens when a gateway is set.
2. `golangci-lint run pkg/model/provider/dmr/...` reports 0 issues; `go build ./...` is clean.
3. End to end: built for linux/arm64 and ran inside a Docker Sandbox microVM with `--model dmr/ai/gemma4 --models-gateway http://host.docker.internal:12434/engines`. The client connects through the sandbox proxy to the host's Model Runner and completes the task. The same invocation on the current release fails with "docker model runner is not available".

The self update 404 and the host CLI's `inspect exec` timeout from #3871 are not addressable from this repo (release pipeline and sbx side respectively), so this PR covers only the gateway half.

### DIFF
--- a/pkg/model/provider/dmr/client.go
+++ b/pkg/model/provider/dmr/client.go
@@ -9,7 +9,9 @@ import (
 	"log/slog"
 	"maps"
 	"net/http"
+	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/openai/openai-go/v3"
@@ -18,6 +20,7 @@ import (
 
 	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/httpclient"
 	"github.com/docker/docker-agent/pkg/model/provider/base"
 	"github.com/docker/docker-agent/pkg/model/provider/dmr/dmrmodels"
 	"github.com/docker/docker-agent/pkg/model/provider/oaistream"
@@ -44,7 +47,8 @@ type Client struct {
 
 	client     openai.Client
 	httpClient *http.Client
-	engine     string
+	// engine is empty in gateway mode: engine-gated request shaping stays on defaults.
+	engine string
 
 	// attachmentCaps records the document MIME types this DMR-hosted model is
 	// declared to accept natively, parsed from provider_opts.supports_images /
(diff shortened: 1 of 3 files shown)


Sayt-0 (approved):
Changes looks good. Thank you !

Sayt-0:
Can you sign your commits ?

aboullaite:
> Can you sign your commits ?

Yes sorry for that. Done ✅

Sayt-0:
hello @aboullaite It looks like they still haven't been signed.

aboullaite:
Pretty sure its fixed now @Sayt-0. At least that's what the UI says

PR #3878, opened 2026-07-31
feat(tui): make current session plans editable

Sayt-0:
## Summary

- label the active session plan as `current session` in `/plans`
- keep shared named plans explicitly identified as workspace-global
- allow editing the current session plan body from the browser and detail views
- keep create, status, and delete unsupported for session plans
- refresh the browser and detail view after a successful edit

## Relationship

Follow-up to #3876.

**Do not merge before #3876.** PR #3876 fixes the terminal and UTF-8 behavior used by the external plan editor flow. Keep this PR in draft until #3876 is merged.

## Feedback mapping

| Feedback | Implementation |
|---|---|
| Only the current session's session plan should be shown | The existing listing already reads only the active session's session plan. This behavior is preserved and covered by tests. |
| Plans belonging to the current session should be identifiable | The session row is now labelled `current session`, while its full session ID remains visible in the footer and detail view. |
| Named plans from other sessions are confusing | Named plans remain explicitly marked `shared`, because they are workspace-global and have no session ownership. |
| Session plans should be user-editable | Pressing `e` edits the session plan body from either the browser or detail view. |
| Session plan metadata operations | Status and delete remain unsupported because session plans are unversioned and have no shared-plan metadata. |

## Validation

- `go test -count=1 ./pkg/tools/builtin/sessionplan ./pkg/plans ./pkg/tui/dialog ./pkg/tui/messages ./pkg/tui ./cmd/root`
- `go vet ./pkg/tools/builtin/sessionplan ./pkg/plans ./pkg/tui/dialog ./pkg/tui/messages ./pkg/tui ./cmd/root`
- `task --force lint`
- `task --force build`
- `task --force check-plan-cross`
- real PTY flow: `/plans` -> verify `current session` -> verify status is refused -> `e` -> external editor -> refreshed detail -> persisted content

## Test-suite note

`task test` was run. The changed packages pass, but the complete suite is blocked in this execution environment by unrelated SSRF/local-address tests because `HTTP_PROXY` and `HTTPS_PROXY` are forced through a socket firewall. The failures are in `pkg/config`, `pkg/httpclient`, `pkg/tools/builtin/api`, `pkg/tools/builtin/fetch`, and `pkg/tools/builtin/openapi`.

### DIFF
--- a/pkg/plans/plans.go
+++ b/pkg/plans/plans.go
@@ -7,10 +7,11 @@
 // The package wraps the existing storage rather than duplicating it. Shared
 // plans go through a caller-supplied plan.Storage — pass plan.SharedStorage()
 // to operate on the same store, and thus the same mutex, as the plan tools of
-// agents running in this process. The session plan is read through the
-// sessionplan helpers. Session plans have no revisions or optimistic locking
-// and belong to their session, so the Service exposes them read/export-only
-// and rejects mutations with a typed *UnsupportedError.
+// agents running in this process. The session plan is read and written
+// through the sessionplan helpers. Session plans have no revisions or
+// optimistic locking: the version-guarded mutations reject them with a typed
+// *UnsupportedError, and the one supported write is UpdateSession, which
+// replaces the body of an existing plan last-write-wins.
 package plans
 
 import (
@@ -26,14 +27,17 @@ const (
 	// collaborate on. Shared plans are versioned and fully mutable.
 	ScopeShared Scope = "shared"
 	// ScopeSession is the per-session plan of the "draft, review, execute"
-	// workflow. At most one exists per session; it has no versions and is
-	// read/export-only through the Service.
+	// workflow. At most one exists per session; it has no versions, so the
+	// Service reads, exports, and replaces its body through UpdateSession
+	// but rejects the version-guarded mutations.
 	ScopeSession Scope = "session"
 )
 
-// Mutable reports whether plans in this scope can be created, updated, and
-// deleted through the Service, so a frontend can disable editing up front
-// instead of provoking an *UnsupportedError.
+// Mutable reports whether plans in this scope support the full set of
+// version-guarded mutations (create, update, set-status, delete), so a
+// frontend can disable those actions up front instead of provoking an
+// *UnsupportedError. Session plans are not Mutable in this sense; their
+// body is still replaceable through UpdateSession.
 func (s Scope) Mutable() bool { return s == ScopeShared }
 
 // Plan is the host-facing view of a plan from either scope.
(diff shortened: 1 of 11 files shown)


aheritier (approved):

## Code Review: feat(tui): make current session plans editable

All tests pass (plans, tui, cmd/root), build is clean, and go vet shows no issues. This is a well-structured feature addition.

### What the PR does

Adds the ability to edit the body of a session plan (the per-session draft/review/execute plan) from the TUI browser and detail dialogs. Session plans were previously shown as read-only. The PR:

- Adds `UpdateSession` to the `plans.Service` interface and implements it on the concrete service
- Routes the 'e' key in the plan browser and detail dialog to either the existing shared-plan guarded update or the new unguarded `UpdateSession` path
- Displays session plan rows with the label "current session" instead of the raw session ID
- Adds comprehensive test coverage at all layers

### Strengths

- **Correct version handling**: The sentinel `ExpectedVersion: 0` for session plans (which have no versions) is well-reasoned. The `handleEditPlan` drift check `currentVersion != msg.ExpectedVersion` correctly short-circuits (0 == 0) for session plans, so the editor always opens.
- **Atomic writes**: `UpdateSession` delegates to `sessionplan.WriteContent`, which uses an atomic rename — readers see old or new content, never a partial write; symlinks in the path are replaced rather than followed.
- **Edit-never-creates contract**: The pre-check with `os.Stat` enforces that `UpdateSession` edits only, returning `*NotFoundError` for missing plans. The acknowledged TOCTOU race window is documented and accepted.
- **Robust error propagation**: `planEditorFailureCmd` handles `NotFoundError` from `UpdateSession` gracefully — the draft is kept, the user gets a useful notification, and no data is lost.
- **Consistent UX**: Session-specific notifications ("Updated the current session plan.", "Session plan left unchanged") avoid the confusing shared-plan version strings that would appear if the session code path used the shared plan messages.
- **Test coverage**: All edge cases are covered — normal edit, last-write-wins, invalid session ID, edit-never-creates, validation failures, expired context, symlink resistance, version drift, empty draft, and plan vanished.

### Minor observations (non-blocking)

1. **Redundant filter check for shared plans** (`plan_browser.go`, line 244-245): In `applyFilter`, both `p.Name` and `planDisplayName(p)` are checked. For shared plans `planDisplayName` returns `p.Name`, so the two terms are equivalent and the second check is never the tiebreaker. For session plans both are needed (session ID and "current session" label). The redundancy is harmless and short-circuits away, but the intent could be clearer.

2. **Comment partially restates code** (`service.go`, line 192-193): The comment `// Read the plan back so the caller gets the stored bytes and the real file modification time.` opens with "Read the plan back" which mirrors what `getSession` obviously does. Per project conventions, the WHY is valuable (real mtime from the atomic rename, not the in-memory content); the WHAT prefix could be dropped.

3. **TOCTOU race documented but worth calling out**: Between `os.Stat(path)` and `sessionplan.WriteContent`, an external deletion would cause the write to recreate the plan — silently violating the edit-never-creates contract. The comment acknowledges this. The test coverage tests a missing plan from the start (not the narrow race itself), which is the realistic and testable path.

None of these require changes — they're observations for context. The implementation is correct, safe, and well-tested.

PR #3884, opened 2026-08-02
fix(server): stop per-session toolsets on session delete

EronWright:
# fix(server): stop per-session toolsets on session delete

## Problem

A session materialised via `teamloader` owns a `team` whose toolsets may hold
external resources — notably **stdio MCP subprocesses**. `DeleteSession` cancels the
runtime context but never calls `team.StopToolSets`, and `BatchDeleteSessions` has
the same gap (`LocalRuntime.Close` only stops background agents; `StopToolSets` was
otherwise reached only via the transient `GetAgentToolCount`). So a deleted session's
MCP subprocesses **leak until the server process exits**.

## Fix

Track the per-session `team` on `activeRuntimes` and route both delete paths through
one `scheduleTeardown` helper. It keeps the entry in `deletedSessions`, waits for the
session's stream to drain, stops the toolsets, then drops the entry. Teardown runs in
a background goroutine that outlives the originating request, on a context detached
with `WithoutCancel` (so it still carries trace/logging values). The drain and
`StopToolSets` carry **separate** budgets — a session that exhausts the 5 min drain
must still get a live context to stop its toolsets with.

`team` is nil for attached runtimes (`AttachRuntime`), whose toolset lifecycle belongs
to the embedder — a no-op there.

Two things beyond the original revision, from review feedback:

- **`BatchDeleteSessions` is now `WaitStopped`-able.** It previously never stored into
  `deletedSessions`, so `WaitStopped` returned `nil` immediately while teardown was
  still running. It now registers the same teardown as `DeleteSession`.
- **`WaitStopped` means "fully torn down".** It waits on a completion channel closed
  after `StopToolSets` instead of polling the streaming mutex. This also closes an
  existing race where `WaitStopped` could win the `TryLock`, delete the entry and
  return `nil` before teardown had run.

## Why this is safe

Teams are never shared between sessions: `loadTeamWithConfig` →
`teamloader.LoadWithConfig` builds a fresh team and fresh toolset instances per call
(`loaderdefaults.Opts()` hands in a new `toolsets.NewDefaultToolsetRegistry()` each
time). The only process-wide singletons — `plan`, and `todo` with `shared: true` — do
not implement `Stop`, so `StopToolSets` cannot reach across sessions.

There is already precedent in-tree: `GetAgentToolCount` loads a throwaway team, starts
every toolset and unconditionally stops them while arbitrary sessions are live. If
teams or toolsets were shared, that endpoint would already be killing live sessions'
MCP servers on every call.

`StopToolSets` is also safe to call twice — `Agent.StopToolSets` skips
`!toolSet.IsStarted()`, and `mcp.Toolset.Stop` documents itself idempotent.

## Test

Four new tests in `pkg/server/session_manager_test.go`, all of which fail without the
fix:

- `TestDeleteSession_StopsSessionToolSets` — the leak itself.
- `TestBatchDeleteSessions_StopsSessionToolSets` — the batch path, incl. `WaitStopped`.
- `TestWaitStopped_WaitsForToolSetTeardown` — teardown parks on an in-flight stream and
  completes once it drains; `WaitStopped` returns only after the toolsets are stopped.
- `TestDeleteSession_StopsSessionToolSetsOnlyOnce` — no double-stop.

`task test` and `task lint` green locally, plus `go test ./pkg/server/ -race`.
Verified against a multi-tenant host: the per-session subprocess count returns to zero
on delete and after server shutdown (previously it stayed pinned until exit).

## Noted, not fixed here

- `LocalRuntime.Close` never calls `StopToolSets`, so `runtimeForSession`'s
  failed-construction `defer run.Close()` does not deliver on its "an unclosed runtime
  would leak its tool sets" comment. Real, but a `pkg/runtime` bug.
- `SessionManager` has no `Close`/`Shutdown`, so toolsets of sessions that are never
  deleted still leak at process exit.
- `Team.StopToolSets` is fail-fast: the first agent error aborts the loop, leaving
  later agents' toolsets running.

### DIFF
--- a/pkg/server/session_manager.go
+++ b/pkg/server/session_manager.go
@@ -39,6 +39,18 @@ type activeRuntimes struct {
 	session  *session.Session        // The actual session object used by the runtime
 	titleGen *sessiontitle.Generator // Title generator (includes fallback models)
 
+	// team is the per-session team this manager built via teamloader in
+	// runtimeForSession. It is stopped (StopToolSets) on session teardown so
+	// stdio MCP subprocesses and other toolset-owned resources are released,
+	// not leaked until process exit. Nil for attached runtimes (AttachRuntime),
+	// whose team and toolset lifecycle belong to the external embedder.
+	team *team.Team
+
+	// stopped is closed once teardown has finished: the stream has drained
+	// and the toolsets have been stopped. Set by scheduleTeardown, which is
+	// the only writer; nil until the session is deleted.
+	stopped chan struct{}
+
 	streaming sync.Mutex // Held while a RunStream is in progress; serialises concurrent requests
 }
 
@@ -823,6 +835,91 @@ func (sm *SessionManager) GetSessions(ctx context.Context) ([]*session.Session, 
 	return sessions, nil
 }
 
+const (
+	// sessionDrainTimeout bounds how long teardown waits for a deleted
+	// session's stream to drain before giving up and stopping its toolsets
+	// anyway.
+	sessionDrainTimeout = 5 * time.Minute
+	// toolSetStopTimeout bounds StopToolSets. It is a separate budget from
+	// sessionDrainTimeout on purpose: a session that exhausted the drain
+	// budget must still get a live context to stop its toolsets with.
+	toolSetStopTimeout = 30 * time.Second
+)
+
+// scheduleTeardown detaches the post-delete teardown of a session's runtime and
+// runs it in the background: wait for the stream to drain, release the session's
+// toolset-owned resources, then drop the deletedSessions entry. Both delete
+// paths (DeleteSession and BatchDeleteSessions) go through here so they share
+// one teardown contract; callers hold sm.mux.
+//
+// The entry stays in deletedSessions for the duration so WaitStopped can observe
+// teardown after the runtime has been deregistered, and rs.stopped is closed
+// before the entry is dropped so a waiter never misses the completion signal.
+//
+// ctx only supplies values (trace and logging context): teardown outlives the
+// request that triggered the delete, so cancellation is stripped and each phase
+// carries its own deadline instead.
+func (sm *SessionManager) scheduleTeardown(ctx context.Context, sessionID string, rs *activeRuntimes) {
+	rs.stopped = make(chan struct{})
+	sm.deletedSessions.Store(sessionID, rs)
+
+	detached := context.WithoutCancel(ctx)
+	go func() {
+		// Deferred LIFO: close(stopped) first, so a WaitStopped that still
+		// holds the entry is released before the entry disappears.
+		defer sm.deletedSessions.Delete(sessionID)
+		defer close(rs.stopped)
+
+		drainSessionStream(detached, rs)
+		sm.stopSessionToolSets(detached, rs)
+	}()
+}
+
+// drainSessionStream blocks until the session's RunStream has released the
+// streaming mutex, or until sessionDrainTimeout expires. Teardown continues
+// either way — the timeout exists so a wedged stream cannot pin the session's
+// toolsets forever.
+func drainSessionStream(ctx context.Context, rs *activeRuntimes) {
+	ctx, cancel := context.WithTimeout(ctx, sessionDrainTimeout)
+	defer cancel()
+
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer ticker.Stop()
+	for {
+		if rs.streaming.TryLock() {
+			rs.streaming.Unlock()
+			return
+		}
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		}
+	}
+}
+
+// stopSessionToolSets releases the toolset-owned resources (e.g. stdio MCP
+// subprocesses) of a per-session team the manager built via teamloader —
+// otherwise they leak until the server process exits. It is a no-op for
+// attached runtimes, whose team is nil and whose toolset lifecycle belongs to
+// the external embedder. Call it only after the session's stream has drained,
+// so a toolset is never stopped mid-turn.
+func (sm *SessionManager) stopSessionToolSets(ctx context.Context, rs *activeRuntimes) {
+	if rs == nil || rs.team == nil {
+		return
+	}
+	ctx, cancel := context.WithTimeout(ctx, toolSetStopTimeout)
+	defer cancel()
+
+	sid := ""
+	if rs.session != nil {
+		sid = rs.session.ID
+	}
+	if err := rs.team.StopToolSets(ctx); err != nil {
+		slog.ErrorContext(ctx, "Failed to stop session tool sets", "session_id", sid, "error", err)
+	}
+}
+
 // DeleteSession deletes a session by ID. It cancels the runtime context and
 // removes the session from all registries. Callers that need to wait for
 // the stream to fully stop should call WaitStopped afterwards.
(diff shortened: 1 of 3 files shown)


aheritier:
The idea is interesting @EronWright Could you rebase your PR and validate that linters and tests are passing.
Before merging it I would like to get a review by a some people like @krissetto @simonferquel or @rumpl because even if the idea looks good I am not sure if it could have functionally some side effects in some cases

pkg/server/session_manager.go
```diff
@@ -1898,6 +1941,30 @@ func (sm *SessionManager) BatchDeleteSessions(ctx context.Context, sessionIDs []
...
+							}
+							select {
+							case <-drainCtx.Done():
+								return
+							case <-ticker.C:
+							}
+						}
+					}(sessionRuntime)
+				}
 				sm.runtimeSessions.Delete(sessionID)
```
docker-agent, line 2190: **[medium] `BatchDeleteSessions` new drain goroutine cannot be awaited via `WaitStopped`**

The PR adds an asynchronous goroutine in `BatchDeleteSessions` that waits for the session stream to drain and then calls `sm.stopSessionToolSets` — the same teardown work that `DeleteSession` does. However, unlike `DeleteSession` (which stores the runtime in `deletedSessions` before returning), `BatchDeleteSessions` never calls `sm.deletedSessions.Store(sessionID, sessionRuntime)`.

`WaitStopped` reads exclusively from `deletedSessions`:
```go
rs, ok := sm.deletedSessions.Load(sessionID)
if !ok {
    return nil // already cleaned up
}
```

So a caller that does `BatchDeleteSessions(...)` followed by `WaitStopped(sessionID, ...)` will receive an immediate `nil` even while the new drain goroutine is still running (waiting to drain the stream and call `StopToolSets`). Before this PR there was nothing to wait for in `BatchDeleteSessions`, but now there is — and the wait contract is silently broken.

**Impact**: Any consumer that uses `WaitStopped` to synchronise on full teardown (e.g. to confirm MCP subprocesses have exited) after a batch delete will get a false-complete signal. Consider mirroring `DeleteSession`'s `deletedSessions.Store` call in `BatchDeleteSessions` (and having the drain goroutine call `deletedSessions.Delete` when done), or documenting that `WaitStopped` is not supported after batch deletes.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 65/100 |

docker-agent (commented):

One medium-confidence finding was identified in the new code added by this PR.

aheritier:
There are linter + tests failures @EronWright

pkg/server/session_manager.go
EronWright: Good catch — fixed rather than documented around.

Both delete paths now go through a shared `scheduleTeardown` helper that stores into `deletedSessions`, drains the stream, stops the toolsets, then drops the entry. `BatchDeleteSessions` gets the same `WaitStopped` semantics as `DeleteSession`.

While there, `WaitStopped` now waits on a completion channel closed after `StopToolSets` instead of polling the streaming mutex — that also closes a pre-existing race where it could win the `TryLock`, delete the entry and return `nil` before teardown had run. Drain and `StopToolSets` also got separate timeout budgets, since a drain timeout previously handed `StopToolSets` an already-expired context.

Covered by four new tests in `pkg/server/session_manager_test.go`. The PR description has been updated with the full details.

aheritier (commented):
Reviewed `e0785f3`. CI is green on the head SHA (`build-and-test`, `lint`, `windows-tests`, `license-check`, both `build-image` legs; the second `save-context` entry is a superseded concurrency-cancelled run).

The diagnosis and the fix both hold up. I verified the load-bearing safety claim rather than taking it on trust:

- `loadTeamWithConfig` (`pkg/server/session_manager.go:1605-1613`) calls `teamloader.LoadWithConfig` on every construction with no team cache, so no two sessions share a `*team.Team`.
- The two genuinely process-wide toolset singletons — `pkg/tools/builtin/plan/plan.go:368` (`sharedToolSet`) and `pkg/tools/builtin/todo/todo.go:189` (`newSharedTodoTool`) — implement no `Stop`, so `StartableToolSet.Stop` (`pkg/tools/startable.go:212-224`) bottoms out in `return nil` for them. `StopToolSets` therefore cannot reach across sessions.
- `lsp.Multiplexer` was the one plausible counterexample — it has a real `Stop` that tears down backend LSP processes — but it is constructed per load at `pkg/teamloader/teamloader.go:840`, not shared.

I also checked the "never stop a toolset mid-turn" invariant against every path that takes `activeRuntimes.streaming`. `RunSession` (`session_manager.go:1006`, `:1036`) and `AddMessage`/`UpdateMessage` (`:1888`, `:1917`) all hold `sm.mux` and load from `runtimeSessions`; `recallSession` — the one path that `TryLock`s *before* taking `sm.mux` (`:1278-1286`) — re-checks `runtimeSessions` under the mutex and releases the lock when the entry is gone. Since both delete paths remove the entry while holding `sm.mux`, no new stream can start after the drain observes the mutex free. `deletedSessions` is a `concurrent.Map` guarded by an `RWMutex`, so the `rs.stopped` write that precedes `Store` is properly published to the `WaitStopped` reader after `Load`, and the LIFO `close(stopped)` → `Delete(sessionID)` ordering closes the missed-signal window. Reworking `WaitStopped` onto a completion channel instead of having it win the `TryLock` and delete the entry itself is a real bug fix in its own right, and routing both delete paths through one `scheduleTeardown` removes the duplicated drain loop rather than adding a second copy.

Two non-blocking findings.

**[should-fix]** `pkg/server/session_manager.go:1020` — the single line that actually closes the leak, `team: tm` in `RunSession`'s `activeRuntimes` literal, is asserted nowhere. `newToolSetSession` (`pkg/server/session_manager_test.go:2022-2025`) stores a hand-built `activeRuntimes{..., team: tm}` straight into `sm.runtimeSessions`, so all four new tests exercise teardown but never the wiring that feeds it. Evidence: `team:` appears exactly once in the test file (line 2025, inside that helper); every call site discards `runtimeForSession`'s new third return as `_` (`session_manager_test.go:1231`, `:1757`, `author_safety_test.go:82`, `:269`); and no test asserts `rs.team != nil` on the entry `RunSession` stored. Delete `team: tm` from line 1020 and the leak comes back with all four new tests still green. This repo already guards against exactly this shape one seam over — see the comment on `TestRuntimeForSession_RegistersSessionScopedElicitationSink` (`session_manager_test.go:1206-1210`): "the latter would keep passing even if runtimeForSession's `run.OnElicitationRequest(...)` registration were deleted, since it never exercises what got wired onto `run`". Cheapest fix: capture the team in one existing `runtimeForSession` test with a `require.NotNil`, and/or assert `rs.team != nil` after `RunSession`.

**[should-fix]** `pkg/server/server.go:487-488` — `deleteSession`'s comment still says `?wait=true` blocks "until the runtime's stream goroutine has fully exited (the streaming mutex is released)", which is precisely the contract this PR replaces. `WaitStopped` now also waits for `StopToolSets`, under the same 10s default budget (`server.go:474`), so a slow toolset teardown can turn a previously-200 `DELETE ?wait=true` into the 202 "stop still in progress" branch. The new behaviour is the right one and it degrades gracefully, but the comment describing it — and the same claim in `server_attached_test.go:303` — now contradicts the code. Worth a one-line update here since it is the same package.

**[optional]** The `Claude-Session: https://claude.ai/code/session_...` trailer resolves only inside the authoring session and has no precedent in the last 400 commits on `main` (0 occurrences). The commit message is self-contained without it; consider dropping it before merge.

The three items under "Noted, not fixed here" are all real and correctly scoped out — `LocalRuntime.Close` not calling `StopToolSets` and `Team.StopToolSets` being fail-fast each deserve their own issue.

aheritier:

PR #3893, opened 2026-08-03
fix(runtime): resolve transfer_task caller from pinned session agent

### LINKED ISSUE #3886: Design boundary? transfer_task inside a run_background_agent sub-agent fails with 'target agent not in sub-agents list' because handler uses CurrentAgent() instead of the pinned session agent
## Summary

I'm trying to confirm whether this is the intended design boundary or a bug: **can a sub-agent dispatched via `run_background_agent` call `transfer_task` to delegate to its own `sub_agents`?**

## Reproduction

A coordinator (`root`) fans out work with `run_background_agent` to a composite agent (`pipeline`), which in turn tries to `transfer_task` to its own sub-agents (`director`, `producer`, `reviewer`):

```yaml
agents:
  root:
    description: Batch scheduler
    sub_agents: [pipeline]
    toolsets:
      - type: background_agents
      - type: think

  pipeline:
    description: Composite pipeline (director -> producer -> reviewer)
    sub_agents: [director, producer, reviewer]
    toolsets:
      - type: think

  director: ...
  producer: ...
  reviewer: ...
```

The `pipeline` agent, running inside the background sub-session, calls `transfer_task(agent="director", ...)` and fails with:

```
Agent root cannot transfer task to director: target agent not in sub-agents list. Available agent IDs are: pipeline
```

The error incorrectly names **root** as the caller (root's sub-agents are only `[pipeline]`), even though the call originates from `pipeline` whose sub-agents do include `director`.

## Root cause analysis

`pkg/runtime/agent_delegation.go`, `handleTaskTransfer` resolves the calling agent via `r.CurrentAgent()`:

```go
a := r.CurrentAgent()
if errResult := validateAgentInList(a.Name(), params.Agent, "transfer task to", "sub-agents list", a.SubAgents()); ...
```

Background sub-sessions are created with `PinAgent: true` + `WithAgentName(cfg.AgentName)` (in `RunAgent` -> `runCollecting`), which pins the **session** to `pipeline` but deliberately does NOT mutate the runtime's shared `currentAgent` field (which remains `root`). So `CurrentAgent()` returns `root`, and the transfer validation fails.

Meanwhile `agentRouter.ResolveSession` (agent_router.go) explicitly handles this case:

```go
// ResolveSession returns the agent for sess: when sess pins a specific
// agent (e.g. background agent tasks), that agent is returned directly
// instead of reading the shared current-agent field; otherwise Current is returned.
```

but `handleTaskTransfer` does not use it.

## Question

Is delegating further from inside a `background_agent` a supported pattern? The docs for `background_agents` describe dispatching work to sub-agents that complete it themselves (all examples are leaf agents with toolsets, none re-delegate). If background agents are **only meant to be leaf executors**, then this is working as designed and my configuration is out of scope — please confirm and I'll restructure.

If background agents *are* meant to support nested delegation, then `handleTaskTransfer` should resolve the caller via the pinned session agent (`resolveSessionAgent(sess)`) rather than `CurrentAgent()`.

## Verification

I wrote a regression test (`TestHandleTaskTransfer_UsesPinnedSessionAgent`) that pins a session to `pipeline` (mimicking `run_background_agent`) with `currentAgent` = `root`, then calls `transfer_task` to `director`:

- Without fix: fails with exactly `Agent root cannot transfer task to director: target agent not in sub-agents list. Available agent IDs are: pipeline`
- With fix (`a := r.resolveSessionAgent(sess)`): passes

Environment: v1.111.0 (runtime), latest `main` (8003b634c) confirmed the code is unchanged.

yangtuooc:
## What

Fixes `transfer_task` failing inside a `run_background_agent` sub-session. See issue #3886.

## Problem

A coordinator (`root`) fans out work via `run_background_agent` to a composite agent (`pipeline`) that has its own `sub_agents` (e.g. `director`). When `pipeline` calls `transfer_task(agent="director")`, it fails with:

```
Agent root cannot transfer task to director: target agent not in sub-agents list. Available agent IDs are: pipeline
```

The error names **root** as the caller even though the call originates from `pipeline`, whose sub-agents do include `director`.

## Root cause

`handleTaskTransfer` resolves the calling agent via `r.CurrentAgent()`, which returns the runtime's **shared** current-agent field. Background sub-sessions are created with `PinAgent: true` + `WithAgentName(cfg.AgentName)` (`RunAgent` → `runCollecting`), which pins the **session** to `pipeline` but deliberately does not mutate `currentAgent` (it stays `root`). So validation runs against `root.SubAgents() = [pipeline]` and rejects `director`.

`agentRouter.ResolveSession` already handles pinned sessions ("when sess pins a specific agent (e.g. background agent tasks), that agent is returned directly"), but `handleTaskTransfer` wasn't using it.

## Fix

Resolve the caller via `resolveSessionAgent(sess)` instead of `CurrentAgent()`. This returns the pinned agent when the session pins one, and falls back to `CurrentAgent()` otherwise — so ordinary `transfer_task` behavior is unchanged.

## Tests

Adds `TestHandleTaskTransfer_UsesPinnedSessionAgent`: a session pinned to `pipeline` (as `run_background_agent` does) transfers to `director` while the runtime's current-agent is `root`.

- Without fix: fails with the exact error above
- With fix: passes

`go test ./pkg/runtime/` passes.

Closes #3886

### DIFF
--- a/pkg/runtime/agent_delegation.go
+++ b/pkg/runtime/agent_delegation.go
@@ -557,7 +557,7 @@ func (r *LocalRuntime) handleTaskTransfer(ctx context.Context, sess *session.Ses
 		return nil, fmt.Errorf("invalid arguments: %w", err)
 	}
 
-	a := r.CurrentAgent()
+	a := r.resolveSessionAgent(sess)
 	if errResult := validateAgentInList(a.Name(), params.Agent, "transfer task to", "sub-agents list", a.SubAgents()); errResult != nil {
 		return errResult, nil
 	}
--- a/pkg/runtime/agent_delegation_test.go
+++ b/pkg/runtime/agent_delegation_test.go
@@ -643,3 +643,53 @@ func TestTransferTask_PropagatesPermissions(t *testing.T) {
 	assert.Equal(t, []string{"safe_tool"}, parentClone.Allow,
 		"parent permissions must remain isolated from child mutations after transfer_task")
 }
+
+// TestHandleTaskTransfer_UsesPinnedSessionAgent verifies that transfer_task
+// inside a background sub-session resolves the caller from the session's
+// pinned agent, not the runtime's shared current-agent. Regression test for
+// the bug where a pipeline agent dispatched via run_background_agent could
+// not transfer_task to its own sub-agents because current-agent remained
+// the root coordinator.
+func TestHandleTaskTransfer_UsesPinnedSessionAgent(t *testing.T) {
+	t.Parallel()
+
+	childStream := newStreamBuilder().AddContent("done").AddStopWithUsage(10, 5).Build()
+	prov := &mockProvider{id: "test/mock-model", stream: childStream}
+
+	// root only knows pipeline; pipeline knows director.
+	director := agent.New("director", "Director agent", agent.WithModel(prov))
+	pipeline := agent.New("pipeline", "Pipeline agent", agent.WithModel(prov))
+	agent.WithSubAgents(director)(pipeline)
+	root := agent.New("root", "Root agent", agent.WithModel(prov))
+	agent.WithSubAgents(pipeline)(root)
+
+	tm := team.New(team.WithAgents(root, pipeline, director))
+	rt, err := NewLocalRuntime(t.Context(), tm,
+		WithSessionCompaction(false),
+		WithModelStore(mockModelStore{}),
+	)
+	require.NoError(t, err)
+
+	// background sub-session pinned to pipeline (as run_background_agent does).
+	sess := session.New(
+		session.WithUserMessage("Please proceed."),
+		session.WithAgentName("pipeline"),
+		session.WithToolsApproved(true),
+	)
+	evts := make(chan Event, 128)
+
+	toolCall := tools.ToolCall{
+		ID:   "call_1",
+		Type: "function",
+		Function: tools.FunctionCall{
+			Name:      "transfer_task",
+			Arguments: `{"agent":"director","task":"write a prompt","expected_output":"prompt text"}`,
+		},
+	}
+
+	result, err := rt.handleTaskTransfer(t.Context(), sess, toolCall, NewChannelSink(evts))
+	require.NoError(t, err)
+	require.NotNil(t, result)
+	assert.False(t, result.IsError,
+		"pipeline (pinned session) should be able to transfer to its sub-agent director; got: %s", result.Output)
+}


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is correct: replacing `r.CurrentAgent()` with `r.resolveSessionAgent(sess)` properly resolves the calling agent from the session's pinned agent in background sub-sessions, fixing the transfer_task regression described in #3886. The test correctly demonstrates the regression scenario. One low-severity note on a new edge case fallback is posted inline.

pkg/runtime/agent_delegation.go
```diff
@@ -557,7 +557,7 @@ func (r *LocalRuntime) handleTaskTransfer(ctx context.Context, sess *session.Ses
 		return nil, fmt.Errorf("invalid arguments: %w", err)
 	}
 
-	a := r.CurrentAgent()
+	a := r.resolveSessionAgent(sess)
```
docker-agent, line 560: **[low] Silent fallback to current agent when session `AgentName` is unrecognized**

`ResolveSession` falls back to `r.Current()` when `sess.AgentName` is non-empty but the named agent is not found in the team (e.g., if an agent was removed or renamed after the session was created). In that edge case, `handleTaskTransfer` will validate the transfer against the **runtime's current agent** instead of the session's intended pinned agent — which could either permit a transfer it should reject, or reject one it should permit.

This is a new code path introduced by this PR: before the change, `CurrentAgent()` was always used (sessions were never pinned), so the `sess.AgentName != ""` fallback didn't exist. In practice the risk is low because a session referencing a non-existent agent is already an erroneous state, and `Current()` is a reasonable fallback. Consider logging a warning when the named agent is not found (inside `ResolveSession` or here) so it's observable when this fallback fires.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

Sayt-0 (changes requested):
The validation fix is correct, but one in-scope issue remains before this can be approved.

`handleTaskTransfer` now resolves the caller from the pinned session. However, `runForwarding` immediately resolves `callerAgent` again from `r.currentAgentName()`, and `SwitchCurrentAgent: true` mutates the runtime's shared current-agent state.

For `root -> background pipeline -> director`, this means:

- validation correctly uses `pipeline`;
- `AgentSwitching`, hooks, and `SubSessionCompleted` are still attributed to `root`;
- the background goroutine temporarily changes the shared current agent to `director`, which can affect the foreground session or another concurrent background task.

A focused test against this commit confirmed that `AgentSwitching.FromAgent` is `root` instead of `pipeline`.

Could the session-resolved caller be carried into `runForwarding`, and could transfers from pinned sessions avoid mutating the shared `currentAgent` state, with tests covering:

1. switching and completion attribution to `pipeline`;
2. the shared current agent remaining `root`;
3. isolation during concurrent background delegation?

The cycle detection and delegation depth guard discussed in #3886 can be tracked as a separate follow-up and does not need to expand this PR.

aheritier:

Commits that are not verified: `f00a794`

yangtuooc:
I confirmed the issue raised in this review. Since #3905 has now been merged and covers both the original #3886 validation problem and the pinned-session isolation and attribution gap, would it be okay for me to close this PR as superseded?

aheritier:

PR #3901, opened 2026-08-03
fix(cli): match command guidance to invocation mode

Dandiggas:
## What changed

- add a shared helper that detects standalone versus Docker CLI plugin invocation
- update setup, doctor, model, alias, sandbox, environment, creator, and TUI guidance to use the runnable command for the current mode
- add regression coverage for both `docker-agent` and `docker agent` output

## Why

Standalone users were shown `docker agent` commands even when they launched the `docker-agent` binary directly. Command guidance was hardcoded across several sibling paths, so fixing only one message would leave the same usability bug elsewhere.

## Impact

User-facing commands now match the active invocation mode. Docker CLI plugin users continue to see `docker agent`, while standalone users see `docker-agent`.

## Validation

- `MODEL_RUNNER_HOST=http://127.0.0.1:59997 task dev`
- standalone built-binary smoke test: `docker-agent version dev`
- Docker CLI plugin smoke test: `docker agent version dev`

## Scope notes

- Only runnable command guidance is dynamic. Remaining `docker agent` strings in source are product-name prose (the tour copy, "docker agent env file", the board title) and are intentionally unchanged.
- Cobra `Example` fields stay hardcoded to `docker-agent`, matching the existing convention in `root.go` and `setup.go`. The dynamic guidance covers `Long` text, printed hints, and error messages.
- The `invocationError` sentinels and `pkg/creator/instructions.txt` share one placeholder token, `{{docker_agent_command}}`, resolved at render time.

### DIFF
--- a/cmd/root/alias.go
+++ b/cmd/root/alias.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/docker/docker-agent/pkg/cli"
+	"github.com/docker/docker-agent/pkg/cli/invocation"
 	"github.com/docker/docker-agent/pkg/config"
 	latestcfg "github.com/docker/docker-agent/pkg/config/latest"
 	pathx "github.com/docker/docker-agent/pkg/path"
@@ -201,9 +202,9 @@ func runAliasAddCommand(cmd *cobra.Command, args []string, flags *aliasAddFlags)
 	}
 
 	if name == "default" {
-		out.Printf("\nYou can now run: docker agent run %s (or even docker agent run)\n", name)
+		out.Printf("\nYou can now run: %s run %s (or even %s run)\n", invocation.DockerAgent(), name, invocation.DockerAgent())
 	} else {
-		out.Printf("\nYou can now run: docker agent run %s\n", name)
+		out.Printf("\nYou can now run: %s run %s\n", invocation.DockerAgent(), name)
 	}
 
 	return nil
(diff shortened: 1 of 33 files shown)


aheritier:
Hi @Dandiggas I am a bit worried about the complexity it adds in the codebase vs the limited impact of the change. I am not sure if many users are using `docker-agent` and not `docker agent`

Dandiggas:
Fair, and happy to split this into smaller PRs if the size is the problem.

I came across it following the install docs, which offer both. I took the standalone path, checked with `docker-agent version`, then the tour text pointed me at `docker agent getting-started`.

I'd assumed this was wanted because `Execute()` already rewrites every `Example` field based on `RunningStandalone()`, so I read it as the project having decided printed commands should match how the binary was invoked. This PR applies the same rule where a tree walk can't reach, since most of the guidance comes from `Printf` and `fmt.Errorf` rather than cobra fields.

The three regression tests assert both modes and use `NotContains` on the wrong form, so a regression in either fails CI rather than needing a pass over every call site.

Cutting scope would leave the same bug in whatever gets left out, but splitting is easy. Would that help?

aheritier:

Commits that are not verified: `25761d1`

aheritier:

Commits that are not verified: `25761d1`

Dandiggas:
Closing this for now. The scope concern is fair, and I'd rather not leave a 33-file diff open indefinitely.

If matching the printed guidance to the invocation mode is something you do want, happy to reopen it as a minimal PR covering a single command path plus its test.

PR #3907, opened 2026-08-04
feat(safety): broaden destructive-command taxonomy

trungutt:
## Why

The shell safety classifier's destructive list misses several command shapes that show up constantly in real sessions, so they fall into the "unknown" bucket instead of being flagged as destructive with a blast radius. Most visibly, the bare `docker rm <id>` shorthand is unclassified while `docker container rm <id>` is destructive — removing a (possibly stateful, running) container got a weaker signal than the long form of the exact same operation. When embedded in a host application, that means the confirmation UI shows a vague "potentially risky" hint instead of a clear destructive warning with severity.

## What

Pattern-list additions in `safety_patterns.json` (plus table-test coverage):

- **docker**: bare `docker rm <id>` and its `-f`/`--force`/`-v` shorthands; `docker stop <id>`; `docker compose rm -f`; the `docker builder prune` alias; long/short flag parity for existing prune entries (`--all` ↔ `-a`, including `docker volume prune -a/--all` which also drops *named* volumes)
- **git** (new section): `reset --hard`, `clean -f/-fd/-fdx` variants, `checkout -- <path>` / `checkout .` / `restore .`, `push --force`/`-f`, `branch -D`, `stash drop`/`clear`
- **filesystem**: `sed -i` in-place edits

No classifier code changes — gating semantics are untouched (unknown and destructive both prompt outside autonomous mode); this improves the label, blast radius, and reason shown to the user, and what custom permission rules can key on.

Also refreshes the taxonomy's self-description, which still referred to the deprecated `safer_shell` hook.


pkg/safety/safety_patterns.json
```diff
@@ -34,46 +34,80 @@
...
     ],
     "docker": [
       { "pattern": "docker volume rm <name>",                                "blast_radius": "HIGH",        "category": "dk-volume-del",    "notes": "irreversible data loss" },
       { "pattern": "docker volume prune",                                    "blast_radius": "MEDIUM",      "category": "dk-volume-del",    "notes": "unused only; can surprise" },
       { "pattern": "docker volume prune -f",                                 "blast_radius": "MEDIUM",      "category": "dk-volume-del",    "notes": "bypasses prompt" },
       { "pattern": "docker volume prune --force",                            "blast_radius": "MEDIUM",      "category": "dk-volume-del",    "notes": "long-form of -f" },
+      { "pattern": "docker volume prune -a",                                 "blast_radius": "HIGH",        "category": "dk-volume-del",    "notes": "includes named volumes, not just anonymous" },
+      { "pattern": "docker volume prune --all",                              "blast_radius": "HIGH",        "category": "dk-volume-del",    "notes": "long form of -a" },
```
docker-agent, line 71: **[medium] `docker volume prune -af` combined flag is not covered — silently kills all volumes without prompting**

The PR adds `-a`/`--all` flag parity for `docker volume prune` (line 62–63), and separately adds `-f`/`--force` entries (pre-existing). However, the combination `docker volume prune -af` (or `-fa`) — which destroys **all** named and anonymous volumes **without an interactive prompt** — has no matching pattern.

Because the classifier uses prefix-matching (`patternToRegexp` is not end-anchored), `docker volume prune -a` does NOT match `docker volume prune -af`: the pattern regex becomes `(?i)(?:^|.*\b)docker volume prune \-a(?:$|\b.*)`, and `-af` fails the `\b` word-boundary check after `-a` (both `-` and `a` and `f` collapse into one token `\-af` in the normalized command, so the pattern literal `-a` is not a word boundary split). Likewise `docker volume prune --all` does not match `-af`. The command therefore falls through to `ClassUnknown`, which gives the user a weaker "unknown risk" signal instead of the `HIGH` blast-radius destructive warning that `-af` deserves — it is arguably the most dangerous volume-prune invocation (no prompt, no scope limit).

**Suggested addition** (after the `docker volume prune --all` line):

```suggestion
      { "pattern": "docker volume prune -a",                                 "blast_radius": "HIGH",        "category": "dk-volume-del",    "notes": "includes named volumes, not just anonymous" },
      { "pattern": "docker volume prune --all",                              "blast_radius": "HIGH",        "category": "dk-volume-del",    "notes": "long form of -a" },
      { "pattern": "docker volume prune -af",                                "blast_radius": "HIGH",        "category": "dk-volume-del",    "notes": "force + all: destroys every volume without prompt" },
      { "pattern": "docker volume prune -fa",                                "blast_radius": "HIGH",        "category": "dk-volume-del",    "notes": "flag-order variant of -af" },
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

- [medium] pkg/safety/safety_patterns.json:45 — Missing `git clean -fX` and other flag-permutation variants that delete ignored files (confidence: weak 52/100)

melmennaoui (dismissed):

docker-agent (commented):

This PR broadens the destructive-command taxonomy with well-considered additions across docker, git, and filesystem patterns. The JSON structure is valid, the best-severity-wins classifier semantics are correctly leveraged, and the test coverage validates the core new entries. Four coverage gaps were found in the `+` lines:

1. `git push --force` regex prefix-matches `--force-with-lease` (the safer variant), producing a false positive destructive classification — the pattern note even acknowledges this.
2. `docker rm -vf <id>` (reversed flag order of the new `-fv` HIGH entry) is unclassified.
3. Four of the six three-flag permutations of `git clean -f/-d/-x` are missing.
4. `git restore <path>` (per-file restore) is absent despite `git restore .` being added.

pkg/safety/safety_patterns.json
```diff
@@ -34,46 +34,80 @@
...
+      { "pattern": "git clean --force",            "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "long form of -f" },
+      { "pattern": "git clean -fd",                "blast_radius": "HIGH",        "category": "git-discard",      "notes": "adds untracked directories" },
+      { "pattern": "git clean -df",                "blast_radius": "HIGH",        "category": "git-discard",      "notes": "flag-order variant" },
+      { "pattern": "git clean -fdx",               "blast_radius": "HIGH",        "category": "git-discard",      "notes": "adds ignored files (build caches, .env)" },
+      { "pattern": "git clean -xfd",               "blast_radius": "HIGH",        "category": "git-discard",      "notes": "flag-order variant" },
+      { "pattern": "git push -f",                  "blast_radius": "MEDIUM",      "category": "git-history",      "notes": "rewrites remote history" },
+      { "pattern": "git push --force",             "blast_radius": "MEDIUM",      "category": "git-history",      "notes": "long form of -f; --force-with-lease also matches" },
```
docker-agent: **[medium] `git push --force` pattern inadvertently matches `git push --force-with-lease`**

The regex generated for `"git push --force"` is `(?i)(?:^|.*\b)git push --force(?:$|\b.*)`. The trailing `(?:$|\b.*)` allows any continuation after `--force`, and since `-` is a non-word character, there is no word-boundary break between `--force` and `-with-lease`. As a result, `git push --force-with-lease origin main` matches this pattern and receives a MEDIUM destructive classification.

`--force-with-lease` is explicitly designed as the safer alternative to `--force` — it aborts if the remote ref has moved since the last fetch, preventing history rewrites caused by concurrent pushes. Flagging it at the same MEDIUM blast_radius as a bare `--force` is a false positive that will erode user trust in the safety classifier.

The note on the entry even acknowledges this: `"long form of -f; --force-with-lease also matches"` — which suggests this was known but accepted. If the intent is to flag `--force-with-lease` as destructive, the note should be a separate entry with lower blast_radius and its own rationale; if the intent is NOT to flag it, the pattern needs a word-boundary fix.

**Suggested fix** — add a word-boundary anchor after `--force` to prevent prefix-matching into `--force-with-lease`:

```suggestion
      { "pattern": "git push --force ",             "blast_radius": "MEDIUM",      "category": "git-history",      "notes": "long form of -f; does not match --force-with-lease" },
```

(Trailing space after `force` ensures the next character must be whitespace, which in the normalised-command form means the token ended.)

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/safety/safety_patterns.json
```diff
@@ -34,46 +34,80 @@
...
+      { "pattern": "docker rm -v <id>",                                      "blast_radius": "HIGH",        "category": "dk-container-del", "notes": "shorthand; also drops attached volumes" },
+      { "pattern": "docker rm -f <id>",                                      "blast_radius": "MEDIUM-HIGH", "category": "dk-container-del", "notes": "force-kills running container" },
+      { "pattern": "docker rm --force <id>",                                 "blast_radius": "MEDIUM-HIGH", "category": "dk-container-del", "notes": "long form of -f" },
+      { "pattern": "docker container rm -f <id>",                            "blast_radius": "MEDIUM-HIGH", "category": "dk-container-del", "notes": "force-kills running container" },
+      { "pattern": "docker container rm --force <id>",                       "blast_radius": "MEDIUM-HIGH", "category": "dk-container-del", "notes": "long form of -f" },
       { "pattern": "docker rm -fv <id>",                                     "blast_radius": "HIGH",        "category": "dk-container-del", "notes": "force-kill + volumes" },
```
docker-agent, line 82: **[medium] `docker rm -vf <id>` flag-order variant is unclassified (silently MEDIUM instead of HIGH)**

The PR adds `docker rm -fv <id>` at HIGH blast_radius, but omits the reverse flag order `docker rm -vf <id>`. Docker's CLI accepts flags in any order, so both forms are valid.

Tested with the actual regex engine: `docker rm -vf web` only matches `docker rm <id>` (MEDIUM) because no HIGH pattern covers the `-vf` order. The classifier's best-severity-wins logic cannot compensate — there is no HIGH-severity candidate to pick. A user running `docker rm -vf mycontainer` receives a MEDIUM prompt instead of HIGH, silently downgrading a force-kill + volume removal.

The pre-existing `docker rm -f -v <id>` (HIGH, expanded form) also does not cover `-vf` since it requires the space-separated form.

**Suggested fix** — add the missing flag-order variant alongside the existing `-fv` entry:

```suggestion
      { "pattern": "docker rm -fv <id>",                                     "blast_radius": "HIGH",        "category": "dk-container-del", "notes": "force-kill + volumes" },
      { "pattern": "docker rm -vf <id>",                                     "blast_radius": "HIGH",        "category": "dk-container-del", "notes": "flag-order variant of -fv" },
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/safety/safety_patterns.json
```diff
@@ -34,46 +34,80 @@
...
+      { "pattern": "git restore .",                "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "discards all worktree changes" },
+      { "pattern": "git clean -f",                 "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "deletes untracked files" },
+      { "pattern": "git clean --force",            "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "long form of -f" },
+      { "pattern": "git clean -fd",                "blast_radius": "HIGH",        "category": "git-discard",      "notes": "adds untracked directories" },
+      { "pattern": "git clean -df",                "blast_radius": "HIGH",        "category": "git-discard",      "notes": "flag-order variant" },
+      { "pattern": "git clean -fdx",               "blast_radius": "HIGH",        "category": "git-discard",      "notes": "adds ignored files (build caches, .env)" },
+      { "pattern": "git clean -xfd",               "blast_radius": "HIGH",        "category": "git-discard",      "notes": "flag-order variant" },
```
docker-agent, line 56: **[medium] Missing `git clean -fxd` and `git clean -dfx` flag-order variants at HIGH severity**

The PR adds four flag-order permutations of `git clean` with `-f`, `-d`, `-x` flags: `-fd`, `-df`, `-fdx`, `-xfd`. However, two valid permutations are absent:

- `git clean -fxd` — equivalent to `-fdx` (deletes untracked files + directories + ignored files)
- `git clean -dfx` — equivalent to `-fdx` and `-xfd`

Git accepts flags in any order. A user running `git clean -fxd` (a common tab-completion variant) will not match any HIGH pattern. Because the classifier's `git clean -f` pattern ends with a word boundary that correctly prevents it from matching multi-character flag clusters (confirmed by regex test: `\b` after `-f` blocks `-fd`, `-fxd`, etc.), there is no fallback — the command falls through to the unknown branch entirely.

Six permutations of three flags exist: `-fdx`, `-fxd`, `-dfx`, `-dxf`, `-xfd`, `-xdf`. The PR covers two; four remain unclassified.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/safety/safety_patterns.json
```diff
@@ -34,46 +34,80 @@
...
-      { "pattern": "chmod -R 777 <path>",          "blast_radius": "LOW",         "category": "fs-permissions",   "notes": "security regression, reversible" }
+      { "pattern": "chmod -R 777 <path>",          "blast_radius": "LOW",         "category": "fs-permissions",   "notes": "security regression, reversible" },
+      { "pattern": "sed -i",                       "blast_radius": "MEDIUM",      "category": "fs-modify",        "notes": "in-place file edit; -i.bak variants also match" }
+    ],
+    "git": [
+      { "pattern": "git reset --hard",             "blast_radius": "HIGH",        "category": "git-discard",      "notes": "discards uncommitted work; unrecoverable" },
+      { "pattern": "git checkout -- <path>",       "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "overwrites local modifications" },
+      { "pattern": "git checkout .",               "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "discards all worktree changes" },
+      { "pattern": "git restore .",                "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "discards all worktree changes" },
```
docker-agent, line 44: **[low] `git restore <path>` is not classified; only `git restore .` is covered**

The PR adds `git restore .` (MEDIUM-HIGH) but omits `git restore <path>`. Running `git restore src/main.go` discards uncommitted changes to that specific file — a common and destructive operation — yet it will receive no destructive classification and fall through to the unknown branch.

This is a narrower coverage gap than `git restore .` (which restores everything), but it affects the most common day-to-day form of the command. The analogous `git checkout -- <path>` entry (also added in this PR) does cover the path variant, so there is a precedent for classifying per-file restores.

**Suggested fix** — add a `git restore <path>` entry alongside `git restore .`:

```suggestion
      { "pattern": "git restore .",                "blast_radius": "MEDIUM-HIGH", "category": "git-discard",      "notes": "discards all worktree changes" },
      { "pattern": "git restore <path>",           "blast_radius": "MEDIUM",      "category": "git-discard",      "notes": "discards uncommitted changes to a specific path" },
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

pkg/safety/safety_patterns.json
trungutt: Added `docker volume prune -af` and `-fa` at HIGH in 8252370bc. One correction to the analysis: the command did not fall to `ClassUnknown` — the base `docker volume prune` pattern still prefix-matched it (verified against the real classifier), so it was destructive at MEDIUM instead of HIGH. Same fix either way.

pkg/safety/safety_patterns.json
trungutt: Added `docker rm -vf <id>` at HIGH in 8252370bc. Note the pre-fix behavior was MEDIUM via `docker rm <id>` (the `<id>` placeholder matches the `-vf` token), not unclassified — but the severity downgrade was real.

pkg/safety/safety_patterns.json
trungutt: Added all six three-flag permutations (`-fdx -fxd -dfx -dxf -xfd -xdf`) plus the two-flag `-fx`/`-xf` forms (force + ignored files, no dirs — same `.env`/cache blast radius) at HIGH in 8252370bc. Confirmed pre-fix that `git clean -fxd` fell through to unknown.

pkg/safety/safety_patterns.json
trungutt: Added `git restore <path>` at MEDIUM in 8252370bc. Caveat noted in the entry: `<path>` is `\S+`, so `git restore --staged <file>` (harmless unstage) also matches and gets a MEDIUM destructive label — the pattern syntax cannot exclude it. That errs toward prompting, which both destructive and unknown do anyway.

pkg/safety/safety_patterns.json
trungutt: The suggested fix is a no-op: patterns run through `normalizeCommand` before compilation, which strips the trailing space, and RE2 has no lookahead, so a JSON-only pattern cannot stop `git push --force` from prefix-matching `--force-with-lease` (the matcher appends `(?:$|\b.*)`).

What 8252370bc does instead: adds an explicit `git push --force-with-lease` entry *before* `--force`. On equal severity the classifier keeps the first match, so the user now sees the accurate reason naming the lease-guarded form. I kept it at MEDIUM deliberately — it still rewrites remote history (the lease only prevents clobbering unseen commits), and a lower blast_radius entry would be dead code because best-severity-wins would still surface the MEDIUM `--force` match. Also worth noting: `unknown` outranks `medium` in this classifier, so the alternative (leaving it unmatched) would be a *stronger* signal, not a gentler one.

pkg/safety/safety_patterns.json
docker-agent: Thanks for the fix! The entry looks good, and I appreciate the caveat note — you're right that the pattern syntax limitation means `git restore --staged <file>` (which only unstages, doesn't discard) will also match and get labeled MEDIUM destructive.

The trade-off is reasonable: erring toward prompting covers both the truly destructive `git restore <path>` and the harmless `--staged` variant, and since both behaviors already prompt anyway (destructive vs. unknown), there's no functional difference in the UX. The explicit note in the entry documents the limitation for future maintainers.

pkg/safety/safety_patterns.json
docker-agent: I appreciate the detailed explanation — you're absolutely right that my suggested fix was mechanically wrong. I misunderstood how `normalizeCommand` pre-processes the patterns before the `(?:$|\b.*)` suffix is applied, so the trailing space approach wouldn't have worked.

Your solution is clever: explicitly matching `--force-with-lease` first leverages the first-match-wins tie-breaking to surface the correct reason. I see now why you kept it at MEDIUM — since `--force` would still match and dominate in best-severity-wins aggregation, a lower severity would be invisible, and leaving it unmatched would indeed bump it to `unknown`, which is a stronger (and misleading) signal.

My original concern was that flagging `--force-with-lease` at MEDIUM would be a false positive, but your rationale that it still rewrites history (just with a safety check) is fair — the lease prevents *some* accidents but doesn't eliminate the destructive nature of the rewrite itself. Thanks for the context on the severity logic; I hadn't traced through the full match precedence. This is a solid fix.

melmennaoui (approved):

pkg/safety/safety_test.go
```diff
@@ -17,6 +17,23 @@ func TestClassifyCommand_DestructivePatterns(t *testing.T) {
...
+		{"docker builder prune", "medium"},
+		{"sed -i 's/a/b/' config.yml", "medium"},
+		{"git reset --hard origin/main", "high"},
+		{"git clean -fd", "high"},
+		{"git clean -fxd", "high"},
+		{"git clean -fx", "high"},
+		{"git checkout -- src/", "high"},
+		{"git restore src/main.go", "medium"},
+		{"git push --force origin main", "medium"},
+		{"git push --force-with-lease origin main", "medium"},
```
docker-agent, line 35: **[medium] 5 newly-added flag-order variant patterns lack test cases**

This PR adds the following patterns to `safety_patterns.json` but does not add corresponding test cases in the table-driven test:

- `git clean -xf` (line 51)
- `git clean -dfx` (line 54)
- `git clean -dxf` (line 55)
- `git clean -xdf` (line 57)
- `docker volume prune -fa` (line 73)

The classifier converts each JSON pattern string to a regex at runtime via `patternToRegexp` — there is no compile-time validation. A typo in any of these five pattern strings (wrong character order, extra space, missing dash) would silently produce a broken pattern that never matches the intended command, leaving it mis-classified as `ClassUnknown` with no destructive signal. Only a test case would catch it.

The PR does test `git clean -fxd`, `git clean -fx`, `docker volume prune -af`, and `docker rm -vf` from the new set, so the pattern is clearly established — the five omissions appear to be an oversight.

Consider adding:
```go
{"git clean -xf", "high"},
{"git clean -dfx", "high"},
{"git clean -dxf", "high"},
{"git clean -xdf", "high"},
{"docker volume prune -fa", "high"},
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
> This incremental review covers only the commits since `e84ce41a8271`.

### Assessment: 🟢 APPROVE

The new flag-order variants and pattern additions are correct and the blast-radius classifications are appropriate. One test coverage gap in the newly-added entries is noted below.

PR #3908, opened 2026-08-04
feat(filesystem): fail fast on foreign-OS paths and explain path resolution

trungutt:
## Why

In a real session on a Windows host, the model called `read_file` / `edit_file` with a WSL-style path (`/mnt/c/.../docker-compose.yml`). `filepath.IsAbs` doesn't recognize foreign-OS absolute paths, so the path was classified as relative and silently joined onto the working directory, failing with:

```
Error reading file: open <wd>\mnt\c\...\docker-compose.yml: The system cannot find the path specified.
```

Nothing told the model what happened — the tool descriptions and instructions never state the working directory, and even claimed "absolute paths … work as expected" — so it retried host paths for the rest of the session, including a blind `edit_file` with guessed `oldText`.

## What

Three changes, all aimed at letting the model self-correct in one turn:

1. **Fail fast on foreign-OS-shaped paths** — `/foo` (and rooted-no-drive `\foo`, drive-relative `C:foo`) on Windows, `C:\foo` on Unix — with an error that names the expected format. UNC paths are untouched; an explicit `./` prefix bypasses the check for the rare legitimate file with such a name. No translation is attempted (`/mnt/*` is only meaningful inside WSL, where GOOS=linux and the path is genuinely absolute).
2. **Resolution hint on not-found errors** (read_file, read_multiple_files, edit_file, list_directory): `not found (resolved to %q; relative paths resolve from the working directory %q)`. write_file is untouched — it creates parent directories, so the failure mode doesn't exist there.
3. **Instructions state the working directory** and the misleading "absolute paths and \"..\" work as expected" line is replaced with the positive rule (host-native format).

Because the instructions now embed a machine-specific path, the e2e cassette matcher normalizes the working-directory line (same approach as the existing `Instructions from:` prompt-file normalization), and the recorded request bodies were updated for the new instructions text — same as previous prompt-only changes (e.g. c94f3c81f).

Behavior note: anyone deliberately using `/foo` on Windows to address `<wd>\foo` now gets an error instead; `./foo` expresses that intent explicitly.

## Diagram

```
before:  /mnt/c/project/compose.yml
           └─ IsAbs? no → join(wd, path) → <wd>\mnt\c\project\compose.yml → "not found" (opaque)

after:   /mnt/c/project/compose.yml
           └─ IsAbs? no → looks POSIX-absolute on Windows
              → error: use C:\... or a path relative to the working directory
```


docker-agent (commented):
### Assessment: 🟢 APPROVE

The foreign-OS path detection logic is sound. `checkForeignPath` correctly handles all documented cases:
- POSIX absolute paths on Windows (`/foo`) → rejected (UNC `//` excluded)
- Rooted-no-drive paths on Windows (`\\foo`) → rejected (UNC `\\\\` excluded)
- Drive-relative paths on Windows (`C:foo`, `C:`) → rejected via `isDriveRelative`
- Windows absolute paths on Unix (`C:\\...`, `C:/...`) → rejected via `hasDrivePrefix`

The `./` bypass is implicit and correct: `./mnt/c/x` starts with `.`, not `/` or `\\`, so none of the detection cases fire.

`resolutionHint` is wired to all four affected tools (read_file, read_multiple_files, edit_file, list_directory) and correctly omitted from write_file. The `workingDirRegex` in `proxy.go` uses the right escape sequences for JSON-encoded cassette bodies and handles both Unix (`[^\\"\r\n]+`) and Windows (`\\\\` alternation) paths.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is correct. All new -line logic was verified:

- **`checkForeignPath`**: Correctly rejects foreign-OS paths in all tested cases. The `./` bypass works naturally via the existing logic (paths starting with `./` don't match any rejection condition). UNC paths pass through correctly. Drive-relative paths (`C:`, `C:foo`) are correctly gated on Windows while allowed on Unix (where they are valid relative names).
- **`resolutionHint`**: Called only after a successful `resolveAndCheckPath` (i.e., path passed all gates), so the resolved path is always valid. Correctly surfaces the joined path and working directory to the model.
- **`workingDirRegex`** in the cassette matcher: The pattern `The working directory is \\"(?:[^\\"\r\n]|\\\\\\\\)+\\"` correctly matches the JSON-encoded form of `%q`-formatted paths — including Windows paths with doubled backslashes and UNC paths.
- **`Instructions()`**: The `if t.workingDir != ""` guard prevents the working-directory line from appearing with an empty string; the `%q` format is consistent with what the cassette normalizer expects.
- **`ReadFileMeta.Path`**: The new field in the not-found error path is additive and correct.

melmennaoui (dismissed):

docker-agent (commented):
### Assessment: 🟢 APPROVE

melmennaoui (approved):

PR #3911, opened 2026-08-05
feat(shell): name the resolved interpreter in the shell tool description

trungutt:
## Why

The shell tool describes itself as *"Executes the given shell command in the user's default shell."* That tells the model nothing about which shell — or even which OS — will interpret the command, so models default to POSIX syntax.

On Windows the toolset actually resolves PowerShell (pwsh/powershell.exe, cmd.exe fallback), and in a real session on a Windows host this cost two wasted turns before the model found its footing:

```
turn 1:  pwd && ls -la   → parse error ('&&' is not valid in Windows PowerShell 5.1)
turn 2:  ls -la          → error (ls is a Get-ChildItem alias; -la is not a parameter)
turn 3:  dir             → finally works
```

The toolset already knows the exact resolved shell binary at construction time; it just never told the model.

## What

- The tool description now names the resolved interpreter and OS, e.g.
  - `Executes the given shell command with powershell on Windows. Use Windows PowerShell 5.1 syntax: chain commands with ";" (not "&&"), and avoid POSIX commands/flags like "ls -la".`
  - `Executes the given shell command with zsh on macOS.`
- Toolset instructions gain a matching first bullet: *Commands are executed by `<shell>` on `<OS>` — write them in that shell's syntax.*
- Shells commonly mistaken for POSIX (powershell, pwsh, cmd) carry an explicit dialect warning; Unix shells need only the name.

Same spirit as #3908: give the model the one machine fact it cannot guess, so it self-corrects in zero turns instead of two.

### DIFF
--- a/pkg/tools/builtin/shell/shell.go
+++ b/pkg/tools/builtin/shell/shell.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	goruntime "runtime"
 	"strings"
 	"sync"
 	"time"
@@ -359,13 +360,15 @@ func formatCommandOutput(timeoutCtx, ctx context.Context, err error, rawOutput s
 }
 
 func (t *ToolSet) Instructions() string {
-	return `## Shell Tools
+	return fmt.Sprintf(`## Shell Tools
 
+- Commands are executed by %s on %s — write them in that shell's syntax
 - Each call runs in a fresh shell session — no state persists between calls
 - Default timeout: 30s. Set "timeout" for longer operations (builds, tests)
 - Use "cwd" parameter instead of cd within commands
 - Combine operations with pipes, redirections, and heredocs
-- Non-zero exit codes return error info with output; timed-out commands are terminated`
+- Non-zero exit codes return error info with output; timed-out commands are terminated`,
+		shellBaseName(t.handler.shell), displayOS())
 }
 
 func (t *ToolSet) Tools(context.Context) ([]tools.Tool, error) {
@@ -373,7 +376,7 @@ func (t *ToolSet) Tools(context.Context) ([]tools.Tool, error) {
 		{
 			Name:                    ToolNameShell,
 			Category:                "shell",
-			Description:             `Executes the given shell command in the user's default shell.`,
+			Description:             shellToolDescription(t.handler.shell),
 			Parameters:              tools.MustSchemaFor[RunShellArgs](),
 			OutputSchema:            tools.MustSchemaFor[string](),
 			Handler:                 tools.NewRuntimeHandler(t.handler.RunShell),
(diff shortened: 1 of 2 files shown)


pkg/tools/builtin/shell/shell.go
```diff
@@ -399,3 +402,62 @@ func (t *ToolSet) Stop(context.Context) error {
...
+// commonly mistake for a POSIX shell. Empty for shells where the name
+// alone is enough of a cue.
+func shellSyntaxHint(name string) string {
+	switch name {
+	case "powershell":
+		// Windows PowerShell 5.1: '&&' / '||' are parse errors and
+		// POSIX flags don't exist ('ls -la' fails on the alias).
+		return `Use Windows PowerShell 5.1 syntax: chain commands with ";" (not "&&"), and avoid POSIX commands/flags like "ls -la".`
+	case "pwsh":
+		return `Use PowerShell syntax; POSIX commands/flags like "ls -la" are not available.`
```
docker-agent, line 430: **[low] `pwsh` syntax hint incorrectly states POSIX commands are unavailable on Linux/macOS**

`shellSyntaxHint` returns the same message for `pwsh` regardless of the host OS:

```go
case "pwsh":
    return `Use PowerShell syntax; POSIX commands/flags like "ls -la" are not available.`
```

PowerShell 7 (`pwsh`) is cross-platform. On Linux and macOS, `ls` is a native system binary — `ls -la` works fine inside a `pwsh` session because the shell resolves it via `$PATH` rather than its own `Get-ChildItem` alias. The combined description a model would see on Linux is:

> "Executes the given shell command with pwsh on Linux. Use PowerShell syntax; POSIX commands/flags like "ls -la" are not available."

…which is factually wrong and could cause the model to avoid valid commands unnecessarily. The `powershell` case (Windows PowerShell 5.1) correctly restricts POSIX flags, but that qualifier doesn't apply to `pwsh` on non-Windows hosts.

**Fix**: Condition the `pwsh` hint on the OS, or soften the wording. For example:

```go
case "pwsh":
    if goruntime.GOOS == "windows" {
        return `Use PowerShell syntax; POSIX commands/flags like "ls -la" are not available.`
    }
    return `Use PowerShell syntax.`
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/tools/builtin/shell/shell.go
```diff
@@ -399,3 +402,62 @@ func (t *ToolSet) Stop(context.Context) error {
...
+}
+
+// shellBaseName reduces a resolved shell path to a lowercase name the
+// model can recognize (C:\...\powershell.exe -> powershell, /bin/zsh -> zsh).
+// Splits on both separators instead of filepath.Base so the result is
+// deterministic regardless of the host OS the path came from.
+func shellBaseName(shellPath string) string {
+	base := shellPath
+	if i := strings.LastIndexAny(base, `/\`); i >= 0 {
+		base = base[i+1:]
```
docker-agent, line 445: **[low] `shellBaseName` returns an empty string when `shellPath` ends with a path separator**

```go
func shellBaseName(shellPath string) string {
    base := shellPath
    if i := strings.LastIndexAny(base, `/\`); i >= 0 {
        base = base[i+1:]   // if shellPath="/bin/bash/", i == len-1, so base = ""
    }
    return strings.ToLower(strings.TrimSuffix(base, filepath.Ext(base)))
}
```

If `shellPath` has a trailing separator (e.g. `"/bin/bash/"`), `LastIndexAny` finds the final character, so `base[i+1:]` becomes `""`. The function then returns `""`, and the tool description becomes `"Executes the given shell command with  on Linux."` — an empty shell name that leaves the model with no useful context.

`detectShell()` always produces well-formed paths in practice, so this is an unlikely trigger. A simple guard makes the defensive intent explicit and protects against misconfigured environments:

```go
func shellBaseName(shellPath string) string {
    base := shellPath
    if i := strings.LastIndexAny(base, `/\`); i >= 0 {
        base = base[i+1:]
    }
    if base == "" {
        return shellPath // fall back to the raw input rather than an empty string
    }
    return strings.ToLower(strings.TrimSuffix(base, filepath.Ext(base)))
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

Both findings are low-severity edge cases in the new helper functions. The core feature — naming the resolved interpreter and OS in the tool description — is well-structured and correct.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly names the resolved interpreter in both the tool description and `Instructions()` string. The logic is correct: `detectShell()` is called once, its result passed through `shellBaseName()` and `shellSyntaxHint()` to compose context-appropriate descriptions — PowerShell/pwsh/cmd paths get an explicit dialect warning, Unix shells get just the name. Tests cover the main new code paths (`TestShellTool_DescriptionNamesInterpreter`, `TestShellTool_Instructions`, `TestShellBaseName`). No bugs introduced by this PR.

melmennaoui (approved):

PR #3912, opened 2026-08-05
feat(filesystem): say when list_directory finds an empty directory

trungutt:
## Why

`list_directory` returns an empty `Output` for an empty directory, and the runtime substitutes a generic `(no output)` placeholder (`pkg/runtime/toolexec/dispatcher.go`). The model cannot tell "the listing succeeded and the directory is empty" from "the tool broke".

In a real session the working directory was an empty scratch dir: the model called `list_directory`, got `(no output)`, distrusted it, and re-listed the same directory through the shell — a wasted turn, and it *still* concluded "your directory is empty or I don't have access to it".

## What

When the listing produces no entries, say so explicitly instead of returning nothing:

- `Directory is empty: <resolved path>` — nothing there at all
- `Directory has no visible entries (N hidden by ignore patterns): <resolved path>` — entries exist but are all VCS/agents-ignore filtered, so the model doesn't go looking for files that are deliberately hidden

Including the resolved path doubles as a resolution hint, same as #3908 did for not-found errors.

### DIFF
--- a/pkg/tools/builtin/filesystem/filesystem.go
+++ b/pkg/tools/builtin/filesystem/filesystem.go
@@ -1052,6 +1052,18 @@ func (t *ToolSet) handleListDirectory(ctx context.Context, args ListDirectoryArg
 		}
 	}
 
+	// An empty Output is rendered as a generic "(no output)" placeholder
+	// upstream, which the model cannot distinguish from a tool failure and
+	// typically retries with a shell command. Say explicitly that the
+	// listing succeeded and the directory is empty.
+	if count == 0 {
+		if len(entries) == 0 {
+			fmt.Fprintf(&result, "Directory is empty: %s\n", resolvedPath)
+		} else {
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

docker-agent (commented):
### Assessment: 🟢 APPROVE

The change is correct and clean. When `list_directory` produces zero visible entries, the new `if count == 0` block now emits an explicit human-readable message instead of returning an empty string that the runtime replaces with a generic `(no output)` placeholder. The two-branch logic correctly distinguishes a truly empty directory (zero `entries`) from one that had entries but all were filtered by ignore patterns (non-zero `entries`, zero `count`), and both messages include the resolved path as a resolution hint — consistent with the pattern established in #3908. No correctness issues found.

melmennaoui (approved):

PR #3913, opened 2026-08-05
fix(filesystem): repair malformed inner JSON in double-serialized edits

trungutt:
## Why

`ParseEditFileArgs` already handles two model failure modes independently: a double-serialized `edits` field (#2144), and brace/bracket-counting mistakes in the outer JSON (repaired via `tryRepairEditFileJSON`). But when both happen at once, parsing fails.

A production session hit exactly that combination — the model sent `edits` as a JSON string *and* miscounted braces inside it:

```json
{"edits": "[{\"oldText\": \"...\", \"newText\": \"...\"}}]", "path": "docker-compose.yml"}
```

The unwrap succeeded, but the inner payload (`...}}]` — extra closing brace) failed with:

```
Error calling tool: failed to parse double-serialized edits string: invalid character '}' after array element
```

The model then recovered by reading the whole file and rewriting it — a full extra round-trip that the existing repair logic was designed to avoid.

## What

Give the unwrapped inner payload the same repair pass as the outer JSON. Genuinely unparseable payloads still error as before.

### DIFF
--- a/pkg/tools/builtin/filesystem/filesystem.go
+++ b/pkg/tools/builtin/filesystem/filesystem.go
@@ -397,10 +397,10 @@ func ParseEditFileArgs(data []byte) (EditFileArgs, error) {
 		if err := json.Unmarshal(repaired, &args.Edits); err != nil {
 			return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string after repair: %w", err)
 		}
+		slog.Debug("Repaired malformed double-serialized edits payload")
 		if err := validateRepairedEdits(args.Edits); err != nil {
 			return EditFileArgs{}, err
 		}
-		slog.Debug("Repaired malformed double-serialized edits payload")
 	}
 
 	return args, nil
--- a/pkg/tools/builtin/filesystem/filesystem.go
+++ b/pkg/tools/builtin/filesystem/filesystem.go
@@ -397,10 +397,27 @@ func ParseEditFileArgs(data []byte) (EditFileArgs, error) {
 		if err := json.Unmarshal(repaired, &args.Edits); err != nil {
 			return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string after repair: %w", err)
 		}
+		if err := validateRepairedEdits(args.Edits); err != nil {
+			return EditFileArgs{}, err
+		}
 		slog.Debug("Repaired malformed double-serialized edits payload")
 	}
 
 	return args, nil
+}
+
+// validateRepairedEdits guards against repair output that is structurally
+// valid but semantically corrupted. An empty oldText is never a meaningful
+// edit (handleEditFile's strings.Replace would silently insert newText at
+// the start of the file), so its presence after a repair means the repair
+// removed a load-bearing character rather than a spurious one.
+func validateRepairedEdits(edits []Edit) error {
+	for i, edit := range edits {
+		if edit.OldText == "" {
+			return fmt.Errorf("repaired edits payload is invalid: edit %d has empty oldText", i+1)
+		}
+	}
+	return nil
 }
 
 // tryRepairEditFileJSON attempts to fix common LLM JSON malformations by
--- a/pkg/tools/builtin/filesystem/filesystem.go
+++ b/pkg/tools/builtin/filesystem/filesystem.go
@@ -388,7 +388,16 @@ func ParseEditFileArgs(data []byte) (EditFileArgs, error) {
 		return EditFileArgs{}, fmt.Errorf("edits field is neither an array nor a JSON string: %w", err)
 	}
 	if err := json.Unmarshal([]byte(editsStr), &args.Edits); err != nil {
-		return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string: %w", err)
+		// The inner payload can carry the same brace/bracket-counting
+		// mistakes as the outer JSON, so give it the same repair pass.
+		repaired, ok := tryRepairEditFileJSON([]byte(editsStr))
+		if !ok {
+			return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string: %w", err)
+		}
+		if err := json.Unmarshal(repaired, &args.Edits); err != nil {
+			return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string after repair: %w", err)
+		}
+		slog.Debug("Repaired malformed double-serialized edits payload")
 	}
 
 	return args, nil
--- a/pkg/tools/builtin/filesystem/filesystem_test.go
+++ b/pkg/tools/builtin/filesystem/filesystem_test.go
@@ -587,6 +587,22 @@ func TestParseEditFileArgs(t *testing.T) {
 			wantErrMsg: "failed to parse double-serialized edits string",
 		},
 		{
+			name:     "repair: double-serialized with extra closing brace in inner payload",
+			input:    `{"edits": "[{\"oldText\": \"a\", \"newText\": \"b\"}}]", "path": "docker-compose.yml"}`,
+			wantPath: "docker-compose.yml",
+			wantEdits: []Edit{
+				{OldText: "a", NewText: "b"},
+			},
+		},
+		{
+			name:     "repair: double-serialized with extra closing bracket in inner payload",
+			input:    `{"path": "f.go", "edits": "[{\"oldText\": \"a\", \"newText\": \"b\"}]]"}`,
+			wantPath: "f.go",
+			wantEdits: []Edit{
+				{OldText: "a", NewText: "b"},
+			},
+		},
+		{
 			name:     "missing edits field (partial/streaming args)",
 			input:    `{"path": "/tmp/test.txt"}`,
 			wantPath: "/tmp/test.txt",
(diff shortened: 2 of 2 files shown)


pkg/tools/builtin/filesystem/filesystem.go
```diff
@@ -388,7 +388,16 @@ func ParseEditFileArgs(data []byte) (EditFileArgs, error) {
 		return EditFileArgs{}, fmt.Errorf("edits field is neither an array nor a JSON string: %w", err)
 	}
 	if err := json.Unmarshal([]byte(editsStr), &args.Edits); err != nil {
-		return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string: %w", err)
+		// The inner payload can carry the same brace/bracket-counting
+		// mistakes as the outer JSON, so give it the same repair pass.
+		repaired, ok := tryRepairEditFileJSON([]byte(editsStr))
```
docker-agent, line 393: **[medium] `tryRepairEditFileJSON` applied to inner edits array may silently corrupt valid JSON by removing structural delimiters**

`tryRepairEditFileJSON` was designed for the outer `EditFileArgs` object (e.g. `{"path":..., "edits":[...]}}` with a trailing extra `}`). When called on the unwrapped inner edits array string (e.g. `[{"oldText":"...", "newText":"..."}}]`), the heuristic is the same — it removes whichever character at the `json.SyntaxError` offset is `}`, `]`, or `\`. In the case of a trailing extra delimiter this is harmless and correct.

The concern is a subtler malformation: if the edit payload's syntax error is reported at an offset that happens to be a structurally important `}` — e.g. the parser has already consumed a prior syntax error token through normal tolerance, and the "extra" character is actually closing a nested edit — removing it can produce JSON that is syntactically valid but semantically wrong (e.g. the last edit object's `newText` key quietly merges with the next object, or an object is silently dropped). The function has no way to distinguish a spurious trailing delimiter from a load-bearing interior one when operating on an array rather than the top-level object it was designed for.

The test cases added cover the straightforward trailing-extra-delimiter pattern (`}}]` and `}]]`). A case like `[{"oldText": "x"}}, {"newText": "y"}]` (malformation inside the array, not at the end) is not tested and could produce a different repair outcome.

**Suggested mitigation**: After calling `json.Unmarshal(repaired, &args.Edits)` succeeds, consider checking that each `Edit` has non-empty `OldText` and `NewText`, or add a test covering an interior malformation to validate repair does not silently drop an edit.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

pkg/tools/builtin/filesystem/filesystem.go
```diff
@@ -388,7 +388,16 @@ func ParseEditFileArgs(data []byte) (EditFileArgs, error) {
...
+		if !ok {
+			return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string: %w", err)
+		}
+		if err := json.Unmarshal(repaired, &args.Edits); err != nil {
```
docker-agent, line 397: **[low] Repaired inner edits payload has no semantic validation — silently corrupted edits could be applied to files**

After `tryRepairEditFileJSON` succeeds and `json.Unmarshal(repaired, &args.Edits)` passes, the resulting `args.Edits` slice is returned and used to patch files. The repair is purely structural: it removes stray `}`, `]`, or `\` characters at syntax-error offsets (up to 3 iterations) and accepts any result that satisfies `json.Valid`. No check is made that each `Edit` in the repaired slice has non-empty `OldText` and `NewText`.

In the unlikely but possible scenario where removing a character produces structurally valid JSON with an edit whose `OldText` is empty (e.g. the closing `"` of `oldText`'s value was removed and the remaining bytes still parse), the edit would apply a full file overwrite (empty `oldText` matches anything) or silently be a no-op — both unintended outcomes.

This is a pre-existing design constraint of the outer repair path too, but the extension to the inner array payload is introduced by this PR. Given the 3-iteration cap and the limited character set (`}`, `]`, `\`), catastrophic corruption is unlikely in practice; this is a defence-in-depth concern rather than an immediately exploitable bug.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 65/100 |

docker-agent (commented):

pkg/tools/builtin/filesystem/filesystem.go
trungutt: Fair point. Addressed in 505230f66:

- Added the interior-malformation test you suggested: an extra `}` **between** array elements repairs cleanly to two intact edits.
- Added a guard (`validateRepairedEdits`) so that when a repair produces a structurally valid array containing an edit with empty `oldText` — the signature of a load-bearing character having been removed — the payload is rejected instead of applied. Rejection matches pre-PR behavior for those inputs (they failed to parse at all), so nothing regresses.

Note `newText` may legitimately be empty (deletion), so the guard only checks `oldText`.

pkg/tools/builtin/filesystem/filesystem.go
trungutt: Addressed for the path this PR introduces in 505230f66: repaired inner payloads are now validated and rejected if any edit has an empty `oldText`.

One correction to the scenario: an empty `oldText` would not overwrite the file — `strings.Replace(s, "", new, 1)` inserts at offset 0 — so the failure mode is a silent prepend, not a full overwrite. Still unintended, hence the guard.

As you note, the outer repair path has the same design constraint and pre-dates this PR; happy to extend the same validation there in a follow-up.

pkg/tools/builtin/filesystem/filesystem.go
docker-agent: I read your changes and they look excellent — both the test and the guard address the concern perfectly.

The interior-malformation test demonstrates that the repair heuristic already handles the scenario I was worried about correctly, and `validateRepairedEdits` ensures that if a future input *does* trigger pathological behavior (a structurally important delimiter removed, leaving an edit with empty `oldText`), it fails loudly instead of corrupting data. The note about `newText` being legitimately empty for deletions is spot on.

Thank you for the thorough fix!

pkg/tools/builtin/filesystem/filesystem.go
docker-agent: Thank you for the fix and the correction on the failure mode — you're absolutely right that `strings.Replace(s, "", new, 1)` would prepend rather than overwrite. I should have traced that through more carefully.

I've confirmed the guard in 505230f66 addresses the risk for the inner repair path introduced by this PR. Given that the outer repair path has the same constraint and pre-dates this change, handling it in a follow-up makes sense and keeps the scope of this PR focused.

docker-agent (commented):
### Assessment: 🟢 APPROVE

> This review covers only the commits pushed since `fe9949095ca3`.

The new `validateRepairedEdits` guard is correctly placed and logically sound. It is called immediately after a successful repair+unmarshal, before returning — exactly the right insertion point. The 1-based index in the error message (`edit %d` using `i+1`) is correct for human-readable output.

The one asymmetry raised during analysis (the guard only applies to the repaired path, not the first-attempt double-serialized path) is pre-existing and not introduced by this PR. The PR's stated scope is the double-serialized + malformed-inner-JSON combination; applying the guard to that specific path is the correct, minimal change. No new bugs were introduced by the changes since `fe9949095ca3`.

melmennaoui (dismissed):

pkg/tools/builtin/filesystem/filesystem.go
```diff
@@ -388,12 +388,38 @@ func ParseEditFileArgs(data []byte) (EditFileArgs, error) {
...
+			return EditFileArgs{}, fmt.Errorf("failed to parse double-serialized edits string after repair: %w", err)
+		}
+		if err := validateRepairedEdits(args.Edits); err != nil {
+			return EditFileArgs{}, err
+		}
+		slog.Debug("Repaired malformed double-serialized edits payload")
```
docker-agent: **[low] `slog.Debug` is never emitted when `validateRepairedEdits` rejects the repair — silent failure in logs**

When `validateRepairedEdits` returns an error (line 400–402 early-returns), the `slog.Debug("Repaired malformed double-serialized edits payload")` on line 403 is never reached. A repair that produced semantically invalid output (e.g. an edit with empty `oldText`) leaves no debug trace — the caller receives the error message, but there is nothing in the log to confirm that a repair was even attempted for that payload.

Moving the log line before the `validateRepairedEdits` call, or adding a `slog.Warn` on the validation-failure return path, would make repair attempts always observable:

```suggestion
		slog.Debug("Repaired malformed double-serialized edits payload")
		if err := validateRepairedEdits(args.Edits); err != nil {
			return EditFileArgs{}, err
		}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The inner-repair path is a clean extension of the existing outer-repair pattern, and `validateRepairedEdits` guards against the one realistic semantic corruption scenario (repair removing a load-bearing character and leaving an edit with empty `oldText`). One minor observability issue noted inline.

pkg/tools/builtin/filesystem/filesystem.go
trungutt: Applied your suggestion in 523c28083 — the debug log now precedes validation, so repair attempts are observable even when the repaired payload is rejected. (The log is accurate in that position: structural repair did succeed; validation is a separate semantic check whose failure is already surfaced in the returned error.)

docker-agent: Thank you for applying the fix! The log now correctly reflects that structural repair succeeded independently of the semantic validation step, which makes debugging much clearer.

melmennaoui (approved):

PR #3916, opened 2026-08-05
fix(server): contain API session working directories

Sayt-0:
## Summary

- add a dedicated opt-in `--session-workingdir-root` boundary for session working directories supplied through the HTTP API
- canonicalize configured roots and candidate directories before rejecting traversal, sibling-directory, and symlink escapes
- preserve unrestricted local-daemon behavior when no root is configured, without treating `--working-dir` or the process cwd as an implicit security boundary
- document the deployment model and add server and CLI regression coverage

Security alert: https://github.com/docker/docker-agent/security/code-scanning/57

## Alert expectations

| --- | --- |
| Prevent untrusted API callers from selecting arbitrary host paths | `WithSessionWorkingDirRoot` constrains `POST /api/sessions` working directories when configured |
| Reject traversal and symlink escapes | Both root and candidate are resolved with `filepath.EvalSymlinks`; containment uses `filepath.Rel` and `filepath.IsLocal` |
| Avoid the behavior regression reverted by #3788 | The default remains unrestricted and neither `runConfig.WorkingDir` nor the process cwd becomes a boundary |
| Support API and attached-run control planes | `serve api` exposes the flag; `run --listen` accepts the same hidden advanced flag |
| Keep the intentional unrestricted flow explicit to CodeQL | The remaining unrestricted sink has a rule-specific local suppression with the compatibility rationale |

## Validation

- `task build`
- `task lint`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy -u ALL_PROXY -u all_proxy task test`
- `npx --yes markdownlint-cli2@0.22.1 features/api-server/index.md features/cli/index.md`
- `./scripts/docs-check-canonical.sh`
- `./scripts/docs-check-llms-txt.sh`
- offline documentation link check with Lychee

The proxy variables are removed for the test command because the local development proxy intercepts private-address SSRF fixtures; the same failing packages pass without that interception.

## Security and compatibility notes

- Containment is opt-in so long-lived local daemons can continue opening arbitrary user-selected workspaces. Multi-user or network-exposed deployments should configure the new root.
- The configured root should be controlled by the operator. Filesystem namespace changes between canonicalization and later tool use remain subject to normal TOCTOU constraints when untrusted principals can mutate the root.
- Existing HTTP error classification is unchanged.
- Alert closure depends on the CodeQL analysis for this branch.


aheritier:

Commits that are not verified: `a948b84`

ronan-thibaut-glitch (approved):

PR #3920, opened 2026-08-06
fix: frontend-agnostic context overflow message, export recorded errors for recovery

trungutt:
### What this PR does

Two fixes motivated by a diagnostics report where a run died with a fatal context overflow:

1. **Make the token-overflow message frontend-agnostic.** The fatal overflow error told users to "Try running /compact", but this error travels on the runtime event stream and reaches every consumer (API/serve clients, embedding hosts), not all of which expose a `/compact` command. It also claimed "automatic compaction is not enabled" even when compaction *was* enabled but its overflow-retry budget was exhausted. The message now names the action ("Compact the conversation…") instead of a command and drops the wrong claim.

2. **Include recorded errors in the session recovery export.** Recorded error items (persisted since v1.100.0) are filtered out by `GetAllMessages()`, so `ExportSessionForRecovery` produced a transcript that just stops mid-conversation with no trace of the failure that ended it. Diagnosing the report above required manually reproducing which error the user saw. The export now carries a separate `errors` key (omitted when the session has no recorded errors, so the shape is unchanged for the common case).

### DIFF
--- a/pkg/modelerrors/modelerrors.go
+++ b/pkg/modelerrors/modelerrors.go
@@ -635,9 +635,14 @@ func FormatError(err error) string {
 	case OverflowKindMedia:
 		return "An image or file in this conversation is too large for the AI provider. " +
 			"Try a smaller file or remove it from context."
+	// Frontend-agnostic by design too: this error reaches every consumer of
+	// the runtime event stream, and not all of them expose a /compact
+	// command, so name the action ("compact") rather than a command. It also
+	// avoids claiming compaction "is not enabled": this path is reached both
+	// when compaction is disabled and when its retries are exhausted.
 	case OverflowKindTokens:
-		return "The conversation has exceeded the model's context window and automatic compaction is not enabled. " +
-			"Try running /compact to reduce the conversation size, or start a new session."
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The two focused fixes in this PR are correct and well-tested:

**1. Frontend-agnostic overflow message** (): The updated text removes the incorrect claim that "automatic compaction is not enabled" (which was wrong when compaction was enabled but had exhausted its retry budget) and replaces the CLI-specific "/compact" command reference with a generic action description. Clean, correct change.

**2. Export recorded errors for recovery** (, ): `GetAllErrors` correctly snapshots items under the read lock via `snapshotItems()`, iterates error items, and recurses into sub-sessions. The conditional `if errs := sess.GetAllErrors(); len(errs) > 0` correctly omits the `errors` key when empty, preserving the existing export shape for sessions with no recorded errors. The implementation is safe and consistent with the existing `GetAllMessages` pattern.

Test coverage is thorough: both the error-present and error-absent export cases are exercised, plus the new `GetAllErrors` unit test validates ordering and sub-session recursion.

melmennaoui (approved):

PR #3921, opened 2026-08-06
test(cache): fix Windows flake in the torn-file concurrency test

trungutt:
### What this PR does

Fixes a Windows-only CI flake in `TestFileCache_concurrentStoreNeverYieldsTornFile` (seen e.g. in [this run](https://github.com/docker/docker-agent/actions/runs/31082255105/job/92553513516)):

```
open C:\...\cache.json: The process cannot access the file because it is being used by another process.
```

The test races a reader loop (`os.ReadFile`) against `Store`'s temp-file+rename publishes to prove readers never observe a torn write. On POSIX that's a clean race: rename is atomic and the destination stays openable. On Windows, `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` can make the destination transiently unopenable, so the reader occasionally gets `ERROR_SHARING_VIOLATION` — a timing artifact of the writer itself, not a torn write.

The reader loop already tolerated `os.ErrNotExist` (file not published yet); it now also tolerates the Windows sharing violation, via a build-tagged test helper following the existing `isLockUnavailable` pattern (`pkg/tools/builtin/plan/lock_windows.go`). The test's actual assertion — successfully read bytes always parse as valid JSON — is unchanged.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -260,7 +260,11 @@ func TestFileCache_concurrentStoreNeverYieldsTornFile(t *testing.T) {
 	// half-written content and json.Unmarshal would error.
 	for range 100 {
 		data, err := os.ReadFile(path)
-		if errors.Is(err, os.ErrNotExist) {
+		// Tolerate transient open failures caused by the writer itself:
+		// the file may not exist yet, and on Windows the open can race a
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

Sayt-0 (approved):

PR #3949, opened 2026-08-07
feat: add startup banner to normal TUI

rumpl:
## Summary
- share the Docker Agent ASCII banner between the lean and normal TUIs
- show it as a centered empty state when an agent has no welcome message
- keep configured welcome messages unchanged

## Validation
- `task test`
- `task lint`
- `task build`


pkg/tui/page/chat/chat.go
```diff
@@ -230,6 +231,7 @@ type chatPage struct {
 	// Track whether we've received content from an assistant response
 	// Used by --exit-after-response to ensure we don't exit before receiving content
 	hasReceivedAssistantContent bool
+	showStartupBanner           bool
```
docker-agent, line 234: **[low] `showStartupBanner` initializes to `false`, so banner is invisible until `AgentInfoEvent` fires**

`showStartupBanner` is not set in `New()` — it starts as `false` (Go zero value). The `messagesView()` logic at line 730 returns early when `!p.showStartupBanner`, so the banner will never appear in the window between TUI creation and the first `AgentInfoEvent`. In practice this gap may be imperceptible for fast-starting agents, but on slow networks or in session-restore paths users will see a blank screen instead of the startup banner.

Consider initializing `showStartupBanner = true` in `New()` and setting it to `false` only once a non-empty `WelcomeMessage` is confirmed:

```suggestion
	showStartupBanner:           true,
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

- [low] pkg/tui/page/chat/runtime_events.go:124 — AgentInfoEvent re-enabling showStartupBanner can flash banner mid-conversation during agent/model switches (confidence: weak 52/100)

pkg/tui/page/chat/chat.go
rumpl: Fixed in 4b2fd8b1e. Chat pages now start with the banner enabled, so it is visible before AgentInfoEvent arrives. A non-empty welcome message disables it, while later empty AgentInfo events do not reactivate it mid-conversation. Added regression tests for both cases.

docker-agent: Perfect! The fix addresses all three aspects:

1. ✅ `showStartupBanner: true` in `New()` ensures the banner is visible immediately
2. ✅ The `AgentInfoEvent` handler now only disables it when a non-empty welcome message arrives (instead of unconditionally setting it based on message emptiness)
3. ✅ The regression tests cover both the initial-state visibility and the mid-conversation non-reactivation edge case

The logic change from `p.showStartupBanner = msg.WelcomeMessage == ""` to the conditional assignment is particularly nice — it prevents the banner from flickering back on if a later `AgentInfo` event arrives with an empty message after the conversation has started. Thanks for the thorough fix!

Sayt-0 (approved):

aheritier (approved):

PR #3950, opened 2026-08-07
feat(tui): render LaTeX in markdown

rumpl:
## Summary

- add a dependency-free LaTeX-to-Unicode terminal renderer
- render inline and display math in TUI Markdown
- preserve unsupported, incomplete, currency, shell-variable, and code content
- cover renderer output with golden tests and Markdown integration tests

## Validation

- `task format`
- `task test` (with `MOONSHOT_API_KEY` unset to skip the opt-in live API test)
- `task lint`
- `task build`


docker-agent (commented):

One NOTABLE finding (medium severity, LIKELY confirmed) was identified in the new inline LaTeX rendering code.

pkg/tui/components/markdown/fast_renderer.go
```diff
@@ -1912,12 +1980,88 @@ func hasInlineMarkdown(text string) bool {
...
+		opening, closing = "$$", "$$"
+	case strings.HasPrefix(source, "$") && (len(source) == 1 || !unicode.IsSpace(rune(source[1]))):
+		opening, closing = "$", "$"
+	default:
+		return "", 0, false
+	}
+
+	end := findUnescapedDelimiter(source, closing, len(opening))
+	if end < 0 {
+		return source, len(source), looksLikePendingMath(source[len(opening):]) || opening[0] == '\\'
```
docker-agent: **[medium] Unclosed `\(` or `\[` inline math consumes the rest of the line, suppressing further markdown processing**

In `renderInlineLatex`, when `\(` or `\[` is encountered but no matching closing delimiter is found on the same line, line 2006 returns:

```go
return source, len(source), looksLikePendingMath(source[len(opening):]) || opening[0] == '\\'
```

Because `opening[0] == '\\'` is always `true` for these delimiters, the function returns the **entire remaining text** (`source = text[i:]`) as a flat rendered string with `consumed = len(source)`. The caller in `renderInlineWithStyleTo` then advances `i` to the end of `text`, skipping all further inline markdown processing.

**Impact**: Any inline markdown appearing after an unclosed `\(` or `\[` on the same line is silently swallowed — e.g., in `See \( x more examples and **note** this`, the `**note**` bold would never be rendered; only `\( x more examples and **note** this` would be emitted as flat text.

**Suggested fix**: Instead of consuming the entire remaining slice, consume only up to the next whitespace boundary or limit the "pending math" heuristic to `\(` only (not `\[`, which is a display-math delimiter unlikely to appear inline), and return `recognized=false` when no reasonable boundary can be found:

```go
if end < 0 {
    if looksLikePendingMath(source[len(opening):]) {
        return source, len(source), true
    }
    return "", 0, false
}
```

This restricts the "eat the rest of the line" behavior to cases where the content actually looks like pending math (contains `_`, `^`, `=`, etc.), rather than blindly consuming for any `\(` or `\[`.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

rumpl: Fixed in 79a3aad76. Unclosed delimiters now fall back to normal inline parsing when the remainder contains Markdown formatting, with regression coverage for bold text after both unclosed `\\(` and `\\[` delimiters.

docker-agent: Excellent fix! The logic is exactly right: when an unclosed delimiter is encountered, you now check both whether it looks like pending math AND whether the remainder contains markdown formatting. If markdown is detected, you return recognized=false to let normal inline parsing continue, preserving bold, code, strikethrough, etc.

The regression tests are perfect — they verify that both unclosed delimiters preserve bold markdown in the remainder, and they confirm the ANSI bold sequence is actually emitted. This prevents the issue from recurring.

Nice work!

Sayt-0 (approved):

aheritier (approved):

PR #3956, opened 2026-08-09
Use typescript types for codemode

rumpl:


pkg/tools/codemode/functions.go
```diff
@@ -3,40 +3,290 @@ package codemode
...
 
-	doc.WriteString(toComment(&tool))
-	fmt.Fprintf(&doc, "function %s(args: Input): Output { ... }\n", tool.Name)
+	if isObjectSchema(input) {
+		fmt.Fprintf(&doc, "interface %s %s\n\n", inputName, objectType(input, input, 0))
+	} else {
+		fmt.Fprintf(&doc, "type %s = %s;\n\n", inputName, schemaType(input, input, 0))
+	}
+	fmt.Fprintf(&doc, "type %s = %s;\n\n", outputName, schemaType(output, output, 0))
+	fmt.Fprintf(&doc, "declare function %s(args: %s): %s;\n", tool.Name, inputName, outputName)
```
docker-agent: **[medium] `tool.Name` used unsanitized in `declare function` — may produce invalid TypeScript**

The `declare function` line emits `tool.Name` verbatim, but TypeScript function names must be valid identifiers. If a tool is named with hyphens (e.g. `search-items`), dots, spaces, or starts with a digit, the output is syntactically invalid TypeScript — `declare function search-items(...)` would be parsed as a subtraction expression, not a declaration.

`baseName` (i.e. `typeName(tool.Name)`) is already computed on line 18 and correctly used for `inputName`/`outputName`, but it is not used for the function name itself. The fix is a one-character change:

```suggestion
	fmt.Fprintf(&doc, "declare function %s(args: %s): %s;\n", baseName, inputName, outputName)
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

pkg/tools/codemode/functions.go
rumpl: Fixed in e6b355bc1. The declaration now uses the sanitized PascalCase name, and code mode registers the same name in the JavaScript runtime so the advertised function is callable. The original tool name remains registered for compatibility.

docker-agent: Confirmed — the fix correctly uses `baseName` for the function declaration, ensuring tools with non-identifier names (hyphens, dots, etc.) produce valid TypeScript. Registering the sanitized name in the runtime so it's callable is the right approach; maintaining the original name for backward compatibility is a good touch.

Thanks for the quick fix!

Sayt-0 (approved):

PR #3958, opened 2026-08-09
feat: persist generic session attributes

rumpl:
## Summary

- add concurrency-safe, namespaced string attributes to sessions
- persist attributes in SQLite and expose independent copies in lightweight summaries
- preserve attributes across JSON round trips, metadata updates, clones, branches, and derived sessions

## Compatibility

- add forward migration 026 with an empty JSON object default
- treat legacy NULL and empty attribute values as empty metadata

## Testing

- `go test ./pkg/session/...`
- `go test -race ./pkg/session/...`
- `go test ./pkg/server/... ./pkg/runtime/...`
- `task lint`
- `task build`


docker-agent (commented):
### Assessment: 🟢 APPROVE

Sayt-0 (approved):

PR #3989, opened 2026-08-15
feat(runtime): resume external harness sessions

rumpl:
## Summary

- capture external harness session IDs and persist them with docker-agent sessions
- resume existing harness sessions on later turns instead of starting a new CLI session
- send only the current user request, without docker-agent system instructions or conversation history
- cover session persistence and resume command construction across runtime recreation

## Testing

- `task lint`
- `task build`
- `go test ./pkg/runtime -run 'TestHarness|TestAfterLLMCallHook_Harness' -count=10`

`task test` was also run; all relevant packages passed, but `pkg/teamloader`'s `TestLoadExamples` failed for the DMR examples because Docker Model Runner was not available at `127.0.0.1:12434`.

### DIFF
--- a/pkg/runtime/harness.go
+++ b/pkg/runtime/harness.go
@@ -21,7 +21,7 @@ import (
 	"github.com/docker/docker-agent/pkg/tools"
 )
 
-func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Session, a *agent.Agent, baseExtra, baseLegacy []chat.Message, baseSources []session.InstructionSource, events EventSink) string {
+func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Session, a *agent.Agent, events EventSink) string {
 	ctx, span := r.startSpan(ctx, "runtime.harness", trace.WithAttributes(traceAttributesForHarness(sess, a)...))
 	defer span.End()
 
@@ -46,10 +46,9 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
 		r.executeTurnEndHooks(context.WithoutCancel(ctx), sess, a, endReason, events)
 	}()
 
-	turnStartMsgs := r.executeTurnStartHooks(ctx, sess, a, events)
-	legacyExtras := append(slices.Clone(baseLegacy), turnStartMsgs.legacyMessages()...)
-	messages := r.messagesWithDynamicContext(ctx, sess, a,
-		instructionSources(baseExtra, nil, turnStartMsgs, baseSources...), legacyExtras)
+	// Harnesses own their context; run lifecycle hooks but do not forward injected instructions.
+	r.executeTurnStartHooks(ctx, sess, a, events)
+	messages := harnessInputMessages(sess)
 	stop, msg, rewritten := r.executeBeforeLLMCallHooks(ctx, sess, a, modelID, 1, messages)
 	if stop {
 		slog.WarnContext(ctx, "before_llm_call hook signalled run termination",
@@ -63,7 +62,16 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
 	}
 	messages = r.applyBeforeLLMCallTransforms(ctx, sess, a, modelID, messages)
 
-	prompt := harnessPromptFromMessages(messages)
+	harnessSessionID := harnessSessionIDFor(sess, a)
+	prompt := strings.TrimSpace(harnessPrompt(messages))
+	if prompt == "" {
+		msg := "cannot run external harness without a user prompt"
+		events.Emit(ErrorWithCodeForSession(sess.ID, ErrorCodeModelError, msg))
+		r.notifyError(ctx, a, sess.ID, msg)
+		span.SetStatus(codes.Error, "harness prompt is empty")
+		endReason = turnEndReasonError
+		return endReason
+	}
 	var streamed strings.Builder
 	var finalResult string
 	var usage *chat.Usage
@@ -116,8 +124,11 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
 		}
 	}
 
-	err = baseharness.Run(ctx, provider, prompt, func(ev baseharness.Event) {
+	var reportedHarnessSessionID string
+	handleEvent := func(ev baseharness.Event) {
 		switch ev.Type {
+		case baseharness.EventSessionID:
+			reportedHarnessSessionID = strings.TrimSpace(ev.SessionID)
 		case baseharness.EventText:
 			if ev.Text == "" {
 				return
@@ -162,7 +173,12 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
 				cost = ev.Usage.TotalCostUSD
 			}
 		}
-	})
+	}
+	if harnessSessionID == "" {
+		err = baseharness.Run(ctx, provider, prompt, handleEvent)
+	} else {
+		err = baseharness.Resume(ctx, provider, harnessSessionID, prompt, handleEvent)
+	}
 	if err != nil {
 		if ctx.Err() != nil {
 			completeRemainingToolCalls(tools.ResultError("External harness was canceled."))
@@ -179,6 +195,9 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
 		span.SetStatus(codes.Error, "harness run error")
 		endReason = turnEndReasonError
 		return endReason
+	}
+	if reportedHarnessSessionID != "" && reportedHarnessSessionID != harnessSessionID {
+		r.rememberHarnessSessionID(ctx, sess, a, reportedHarnessSessionID)
 	}
 
 	completeRemainingToolCalls(harnessToolCompletedResult())
@@ -380,16 +399,44 @@ func (r *LocalRuntime) recordHarnessAssistantMessage(sess *session.Session, a *a
 	}
 }
 
-func harnessPromptFromMessages(messages []chat.Message) string {
-	var b strings.Builder
-	for _, msg := range messages {
-		content := harnessMessageContent(msg)
-		if strings.TrimSpace(content) == "" {
-			continue
-		}
-		fmt.Fprintf(&b, "<%s>\n%s\n</%s>\n\n", msg.Role, content, msg.Role)
+func harnessSessionAttributeKey(sess *session.Session, a *agent.Agent) string {
+	return fmt.Sprintf("docker-agent.harness.session.%s.%s.%s", sess.ID, a.Name(), codingharness.Label(a.Harness()))
+}
+
+func harnessSessionIDFor(sess *session.Session, a *agent.Agent) string {
+	return sess.AttributesSnapshot()[harnessSessionAttributeKey(sess, a)]
+}
+
+func (r *LocalRuntime) rememberHarnessSessionID(ctx context.Context, sess *session.Session, a *agent.Agent, harnessSessionID string) {
+	sess.SetAttribute(harnessSessionAttributeKey(sess, a), harnessSessionID)
+	if sess.IsSubSession() {
+		// SubSessionCompleted persists the full child after its stream drains.
+		return
 	}
-	return strings.TrimSpace(b.String())
+	if r.sessionStore == nil {
+		return
+	}
+	if err := r.sessionStore.UpdateSession(context.WithoutCancel(ctx), sess); err != nil {
+		slog.WarnContext(ctx, "Failed to persist harness session ID", "session_id", sess.ID, "agent", a.Name(), "error", err)
+	}
+}
+
+func harnessInputMessages(sess *session.Session) []chat.Message {
+	for _, item := range slices.Backward(sess.MessagesSnapshot()) {
+		if item.Message != nil && !item.Message.Implicit && item.Message.Message.Role == chat.MessageRoleUser {
+			return []chat.Message{item.Message.Message}
+		}
+	}
+	return nil
+}
+
+func harnessPrompt(messages []chat.Message) string {
+	for _, message := range slices.Backward(messages) {
+		if message.Role == chat.MessageRoleUser {
+			return harnessMessageContent(message)
+		}
+	}
+	return ""
 }
 
 func harnessMessageContent(msg chat.Message) string {
--- a/pkg/runtime/harness_test.go
+++ b/pkg/runtime/harness_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/docker/docker-agent/pkg/js"
 	"github.com/docker/docker-agent/pkg/paths"
 	"github.com/docker/docker-agent/pkg/session"
+	"github.com/docker/docker-agent/pkg/session/sqlitestore"
 	"github.com/docker/docker-agent/pkg/team"
 	"github.com/docker/docker-agent/pkg/tools"
 )
@@ -43,7 +44,9 @@ func TestMain(m *testing.M) {
 		panic(err)
 	}
 	for _, name := range []string{"codex", "claude"} {
-		script := fmt.Sprintf("#!/bin/sh\ncat %q\n", filepath.Join(dir, name+".out"))
+		script := fmt.Sprintf("#!/bin/sh\nprintf '%%s\\n' \"$@\" > %q\ncat %q\nif [ -f %q ]; then exit \"$(cat %q)\"; fi\n",
+			filepath.Join(dir, name+".args"), filepath.Join(dir, name+".out"),
+			filepath.Join(dir, name+".exit"), filepath.Join(dir, name+".exit"))
 		if err := os.WriteFile(filepath.Join(dir, name), []byte(script), 0o755); err != nil {
 			panic(err)
 		}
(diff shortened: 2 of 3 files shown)


pkg/runtime/harness.go
```diff
@@ -162,7 +164,15 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
...
 			}
 		}
-	})
+	}
+	if harnessSessionID == "" {
+		err = baseharness.Run(ctx, provider, prompt, handleEvent)
+	} else {
+		err = baseharness.Resume(ctx, provider, harnessSessionID, prompt, handleEvent)
+	}
+	if reportedHarnessSessionID != "" && reportedHarnessSessionID != harnessSessionID {
```
docker-agent: **[medium] Harness session ID persisted even when the run fails, potentially replacing a valid ID with one from a failed session**

`rememberHarnessSessionID` is called at lines 173–174 **before** the `if err != nil` check at line 176. An external harness (e.g. Codex, Claude Code) may emit an `EventSessionID` event early in its output and then exit with an error. When that happens:

1. `handleEvent` fires with `EventSessionID`, setting `reportedHarnessSessionID` to the new value.
2. `baseharness.Run`/`Resume` returns a non-nil `err`.
3. The new session ID is persisted unconditionally via `rememberHarnessSessionID`.
4. On the next turn, docker-agent tries to resume using the ID from the failed run — which may no longer exist — instead of the last known-good ID.

Move the persistence call inside the success path (after confirming `err == nil`):

```suggestion
	if reportedHarnessSessionID != "" && reportedHarnessSessionID != harnessSessionID {
		r.rememberHarnessSessionID(ctx, sess, a, reportedHarnessSessionID)
	}
	if err != nil {
```

Should become:

```go
if err != nil {
    // ... existing error handling
}
if reportedHarnessSessionID != "" && reportedHarnessSessionID != harnessSessionID {
    r.rememberHarnessSessionID(ctx, sess, a, reportedHarnessSessionID)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

- [low] pkg/runtime/harness.go:51 — Return value of executeTurnStartHooks discarded, dropping hook-injected messages (confidence: weak 52/100)

pkg/runtime/harness.go
```diff
@@ -380,16 +390,40 @@ func (r *LocalRuntime) recordHarnessAssistantMessage(sess *session.Session, a *a
...
+	return fmt.Sprintf("docker-agent.harness.session.%s.%s.%s", sess.ID, a.Name(), codingharness.Label(a.Harness()))
+}
+
+func harnessSessionIDFor(sess *session.Session, a *agent.Agent) string {
+	return sess.AttributesSnapshot()[harnessSessionAttributeKey(sess, a)]
+}
+
+func (r *LocalRuntime) rememberHarnessSessionID(ctx context.Context, sess *session.Session, a *agent.Agent, harnessSessionID string) {
+	sess.SetAttribute(harnessSessionAttributeKey(sess, a), harnessSessionID)
+	if sess.IsSubSession() || r.sessionStore == nil {
```
docker-agent: **[medium] Sub-session harness resumption silently broken when sub-session object is reconstructed**

`rememberHarnessSessionID` calls `sess.SetAttribute(...)` unconditionally (line 402) but skips `UpdateSession` for sub-sessions (line 403, `sess.IsSubSession()`). The in-memory attribute is set, but if the sub-session object is later reconstructed from the store (e.g. on a session reload), the harness session ID will not be there, and the next invocation will call `baseharness.Run` with a fresh session instead of `baseharness.Resume`.

This means sub-session harness resumption is only durable within the lifetime of the current in-memory sub-session object. Any scenario that reconstructs the sub-session (reload, restart) silently reverts to starting a new harness session.

If sub-sessions in this codebase are always ephemeral (never reloaded from the store independently), this is harmless. If they can be loaded and continued, this is a session-ID loss bug introduced by this PR.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

pkg/runtime/harness.go
```diff
@@ -380,16 +390,40 @@ func (r *LocalRuntime) recordHarnessAssistantMessage(sess *session.Session, a *a
...
+
+func harnessPrompt(messages []chat.Message) string {
+	for _, message := range slices.Backward(messages) {
+		if message.Role == chat.MessageRoleUser {
+			return harnessMessageContent(message)
 		}
-		fmt.Fprintf(&b, "<%s>\n%s\n</%s>\n\n", msg.Role, content, msg.Role)
 	}
-	return strings.TrimSpace(b.String())
+	return ""
```
docker-agent, line 439: **[medium] Empty prompt passed to external harness when session has no user messages or after hook rewrites**

`harnessPrompt` returns `""` when no `MessageRoleUser` message is found in the slice. There are two paths to this:

1. **No user messages**: `harnessInputMessages` returns `nil` when the session has no user messages. A `nil` slice passes through to `harnessPrompt(nil)`, which returns `""`.
2. **Hook rewrite**: `executeBeforeLLMCallHooks` can replace `messages` with a `rewritten` slice (line 59–61). If `rewritten` contains no user message, `harnessPrompt` returns `""` after `applyBeforeLLMCallTransforms`.

`baseharness.Run(ctx, provider, "", handleEvent)` or `baseharness.Resume(ctx, provider, id, "", handleEvent)` is then called with an empty prompt. The external CLI tool's behavior for an empty prompt is undefined and not surfaced in the diff — it may fail silently, misinterpret the request, or produce unexpected output.

Consider an explicit guard before calling `Run`/`Resume`:

```go
if prompt == "" {
    // surface a clear error rather than invoking the external harness with no input
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

pkg/runtime/harness.go
aheritier: **[medium] Harness session ID persisted even when the run fails, potentially replacing a valid ID with one from a failed session**

`rememberHarnessSessionID` is called at lines 173–174 **before** the `if err != nil` check at line 176. An external harness (e.g. Codex, Claude Code) may emit an `EventSessionID` event early in its output and then exit with an error. When that happens:

1. `handleEvent` fires with `EventSessionID`, setting `reportedHarnessSessionID` to the new value.
2. `baseharness.Run`/`Resume` returns a non-nil `err`.
3. The new session ID is persisted unconditionally via `rememberHarnessSessionID`.
4. On the next turn, docker-agent tries to resume using the ID from the failed run — which may no longer exist — instead of the last known-good ID.

Move the persistence call inside the success path (after confirming `err == nil`):

```suggestion
	if reportedHarnessSessionID != "" && reportedHarnessSessionID != harnessSessionID {
		r.rememberHarnessSessionID(ctx, sess, a, reportedHarnessSessionID)
	}
	if err != nil {
```

Should become:

```go
if err != nil {
    // ... existing error handling
}
if reportedHarnessSessionID != "" && reportedHarnessSessionID != harnessSessionID {
    r.rememberHarnessSessionID(ctx, sess, a, reportedHarnessSessionID)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/runtime/harness.go
aheritier: **[medium] Sub-session harness resumption silently broken when sub-session object is reconstructed**

`rememberHarnessSessionID` calls `sess.SetAttribute(...)` unconditionally (line 402) but skips `UpdateSession` for sub-sessions (line 403, `sess.IsSubSession()`). The in-memory attribute is set, but if the sub-session object is later reconstructed from the store (e.g. on a session reload), the harness session ID will not be there, and the next invocation will call `baseharness.Run` with a fresh session instead of `baseharness.Resume`.

This means sub-session harness resumption is only durable within the lifetime of the current in-memory sub-session object. Any scenario that reconstructs the sub-session (reload, restart) silently reverts to starting a new harness session.

If sub-sessions in this codebase are always ephemeral (never reloaded from the store independently), this is harmless. If they can be loaded and continued, this is a session-ID loss bug introduced by this PR.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

pkg/runtime/harness.go
aheritier: **[medium] Empty prompt passed to external harness when session has no user messages or after hook rewrites**

`harnessPrompt` returns `""` when no `MessageRoleUser` message is found in the slice. There are two paths to this:

1. **No user messages**: `harnessInputMessages` returns `nil` when the session has no user messages. A `nil` slice passes through to `harnessPrompt(nil)`, which returns `""`.
2. **Hook rewrite**: `executeBeforeLLMCallHooks` can replace `messages` with a `rewritten` slice (line 59–61). If `rewritten` contains no user message, `harnessPrompt` returns `""` after `applyBeforeLLMCallTransforms`.

`baseharness.Run(ctx, provider, "", handleEvent)` or `baseharness.Resume(ctx, provider, id, "", handleEvent)` is then called with an empty prompt. The external CLI tool's behavior for an empty prompt is undefined and not surfaced in the diff — it may fail silently, misinterpret the request, or produce unexpected output.

Consider an explicit guard before calling `Run`/`Resume`:

```go
if prompt == "" {
    // surface a clear error rather than invoking the external harness with no input
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

aheritier (changes requested):
## Request changes

The harness-session resumption implementation has correctness gaps that can discard a last-known-good session ID, fail to preserve sub-session state across reconstruction, or invoke an external harness with an empty prompt.

Reviewed commit: `506b602c7489cdaf8a621eecf675363b781b8f61`. CI is green on that SHA.

### Acknowledgement

I looked at this independently and reached the same conclusion as the existing automated review: the detailed blocking feedback is already captured in that review and its inline comments on `pkg/runtime/harness.go` (still unresolved). Those findings are relevant and worth addressing before merge, so I'm deliberately not restating the technical detail here.

Please address the existing inline threads — or explain why they don't apply — and I'll re-review.

pkg/runtime/harness.go
```diff
@@ -380,16 +390,40 @@ func (r *LocalRuntime) recordHarnessAssistantMessage(sess *session.Session, a *a
...
+		return
+	}
+	if err := r.sessionStore.UpdateSession(context.WithoutCancel(ctx), sess); err != nil {
+		slog.WarnContext(ctx, "Failed to persist harness session ID", "session_id", sess.ID, "agent", a.Name(), "error", err)
+	}
+}
+
+func harnessInputMessages(sess *session.Session) []chat.Message {
+	for _, item := range slices.Backward(sess.MessagesSnapshot()) {
+		if item.Message != nil && item.Message.Message.Role == chat.MessageRoleUser {
```
docker-agent: **[medium] harnessInputMessages returns implicit user messages instead of the real user request**

`harnessInputMessages` scans backward through session messages and returns the first `chat.MessageRoleUser` message it finds, without checking `item.Message.Implicit`. When a harness agent is invoked via delegation, the session is created with an implicit user message (e.g., `"Please proceed."`) added as the last user message. The function would return this implicit message as the harness prompt, instead of the real user request — sending a meaningless instruction to the external harness.

The `Implicit` flag exists precisely to distinguish auto-generated protocol messages from real user input. Skipping the check here means implicit messages can silently slip through as harness prompts.

**Suggested fix:** add an `Implicit` guard to the condition:

```suggestion
		if item.Message != nil && !item.Message.Implicit && item.Message.Message.Role == chat.MessageRoleUser {
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/runtime/harness.go
```diff
@@ -46,10 +46,8 @@ func (r *LocalRuntime) runHarnessAgent(ctx context.Context, sess *session.Sessio
 		r.executeTurnEndHooks(context.WithoutCancel(ctx), sess, a, endReason, events)
 	}()
 
-	turnStartMsgs := r.executeTurnStartHooks(ctx, sess, a, events)
-	legacyExtras := append(slices.Clone(baseLegacy), turnStartMsgs.legacyMessages()...)
-	messages := r.messagesWithDynamicContext(ctx, sess, a,
-		instructionSources(baseExtra, nil, turnStartMsgs, baseSources...), legacyExtras)
+	r.executeTurnStartHooks(ctx, sess, a, events)
```
docker-agent, line 50: **[medium] turn_start hook instruction output silently dropped for harness agents**

The old code used the return value of `executeTurnStartHooks` (`turnStartMsgs`) to populate both `legacyExtras` and the instruction sources passed to `messagesWithDynamicContext`. The new code calls the hook only for side effects and drops the `instructionObservation` it returns.

This is stated as intentional — the external harness manages its own context — but the change creates a silent behavioral gap: any `turn_start` hook that injects instructions (e.g. user identity, workspace context, safety policy reminders) will appear to succeed (the hook fires and emits events) but its injected content will never reach the harness prompt. Authors of such hooks have no indication their output is being discarded.

If this is intentional, a code comment explaining that `turn_start` instructions are deliberately not forwarded to the external harness would prevent future confusion and accidental regression.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

Two medium-severity findings in the new harness session persistence logic. The implicit-message filter is a clear code gap; the hook-output discard is likely intentional but warrants documentation.

pkg/runtime/harness.go
rumpl: Documented in 992c8c0ac. turn_start still runs for lifecycle side effects, while its injected instructions are intentionally not forwarded because the external harness owns its context.

pkg/runtime/harness.go
rumpl: Fixed in 992c8c0ac. Empty/whitespace prompts now emit a clear ErrorEvent and return without invoking the external harness. The test covers an implicit-only session, and the guard also covers hook rewrites that remove the user prompt.

pkg/runtime/harness.go
rumpl: Sub-sessions are deliberately not written with UpdateSession while running because SubSessionCompleted persists the full child after the stream drains; eagerly inserting it would conflict with AddSubSession. I documented that path and added a SQLite reconstruction test proving the captured harness ID survives SubSessionCompleted persistence and reload.

pkg/runtime/harness.go
rumpl: Fixed in 992c8c0ac. harnessInputMessages now excludes implicit user messages, and an implicit-only session fails explicitly without launching the CLI.

pkg/runtime/harness.go
rumpl: Fixed in 992c8c0ac. The reported session ID is now persisted only after Run/Resume succeeds, and the regression test verifies that an ID emitted by a failing resume does not replace the last known-good ID.

pkg/runtime/harness.go
docker-agent: You're absolutely right — I misread the persistence flow. The skip at line 413 is correct: SubSessionCompleted persists the full child (including all attributes) via AddSubSession after the stream completes, so eagerly calling UpdateSession mid-run would indeed conflict with the batch insertion.

I verified TestHarnessSubSessionIDSurvivesReconstruction proves the harness ID survives SQLite reconstruction. Thanks for the context and the test coverage!

docker-agent (commented):

- [medium] pkg/runtime/harness.go:410 — `rememberHarnessSessionID` stores harness session ID on child sub-session using a child-ID-scoped key; the same root cause as the inline finding above but observed from the write side (confidence: weak 31/100)

pkg/runtime/harness.go
```diff
@@ -380,16 +399,44 @@ func (r *LocalRuntime) recordHarnessAssistantMessage(sess *session.Session, a *a
...
 }
 
-func harnessPromptFromMessages(messages []chat.Message) string {
-	var b strings.Builder
-	for _, msg := range messages {
-		content := harnessMessageContent(msg)
-		if strings.TrimSpace(content) == "" {
-			continue
+func harnessSessionAttributeKey(sess *session.Session, a *agent.Agent) string {
+	return fmt.Sprintf("docker-agent.harness.session.%s.%s.%s", sess.ID, a.Name(), codingharness.Label(a.Harness()))
```
docker-agent, line 403: **[medium] `harnessSessionAttributeKey` embeds `sess.ID` in the key, breaking harness session resumption for sub-sessions**

`harnessSessionAttributeKey` builds the attribute key as:
```
docker-agent.harness.session.<sess.ID>.<agentName>.<label>
```

When a harness agent runs inside a sub-session, `rememberHarnessSessionID` stores the harness session ID on the child session using a key that includes the child's `sess.ID`. On the next turn, if the sub-session is re-created with a new ID (as is typical for ephemeral sub-sessions), `harnessSessionIDFor` looks up the key using the new child's `sess.ID` — the lookup misses, the harness session ID is not found, and a fresh harness session is started instead of resuming the existing one. The feature silently degrades to the pre-PR behaviour for any harness agent run via sub-session delegation.

Since session attributes are already scoped to a specific `*session.Session` object (they are not shared across sessions), embedding `sess.ID` in the key name is redundant in the flat-session case and breaks resumption in the sub-session case. A key like `docker-agent.harness.session.<agentName>.<label>` would be sufficient and correct in both cases.

```suggestion
	return fmt.Sprintf("docker-agent.harness.session.%s.%s", a.Name(), codingharness.Label(a.Harness()))
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

aheritier (approved):

pkg/runtime/harness.go
aheritier: WDYT of this @rumpl ?

docker-agent: I'm available if you need clarification on the finding. Happy to provide more detail on the sub-session attribute flow if it would help.

PR #3991, opened 2026-08-16
feat: include cost in session summaries

rumpl:
## Summary
- add session cost to `session.Summary`
- populate summary cost from in-memory and SQLite stores
- cover persisted, in-memory, and concurrent cost reads

## Testing
- `task build`
- `task lint`
- `go test -race ./pkg/session`


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are correct and well-tested. Cost is consistently read via the session mutex-protected `TokensAndCost()` in the in-memory store, and via the `cost` column (added with `DEFAULT 0`) in the SQLite store — no scan or concurrency issues. The new race test and round-trip tests cover the key invariants.

Sayt-0 (approved):
good

PR #3999, opened 2026-08-18
fix: add Desktop PAC-aware egress with SSRF protections

### LINKED ISSUE #3998: URL agent sources bypass Docker Desktop's proxy: SSRF-safe transport uses ProxyFromEnvironment, breaking PAC-only environments
### Description
When `docker-agent serve api` loads an agent from an https URL source, `urlSource.Read()` in `pkg/config/sources.go` constructs the Desktop proxy-aware client via `httpclient.NewHTTPClient(ctx)` but then (unless the source is marked unsafe or is localhost) replaces it with `httpclient.NewSSRFSafeTransport()`, whose `Proxy` is `http.ProxyFromEnvironment` (`pkg/httpclient/ssrf.go`). Identified by source inspection at v1.115.0 (`sources.go:329-344`), v1.124.0, v1.125.0, and current HEAD; not yet independently reproduced in a live rig. In environments where outbound HTTPS is only possible via a PAC-file proxy (no static `HTTPS_PROXY` env var — common in corporate networks), the fetch connects direct and fails at DNS/connect. Because Docker Desktop's managed Gordon agent is loaded from a URL source, the server starts with zero loaded agents: `source_loader.go` loads synchronously once, `getAgents()` silently skips the failed source and returns `[]`, the Desktop UI falls back to requesting `docker_gordon.yaml`, and `resolveSource()` (`pkg/server/session_manager.go`) cannot match it → HTTP 500 "agent not found". The OCI source path does not have this bug (`pkg/remote/pull.go` uses the Desktop-aware transport via `crane.WithTransport`). Additionally, the URL cache key is the sha256 of the full raw URL (`hashURL`); since the shipped URL embeds `desktopVersion` and `gordonTag`, every Desktop upgrade cold-starts the cache and the failure recurs.

### Expected Behavior
URL sources for Docker-hosted endpoints load successfully behind a PAC-only proxy, using the same Desktop proxy-aware transport as OCI pulls; failures should surface as a meaningful error, not a 500 "agent not found".

### Actual Behavior
Behind a PAC-only proxy, the initial source fetch fails (direct connect), the agent list is empty, and Desktop's Gordon UI receives HTTP 500 on `POST /api/sessions/<id>/agent/docker_gordon.yaml`. Because of the version-bearing cache key, the failure recurs after every Desktop upgrade even if a previous cache existed.

### Steps to Reproduce
**Expected reproduction, derived from source inspection — untested:**

1. On a host whose only outbound path is a PAC-file proxy (or simulate: block direct egress to the source host with a firewall rule; provide a working proxy configured only via PAC/Desktop proxy settings, with no `HTTP(S)_PROXY` env vars).
2. Configure Docker Desktop's proxy settings with the PAC (Desktop's own httpproxy will CONNECT to the source host successfully).
3. Cold cache: remove `<data-dir>/url_cache`.
4. Run `docker-agent serve api --listen <sock> --data-dir <dir> https://..../proxy/gordon-agent?...` (the argv Desktop uses) or any https URL source resolvable only via the proxy.
5. Expected: the source fetch dials direct and fails; `GET /api/agents` returns `[]`; any session POST for the agent returns 500.

### Docker Agent version
Identified by source inspection at v1.115.0; code present unchanged at v1.124.0, v1.125.0, and HEAD (`pkg/config/sources.go`, `urlSource.Read`).

### OS & terminal
Not OS-specific — network topology (PAC-only egress) is the trigger.

### Error output
```shell
Generic: the source load error is swallowed — `getAgents()` returns empty; the client sees `HTTP 500` / `agent not found` from `resolveSource()`.
```

### Additional context / Proposed remediation
- [ ] (1) For https `docker.com`/`*.docker.com` source URLs, use the Desktop proxy-aware transport with an explicit `Timeout` and `CheckRedirect` (`NewHTTPClient` sets neither); keep `NewSSRFSafeTransport` for all other non-localhost URLs. Do NOT reuse `environment.IsTrustedDockerURL` as the predicate — it admits localhost/loopback over plain http and gates JWT injection, not SSRF. Alternative worth security review: teach `NewSSRFSafeTransport` the Desktop proxy socket while keeping its dial-time allowlist (preserves anti-DNS-rebinding hardening for all hosts).
- [ ] (2) Bounded startup retry with backoff in `source_loader.go` when the initial synchronous load fails (today only the 60-min ticker retries, and only when `refreshInterval > 0`).
- [ ] (3) Surface unresolvable-agent as 404/503-class with the underlying source-load failure category instead of 500.

aheritier:
## Summary
Closes #3998.

Adds Desktop-optional, PAC-aware egress for configured HTTP clients while preserving standalone proxy behavior and SSRF protections. The implementation covers all supported consumers, including agent/source fetches, sessions, tools, MCP HTTP transports, and MCP OAuth flows. Remote MCP Streamable HTTP/SSE remains intentionally excluded; MCP OAuth flows are supported.

## Behavior
- Uses Docker Desktop's PAC-aware proxy when available, without making Desktop a runtime requirement.
- Retains SSRF protections and transport composition for every supported consumer.
- Distinguishes source-unavailable (`404`) from upstream/source failure (`502`) semantics.
- Preserves bounded retries and transport state/cooldown behavior.
- `DOCKER_AGENT_DISABLE_DESKTOP_PROXY=1` disables Desktop proxy use (kill switch).
- Standalone deployments should configure `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` as appropriate.
- docker-agent does not directly evaluate PAC files.

## Commit / phase map
1. Compose configured egress with Desktop PAC transport and SSRF guards.
2. Route HTTP clients through the Desktop-aware transport.
3. Preserve PAC behavior for generic clients.
4. Distinguish unavailable agent sources from upstream failures.
5. Harden Desktop PAC egress checks.
6. Retain cached PAC transport state and cooldown behavior.
7. Preserve wrapped Desktop transports.
8. Restore the SSRF protection documentation anchor.
9. Document Desktop PAC egress controls and standalone proxy guidance.
10. Rename and document the Desktop proxy kill switch, including MCP OAuth coverage.

## Validation
- `task build` passed.
- `task test` passed.
- `task lint` passed.
- All commits are signed; branch reviewed and ready to merge.

Manual PAC-only / Desktop-host smoke validation was not run and remains environment-limited coverage.


pkg/httpclient/desktop_transport.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (t *desktopAwareTransport) proxySafe(ctx context.Context, host string) bool {
+	if ip := net.ParseIP(host); ip != nil {
+		return IsPublicIP(ip)
+	}
+	ips, err := t.resolver(ctx, host)
+	if err != nil {
+		// A PAC proxy can resolve names unavailable to the local resolver.
+		return true
```
Sayt-0: **[blocking — SSRF fail-open]** Any resolver error routes the request to the PAC proxy. `fetch` takes URLs chosen by the model, so this is reachable through prompt injection: `fetch http://intranet.corp/` → NXDOMAIN locally (split-horizon DNS) → `proxySafe` returns `true` → the request is handed to the corporate proxy, which resolves it. No `allow_private_ips: true` is required, and a transient SERVFAIL or timeout is enough to flip the routing on an otherwise healthy host.

The comment explains *why* the proxy may resolve names the local resolver cannot — that is true, but it is also exactly the capability an attacker wants. Erring toward delegation is the wrong default for the one tool whose URLs are untrusted.

Minimal fix: fail closed here (`return false`) and gate proxy-side resolution behind an explicit host allowlist or `allow_private_ips`. If delegation must remain the default for Docker-hosted source URLs, scope it to that consumer rather than to every guarded client.

pkg/desktop/transport/transport.go
```diff
@@ -11,51 +11,96 @@ import (
...
+		return baseTransport
+	}
+	proxyTransport := transport.Clone()
+	proxyTransport.Proxy = http.ProxyURL(&url.URL{
+		Scheme: "http",
+	})
+	proxyTransport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		return socket.DialUnix(ctx, desktop.Paths().ProxySocket)
 	}
+	return newFallbackTransport(proxyTransport, transport)
```
Sayt-0, line 188: **[blocking — the dial-time SSRF check is bypassed on this branch]** `transport.Clone()` copies the guarded `DialContext` from `NewSSRFSafeTransport`, and line 75 then overwrites it. `SSRFDialControl` — documented at `pkg/httpclient/ssrf.go:62-80` as post-DNS and anti-rebinding — therefore never runs for proxied requests.

That leaves `proxySafe`'s pre-flight lookup as the only guard, while the proxy re-resolves the name independently. A hostname with TTL 0 that answers public at pre-flight time and `127.0.0.1` / RFC1918 / `169.254.169.254` at the proxy's own resolution has nothing left to stop it. This is the precise scenario the dial-time check exists for.

Note that this weakens the guarantee for *all* guarded consumers, far beyond the `docker.com` source URLs #3998 was about: `fetch`, `api`, `openapi`, `a2a`, `webhook`, `skills`, `toolinstall`, MCP OAuth, `tui/image`.

The cleanest shape is the one #3998 suggested: keep `IsPublicIP` enforcement at whatever performs the final dial, so the guarantee survives the proxy hop.

pkg/desktop/transport/transport.go
```diff
@@ -11,51 +11,96 @@ import (
...
+// NewDesktopTransport returns a Docker Desktop proxy transport with direct fallback.
+// A nil direct uses http.DefaultTransport.
+func NewDesktopTransport(direct http.RoundTripper) http.RoundTripper {
+	baseTransport := directTransport(direct)
+	transport, ok := baseTransport.(*http.Transport)
+	if !ok {
+		return baseTransport
+	}
+	proxyTransport := transport.Clone()
+	proxyTransport.Proxy = http.ProxyURL(&url.URL{
```
Sayt-0, line 182: **[major — silently overrides the operator's proxy configuration]** Hard-setting `Proxy` together with the socket `DialContext` means `http.ProxyFromEnvironment` — and with it `NO_PROXY` — is never consulted on this branch.

Reproduced locally with a fake Desktop socket plus a real env-proxy listener and `HTTP_PROXY`/`HTTPS_PROXY` pointed at it: the request leaves through the Desktop socket and the configured proxy is **never contacted**. Concretely, a host the operator deliberately listed in `NO_PROXY` (internal service, private registry) now goes to the Desktop proxy as soon as Desktop is running.

This is new for guarded clients — baseline `NewSSRFSafeTransport` honoured the environment and even maintained a proxy dial allowlist. It was already the case for `NewHTTPClient`.

Explicit configuration should normally win over auto-detection: skip the Desktop branch when the request URL matches `NO_PROXY`, or when an explicit `HTTPS_PROXY`/`ALL_PROXY` applies. At minimum the precedence needs to be stated in the docs and pinned by a test — `grep NO_PROXY` across `pkg/httpclient/*_test.go` only hits the pre-existing `ssrf_test.go`.

pkg/httpclient/desktop_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	for _, ip := range ips {
+		if !IsPublicIP(ip) {
+			return false
+		}
+	}
+	return true
+}
+
+func desktopProxyDisabled() bool {
+	return os.Getenv(disableDesktopProxyEnv) == "1"
```
Sayt-0: **[major — the emergency kill switch has a strict-equality footgun]** `== "1"` means `DOCKER_AGENT_DISABLE_DESKTOP_PROXY=true` (or `yes`, `on`, or `1 ` with a stray space) silently does nothing. Every other boolean env var in this repo is permissive:

- `pkg/hubauth/exchange.go:92-97` — anything except `""` / `0` / `false`, case-insensitive
- `pkg/selfupdate/selfupdate.go:707-713` — `isTruthy`: `1` / `true` / `yes` / `on`
- `pkg/telemetry/genai/content.go:28` — same set

And `docs/configuration/overview/index.md:164` documents `DOCKER_AGENT_AUTO_UPDATE` as accepting `1`, `true`, `yes`, `on` — three rows above the new entry. Anyone reaching for this variable is by definition mid-incident; failing silently is the worst available behaviour.

Suggestion: reuse `isTruthy`-style parsing and log a warning on a non-empty unrecognised value.

Related, and fine as a follow-up: there is no positive signal anywhere about which egress path was selected (Desktop proxy / direct / kill-switch bypass / blocked by pre-flight). The only log is `transport.go:165`, emitted *after* the proxy has already failed, so there is currently no way to prove a bypass took effect.

pkg/httpclient/desktop_transport.go
```diff
@@ -0,0 +1,141 @@
...
+
+func cloneDefaultTransport() *http.Transport {
+	if base, ok := http.DefaultTransport.(*http.Transport); ok {
+		return base.Clone()
+	}
+	return &http.Transport{Proxy: http.ProxyFromEnvironment}
+}
+
+func (t *desktopAwareTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if desktopProxyDisabled() || isLoopbackHost(req.URL.Hostname()) || (t.guarded && !t.proxySafe(req.Context(), req.URL.Hostname())) {
```
Sayt-0: **[major — a wasted DNS lookup on every request, including where Desktop does not exist]** The `t.proxySafe(...)` call in this condition resolves DNS, and it is evaluated *before* `desktoptransport.DesktopRunning()` on line 71. The extra `LookupIP` is therefore paid:

- on every request, even on hosts with no Docker Desktop installed (CI, Linux servers, standalone);
- on every redirect hop, since net/http re-enters `RoundTrip` per hop;
- twice per host for `fetch`, which issues a `robots.txt` request in addition to the fetch itself.

Measured against a fake authoritative DNS server: **at least 2 DNS queries for a single guarded RoundTrip** (one pre-flight, one at dial). Baseline performed no per-request pre-flight at all.

The fix is ordering — consult the memoized `DesktopRunning()` first, and pre-flight only when the Desktop branch is genuinely a candidate:

```go
if desktopProxyDisabled() || isLoopbackHost(req.URL.Hostname()) {
    return t.direct.RoundTrip(req)
}
desktopRunning, err := desktoptransport.DesktopRunning(req.Context())
if err != nil || !desktopRunning {
    return t.direct.RoundTrip(req)
}
if t.guarded && !t.proxySafe(req.Context(), req.URL.Hostname()) {
    return t.direct.RoundTrip(req)
}
```

Side effect worth noting: the pre-flight also discloses the destination to the local resolver, including for names that were meant to be resolved proxy-side.

pkg/server/session_manager.go
```diff
@@ -1594,7 +1600,11 @@ func (sm *SessionManager) loadTeam(ctx context.Context, agentFilename string, ru
 		return nil, err
 	}
 
-	return teamloader.Load(ctx, agentSource, runConfig, loaderdefaults.Opts()...)
+	t, err := teamloader.Load(ctx, agentSource, runConfig, loaderdefaults.Opts()...)
+	if err != nil {
+		return nil, fmt.Errorf("%w: load %q: %w", ErrAgentSourceUnavailable, agentFilename, err)
```
Sayt-0: **[major — wrong status class for local errors]** This wraps *every* error from `teamloader.Load` as `ErrAgentSourceUnavailable`, which `server.go:252` maps to 502. Same at lines 1621 and 1636.

Verified with a `config.NewBytesSource` holding deliberately malformed YAML:

```
GET  /api/agents/broken-yaml             -> 502  (baseline: 404)
POST /api/sessions/:id/agent/broken-yaml -> 502  (baseline: 500)
   "agent source unavailable: load \"broken-yaml\": ... sequence end token ']' not found"
```

No network is involved. The same applies to an unknown model, a missing provider API key, an invalid toolset, and `context.Canceled` from a client disconnect. A 502 instructs the caller to treat the failure as a transient upstream hiccup and retry it, which is both wrong and unactionable for a YAML syntax error.

The distinction being introduced here is genuinely valuable — it just needs a third bucket. Suggestion: keep `ErrAgentSourceUnavailable` for fetch/transport failures (wrap inside `urlSource.Read` / `sourceLoader`, where the I/O nature is known) and add something like `ErrAgentConfigInvalid` → 422/400 for parse and validation failures.

docs/tools/fetch/index.md
```diff
@@ -111,6 +111,14 @@ toolsets:
...
+
+For Docker Desktop proxy configuration, see [Docker Desktop proxy settings](https://docs.docker.com/desktop/settings-and-maintenance/settings/#proxies). Docker's [PAC files](https://docs.docker.com/enterprise/security/hardened-desktop/air-gapped-containers/#proxy-auto-configuration-pac-files) documentation describes the `containersProxy` setting for managed container and image-pull traffic, not Docker Agent's host-proxy adapter path.
+
+Set `DOCKER_AGENT_DISABLE_DESKTOP_PROXY=1` to bypass only Docker Desktop's PAC adapter temporarily when diagnosing egress problems. This does not disable SSRF protection; use the tool's documented `allow_private_ips: true` option only for trusted internal services.
+
 ### SSRF protection and reaching localhost
 
 By default, the fetch tool refuses connections to **non-public IP addresses** — even when DNS for an otherwise-public host resolves to one of them (so DNS rebinding is also blocked). The check happens at dial time, after DNS resolution, and rejects:
```
Sayt-0: **[major — the docs no longer match the code]** This section still states unconditionally that the check "happens at dial time, after DNS resolution" and that "DNS rebinding is also blocked". With Docker Desktop running, neither holds on the PAC branch: `NewDesktopTransport` replaces the guarded `DialContext`, so `SSRFDialControl` does not run, and `proxySafe` fails open on any resolver error.

The new section above is more honest ("A DNS lookup failure may be delegated to the proxy"), but a reader landing on the SSRF section — the one linked from every security discussion — still gets the old guarantee verbatim. `docs/tools/openapi/index.md` and `docs/tools/api/index.md` carry the same unqualified claim.

Whatever the code settles on, this section needs to state which guarantee applies on which path, and cross-link the Desktop-proxy caveat.

pkg/httpclient/desktop_transport.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (t *desktopAwareTransport) transportForDesktopState(ctx context.Context, desktopRunning bool) http.RoundTripper {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if !t.detectionObserved || t.desktopDetected != desktopRunning {
+		t.detectionObserved = true
+		t.desktopDetected = desktopRunning
+		if desktopRunning {
+			t.desktopTransport = t.newDesktopTransport(ctx, t.direct)
```
Sayt-0: **[minor — a Desktop flap rebuilds the transport and drops the cooldown]** Every *transition into* `true` calls `newDesktopTransport` again. Reproduced by driving the state machine `true → false → true`: a brand-new connection is opened, the previous `*http.Transport` pair is orphaned (its idle connections survive until `IdleConnTimeout`, 90s), and the `fallbackTransport`'s `disabledUntilUnixNano` cooldown is reset — so a proxy known to be broken five seconds earlier is retried immediately instead of waiting out its 30s cooldown.

This sits awkwardly with c6a3d87 ("retain Desktop PAC transport state"): `TestDesktopAwareTransportRetainsDesktopBranchCooldown` performs a single transition and never flaps back, so it cannot catch this.

Fix: build the Desktop transport lazily once and keep it across `false` periods rather than rebuilding on re-entry into `true`.

pkg/httpclient/desktop_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return true
+}
+
+func desktopProxyDisabled() bool {
+	return os.Getenv(disableDesktopProxyEnv) == "1"
+}
+
+func isLoopbackHost(host string) bool {
+	if host == "localhost" {
```
Sayt-0: **[minor — "loopback always stays direct" is not quite true]** Exact-string `localhost` plus IP literals misses the following, all verified: `LOCALHOST`, `Localhost`, `localhost.` (trailing dot), `foo.localhost` (RFC 6761 requires these to be loopback), `2130706433`, `0x7f000001`. `::ffff:127.0.0.1` and `::1` are correctly caught by `net.ParseIP`.

For the guarded transport this is mostly harmless, since `proxySafe` still runs. For the **unguarded** `NewAllowPrivateIPsClient`, `proxySafe` is skipped entirely (`t.guarded && ...`), so this function decides alone: those spellings go to the Desktop proxy, which typically cannot reach the host's loopback interface. And `isProxySocketError` only falls back to direct on *socket-level* errors, so a 403/502 from the proxy surfaces as a plain request failure rather than a retry.

`safeclient.go:18` and `docs/tools/fetch/index.md:116` both promise that loopback is always direct, so this is a documented promise the code does not keep. Lowercasing the host, stripping a trailing dot and treating the `.localhost` suffix as loopback would close it.

pkg/desktop/transport/transport.go
```diff
@@ -11,51 +11,96 @@ import (
...
+}
 
-	desktopRunning, err := memoizer.Memoize("desktopRunning", func() (bool, error) {
-		// Memoized once per process: detach the first caller's cancellation
-		// (so a cancelled caller can't poison the cached result) while keeping
-		// its trace context.
-		return desktop.IsDockerDesktopRunning(context.WithoutCancel(ctx)), nil
+// DesktopRunning reports the memoized Docker Desktop availability state.
+func DesktopRunning(ctx context.Context) (bool, error) {
+	detectionMu.Lock()
```
Sayt-0: **[minor — detection moved from once-per-client to once-per-request, behind a process-global lock]** `DesktopRunning` is now called from `desktopAwareTransport.RoundTrip` on every request, takes the package-global `detectionMu`, and holds it across `desktop.IsDockerDesktopRunning` (`pkg/desktop/running.go:9` — a `/ping` with a 3s timeout) on each 60s memoize expiry.

To be fair about the magnitude: the warm path is cheap — measured at 47.7µs for 50 concurrent warm-cache calls — and `pkg/memoize` already de-duplicates concurrent misses through `singleflight`, which makes `detectionMu` largely redundant. What does change is that a stall of up to 3s on all in-flight egress in the process can now recur every 60s for the process's entire life, whereas baseline paid it once at client construction. On a streaming agent runtime that lands on connection setup.

`context.WithoutCancel` also means the triggering request's cancellation cannot cut the probe short.

Suggestion: hold the cached state in an `atomic` for the hot path and refresh it out of band, so no request ever waits on a global lock held across I/O.

pkg/server/server.go
```diff
@@ -237,24 +238,22 @@ func agentsAPIEntry(name string, cfg *latest.Config) (api.Agent, bool) {
...
 
-	return echo.NewHTTPError(http.StatusNotFound)
+func agentSourceHTTPError(operation string, err error) error {
+	switch {
+	case errors.Is(err, ErrAgentNotFound):
+		return echo.NewHTTPError(http.StatusNotFound, err.Error())
+	case errors.Is(err, ErrAgentSourceUnavailable):
+		return echo.NewHTTPError(http.StatusBadGateway, fmt.Sprintf("%s: %v", operation, err))
+	default:
+		return nil
```
Sayt-0: **[minor, one-line fix — `default: return nil` can produce HTTP 200 with an empty body]** An echo handler that returns `nil` without writing anything yields 200 with an empty body; confirmed against the exact version in `go.mod` (echo v4.15.4).

`getAgentConfig:243` returns this value directly, so any error escaping the two sentinels becomes a silent success. It is unreachable today — every error return of `resolveSource` (only `ErrAgentNotFound`) and of `LoadAgentConfig` (everything wrapped as `ErrAgentSourceUnavailable`) was enumerated — so there is no live impact. But the other three call sites (`:295`, `:436`, `:527`) all guard with `if sourceErr := ...; sourceErr != nil`, and `getAgentConfig` is the only one that does not, leaving the invariant one refactor away from breaking silently.

Either return a 500 from `default`, or apply the same guarded idiom in `getAgentConfig`.

pkg/httpclient/safeclient.go
```diff
@@ -29,7 +42,7 @@ func NewSafeClient(timeout time.Duration, unsafe bool) *http.Client {
 	}
 	return &http.Client{
 		Timeout:       timeout,
-		Transport:     NewSSRFSafeTransport(),
+		Transport:     NewDesktopAwareSSRFSafeTransport(),
```
Sayt-0: **[minor — the three-way client API is a trap]** There are now `NewSafeClient(t, false)`, `NewSafeClient(t, true)` and `NewAllowPrivateIPsClient(t)`, with subtly different policies, and the doc comment above still says the `unsafe` branch "exists ONLY for tests" — which after this PR is literally true: the only remaining `true` callers are `pkg/hooks/builtins/builtins_test.go:199` and `pkg/skills/main_test.go:14`.

Meanwhile the same three lines are copy-pasted at five production sites (`a2a.go:188`, `api.go:42`, `openapi.go:126` and `:486`, `fetch.go:110`). The obvious move is to let `NewSafeClient` dispatch on the flag itself (`if unsafe { return NewAllowPrivateIPsClient(timeout) }`) and delete the five branches — or replace the boolean with an explicit policy type. As written, the next caller has three near-identical constructors to choose wrongly from.

Related, at `fetch.go:111`: that call builds a whole `*http.Client` (with a `Timeout` and a `BoundedRedirects(10)` closure) only to keep `.Transport` and discard the rest, since the real policy is set at `fetch.go:116`. Exporting a `NewAllowPrivateIPsTransport()` alongside `NewDesktopAwareSSRFSafeTransport()` would make both branches symmetric.

pkg/desktop/transport/transport.go
```diff
@@ -11,51 +11,96 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"sync/atomic"
+	"testing"
```
Sayt-0: **[minor — production code importing `testing`]** This file now imports the standard `testing` package in a non-test file, to accept a `testing.TB` in `SetDesktopRunningForTest` at line 93, alongside two mutable package-level function globals (`desktopRunning`, `desktopRunningOverride`).

The repo already has a pattern for exactly this seam that keeps `testing` out of the shipped package: `SetBrowserOpenerForTesting(opener) (restore func())` in `pkg/tools/mcp/oauthflow/oauthflow.go`, which returns a restore closure instead.

In fairness, the alarming parts check out: `golangci-lint run ./pkg/desktop/transport/...` reports 0 issues (the `depguard` rule only bans testify in production code), and importing `testing` registers no `-test.*` flags in a normal binary — verified with a plain `func main()` importing the package transitively, which shows 0 entries on `flag.CommandLine`. So this is style and architecture rather than a functional problem. The house pattern is right there, and it keeps the override out of the production surface.

Sayt-0 (changes requested):
Thanks for tracking this down — #3998 is a real, well-diagnosed bug, and the OCI-vs-URL asymmetry (`pkg/remote/pull.go` already uses the Desktop transport, `urlSource.Read` does not) is convincing evidence. The commit-per-phase split, the hermetic detection hook, keeping loopback direct, shipping a kill switch and documenting the behaviour across every affected tool page are all solid.

Requesting changes on one substantive point plus four cheap ones. Details are inline.

**Blocking — the security property, not the feature.** For guarded clients this PR replaces an *enforced* SSRF control with an *advisory, fail-open* one whenever Docker Desktop runs. `NewDesktopTransport` overwrites the guarded `DialContext`, so `SSRFDialControl` never executes on the proxy branch, and `proxySafe` then returns `true` on any resolver error. Since `fetch` takes model-chosen URLs, a prompt-injected `http://intranet.corp/` that fails local resolution is handed to the corporate PAC proxy, which resolves it — with no `allow_private_ips: true` involved.

#3998 anticipated this trade-off and asked for it to be arbitrated: remediation (1) scoped the Desktop transport to `docker.com` source URLs and flagged the broader option as *"worth security review: teach `NewSSRFSafeTransport` the Desktop proxy socket **while keeping its dial-time allowlist**"*. This PR takes the broad route — every guarded consumer (`fetch`, `api`, `openapi`, `a2a`, `webhook`, `skills`, `toolinstall`, MCP OAuth, `tui/image`, URL sources) — without preserving the enforcement point. That decision deserves to be explicit, ideally with a security reviewer.

Any of these unblocks: scope the PAC branch to trusted Docker hosts; or fail **closed** on local DNS errors with proxy-side resolution behind an explicit allowlist; or move `IsPublicIP` enforcement into whatever performs the final dial.

**Blocking-adjacent:** an explicitly configured `HTTPS_PROXY`/`NO_PROXY` is now silently ignored for guarded clients — reproduced locally, the configured proxy is never contacted. New versus baseline, undocumented and untested.

**Cheap fixes before merge**, all covered inline: the kill switch accepts only the exact string `"1"` while every other boolean env var in the repo is permissive; the `proxySafe` DNS lookup runs *before* the `DesktopRunning()` check, costing an extra lookup per request and per redirect hop even where Docker Desktop is absent; every `teamloader.Load`/`config.Load` error becomes a 502, so malformed local YAML now reports as a gateway failure; and `agentSourceHTTPError` returning `nil` in its `default` branch lets `getAgentConfig` answer 200 with an empty body.

**On validation.** The description notes that PAC-only/Desktop-host smoke validation was not run. Given that PAC behaviour is the entire point of the change and that an SSRF control is being relaxed, `task build/test/lint` alone looks insufficient. Minimum ask: one PAC-only manual run, a test pinning the `NO_PROXY` interaction, and a test for the `true → false → true` detection flap — the current cooldown test only performs a single transition.

**Suggestion.** The HTTP-status refactor and the startup retry are independent and uncontroversial, and between them they cover remediations (2) and (3) of #3998. Splitting them into their own PR would ship most of the user-visible fix now and let the transport work take the security review and PAC validation it needs; as it stands, 881 additions across 12 commits mix four concerns and none can be reverted independently.

One caveat on the evidence: findings 1 and 2 are code-proven and the proxy-precedence one was reproduced locally, but no Docker Desktop + PAC environment was available, so whether Desktop's own proxy refuses private destinations and compensates for the missing client-side guard remains unverified. If it does, the first finding downgrades considerably — which is precisely the fact worth establishing before merge.

pkg/desktop/transport/transport.go
aheritier: Intentional design decision: when Desktop is available, guarded traffic routes Desktop-first; DOCKER_AGENT_DISABLE_DESKTOP_PROXY provides the explicit opt-out for operators who need direct/environment-proxy behavior.

pkg/desktop/transport/transport.go
aheritier: Intentional, documented trust-boundary decision: the Desktop proxy is trusted to perform the final resolution/dial, while local preflight only rejects known-private results; NXDOMAIN-only preflight delegation is a documented limitation of this proxy path.

Sayt-0 (commented):
**Verdict: REQUEST CHANGES**

Thanks for the follow-up work — a lot of the previous round landed cleanly. Confirmed fixed at `b3620f83`: NXDOMAIN-only resolver delegation, `DesktopRunning` checked *before* the DNS preflight (verified: **0 resolver calls for 10 requests when Desktop is absent**), `default` -> 500 in `agentSourceHTTPError`, malformed local YAML back to 500, truthy kill-switch parsing, Desktop transport + cooldown surviving a `true->false->true` flap, and the `testing` import removed from production code. The OCI-vs-URL asymmetry (`pkg/remote/pull.go:69` vs the old `pkg/config/sources.go:343`) is a convincing diagnosis, and the PR also quietly fixes a latent misuse — after this change **no production caller passes `unsafe=true`** to `NewSafeClient`.

Five things still block. Four of them are cheap. Details inline.

1. **A data race, reproduced under `-race`** — lazy Desktop-transport creation writes `DisableCompression` on the *shared* `*http.Transport` that concurrent requests are reading. The PR's own "concurrent" test cannot catch it.
2. **`*.localhost` in `validateAgentURL`** now accepts plaintext `http://evil.localhost/agent.yaml` (base rejected it) and serves it over `http.DefaultTransport` with **no SSRF guard at all**. Unrelated to PAC egress, and untested.
3. **Six security contracts still promise the guarantee this PR scopes down** — including `agent-schema.json`, which feeds IDE autocomplete. The PR body's "Retains SSRF protections ... for every supported consumer" is contradicted by your own resolution comment and should be corrected before merge.
4. **`sanitizeURLForLog` keeps the URL path**, and `NewSafeClient(_, false)` now reaches that log site for the first time — webhook secrets (Slack/Discord/Telegram) live in the path.
5. **The design decision itself is still unvalidated.** Your own note says as much. The single fact that decides whether this is a blocker or a footnote: **does Docker Desktop's proxy refuse RFC1918 / link-local / metadata destinations, including on PAC `DIRECT`?** I had no Desktop+PAC environment, so I could not establish it. If it does, the residual risk drops sharply. If it does not, `fetch` takes model-chosen URLs and this is prompt-injectable SSRF.

Plus, not anchorable inline because the files are untouched — **five stale security comments**, all reachable through a now-PAC-routed client and all inaccurate whenever Desktop is running:

- `agent-schema.json:2354` — *"after DNS resolution, so DNS rebinding is also blocked"*. No generator: `AGENTS.md` and `.agents/skills/bump-config-version/SKILL.md` require a manual edit coordinated with `types.go`; `pkg/config/schema_test.go` only cross-validates.
- `pkg/config/latest/types.go:1603-1611` — same claim.
- `pkg/toolinstall/registry.go:167-173` — *"NewSafeClient enforces dial-time SSRF protection"*.
- `pkg/skills/cache.go:26-33` — *"refuses such targets at dial time, after DNS resolution, defeating DNS rebinding"*.
- `pkg/httpclient/ssrf.go:246` — `LocalhostOnlyRedirects` still hard-codes exact `localhost`, now inconsistent with the two widened predicates (see the `sources.go` thread).

I have a handful of minor/nit items (an `ALL_PROXY` claim repeated in 8 doc paragraphs that the code does not implement, raw `err.Error()` incl. the source URL in the new 502 bodies, no verdict cache on the per-request DNS preflight, and a latent probe-amplification path in `desktopDetectionCache`) — happy to post those separately if useful, but they should not gate this.

**Suggestion on scoping:** the HTTP-status refactor and the startup retry are independent and uncontroversial, and between them cover remediations (2) and (3) of #3998. Splitting them out would ship most of the user-visible fix now and let the transport work take the PAC validation and security sign-off it needs. As it stands, 17 commits across 31 files mix four concerns and none can be reverted independently.

For the record on what I checked: CI is fully green (20 checks), `golangci-lint run ./...` -> 0 issues, the custom cop suite -> no offenses, `go mod tidy -diff` clean, wasm build OK, and `go test -race -count=10 -shuffle=on ./pkg/desktop/transport/... ./pkg/httpclient/...` clean. None of the findings below are lint-visible. The one `pkg/tools/builtin/openapi` failure I hit was environmental (ambient `HTTP_PROXY` in my sandbox) and reproduces identically at base — not yours.

pkg/httpclient/desktop_transport.go
```diff
@@ -0,0 +1,204 @@
...
+	running, err := desktoptransport.DesktopRunning(ctx)
+	if err != nil || !running {
+		return t.direct
+	}
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if t.desktopTransport == nil {
+		t.desktopTransport = t.newDesktopTransport(ctx, t.direct)
+		if t.disableCompression {
+			disableCompression(t.desktopTransport)
```
Sayt-0, line 121: **blocker — data race: lazy Desktop-transport creation mutates a shared, in-use `*http.Transport`**

`disableCompression(t.desktopTransport)` here reaches `fallbackTransport.DisableCompression` (`pkg/desktop/transport/transport.go:231`), which writes `f.direct.DisableCompression = true`. But `f.direct` **is the same pointer** as `t.direct`: `NewDesktopTransport(direct)` -> `directTransport(direct)` returns its argument unchanged when non-nil. Meanwhile another goroutine can be inside `t.direct.RoundTrip` (line 93/98), where `net/http` reads that field with no synchronisation.

Reproduced with a real `httptest.Server`, 32 goroutines, and a Desktop `false->true` flip:

```
WARNING: DATA RACE
Write at 0x00c0001b92c9 by goroutine 8:
  desktop/transport.(*fallbackTransport).DisableCompression()  transport.go:231
  httpclient.disableCompression()                              desktop_transport.go:196
  httpclient.(*desktopAwareTransport).selectedTransportFor()   desktop_transport.go:113
Previous read at 0x00c0001b92c9 by goroutine 26:
  net/http.(*persistConn).roundTrip()   transport.go:2841   // pc.t.DisableCompression
```

**This is new.** At base, `transport.New(ctx)` picked proxy-vs-direct **once at construction** and `newTransport` disabled compression **before the client was published** — no post-publication write existed. The lazy `t.desktopTransport` cache is what introduces the window.

**`TestDesktopAwareTransportDisableCompressionConcurrentDesktopCreation` cannot catch this**: it substitutes a fake `compressionTransport` that does not alias `t.direct`, so the real field is never touched. I ran it 10x under `-race` — green.

Trigger in production: any client from `httpclient.NewHTTPClient` (which sets `disableCompression = true` unconditionally at construction) shared across goroutines and alive across a Desktop availability transition — the cached provider clients in `pkg/model/provider/*/client.go` are exactly that shape.

Cheapest fix: stop mutating `f.direct` after publication. `desktopAwareTransport.DisableCompression` already handles `t.direct` itself, so the fallback only needs to touch its own proxy transport — see the `transport.go:231` thread. Alternatively pass the compression flag into `NewDesktopTransport`, or build the Desktop transport eagerly in `newDesktopAwareTransport`.

pkg/desktop/transport/transport.go
```diff
@@ -11,77 +11,221 @@ import (
...
-		direct: direct,
-	}
+	return &fallbackTransport{proxy: proxy, direct: direct}
 }
 
-// DisableCompression disables automatic gzip compression on both transports.
-// This is needed for SSE streaming compatibility.
 func (f *fallbackTransport) DisableCompression() {
 	f.proxy.DisableCompression = true
 	f.direct.DisableCompression = true
```
Sayt-0: **blocker (write side of the race above) — `f.direct` is the caller's transport, not a clone**

`NewDesktopTransport` passes the *same* `*http.Transport` into `newFallbackTransport(proxyTransport, transport)`, so this line mutates a transport that `desktopAwareTransport` also owns and is concurrently serving requests on (see `pkg/httpclient/desktop_transport.go:113`).

`desktopAwareTransport.DisableCompression` already sets the flag on `t.direct` at construction time, before publication — so this line is redundant for that caller and harmful for it. The only caller that would still need it is a hypothetical one building a `fallbackTransport` directly.

```go
// DisableCompression disables automatic gzip on the proxy branch.
// The direct transport is owned by the caller (see NewDesktopTransport:
// it is passed in, not cloned), so mutating it here would race with
// requests already in flight on it.
func (f *fallbackTransport) DisableCompression() {
	f.proxy.DisableCompression = true
}
```

Unrelated nit, same function area: the doc comments removed from `fallbackTransport`, `newFallbackTransport`, `DisableCompression`, `proxyEnabled` and `isProxySocketError` in this diff were genuinely useful — the *"CAS (not Store) so a concurrent disableProxy() can't be stomped"* note especially, since it explains a non-obvious invariant. Any reason to drop them?

pkg/config/sources.go
```diff
@@ -477,7 +481,12 @@ func isLocalhostHTTP(rawURL string) bool {
...
 		return false
 	}
-	return u.Scheme == "http" && u.Hostname() == "localhost"
+	return u.Scheme == "http" && isLocalhostHTTPHost(u.Hostname())
+}
+
+func isLocalhostHTTPHost(host string) bool {
+	host = strings.TrimSuffix(host, ".")
+	return strings.EqualFold(host, "localhost") || strings.HasSuffix(strings.ToLower(host), ".localhost")
 }
```
Sayt-0, line 494: **blocker — this widening downgrades the HTTPS-only requirement for agent sources, and it is unrelated to PAC egress**

Base was `u.Hostname() == "localhost"` (exact). With the `.localhost` suffix match, `validateAgentURL` now accepts plaintext HTTP:

```
validateAgentURL("http://evil.localhost/agent.yaml")  -> err=<nil>     // base: "only https:// URLs are allowed"
validateAgentURL("http://sub.localhost./agent.yaml")  -> err=<nil>
```

And `urlSource.Read:340-343` then picks `&http.Client{Timeout: 60s, CheckRedirect: LocalhostOnlyRedirects(10)}` with a **nil `Transport`** => `http.DefaultTransport`: no SSRF dial guard, no PAC. On the agent configuration — the highest-trust input in the product, since it defines instructions, tools and commands.

RFC 6761 is a convention, not a resolver guarantee. Go's own resolver has **no** `.localhost` special case (grepped `src/net/{lookup,dnsclient_unix,hosts}.go`); macOS and systemd-resolved do, but musl (Alpine), Docker's embedded resolver at `127.0.0.11`, or a corporate/VPN DNS will forward the query upstream. On those, an attacker-chosen or misconfigured answer for `evil.localhost` is fetched **unguarded, in cleartext**.

`isLoopbackHost` in `pkg/httpclient/desktop_transport.go:171` has a legitimate reason to accept `*.localhost` (local-dev subdomains like `myapp.localhost:3000`). But `isLocalhostHTTPHost` governs a **different security boundary** and nothing in #3998 requires touching it. Suggest keeping the source predicate exact:

```suggestion
func isLocalhostHTTPHost(host string) bool {
	// Deliberately exact: this predicate decides whether an agent source may
	// be fetched over plaintext HTTP with no dial-time SSRF guard (see
	// urlSource.Read). *.localhost is an RFC 6761 convention, not a resolver
	// guarantee -- musl, Docker's embedded DNS and corporate resolvers forward
	// it upstream. Use httpclient.isLoopbackHost for transport-level checks.
	return strings.EqualFold(strings.TrimSuffix(host, "."), "localhost")
}
```

Two follow-ons in the same area:

- **Lines 492-497** (`validateAgentURL`'s doc comment) are now doubly stale: they still claim SSRF is enforced "at dial time by `[httpclient.NewSSRFSafeTransport]`", but line 343 now calls `NewDesktopAwareSSRFSafeTransport`, and that enforcement no longer applies on the Desktop branch.
- **Third definition of "localhost"**: `pkg/httpclient/ssrf.go:246` (`LocalhostOnlyRedirects`) still requires exact `localhost`, so a `*.localhost` source that redirects — *even to itself* — now fails. Reproduced: `refusing redirect to non-localhost URL "http://evil.localhost/next"`. Fail-closed rather than a bypass, but it means the widened path is broken anyway. Whatever you decide, please land **one** shared predicate across the three sites.

No test covers the new branch — `sources_test.go` only exercises exact `localhost`/`localhost:port`.

pkg/desktop/transport/transport.go
```diff
@@ -139,32 +277,39 @@ func (f *fallbackTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+	}
+	sanitized := *u
+	sanitized.User = nil
+	sanitized.RawQuery = ""
+	sanitized.ForceQuery = false
+	sanitized.Fragment = ""
+	sanitized.RawFragment = ""
+	// This duplicates pkg/tools/mcp's URL sanitizer because this leaf package
+	// cannot import that higher-level package; log injection also needs C0 removal.
+	return sanitizeForLog(sanitized.String())
```
Sayt-0: **blocker — the sanitizer strips userinfo/query/fragment but keeps the path, and secrets live in the path**

```
sanitizeURLForLog("https://hooks.slack.com/services/T000/B000/SUPERSECRET?x=1#f")
 -> "https://hooks.slack.com/services/T000/B000/SUPERSECRET"
```

This repo's own type comment (`pkg/config/latest/types.go:1462-1477`) names Slack, Mattermost, Discord, IFTTT and Telegram as webhook targets — all of them put the credential in the path. `pkg/tools/builtin/webhook/webhook.go:139` uses `NewSafeClient(timeout, false)`, which **because of this PR** can reach this `slog.Warn` for the first time: at base, `NewSafeClient(_, false)` used `NewSSRFSafeTransport()` directly and never went through `fallbackTransport`. Same new exposure for `skills`, `toolinstall`, `hooks`, `tui/image` and MCP OAuth. And `docs/community/troubleshooting/index.md` asks users to attach logs to issue reports.

Since this is a "the proxy socket is gone" diagnostic, scheme+host is all the log actually needs:

```go
func sanitizeURLForLog(u *url.URL) string {
	if u == nil || u.Host == "" {
		return ""
	}
	// Scheme+host only: paths carry credentials for common webhook targets
	// (Slack /services/T/B/SECRET, Telegram /bot<token>/...). The proxy-socket
	// diagnostic does not need the path.
	return sanitizeForLog(u.Scheme + "://" + u.Host)
}
```

Two things I checked so you do not have to:

- **The `error` field is fine.** It comes from `f.proxy.RoundTrip`, below the `http.Client` layer, so it is *not* `*url.Error`-wrapped and carries no URL. No need to touch it.
- **`sanitizeForLog` and Unicode:** it maps only `r < 0x20 || r == 0x7f`. `slog`'s TextHandler quotes U+2028/U+2029/U+202E, and JSONHandler escapes U+2028/U+2029 — but **U+202E (RTL override) passes through raw in JSON output**. Cosmetic log-spoofing only; optional.

Also: the comment on line ~289 says this duplicates `pkg/tools/mcp`'s sanitizer because of import direction. That is the right diagnosis of the constraint — but a small leaf package (`pkg/logredact`?) would let both sites share one implementation and stop the two copies from drifting.

pkg/httpclient/safeclient.go
```diff
@@ -27,9 +60,5 @@ func NewSafeClient(timeout time.Duration, unsafe bool) *http.Client {
 	if unsafe {
 		return &http.Client{Timeout: timeout}
 	}
-	return &http.Client{
-		Timeout:       timeout,
-		Transport:     NewSSRFSafeTransport(),
-		CheckRedirect: BoundedRedirects(10),
-	}
+	return ClientForAllowPrivateIPs(timeout, false)
```
Sayt-0, line 71: **blocker — `NewSafeClient`'s own doc comment (lines 52-54) now overstates the guarantee**

This line routes every `NewSafeClient(_, false)` caller to the shared `NewDesktopAwareSSRFSafeTransport`, yet lines 52-54 still say:

> The default refuses connections to non-public IPs at dial time — defeating DNS rebinding to loopback / RFC1918 / link-local incl. cloud metadata at 169.254.169.254 — and bounds the redirect chain at 10 hops.

True only on the direct/fallback branch. When Desktop is running and `proxySafe` passes, `NewDesktopTransport` overwrites `DialContext` with the unix-socket dial, so `SSRFDialControl` never runs; protection is a *pre-request* DNS preflight with a TOCTOU window, not a dial-time check.

You did this reformulation carefully in the Markdown docs (`docs/tools/fetch/index.md:116-124` is genuinely well written) — the Go/JSON contracts just did not get the same pass. The same claim is unchanged at `agent-schema.json:2354`, `pkg/config/latest/types.go:1603-1611`, `pkg/toolinstall/registry.go:167-173`, `pkg/skills/cache.go:26-33` and `pkg/config/sources.go:492-497`. Note `agent-schema.json` is **not** generated — `AGENTS.md` requires a manual edit coordinated with `types.go`, and `pkg/config/schema_test.go` only cross-validates the two.

Suggested wording for this one:

```go
// On its direct path -- Docker Desktop unavailable, disabled via
// DOCKER_AGENT_DISABLE_DESKTOP_PROXY, or ineligible -- the client refuses
// non-public IPs at dial time, defeating DNS rebinding to loopback /
// RFC1918 / link-local incl. cloud metadata at 169.254.169.254.
//
// When Docker Desktop is running, eligible non-loopback requests go through
// its PAC proxy instead. There the only local control is a pre-request DNS
// preflight (see desktopAwareTransport.proxySafe): Desktop selects and
// resolves the destination itself, so dial-time enforcement does not apply.
// Either way the redirect chain is bounded at 10 hops.
```

Related, and worth a line here: `TransportForAllowPrivateIPs` hands out a **process-wide mutable `*http.Transport`**. Today only `pkg/httpclient/client.go:newTransport` mutates a transport, and it builds a fresh one — but a future caller doing `TransportForAllowPrivateIPs(true).(*http.Transport).DisableCompression = true` would silently change behaviour for `fetch`, `api`, `openapi`, `a2a` and MCP OAuth at once. Worth documenting as read-only.

Credit where due: sharing these transports is a real improvement — `fetch`/`api`/`openapi` previously built a fresh transport per call and never reused connections.

pkg/httpclient/desktop_transport.go
```diff
@@ -0,0 +1,204 @@
...
+	if len(ips) == 0 {
+		return false
+	}
+	for _, ip := range ips {
+		if !IsPublicIP(ip) {
+			return false
+		}
+	}
+	return true
+}
```
Sayt-0, line 146: **blocker (design decision, not a coding defect) — NXDOMAIN delegation + TOCTOU, and this is the item that needs an explicit sign-off**

The narrowing from "any resolver error" to `dnsErr.IsNotFound` is a real improvement over the previous round. But NXDOMAIN is arguably the *most* load-bearing case: in split-horizon corporate DNS, internal names are precisely the ones that NXDOMAIN locally and resolve through the proxy. Since `fetch` takes **model-chosen** URLs, a prompt-injected `http://jenkins.internal/` or `http://intranet.corp/` is handed to the PAC proxy with **no `allow_private_ips: true` anywhere**. The reachable set goes from *zero* proxy-only internal names to *all* proxy-reachable internal names.

Separately, `proxySafe` resolves and Desktop then resolves again independently — a check-time/use-time gap. A domain that answers public here and `169.254.169.254` at connect time is exactly the rebinding that `SSRFDialControl`'s own comment (`ssrf.go:64-69`) was written to defeat.

**To be fair on severity, and I want this on the record:** base already trusted an explicitly configured `HTTP_PROXY`/`HTTPS_PROXY` to enforce destination policy — that is the documented `proxyDialAllowlist` exception at `ssrf.go:99-104`. So this is not different in *kind*. It is different in *default*: (a) active merely because Desktop is running, with no operator action; (b) extended to every guarded consumer rather than the #3998 source-fetch case; (c) PAC `DIRECT` also escapes the guard; (d) `NO_PROXY` no longer provides an exit.

#3998 anticipated exactly this and asked for it to be arbitrated — remediation (1) scoped the Desktop transport to `docker.com` source URLs and flagged the broad option as *"worth security review"*. Your resolution comment is refreshingly honest that neither the live PAC smoke test nor the security acceptance has happened. I could not close that gap either: no Desktop+PAC environment here.

**The one fact that settles this:** does Docker Desktop's host proxy refuse RFC1918 / link-local / metadata destinations, including when PAC returns `DIRECT`? If yes, residual risk drops sharply and this becomes a documentation matter. If no, it is prompt-injectable SSRF. Please get that answered by the Desktop networking owners and record it on the PR.

Any of these also unblocks without needing the answer: scope the PAC branch to trusted Docker hosts; fail **closed** on NXDOMAIN for guarded transports and require an explicit host allowlist or `allow_private_ips` for proxy-only names; or make Desktop routing opt-in for guarded consumers.

**Test gap worth closing regardless:** there is no end-to-end test of the guarded happy path. `newDesktopAwareTransport(true)` appears only in `TestDesktopAwareTransportProxySafe` (unit-tests `proxySafe`, never round-trips) and in `TestDesktopAwareTransportDesktopWinsOverNoProxyUntilKillSwitch` (fakes `newDesktopTransport`). `NewDesktopAwareSSRFSafeTransport()` — the actual production constructor — is **never called from any test**.

Sayt-0 (approved):

pkg/httpclient/desktop_transport.go
aheritier: Addressed in the follow-up implementation: compression configuration is no longer mutated through the shared direct transport after publication, eliminating the lazy-creation race. The concurrent/race coverage was updated as part of the cleanup.

pkg/desktop/transport/transport.go
aheritier: Addressed in the follow-up cleanup: the fallback compression operation only affects the proxy transport, while the caller-owned direct transport is configured before publication. The non-obvious lifecycle/CAS comments were retained or clarified.

pkg/config/sources.go
aheritier: Addressed in the follow-up implementation: source URL validation keeps the plaintext exception restricted to exact localhost, while transport-level loopback handling remains separate. Redirect validation and tests were aligned with that boundary.

pkg/desktop/transport/transport.go
aheritier: Addressed in the follow-up cleanup: proxy-failure URL diagnostics now redact the path and retain only scheme/host, preventing webhook credentials embedded in paths from entering logs.

pkg/httpclient/safeclient.go
aheritier: Addressed in the follow-up documentation/contracts pass: the direct/fallback dial-time guarantee is distinguished from the Desktop-selected path, whose final resolution/dial is delegated to Desktop. The schema and affected Go comments were aligned with that documented trust boundary.

PR #4009, opened 2026-08-20
fix(tools): prevent slow startup from blocking turns

### LINKED ISSUE #4001: Slow toolset Start() blocks conversation turns unboundedly; RAG file watcher's context gets cancelled by the startup probe
### Summary

Two related bugs make a slow `Startable` toolset (in our case, a RAG toolset doing a large incremental reindex) both (a) silently and permanently break its own background maintenance work, and (b) turn a cosmetic 30s startup warning into an unbounded hang on the very next user turn.

### Bug 1 — the turn-critical path has no start timeout, unlike the startup sidebar path

`EmitStartupInfo` bounds each toolset's `Start()` call with a 30s deadline via `startToolsetWithTimeout` (`pkg/runtime/runtime.go:1920`), racing the call against the timeout in a goroutine+select. When it fires, the abandoned `Start()` keeps running in the background — by design, per the comment on `startToolsetWithTimeout` — holding `StartableToolSet`'s single-flight lock (`pkg/tools/startable.go:88`, a plain `sync.Mutex`, not context-aware).

But `Agent.Tools()` (`pkg/agent/agent.go:402-403`), which every conversation turn calls to gather the tool list, goes through `ensureToolSetsAreStarted` (`pkg/agent/agent.go:550`) → `toolSet.Start(ctx)` with **no timeout at all**. If the toolset's original `Start()` is still running (because it exceeded the 30s sidebar deadline), this call blocks on the mutex until that original call finishes — which can be well past 30s, well past the deadline that was supposedly enforced. The 30s bound in `startToolsetWithTimeout` protects only the sidebar display; it does not protect the actual message-sending path, so a toolset that's merely slow (not stuck) still causes a real, unbounded hang on the user's next message.

Suggested fix: apply the same bounded-timeout wrapper to `ensureToolSetsAreStarted`'s `Start()` calls, and when a toolset isn't ready in time, proceed with the tools available so far (skip that toolset for this turn) instead of blocking.

### Bug 2 — a toolset's background watcher can inherit a context that gets cancelled out from under it

`rag.ToolSet.Start()` (`pkg/tools/builtin/rag/rag.go:93-99`) spawns its background file watcher using the *same* `ctx` passed into `Start`:

```go
func (t *ToolSet) Start(ctx context.Context) error {
	...
	if err := t.manager.Initialize(ctx); err != nil { ... }
	go func() {
		if err := t.manager.StartFileWatcher(ctx); err != nil { ... }
	}()
	return nil
}
```

On the common startup path, that `ctx` is the bounded `startCtx` created in `startToolsetWithTimeout` (`pkg/runtime/runtime.go:1924`):

```go
startCtx, cancel := context.WithTimeout(ctx, timeout)
defer cancel()
```

`defer cancel()` fires as soon as `startToolsetWithTimeout` returns — which is normal and expected for the one-shot `Initialize()` call, but the file-watcher goroutine captured the *same* `startCtx`, so its background reindexing dies the moment that deferred cancel fires, and never comes back (there's no restart/refresh of the watcher's context afterward). Since `StartableToolSet.Start` is single-flight and only calls the underlying `Start()` once per process, the watcher never gets a second chance at a longer-lived context.

Observed directly in a debug log from a real session (paths lightly genericized):

```
level=WARN msg="Embedding input builder failed; falling back to raw chunk content" strategy=semantic-embeddings path=~/.tampa/runtime/agent/tampa-knowledge/systems/snowflake.md chunk_index=1 error="error receiving from semantic generation stream: context canceled"
... (repeated for chunk_index 0,2,3,4,5,6)
level=ERROR msg="Failed to re-index file" path=~/.tampa/runtime/agent/tampa-knowledge/systems/snowflake.md error="failed to generate embeddings: batch 1 failed: failed to create embeddings: context canceled"
level=ERROR msg="Failed to cleanup orphaned documents" error="failed to collect files: context canceled"
```

This happened well into the session (long after the `[RAG Manager] Initialization complete` log line), triggered by a live file edit — i.e. the watcher was already dead by the time it tried to act, and every subsequent live edit to that KB silently fails to index for the rest of the process's life. The backlog only gets cleared on the next full `Initialize()` — i.e. the next relaunch/new session — which combined with Bug 1 means that next relaunch can block the first message for however long the accumulated backlog takes to reindex.

Suggested fix: give the file watcher (and any other long-lived background work started from inside `Start()`) an independent, process-scoped context — e.g. `context.WithoutCancel(ctx)`, mirroring the pattern already used elsewhere in the codebase for MCP connectors that need to detach from a short-lived caller context.

### Repro sketch

1. Configure a `rag` toolset with `type: semantic-embeddings` (or any `Startable` toolset whose `Initialize`/`Start` can take >30s on a cold or large-delta index).
2. Launch, and let the startup sidebar probe time out (`"... is taking too long to start (>30s)"`).
3. Edit one of the indexed files while the session is running. Observe the background reindex attempt fail with `context canceled` once the sidebar probe's `startCtx` deadline has passed.
4. Send a message before the original `Start()` call has returned: the turn hangs with no user-visible indication of why, well past the 30s the sidebar implied.

### Environment

- `docker-agent` commit: `8003b634c0068ee95cdf599fdbada22b3e715275`

Sayt-0:
## Summary

- keep conversation turns responsive when a toolset start is already in progress or exceeds the start budget
- make toolset shutdown context-aware while still reaping starts that finish after shutdown times out
- detach the long-lived RAG watcher from the short-lived startup context and keep `Stop` responsible for cancellation

## Issue expectations

| Expectation from #4001 | Implementation |
| --- | --- |
| Bound `Start()` on the turn-critical path | `Agent` races turn-initiated starts against a 30-second timeout |
| Do not wait behind the startup probe's abandoned `Start()` | non-blocking `TryStart` skips an in-flight lifecycle operation |
| Continue with tools that are already available | non-blocking readiness checks omit only toolsets that are not immediately ready |
| Preserve bounded shutdown | context-aware `StopIfStarted` honors shutdown deadlines and reaps a late successful start |
| Keep the RAG watcher alive after the startup probe returns | watcher context uses `context.WithoutCancel` while preserving context values |
| Stop long-lived RAG work during lifecycle cleanup | `ToolSet.Stop` remains the owner of watcher cancellation and waits for managed goroutines |

## Validation

- `go test -race -count=1 ./pkg/tools ./pkg/agent ./pkg/runtime ./pkg/tools/builtin/rag ./pkg/tools/mcp ./pkg/tools/builtin/mcpcatalog ./pkg/team ./pkg/a2a ./pkg/acp ./pkg/chatserver ./pkg/mcp ./pkg/server`
- lifecycle and timeout regression tests repeated 100 times under the race detector with `GOMAXPROCS=1` and `GOMAXPROCS=8`
- end-to-end runtime probe timeout validation repeated 50 times under the race detector
- RAG watcher cancellation validation repeated 50 times under the race detector, including a mutation check that fails when `context.WithoutCancel` is removed
- `go test ./...` with local HTTP proxy variables removed
- `golangci-lint run ./pkg/tools/... ./pkg/agent/... ./pkg/runtime/...`
- `go vet` on directly impacted packages and consumers
- `go run ./lint .`
- `go mod tidy --diff`
- `./scripts/build.sh`

The repository-wide race run also reports existing races in `pkg/evaluation` and `pkg/tui`; both reproduce on an unchanged `main` checkout and are unrelated to this change.

Closes #4001

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"math/rand"
@@ -561,47 +562,33 @@ func (a *Agent) ToolSets() []tools.ToolSet {
 	return toolSets
 }
 
-// toolsetStartTimeout bounds a toolset start initiated on the turn path
-// when the turn context carries no deadline of its own. It mirrors the
-// runtime's startup-probe timeout (defaultToolStartTimeout) so both paths
-// give up on a wedged toolset after the same grace period; the abandoned
-// attempt keeps running in the background under the toolset's single-flight
-// lock and its outcome is picked up on a later turn.
-const toolsetStartTimeout = 30 * time.Second
-
 // tryStartToolSet starts one toolset without ever letting it stall the turn
-// (#4001):
+// (#4001). It runs the toolset's bounded, non-blocking TryStartWithTimeout
+// with the shared tools.DefaultStartTimeout budget — the same grace period
+// as the runtime's startup probe — and translates the outcome:
 //
 //   - A Start already in flight (e.g. a startup probe that timed out upstream
 //     but kept going) is skipped silently — TryStart never joins an in-flight
 //     attempt — so the turn proceeds with the toolsets that are ready.
-//   - A start initiated here runs in its own goroutine raced against the
-//     context/timeout, because a wedged toolset can ignore cancellation (the
-//     MCP connector detaches the context it is handed). On timeout the attempt
-//     is abandoned to finish in the background and the toolset is skipped for
-//     this turn — silently, since the failure reporters share the toolset's
-//     single-flight lock and consulting them would block on the very start we
-//     just abandoned.
+//   - A start initiated here that outlives the budget (a wedged toolset can
+//     ignore cancellation) is abandoned to finish in the background and the
+//     toolset is skipped for this turn — silently, since the failure
+//     reporters share the toolset's single-flight lock and consulting them
+//     would block on the very start we just abandoned. Its outcome is picked
+//     up on a later turn.
 func (a *Agent) tryStartToolSet(ctx context.Context, toolSet *tools.StartableToolSet) error {
-	startCtx, cancel := context.WithTimeout(ctx, toolsetStartTimeout)
-	defer cancel()
-
-	done := make(chan error, 1) // buffered so a late send never blocks
-	go func() {
-		started, err := toolSet.TryStart(startCtx)
-		if !started && err == nil {
-			slog.DebugContext(ctx, "Toolset start already in flight; skipping for this turn", "agent", a.Name(), "toolset", tools.DescribeToolSet(toolSet))
+	started, err := toolSet.TryStartWithTimeout(ctx, tools.DefaultStartTimeout)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+			slog.DebugContext(ctx, "Toolset start still running; skipping for this turn", "agent", a.Name(), "toolset", tools.DescribeToolSet(toolSet), "cause", err)
+			return nil
 		}
-		done <- err
-	}()
-
-	select {
-	case <-startCtx.Done():
-		slog.DebugContext(ctx, "Toolset start still running; skipping for this turn", "agent", a.Name(), "toolset", tools.DescribeToolSet(toolSet), "cause", context.Cause(startCtx))
-		return nil
-	case err := <-done:
 		return err
 	}
+	if !started {
+		slog.DebugContext(ctx, "Toolset start already in flight; skipping for this turn", "agent", a.Name(), "toolset", tools.DescribeToolSet(toolSet))
+	}
+	return nil
 }
 
 // ensureToolSetsAreStarted starts every toolset, surfacing the first
@@ -690,16 +677,20 @@ func (a *Agent) DrainWarnings() []string {
 }
 
 func (a *Agent) StopToolSets(ctx context.Context) error {
+	var errs []error
 	for _, toolSet := range a.toolsets {
 		// StopIfStarted checks-and-stops atomically under the toolset's
 		// lifecycle lock, so a toolset that was never started is left
 		// untouched, an in-flight Start that settles in time is stopped
 		// rather than leaked, and a Start wedged past ctx's deadline
 		// surfaces ctx.Err() instead of blocking shutdown forever (#4001).
+		// One failed stop must not abandon the rest: later toolsets are
+		// still stopped (an expired ctx still stops responsive ones) and
+		// the failures surface together.
 		if err := toolSet.StopIfStarted(ctx); err != nil {
-			return fmt.Errorf("failed to stop toolset: %w", err)
+			errs = append(errs, fmt.Errorf("failed to stop toolset %s: %w", tools.DescribeToolSet(toolSet), err))
 		}
 	}
 
-	return nil
+	return errors.Join(errs...)
 }
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -472,7 +472,10 @@ func (a *Agent) collectTools(ctx context.Context) ([]tools.Tool, error) {
 		started bool
 	}
 	results := concurrent.MapSlice(a.toolsets, func(toolSet *tools.StartableToolSet) listResult {
-		if !toolSet.IsStarted() {
+		// TryIsStarted: a toolset whose lifecycle operation is still in
+		// flight (e.g. a wedged Start abandoned by tryStartToolSet) is
+		// skipped for this turn instead of blocking the listing (#4001).
+		if !toolSet.TryIsStarted() {
 			return listResult{}
 		}
 		ts, err := toolSet.Tools(ctx)
@@ -558,6 +561,49 @@ func (a *Agent) ToolSets() []tools.ToolSet {
 	return toolSets
 }
 
+// toolsetStartTimeout bounds a toolset start initiated on the turn path
+// when the turn context carries no deadline of its own. It mirrors the
+// runtime's startup-probe timeout (defaultToolStartTimeout) so both paths
+// give up on a wedged toolset after the same grace period; the abandoned
+// attempt keeps running in the background under the toolset's single-flight
+// lock and its outcome is picked up on a later turn.
+const toolsetStartTimeout = 30 * time.Second
+
+// tryStartToolSet starts one toolset without ever letting it stall the turn
+// (#4001):
+//
+//   - A Start already in flight (e.g. a startup probe that timed out upstream
+//     but kept going) is skipped silently — TryStart never joins an in-flight
+//     attempt — so the turn proceeds with the toolsets that are ready.
+//   - A start initiated here runs in its own goroutine raced against the
+//     context/timeout, because a wedged toolset can ignore cancellation (the
+//     MCP connector detaches the context it is handed). On timeout the attempt
+//     is abandoned to finish in the background and the toolset is skipped for
+//     this turn — silently, since the failure reporters share the toolset's
+//     single-flight lock and consulting them would block on the very start we
+//     just abandoned.
+func (a *Agent) tryStartToolSet(ctx context.Context, toolSet *tools.StartableToolSet) error {
+	startCtx, cancel := context.WithTimeout(ctx, toolsetStartTimeout)
+	defer cancel()
+
+	done := make(chan error, 1) // buffered so a late send never blocks
+	go func() {
+		started, err := toolSet.TryStart(startCtx)
+		if !started && err == nil {
+			slog.DebugContext(ctx, "Toolset start already in flight; skipping for this turn", "agent", a.Name(), "toolset", tools.DescribeToolSet(toolSet))
+		}
+		done <- err
+	}()
+
+	select {
+	case <-startCtx.Done():
+		slog.DebugContext(ctx, "Toolset start still running; skipping for this turn", "agent", a.Name(), "toolset", tools.DescribeToolSet(toolSet), "cause", context.Cause(startCtx))
+		return nil
+	case err := <-done:
+		return err
+	}
+}
+
 // ensureToolSetsAreStarted starts every toolset, surfacing the first
 // failure of each streak as a user-visible warning and silently retrying
 // on every subsequent turn. A successful Start() automatically resets the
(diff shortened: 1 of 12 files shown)


aheritier (changes requested):
Reviewed at a4efe53 with CI green (build-and-test, windows-tests, lint, license-check, CodeQL, build-image amd64/arm64).

The core design holds up. I traced the publish/consume handshake between `StopIfStarted` and `unlock` and could not construct a stale-request scenario: a request published before a holder's check is consumed by that holder's release, and one published after it always finds `mu` free (or a new holder that will consume it on release), so it cannot survive to reap an unrelated later start. The new tests are meaningful, not decorative — reverting `collectTools` to `IsStarted()` plus a legacy check-then-stop makes `pkg/agent` deadlock and fail, and removing `context.WithoutCancel` in `rag.go` fails `TestStartDetachesWatcherFromCallerContext`. Keeping `Initialize` on the caller's ctx while detaching only the watcher is the right split.

Three things I'd like addressed.

**[should-fix] A single wedged toolset makes shutdown abandon every later toolset.** `Agent.StopToolSets` (pkg/agent/agent.go:692-703) returns on the first error, and `Team.StopToolSets` (pkg/team/team.go:158-166) does the same across agents. Now that `StopIfStarted` can return `ctx.Err()`, one wedged toolset silently leaves the remaining *healthy, responsive* toolsets running (MCP servers/containers never stopped). Failing test against this branch:

```go
func TestReviewStopToolSetsAbandonsLaterToolsetsOnDeadline(t *testing.T) {
	synctest.Test(t, func(t *testing.T) {
		wedged := &hungStartToolSet{entered: make(chan struct{}), release: make(chan struct{})}
		healthy1 := &hungStartToolSet{entered: make(chan struct{}), release: make(chan struct{})}
		close(healthy1.release)
		healthy2 := &hungStartToolSet{entered: make(chan struct{}), release: make(chan struct{})}
		close(healthy2.release)
		a := New("root", "test", WithToolSets(wedged, healthy1, healthy2))
		require.NoError(t, a.toolsets[1].Start(t.Context()))
		require.NoError(t, a.toolsets[2].Start(t.Context()))

		startDone := make(chan error, 1)
		go func() { startDone <- a.toolsets[0].Start(t.Context()) }()
		<-wedged.entered

		ctx, cancel := context.WithTimeout(t.Context(), time.Second)
		defer cancel()
		require.ErrorIs(t, a.StopToolSets(ctx), context.DeadlineExceeded)
		assert.EqualValues(t, 1, healthy1.stops.Load(), "healthy toolset 1 should still be stopped")
		assert.EqualValues(t, 1, healthy2.stops.Load(), "healthy toolset 2 should still be stopped")

		close(wedged.release)
		synctest.Wait()
		<-startDone
	})
}
```

```
    zz_review_test.go:42: Error: expected: int(1) actual: int32(0)
        Messages: healthy toolset 1 should still be stopped
    zz_review_test.go:43: Error: expected: int(1) actual: int32(0)
        Messages: healthy toolset 2 should still be stopped
--- FAIL: TestReviewStopToolSetsAbandonsLaterToolsetsOnDeadline (0.00s)
```

Old code hung here instead, so this is not a regression — but bounding shutdown is only half the win if the rest of the toolsets are then skipped. `errors.Join` over the loop instead of an early `return` fixes both call sites, and it fits the PR's stated "preserve bounded shutdown" goal.

**[should-fix] The bounded-start dance and its 30s budget are now duplicated, and the "mirrors" claim is not enforceable.** `Agent.tryStartToolSet` + `toolsetStartTimeout` (pkg/agent/agent.go:564-604) reimplements `startToolsetWithTimeout` (pkg/runtime/runtime.go:1927-1944) — same `WithTimeout`, same buffered channel, same goroutine race, same rationale comment — against a second 30s constant (pkg/runtime/defaults.go:68). The comment at agent.go:566-567 says it "mirrors the runtime's startup-probe timeout (defaultToolStartTimeout)", but the runtime's value is configurable via `WithToolStartTimeout` (runtime.go:600) and `pkg/runtime/runtime_test.go:1388` already sets it to 50ms while the turn path stays pinned at 30s. Please share one helper (a `TryStartWithTimeout`-style function in `pkg/tools` would let both callers pass their own budget) and one constant, or drop the mirroring claim.

**[should-fix] The invariant that keeps shutdown working under an already-expired ctx is untested.** `lifecycleMutex.LockContext` (pkg/tools/startable.go:100-113) deliberately acquires an uncontended lock even when `ctx` is already done, and several shutdown paths depend on exactly that: `pkg/mcp/server.go:76` and `pkg/acp/agent.go:71` call `StopToolSets` with the ctx whose cancellation triggered the cleanup. Today the behaviour is correct (verified, `-race`, `-count=5`), but nothing pins it — a future simplification of `LockContext` to a single `select` would turn those shutdowns into silent no-ops with all tests still green. Worth a small test:

```go
func TestStopToolSetsWithAlreadyCanceledContext(t *testing.T) {
	ts := &hungStartToolSet{entered: make(chan struct{}), release: make(chan struct{})}
	close(ts.release)
	a := New("root", "test", WithToolSets(ts))
	require.NoError(t, a.toolsets[0].Start(t.Context()))

	ctx, cancel := context.WithCancel(t.Context())
	cancel()
	require.NoError(t, a.StopToolSets(ctx))
	assert.EqualValues(t, 1, ts.stops.Load(), "responsive toolset must still be stopped under an expired ctx")
}
```

**[optional] `TryIsStarted` can report a toolset it just reaped.** `TryStart` re-checks after the release precisely because the handshake may stop the toolset the attempt just started (startable.go:236-242), but `TryIsStarted` (startable.go:204-210) computes its result before the deferred `unlock`, so it can return `true` for a toolset that its own release handshake then stopped. `collectTools` (agent.go:478-482) proceeds to `Tools()` on it and can surface a spurious "list failed: … not started" warning. Narrow, shutdown-only window; a `TryStart`-style re-check would close it.

**[optional] Reads can now perform a blocking `Stop`.** After this change `IsStarted()`, `Tools()` and the `ShouldReport*` predicates may all run the underlying `Stop()` inside their release handshake (startable.go:414-441), which is invisible from their godocs. `mcpcatalog.go:933` builds a JSON status listing out of `ts.IsStarted()` and can now block for the duration of an MCP shutdown. Please note the side effect on `IsStarted`/`Tools` doc comments, since it is a public contract change for callers outside `pkg/tools`.

aheritier (commented):
Reviewed at `fa25080` with CI green (build-and-test, windows-tests, lint, license-check, CodeQL ×3, build-image amd64/arm64; the 4 `skipping` jobs are the merge/push publish jobs, and the combined legacy status has 0 contexts).

The previous round's items are all addressed: shutdown now aggregates instead of bailing on the first failure (`pkg/agent/agent.go:679-696`, `pkg/team/team.go:158-169`, pinned by `TestAgentStopToolSetsContinuesPastStopFailure`, `TestAgentStopToolSetsStopsLaterToolsetsPastDeadline`, `TestStopToolSetsContinuesAcrossAgents`); the bounded-start dance and its budget are now shared (`tools.TryStartWithTimeout` / `tools.DefaultStartTimeout`, with `startToolsetWithTimeout` deleted and `pkg/runtime/defaults.go:71` aliasing the shared constant); the expired-ctx fast path in `LockContext` is pinned (`TestStartableToolSet_StopIfStartedExpiredContextStopsResponsiveToolset`); and `TryIsStarted` now re-checks the release handshake (`pkg/tools/startable.go:207-222`).

I re-traced the publish/consume handshake for the one hole I could imagine — `StopIfStarted` publishes its request, `LockContext`'s fast `TryLock` fails, and the second `select` then picks `ctx.Done()` even though the lock has become free (Go picks uniformly when both cases are ready), leaving an orphan request nobody consumes. It is not reachable: a failed `TryLock` at `t1` (after the publish) proves some holder owned the lock at `t1`, and that holder's `stopRequestMu`-guarded check in `settleStopRequestAndRelease` necessarily runs after `t1`, so the request is always consumed. The `mu`-before-`stopRequestMu` release order in `settleStopRequestAndRelease` is what makes that argument hold — worth keeping that comment intact.

No blocking findings. Two things I'd still like addressed.

**[should-fix] A turn-initiated start that times out is completely silent, including on surfaces where the turn path is the only starter.** `tryStartToolSet` maps `DeadlineExceeded` to a `slog.Debug` and `nil` (`pkg/agent/agent.go:579-590`), and `TestAgentToolsBoundsTurnInitiatedStart` pins that silence. The stated rationale is that the failure reporters share the lifecycle lock — true, but `Agent.AddToolWarning` (`pkg/agent/agent.go:626`) does not touch that lock, which is exactly how the startup probe reports the identical condition (`pkg/runtime/runtime.go:1827-1832`: "…is taking too long to start (>%s) — it keeps starting in the background…"). Turn-path warnings do reach the user on every surface (`pkg/runtime/loop.go:375`, `:479`, `:1858` → `emitAgentWarnings`). On `--exec`, `serve api`, `serve mcp` and `serve a2a` there is no `EmitStartupInfo` pass, so a wedged toolset now costs 30 s of dead air and its tools silently disappear from the turn with nothing in the UI to explain it. Suggest emitting the same warning (and only in the `DeadlineExceeded` branch, so an in-flight skip stays quiet — subsequent turns skip immediately and cannot flood).

**[should-fix] The turn-critical path still contains two unbounded waits on the lifecycle lock.** `collectTools` was hardened at the readiness check (`pkg/agent/agent.go:479`), but the check does not hold the lock across the listing: `StartableToolSet.Tools` calls the underlying `Tools` first and only then takes the lock unconditionally (`pkg/tools/startable.go:386-388`), and the error branch calls `ShouldReportListFailure()` (`pkg/agent/agent.go:497` → `pkg/tools/startable.go:521-522`), which does the same. If any lifecycle operation wedges in that window — e.g. a recovery `Start` initiated by a concurrent startup/refresh pass while the MCP list round-trip is in flight — the turn blocks indefinitely, which is the #4001 failure mode again (narrower, and not a regression: pre-change `collectTools` blocked one line earlier in `IsStarted()`). Suggest `mu.LockContext(ctx)` in `Tools` (returning `ctx.Err()`) plus a `Try`-flavoured streak reporter, or, if that is out of scope, say so explicitly in the `Tools` godoc so the residual window is documented rather than implied closed.

**[optional] `TryStart` can report `(false, nil)` with no reap involved.** The post-release re-check re-enters `TryIsStarted` (`pkg/tools/startable.go:243-252`), which also returns false when an unrelated goroutine happens to hold the lock at that instant — so a start that genuinely succeeded is reported as not-started, outside the documented outcome list (`pkg/tools/startable.go:230-238`), and the startup pass then skips the toolset for that pass (`pkg/runtime/runtime.go:1877-1887`). Re-reading `s.started` under a re-acquired lock (or only re-checking when a request was actually settled) would keep the return value honest.

**[optional] The blocking-`Stop` side effect is documented unevenly.** `IsStarted` and `Tools` now carry the "may invoke — and block on — the underlying Stop" note, but `TryIsStarted` / `TryStart` are still introduced as the non-blocking variants (`pkg/tools/startable.go:201-206`, `:230-238`) while their release handshake can run `stopLocked` synchronously (`:473-506`), and the three `ShouldReport*` predicates (`:511`, `:521`, `:537`) say nothing about it.

**[optional] `pkg/agent/agent.go:567`** claims the turn budget is "the same grace period as the runtime's startup probe". That holds for the default only — the probe's budget is configurable through `WithToolStartTimeout` (tests use 50 ms) while the turn path is pinned to `tools.DefaultStartTimeout`.

aheritier (approved):

PR #4010, opened 2026-08-20
fix(runtime): avoid false on_user_input notifications

### LINKED ISSUE #4004: on_user_input hook fires on every stream teardown, not just real elicitation (false 'needs input' notifications for background agents)
## Summary

The `on_user_input` hook — surfaced to users as a "needs input" notification — fires unconditionally whenever *any* agent stream stops, including streams that completed normally with zero questions. This makes the notification unreliable for background agent tasks: a `run_background_agent` sub-task that finishes cleanly triggers the exact same "needs input" alert as one that is genuinely blocked waiting on the user.

## Root cause

`executeOnUserInputHooks` (`pkg/runtime/hooks.go:570-583`) is invoked from four call sites:

1. `toolexec/dispatcher.go:810` — a real tool-confirmation prompt (legitimate)
2. `elicitation.go:489` — a real elicitation/question (legitimate)
3. `loop_steps.go:70` — max iterations reached (arguably legitimate — agent is stuck)
4. **`loop.go:216`**, inside `finalizeEventChannel` — called **unconditionally at the end of every stream**, success or failure, with no check for whether the stream actually ended in a pending-input state

Background sub-agent execution (`run_background_agent`, via `agent_delegation.go` `runCollecting` at lines 346-396) calls `RunStream`, which tears down through the same `finalizeEventChannel` path. So a background agent that completes its task with no outstanding question still fires `on_user_input`.

Compounding this, the hook payload does not carry any field distinguishing *why* it fired — `userconfig_test.go:1181` shows real hook configs mapping `on_user_input` straight to a fixed notification string (e.g. `terminal-notifier -message "needs input"`), with no way for the consumer to tell "genuinely waiting on you" apart from "a stream just ended."

## Impact

Any client that surfaces `on_user_input` as a user-facing notification (e.g. background-agent-fanout systems built on docker-agent) gets a notification storm where most "needs input" alerts are false positives — a background task simply finishing. This defeats the purpose of the notification and trains users to ignore it.

## Suggested fix

Gate the `loop.go:216` call on the stream actually having ended in a pending-elicitation/pending-confirmation state, rather than firing on all teardown paths. Alternatively, include a reason/cause field in the hook payload (e.g. `elicitation`, `tool_confirmation`, `max_iterations`, `stream_end`) so consumers can filter to genuine input-needed events themselves.

## Environment

- Confirmed against docker-agent HEAD `8003b63`
- Reproduced via `run_background_agent` fan-out in a docker-agent-based agent system (multiple concurrent sub-agent tasks configured via `background_agents` toolset)

Sayt-0:
## Summary

- fire `on_user_input` at stream teardown only when an interactive root session is waiting for the next prompt
- skip the hook for sub-sessions, background agents, non-interactive runs, and canceled max-iteration waits
- preserve genuine notifications for tool confirmations, elicitations, and interactive `max_iterations`
- attribute notifications to the resolved session agent and include the conversation ID for elicitations
- add regression coverage for background-agent completion and input-wait behavior

Fixes #4004

## Issue expectations

| --- | --- |
| Background-agent completion must not emit a false input notification | Sub-session and non-interactive teardowns no longer fire `on_user_input` |
| Genuine input waits must still notify | Root interactive completion, tool confirmation, elicitation, and interactive `max_iterations` remain covered |
| `max_iterations` behavior must be explicit | Interactive sessions notify before waiting; non-interactive and already-canceled sessions stop without notifying |
| Consumers currently cannot distinguish teardown reasons | Gating was selected instead of extending the hook payload, avoiding unnecessary false events |
| Session-pinned agents must use the correct hook configuration | The resolved agent is propagated to the hook dispatcher |

## Validation

- `task build`
- `task lint`
- `go vet ./...`
- `task test`
- targeted runtime and tool-dispatch tests with the race detector

## Residual consideration

`toolexec.HookDispatcher.NotifyUserInput` now receives the resolved agent explicitly. This is technically an exported interface signature change, but repository-wide search found only the runtime implementation and its test stub.

### DIFF
--- a/pkg/runtime/elicitation.go
+++ b/pkg/runtime/elicitation.go
@@ -486,7 +486,11 @@ func (r *LocalRuntime) elicitationHandler(ctx context.Context, req *mcp.ElicitPa
 		}, nil
 	}
 
-	r.executeOnUserInputHooks(ctx, "", "elicitation")
+	// No *agent.Agent is threaded into MCP handler callbacks, so fall back
+	// to the current agent here. The session ID is the conversation ID the
+	// run loop seeded into ctx (empty for elicitations outside a run, e.g.
+	// startup OAuth probes).
+	r.executeOnUserInputHooks(ctx, r.CurrentAgent(), genai.ConversationIDFromContext(ctx), "elicitation")
 
 	// The registry key (and the ElicitationID surfaced to clients for
 	// ResumeElicitation routing) is always a freshly generated, internal
--- a/pkg/runtime/hooks.go
+++ b/pkg/runtime/hooks.go
@@ -570,10 +570,12 @@ func (r *LocalRuntime) executeAfterLLMCallHooks(ctx context.Context, sess *sessi
 
 // executeOnUserInputHooks fires on_user_input when the runtime is about
 // to wait for the user (tool confirmation, elicitation, max iterations,
-// stream stopped). Resolves the agent itself so callsites in code paths
-// without an agent handle (like the elicitation handler) stay short.
-func (r *LocalRuntime) executeOnUserInputHooks(ctx context.Context, sessionID, logContext string) {
-	a := r.CurrentAgent()
+// stream stopped). The agent is passed explicitly so callsites that
+// already resolved the session's agent (pinned sessions included)
+// attribute the event to that agent; paths without an agent handle
+// (like the elicitation handler) fall back to CurrentAgent at the
+// callsite. A nil agent is a no-op.
+func (r *LocalRuntime) executeOnUserInputHooks(ctx context.Context, a *agent.Agent, sessionID, logContext string) {
 	if a == nil {
 		return
 	}
--- a/pkg/runtime/loop.go
+++ b/pkg/runtime/loop.go
@@ -214,7 +214,13 @@ func (r *LocalRuntime) finalizeEventChannel(ctx context.Context, sess *session.S
 	// cleanup hooks run even when the stream was interrupted (e.g. Ctrl+C).
 	r.executeSessionEndHooks(context.WithoutCancel(ctx), sess, a)
 
-	r.executeOnUserInputHooks(ctx, sess.ID, "stream stopped")
+	// on_user_input means "the agent is now waiting for the user". Only a
+	// root interactive stream hands control back to a user when it ends;
+	// sub-session and non-interactive teardowns (background agents, MCP
+	// serve, A2A, evals) have nobody to wait for and must not fire it (#4004).
+	if !sess.IsSubSession() && !sess.NonInteractive {
+		r.executeOnUserInputHooks(ctx, a, sess.ID, "stream stopped")
+	}
 
 	r.telemetry.RecordSessionEnd(ctx)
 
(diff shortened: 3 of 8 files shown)


aheritier (approved):
Verified locally on 6d5ad03: `go build ./...`, `go test ./pkg/runtime/...`, and the new tests under `-race` all pass. CI is green on this SHA.

The core reasoning holds up. `on_user_input` means "control is being handed back to a human", and the three gates match that: `loop.go:221` skips sub-session and non-interactive teardowns, `loop_steps.go:91-105` moves the hook after the non-interactive auto-stop and behind a `ctx.Err()` guard so a cancelled run can't announce a wait it will never perform, and the elicitation callsite now uses `genai.ConversationIDFromContext(ctx)` — consistent with `elicitation.go:524`, which already derived its session id the same way, so the previously-empty `session_id` is now correct.

Test quality is the strongest part: I applied `pkg/runtime/on_user_input_test.go` unchanged to `origin/main`, and 7 of the 8 tests fail there (only `RootInteractiveStreamEnd_FiresOnce`, which pins pre-existing semantics, passes). That is real regression coverage rather than restated implementation.

The exported signature change is contained: `git grep NotifyUserInput` finds only `hookDispatcher` (`tool_dispatch.go:129`) and the test stub (`toolexec/helpers_test.go:62`), matching what the description claims.

### [should-fix] One-shot `--exec` runs still emit the false notification

Not a regression from this PR, and not a blocker — but the description says non-interactive runs are now skipped, and that is only true for A2A, MCP serve, chatserver, and evals, which set `session.NonInteractive` explicitly. Nothing under `cmd/` ever sets it (`git grep NonInteractive -- cmd` returns nothing), so a CLI run keeps `NonInteractive == false`:

- `pkg/cli/runner.go:291-298` — `len(userMessages) > 0`: runs `oneLoop` per message, then returns and the process exits
- `pkg/cli/runner.go:279-289` / `:299-312` — the `-` and non-tty-stdin branches: single `oneLoop`, then exit

Each of those streams tears down through `loop.go:221` as a root, "interactive" session, so the hook fires — the exact shape asserted by `TestOnUserInputHooks_RootInteractiveStreamEnd_FiresOnce`. Concretely, `docker agent run --exec agent.yaml "do X"` in CI still triggers the desktop notification documented at `docs/guides/tips/index.md:389`, with no user to notify. Only `pkg/cli/runner.go:314-336` (the bare interactive prompt loop) genuinely waits. Worth a follow-up: mark the session non-interactive in the one-shot branches, or have `cli.Run` signal which mode it is in.

### [optional] Third copy of the recording-builtin harness

`runtimeWithRecordedUserInput` (`on_user_input_test.go:31-58`) is now the third near-identical copy alongside `runtimeWithRecordedAgentSwitch` (`on_agent_switch_test.go:55-83`) and `runtimeWithRecordedSessionResume` (`on_session_resume_test.go:19-41`). Following the established pattern is defensible; at three copies a shared helper parameterized by event + builtin name would pay for itself.

### [optional] Docs could state the narrowed contract

`docs/configuration/hooks/index.md:53` ("When the agent is waiting for user input") is not stale — the fix makes behavior match it more closely. Since consumers deliberately cannot distinguish teardown reasons, a clause noting that sub-sessions, background agents, and non-interactive runs never fire it would save readers a source dive.

### [optional] Label

Labelled `kind/feat` while the commit is `fix(runtime):`; that will place this under "What's New" instead of "Bug Fixes" in the generated changelog.

Nice touches: every non-obvious gate carries a short WHY comment with a repository-qualified `#4004` reference rather than a bare assertion, and the cancelled-context test also asserts `MaxIterationsReached` is still emitted, pinning that the guard narrows only the input wait.

PR #4014, opened 2026-08-21
feat(tui): add a show/hide banner setting

rumpl:
## What

Adds a user setting to show/hide the ASCII-art startup banner, exposed as **Show startup banner** on the **Appearance** tab of `/settings` and persisted as `settings.show_banner` in `~/.config/cagent/config.yaml`.

Settings dialog — Appearance tab with the new "Show startup banner" toggle

## Why

The banner is the first thing an empty conversation shows. It's nice on a fresh start and pure noise for people who live in the TUI all day — until now there was no way to turn it off.

## How

- `userconfig.Settings.ShowBanner *bool` + `GetShowBanner()`, defaulting to `true` (unset means shown, and the default is never written to the config file).
- `/settings` → Appearance gains a toggle row; the value round-trips through `messages.Preferences` like the other dialog-managed settings.
- Applying the setting takes effect immediately on every open tab (`Page.SetShowBanner`) and is inherited by pages created later (new tab, `/clear`, `/fork`, session load) via `chat.WithShowBanner`.
- The lean TUI honors it too (`leantui.Config.ShowBanner`); the "Type a message…" hint line stays regardless.

## Testing

- New unit tests: config round trip + default, dialog toggle/navigation/apply, chat-page rendering with the banner disabled, live propagation to existing and future pages, startup path from a persisted `show_banner: false`, and the lean TUI welcome block.
- `go build ./...`, `go test ./pkg/tui/... ./pkg/userconfig/... ./pkg/leantui/... ./cmd/...`, `go run ./lint .`, `go vet`, `gofmt` all clean.

## Docs

- `docs/configuration/user-settings/index.md`: `show_banner` row in the settings reference, complete example, and precedence list.
- `docs/features/tui/index.md`: mention in the `/settings` Appearance section.

### DIFF
--- a/pkg/leantui/leantui.go
+++ b/pkg/leantui/leantui.go
@@ -28,6 +28,9 @@ type Config struct {
 	AppName          string
 	DisabledCommands []string
 	RenderImages     *bool
+	// ShowBanner displays the ASCII-art welcome banner. Defaults to true
+	// when nil.
+	ShowBanner *bool
 
 	// Banner overrides the ASCII-art welcome banner. When nil the built-in
 	// bannerLines ("docker agent") is used; embedders set it to brand the lean
@@ -168,6 +171,8 @@ type model struct {
 	banner           []string
 	disabledCommands map[string]bool
 	renderImages     bool
+	// hideBanner drops the ASCII-art welcome banner; the zero value keeps it.
+	hideBanner bool
 }
 
 func newModel(term *ui.Terminal, cfg Config) *model {
@@ -202,6 +207,7 @@ func newModel(term *ui.Terminal, cfg Config) *model {
 		banner:           cfg.Banner,
 		disabledCommands: disabled,
 		renderImages:     renderImages,
+		hideBanner:       cfg.ShowBanner != nil && !*cfg.ShowBanner,
 	}
 }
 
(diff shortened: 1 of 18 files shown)


aheritier (approved):

pkg/leantui/leantui.go
```diff
@@ -224,6 +230,9 @@ func (m *model) commitWelcome() {
 	if len(banner) == 0 {
 		banner = bannerLines
 	}
+	if m.hideBanner {
```
docker-agent, line 233: **[low] Top-padding blank line still emitted when banner is hidden in lean TUI**

When `hideBanner` is `true`, the new code sets `banner = nil` to suppress the ASCII art, but the surrounding closure in `commitWelcome` still runs `for range bannerTopPadding` unconditionally — emitting 1 blank line before the hint regardless of the banner setting. With the banner hidden, the transcript block produces `["", "", "  Type a message..."]` (top-padding blank + separator blank + hint) rather than just `["  Type a message..."]`.

A minimal fix is to guard the top-padding loop the same way the art loop is already guarded — only emit padding when there is art to pad:

```suggestion
	if m.hideBanner {
		banner = nil
	}
```

Alternatively, skip the padding when `banner` is nil inside the closure:
```go
if banner != nil {
    for range bannerTopPadding {
        lines = append(lines, "")
    }
}
```

The existing test `TestCommitWelcomeSkipsHiddenBanner` verifies the art text is absent and the hint survives, but does not assert that no blank lines appear — so the extra blank passes silently.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR is well-structured — the setting is correctly wired from `userconfig.Settings.ShowBanner` through the dialog, all open pages, and the lean TUI. Default semantics (nil → shown) are sound, and the `boolPreference` round-trip matches the established pattern used by `RenderImages` and others. One low-severity cosmetic issue is noted inline.

Issue #3996, opened 2026-08-17
Epic: Gemini image output with workspace deliverables and portable session copies


aheritier:
## Purpose

Reliable Gemini image workflows across the Docker models gateway, direct Gemini API, and Vertex request surface, with workspace deliverables, portable session bytes, and manifest-authorized TUI rendering. This is Gemini-focused, not universal multimodal provider parity.

## Configuration and request behavior

- Explicit `output_capabilities.image` true/false wins. Omitted, empty, or unset metadata uses exact provider-qualified models.dev output metadata; unknown or unavailable entries remain disabled. Model names are not capability evidence.
- Eligible ordinary Google calls request TEXT+IMAGE on gateway, direct API, and Vertex. Custom function tools and structured output are rejected before dispatch; Google Search, Maps, and code execution remain allowed. Utility calls bypass the guard and omit image response modalities.
- Audio/video input capability overrides and request-model filtering remain supported; this does not add universal upload or conversion support.
- The missing-image notice is phrase-based, with accepted negation/quotation/context false positives and wording false negatives. It is not NLP, has no retry, and does not predicate on output capability.

## Generated-media lifecycle

- Complete media bytes accumulate internally; this is not progressive partial-image display or remote binary streaming.
- Every generated item is written inside the owning workspace. Relative nesting, MIME correction, atomic hard-link no-overwrite writes (unsupported filesystems fail safely), and partial success are retained. Absolute, traversal, home-relative, external, and symlink-escaping requests are redirected to a sanitized workspace basename with a bounded warning. There is no external generated-media write or confirmation flow.
- The manifest records owner, path, and workspace root before an optional complete portable blob is stored, without an application-level generated-media cap. Blob failure warns and preserves the workspace file. Session deletion removes manifest/blob rows, not workspace files.
- The root-kind migration remains for compatibility; old unknown or external rows fail closed. Every resolution rechecks the current manifest, so session deletion revokes references. All reads authorize against the manifest first, prefer blobs, and fall back only when the blob interface is absent or the blob is not found, and only to the owning workspace. Other blob errors fail closed.
- A blob can survive missing WorkingDir provenance and then uses the recorded relative display path, not a verified absolute path. Ordinary legacy file contents are not integrity-verified.
- Naming precedence is marker, exact one-image prompt filename, provider filename, then `generated-N`. Private markers are removed from visible and persisted text.
[...]

PR #4016, opened 2026-08-22
feat(#3996): resolve and filter input media per request model

aheritier:
## What and why

Resolve audio/video input capability independently and support latest-only configuration overrides. Filter unsupported image/audio/video parts using the model selected for each request, preserve provider-specific image fallbacks, and prepare history again for fallback models. Stored history and supported text/document ordering remain intact. Input audio/video capability invariants are retained; this does not add universal upload or conversion support.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, `main`, rather than the aggregate stack against `main`.

## Commit inventory

Head: `309a95662d205a795722eae67a1b3c9d906f99bc`; parent SHA: `316a5d2fe3c1b3a97fe0e6a867f6715d173cbd70`.

- `d8ca5eea7173a77a34021b57d357ffba77d4ea8b` — feat(#3996): detect audio/video input modalities
- `98639c4860fe8f42193f377033a1866c6c1757e4` — feat(#3996): add audio/video capability overrides
- `309a95662d205a795722eae67a1b3c9d906f99bc` — feat(#3996): strip unsupported audio/video media parts

## Validation

```sh
go test -v -count=1 ./pkg/modelinfo ./pkg/config/latest ./pkg/runtime ./pkg/hooks/builtins -run 'TestResolveCaps_OverrideAudioVideo|TestLoadCaps_UnknownModelHasNoAudioVideo|TestStripUnsupportedModalitiesTransform|TestRunStream_FallbackUsesItsOwnCapabilities|TestRegisterInstallsAllBuiltins|TestModelConfigCapabilitiesAudioVideo'
```

Matched top-level tests: `pkg/config/latest`: **2**; `pkg/hooks/builtins`: **1**; `pkg/modelinfo`: **2**; `pkg/runtime`: **3**.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1756,7 +1756,7 @@
         },
         "capabilities": {
           "$ref": "#/definitions/CapabilitiesConfig",
-          "description": "Explicit attachment capability override for models the models.dev catalogue does not describe correctly (custom OpenAI-compatible providers, local models like Ollama, or dropped model versions). When set, the declared flags are authoritative and no models.dev lookup is performed; when omitted, capabilities are detected automatically. Without it, such models fall back to text-only and their image/PDF attachments are silently dropped."
+          "description": "Explicit attachment capability override for models the models.dev catalogue does not describe correctly (custom OpenAI-compatible providers, local models like Ollama, or dropped model versions). When set, the declared flags are authoritative and no models.dev lookup is performed; when omitted, capabilities are detected automatically. Without it, such models fall back to text-only and their image/PDF/audio/video attachments are silently dropped."
         },
         "cost": {
           "$ref": "#/definitions/CostConfig",
@@ -1776,6 +1776,14 @@
         "pdf": {
           "type": "boolean",
           "description": "Whether the model accepts PDF (application/pdf) attachments"
+        },
+        "audio": {
+          "type": "boolean",
+          "description": "Whether the model accepts audio attachments"
+        },
+        "video": {
+          "type": "boolean",
+          "description": "Whether the model accepts video attachments"
         }
       },
       "additionalProperties": false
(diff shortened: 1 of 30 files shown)


aheritier (commented):
**Review of 26adf11a — CI green (16 success / 3 skipped for this SHA), build/tests/lint/mutation checks pass locally. One blocking finding.**

**[blocking] Image input regresses for models models.dev doesn't know but the provider does.**
`pkg/runtime/loop.go:814-816` resolves caps with `ResolveCapsFromModel(m, override)`; on a catalogue miss `m == nil` → text-only → `strip_modalities.go:84` strips image parts before the provider runs. The previous transform was an explicit pass-through on a catalogue miss. Two concrete victims:
- **DMR** — `dmr/client.go:55` says models.dev has no `dmr` provider; `client.go:190-193` applies `provider_opts.supports_images` at conversion, now unreachable for images. Shipped `examples/dmr.yaml:27-36` ("declare the capabilities explicitly so attachments are forwarded") now silently drops images.
- **Anthropic** — `anthropic/attachments.go:28-31` `IsClaude` heuristic (custom/aliased Claude, Bedrock IDs) grants image+PDF on a miss; defeated upstream.

Repro: `ResolveCapsFromModel(nil,nil)` + image Document part → part stripped. Suggest either keeping pass-through when `m == nil && override == nil` (audio/video still dropped at `attachment.Decide`), or letting providers contribute caps (e.g. DMR projects `provider_opts` into `CapsOverride`) — and update `examples/dmr.yaml` + the "Models absent from the catalogue" paragraph in `docs/configuration/models/index.md` accordingly.

**[should-fix]** Stripping is debug-log only (`strip_modalities.go:85-88`); no user-visible `Warning` event, so a user attaching audio to a non-audio model gets no indication it was discarded. Fine as a follow-up if noted.

**[optional]** Caps are resolved for the primary model; a fallback model may differ. Pre-existing, noting once.

Verified: override wins over catalogue, history not mutated (copy), gateway/all providers covered via the runtime loop, harness path intentionally skips, only `pkg/config/latest` touched, schema + examples updated, tests load-bearing (3/3 mutants killed).

dgageot (approved):

PR #4017, opened 2026-08-22
fix(#3996): diagnose Gemini requests and sanitize API failures

aheritier:
## What and why

Record Gemini request shape without prompt, schema, media-payload, or credential fields, and classify invalid arguments into bounded actionable categories. Keep sanitized SDK context while preserving the original cause for classification. Shared UTF-8 helpers bound display metadata. This PR adds diagnostics, not image-output request policy.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4016, rather than the aggregate stack against `main`.

## Commit inventory

Head: `2c5fd913464ef376471cc098b8e9481367f16b6a`; parent SHA: `309a95662d205a795722eae67a1b3c9d906f99bc`.

- `a8d51970aa6c3737d0573a1c5686d7aa754db34e` — feat(#3996): capture request-shape diagnostics for image requests
- `1cc2ffa077f613d2605441a2d010373a85113411` — feat(#3996): add shared UTF-8-safe display-name sanitization helpers
- `2c5fd913464ef376471cc098b8e9481367f16b6a` — fix(#3996): classify Gemini API 400s into actionable categories

## Validation

Build, test compilation, owning-package tests and the named fixture passed at this PR head.

Exact deterministic fixture command:

```sh
go test -v -count=1 ./pkg/model/provider/gemini -run 'TestNewRequestShape|TestRequestShape_LogAttrsNeverLeaksToolSchemas|TestWrapGeminiError'
```

Matched top-level tests: `pkg/model/provider/gemini`: **14**.

Deterministic scope: the named local fixture exercises this PR boundary with disposable configuration/stores and fake or loopback providers as applicable. Every listed package ran nonzero matching top-level tests.

Deferred/live scope: Live provider, remote CI, and platform execution are not claimed by this deterministic receipt. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.

PR #4019, opened 2026-08-22
feat(#3996): resolve image output capability from models.dev

aheritier:
## What and why

Add latest-only `output_capabilities.image` configuration/schema support. Explicit true or false wins. Omitted, empty, or unset metadata consults exact provider-qualified models.dev output metadata; unknown models, unavailable providers, and lookup failures remain disabled. No capability is inferred from a model-name prefix. Numbered configuration packages remain frozen.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4017, rather than the aggregate stack against `main`.

## Commit inventory

Head: `51559ff510bf01ea053e3923c329c80ba724dca5`; parent SHA: `2c5fd913464ef376471cc098b8e9481367f16b6a`.

- `b705039ca893933097887e9bbc66ba90231e34f3` — feat(#3996): add output_capabilities.image model override
- `51559ff510bf01ea053e3923c329c80ba724dca5` — feat(#3996): resolve image output capability from models.dev

## Validation

```sh
go test -v -count=1 ./pkg/modelinfo ./pkg/config/latest -run 'TestResolveOutputImage|Test.*OutputCapabilities'
```

Matched top-level tests: `pkg/config/latest`: **7**; `pkg/modelinfo`: **1**.

Deferred/live scope: No new live catalogue/provider run was executed; deterministic resolution covers explicit and omitted metadata. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.

### DIFF
--- a/agent-schema.json
+++ b/agent-schema.json
@@ -1758,6 +1758,10 @@
           "$ref": "#/definitions/CapabilitiesConfig",
           "description": "Explicit attachment capability override for models the models.dev catalogue does not describe correctly (custom OpenAI-compatible providers, local models like Ollama, or dropped model versions). When set, the declared flags are authoritative and no models.dev lookup is performed; when omitted, capabilities are detected automatically. Without it, such models fall back to text-only and their image/PDF/audio/video attachments are silently dropped."
         },
+        "output_capabilities": {
+          "$ref": "#/definitions/OutputCapabilitiesConfig",
+          "description": "Explicit, owner-declared generative output capabilities for this model. Never inferred from the model name or any catalogue — omit it, or leave a flag unset/false, to preserve existing behavior. Consumed by capability-gated request contracts (e.g. the Gemini gateway image-output route) that must not activate for a model unless its owner has affirmatively opted it in. Cannot be combined with first_available (set output_capabilities.image on the candidate models instead)."
+        },
         "cost": {
           "$ref": "#/definitions/CostConfig",
           "description": "Explicit token pricing (USD per 1M tokens), overriding the models.dev catalogue. Used for per-turn cost computation, session cost tracking, the /model picker, and the after_llm_call hook's cost field. Makes an uncatalogued model (custom base_url provider, local or private deployment) 'priced' instead of billing $0. Prices must not be negative; an all-zero table means 'priced, free'. Cannot be combined with first_available."
@@ -1784,6 +1788,17 @@
         "video": {
           "type": "boolean",
           "description": "Whether the model accepts video attachments"
+        }
+      },
+      "additionalProperties": false
+    },
+    "OutputCapabilitiesConfig": {
+      "type": "object",
+      "description": "Explicit, owner-declared generative output capabilities for a model. There is no automatic detection — no catalogue of output-capable models exists, and matching on the model name string is deliberately avoided as unreliable — so this is nil/omitted by default. Omitted or false preserves existing behavior; only an explicit true opts a model into output-capability-gated behavior.",
+      "properties": {
+        "image": {
+          "type": "boolean",
+          "description": "Whether the model is declared able to generate image output. Never inferred; must be set explicitly by the owner."
         }
       },
       "additionalProperties": false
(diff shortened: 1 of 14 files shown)


aheritier (commented):
Reviewed at head `d1bcf0a`. CI is green for this SHA (16 successful check-runs, 3 skipped publish jobs, nothing pending; no legacy commit statuses). Locally on the PR head: `go build ./...` OK, `go test -count=1 -race ./pkg/modelinfo/... ./pkg/model/provider/base/... ./pkg/config/...` green, `golangci-lint` (v2.13.2) 0 issues. The new tests are load-bearing — three targeted mutants (dropping the override short-circuit, `EqualFold`→`==`, flipping the nil-store default) all fail the suite. Still a draft, so this is a comment rather than an approval.

The mechanics are good: conservative-by-default resolution with a 10s bound, `*bool` tri-state so an explicit `false` is authoritative and distinguishable from omitted, deep-copy/round-trip/shorthand behaviour pinned, `first_available` rejected with an actionable message, frozen `pkg/config/v*` untouched, strict schema, and `examples/gemini_image_output.yaml` exercised by `TestParseExamples`. Note `Config.ImageOutputEnabled` has no production caller at this head (it's consumed in #4020) — worth a line in the PR body so reviewers don't read it as dead code.

**Blocking — the docs in this PR say the opposite of what it does.** `docs/configuration/models/index.md:154-161` states "There is no automatic detection for output capabilities… always unknown/off unless the owner declares them" (plus `:79` "Never inferred" and `:176` "Omitting `output_capabilities`… always preserves existing behavior"), and `examples/gemini_image_output.yaml:6-8` says "never inferred from the model name or any catalogue". But `pkg/modelinfo/modelinfo.go:645-666` resolves an omitted flag from the models.dev output modalities. `d1bcf0a` updated the schema descriptions (`agent-schema.json:1746`, `:1780`, `:1786`) but not the prose or the example, though its commit message claims a documentation update.

**Blocking — this diverges from #3996's stated contract, and #4020 makes it user-visible.** The epic says image generation is "opt-in and owner-declared", that "the configuration remains the authority for image-output eligibility", and that models without the declaration "remain text-only". With catalogue inference, any model listed with an `image` output modality is opted in with no declaration — and since #4020 gates on the *resolved* value (`config.ResponseModalities = [TEXT, IMAGE]`, plus the hard rejection "…does not support tools in the same request"), an existing `google/gemini-2.5-flash-image` + tools setup would go from working to a request-time error. #4020 also reports `OutputCapabilityKnown` from the *declaration* while gating on the *inferred* value, so diagnostics would say "unknown" for requests the guard rejects. Please decide explicitly: keep inference and retract/adjust the epic's owner-declared-only limit (documenting the behaviour change), or keep declared-only and confine the catalogue lookup to diagnostics. Then make code, schema, docs, example and diagnostics agree.

Should-fix:
- `a205303` doesn't compile on its own: its tests reference `ResolveOutputImage` (`pkg/modelinfo/output_capabilities_test.go:40`) and `Config.ImageOutputEnabled` (`pkg/model/provider/base/base_test.go:93`), both added only in `d1bcf0a` — `go vet ./pkg/modelinfo/... ./pkg/model/provider/base/...` fails at that commit. Moving those test files into `d1bcf0a` restores atomic/bisectable commits.
- `pkg/config/latest/types.go:131-139` adds `func Bool(value bool) *bool`, born deprecated with a `//nolint:modernize`, and with zero callers repo-wide (the PR's own literals use `new(true)`). Looks like a rebase leftover — please drop it.
- The new code was inserted between the two lines of `capsFromModalities`' doc comment: `pkg/modelinfo/modelinfo.go:687` now reads `// booleans it grants. Unknown modality names are ignored.` with the subject line gone.

Optional: adding `OutputCapabilitiesConfig` to `definitionMap` in `TestSchemaMatchesGoTypes` (`pkg/config/schema_test.go:307`) would guard nested schema drift for future fields; and note that inference-by-default puts every undeclared model on the catalogue path, where a cold cache with models.dev unreachable can cost up to `loadCapsTimeout` per resolution because the embedded snapshot is deliberately not memoized (`pkg/modelsdev/store.go:176-184`).

dgageot (approved):

PR #4020, opened 2026-08-22
feat(#3996): guard image-output requests across Google surfaces

aheritier:
## What and why

Request TEXT+IMAGE for eligible ordinary Google calls through the Docker models gateway, direct Gemini API, and Vertex. Reject custom function tools and structured output before dispatch; allow Google Search, Maps, and code-execution built-ins. Title generation and compaction bypass the guard and omit image response modalities. Utility behavior is unchanged. Vertex coverage is deterministic guard/predicate coverage; live Vertex validation remains deferred.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4019, rather than the aggregate stack against `main`.

## Commit inventory

Head: `ac6390e97f8c0f8cb20f890ac59e555dccdcc0d2`; parent SHA: `51559ff510bf01ea053e3923c329c80ba724dca5`.

- `ac6390e97f8c0f8cb20f890ac59e555dccdcc0d2` — feat(#3996): reject incompatible image-output Google requests

## Validation

```sh
go test -v -count=1 ./pkg/model/provider/gemini -run 'TestCheckImageOutputRequestCompatibility|TestImageOutputRequestIncompatibleError|TestCreateChatCompletionStream_ImageOutputGuard|TestCreateChatCompletionStream_ImageResponseModalities|TestWantsImageResponseModalities'
```

Matched top-level tests: `pkg/model/provider/gemini`: **7**.

Deferred/live scope: No direct/gateway live reproduction was rerun; live Vertex remains deferred. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.

### DIFF
--- a/docs/providers/google/index.md
+++ b/docs/providers/google/index.md
@@ -60,6 +60,34 @@ models:
 | `gemini-2.5-flash`        | Fast inference, cost-effective  |
 | `gemini-2.5-pro`          | Strong reasoning, large context |
 
+## Generated Images
+
+Some Gemini models (e.g. `gemini-2.5-flash-image`) are designed to generate
+an image directly as part of their reply, not just describe one. Docker
+Agent's Gemini request path doesn't yet ask for that image output — that
+support is still being completed — so today a request like this gets a
+text-only reply. See
+[Generated Media](../../features/tui/index.md#generated-media) for the
+current, verified state.
+
+```yaml
+agents:
+  root:
+    model: google/gemini-2.5-flash-image
+```
+
+When the model is accessed through a Docker AI Gateway and explicitly
+declared image-output-capable with
+[`output_capabilities.image: true`](../../configuration/models/index.md#output-capabilities),
+Docker Agent has verified that request combined with custom function tools,
+a built-in tool (e.g. `google_search`), or structured output gets rejected
+by the gateway with an opaque, empty-body HTTP 400. To avoid that, Docker
+Agent rejects such a combination itself, before any request is sent, with a
+clear error naming which feature is incompatible. Plain text requests to
+that model (no tools, no structured output) are unaffected, as is every
+other route: direct Gemini API/Vertex AI calls, and gateway calls to a model
+without the declaration.
+
 ## Thinking Budget
 
 Gemini supports two approaches depending on the model version:
--- a/pkg/model/provider/base/base.go
+++ b/pkg/model/provider/base/base.go
@@ -79,6 +79,11 @@ func (c *Config) CapsOverride() *modelinfo.CapsOverride {
 	return &modelinfo.CapsOverride{Image: caps.Image, PDF: caps.PDF, Audio: caps.Audio, Video: caps.Video}
 }
 
+// ToolCallSupport resolves the model's tool-call capability from models.dev.
+func (c *Config) ToolCallSupport(ctx context.Context) modelinfo.ToolCallSupport {
+	return modelinfo.ResolveToolCallSupport(ctx, c.ModelOptions.ModelsDevStore(), c.ID())
+}
+
 // ImageOutputEnabled resolves the model's image-output capability from its
 // explicit tri-state configuration and, when unset, the models.dev catalogue.
 func (c *Config) ImageOutputEnabled(ctx context.Context) bool {
--- a/pkg/model/provider/base/base_test.go
+++ b/pkg/model/provider/base/base_test.go
@@ -59,6 +59,22 @@ func TestConfigCapsOverride(t *testing.T) {
 	})
 }
 
+func TestConfigToolCallSupport(t *testing.T) {
+	t.Parallel()
+
+	store := modelsdev.NewDatabaseStore(&modelsdev.Database{Providers: map[string]modelsdev.Provider{
+		"google": {Models: map[string]modelsdev.Model{
+			"tool-model": {ToolCall: true},
+		}},
+	}})
+	cfg := Config{
+		ModelConfig:  latest.ModelConfig{Provider: "google", Model: "tool-model"},
+		ModelOptions: options.Apply(options.WithModelsDevStore(store)),
+	}
+
+	assert.Equal(t, modelinfo.ToolCallSupported, cfg.ToolCallSupport(t.Context()))
+}
+
 func TestConfigImageOutputEnabled(t *testing.T) {
 	t.Parallel()
 
--- a/pkg/model/provider/gemini/client.go
+++ b/pkg/model/provider/gemini/client.go
@@ -729,6 +729,17 @@ func stringifyEnumValues(values []any) []string {
 	return out
 }
 
+// wantsImageResponseModalities reports whether this ordinary chat request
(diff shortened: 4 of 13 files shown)


aheritier (commented):
Reviewed this PR's own delta (three-dot vs `gemini-output-capability`) at d1eeba5. CI is green for this head (16 success / 3 skipped). Validated in a throwaway checkout: `go build ./...` clean, `go test -count=1 -race` clean on gemini/base/modelinfo/runtime/tui/modelerrors, `golangci-lint run` 0 issues. Draft, so this is a comment rather than an approval — and there are two blocking items below.

First, credit where it's due: the error-seam work here is unusually rigorous. `TestImageOutputRequestIncompatibleError_RoutesThroughExistingErrorSeam` pins that the error is not overflow-shaped (so `FormatError` passes it through and no compaction loop is triggered) and that `ClassifyModelError` won't retry it. The runtime and TUI integration tests assert real message-list state via the new `MessageTypeCount` rather than call counters, and correctly attribute spinner removal to `AddErrorMessage`'s internal path instead of leaning on the stop-time `RemoveSpinner`. The zero-provider-call assertions are genuine. Neutering the guard kills tests, as it should.

**[blocking] The guard fires on catalogue-inferred image output, contradicting both this PR's docs and its own diagnostics — and undeclared catalogue image models with tools go from working to rejected.**

- `pkg/model/provider/gemini/client.go:754` resolves `imageOutputEnabled := c.ImageOutputEnabled(ctx)`, which falls through to the models.dev catalogue when no override is set (`pkg/model/provider/base/base.go:88-95` → `modelinfo.ResolveOutputImage`).
- `pkg/model/provider/gemini/image_output_guard.go:70-72` gates on that resolved value, while `pkg/model/provider/gemini/diagnostics.go:85` sets `OutputCapabilityKnown` from the *declaration* only. The guard acts on one notion of "known" and reports another.
- `docs/providers/google/index.md` (new "Generated Images" section) says the rejection applies only when a model is "explicitly declared image-output-capable with `output_capabilities.image: true`", and that "every other route" is unaffected including "gateway calls to a model without the declaration".
- Probe with production wiring (teamloader always passes `options.WithModelsDevStore`), undeclared `google/gemini-2.5-flash-image` + one custom function tool: rejected before dispatch, message `"this model does not support tool calls; use a tool-capable model or remove tools from the request"`, diagnostics `OutputCapabilityKnown=false OutputCapabilityEnabled=true`. The bundled `pkg/modelsdev/snapshot.json` (2026-08-31) lists that model with output modalities `[text image]`, so this is the default offline behaviour, not an edge case. Note the message never mentions image output at all — the user is told to swap models for a tool-capability reason that isn't the real cause.
- The same inferred value also sets `ResponseModalities=[TEXT,IMAGE]` for that undeclared model (`client.go:756-758`), which the first new docs paragraph explicitly denies ("Docker Agent's Gemini request path doesn't yet ask for that image output").
- Tests can't catch this: the `gateway undeclared: never rejects even with tools` case in `image_output_guard_test.go` computes the `imageOutputEnabled` argument by hand from the declaration, so it asserts nothing about the real gate; and the end-to-end `declaration missing: reaches provider` case uses `gemini-2.5-flash` with no models.dev store wired. Confirmed by mutation: re-gating the guard on the declaration (i.e. the documented behaviour) leaves every end-to-end test green.

The root cause is the inference in `ResolveOutputImage` (raised on #4019), but it only becomes user-visible breakage here, and this PR independently ships documentation asserting the opposite — so it can't land at this head as-is. Minimal fix here: key both the guard and the `ResponseModalities` decision on the declaration, i.e. the same predicate `OutputCapabilityKnown` reports. If #4019 moves to declaration-only, this gate becomes correct with no further change.

**[blocking] Docs scope the guard to the gateway; the code rejects on direct Gemini API and Vertex AI too, and that part is explicitly unverified.**

`image_output_guard.go` contains no `apiSurface` check — unlike `wantsImageResponseModalities` (`client.go:734-741`), which does switch on surface. The tests pin the broader behaviour (`direct Gemini API declared true + tools: rejected`, `Vertex AI declared true + tools: rejected`, plus the e2e `direct Gemini call with tools ... rejected before dispatch`), yet `docs/providers/google/index.md` tells the reader "every other route" is unaffected, naming "direct Gemini API/Vertex AI calls" specifically. The commit message is honest that only gateway probing produced the empty-body 400 and that direct/Vertex rejection is precautionary "until live ... verification proves it is accepted there" — but epic #3996 states "Direct Gemini API was manually validated" for the image-output workflow, so pre-emptively rejecting there deserves either a narrower guard or docs that say so plainly. The PR title ("gateway requests") matches the docs; the commit message ("Google requests") matches the code.

**[should-fix] The guard keys on `imageOutputEnabled` instead of `wantsImageResponseModalities`, so title-generation and compaction requests are rejected even though they deliberately don't ask for image output.**

`client.go:740` excludes `GeneratingTitle()`/`Compacting()` from `ResponseModalities`, but `client.go:787` hands the guard the raw `imageOutputEnabled`. `provider_opts` survive `CloneWithOptions` (`pkg/model/provider/clone.go:16` copies the base `ModelConfig`; `builtInTools` reads `ModelConfig.ProviderOpts` at `client.go:543`), so with e.g. `google_search: true` a declared image model gets `wantsImageResponseModalities=false` for title generation and compaction yet is still rejected with "...does not support built-in tools in the same request" — a claim about a request that never asked for image output. Impact is bounded because ordinary chat is already rejected in that configuration, and title generation is handled downstream in #4021; compaction (`pkg/runtime/compactor/compactor.go:203-206`) isn't touched there, so it stays affected across the stack. Gating the guard on `wantsImageResponseModalities` would make it mean what it says.

**[should-fix] `ResolveToolCallSupport` can't distinguish "catalogue says false" from "catalogue omits the field", contrary to its own doc comment — and the omitted case is the flagship model.**

The comment promises "Missing catalogue data remains unknown rather than being treated as a negative capability claim", but only a nil store, a lookup error, or a wholly missing model yields `ToolCallSupportUnknown`. `modelsdev.Model.ToolCall` is a plain `bool` with `json:"tool_call,omitempty"` (`pkg/modelsdev/types.go:29`), so an absent field is indistinguishable from an explicit false. The bundled snapshot has no `tool_call` for `gemini-2.5-flash-image`, so that model resolves to `ToolCallUnsupported` and the user is told "this model does not support tool calls" on the strength of absent data. The tests cover explicit true/false and a missing model, but not the absent-field case. (Live models.dev currently sends `tool_call: false` explicitly, so this is honesty/robustness rather than a live-data mismatch.)

**[optional]** `config.ResponseJsonSchema != nil` in the structured-output check is redundant — `buildConfig` (`client.go:478-481`) always sets `ResponseMIMEType` alongside it, and a mutation dropping that clause survives the suite. Harmless; flagging only so it isn't mistaken for covered behaviour.

dgageot (approved):

PR #4021, opened 2026-08-22
fix(#3996): keep image-output models eligible for session titles

aheritier:
## What and why

Build ordered session-title candidates from `title_model` and the normal agent model. Keep image-output-capable candidates eligible, skip nil candidates, and fall through after an empty or failed candidate. This changes title selection only; it does not guarantee a provider returns a nonempty title.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4020, rather than the aggregate stack against `main`.

## Commit inventory

Head: `13afae4a093aac86a4af104510eac01a9be1c70d`; parent SHA: `ac6390e97f8c0f8cb20f890ac59e555dccdcc0d2`.

- `13afae4a093aac86a4af104510eac01a9be1c70d` — fix(#3996): keep image-output models eligible for session titles

## Validation

```sh
go test -v -count=1 ./pkg/sessiontitle ./pkg/runtime -run 'TestNew_ImageOutputCandidateFallsThroughOnFailure|TestNew_AllImageOutputCandidatesRemainEligible|TestLocalRuntime_TitleGenerator|TestRunStream_ImageOutputOnlyModel_MainStreamUnaffected'
```

Matched top-level tests: `pkg/runtime`: **3**; `pkg/sessiontitle`: **2**.


aheritier (commented):
Review of d5276dbe (delta vs gemini-image-request-guard only). CI green for this SHA (16/16 non-skipped checks). Built, `go test -race` (sessiontitle/runtime/server/cmd/root) and golangci-lint clean locally.

**[blocking] Docs contradict the code.** `docs/configuration/models/index.md:182-187` (new) says an `image: true` model "is skipped as a session-title candidate" and titles fall to "the first non-image-output candidate". `sessiontitle.New` (`generator.go:51-61`) has no such filter and the new tests (`TestNew_AllImageOutputCandidatesRemainEligible` etc.) assert the opposite. Same stale wording in `pkg/runtime/title_generator_test.go:130` and the PR body ("preserving the title-generation model filter"). Please drop/rewrite.

**[blocking] Title requests on image-output Gemini models with built-in tools still fail.** The only functional change is `New` returning nil on an empty list; `pkg/model` is byte-identical to base. Driving a real Gemini client through `sessiontitle.Generate` at this head: with `output_capabilities.image: true` + `provider_opts.google_search: true`, the title attempt is rejected by the #4020 guard ("does not support built-in tools") with zero wire requests — `builtInTools()` (`gemini/client.go:531-549`) ignores `GeneratingTitle()`, unlike `wantsImageResponseModalities` (`client.go:734-741`). The plain image-output case works, but that's base behaviour. Suggest making `builtInTools()` return nil when `GeneratingTitle() || Compacting()` (ideally in #4020 — it fixes compaction's identical gap too) and reframing this PR as the constructor refactor + regression pins.

**[should-fix] Pin the load-bearing property.** Tests only prove `New` doesn't filter. The comment "Gemini title requests deliberately omit response modalities" isn't guarded anywhere: removing `!GeneratingTitle() && !Compacting()` from `wantsImageResponseModalities` passes every test in gemini/sessiontitle/runtime/server. Add a gemini test: image-output client + `WithGeneratingTitle()` → no `responseModalities`, no built-in tools, exactly one provider call.

Verified OK: candidate order title_model → model → fallbacks preserved; nil generator handled by all consumers (app.go, session_manager.go, run.go, new.go, debug.go); no accidental exclusion of text-only/tools-disabled models; no interaction with the NoThinking clone sentinel or the content-free done chunk.

dgageot (approved):

PR #4022, opened 2026-08-22
fix(#3996): filter gateway SSE keepalives and test image requests

aheritier:
## What and why

Add a gateway-only, payload-free SSE keepalive filter before the pinned GenAI parser. Direct Gemini and Vertex do not opt in. Preserve meaningful named payloads and unrelated frames. Deterministic request fixtures cover gateway and direct TEXT+IMAGE construction and resolved capability; Vertex remains predicate-only. The upstream direct keepalive fixture follows the binding transport policy.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4021, rather than the aggregate stack against `main`.

## Commit inventory

Head: `099eed7f742a6ff599705e36fdaa76316ac4b296`; parent SHA: `13afae4a093aac86a4af104510eac01a9be1c70d`.

- `5686038171c493377093b80c96d4b82cd937a6b2` — fix(#3996): drop gateway SSE keepalive frames before the genai parser
- `099eed7f742a6ff599705e36fdaa76316ac4b296` — test(#3996): pin TEXT+IMAGE request construction for image-output models

## Validation

```sh
go test -v -count=1 ./pkg/httpclient ./pkg/model/provider/gemini -run 'TestSSEFilter_KeepaliveMode|TestSSEFilter_SharedPathKeepsKeepaliveFrames|TestNewHTTPClient_SSEKeepaliveFilterOptIn|TestWantsImageResponseModalities|TestCreateChatCompletionStream_ImageResponseModalities'
```

Matched top-level tests: `pkg/httpclient`: **4**; `pkg/model/provider/gemini`: **4**.

Deferred/live scope: No live image generation or live SSE route was exercised. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.

### DIFF
--- a/pkg/httpclient/client.go
+++ b/pkg/httpclient/client.go
@@ -24,6 +24,10 @@ type HTTPOptions struct {
 	Header http.Header
 	Query  url.Values
 
+	// dropSSEKeepaliveEvents enables keepalive-frame dropping in the SSE
+	// filter transport; see WithSSEKeepaliveFilter.
+	dropSSEKeepaliveEvents bool
+
 	// cagentID resolves the persistent install UUID stamped as
 	// `X-Cagent-Id` on gateway-bound requests. It defaults to
 	// [userid.Get]; tests inject their own source via
@@ -89,7 +93,10 @@ func NewHTTPClient(ctx context.Context, opts ...Opt) *http.Client {
 
 	var wrapped http.RoundTripper = &userAgentTransport{
 		httpOptions: httpOptions,
-		rt:          &sseFilterTransport{base: rt},
+		rt: &sseFilterTransport{
+			base:                rt,
+			dropKeepaliveEvents: httpOptions.dropSSEKeepaliveEvents,
+		},
 	}
 	if httpOptions.refreshAuth != nil {
 		// Outermost, so a replayed request goes through the whole chain again.
@@ -232,6 +239,16 @@ func WithModelName(name string) Opt {
 func WithQuery(query url.Values) Opt {
 	return func(o *HTTPOptions) {
 		o.Query = query
+	}
+}
+
+// WithSSEKeepaliveFilter strips payload-free events named "keepalive".
+// The Gemini gateway emits these transport frames, but the GenAI SDK rejects
+// event-prefixed lines even when their only data is {}. Other names and
+// keepalives with meaningful payloads are deliberately left unchanged.
+func WithSSEKeepaliveFilter() Opt {
+	return func(o *HTTPOptions) {
+		o.dropSSEKeepaliveEvents = true
 	}
 }
 
--- a/pkg/httpclient/sse_filter.go
+++ b/pkg/httpclient/sse_filter.go
@@ -10,7 +10,7 @@ import (
 
 // sseFilterTransport wraps a base RoundTripper and, when the response is a
 // `text/event-stream`, replaces the body with one that strips SSE events
-// containing no `data:` lines or named `keepalive`.
+// containing no `data:` lines.
 //
 // Why this exists: some upstreams (notably OpenRouter) inject comment-only
 // keep-alive frames into their streams:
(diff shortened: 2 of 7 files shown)


aheritier (commented):
Reviewed the three-dot delta of this PR only (head `a4ac3c343`). CI is green for the head SHA (13 check-runs, 0 failures). Since this is a draft, this is a review comment rather than an approval.

The transport fix itself is right and well pinned. `pkg/httpclient/sse_filter.go` is byte-identical to the version I reviewed on 2026-07-30, and I re-verified the constraint it exists for: `google.golang.org/genai@v1.70.0` `iterateResponseStream` (`api_client.go:511-529`) fails the whole stream with `invalid stream chunk: <prefix>:<data>` for any non-blank line whose field is not `data`, so an `event: keepalive` line is fatal. Verified locally on a throwaway checkout: `go build ./...`, `go test -race` on `pkg/httpclient` + `pkg/model/provider/gemini`, `golangci-lint` (0 issues). Also verified by probe that keepalive dropping is correct across fragmented `Read` boundaries (1/2/3/7/13/26-byte chunks) and with CRLF framing, that the opt-in is gateway-only (`pkg/model/provider/gemini/client.go:186`), and that non-SSE bodies — including `application/json` 400/500 error payloads — pass through byte-identical, so the #4017 diagnostics classifier still sees the original bytes. Mutation-testing the suite (`go test -overlay`) killed the mutants that matter: dropping the `hasMeaningfulData` guard, treating `{}` as meaningful, removing the opt-in gate, and removing the gateway wiring all fail tests.

**[blocking] `pkg/model/provider/gemini/image_response_modalities_test.go` asserts a "declared-only" contract the code does not implement.** The docstring (`:23-26`) claims the suite "exhaustively covers … only for a model explicitly declared image-output-capable via `output_capabilities.image`", and cases named `direct Gemini API, declaration missing: never` (`:44`) and `gateway, declaration missing: absent` (`:243-249`) claim missing-declaration means no image output. But production resolves the flag through `base.Config.ImageOutputEnabled` → `modelinfo.ResolveOutputImage` (`pkg/modelinfo/modelinfo.go:681-702`), which only returns `false` for a missing override when the models.dev store is `nil`/unreachable. `TestWantsImageResponseModalities` bypasses that resolution entirely (it passes the boolean in from the declaration), and the streaming cases pass no `WithModelsDevStore`, so they are vacuous: with a populated catalogue entry for `gemini-2.5-flash-image` the very cases named "never/absent" would flip to requesting `TEXT+IMAGE`. The commit message of `a7d2d76a3` additionally claims it pins "catalogue resolution, override precedence", but the file references no models.dev store at all. This is not a re-report of the inference-vs-declaration behaviour raised on #4019 — the new defect is that this PR adds the artifact that makes that behaviour look pinned and epic AC ("models/routes without that declaration … remain text-only") look satisfied. Fix: rename/re-scope these cases to what they actually assert, and add at least one case with a populated `WithModelsDevStore` demonstrating the real resolution.

**[should-fix] The keepalive matcher is keyed to one literal frame shape, with no recorded evidence of what the gateway actually emits.** `pkg/httpclient/sse_filter.go:120-126` drops a frame only when the event name is exactly `keepalive` and every `data:` payload is empty or `{}`. On this opt-in path *any* `event:` line is fatal to genai, so a payload-free `event: ping` / `event: heartbeat` — or a keepalive that ever gains a payload — silently returns to the original stream-killing bug. Nothing pins the choice either: mutating the name check to a `keep` prefix, to "any non-empty event name", or to a leading-space-tolerant matcher all leave the suite green. Two asks: (1) cite the evidence for the frame shape (captured stream / gateway version / upstream issue) in the comment at `pkg/httpclient/client.go:197-204`, and state that the durable remedy is gateway-side (or genai-side) with this as a scoped workaround; (2) consider dropping any payload-free named event on this gateway-only path, and add a test that pins whichever rule you choose.

**[should-fix] Scope: the delta carries a 405-line test file for a different PR's feature.** Commit `a7d2d76a3` adds `image_response_modalities_test.go`, which tests `wantsImageResponseModalities` — code introduced by the preceding guard commit (#4020), as its own message says. Per AGENTS.md ("Make commits logical and atomic … avoid mixing unrelated changes") those tests belong with the code they pin, where they would also have exercised that PR's review findings; the PR body here describes only the SSE filter and does not mention them.

**[optional] Pre-existing (not this delta), noted for context:** an error body served with `Content-Type: text/event-stream` is swallowed entirely by the filter (probe: a 400 with `{"error":{...}}` arrives as an empty body) — harmless for the JSON error shape I verified, but worth confirming the gateway never returns event-stream-typed errors on the streaming route; and `pending` grows without bound while a frame has no terminating blank line (probe: 2.8 MB buffered, 0 bytes emitted, no error). Both predate this PR (`pkg/httpclient/sse_filter.go` from `1c80d4e31` on main).

Issue/plan checklist for this PR (epic #3996 AC "SSE keepalive frames are ignored before GenAI parsing"): keepalive frames dropped before genai ✅ (for the `event: keepalive` + payload-free shape); gateway-scoped only, direct Gemini/Vertex untouched ✅ (pinned by tests both ways); other providers' `event:` framing preserved ✅; non-SSE/error bodies unmodified ✅; unit-test evidence ✅ but with the misleading modalities suite above.

dgageot (approved):

PR #4024, opened 2026-08-22
feat(#3996): add workspace media writes and session provenance

aheritier:
## What and why

Add collision-safe workspace media writes, nested parent creation, MIME-corrected extensions, no-overwrite publication, and cleanup on failure. Persist absolute local workspace provenance across session creation surfaces and resolve it through a bounded parent walk. Relative nesting, MIME correction, collision handling, no overwrite, and partial success are retained. This is production infrastructure, not a test-only change.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4023, rather than the aggregate stack against `main`.

Publication hard-links a complete sibling temporary file to an unused final name atomically; concurrent destination creation cannot be overwritten. Filesystems without hard-link support fail safely.

## Commit inventory

Head: `8cd7c8d6182ec1a60d9a404268f88887f19c80fa`; parent SHA: `dfca90319859bd683dd1103d48ed43ed043bddd2`.

- `0b7300ac4342355e3d7a3b06d98b336f22091bf9` — feat(#3996): persist absolute WorkingDir provenance at every local creation path
- `a532ba0a62ac42d3a1d853b8fd41991ca865200b` — feat: publish workspace media without replacing existing files
- `71ec85cb5c31543e57e25d29df648c15364ac1c9` — test(#3996): cover atomic-file cleanup after partial writes
- `8cd7c8d6182ec1a60d9a404268f88887f19c80fa` — test(#3996): avoid shadowing the session package in MCP fixtures

## Validation

```sh
go test -v -count=1 ./pkg/atomicfile ./pkg/workspacemedia ./pkg/session -run 'TestWrite|Test.*WorkingDir'
```

Matched top-level tests: `pkg/atomicfile`: **6**; `pkg/session`: **11**; `pkg/workspacemedia`: **21**.

### DIFF
--- a/cmd/root/backend.go
+++ b/cmd/root/backend.go
@@ -214,6 +214,9 @@ func (b *remoteBackend) CreateSession(ctx context.Context, _ *teamloader.LoadRes
 	sessTemplate := session.New(
 		session.WithToolsApproved(req.ToolsApproved),
 		session.WithSafetyPolicy(req.SafetyPolicy),
+		// WorkingDir is intentionally not sent: the client checkout is not the
+		// remote server's workspace. The server establishes its own workspace
+		// provenance for the session it creates (see server.SessionManager).
 	)
 
 	sess, err := client.CreateSession(ctx, sessTemplate)
--- a/cmd/root/flags.go
+++ b/cmd/root/flags.go
@@ -59,12 +59,17 @@ func sessionDBPath(flagValue string) string {
 	return filepath.Join(paths.GetDataDir(), "session.db")
 }
 
-func setupWorkingDirectory(workingDir string) error {
-	if workingDir == "" {
+// setupWorkingDirectory applies a --working-dir override: it chdirs into the
+// directory and stores the absolute path back into runConfig.WorkingDir so
+// downstream consumers (session workspace provenance, toolsets, servers) see
+// an absolute root — after the chdir a relative flag value would otherwise
+// resolve against itself.
+func setupWorkingDirectory(runConfig *config.RuntimeConfig) error {
+	if runConfig.WorkingDir == "" {
 		return nil
 	}
 
-	absWd, err := filepath.Abs(workingDir)
+	absWd, err := filepath.Abs(runConfig.WorkingDir)
 	if err != nil {
 		return fmt.Errorf("invalid working directory: %w", err)
 	}
@@ -79,6 +84,7 @@ func setupWorkingDirectory(workingDir string) error {
 	}
 
 	_ = os.Setenv("PWD", absWd)
+	runConfig.WorkingDir = absWd
 	slog.Debug("Working directory set", "path", absWd)
 
 	return nil
@@ -164,7 +170,7 @@ func addGatewayFlags(cmd *cobra.Command, runConfig *config.RuntimeConfig, loadUs
 			runConfig.Providers = userCfg.GetProviders()
 		}
 
-		return setupWorkingDirectory(runConfig.WorkingDir)
+		return setupWorkingDirectory(runConfig)
 	}
 }
 
--- a/cmd/root/flags_test.go
+++ b/cmd/root/flags_test.go
@@ -409,3 +409,31 @@ func TestEnvFromFileErrorsAbortPreRun(t *testing.T) {
 		assert.Contains(t, err.Error(), "bad.env")
 	})
 }
+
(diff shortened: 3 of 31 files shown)


aheritier (commented):
Reviewed at head `9ffb3af940`. CI is green for this SHA (13/13 check-runs, no failures — `build-and-test`, `windows-tests`, `lint`, both `build-image` arches, CodeQL/Analyze, `license-check`; the three `*push*` jobs are skipped by design). Draft, so no approval — comments only.

The atomicfile work is good and, importantly, load-bearing: I re-ran the package with `go test -overlay` substituting a `write_other.go` that leaks the temp file on error, and both `TestWrite_CleansUpTempFileOnMidWriteFailure` and `…_ExistingDestination` fail. `partialFailReader` gives a deterministic post-partial-write seam (no timing dependence), the assertions check whole-directory emptiness so a leak under any temp name is caught, and `-count=3 -race` is stable. The new tests need no Windows skip (`write_windows.go:29,36` creates and removes the temp in the target directory) and introduce no root-dependent or `0o000` setup.

**[blocking] The title and description don't match the delta.** The three-dot diff against `generated-media-streaming` is 31 files / +1448−65 across four commits and is not test-only: it adds the new `pkg/workspacemedia` package (`writer.go`, 381 lines), the new public `pkg/session/working_dir.go` API (`CaptureLocalWorkingDir`, `ResolveWorkingDir`, `ErrWorkingDirUnavailable`), and behavior changes in 15 production files — including `cmd/root/flags.go` writing the absolutized `--working-dir` back into `RuntimeConfig`, plus MCP/A2A/chatserver/ACP/session-manager/delegation/compactor wiring. The content belongs in this foundations layer, but a `test(#3996):` title and a body that says only "adds regression coverage" hide a new public package and a flag-semantics change from reviewers and from the merge message. Please retitle (`feat(#3996): …`) and describe all three workstreams before marking ready for review.

**[should-fix] `TestWriteCreatesFileWithMode` does not actually pin the chmod (`pkg/atomicfile/atomicfile_test.go:15-38`).** With `write_other.go` reduced to `return atomic.WriteFile(path, r)` (chmod deleted), only `TestWriteOverwritesAndRetightensMode` fails; the create-path test still passes because `os.CreateTemp`'s hard-coded 0o600 inside `natefinch/atomic` happens to equal the asserted mode. The comment at `:15-19` claims the opposite, and calls that would silently regress use other modes: `pkg/selfupdate/selfupdate.go:640` (0o755 — the executable bit) and `pkg/board/store.go:82` (0o644). Asserting 0o640 or 0o755 makes the test load-bearing (and the "umask-derived default" wording should go).

**[should-fix] The rename-error cleanup path is unasserted.** The no-residue guarantee also covers a failing publish-rename, but on POSIX that behavior lives entirely in the dependency (`natefinch/atomic@v1.0.1/atomic.go:29-34`) and on Windows in `write_windows.go:36`; nothing exercises it, so a dependency bump could drop it unnoticed. `pkg/workspacemedia/writer_test.go` has the same gap for `writer.go:177`.

**[optional]** `atomicfile_test.go:1` switches from `package atomicfile_test` to `package atomicfile`, churning the three pre-existing tests and giving up the black-box check, although no new test needs an unexported identifier — unlike `pkg/workspacemedia/writer_test.go:266`, which genuinely needs internal `write`.

**[optional]** The comments at `atomicfile_test.go:15-19, 40-43, 65-67, 76-86` narrate plan/review history ("the plan's restored…", "inadvertently dropped when this file was rewritten") rather than behavior, which future readers can't verify — per AGENTS.md, one terse behavioral line each is enough. The two new tests also drop the `t.Parallel()` the rest of the file uses.

Validation on a throwaway checkout of this head: `go build ./...`, `golangci-lint run` (0 issues), `GOOS=windows go build ./... && go vet`, and `go test -count=1 -race` across all touched packages pass. `pkg/evaluation.TestRunDockerAgentInContainerCancelInterruptsDocker` fails under `-race`, but it fails identically on the base branch — pre-existing, unrelated to the `save.go` change here.

Note for the stack: `pkg/chat/media.go:12` still points at `[DocumentSource.ArtifactPath]`, which does not exist at this head (`pkg/chat/document.go:13` has only `InlineText`/`InlineData`) — flagged on #4023, unchanged here, presumably resolved in the materialization layer.

dgageot (dismissed):

dgageot (approved):

PR #4026, opened 2026-08-22
feat(#3996): keep generated media workspace-only with portable session copies

aheritier:
## What and why

Keep all generated-media writes inside the owning workspace. Absolute, traversal, home-relative, external, and symlink-escaping requests are redirected to a sanitized basename in the workspace with a bounded warning; there is no external-write behavior, confirmation/dialog/schema, or `ArtifactRootExternal` path. ACP foreground sessions complete and persist assistant text plus workspace artifact metadata without an elicitation consumer. After file and manifest recording, optionally store the complete bytes as a portable session blob without an application-level cap; blob failure warns and preserves the workspace file. Deleting a session removes manifest/blob rows, never workspace files. The root-kind migration remains for compatibility, and old unknown/external rows fail closed.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4025, rather than the aggregate stack against `main`.

## Commit inventory

Head: `1409e8b414370be9a472cad3d889d0cc8a2316c9`; parent SHA: `9d7da808954352fc5fd7c54aeeacc8240c0bcc2d`.

- `e58a910c6834250df13b2a802cfc63d7bca3349e` — feat: keep generated-media saves inside the owning workspace
- `1409e8b414370be9a472cad3d889d0cc8a2316c9` — feat(#3996): persist portable generated media blobs

## Validation

```sh
go test -v -count=1 ./pkg/runtime ./pkg/acp ./pkg/session ./pkg/workspacemedia ./pkg/cli -run 'TestMaterializeGeneratedMedia_Redirect|TestRunStream_EscapingGeneratedMediaCompletesWithoutElicitation|TestPrompt_EscapingGeneratedMediaCompletesWithoutElicitation|TestGeneratedMediaBlobStore|TestGeneratedMediaManifest_DeleteSessionPrunesBlob|TestRequestedBasename|TestClassifyRequestedPath|TestNonOAuthElicitationDeclinedAndStreamDrained'
```

Matched top-level tests: `pkg/acp`: **1**; `pkg/cli`: **1**; `pkg/runtime`: **4**; `pkg/session`: **2**; `pkg/workspacemedia`: **2**.

Deferred/live scope: No active database was read, copied, migrated, or mutated. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.

### DIFF
--- a/pkg/acp/agent.go
+++ b/pkg/acp/agent.go
@@ -635,11 +635,11 @@ func (a *Agent) readResourceLink(ctx context.Context, sessionID string, rl *acp.
 }
 
 func resourceLinkName(rl *acp.ContentBlockResourceLink) string {
-	if rl.Name != "" {
-		return rl.Name
+	if name := chat.SanitizeDisplayName(rl.Name); name != "" {
+		return name
 	}
 	if path, ok := resourceLinkPath(rl.Uri); ok {
-		if base := filepath.Base(path); base != "." && base != string(filepath.Separator) {
+		if base := chat.SanitizeDisplayName(filepath.Base(path)); base != "" && base != "." && base != string(filepath.Separator) {
 			return base
 		}
 	}
--- a/pkg/acp/agent_test.go
+++ b/pkg/acp/agent_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	acpsdk "github.com/coder/acp-go-sdk"
@@ -68,6 +69,19 @@ func TestBuildUserContent_ResourceLinkFallbackDoesNotExposeAbsoluteURI(t *testin
 	assert.Contains(t, content, "content unavailable")
 	assert.NotContains(t, content, uri)
 	assert.NotContains(t, content, "/var/folders")
+}
+
+func TestResourceLinkNameIsSafeAndBounded(t *testing.T) {
+	t.Parallel()
+
+	longName := strings.Repeat("é", 100) + "\nforged"
+	assert.Equal(t, chat.SanitizeDisplayName(longName), resourceLinkName(&acpsdk.ContentBlockResourceLink{Name: longName}))
+
+	got := resourceLinkName(&acpsdk.ContentBlockResourceLink{Uri: "file:///tmp/unsafe%0Aname.png"})
+	assert.Equal(t, "unsafe_name.png", got)
+	assert.LessOrEqual(t, len(got), chat.MaxSanitizedFieldBytes)
+
+	assert.Equal(t, "resource", resourceLinkName(&acpsdk.ContentBlockResourceLink{Uri: "https://example.com/private.png"}))
 }
 
 func TestBuildUserMessage_ImageContent(t *testing.T) {
--- a/pkg/acp/runagent_test.go
+++ b/pkg/acp/runagent_test.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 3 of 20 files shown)


aheritier (commented):
Reviewed the three-dot delta of this PR only (base `workspace-media-materialization` @ 41c56eba5), at head `161aaa6f0`. CI is green on that SHA (13/13 check-runs, incl. `windows-tests` — the #4025 `pkg/hooks` flake did not recur). Re-validated locally in a throwaway extract: `go build ./...`, `GOOS=windows go build ./...`, `go test -count=1 -race` on `pkg/{runtime,workspacemedia,session,cli,chat,tui/dialog}/...`, `golangci-lint` — all green, 0 issues. COMMENT rather than approve because the PR is still a draft.

Strong slice. The confirmation is fail-closed by construction rather than by convention: the elicitation carries an enum schema with the safe choice first (`pkg/runtime/media_escape.go:141-176`) **and** the runtime independently requires the exact affirmative value (`:206-211`), so accept-with-empty/free-form/other content cannot authorize a write. The TUI side is pinned against the real dialog (`pkg/tui/dialog/media_escape_schema_test.go:38-74`), not a mock. Directory targets are pre-resolved to the final filename before the prompt (`:240-242`) and a directory appearing at the confirmed path afterwards is refused outright (`pkg/workspacemedia/external.go:56-58`). The CLI now declines without abandoning the stream (`pkg/cli/runner.go:248-258`), proven on an unbuffered stream (`runner_test.go:607-654`). I mutation-tested the suite (`go test -overlay`): dropping the content check, the directory pre-resolution, or the CLI `continue` each fails tests.

**[should-fix] A2A and ACP have no fail-safe — the confirmation parks the turn.** `pkg/a2a/adapter.go:113-123` and `pkg/acp/agent.go:140-151` construct the runtime without `runtime.WithNonInteractive` (contrast `pkg/mcp/server.go:275`), neither package references elicitation at all, and both consume `rt.RunStream` directly (`adapter.go:131`, `agent.go:607`) without ever calling `ResumeElicitation`. So `hasElicitationSink()` is false, nothing answers the event, and `tools.InteractivePromptsAllowed` defaults true (`pkg/tools/interactive.go:38-41`) — neither fast-decline at `pkg/runtime/elicitation.go:498` nor `:523` fires, and `confirmMediaEscape` blocks in the `select` at `:597-611` until the client cancels. The commit message's "non-interactive/headless surfaces the bytes are redirected" is not true for these two. Latent today (nothing populates `chat.MediaDelta.RequestedPath` at this head — only `loop.go:1397` reads it), so it becomes reachable when marker extraction lands in #4027; please fix by then. A bounded wait that falls into the existing redirect path is probably the cheapest option — note that a plain `!hasElicitationSink() ⇒ decline` would break the CLI, which answers off the stream.

**[should-fix] The symlink-parent escape branch is untested.** `media_escape.go:66-71` handles the one escape lexical classification cannot see (a contained path whose parent symlinks outside the root). Deleting that branch under `-overlay` leaves the whole `./pkg/runtime` suite green; `materialize_generated_media_test.go:387-391` only covers the provider-named retry. Worth a test: workspace root with `link -> <outside>`, `RequestedPath: "link/cat.png"` ⇒ redirect + warning, nothing written outside.

**[optional]** A Windows-style absolute path on a Unix host is classified `PathEscaping` (`classify_test.go:24`) but `externalMediaTarget:226-238` anchors it under the workspace root, so the prompt says "OUTSIDE the workspace" for `<root>/C:/abs/cat.png`. Exact path shown, so nothing unsafe — just misleading wording.

**[optional]** A collision silently writes a name the user did not confirm: confirming `…/cat.png` when it exists writes `…/cat-1.png` with no notice (`external_test.go:30-46`, `media_escape_test.go:133-155`, which asserts no warnings). Never overwriting is right; a bounded notice when the final path differs from the confirmed one would preserve the "what you confirmed is what was written" promise.

**[optional]** `generatedDocumentName:254-259` uses `filepath.Base` of the confirmed path instead of `chat.SanitizeDisplayName`, so an external item's display name can carry `<`/`>` and exceed `MaxSanitizedFieldBytes` (`pkg/chat/display_name.go:49-72`), unlike every other display name.

Checked and fine: accept-then-write-failure drops the item with a sanitized warning (no path/raw-error leak) consistent with the existing per-item contract; the manifest add boundary rejects mis-rooted records and unknown roots, and unrecorded absolute lookups still fail closed (`generated_media_manifest_test.go:114-180`); migration 29's missing `DownSQL` is not a deviation (11/29 lack one, and `DownSQL` is never executed).

Cross-PR note: this PR's external write path has no per-item size cap either — same gap already raised on #4025, not re-raised here. When the blob round lands, `root_kind` needs to renumber 29 → 30 with the ID, zero-padded name, pinned digest and sequential test moving in lockstep.

dgageot (dismissed):

dgageot (dismissed):

dgageot (approved):

PR #4027, opened 2026-08-22
feat(#3996): name generated media with private markers and prompt fallback

aheritier:
## What and why

Request private `[media-file: ...]` markers, remove them from visible and persisted text, and apply naming precedence: marker, exact one-image user-prompt filename, provider name, then `generated-N`. Prompt fallback applies only when exactly one image was returned. Requested paths remain untrusted and pass through workspace-only containment, MIME correction, and collision-safe publication.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4026, rather than the aggregate stack against `main`.

## Commit inventory

Head: `da8faa29eace36625b546c4d7103b4d08543d57e`; parent SHA: `1409e8b414370be9a472cad3d889d0cc8a2316c9`.

- `69155f939bcc878aa22b53188602aeea8fbdaeb3` — feat(#3996): name generated media via [media-file:] response markers
- `da8faa29eace36625b546c4d7103b4d08543d57e` — feat(#3996): fall back to an explicit user-prompt filename for one unnamed image

## Validation

Build, test compilation, owning-package tests and the named fixture passed at this PR head.

Exact deterministic fixture command:

```sh
go test -v -count=1 ./pkg/runtime ./pkg/model/provider/gemini -run 'TestExtractExplicitOutputFilename|TestApplyUserPromptRequestedPath|TestUserPromptOfPhraseFilename|TestMarkerOverridesUserPromptFilename|TestUserPromptFilenameSkipsMultiBlobTurns|TestComparativeReferencePromptsNeverNameGeneratedMedia|TestHandleStream_MediaFileMarker|TestCreateChatCompletionStream_MediaFileInstruction'
```

Matched top-level tests: `pkg/model/provider/gemini`: **2**; `pkg/runtime`: **8**.

Deterministic scope: the named local fixture exercises this PR boundary with disposable configuration/stores and fake or loopback providers as applicable. Every listed package ran nonzero matching top-level tests.

Deferred/live scope: Live provider, remote CI, and platform execution are not claimed by this deterministic receipt. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.

PR #4028, opened 2026-08-22
fix(#3996): leave reasoning headroom for Gemini session titles

aheritier:
## What and why

Omit thinkingConfig for session-title calls without returning early from request construction. Use a shared 128-token title budget for reasoning headroom with a separate 50-character visible limit. Ordinary request thinking behavior is unchanged. This is a mitigation, not a provider-success guarantee.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4027, rather than the aggregate stack against `main`.

## Commit inventory

Head: `8eed51894134f104fa449f2cfbe7e510a22d3956`; parent SHA: `da8faa29eace36625b546c4d7103b4d08543d57e`.

- `8eed51894134f104fa449f2cfbe7e510a22d3956` — fix(#3996): leave reasoning headroom for Gemini session titles

## Validation

```sh
go test -v -count=1 ./pkg/model/provider/gemini ./pkg/sessiontitle -run 'TestBuildConfig_NoThinking|TestBuildConfig_Gemini25_ThinkingBudget|TestBuildConfig_Gemini3_ThinkingLevel|TestGenerateOnceUsesTitleHeadroomAndClearsStructuredOutput|TestGenerator_Generate_FallsBackOnEmptyOutput'
```

Matched top-level tests: `pkg/model/provider/gemini`: **4**; `pkg/sessiontitle`: **2**.

### DIFF
--- a/pkg/model/provider/gemini/client.go
+++ b/pkg/model/provider/gemini/client.go
@@ -425,7 +425,7 @@ func extractMimeType(dataURLPrefix string) string {
 	return "image/jpeg" // Default fallback
 }
 
-// buildConfig creates GenerateContentConfig from model config
+// BuildConfig creates GenerateContentConfig from model config.
 func (c *Client) buildConfig() *genai.GenerateContentConfig {
 	config := &genai.GenerateContentConfig{}
 	if c.ModelConfig.MaxTokens != nil {
@@ -453,7 +453,11 @@ func (c *Client) buildConfig() *genai.GenerateContentConfig {
 	// Apply thinking configuration for Gemini models.
 	// See https://ai.google.dev/gemini-api/docs/thinking
 	if c.ModelOptions.NoThinking() {
-		// NoThinking requested (e.g. title generation). For Gemini 3+ models
+		if c.ModelOptions.GeneratingTitle() {
+			return config
+		}
+
+		// NoThinking requested (e.g. MCP sampling). For Gemini 3+ models
 		// that always think, use the lowest level and bump MaxOutputTokens so
 		// internal reasoning doesn't consume the entire budget. Gemini 2.5 and
 		// older can fully disable thinking with ThinkingBudget=0.
--- a/pkg/model/provider/gemini/client_test.go
+++ b/pkg/model/provider/gemini/client_test.go
@@ -10,9 +10,73 @@ import (
 	"github.com/docker/docker-agent/pkg/chat"
 	"github.com/docker/docker-agent/pkg/config/latest"
 	"github.com/docker/docker-agent/pkg/model/provider/base"
+	"github.com/docker/docker-agent/pkg/model/provider/options"
 	"github.com/docker/docker-agent/pkg/modelsdev"
 	"github.com/docker/docker-agent/pkg/tools"
 )
+
+func TestBuildConfig_NoThinking(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name          string
+		model         string
+		opts          []options.Opt
+		wantThinking  bool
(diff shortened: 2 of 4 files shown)


aheritier (commented):
Reviewed the three-dot delta vs `workspace-media-naming` (b18b517f7): 1 commit, 2 files (+70/-2). CI is green for 97877c1f8 (13 check-runs, incl. lint, build-and-test, CodeQL, windows-tests; publish jobs skipped). Locally: build + GOOS=windows build clean, `go test -race` green on pkg/model/provider/gemini, pkg/sessiontitle, pkg/model/provider, golangci-lint 0 issues, and the new test is load-bearing (reverting the production hunk fails `title_generation_omits_thinking_config`).

**[blocking] Gemini 3 title generation loses the safeguard the comment still documents.** `client.go:455-458` returns before the Gemini-3 branch at :464-472, which set ThinkingLevelLow *and* raise MaxOutputTokens to a 200-token floor. Title generation clones with MaxTokens=20 (sessiontitle/generator.go:32,:148 -> clone.go:56-58), so the effective config for gemini-3-flash goes from `{ThinkingLevel:LOW, MaxOutputTokens:200}` to `{ThinkingConfig:nil, MaxOutputTokens:20}`. Per the comment retained at :460-463, Gemini 3+ models always think, so an omitted thinkingConfig means default thinking against a 20-token cap -> MAX_TOKENS with no visible text, which generateOnce reports as "empty title output" (generator.go:164-167) and Generate then burns every fallback (:135). Please narrow the omission to models that actually reject thinkingConfig (image-output models), or keep the MaxOutputTokens floor when omitting it, and add a title case asserting that floor.

**[blocking] Title/body describe this change inversely.** The only production change gates the existing NoThinking branch on GeneratingTitle(); the user-budget branch (:479-486) is untouched, so ordinary/image-output requests are byte-identical to base. The body claims the opposite ("preserves ... for ordinary image-output requests while retaining the special title-generation behavior") and lists modality/guard/runtime-TUI validation that is not in this delta. The commit body is accurate — please align title + body (and note that Compacting() is deliberately unaffected: compactor.go:202-206 never sets NoThinking).

**[should-fix]** `return config` at :457 also bypasses the structured-output block at :488-491 (verified: title+structured yields ResponseMIMEType "" vs "application/json" on the non-title path). Harmless today only because sessiontitle passes WithStructuredOutput(nil) (generator.go:147). Prefer scoping the condition: `if c.ModelOptions.NoThinking() && !c.ModelOptions.GeneratingTitle() { ... }`.

**[should-fix]** :428 doc comment says `BuildConfig` for the unexported `buildConfig` — and it's an unrelated drive-by edit in an otherwise focused diff.

**[optional]** The added cases use gemini-3-flash / gemini-2.5-flash; the motivating image-capable model (e.g. gemini-2.5-flash-image) isn't exercised, and nothing pins that ordinary image-model requests still carry the user's budget.

Checked and clean: no transport split (buildConfig has no apiSurface branching); ResponseModalities is applied after buildConfig (:763-768) so the #4020 guard is intact; the reranking caller (:881) overrides ThinkingConfig explicitly; no clone shared-state mutation (clone.go copies by value, -race clean); `thinking_budget: 0` still disables thinking; modelerrors untouched. The pkg/mcp `clientSession` rename appears exactly once in the stack (9ffb3af94, #4024 lane) — no duplication.

dgageot (dismissed):

dgageot (dismissed):

dgageot (dismissed):

dgageot (approved):

dgageot (approved):

PR #4029, opened 2026-08-22
feat(#3996): render generated images from manifest-gated portable blobs

aheritier:
## What and why

Revalidate generated-file authorization against the current manifest on every resolution, by owner, path, and workspace root. Session deletion revokes references even when workspace provenance is cached. Invalid POSIX/Windows paths and unknown/external roots are rejected before any lookup. Prefer portable blobs. Fall back only when the optional blob interface is absent or the blob is not found, and only to the owning workspace; other blob errors fail closed. Portable bytes may survive a missing WorkingDir and then use the recorded relative display path, not a verified absolute path. Legacy ordinary-file contents are not integrity-verified. Resolve bytes off the TUI Update path and preserve adjacent text.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4028, rather than the aggregate stack against `main`.

## Commit inventory

Head: `950d801ff3a2eb4ee5aaf56265bf6fe8faa54110`; parent SHA: `8eed51894134f104fa449f2cfbe7e510a22d3956`.

- `8c2a76fad28771134b95cd1d797221403ea4f3f8` — feat: render generated images through workspace manifest authorization
- `950d801ff3a2eb4ee5aaf56265bf6fe8faa54110` — feat: restore generated media from portable session blobs

## Validation

```sh
go test -v -count=1 ./pkg/runtime ./pkg/tui/components/message ./pkg/tui/components/messages ./pkg/tui/page/chat -run 'TestResolveGeneratedFile|TestAppendAssistantMediaPreservesStreamedTextRendering|TestLoadFromSession.*Media|TestUpdateAssistantMedia|TestAssistantMedia|Test.*GeneratedMedia'
```

Matched top-level tests: `pkg/runtime`: **52**; `pkg/tui/components/message`: **4**; `pkg/tui/components/messages`: **5**; `pkg/tui/page/chat`: **3**.

Deferred/live scope: No active database or remote runtime was exercised. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -1267,6 +1267,36 @@ type contextBreakdownProvider interface {
 	ContextBreakdown(ctx context.Context, sess *session.Session) (*runtime.ContextBreakdown, error)
 }
 
+// generatedFileResolver is an optional runtime capability: resolving one
+// recorded generated-media reference to its bytes and validated canonical
+// path, gated on the generated-media manifest and the owning session's
+// workspace (see [runtime.LocalRuntime.ResolveGeneratedFile]). Only the
+// local runtime implements it; remote runtimes never deliver generated-file
+// payloads, so UIs treat the missing capability as "nothing to resolve".
+type generatedFileResolver interface {
+	ResolveGeneratedFile(ctx context.Context, ref runtime.GeneratedFileRef) (*runtime.ResolvedGeneratedFile, error)
+}
+
+// CanResolveGeneratedFiles reports whether the runtime can resolve
+// generated-media references at all, letting UIs skip resolution work
+// entirely on runtimes without the capability.
+func (a *App) CanResolveGeneratedFiles() bool {
+	_, ok := a.runtime.(generatedFileResolver)
+	return ok
+}
+
+// ResolveGeneratedFile resolves one recorded generated-media reference.
+// Returns an error wrapping [runtime.ErrUnsupported] when the runtime does
+// not own local generated media (e.g. remote runtimes). Callers must treat
+// any error as "unavailable" — never surface its text to the user.
+func (a *App) ResolveGeneratedFile(ctx context.Context, ref runtime.GeneratedFileRef) (*runtime.ResolvedGeneratedFile, error) {
+	resolver, ok := a.runtime.(generatedFileResolver)
+	if !ok {
+		return nil, fmt.Errorf("generated file resolution: %w", runtime.ErrUnsupported)
+	}
+	return resolver.ResolveGeneratedFile(ctx, ref)
+}
+
 // ContextBreakdown returns the estimated context-window composition for the
 // current session. Returns an error wrapping [runtime.ErrUnsupported] when
 // the runtime cannot compute it (e.g. remote runtimes).
--- a/pkg/app/generated_file_test.go
+++ b/pkg/app/generated_file_test.go
@@ -1,0 +1,57 @@
+package app
+
+import (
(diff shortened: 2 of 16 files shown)


aheritier (commented):
Reviewed the #4029 delta only (97877c1f8...20a7918fb). CI green for 20a7918fb (build-and-test, lint, windows-tests, CodeQL); locally `go build`, `-race` tests on touched pkgs and golangci-lint are clean, and an `-overlay` mutation removing the manifest lookup in `ResolveGeneratedFile` fails `TestResolveGeneratedFile_ForgedReferenceRefused`, so the gate is load-bearing.

Verified:
- Manifest-gated: the TUI only extracts `Document.Source.ArtifactPath/Root/OwnerSessionID` parts persisted by materialization (`generated_media.go: generatedImageMedia`), never model text; `ResolveGeneratedFile` requires an (owner, path) manifest record with matching root kind, confines workspace reads with `os.OpenRoot` + per-component `Lstat` symlink rejection + post-open `SameFile` re-check, 20 MiB size cap before and during read. Negative tests cover forged `.env`/`main.go`, `../` traversal, symlink leaf and symlink parent inside the workspace, cross-workspace owner, external leaf symlink, oversized, missing file, legacy/unknown root.
- Sub-agent artifacts (owner = child session) resolve live via the materialization-seeded cache and lazily via `session.ResolveWorkingDir` parent fallback (`TestResolveGeneratedFile_ParentWorkingDirFallback`); they are not re-rendered on restore because the parent `sess.Messages` doesn't contain child-session messages — consistent with sub-agent text today.
- Off the update loop: placeholders attach synchronously, bytes resolve inside a `tea.Cmd` recorded like routed timers (survives background tabs); asserted by `rt.resolveCalls()==0` before the cmd runs. Positive-only per-owner cache; store hit counts pinned.
- Fallbacks: remote runtime (`Message` is `json:"-"` → nil, and no `generatedFileResolver` capability) → no-op; graphics disabled / undecodable / deleted file → sanitized text fallback, no raw error/owner-ID/ref leak (tests pin this).

[optional] `tuiimage.FromBytes` → `chat.ResizeImage` fully decodes before the `maxDecodedDimension` check, and on resize error `FromBytes` falls through to `stdimage.Decode` on the original bytes, so the dimension guard is not enforced on this path. Pre-existing in `main` (markdown images share it), not this PR's delta — a `DecodeConfig` pre-check in `FromBytes` would harden both paths; fine as a follow-up.

Stack note (rev135 DB-blob rewrite): the TUI side (`generated_media.go`, `messages.go`, `message.go`, `types.go`, `app.go` capability seam) consumes only `ResolvedGeneratedFile{Data,Path}` and would survive unchanged; the rewrite touches `pkg/runtime/generated_file.go` internals (DB-first lookup, `maxGeneratedFileBytes` → `session.MaxGeneratedBlobBytes`) and flips `TestResolveGeneratedFile_MissingFileRefused` semantics (deleted file → still renders from the blob).

dgageot (dismissed):

dgageot (dismissed):

dgageot (dismissed):

dgageot (dismissed):

dgageot (approved):

dgageot (approved):

trungutt (approved):

PR #4030, opened 2026-08-22
docs(#3996): describe portable generated media and Gemini output limits

aheritier:
## What and why

Align model, Google-provider, session, TUI, and example references with catalogue/override eligibility, ordinary-request guards, safe workspace-only naming, portable session blobs, no-resend history, and no generated-media size cap. Document the phrase-warning blind spots and exactly-one-image filename fallback. Database compatibility prose is caution-only: migrations happen in place, older binaries may not understand the upgraded schema, and compatibility tests use isolated disposable stores. It provides no operational database procedure.

Part of docker/docker-agent#3996. Review this PR against its immediate parent, docker/docker-agent#4029, rather than the aggregate stack against `main`.

## Commit inventory

Head: `4b04a89084ac6ea9d35630b3e1111ee73ceeb18a`; parent SHA: `950d801ff3a2eb4ee5aaf56265bf6fe8faa54110`.

- `0df48c0b85490c0b3bb430e3bdff187d6b75acf5` — docs: explain the workspace artifact lifecycle
- `4c0d297f5df9410a1f3222264d790c4814400f17` — docs: describe generated-image requests and workspace-only saves
- `4b04a89084ac6ea9d35630b3e1111ee73ceeb18a` — docs: clarify uncapped generated-media storage and resolution

## Validation

```sh
go test -v -count=1 ./pkg/modelinfo ./pkg/model/provider/gemini ./pkg/runtime -run 'TestResolveOutputImage|TestCheckImageOutputRequestCompatibility|TestExtractExplicitOutputFilename|TestResolveGeneratedFile_PortableBlobWinsOverWorkspace|TestResolveGeneratedFile_LegacyManifestRoundTripsLargeWorkspaceFile|TestImageGenerationTextOnlyWarning'
```

Matched top-level tests: `pkg/model/provider/gemini`: **2**; `pkg/modelinfo`: **1**; `pkg/runtime`: **4**.

Deferred/live scope: Markdown lint and canonical-front-matter checks passed. Live provider and active-database checks were not run. The final stack head passed build, lint, full tests, an uncached full suite, focused race tests and documentation checks in disposable environments. Remote CI is tracked by the checks below; no new paid-provider or active-database validation was run.


docs/features/sessions/index.md
```diff
@@ -29,6 +29,22 @@ Override the location with `-s`/`--session-db`, or by overriding the data direct
 $ docker agent run agent.yaml --session-db ./sessions.db
 ```
 
+## Generated Media Files
+
+Some models (e.g. Gemini image-output models) can generate an image as part
+of a reply. Docker Agent saves each generated image into the session's
+workspace — the directory the session was started in — as an ordinary,
+untracked file, not into `session.db` or the data directory. The session
```
rumpl, line 37: We should keep it in the database, sessions should be portable

rumpl (dismissed):

docs/features/sessions/index.md
aheritier: ok I can try this @rumpl. 
I might limit their size to 20Mb in the DB WDYT?

aheritier (commented):
Docs-only review of this PR's own delta vs `generated-image-inline-tui` (2 commits, 6 files). CI is fully green at d04b1723 (16 success, 3 skipped, 0 failing) — including every docs gate. Most of the prose checks out line-for-line against the code in the stack; the findings below are the exceptions.

**[blocking] The unresolved review request on this exact paragraph is still unaddressed.**
`docs/features/sessions/index.md:35-37` — "saves each generated image into the session's workspace … **not into `session.db` or the data directory**" is the line reviewed in https://github.com/docker/docker-agent/pull/4030#discussion_r3841042282 ("We should keep it in the database, sessions should be portable"), still unresolved and not outdated at this head. The reply offering a 20 MiB in-DB cap is unanswered, and no DB-backed blob storage exists in the stack yet: `pkg/session/migrations.go` tops out at migration 29 (`029_add_root_kind_to_generated_media_manifest`) — manifest rows only, no blob table, no blob write path. The sentence is truthful about today's code but publishes an architecture that has been objected to. The dependent claim at `docs/features/tui/index.md:344-346` ("a file replaced by another valid image at the same path shows its current bytes") documents replace semantics that DB-backed storage would invert. Resolve the thread first — either the stack lands blobs and both pages are rewritten, or workspace-only storage is confirmed and the thread closed.

**[blocking] The title-generation / compaction claims are the opposite of what the guard does.**
`docs/configuration/models/index.md:181-182` — "Session-title requests remain text-only even for image-output-capable models, so those models remain eligible as title candidates" — and `docs/providers/google/index.md:88` — "Title generation and compaction remain text-only."
The modality half is true: `pkg/model/provider/gemini/client.go:744-750` is the only place `GeneratingTitle()`/`Compacting()` are honoured. But the request guard is not scoped that way: `pkg/model/provider/gemini/image_output_guard.go:63-67` early-returns only on `!imageOutputEnabled` and never consults those options, and it is called at `client.go:798` with `builtInTools` from `client.go:541-559`, i.e. model-level `provider_opts` (`google_search`, `google_maps`, `code_execution`). So for an image-output-capable model with any Gemini built-in tool enabled — the combination documented two sections below on the same page (`docs/providers/google/index.md:152`, "Built-in Tools (Grounding)") — a text-only title request (`pkg/sessiontitle/generator.go:154`, nil tools) or compaction request (`pkg/runtime/compactor/compactor.go:202-205`) is rejected with "…does not support built-in tools in the same request…", ending in `all N title model(s) failed` (`generator.go:135`). The guard's over-broad scope was already raised on #4021; what is new here is documentation asserting the intended behaviour. Either narrow the guard (skip when `!wantsImageResponseModalities`) or state the exception in both places.

**[should-fix] This PR deletes the only documentation of the no-resend guarantee, which is still live code.**
The removed `docs/features/tui/index.md` paragraph ("Generated media is **not** automatically resent to the model on later turns…") has no replacement anywhere in `docs/` (no remaining hit for "resent"), yet the behaviour is intact: `stripGeneratedMediaTransform` (`pkg/runtime/strip_generated_media.go:82`), registered at `pkg/runtime/runtime.go:741`, swaps each generated artifact for a bounded text placeholder in outgoing history. Epic #3996 lists "Generated media is not resent as conversation history" as a user-visible workflow. Worth one sentence back, including the placeholder users actually see.

**[should-fix] The prompt-filename claim is unconditional; the fallback is single-image-only.**
`docs/features/tui/index.md:319-322` — "A filename named explicitly in your prompt … is honored". `pkg/runtime/generated_media_prompt_filename.go:166-171` returns immediately unless `len(media) == 1` and no marker already filled `RequestedPath`. Verified against this head: a single-blob turn with "Generate an image of a lighthouse at sunset as assets/lighthouse.png" yields `assets/lighthouse.png`, while a two-blob turn with the same prompt leaves both paths empty. So on a multi-image turn where the model ignores the `[media-file:]` markers, prompt naming silently does not apply. Epic #3996 words it precisely ("a single unnamed image can fall back to an explicit user-prompt filename"); the doc drops the qualifier. Also worth saying the marker line is the primary mechanism and the prompt scan a narrow, deliberately non-NLP backstop (same file:11-26).

**[should-fix] Heading demotion misfiles the section.**
`docs/configuration/models/index.md:154` turns `## Output Capabilities` into `### Output capabilities`, nesting it under `## Attachment Capability Overrides` (:87) — the very concept its first sentence contrasts itself against (input vs output). The anchor still resolves (:79, `docs/features/tui/index.md:298`) so nothing breaks, but the page TOC now presents output capabilities as a sub-topic of attachment overrides, and the cross-reference label at :79 still reads "Output Capabilities" in Title Case like every other heading in the file. Suggest keeping `##`.

**[needs-decision] models.dev inference vs the epic's acceptance criterion.**
The new prose correctly describes the implemented contract — `output_capabilities` is an override, omitted flags come from models.dev (`pkg/model/provider/base/base.go:87-95` → `modelinfo.ResolveOutputImage`; docs at `models/index.md:156-159`, `:174-179`, `google/index.md:85-87`). But epic #3996 states "Image generation is opt-in and owner-declared through model `output_capabilities.image` … The configuration remains the authority for image-output eligibility". This PR is where that divergence becomes published behaviour. Owner call needed: update the epic, or revert the inference (code side belongs to #4019/#4020).

For the record, these claims were checked and hold at this head: extension-correction notice (`pkg/runtime/loop.go:1440-1443`), dash-suffix collision naming (`pkg/workspacemedia/writer.go:133`), per-item save-failure isolation keeping text and siblings (`loop.go:1387-1404`, `1425-1428`), fail-closed escape confirmation with "keep it in the workspace" as the default and a verified submitted choice (`pkg/runtime/media_escape.go:139-150`, `176-213`), redirect-with-warning and "bytes are never discarded" (`media_escape.go:117-135`), subdirectory creation (`writer.go:100`), the "unavailable" note covering deleted/moved/symlinked/undecodable/oversized files via the 20 MiB display bound (`pkg/runtime/generated_file.go:29`, `275-289`; `pkg/tui/page/chat/generated_media.go:141`, `192`), the remote-runtime caveat (`generated_media.go:66`, `92`), graphics/`render_images` gating shared with Markdown images (`pkg/tui/image/image.go:62-63`, `339-341`; `pkg/tui/components/message/message.go:764-777`), the `first_available` validation constraint (`pkg/config/latest/validate.go:175-177`), the honest "Vertex must pass live acceptance checks" scoping, and `examples/gemini_image_output.yaml` (schema-valid; all three "Try it out" prompts extract as documented). No secrets, tokens, or internal URLs in the diff.

dgageot (dismissed):

dgageot (dismissed):

dgageot (dismissed):

dgageot (approved):

dgageot (approved):

PR #4037, opened 2026-08-22
TUI - Optimize rendering of long streaming responses

krissetto:
Improves cpu usage during streaming of long responses by caching better and not re-rendering the entire response on each new chunk that arrives. 

Videos below show an example of the before and after using the same thread

2/3 of the diff are tests in a separate commit

---

#### Before

main@2a70168d4
initial ram usage seems lower, but rises and bounces around during streaming
cpu usage rises as the response continues, peaking above 50% in this example

[Screencast From 2026-08-22 20-24-47.webm](https://github.com/user-attachments/assets/b082fd19-955d-4724-802f-9331ccfbe060)

---

#### After
initial ram usage is higher, but remains mostly flat or goes down a tiny bit during streaming. final usage is not meaningfully higher
cpu usage remains essentially flat around 9-12% in this example for the entire duration of the response

[Screencast From 2026-08-22 20-21-38.webm](https://github.com/user-attachments/assets/13d792a7-adc1-4cd0-b711-1417ff11deb7)

### DIFF
--- a/pkg/tui/components/message/message.go
+++ b/pkg/tui/components/message/message.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	tea "charm.land/bubbletea/v2"
+	"charm.land/lipgloss/v2"
 	"github.com/charmbracelet/x/ansi"
 
 	"github.com/docker/docker-agent/pkg/tui/animation"
@@ -28,6 +29,7 @@ type Model interface {
 	layout.Model
 	layout.Sizeable
 	SetMessage(msg *types.Message) tea.Cmd
+	AppendContent(content string) tea.Cmd
 	SetSelected(selected bool)
 	SetHovered(hovered bool)
 	CodeBlocks() []markdown.CodeBlock
@@ -41,6 +43,19 @@ type Model interface {
 	// to assert that finalized views have actually released their per-message
 	// render state without reaching into unexported fields via reflection.
 	HasLiveRenderState() bool
+	// RenderedSegments exposes immutable header/stable blocks separately from
+	// the mutable tail so transcript follow-tail rendering need not flatten the
+	// complete active response on every chunk.
+	RenderedSegments(width int) (AssistantSegments, bool)
+}
+
+// AssistantSegments is a line-oriented active assistant rendering. Header and
+// Stable are retained by the message view and immutable until the next width
+// change; Tail contains only the mutable markdown block.
+type AssistantSegments struct {
+	Header []string
+	Stable []string
+	Tail   []string
 }
 
 // messageModel implements Model
@@ -72,7 +87,11 @@ type messageModel struct {
 	// mdRenderer is reused across renders of an assistant message so that
 	// streamed-in chunks only re-render the trailing block instead of the whole
 	// accumulated markdown each time.
-	mdRenderer *markdown.IncrementalRenderer
+	mdRenderer        *markdown.IncrementalRenderer
+	streamLines       assistantStreamLines
+	segmentCodeBlocks []markdown.CodeBlock
+	contentBuf        strings.Builder
+	imageScanOffset   int
 
 	// finalized is set by Finalize() once the message is no longer the active
 	// streaming view. After it is set, Render() still produces correct output,
@@ -85,6 +104,14 @@ type messageModel struct {
 	markdownImages  map[string]tuiimage.Inline
 	loadingImages   map[string]bool
 	markdownImageID int
+}
+
+type assistantStreamLines struct {
+	width       int
+	stable      string
+	stableLines []string
+	headerKey   string
+	headerLines []string
 }
 
 type markdownImagesLoadedMsg struct {
@@ -119,13 +146,19 @@ type renderCache struct {
 
 // New creates a new message view
 func New(ar *animation.Runtime, msg, previous *types.Message) *messageModel {
+	imageScanOffset := -1
+	if msg != nil && msg.Type == types.MessageTypeAssistant {
+		refs := tuiimage.MarkdownReferences(msg.Content)
+		imageScanOffset = nextUnresolvedImageOpener(msg.Content, 0, refs)
+	}
 	return &messageModel{
-		message:  msg,
-		previous: previous,
-		width:    80, // Default width
-		height:   1,  // Will be calculated
-		focused:  false,
-		spinner:  spinner.New(ar, spinner.ModeBoth, styles.SpinnerDotsAccentStyle),
+		message:         msg,
+		previous:        previous,
+		width:           80, // Default width
+		height:          1,  // Will be calculated
+		focused:         false,
+		imageScanOffset: imageScanOffset,
+		spinner:         spinner.New(ar, spinner.ModeBoth, styles.SpinnerDotsAccentStyle),
 	}
 }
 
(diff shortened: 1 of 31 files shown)


pkg/tui/tui_stream_motion_scaling_test.go
```diff
@@ -0,0 +1,104 @@
...
+		return m, nil
+	}
+	updated, cmd := m.root.Update(msg)
+	m.root = updated.(*appModel)
+	return m, cmd
+}
+
+func (m *streamingMotionModel) View() tea.View {
+	m.views.Add(1)
+	if !m.root.viewCacheValid {
```
docker-agent, line 56: **[medium] Plain `bool` field `viewCacheValid` read in `View()` concurrent with `Update()` writes — potential data race**

`streamingMotionModel.View()` reads the bare `bool` field `m.root.viewCacheValid` on line 49 to decide whether to count a composition:

```go
func (m *streamingMotionModel) View() tea.View {
    m.views.Add(1)
    if !m.root.viewCacheValid {   // ← reads plain bool
        m.compositions.Add(1)
    }
    v := m.root.View()
    ...
}
```

BubbleTea (even v2) calls `View()` from a dedicated render goroutine, while `Update()` — which writes `viewCacheValid` on `*appModel` — runs on a separate event-loop goroutine. Reading and writing a plain `bool` from two goroutines without synchronization is a data race under the Go memory model and will be flagged by `go test -race`.

**Impact**: With `-race` enabled the test will report a data race, potentially causing it to fail in CI. Even without `-race`, the Go memory model does not guarantee that a write in `Update()` is ever visible to a concurrent `View()` read — so `compositions` counts could be permanently wrong (always seeing the stale initial value), making the bounded-composition assertion meaningless.

**Fix**: Use `sync/atomic` or a mutex around the `viewCacheValid` check, or — since `viewCacheValid` tracks cached render state on `appModel` — ensure that field itself is already accessed only from the BubbleTea event loop and that the test only reads it via the same synchronisation mechanism (e.g., read it inside `Update()` via a `streamingMotionRead` message, rather than directly in `View()`).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

- [medium] `pkg/tui/tui_stream_scrolledup_test.go:59` — Racy reset of atomic counters (`compositions`, `writes`) while render goroutine may be incrementing them (confidence: weak 52/100)
- [medium] `pkg/tui/tui_stream_scrolledup_test.go:63` — Sleep-based timing assumptions make render-count assertions flaky on slow CI (confidence: weak 52/100)
- [low] `pkg/tui/components/scrollview/scrollview.go:327` — `composePadded` silently drops trailing scrollbar lines when `sbLines` is longer than the content lines slice (informational)
- [low] `pkg/tui/tui_stream_motion_scaling_test.go:103` — `root.ar.Stop()` called after `program.Quit()` / `<-done`; if the animation runtime sends into the quit program a goroutine leak or panic may occur (informational)
- [low] `pkg/tui/components/message/message.go:219` — `imageScanTail` sliced by raw byte index may produce invalid UTF-8 when the final character is a multi-byte rune (informational)

pkg/tui/tui_stream_motion_scaling_test.go
krissetto: Fixed in 2235eafcd: the program test model now serializes Update/View access, counter resets were replaced with quiesced baselines, and sleep-based flush waits were replaced by bounded quiescence polling. Validated with focused tests x10 and the two program tests under `go test -race` x3.

docker-agent: Excellent fix! The sync.Mutex serializing Update()/View() access (lines 32, 51) eliminates the data race entirely — both the viewCacheValid read and the counter increments are now protected. Moving from reset-based deltas to quiesced baselines and replacing sleeps with bounded polling makes the test deterministic and race-detector-safe.

Validated under -race x3 is the right level of rigor for concurrent program tests. Thanks for the thorough resolution!

docker-agent (commented):

4 medium-severity findings in the new caching and deferred-tail logic. No high-severity or security issues detected.

pkg/tui/components/messages/messages.go
```diff
@@ -944,15 +1034,85 @@ func (m *model) scrollToTop() {
...
+	msg := m.messages[m.deferredTailIndex]
+	var b strings.Builder
+	b.Grow(len(msg.Content) + deferredBytes(m.deferredTail))
+	b.WriteString(msg.Content)
+	for _, chunk := range m.deferredTail {
+		b.WriteString(chunk)
+	}
+	msg.Content = b.String()
+	index := m.deferredTailIndex
+	_ = m.views[index].(message.Model).SetMessage(msg)
```
docker-agent: **[medium] `materializeDeferredTail` discards the `tea.Cmd` returned by `SetMessage`, silently dropping image-load side-effects**

`SetMessage` returns a `tea.Cmd` that triggers `loadMarkdownImageReferences` for any markdown images in the message content. In `materializeDeferredTail`, this return value is explicitly discarded with `_ = m.views[index].(message.Model).SetMessage(msg)`.

The deferred tail exists precisely because the user scrolled away during streaming — so the final chunks (which may introduce new image references) never went through `AppendContent`'s image scanning path. When the tail is materialized, `SetMessage` may produce a non-nil image-load command, but that command is dropped. Any inline images that appear only in the deferred portion of the response will remain unresolved.

**Suggested fix:** capture and return the cmd from `materializeDeferredTail`, then include it in the command batch at each call site (e.g. `materializeDeferredTailForInteraction` / `FinalizeStream`).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/tui/page/chat/runtime_events.go
```diff
@@ -345,6 +345,10 @@ func (p *chatPage) handleAgentSwitching(msg *runtime.AgentSwitchingEvent) tea.Cm
 }
 
 func (p *chatPage) handleStreamStopped(msg *runtime.StreamStoppedEvent) tea.Cmd {
+	// Stream completion is an exact-content boundary even when the viewport is
+	// scrolled above the active response. Reconcile buffered chunks before any
+	// queue/export/copy consumer can observe the final message.
+	p.messages.FinalizeStream()
```
docker-agent: **[medium] `FinalizeStream` called unconditionally before the sub-agent depth check in `handleStreamStopped`**

`p.messages.FinalizeStream()` is called at the very top of `handleStreamStopped`, before `p.streamDepth` is inspected. `FinalizeStream` materializes the deferred tail and resets `deferredTailIndex` to `-1`.

In a sub-agent delegation, when a sub-agent's `StreamStopped` arrives while the parent stream is still running (`streamDepth > 1` before the decrement), the parent's deferred tail — accumulated because the user has scrolled — is flushed prematurely. After `deferredTailIndex` is reset to `-1`, subsequent chunks from the parent stream that arrive while the user remains scrolled away are re-deferred into a fresh `deferredTail` slice. The chunks that landed before the premature flush were already folded into the message, so the content is not lost, **but** the flush triggers `ScrollToBottom` at the top of the `streamDepth > 0` return path, forcing an unwanted scroll-to-bottom on the still-active parent response.

Consider guarding `FinalizeStream()` on `p.streamDepth <= 1` (i.e., only finalize for the outermost stopped stream), matching the pattern used for the rest of the cleanup logic below.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

pkg/tui/components/message/message.go
```diff
@@ -260,6 +357,78 @@ func (mv *messageModel) IsToggleLine(lineIdx int) bool {
...
+		cache.stableLines = append(cache.stableLines, styledAssistantLines(messageStyle, width, delta)...)
+		cache.stable = parts.StablePrefix
+	}
+	header := actionRow(innerWidth, mv.hovered, types.MessageCopyLabel)
+	prefix := ""
+	if !mv.sameAgentAsPrevious(msg) {
+		prefix = mv.senderPrefix(msg.Sender)
+	}
+	headerKey := prefix + header
+	if cache.headerKey != headerKey || cache.width != width {
```
docker-agent: **[medium] Header cache not invalidated on terminal width change when `headerKey` is unchanged**

In `RenderedSegments`, the width-change guard at line 377 resets the stable-line cache AND immediately writes the new width into `cache.width`:

```go
if cache.width != width || !strings.HasPrefix(parts.StablePrefix, cache.stable) {
    cache.width, cache.stable = width, ""   // cache.width updated HERE
    cache.stableLines = nil
}
```

Then the header-line invalidation check at line 395 is:

```go
if cache.headerKey != headerKey || cache.width != width {
```

Because `cache.width` was just set to `width` in the block above, the `cache.width != width` sub-condition is **always false** after a width change. Header lines are only re-rendered when `headerKey` also changes. If the sender name and hover state are unchanged, `headerKey` stays the same, and the cached header lines (rendered at the old width) are served at the new terminal width.

Header lines include the sender prefix styled via `lipgloss`, which is width-sensitive. A width change without a re-render of headers produces misaligned or truncated header text until the next hover or action change forces a `headerKey` update.

**Suggested fix:** add an explicit width check for the header cache, separate from the `headerKey` check:

```go
if cache.headerKey != headerKey || cache.headerWidth != width {
    cache.headerKey = headerKey
    cache.headerWidth = width
    // rebuild header lines...
}
```

Alternatively, reset `cache.headerKey = ""` inside the width-change block at line 377–380 so the subsequent check re-fires unconditionally.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/tui/components/scrollview/scrollview.go
```diff
@@ -296,6 +296,57 @@ func (m *Model) ViewWithLines(visibleLines []string) string {
...
 
+// ViewWithPaddedLines renders pre-sliced lines that are already exactly
+// ContentWidth columns wide. It skips ANSI/grapheme measurement and wrapping;
+// callers must uphold the width contract.
+func (m *Model) ViewWithPaddedLines(visibleLines []string) string {
+	if m.width <= 0 || m.height <= 0 {
+		return ""
+	}
+	m.syncScrollbar()
+	if m.NeedsScrollbar() && len(visibleLines) < m.height {
```
docker-agent, line 307: **[medium] `ViewWithPaddedLines` only pads to viewport height for `NeedsScrollbar`, skipping the `reserveScrollbarSpace` case**

`ViewWithPaddedLines` pads the input slice to `m.height` only when `NeedsScrollbar()` is true:

```go
if m.NeedsScrollbar() && len(visibleLines) < m.height {
    result := make([]string, m.height)
    copy(result, visibleLines)
    visibleLines = result
}
return m.composePadded(visibleLines)
```

`composePadded`'s `reserveScrollbarSpace` branch iterates over whatever length `lines` has — it does not pad to `m.height`. When `reserveScrollbarSpace` is `true` but `NeedsScrollbar()` is `false` (content is shorter than the viewport), `ViewWithPaddedLines` passes the unpadded, shorter slice to `composePadded`, which renders fewer rows than the viewport height. This breaks the fixed-height contract the callers rely on and can cause the layout below the message list to shift upward.

The parallel `viewWithLines` method does not have this gap — it pads in both scrollbar and non-scrollbar cases identically. `ViewWithPaddedLines` should mirror that behavior:

```suggestion
	if (m.NeedsScrollbar() || m.reserveScrollbarSpace) && len(visibleLines) < m.height {
		result := make([]string, m.height)
		copy(result, visibleLines)
		visibleLines = result
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

aheritier (approved):

PR #4053, opened 2026-08-24
test(hooks): skip the stdin-JSON hook test when jq is missing

### LINKED ISSUE #4052: pkg/hooks tests fail with a misleading assertion when jq is not installed
### Description

`pkg/hooks` has one test helper that needs a JSON parser in the shell:

```go
func printStdinJSONFieldCmd(field string) string {
	return `cat | jq -r '.` + field + `'`
}
```

`jq` is not part of the Go toolchain and is absent from stock container images
(`golang:1.27` does not ship it). When it is missing the hook still runs and exits cleanly — it
just produces no output — so the test does not fail on "command not found". It fails later, on
its content assertion, with an error that points at the hook plumbing instead of at the machine.

### Expected Behavior

A missing external tool is reported as a missing external tool: the test skips with a reason
naming `jq`, and the package still reports `ok`.

### Actual Behavior

`TestExecuteStopReceivesResponseContent` fails with:

```
"" does not contain "final answer content"
```

Nothing in that message mentions `jq`. It reads as "the stop hook is not receiving the response
content", i.e. a product defect in hook input plumbing.

### Steps to Reproduce

1. Use a machine or container without `jq` (e.g. the stock `golang:1.27` image).
2. `go test ./pkg/hooks/`
3. Observe the failure above.

### Docker Agent version

_No response_

### OS & terminal

_No response_

### Model used

_No response_

### Error output

```shell
=== RUN   TestExecuteStopReceivesResponseContent
    hooks_test.go:586:
        	Error Trace:	pkg/hooks/hooks_test.go:586
        	Error:      	"" does not contain "final answer content"
        	Test:       	TestExecuteStopReceivesResponseContent
--- FAIL: TestExecuteStopReceivesResponseContent (0.00s)
FAIL	github.com/docker/docker-agent/pkg/hooks	20.059s


Confirmed as environmental, not a regression — it reproduces on unmodified `main`
(`f2807d1bb`), 3 runs out of 3, and disappears the moment `jq` is installed.
```

### Screenshots

_No response_

### Additional context

The cost of this is diagnostic time spent in the wrong place. Found while verifying an unrelated
branch: the failure was initially read as a merge regression, and it took a baseline run against
untouched `main` plus tracing the helper to identify a missing package as the cause.

It also makes the package quietly hostile to contributors whose machines lack `jq` — the first
run of the suite reports what looks like a real bug.

Only one test uses the helper, so the blast radius is small and the fix is contained.

dwin-gharibi:
`printStdinJSONFieldCmd` shells out to `jq`. When `jq` is absent the hook still runs and exits
cleanly, producing no output, so the test fails on its *content* assertion rather than on a
missing command — reporting a hook-plumbing defect that does not exist.

Closes #4052.

## Before

On a machine without `jq` (e.g. stock `golang:1.27`):

```text
--- FAIL: TestExecuteStopReceivesResponseContent (0.00s)
    hooks_test.go:586:
        	Error: "" does not contain "final answer content"
FAIL	github.com/docker/docker-agent/pkg/hooks
```

## After

```text
    hooks_test.go:573: jq is not installed; it is required to read a JSON field from hook stdin
--- SKIP: TestExecuteStopReceivesResponseContent (0.00s)
ok  	github.com/docker/docker-agent/pkg/hooks
```

With `jq` present, unchanged:

```text
--- PASS: TestExecuteStopReceivesResponseContent (0.00s)
ok  	github.com/docker/docker-agent/pkg/hooks
```

## What changed

The helper takes `*testing.T` and guards on `exec.LookPath`:

```go
func printStdinJSONFieldCmd(t *testing.T, field string) string {
	t.Helper()
	if _, err := exec.LookPath("jq"); err != nil {
		t.Skip("jq is not installed; it is required to read a JSON field from hook stdin")
	}
	return `cat | jq -r '.` + field + `'`
}
```

The Windows mirror takes the same parameter for signature parity but needs no guard —
PowerShell's `ConvertFrom-Json` is built in. One caller updated.

## Why skip rather than drop the dependency

Removing `jq` would mean either parsing JSON with `sed`/`awk` — fragile, and a fragile test is
worse than a skipped one — or weakening the assertion to "the value appears somewhere in the
payload", which stops testing that the value arrives in the *named field*.

Skipping keeps the assertion exact wherever `jq` exists (CI, most dev machines) and makes the
failure mode honest everywhere else. Coverage is unchanged on any machine that had `jq`.

## Verification

Tested both paths in a container, by moving `/usr/bin/jq` aside and back:

| | `jq` present | `jq` absent |
|---|---|---|
| before | `ok` | `FAIL` — misleading assertion |
| after | `ok` (PASS) | `ok` (SKIP, names jq) |

Full `pkg/hooks` package passes in both states; `go vet` clean.

## Note

This is a test-only change. No production code is touched, and behaviour on machines with `jq`
installed is byte-for-byte identical.

### DIFF
--- a/pkg/hooks/helpers_other_test.go
+++ b/pkg/hooks/helpers_other_test.go
@@ -3,6 +3,7 @@
 package hooks
 
 import (
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -31,7 +32,20 @@ func emitContextEnvPwdCmd(envVars ...string) string {
 
 // printStdinJSONFieldCmd returns a command printing one field of the JSON
 // document the hook receives on stdin.
-func printStdinJSONFieldCmd(field string) string {
+//
+// This is the only test helper that needs a JSON parser in the shell, and POSIX
+// has no built-in one (the Windows mirror can lean on PowerShell's
+// ConvertFrom-Json). It shells out to jq and skips when jq is absent.
+//
+// The skip matters: without it the hook still runs, produces no output, and the
+// test fails on its content assertion instead -- reporting `"" does not contain
+// "final answer content"`, which reads as a defect in the hook plumbing rather
+// than a missing tool on the machine.
+func printStdinJSONFieldCmd(t *testing.T, field string) string {
+	t.Helper()
+	if _, err := exec.LookPath("jq"); err != nil {
+		t.Skip("jq is not installed; it is required to read a JSON field from hook stdin")
+	}
 	return `cat | jq -r '.` + field + `'`
 }
 
--- a/pkg/hooks/helpers_windows_test.go
+++ b/pkg/hooks/helpers_windows_test.go
@@ -31,7 +31,11 @@ func emitContextEnvPwdCmd(envVars ...string) string {
 
 // printStdinJSONFieldCmd returns a command printing one field of the JSON
 // document the hook receives on stdin.
-func printStdinJSONFieldCmd(field string) string {
+// Signature mirrors the POSIX helper, which needs *testing.T to skip when jq is
+// missing. PowerShell's ConvertFrom-Json is built in, so there is nothing to
+// skip on here.
+func printStdinJSONFieldCmd(t *testing.T, field string) string {
+	t.Helper()
 	return `([Console]::In.ReadToEnd() | ConvertFrom-Json).` + field
 }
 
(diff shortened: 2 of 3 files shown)


Sayt-0 (approved):
this one is good

dgageot (approved):

Issue #4060, opened 2026-08-27
RAG semantic-embeddings indexing can trigger a self-sustaining rate-limit retry storm


steilerDev:
## Summary

An agent configured with a `rag` `semantic-embeddings` strategy over a knowledge base large enough that per-chunk LLM enrichment can't finish within the toolset-start timeout will fall into a self-sustaining retry storm against the configured model provider, because several independent pieces of behavior compound:

1. Toolset start (including RAG indexing) is bounded by a hardcoded, non-configurable timeout (`DefaultStartTimeout = 30 * time.Second`, [`pkg/tools/startable.go:381`](https://github.com/docker/docker-agent/blob/main/pkg/tools/startable.go#L381)), shared between the cold-start probe and every subsequent restart attempt. There's no way to grant a RAG-heavy toolset a larger start budget.
2. A failed/aborted toolset start leaves `s.started == false` with no backoff or cooldown before the next attempt (`startLocked`, [`pkg/tools/startable.go:424`](https://github.com/docker/docker-agent/blob/main/pkg/tools/startable.go#L424)).
3. `Agent.Tools()` calls `ensureToolSetsAreStarted` on **every agentic loop iteration**, not once per user turn — confirmed by the loop's own comments that `getTools()` is "the single authoritative source ... each iteration" ([`pkg/runtime/loop.go:1634`](https://github.com/docker/docker-agent/blob/main/pkg/runtime/loop.go#L1634), see also L918-922 and the call sites at L382/L486/L1643). So a stuck/failing RAG toolset gets a fresh start attempt on every loop step of the current turn, with zero delay between attempts.
4. Per-file indexing progress is only persisted after a file's embeddings fully succeed (`indexFile`, [`pkg/rag/strategy/vector_store.go:646-658`](https://github.com/docker/docker-agent/blob/main/pkg/rag/strategy/vector_store.go#L646-L658)). Any abort mid-file (timeout, canceled context, or a classified non-retryable model error) discards all progress for that file, so the next attempt reprocesses the same chunks from scratch.
5. A 429 from the model provider is correctly classified as non-retryable for the *current* request (avoiding hammering a single failing call — see `classifyModelCallError`/`errIndexingAborted` in `pkg/rag/strategy/indexing_errors.go` and `isRetryableStatusCode` in `pkg/modelerrors/modelerrors.go`), but nothing throttles the *next full restart* triggered by (2)+(3). The circuit breaker resets instantly instead of backing off.

[...]

PR #4062, opened 2026-08-27
fix(#4060): add bounded jittered backoff to StartableToolSet retry path

aheritier:

Refs #4060 (partial — MCP/LSP startup errors and RAG 5xx/408 pacing deferred to follow-up)

## What

Adds bounded, cancellable, jittered exponential backoff to `StartableToolSet`'s non-blocking start path. A rate-limited embedding-provider 429 that previously triggered a full concurrent re-index on every agent turn is now paced to at most one retry per 15 s – 5 min window.

## Design

**Gate location — `tryStartLocked` (TryStart/TryStartWithTimeout only).** Blocking `Start()` bypasses the gate so mcpcatalog enable and skill sub-session startup always get a fresh attempt.

**Classifier — `startBackoffRetryable`.** Requires a `*modelerrors.StatusError` in the error chain via `errors.As`. Port numbers and chunk-progress counters in plain error strings cannot arm the gate.

**Bounds — base 15 s, cap 5 min, additive jitter `[d, 1.2d]`.** The additive floor guarantees the full nominal wait; jitter de-synchronises concurrent toolset sources.

**Retry-After.** When the 429 response carries a `Retry-After` header, that hint overrides the computed delay (capped at 5 min, with the same additive jitter applied to avoid re-synchronisation).

**Embedding-provider gap fixed.** `openai/client.go` and `dmr/embed.go` now call `oaistream.WrapOpenAIError` on their embedding errors. `WrapOpenAIError` wraps `*openaisdk.Error` in `*modelerrors.StatusError` carrying the HTTP status code, so a 429 from the embedding provider reaches the gate as `*StatusError` and arms it.

**Known limitation.** Code-mode composites (`codemode.Wrap`) return `PartialStartError` on a partial failure; the partial-start branch resets the gate and leaves the failed-subset retry unpaced. Tracked in #4067.

## Changed files

| File | What changed |
|---|---|
| `pkg/modelerrors/modelerrors.go` | New exported `RetryableHTTPStatus(err)` classifier |
| `pkg/modelerrors/modelerrors_test.go` | Tests for the classifier (incl. regex false-positive cases) |
| `pkg/tools/startable_backoff.go` | New: constants, `computeStartBackoff`, `startBackoffRetryable`, `additiveJitter`, `retryAfterHint` |
| `pkg/tools/startable_backoff_test.go` | Unit tests: gate fires on 429/408/5xx, not on plain text; Retry-After honoured/ignored/capped; jitter spread |
| `pkg/tools/startable_backoff_regression_test.go` | Consumer-shaped regression suite (RAG/MCP/LSP shapes, no-leak, latch, blocking Start ungated) |
| `pkg/tools/startable.go` | `tryStartLocked`, `StartableOption`, `WithStartRetryJitter`, `WithStartRetryClock`, variadic `NewStartable` |
| `pkg/tools/export_test.go` | Exported helpers for deterministic test control |
| `pkg/tools/builtin/rag/rag_backoff_test.go` | Real-toolset integration test via `rag.New` + fake clock |
| `pkg/model/provider/openai/client.go` | Wrap batch-embedding error in `oaistream.WrapOpenAIError` |
| `pkg/model/provider/openai/embed_test.go` | httptest 429 → proves `*StatusError` chain; 5xx equivalent |
| `pkg/model/provider/dmr/embed.go` | Same — wrap DMR embedding error in `oaistream.WrapOpenAIError` |
| `pkg/model/provider/dmr/embed_test.go` | httptest 429 → proves `*StatusError` chain |
| `pkg/agent/agent.go` | Update stale log messages |
| `pkg/tools/mcp/mcp.go` | Update stale log message |
| `docs/tools/rag/index.md` | New section: trigger table (429-only), retry schedule, Retry-After note, troubleshooting |
| `docs/tools/mcp/index.md` | Lifecycle note: local startup failures are not paced by the gate |
| `docs/tools/lsp/index.md` | Same note for LSP |
| `docs/community/troubleshooting/index.md` | Cross-reference to RAG backoff section |

### DIFF
--- a/docs/community/troubleshooting/index.md
+++ b/docs/community/troubleshooting/index.md
@@ -199,6 +199,8 @@ MCP tools using stdio transport must complete the initialization handshake befor
 
 If a toolset keeps crashing in a tight loop, tune the [`lifecycle`](../../configuration/tools/index.md#toolset-lifecycle) block on the toolset (e.g. raise `backoff.initial`, lower `max_restarts`, or switch to the `best-effort` profile) so a flaky dependency does not amplify into a restart storm.
 
+If a **RAG knowledge base** is failing to index because the embedding provider is rate-limiting requests (HTTP 429), Docker Agent automatically backs off and retries — see [Indexing failures, retries and backoff](../../tools/rag/index.md#indexing-failures-retries-and-backoff) for the retry schedule and the `max_indexing_concurrency` / `max_embedding_concurrency` knobs that control how much concurrent load is generated.
+
 ## Configuration Errors
 
 ### YAML syntax issues
--- a/docs/tools/lsp/index.md
+++ b/docs/tools/lsp/index.md
@@ -197,6 +197,8 @@ Available Capabilities:
 
 LSP toolsets are managed by the same supervisor as MCP toolsets, so a crashed `gopls` (or any other language server) is reconnected automatically with exponential backoff. Use the [`lifecycle`](../../configuration/tools/index.md#toolset-lifecycle) block to tune the policy per toolset — for example, mark `gopls` as `strict` if your CI flow requires it to be available, or use `/toolset-restart gopls` from the TUI to force a reconnect when the server gets stuck.
 
+**Startup failure behaviour:** local LSP server startup failures (missing binary, server-unavailable) fail fast — each turn retries immediately with no artificial delay. The rate-limit backoff gate applies only to model-provider embedding calls (see [Indexing failures, retries and backoff](../rag/index.md#indexing-failures-retries-and-backoff)); it does not apply to LSP server startup.
+
 ```yaml
 toolsets:
   - type: lsp
--- a/docs/tools/mcp/index.md
+++ b/docs/tools/mcp/index.md
@@ -258,6 +258,8 @@ toolsets:
 
 See [Toolset Lifecycle](../../configuration/tools/index.md#toolset-lifecycle) for all profiles and tuning knobs, and [`/toolset-restart`](../../features/tui/index.md) to force a reconnect from the TUI.
 
+**Startup failure behaviour:** local MCP startup failures (missing binary, connection refused, authentication error) fail fast — each turn retries immediately with no artificial delay. The rate-limit backoff gate applies only to model-provider embedding calls (see [Indexing failures, retries and backoff](../rag/index.md#indexing-failures-retries-and-backoff)); it does not apply to MCP server startup.
+
 ## Combined Example
 
 ```yaml
--- a/docs/tools/rag/index.md
+++ b/docs/tools/rag/index.md
@@ -174,6 +174,92 @@ chunking:
 >
 > Currently supports Go (`.go`) files. More languages will be added. Falls back to plain text chunking for unsupported file types.
 
+## Indexing failures, retries and backoff
+
+When a knowledge-base fails to start — because the embedding provider is rate-limiting
+your requests or returning a transient server error — Docker Agent spaces out retry
+attempts with bounded exponential backoff instead of hammering the provider on
+every agent turn.
+
+### What triggers backoff
+
+Backoff applies only to **HTTP 429 rate-limit** responses from the embedding or
+model provider — the one signal that reliably reaches the toolset gate. Other
+errors (5xx, 408) are handled per-file within the indexing run and do not arm
+the gate. These are the current gate triggers:
+
+| Failure kind | Behaviour |
+|---|---|
+| HTTP 429 (rate limit) | Backoff: next attempt delayed |
+| Other failures (5xx, 408, config errors, auth) | Fail fast: retried every turn with no added delay |
+| Context cancellation or agent shutdown | Immediate: no delay |
+
+> [!NOTE]
+> 5xx and 408 errors from the embedding provider are retried per-file and do not
+> propagate to the toolset gate. Only 429 (rate-limit) terminates the indexing run
+> early and surfaces the gate so Docker Agent can pace the next attempt.
+
+### Retry policy and parameters
+
+The backoff is **bounded exponential with additive jitter**:
+
+- **Base delay**: 15 seconds
+- **Maximum delay**: up to ~6 minutes (5-minute cap plus up to 20% additive jitter)
+- **Growth**: doubles after each consecutive retryable failure (15s → 30s → 1m → 2m → 4m → 5m)
+- **Jitter**: each wait is a random value in `[nominal, 1.2×nominal]` (additive 0–20%)
+  so concurrent knowledge-base sources spread their retries and avoid
+  hammering the provider together
+- **Retry-After override**: if the embedding provider responds with a `Retry-After`
+  header, that hint overrides the computed delay (capped at the 5-minute maximum,
+  with the same additive jitter applied to spread concurrent retries)
+
+The gate is a lightweight wall-clock check — it creates no background threads or
+timers. A Stop command or agent shutdown takes effect immediately regardless of
+how much of the backoff window remains.
+
+### Operational impact
+
+**Before**: a rate-limited knowledge base was re-indexed on every agent turn —
+`max_indexing_concurrency × max_embedding_concurrency` concurrent provider calls
+could relaunch within milliseconds, easily tripping rate limits for both the
+knowledge base and the agent's own model calls.
+
+**After**: retries are spaced out and jittered so the provider has room to recover
+before the next attempt. The agent continues working with any other toolsets that
+are not affected.
+
+### What you will see
+
+- Docker Agent logs a single warning when a knowledge base first fails to start.
+  Repeated failures in between are logged at debug level only, so you are not
+  flooded with alerts on every turn. Recovery is intentionally silent — the
+  tool appearing in the agent's tool list is the signal that indexing succeeded.
+- The knowledge-base tool does not appear in the agent's tool list until indexing
+  succeeds. A successful start is silent — the tool is listed and the agent uses it.
+
+### Troubleshooting repeated 429 errors
+
+If you see persistent `429` errors in the logs:
+
+1. **Check provider rate limits.** Your embedding API key may have a low requests-per-minute
+   quota. Upgrading the plan or using a different API key can help.
+2. **Reduce concurrency.** The chunked-embeddings and semantic-embeddings strategies
+   accept `max_indexing_concurrency` (default `3`) and `max_embedding_concurrency`
+   (default `3`) parameters. Lowering these reduces simultaneous requests:
+
+   ```yaml
+   rag:
+     docs:
+       docs: [./knowledge-base]
+       strategies:
+         - type: chunked-embeddings
+           max_indexing_concurrency: 1
+           max_embedding_concurrency: 1
+   ```
+
+3. **Use a model with a higher quota.** Some providers offer higher rate limits on
+   specific embedding model tiers.
+
 ## Debugging RAG
 
 Enable debug logging to see retrieval details:
@@ -218,6 +304,7 @@ Look for log tags: `[RAG Manager]`, `[Chunked-Embeddings Strategy]`, `[BM25 Stra
 | `limit`                     | int    | `5`                 | Max results from this strategy                               |
 | `embedding_batch_size`      | int    | `50`                | Chunks per embedding request                                 |
 | `max_embedding_concurrency` | int    | `3`                 | Max concurrent embedding requests                            |
+| `max_indexing_concurrency`  | int    | `3`                 | Max concurrent file-indexing tasks                           |
 | `chunking.size`             | int    | `1500`              | Chunk size in characters (`4000` when `code_aware` is set)   |
 | `chunking.overlap`          | int    | `75`                | Overlap between chunks in characters                         |
 | `chunking.code_aware`       | bool   | `false`             | AST-based chunking (Go files only)                           |
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -645,10 +645,10 @@ func (a *Agent) ensureToolSetsAreStarted(ctx context.Context) {
 			continue
 		}
 		if toolSet.ShouldReportFailure() {
-			slog.WarnContext(ctx, "Toolset start failed; will retry on next turn", "agent", a.Name(), "toolset", desc, "error", err)
+			slog.WarnContext(ctx, "Toolset start failed; will retry (backoff may apply)", "agent", a.Name(), "toolset", desc, "error", err)
 			a.AddToolWarning(fmt.Sprintf("%s start failed: %v", desc, err))
 		} else {
-			slog.DebugContext(ctx, "Toolset still unavailable; retrying next turn", "agent", a.Name(), "toolset", desc, "error", err)
+			slog.DebugContext(ctx, "Toolset still unavailable; will retry (backoff may apply)", "agent", a.Name(), "toolset", desc, "error", err)
 		}
 	}
 }
--- a/pkg/model/provider/dmr/embed.go
+++ b/pkg/model/provider/dmr/embed.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openai/openai-go/v3"
 
 	"github.com/docker/docker-agent/pkg/model/provider/base"
+	"github.com/docker/docker-agent/pkg/model/provider/oaistream"
 	"github.com/docker/docker-agent/pkg/rag/types"
 )
 
@@ -50,7 +51,7 @@ func (c *Client) CreateBatchEmbedding(ctx context.Context, texts []string) (*bas
 		Model: c.ModelConfig.Model,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("failed to create embeddings: %w", err)
+		return nil, fmt.Errorf("failed to create embeddings: %w", oaistream.WrapOpenAIError(err))
 	}
 
 	if len(response.Data) != len(texts) {
--- a/pkg/model/provider/dmr/embed_test.go
+++ b/pkg/model/provider/dmr/embed_test.go
@@ -1,0 +1,51 @@
+package dmr
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	openaisdk "github.com/openai/openai-go/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/modelerrors"
+)
+
+// TestCreateBatchEmbedding_429SurfacesAsStatusError proves that when the
+// embedding endpoint returns HTTP 429 the error propagates as
+// *modelerrors.StatusError so the StartableToolSet backoff gate can arm.
+func TestCreateBatchEmbedding_429SurfacesAsStatusError(t *testing.T) {
+	t.Parallel()
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusTooManyRequests)
+		_, _ = w.Write([]byte(`{"error":{"message":"rate limit exceeded","type":"requests","code":"rate_limit_exceeded"}}`))
+	}))
+	t.Cleanup(srv.Close)
+
+	cfg := &latest.ModelConfig{
+		Provider: "dmr",
+		Model:    "ai/nomic-embed-text",
+		BaseURL:  srv.URL + "/v1",
+	}
+	client, err := NewClient(t.Context(), cfg, nil)
+	require.NoError(t, err)
+
+	_, err = client.CreateBatchEmbedding(t.Context(), []string{"hello"})
+	require.Error(t, err, "expected error from 429 response")
+
+	// The critical assertion: the OpenAI SDK error must be wrapped in
+	// *modelerrors.StatusError by WrapOpenAIError so the backoff gate can arm.
+	var se *modelerrors.StatusError
+	require.ErrorAs(t, err, &se,
+		"error must contain *modelerrors.StatusError so the backoff gate can arm")
+	assert.Equal(t, http.StatusTooManyRequests, se.StatusCode)
+
+	// Cross-check: the underlying SDK error is also reachable.
+	var apiErr *openaisdk.Error
+	assert.ErrorAs(t, err, &apiErr,
+		"underlying *openaisdk.Error must be reachable through the chain")
+}
--- a/pkg/model/provider/openai/client.go
+++ b/pkg/model/provider/openai/client.go
@@ -1165,7 +1165,7 @@ func (c *Client) CreateBatchEmbedding(ctx context.Context, texts []string) (*bas
 	response, err := client.Embeddings.New(ctx, params)
 	if err != nil {
 		slog.ErrorContext(ctx, "OpenAI batch embedding request failed", "error", err)
-		return nil, fmt.Errorf("failed to create batch embeddings: %w", err)
+		return nil, fmt.Errorf("failed to create batch embeddings: %w", oaistream.WrapOpenAIError(err))
 	}
 
 	if len(response.Data) != len(texts) {
--- a/pkg/model/provider/openai/embed_test.go
+++ b/pkg/model/provider/openai/embed_test.go
@@ -1,0 +1,91 @@
+package openai
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	openaisdk "github.com/openai/openai-go/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/modelerrors"
+)
+
+// TestCreateBatchEmbedding_429SurfacesAsStatusError mirrors the DMR embed test
+// for the OpenAI provider path (pkg/model/provider/openai/client.go:1168).
+// It proves that a 429 from the embedding endpoint propagates as
+// *modelerrors.StatusError so the StartableToolSet backoff gate can arm.
+func TestCreateBatchEmbedding_429SurfacesAsStatusError(t *testing.T) {
+	t.Parallel()
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusTooManyRequests)
+		_, _ = w.Write([]byte(`{"error":{"message":"rate limit exceeded","type":"requests","code":"rate_limit_exceeded"}}`))
+	}))
+	t.Cleanup(srv.Close)
+
+	cfg := &latest.ModelConfig{
+		Provider:     "openai",
+		Model:        "text-embedding-3-small",
+		BaseURL:      srv.URL,
+		ProviderOpts: map[string]any{"api_type": "openai_chatcompletions"}, // custom provider → no auth required
+	}
+	client, err := NewClient(t.Context(), cfg, nil)
+	require.NoError(t, err)
+
+	_, err = client.CreateBatchEmbedding(t.Context(), []string{"hello"})
+	require.Error(t, err, "expected error from 429 response")
+
+	// The critical assertion: the error must contain *modelerrors.StatusError
+	// so the StartableToolSet backoff gate can arm.
+	var se *modelerrors.StatusError
+	require.ErrorAs(t, err, &se,
+		"error must wrap *modelerrors.StatusError so the backoff gate can arm")
+	assert.Equal(t, http.StatusTooManyRequests, se.StatusCode)
+
+	// The underlying SDK error must also be reachable.
+	var apiErr *openaisdk.Error
+	assert.ErrorAs(t, err, &apiErr,
+		"underlying *openaisdk.Error must be reachable through the chain")
+}
+
+// TestCreateBatchEmbedding_5xxSurfacesAsStatusError verifies that 5xx errors
+// from the embedding endpoint also surface as *modelerrors.StatusError — they
+// are wrapped by WrapOpenAIError the same way 429s are. (The indexing strategy
+// classifies them as per-file retryable rather than aborting, but the wrapper
+// is still in place for any caller that needs the typed error.)
+func TestCreateBatchEmbedding_5xxSurfacesAsStatusError(t *testing.T) {
+	t.Parallel()
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusServiceUnavailable)
+		_, _ = w.Write([]byte(`{"error":{"message":"service unavailable","type":"server_error"}}`))
+	}))
+	t.Cleanup(srv.Close)
+
+	cfg := &latest.ModelConfig{
+		Provider:     "openai",
+		Model:        "text-embedding-3-small",
+		BaseURL:      srv.URL,
+		ProviderOpts: map[string]any{"api_type": "openai_chatcompletions"}, // custom provider → no auth required
+	}
+	client, err := NewClient(t.Context(), cfg, nil)
+	require.NoError(t, err)
+
+	_, err = client.CreateBatchEmbedding(t.Context(), []string{"hello"})
+	require.Error(t, err)
+
+	// 5xx is also wrapped by WrapOpenAIError — it surfaces as *StatusError
+	// but is classified retryable-per-file by classifyModelCallError.
+	var se *modelerrors.StatusError
+	require.ErrorAs(t, err, &se)
+	assert.Equal(t, http.StatusServiceUnavailable, se.StatusCode)
+
+	// Verify the chain is intact.
+	var apiErr *openaisdk.Error
+	assert.ErrorAs(t, err, &apiErr)
+}
--- a/pkg/modelerrors/modelerrors.go
+++ b/pkg/modelerrors/modelerrors.go
@@ -781,3 +781,18 @@ func scalarString(v any) string {
 		return fmt.Sprint(v)
 	}
 }
+
+// RetryableHTTPStatus reports whether err contains an HTTP status that warrants
+// backoff (429 rate-limit, 408 request-timeout, or a 5xx server error). It
+// checks for a *StatusError in the chain first; if none is found it falls back
+// to matching \b([45]\d{2})\b in the error message, which can produce false
+// positives for port numbers or similar numeric patterns. Callers that need
+// strict StatusError-only classification should pre-filter with errors.As.
+// No context-error handling is performed.
+func RetryableHTTPStatus(err error) bool {
+	code := extractHTTPStatusCode(err)
+	if code == 0 {
+		return false
+	}
+	return code == http.StatusTooManyRequests || isRetryableStatusCode(code)
+}
--- a/pkg/modelerrors/modelerrors_test.go
+++ b/pkg/modelerrors/modelerrors_test.go
@@ -944,3 +944,57 @@ func TestScalarStringEdgeCases(t *testing.T) {
 		})
 	}
 }
+
+func TestRetryableHTTPStatus(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		err      error
+		expected bool
+	}{
+		// Retryable HTTP codes via *StatusError.
+		{name: "429 rate-limit StatusError", err: &StatusError{StatusCode: 429, Err: errors.New("rate limited")}, expected: true},
+		{name: "408 request timeout StatusError", err: &StatusError{StatusCode: 408, Err: errors.New("timeout")}, expected: true},
+		{name: "500 server error StatusError", err: &StatusError{StatusCode: 500, Err: errors.New("internal server error")}, expected: true},
+		{name: "503 unavailable StatusError", err: &StatusError{StatusCode: 503, Err: errors.New("service unavailable")}, expected: true},
+		{name: "529 overloaded StatusError", err: &StatusError{StatusCode: 529, Err: errors.New("overloaded")}, expected: true},
+		// Non-retryable HTTP codes via *StatusError.
+		{name: "400 bad request StatusError", err: &StatusError{StatusCode: 400, Err: errors.New("bad request")}, expected: false},
+		{name: "401 unauthorized StatusError", err: &StatusError{StatusCode: 401, Err: errors.New("unauthorized")}, expected: false},
+		{name: "403 forbidden StatusError", err: &StatusError{StatusCode: 403, Err: errors.New("forbidden")}, expected: false},
+		{name: "404 not found StatusError", err: &StatusError{StatusCode: 404, Err: errors.New("not found")}, expected: false},
+		// Plain string errors without HTTP codes: must stay non-retryable so
+		// MCP/LSP "connection refused" errors are not paced.
+		{name: "connection refused", err: errors.New("connection refused: dial tcp 127.0.0.1:9999"), expected: false},
+		{name: "no such host", err: errors.New("no such host: example.invalid"), expected: false},
+		// Plain text that DOES contain a retryable HTTP code: the regex fallback
+		// in extractHTTPStatusCode finds it. NOTE: RetryableHTTPStatus returns
+		// true here, but startBackoffRetryable (the toolset gate classifier)
+		// requires a *StatusError and returns false for the same input — the
+		// narrowing is deliberate to prevent port-number/chunk-count false positives.
+		{name: "503 in plain text", err: errors.New("upstream: 503 Service Unavailable"), expected: true},
+		{name: "429 in plain text", err: errors.New("provider said: 429 Too Many Requests"), expected: true},
+		// HTTP-status precedence: a StatusError{429} wrapped alongside
+		// context.DeadlineExceeded must return true — the HTTP signal wins.
+		{
+			name: "429 StatusError + DeadlineExceeded in chain",
+			err: fmt.Errorf("start budget exceeded: %w, provider said: %w",
+				context.DeadlineExceeded,
+				&StatusError{StatusCode: 429, Err: errors.New("rate limited")}),
+			expected: true,
+		},
+		// Bare context errors: no HTTP code, must return false.
+		{name: "bare DeadlineExceeded", err: context.DeadlineExceeded, expected: false},
+		{name: "bare Canceled", err: context.Canceled, expected: false},
+		{name: "nil error", err: nil, expected: false},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			got := RetryableHTTPStatus(tc.err)
+			assert.Equal(t, tc.expected, got)
+		})
+	}
+}
--- a/pkg/tools/builtin/rag/rag_backoff_test.go
+++ b/pkg/tools/builtin/rag/rag_backoff_test.go
@@ -1,0 +1,129 @@
+package rag
+
+// Regression tests proving that the real RAG ToolSet, when wrapped in
+// tools.NewStartable, correctly engages the backoff gate for structured
+// HTTP-status failures (e.g. a 429 rate-limit from the embedding provider)
+// and does NOT throttle plain (non-StatusError) failures.
+//
+// Gate enforcement is in tryStartLocked (TryStart paths only); all gate
+// assertions here use TryStart, not Start. A frozen fake clock supplied via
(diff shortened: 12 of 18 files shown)


aheritier (commented):
Reviewed at head `5974d12`. CI is green (11 checks pass, rest skipped; no legacy statuses). I also re-ran locally on the PR head: `go build ./...`, `task lint` (0 issues), `go test ./pkg/tools/...`, and `go test -race -count=1 ./pkg/tools/ ./pkg/agent/... ./pkg/runtime/...` — all clean. Linked context (#4060 + its remediation comment) is fully accessible; there are no Jira references, so no external context is missing.

What's good: the gate is wall-clock only — no timers, no goroutines, no sleeping under the lifecycle lock — state lives under the existing `mu`, `lifecycleMutex`/`unlock`/stop-reaping are untouched, the partial-start latch is preserved, context cancellation is excluded, the doubling loop is overflow-safe, and the failure-streak semantics are correctly left alone (the gated early-return happens before `startStreak.fail()`, so warnings aren't re-queued). That's the right shape for this seam.

However, I don't think this delivers PR1's contract yet. Findings below are evidence-backed; I reproduced the error chains with a throwaway test that replays RAG's exact wrapping (`classifyModelCallError` → `vector_store` → `Manager.Initialize` → `rag.ToolSet.Start`).

### [blocking] The fix is not generic — only model-provider errors are ever paced

`startBackoffRetryable` hard-requires `*modelerrors.StatusError`. That type is produced in exactly five places, all model-provider adapters:

```
pkg/model/provider/{gemini,bedrock,oaistream,anthropic}/wrap.go → modelerrors.WrapHTTPError(...)
```

`grep -rln modelerrors pkg/tools/` matches only the two new files in this PR. So MCP, LSP and A2A start failures — including a real `503` — never arm the gate. Confirmed: an MCP-shaped `503 Service Unavailable` error does not arm it.

This contradicts the requirement in #4060's remediation comment ("Ensure the implementation is generic and reusable by RAG, MCP, LSP, and other startable tool sets") and the plan's explicit rejection of "changing only RAG (would leave the same generic amplification in MCP/LSP/A2A)". The code is generic in *location* but RAG-only in *effect*, which is the outcome PR1 was scoped to avoid.

### [blocking] The issue's headline trigger — indexing exceeding the start budget — is unpaced

#4060's mechanism (1)+(2) is a start aborted by the 30s budget. That returns `context.DeadlineExceeded`, which `startBackoffRetryable` excludes up front, so the gate never arms for the timeout-driven phase — the very phase that *generates* the 429s in the causal chain.

Worse, the exclusion short-circuits **before** `StatusError` extraction, so a 429 that races the deadline also disarms the gate:

```go
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
    return false   // wins even when a 429 StatusError is also in the chain
}
```

Verified: a chain carrying both a 429 `StatusError` and `DeadlineExceeded` does **not** arm the gate. This is reachable in production — `Manager.Initialize` races `case <-ctx.Done(): return ctx.Err()` against its results channel, so under load the deadline can mask the rate-limit signal precisely when pacing matters most. Suggest narrowing the exclusion to *caller* cancellation and still honouring a `StatusError` found elsewhere in the chain.

### [blocking] Backoff bounds are ~10–20× weaker than the authorized PR1 defaults, with no stated rationale

The plan authorizes 15s initial, ×2, **5min** base cap, additive-only 0–20% jitter (effective max 6min). This PR ships base **1s**, cap **30s**, equal jitter `[nominal/2, nominal]`. Measured sequence:

```
attempt 1 → ~0.9s     attempt 4 → ~5.0s
attempt 2 → ~1.6s     attempt 5 → ~13.0s
attempt 3 → ~3.5s     attempt 6 → ~28.5s   (cap; ~24–52s cumulative)
```

Steady state is therefore **one full concurrent indexing burst every 15–30s, indefinitely** while the KB stays unindexed — each burst being `max_indexing_concurrency × max_embedding_concurrency` concurrent LLM calls. Against an incident that saturated a shared gateway and failed the agent's own conversational turns, a 30s ceiling still permits substantial sustained load; the authorized 5min cap yields one burst per 5–6 min. The first retry at ~0.5–1s is effectively no pacing at all.

If these bounds were changed deliberately, please state the evidence and record the deviation — right now neither the PR body nor the commit message mentions that the authorized defaults were changed.

### [blocking] Gating inside `startLocked` was an explicitly rejected alternative, and it does gate explicit starts

The plan enforces the schedule **only** in the non-blocking `TryStart`/`TryStartWithTimeout`, and rejects gating `startLocked` because it "would delay explicit starts and break mcpcatalog enable". The gate is in `startLocked`, so blocking `Start` is gated too — and blocking `Start` is what user/model-driven paths use:

- `pkg/tools/builtin/mcpcatalog/mcpcatalog.go:168` — `startToolset` defaults to `ts.Start(ctx)`, reached from `handleEnable` (`:765`), whose whole contract is a same-turn deterministic answer.
- `pkg/runtime/loop.go` skill sub-session startup — `startable.Start(ctx)`.

This is currently *latent* only because the classifier is so narrow that MCP errors never arm the gate — i.e. two defects are masking each other. The moment the classifier is broadened (Follow-up A, or fixing the finding above), explicit enable silently starts returning a stale cached error without attempting a connect. Please move enforcement to the `TryStart` paths as specified.

### [should-fix] Stale window survives a successful `/toolset-restart`

`LocalRuntime.RestartToolset` dispatches to the **inner** `Restartable` (`pkg/runtime/runtime.go:1156-1157`); the wrapper never sees it. After a total failure the wrapper has `started == false`, so `startLocked` skips the `if s.started { ... reporter.IsStarted() ... }` recovery block entirely and hits the gate, returning the retained error **without ever consulting the now-live `StartReporter`**. Tools stay unlisted for up to 30s after a successful manual recovery. The plan requires this explicitly: "a later `TryStart` observation that a latched `StartReporter` is live clears failure count/schedule … makes direct `/toolset-restart` recovery immediately authoritative rather than leaving a stale window." Checking a live reporter before the gate would close it.

### [should-fix] No test seam for PR2's mandated cross-package tests

`ExportedSetJitter` lives in `export_test.go`, so it is visible only to package `tools`' own tests — and there is no clock seam at all. PR2 requires deterministic RAG/MCP/LSP regression tests in *other* packages driving a "fake retry clock"; those packages can neither control jitter nor advance policy time. The plan specified exported constructor options (`WithStartRetryClock`, `WithStartRetryJitter` via `NewStartable(ts, opts...)`) precisely for this. Adding them now avoids reworking this file in PR2.

### [should-fix] A test's doc comment claims coverage it doesn't have

`TestStartableToolSet_BackoffNoDoubleStartWithinWindow` states it holds "even under concurrent callers", but the test is strictly sequential — four in-order `s.Start` calls, no goroutines. Please either make it concurrent or drop the claim. Relatedly, PR1's gate list requires a concurrent **exact-boundary** test proving many simultaneous `TryStart` calls invoke the inner `Start` exactly once, plus an at-boundary test; neither is present.

### [optional] Stale test doc name

The comment block above `TestStartableToolSet_BackoffDormantForPlainErrors` opens with `TestStartableToolSet_RetryableBackoffWithExistingFlappyTests` — leftover from a rename.

### Scope note

PR1 as planned also includes sentinel handling in `pkg/agent/agent.go` and `pkg/runtime/runtime.go` plus focused `agent_test.go`/`runtime_test.go` partial-composite cases, described as merge gates for the compatibility contract. This PR returns the raw retained error instead of a scheduled-outcome sentinel, so those changes are absent. That's self-consistent, but it means the planned compatibility gates aren't demonstrated. Worth confirming that's an intentional simplification rather than an omission.

Happy to re-review promptly once the classifier reach, the deadline/429 precedence, and the enforcement point are addressed. No schema, config or docs changes are needed for PR1, and I agree with deferring `DefaultStartTimeout`.

aheritier (commented):
Re-reviewed at head `072cc4a` (delta `5974d12..072cc4a`, +415/−163 across 6 files). Thanks — this is a big step forward: **all four blockers and all three should-fixes from my previous review are addressed.** CI is green at this head (10 checks success, 4 skipped, no legacy statuses); `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` only for the missing required approving review. I re-validated locally on the head: `go build ./...`, `golangci-lint run ./pkg/tools/... ./pkg/modelerrors/...` (0 issues), and `go test -race -count=1 ./pkg/tools/... ./pkg/agent/... ./pkg/runtime/...` — all clean. Linked context (#4060 + its remediation comment, unchanged since 08:36) is fully accessible; no Jira references, so no external context is missing.

## Prior findings — resolved

- **[was blocking] classifier not generic** → fixed. `modelerrors.RetryableHTTPStatus` (`pkg/modelerrors/modelerrors.go:802-808`) routes through `extractHTTPStatusCode`, so any HTTP-status-bearing start failure arms the gate regardless of Go type. MCP/LSP/A2A/RAG are now paced consistently, with table coverage in `TestRetryableHTTPStatus`.
- **[was blocking] deadline short-circuited the 429 signal** → fixed. The context pre-check is gone and precedence is now explicit and tested: `429 StatusError + DeadlineExceeded in chain → true`, bare context errors → `false`. (See the residual scope note below.)
- **[was blocking] backoff bounds 10–20× weaker than authorized** → fixed exactly to plan: `startBackoffBase = 15s`, `startBackoffMax = 5m` (`pkg/tools/startable_backoff.go:18-19`), additive jitter `[nominal, 1.2×nominal]` (`:82-86`). Sequence 15s/30s/60s/120s/240s/300s, effective max 6min, pinned by `TestStartBackoffExponentialGrowth` and `TestStartBackoffBoundedAndJittered`.
- **[was blocking] gate in `startLocked` delayed explicit starts** → fixed. Enforcement moved to `tryStartLocked` (`pkg/tools/startable.go:503-522`); blocking `Start` calls `startLocked` directly (`:390-394`), so mcpcatalog enable and skill sub-session startup are never delayed. (One coverage gap, below.)
- **[was should-fix] stale window after a successful `/toolset-restart`** → fixed by the live-reporter adoption in `tryStartLocked` (`:509-517`), covered by `TestStartableToolSet_ExternalRecoveryClearsBackoffGate` (`startable_backoff_test.go:492`).
- **[was should-fix] no test seam for PR2** → fixed. `NewStartable(ts, opts...)` plus exported `WithStartRetryClock` / `WithStartRetryJitter` (`:322`, `:329`) give other packages a fake clock and deterministic jitter.
- **[was should-fix] doc comment claimed concurrency it didn't have** → fixed: `TestStartableToolSet_BackoffNoDoubleStartWithinWindow` (`:231`) now really runs 20 goroutines and asserts exactly one inner `Start`, and `TestStartableToolSet_BackoffAtBoundary` (`:256`) adds the 1ns-before / exactly-at boundary contract.
- **[was optional] stale test doc name** → fixed.

## [blocking] Partial-start clears the gate, which leaves the storm unpaced for code-mode agents

`startLocked` resets all backoff state on a `PartialStartError` (`pkg/tools/startable.go:601-604`), with the rationale that "the failed subset is retried via the composite's own recovery mechanism, not the cold-start gate". That recovery mechanism *is* the unpaced per-turn retry:

- `codemode.Wrap(toolSets...)` collapses **every** toolset into one composite when `code_mode` is set on the agent or globally (`pkg/teamloader/teamloader.go:859-860`), and that composite is what gets wrapped by `NewStartable` (`:932`).
- `codeModeTool.Start` retries each failed inner on every call and returns `PartialStartError` whenever at least one inner is healthy (`pkg/tools/codemode/codemode.go:222-259`).

So for a code-mode agent whose RAG toolset 429s while other toolsets are fine: turn N arms nothing (partial → reset), turn N+1 re-enters `codeModeTool.Start` → RAG `Manager.Initialize` → a fresh full-speed indexing burst. That is exactly #4060, unfixed, in a supported configuration — and it is the one path where the current comment tells a future reader it's handled. `TestStartableToolSet_PartialStartClearsBackoffGate` (`:392`) pins the exemption as intended behaviour, so nothing flags it.

The plan's contract avoids this: pace the failed members *while preserving the latch* — "a latched partial composite remains usable while its failed members are paced" — via the `ErrStartRetryScheduled`/`RetryAt` sentinel returned as `(s.started, scheduledErr)`, with the agent/runtime sentinel handling and partial-composite tests listed as PR1 merge gates. Either of these resolutions works for me:

1. Arm the gate when a partial cause is retryable, keep `s.started = true`, and return the scheduled sentinel so `pkg/agent/agent.go` and `pkg/runtime/runtime.go` still list the healthy subset (plus the two focused partial tests the plan calls merge gates); or
2. Keep the exemption deliberately, but say so in the comment (naming the code-mode composite as the known unpaced path), note it in the PR body, and open the follow-up issue — rather than leaving a comment that implies coverage.

## [should-fix] The regex fallback arms the gate on errors that carry no HTTP status, and the doc comment says it can't

`RetryableHTTPStatus`'s doc claims it "performs NO string-pattern heuristics, so plain network errors such as 'connection refused' or 'no such host' always return false". `extractHTTPStatusCode` does fall back to `statusCodeRegex = \b([45]\d{2})\b` over the whole message (`pkg/modelerrors/modelerrors.go:289-319`), which matches any bare 4xx/5xx-looking number. Reproduced against that exact regex:

```
match=true   Post "http://localhost:503/mcp": dial tcp 127.0.0.1:503: connect: connection refused
match=true   rag: indexing aborted: chunk 429 of 812 failed
match=true   mcp: server exited: signal: killed (pid 504)
match=false  dial tcp 10.0.0.5:8080: i/o timeout
match=false  context deadline exceeded after 500ms
```

The regex itself is pre-existing, but this PR widens its input domain from provider HTTP response bodies to *arbitrary* toolset start errors (MCP stderr, LSP spawn failures, RAG progress text). A local MCP server on port 503 that is simply down now gets paced 15s → 5min instead of retrying each turn — the opposite of the stated design intent for local process failures. `TestRetryableHTTPStatus`'s "connection refused" case passes only because of the port it happens to use, so it generalises a property that doesn't hold.

Minimum: correct the doc comment and add the false-positive cases to the table so the behaviour is at least honest. Better: require an HTTP-ish context for the fallback (adjacent status text or an `HTTP`/`status` token), or accept only `*StatusError` for the toolset classifier and treat the regex path as model-provider-only.

## [should-fix] Nothing pins the blocking-`Start` bypass — the fix for the last blocker is untested

`startable_backoff_test.go` never calls `s.Start(...)` (only `s.Stop` at `:126` and `:215`); every case drives `TryStart`. The invariant that blocking `Start` must skip the gate is exactly what makes mcpcatalog enable and skill sub-session startup deterministic in-turn, and it is now a one-line-refactor away from silent regression (route `Start` through `tryStartLocked` and all tests still pass). Please add: arm the gate with a retryable failure, then assert a blocking `Start` reaches the inner (`inner.starts` increments) while a concurrent-window `TryStart` does not.

## [should-fix] Live-reporter adoption is unconditional, changing non-gated `TryStart` semantics

The adoption block in `tryStartLocked` (`pkg/tools/startable.go:509-517`) runs whenever `!s.started` and the inner `StartReporter` is live — **not** only when a window is armed. Previously that state fell through to `startable.Start(ctx)`; now the wrapper latches `started = true` and returns `nil` without ever invoking the inner `Start`, for every `StartReporter` toolset (MCP after a background supervisor reconnect, LSP after a supervisor retry), whether or not backoff is involved. That's a broader behavioural change than the stale-window fix needs. Guarding it with `!s.startBackoffUntil.IsZero()` keeps the fix scoped to the case it was written for.

## [should-fix] Stale "retry on next turn" messages the plan asked to correct

Retries can now be up to ~6 minutes apart, but these still promise next-turn behaviour:

- `pkg/agent/agent.go:648` — `"Toolset start failed; will retry on next turn"`
- `pkg/agent/agent.go:651` — `"Toolset still unavailable; retrying next turn"`
- `pkg/tools/mcp/mcp.go:99` — `"MCP command not yet available, will retry on next turn"`

Log-only (the TUI warning string is unaffected), but the plan's PR1 step explicitly includes grepping and correcting these plus the contract comments in `startable.go` / `agent.go` / `runtime.go`.

## [optional] `ExportedSetJitter` is now dead code

`ExportedSetJitter` (`pkg/tools/export_test.go`) has no remaining callers — every test uses `WithStartRetryJitter`. Since it also carries the "writes `startJitter` without `mu`" caveat, dropping it removes both the dead code and the caveat.

## Scope notes (no action required in this PR)

- **Bare-deadline starts remain unpaced, and that is plan-conformant** — acceptance gate 2 says a timeout alone must not schedule. Worth knowing precisely what that leaves behind: `Manager.Initialize` returns bare `ctx.Err()` on deadline (`pkg/rag/manager.go:227-229`), with no HTTP code, so a large corpus that always exceeds the 30s budget still starts one fresh indexing burst per turn. The justification in `pkg/tools/startable_backoff.go:36-39` ("already naturally paced to ~budget intervals while it holds the single-flight lock") doesn't hold here: RAG honours the context and returns at the deadline, releasing the single-flight lock immediately. Please soften that comment and let the deferred `DefaultStartTimeout` follow-up own the gap explicitly.
- The `ErrStartRetryScheduled` / `RetryAt` sentinel and the focused `agent_test.go` / `runtime_test.go` partial-composite cases are still absent. That's self-consistent with returning the raw retained error, and it's benign today (a gated attempt never reaches `startStreak.fail()`, so `ShouldReportFailure` stays false and users don't get a warning per turn) — but it is coupled to the blocking finding above: implementing resolution 1 requires the sentinel.

Overall the core 429/5xx pacing path now looks right, and the bounds, gate placement, test seams and boundary/concurrency coverage all match the plan. Filing this as a comment rather than a formal request-for-changes only because I'm the PR author and GitHub won't let me review my own PR; please treat the blocking finding as a merge gate. Happy to re-review promptly once the partial-composite path is either paced or explicitly scoped.

aheritier (commented):
Re-reviewed at head `63109400` (force-pushed squash of the previously reviewed `072cc4a`, now a single commit onto `b407a13a`). CI is green at this head: 10 checks success, 3 skipped (post-merge publish jobs), 0 failing, 0 pending; the legacy `commits/{sha}/status` endpoint reports `state=pending` only because `statuses` is empty — no legacy contexts are configured. `mergeable=MERGEABLE` (no conflicts); `mergeStateStatus=BLOCKED` solely on the missing required approving review. Linked context (#4060 + its remediation comment, unchanged, and the new #4067) is fully accessible; there are no Jira references, so no external context is missing.

**All eight findings from my previous review are resolved at this head** — verified individually: bounds are now `startBackoffBase = 15s` / `startBackoffMax = 5m` with additive `[d, 1.2d]` jitter (`pkg/tools/startable_backoff.go:12-13`, `:54-57`); enforcement moved to `tryStartLocked` (`pkg/tools/startable.go:477-501`) with blocking `Start` bypassing it, and `TryStartWithTimeout` correctly routing through `TryStart`; the bypass is now pinned by `TestStartableToolSet_BlockingStartSkipsGate`; live-reporter adoption is guarded by `!s.startBackoffUntil.IsZero()` (`:484`); `RetryableHTTPStatus`' doc comment now describes the regex fallback honestly (`pkg/modelerrors/modelerrors.go:785-791`); the partial-start exemption is documented in code (`pkg/tools/startable.go:573-576`) and tracked as #4067; the stale "retry on next turn" log strings are corrected (`pkg/agent/agent.go:648/651`, `pkg/tools/mcp/mcp.go:99`); `ExportedSetJitter` is gone; and the incorrect "naturally paced to ~budget intervals" comment is removed. Thank you — the gate shape, bounds, placement, test seams and boundary/concurrency coverage now match the plan.

Two blocking items remain.

## [blocking] The PR description contradicts the shipped implementation on the exact points that were previously blocking

The body still describes the abandoned first iteration:

- "bounded exponential backoff with equal jitter (base 1s, cap 30s, [nominal/2, nominal])" — the code ships `startBackoffBase = 15 * time.Second` / `startBackoffMax = 5 * time.Minute` (`pkg/tools/startable_backoff.go:12-13`) with **additive** jitter `d + rand.N(d/5+1)` → `[d, 1.2d]` (`:54-57`). Those are the corrected, plan-authorised values — but the body advertises the 10–20×-too-weak ones that were a merge blocker.
- "wall-clock gate inside `startLocked`" — the gate is in `tryStartLocked` (`pkg/tools/startable.go:477-501`), and `startLocked` carries an explicit comment saying it is *not* there (`:512-513`). Gating `startLocked` was the previously blocking, explicitly rejected placement.
- "Changed files" lists 4; the PR changes 8, omitting the new **exported public API** `modelerrors.RetryableHTTPStatus` (`pkg/modelerrors/modelerrors.go:785-796`), its tests, and the log-string edits in `pkg/agent/agent.go` / `pkg/tools/mcp/mcp.go`.
- The body never mentions the partial-start / code-mode exemption that the code documents (`pkg/tools/startable.go:573-576`) and #4067 tracks — noting it in the PR body was one of the two stated conditions for accepting that exemption.

The code is right; the description is wrong, and it is the durable record a squash-merge carries. A reader comparing it against the prior review would reasonably conclude the blockers were not fixed. This is a body edit only.

## [blocking] The commit message's own stated trigger — a 429 from the *embedding* provider — is not paced

`startBackoffRetryable` requires a `*modelerrors.StatusError` in the chain (`pkg/tools/startable_backoff.go:22-28`). That type is produced in exactly one place, `modelerrors.WrapHTTPError` (`pkg/modelerrors/modelerrors.go:52-60`), whose only callers are the four **chat-completion** adapters: `pkg/model/provider/anthropic/wrap.go:39`, `bedrock/wrap.go:32,34`, `gemini/wrap.go:25`, `oaistream/wrap.go:23`.

The RAG **embedding** call goes through none of them:

```
pkg/model/provider/openai/client.go:1165-1169
    response, err := client.Embeddings.New(ctx, params)
    if err != nil { return nil, fmt.Errorf("failed to create batch embeddings: %w", err) }  // raw SDK error
pkg/model/provider/dmr/embed.go:48                                                           // same, unwrapped
```

and every hop up to `startLocked` preserves the error with `%w`, so the *type* is decisive:

```
pkg/rag/strategy/vector_store.go:612   fmt.Errorf("failed to generate embeddings: %w", classifyModelCallError(err))
pkg/rag/strategy/indexing_errors.go:33 ClassifyModelError → 429 found only via the regex fallback → errIndexingAborted
pkg/rag/strategy/vector_store.go:329   isIndexingAborted → aborts the run
pkg/rag/manager.go:234                 fmt.Errorf("one or more strategies failed to initialize: %w", firstError)
pkg/tools/builtin/rag/rag.go:105-109   fmt.Errorf("failed to initialize RAG manager %q: %w", ...)
→ no *StatusError anywhere in the chain → startBackoffRetryable == false → the gate never arms
```

`TestStartableToolSet_BackoffDormantForPlainErrors` pins this as intended, so nothing flags it. Net effect: an embedding-provider 429 during `semantic-embeddings` indexing still relaunches a full concurrent burst every loop iteration — #4060 unfixed for the embedding half of `max_indexing_concurrency × max_embedding_concurrency` — while the commit message states the fix addresses "a 429 from the embedding provider".

The same narrowness leaves MCP/LSP/A2A unpaced: `pkg/tools/mcp/mcp.go:433` `Restart` returns plain errors, so `setStartBackoff` in the `Restartable` branch (`pkg/tools/startable.go:547`) always *resets* — a remote MCP server genuinely returning `503` is never paced. That is #4060's requirement "Ensure the implementation is generic and reusable by RAG, MCP, LSP, and other startable tool sets" met in location but not in effect.

To be clear, the chat-model half is correctly covered (`pkg/rag/strategy/semantic_embeddings.go:267` → `CreateChatCompletionStream` → `pkg/model/provider/openai/response_stream.go:86` → `WrapOpenAIError`), so this is a coverage gap, not a no-op. Any of these resolves it:

1. Wrap the embedding paths in `WrapHTTPError` (`openai/client.go:1165`, `dmr/embed.go:48`) so the RAG embedding half is structurally classified — smallest change, no classifier loosening.
2. Accept `extractHTTPStatusCode` when the message carries HTTP-ish context (adjacent status text or an `HTTP`/`status` token), avoiding the `:503` port and `chunk 429 of 812` false positives.
3. Keep StatusError-only deliberately, state the gap in the PR body, open a follow-up issue (the #4067 pattern), and add a test pinning the exclusion.

## [should-fix] The commit message misdescribes the new exported function

> "Add modelerrors.RetryableHTTPStatus(err): HTTP-status classifier that recognises 429, 408, and 5xx signals via *StatusError (no regex fallback, so port numbers / chunk counters cannot trigger backoff)"

`RetryableHTTPStatus` (`pkg/modelerrors/modelerrors.go:792-796`) delegates to `extractHTTPStatusCode`, which **does** fall back to `statusCodeRegex = \b([45]\d{2})\b` (`:289`, `:311-317`) — its own doc comment says so, and `TestRetryableHTTPStatus` asserts `"upstream: 503 Service Unavailable"` → `true`. The "no regex fallback" property belongs to `startBackoffRetryable`, not to the exported function. This message lands permanently in `main`'s history describing a public API incorrectly.

## [should-fix] A test comment asserts the opposite of shipped behaviour, and the distinguishing case is untested

`pkg/modelerrors/modelerrors_test.go`, on the "503 in plain text" / "429 in plain text" cases, says the regex fallback "must find it so any toolset (not just those wrapping `*StatusError`) is paced for real server errors." That does not hold — `pkg/tools/startable_backoff.go:22-26` filters the regex path out first. Meanwhile `TestStartableToolSet_BackoffDormantForPlainErrors` uses only `errors.New("boom")`, which carries no status at all, so it cannot distinguish "no status" from "status via string match" — the actual `pkg/tools` contract is unpinned. Please correct the comment and add a `pkg/tools` case asserting `errors.New("upstream: 503 Service Unavailable")` does **not** arm the gate; that documents the deliberate narrowing and guards against a silent loosening later.

## [optional] `RetryableHTTPStatus` is exported with a documented footgun no production caller uses

Its sole caller pre-filters to `*StatusError` (`pkg/tools/startable_backoff.go:23-27`), making the call equivalent to reading `se.StatusCode`. The regex branch is exercised only by tests, while the exported surface invites exactly the false positives its doc warns about. Consider keeping it unexported, or exporting `RetryableStatusCode(code int) bool` instead.

## [optional] A single non-retryable failure resets the exponent

`setStartBackoff` calls `resetStartBackoff()` for any non-retryable error (`pkg/tools/startable.go:465-470`), zeroing `startBackoffAttempt`, so a toolset alternating `503 → 401 → 503` restarts each window at the 15s base instead of escalating. Benign today (401 is terminal in practice) — worth a note in the comment if intentional.

## Checklist against #4060's PR1 contract

Met: retryable-only pacing; bounded exponential backoff with jitter; immediate cancellation/shutdown/deadline handling (no sleeps, timers or goroutines; bare context errors never arm; `Stop` clears); successful-start path and retry limits preserved; terminal errors keep their cause; concurrent starts do not multiply retry activity (20 goroutines → exactly one inner `Start`); non-retryable failures fail promptly; already-started toolsets are not restarted; behaviourally narrow with no config/schema change; build/tests/lint green. Docs are PR2's scope per the plan, and deferring `DefaultStartTimeout` remains correct.

Outstanding: the "generic across RAG/MCP/LSP" requirement (the blocking finding above), plus the PR-body and commit-message accuracy items.

Filing this as a comment rather than a formal request-for-changes only because I authored the PR and GitHub will not accept a self-review verdict; please treat both blocking findings as merge gates. Happy to re-review promptly once the classifier reach is resolved (or explicitly scoped with a follow-up) and the body/commit message match the shipped design.

aheritier (commented):
**Summary:** Solid, well-tested implementation of the `StartableToolSet` backoff gate (bounded exponential + additive jitter, `tryStartLocked` gating only non-blocking `TryStart`, external-recovery adoption, embedding-provider `StatusError` wrapping). CI is green across lint/tests/build/CodeQL. One blocking doc defect and two non-blocking follow-ups below.

**[blocking] Invalid config-schema YAML example in `docs/tools/rag/index.md` ("Reduce concurrency" step under the new "Troubleshooting repeated 429 errors" section):**

```yaml
rag:
  name: docs
  path: ./knowledge-base
  strategy:
    type: chunked-embeddings
    params:
      max_indexing_concurrency: 1
      max_embedding_concurrency: 1
```

This does not validate against `agent-schema.json` / `pkg/config/latest.RAGConfig`:
- `rag:` is a map keyed by the source **name** (`additionalProperties: RAGConfig` in `agent-schema.json`); it has no `name` or `path` fields — document paths go in `docs: []string`.
- `strategy:` (singular object) doesn't exist — the field is `strategies:` (an array, `minItems: 1`).
- `params:` is not a real wrapper key — `RAGStrategyConfig.UnmarshalYAML` (`pkg/config/latest/types.go`) flattens strategy-specific fields (like `max_indexing_concurrency`, `max_embedding_concurrency`) directly as siblings of `type`.

A user copy-pasting this fails schema validation. Every other example in this same file (Quick Start, Hybrid Search) uses the correct shape. Suggested fix:

```yaml
rag:
  docs:
    docs: [./knowledge-base]
    strategies:
      - type: chunked-embeddings
        max_indexing_concurrency: 1
        max_embedding_concurrency: 1
```

**[should-fix] `docs/tools/rag/index.md` Configuration Reference table (Chunked-Embeddings Strategy section) is missing `max_indexing_concurrency`.** The new "Reduce concurrency" prose says *"the chunked-embeddings and semantic-embeddings strategies accept `max_indexing_concurrency`… and `max_embedding_concurrency`… parameters,"* and `pkg/rag/strategy/chunked_embeddings.go:48` confirms chunked-embeddings does read `max_indexing_concurrency` — but the reference table above only lists `max_embedding_concurrency` for that strategy. Worth adding the missing row while touching this file.

**[optional] `docs/community/troubleshooting/index.md` wasn't updated.** The `remediate-rag-retry-storm-4060` plan called for adding generic automatic-backoff/safe-retry guidance there alongside the RAG/MCP/LSP doc updates. Not blocking — the new backoff behavior is documented in the three toolset docs — but a cross-reference from the general troubleshooting page would help readers who land there first after a 429.

aheritier (commented):
Re-reviewed at head `fe986e7` (single commit onto `b407a13a`, +1431/−13 across 17 files). CI is green at this head: 16 check-runs success, 3 skipped (post-merge publish jobs), 0 failing, 0 pending; the legacy `commits/{sha}/status` endpoint reports `state=pending` only because `statuses` is empty — no legacy contexts are configured. `mergeable=MERGEABLE`; `mergeStateStatus=BLOCKED` solely on the missing required approving review. Linked context (#4060 + its remediation comment, and #4067) is fully accessible; there are no Jira references, so no external context is missing. Re-validated locally on the head: `go build ./...`, `golangci-lint run ./pkg/tools/... ./pkg/modelerrors/... ./pkg/model/provider/...` (0 issues), and `go test -race -count=1 ./pkg/tools/ ./pkg/modelerrors/ ./pkg/tools/builtin/rag/ ./pkg/model/provider/dmr/` — all clean.

## Previously-blocking findings — verified resolved

**The embedding-provider 429 now structurally reaches the gate.** I traced every hop and confirmed `%w` is used throughout, so the `*StatusError` type survives to `startBackoffRetryable`:

```
pkg/model/provider/openai/client.go:1168   oaistream.WrapOpenAIError(err)   → *modelerrors.StatusError
pkg/model/provider/dmr/embed.go:54         same
pkg/rag/embed/embed.go:142                 fmt.Errorf("batch %d failed: %w", ...)
pkg/rag/strategy/vector_store.go:612       fmt.Errorf("failed to generate embeddings: %w", classifyModelCallError(err))
pkg/rag/strategy/indexing_errors.go:34     429 → retryable=false → fmt.Errorf("%w: %w", errIndexingAborted, err)
pkg/rag/strategy/vector_store.go:329,358   isIndexingAborted → errgroup returns → g.Wait() propagates
pkg/rag/manager.go:234                     fmt.Errorf("one or more strategies failed to initialize: %w", firstError)
pkg/tools/builtin/rag/rag.go:108           fmt.Errorf("failed to initialize RAG manager %q: %w", ...)
→ *StatusError reachable → startBackoffRetryable == true → gate arms
```

`TestCreateBatchEmbedding_429SurfacesAsStatusError` (`pkg/model/provider/dmr/embed_test.go`) pins the wrap site against a real `httptest` 429, and `TestRAGStartableBackoff_StatusErrorEngagesGate` (`pkg/tools/builtin/rag/rag_backoff_test.go:73`) pins the end-to-end gate through `rag.New` with a fake clock. The core defect in #4060 is now genuinely closed for the embedding half.

**The RAG doc/code mismatch is fixed.** `docs/tools/rag/index.md` now scopes the trigger table to 429 only and adds an explicit NOTE that 5xx/408 are swallowed per-file and never reach the gate — which matches `classifyModelCallError` (`pkg/rag/strategy/indexing_errors.go:27-36`). The MCP/LSP docs are consistent with it.

**The invalid YAML example is fixed.** The "Reduce concurrency" snippet now uses the schema-valid `rag:` → source-name → `docs: []` / `strategies: []` shape with flattened strategy params, and the Configuration Reference table gained the missing `max_indexing_concurrency` row.

Also confirmed good: the gate is wall-clock only (no timers, goroutines, or sleeping under the lifecycle lock), enforcement is in `tryStartLocked` so blocking `Start` is never delayed (pinned by `TestStartableToolSet_BlockingStartSkipsGate`), live-reporter adoption is window-guarded, `Stop` clears the window, the doubling loop is overflow-safe, and gated attempts return before `startStreak.fail()` so users don't get a warning per turn.

## [blocking] The `Retry-After` override branch has zero test coverage

`setStartBackoff` gained a `Retry-After` override:

```go
// pkg/tools/startable.go:471-479
s.startBackoffAttempt++
delay := computeStartBackoff(s.startBackoffAttempt, s.startJitter)
if hint := retryAfterHint(err); hint > delay {
    delay = min(hint, startBackoffMax)   // ← never executed by any test
}
s.startBackoffUntil = s.nowFn().Add(delay)
```

Here is the evidence. `go test -race -count=1 -coverprofile=cov ./pkg/tools/` (86.1% overall) reports the branch body with an execution count of **0**:

```
pkg/tools/startable.go:475.2,475.47       3  23    ← the `if` is reached 23×
pkg/tools/startable.go:476.3,477.1        1   0    ← the override never runs
pkg/tools/startable_backoff.go:66.2,66.10 1   0    ← retryAfterHint's non-StatusError return
```

`grep -n "RetryAfter" pkg/tools/*_test.go` returns nothing — no test in `pkg/tools` constructs a `StatusError` with a non-zero `RetryAfter`, so neither of the two behaviours this branch introduces is pinned. I confirmed both are live with a throwaway probe (429 + `RetryAfter: 120s`, identity jitter, frozen clock via `WithStartRetryClock`):

- advancing 16s (past the 15s computed window) → underlying `Start` **not** called: the hint correctly extends the window;
- advancing exactly 120s → underlying `Start` called: the window equals the hint **exactly**.

New behaviour on the hot path of the very failure mode this PR exists to fix, with no test, is a merge gate. Please add coverage for: hint > computed delay extends the window; hint < computed delay is ignored; a hint above `startBackoffMax` is capped.

## [should-fix] The `Retry-After` path bypasses jitter, re-synchronising exactly what jitter exists to spread

The probe above shows the window equals the hint exactly — a jittered hint would have been ≤ 144s. Because `retryAfterHint` is applied *after* `computeStartBackoff`, the returned delay carries no jitter at all.

That contradicts this PR's own stated rationale ("the jitter de-synchronises concurrent toolset sources", PR body; and `docs/tools/rag/index.md:209-211`: "so concurrent knowledge-base sources spread their retries and avoid hammering the provider together"). A provider under load hands the *same* `Retry-After` to every concurrent source, so all of them resume in lockstep at the same instant — the synchronised burst the additive jitter was introduced to prevent, on the one code path where the provider has explicitly told us it is overloaded. Suggest `delay = min(jitterFn(hint), startBackoffMax)` reusing the same jitter function, and asserting the jittered range in the new test.

## [should-fix] `Retry-After` handling appears nowhere in the durable record

It is absent from the PR body, from the commit message, and from `docs/tools/rag/index.md`'s "Retry policy and parameters" list (`:204-211`), which presents the schedule as `15s → 30s → 1m → 2m → 4m → 5m` with jitter as the only variation. An operator reading that will not understand why a 429 carrying `Retry-After: 120` waited two minutes on the first retry. Since this section is user-facing documentation of the retry policy, the server-hint override belongs in it (with the `startBackoffMax` cap noted).

## [should-fix] `Closes #4060` auto-closes an issue whose remediation is incomplete

#4060's "Expected behavior" lists three items; this PR delivers the first only:

- **Configurable/longer start timeout** — untouched. `DefaultStartTimeout = 30 * time.Second` is unchanged (`pkg/tools/startable.go:409-411`), and this PR's own commit message says so: *"Scope: DefaultStartTimeout (30s) unchanged — deferred."*
- **Chunk-level progress checkpointing** — untouched; `pkg/rag/strategy/vector_store.go` is not in the diff.
- The code-mode composite partial-start subset is explicitly exempt from pacing and tracked as the still-open #4067 (`pkg/tools/startable.go:578-581`).

This was raised in my first pass (#issuecomment-5453119687) and again in the third (#issuecomment-5454627600) and is still present. On squash-merge #4060 closes silently and the two deferred items lose their tracking thread. Please switch to `Refs #4060` and open follow-ups for the deferred timeout and checkpointing, or note in #4060 what remains before closing it.

## [should-fix] The PR body's "Changed files" table is incomplete and internally inconsistent

The table lists 10 rows; the PR changes **17** files. Omitted entirely:

| Omitted file | Size |
|---|---|
| `docs/tools/rag/index.md` | +84 |
| `docs/tools/mcp/index.md` | +2 |
| `docs/tools/lsp/index.md` | +2 |
| `docs/community/troubleshooting/index.md` | +2 |
| `pkg/tools/startable_backoff_regression_test.go` | +301 |
| `pkg/tools/builtin/rag/rag_backoff_test.go` | +129 |
| `pkg/model/provider/dmr/embed_test.go` | +51 |

The 84-line user-facing RAG documentation section — the operator-visible half of this change — is invisible in the description that a squash-merge carries into `main`.

Separately, the Design section says the two sites "now wrap embedding errors in `WrapHTTPError`" while the table row for the same change says `WrapOpenAIError`. The code calls `oaistream.WrapOpenAIError` (`openai/client.go:1168`, `dmr/embed.go:54`), which reaches `modelerrors.WrapHTTPError` **only** for `*openaisdk.Error` and passes every other error through unchanged (`pkg/model/provider/oaistream/wrap.go:16-25`). That distinction is exactly what determines whether the gate arms, so it is worth stating precisely rather than two different ways. Body edit only.

## [optional] Docs say "Maximum delay: 5 minutes"; the effective ceiling is 6

`docs/tools/rag/index.md:207` states `**Maximum delay**: 5 minutes`, but additive jitter is applied *after* the cap, so the realised ceiling is 1.2 × 5min = 6min — as this PR's own test asserts: `assert.Check(t, d <= backoffMax+backoffMax/5, ...)` (`pkg/tools/startable_backoff_test.go:323-324`). A reader can derive it from the jitter bullet two lines down, so this is minor; `5 minutes (up to 6 minutes with jitter)` removes the need to.

## [optional] Only one of the two embedding wrap sites has a test

`pkg/model/provider/dmr/embed_test.go` proves the DMR site end-to-end against a live 429; `grep -n CreateBatchEmbedding pkg/model/provider/openai/*_test.go` returns nothing, so the `openai` site (`client.go:1168`) is unpinned. The wrapper is shared so risk is low — but the OpenAI-compatible client is the path #4060 was actually reported through (Anthropic behind an OpenAI-shaped gateway), and it is the site most likely to regress if someone later "simplifies" the wrap back to a bare `%w`. A near-copy of the DMR test would close it.

## Checklist against #4060's PR1/PR2 contract

**Met:** retryable-only pacing; bounded exponential backoff with jitter (15s base, 5min cap, `[d, 1.2d]`); immediate cancellation/shutdown handling (no timers, goroutines or sleeps — pinned by `TestBackoffRegression_GateSpawnsNoTimersOrGoroutines`; bare context errors never arm; `Stop` clears the window); successful-start path, retry limits and partial-start latch preserved; concurrent starts do not multiply retry activity; non-retryable failures still fail promptly; already-started toolsets not restarted; blocking `Start` ungated so `mcpcatalog enable` and skill sub-sessions stay immediate; RAG/MCP/LSP-shaped regression coverage plus a real-toolset integration test; retry/cancellation/classification semantics and troubleshooting documented; no config or schema change; build, tests (with `-race`) and lint green.

**Outstanding:** the untested `Retry-After` branch (blocking), its jitter bypass, and the issue-closure / PR-body accuracy items above.

Filing this as a comment rather than a formal request-for-changes only because I authored the PR and GitHub will not accept a self-review verdict; please treat the blocking finding as a merge gate. The core design is right and the previously-blocking gaps are genuinely closed — happy to re-review promptly once `Retry-After` is covered by a test and the closure keyword / description are corrected.

aheritier (commented):

Re-reviewed at head `e4ae8fd7` (identical content to the previously-reviewed `c82351f6`, force-pushed as a rebase onto a fresher `main` — same 18 files, same 1690/13 insertion/deletion counts, commit message unchanged; verified the rebase was mechanically conflict-free since the only intervening `main` commit, #4071, touched only `CHANGELOG.md`). CI is green: 16 checks pass, 3 skipping (post-merge publish jobs), 0 failing. `mergeable=MERGEABLE`; `mergeStateStatus=BLOCKED` solely on the missing required approving review.

This addresses my previous review (`fe986e7`, #pullrequestreview-5052930334). Going through each finding against the current head:

## [blocking] Retry-After override branch — now covered

Four dedicated tests exercise every sub-case in `pkg/tools/startable_backoff_test.go`:

- `TestStartableToolSet_RetryAfterHonoured` (line 613): hint > computed delay → window extends to the hint exactly.
- `TestStartableToolSet_RetryAfterBelowComputed` (line 696): hint (5s) < computed delay (30s at attempt 2) → hint ignored, computed window holds.
- `TestStartableToolSet_RetryAfterCapAtMax` (line 730): a 24h hint is capped at `startBackoffMax` (5min).
- `TestStartableToolSet_RetryAfterJittered` (line 656): 60 independent gates armed with the same hint, probed at the midpoint of `[hint, 1.2×hint]` — confirms windows are spread across the jitter band, not all set to the exact hint.

`go tool cover` on `./pkg/tools/` (86.4%) now shows `setStartBackoff` (`pkg/tools/startable.go:466`) at **100%** and no zero-coverage lines remain in `startable.go` or `startable_backoff.go` — the previously-unexecuted override branch (was lines 476-477, 0 executions) is exercised. All four tests pass:

```
--- PASS: TestStartableToolSet_RetryAfterHonoured
--- PASS: TestStartableToolSet_RetryAfterBelowComputed
--- PASS: TestStartableToolSet_RetryAfterCapAtMax
--- PASS: TestStartableToolSet_RetryAfterJittered
```

## [should-fix] Retry-After jitter bypass — fixed

`setStartBackoff` (`pkg/tools/startable.go:476-481`) now applies the same jitter function to the Retry-After branch as the computed-delay path:

```go
if hint := retryAfterHint(err); hint > delay {
    if s.startJitter != nil {
        delay = s.startJitter(min(hint, startBackoffMax))
    } else {
        delay = additiveJitter(min(hint, startBackoffMax))
    }
}
```

`TestStartableToolSet_RetryAfterJittered` statistically confirms the spread (not all 60 samples land on the same side of the midpoint).

## [should-fix] Retry-After undocumented — fixed

`docs/tools/rag/index.md:212` now has a "Retry-After override" bullet under "Retry policy and parameters", stating the hint overrides the computed delay, is capped at the 5-minute maximum, and gets the same additive jitter. This also incidentally closes the optional finding below about the 5min→6min ceiling — the same section's "Maximum delay" line now reads "up to ~6 minutes (5-minute cap plus up to 20% additive jitter)".

## [should-fix] `Closes #4060` → fixed

The commit message and PR description both now read `Refs #4060 (partial — ...)`. Squash-merging this PR will no longer auto-close #4060.

## [should-fix] PR body "Changed files" table incomplete — fixed

The table now lists all 18 files in the diff (verified against `git show --stat` on the current head), including the previously-omitted `docs/tools/rag/index.md`, `docs/tools/mcp/index.md`, `docs/tools/lsp/index.md`, `docs/community/troubleshooting/index.md`, `pkg/tools/startable_backoff_regression_test.go`, `pkg/tools/builtin/rag/rag_backoff_test.go`, and `pkg/model/provider/dmr/embed_test.go`. The Design section's embedding-wrap description is also now precise: it names `oaistream.WrapOpenAIError` (not the imprecise `WrapHTTPError`) and explains that it wraps `*openaisdk.Error` specifically.

## [optional] "5 minutes" → "up to 6 minutes" — fixed (see above)

## [optional] Only DMR embed site tested — fixed

`pkg/model/provider/openai/embed_test.go` now exists with `TestCreateBatchEmbedding_429SurfacesAsStatusError` and a 5xx equivalent, mirroring the DMR test. Both embedding-wrap sites are now pinned.

## Re-validation on the current head

```
go build ./...                                                          # clean
go test ./pkg/tools/... ./pkg/modelerrors/... ./pkg/model/provider/...   # all green
golangci-lint run ./pkg/tools/ ./pkg/tools/mcp/ ./pkg/modelerrors/       # 0 issues
```

(`-race` isn't available in this environment — no `gcc` for cgo — so I could not re-run the race-detector pass from my previous review; the non-race suite is green and nothing in this delta touches concurrency-sensitive code paths beyond what was already reviewed.)

All previously-blocking and should-fix findings are resolved. Filing as a comment rather than a formal approval only because I authored the PR and GitHub blocks self-approval, same as every prior round.

dgageot (approved):

PR #4064, opened 2026-08-27
Persist compaction atomically

krissetto:
Persist compaction summaries atomically with the session metadata, so manually and automatically compacted sessions restore correctly on session reload.

Currently if you reload a long session that has had compactions and you try to continue it, no compaction will be found, auto compaction will trigger on the full chat, and the request will fail if the full chat goes over the max context length of the model in use


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/docker-agent/actions/runs/33065204770).

Sayt-0 (approved):

PR #4072, opened 2026-08-28
hooks/builtins: name the resolved shell in the env block

trungutt:
## The problem

On Windows, docker-agent sessions repeatedly emit POSIX syntax — `&&`, `grep`, `head`, `2>/dev/null`, `ls -la`, `$(pwd)` — under Windows PowerShell 5.1 or cmd.exe. Each attempt fails with a distinctive error, the model re-derives the dialect from scratch, and often emits the same wrong syntax on the next turn. Long sessions burn many tool calls this way.

## Why

The env block at the top of every session says:

```
<env>
  Operating System: Windows
  ...
</env>
```

`Windows` alone doesn't disambiguate the shell (PowerShell 5.1? 7? cmd.exe? — different dialects). The model defaults to POSIX by habit since that's most of its training data. The shell tool description does carry a shell-specific hint (`shellSyntaxHint`), but it sits far from where the model is paying attention.

## The change

Add a `Shell:` line to the env block, resolved via a new `shellpath.ShellBaseName` helper:

```
<env>
  Operating System: Windows
  ...
  Shell: powershell (C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe)
</env>
```

Same treatment for `pwsh`, `cmd`, `zsh`, `bash`, etc. Prompt-cached, so no per-turn cost.


pkg/shellpath/shellpath.go
```diff
@@ -7,8 +7,19 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
 
+// ShellBaseName returns the lowercase shell name without extension. Splits on
+// both separators so results are stable when the path came from another host OS.
+func ShellBaseName(shellPath string) string {
```
Sayt-0, line 15: `ShellBaseName` is byte-for-byte identical to the private `shellBaseName` in `pkg/tools/builtin/shell/shell.go` (line 442), which also carries its own table test in `shell_test.go` (line 243). Now that the logic is exported here, `shell.go` can delegate to `shellpath.ShellBaseName` and drop the duplicate implementation and redundant test cases; otherwise the two copies will drift.

pkg/hooks/builtins/add_environment_info.go
```diff
@@ -6,34 +6,37 @@ import (
...
 		return nil, nil
 	}
 	return hooks.NewAdditionalContextOutput(hooks.EventSessionStart, environmentInfo(in.Cwd)), nil
 }
 
-// environmentInfo formats the env block injected at session_start:
-// working directory, git-repo status, and human-readable OS / arch.
+// environmentInfo builds the <env> block. Long-form dialect rules live in
+// shellSyntaxHint (tool description) and shellDialectHint (reactive) so this
+// stays terse.
```
Sayt-0: `shellDialectHint` does not exist anywhere in the repo; only `shellSyntaxHint` does. If it belongs to a companion branch, the reference can be added when that change lands. For now the mention should be dropped to avoid a dead pointer.

```suggestion
// environmentInfo builds the <env> block. Long-form dialect rules live in
// shellSyntaxHint (tool description) so this stays terse.
```

pkg/hooks/builtins/add_environment_info.go
```diff
@@ -6,34 +6,37 @@ import (
...
 func environmentInfo(workingDir string) string {
 	gitRepo := "No"
 	if isGitRepo(workingDir) {
 		gitRepo = "Yes"
 	}
+	shellPath, _ := shellpath.DetectShell()
```
Sayt-0: Nit: the discarded second value is `argsPrefix`, not an error, but `_` reads like an ignored error at a glance. A short clarification avoids the misread:

```suggestion
	shellPath, _ := shellpath.DetectShell() // second value is argsPrefix, unused here
```

pkg/hooks/builtins/add_environment_info.go
```diff
@@ -6,34 +6,37 @@ import (
...
 	"github.com/docker/docker-agent/pkg/hooks"
+	"github.com/docker/docker-agent/pkg/shellpath"
 )
 
 // AddEnvironmentInfo is the registered name of the add_environment_info builtin.
 const AddEnvironmentInfo = "add_environment_info"
 
-// addEnvironmentInfo emits cwd / git / OS / arch info as session_start
-// additional context. No-op when Cwd is empty.
+// addEnvironmentInfo emits cwd/git/OS/arch/shell as session_start context.
```
Sayt-0, line 15: The previous doc comment noted the no-op behavior when `Cwd` is empty; that detail is worth keeping since it explains the early return:

```suggestion
// addEnvironmentInfo emits cwd/git/OS/arch/shell as session_start context.
// No-op when Cwd is empty.
```

Sayt-0 (changes requested):
The `Shell:` line in the env block addresses a real gap: sessions on Windows repeatedly emit POSIX syntax under PowerShell 5.1 / cmd.exe, and the existing `shellSyntaxHint` sits too far from the model's attention. The approach (reuse `shellpath.DetectShell`, prompt-cached line) is sound. Build and tests pass on the branch.

Two items to resolve before merge:

| # | Issue | Files | Severity |
|---|-------|-------|----------|
| 1 | `ShellBaseName` duplicates an existing private helper | `pkg/shellpath/shellpath.go`, `pkg/tools/builtin/shell/shell.go` | blocking |
| 2 | Doc comment references `shellDialectHint`, which does not exist in the repo | `pkg/hooks/builtins/add_environment_info.go` | blocking |

Details in the inline comments. Two minor nits included as well, take or leave.

pkg/shellpath/shellpath.go
trungutt: Done in 715df7bae. Dropped `shellBaseName`/its table test from `pkg/tools/builtin/shell/shell.go` and pointed the two call sites (plus their test asserts) at `shellpath.ShellBaseName`. The equivalent table test already lives in `pkg/shellpath/shellpath_test.go`.

pkg/hooks/builtins/add_environment_info.go
trungutt: Done in 360991093. Dropped the dead `shellDialectHint` reference.

pkg/hooks/builtins/add_environment_info.go
trungutt: Done in 360991093. Added the inline note that the second value is `argsPrefix`.

pkg/hooks/builtins/add_environment_info.go
trungutt: Done in 360991093. Restored the "No-op when Cwd is empty." line.

aheritier (commented):
CI is green on 3609910 (build-and-test, lint, license-check, windows-tests, build-image linux/amd64+arm64, CodeQL go/js/actions), and all three commits are signed and verified. Verified independently on the head commit: `go build ./...`, `GOOS=windows GOARCH=amd64 go build ./pkg/...`, and `go test ./pkg/shellpath/... ./pkg/hooks/builtins/... ./pkg/tools/builtin/shell/...` all pass.

The change does what the description says, and the previous round is fully addressed:

- `shellpath.ShellBaseName` is now the single implementation: `grep -rn '\bshellBaseName\b'` returns nothing, both call sites (`shell.go:371`, `shell.go:412`) delegate, and the moved table test in `pkg/shellpath/shellpath_test.go:10` is a superset of the removed one (adds `/bin/sh` and `""`).
- `grep -rn shellDialectHint` returns nothing — the dead pointer is gone.
- The "No-op when Cwd is empty." note and the `argsPrefix` clarification are back.
- The core invariant holds: the hook resolves via `shellpath.DetectShell()`, the same source as `shellHandler.shell` (`shell.go:294`), so the env block cannot name a shell different from the one the tool executes.
- The prompt-cache claim checks out — `AddEnvironmentInfo` output sits behind the `cache_control` marker (`pkg/session/session.go:2227`).

## Findings

**[should-fix] Docs enumerate the env block without the new shell field.**
Two reference tables list the injected fields exhaustively and are now incomplete:
- `docs/configuration/hooks/index.md:209` — "Adds the working directory, git-repo status, OS, and CPU architecture."
- `docs/configuration/agents/index.md:96` — "injects working directory, OS, CPU architecture, and git info into context."

Adding the resolved shell to both keeps them accurate in the same change.

**[should-fix] `465e8955`'s commit body will land in `main` with stale references.**
Squash merge is disabled on this repo (`allow_squash_merge: false`), so all three commit messages are permanent history. That body cites `shellDialectHint` — which does not exist in the repo — and "follow-up commit", which stops resolving once the message is in history; it also states "the pre-existing copy in the shell toolset moves over here too", which only became true in `715df7ba`. The code comment was corrected in `3609910`, but the commit body still carries the dead pointer. Worth a reword on the next rebase.

**[optional] The env-block / shell-tool agreement isn't tested.**
`add_environment_info_test.go:54,61` builds the expected string from the same `shellpath.DetectShell()` and `shellpath.ShellBaseName()` calls production uses, so it pins the line's placement and format but cannot fail if the hook's resolution later diverges (e.g. someone swaps in `DetectUnixShell`) — which is the invariant this change exists to protect. It matches the existing `displayOS()`/`displayArch()` style in that test, so take or leave.

**[optional] `pkg/shellpath`'s package comment (`shellpath.go:1`)** still scopes the package to "safe shell binary resolution to prevent PATH hijacking attacks (CWE-426)", while `ShellBaseName` is a pure formatting helper. A one-line broadening keeps the doc honest now that the package hosts both.

Leaving this as a comment rather than an approval so the two `[should-fix]` items — the docs enumerations and the `465e8955` commit body — can be picked up first; neither is a correctness problem, and the code itself looks good to me. Note that merge is separately gated by the earlier CHANGES_REQUESTED review, which resolving threads does not dismiss; the original reviewer needs to re-review or that review needs dismissing.

Sayt-0 (approved):

PR #4077, opened 2026-08-30
test(runtime): add persistence observer streaming benchmarks

salignatmoandal:
## Summary

Adds characterization tests and benchmarks for `PersistenceObserver` streaming persistence — the path that mirrors assistant token deltas (`AgentChoice` / `AgentChoiceReasoning`) into a single growing message row in the session store.

This establishes a baseline before any future optimization (e.g. debounced flushes) and documents the per-chunk store write contract that the observer currently implements.

## Context

During a streaming assistant turn, the runtime emits one `AgentChoiceEvent` per delta. `PersistenceObserver.persistStreamingContent` keeps a single in-flight row and:

1. **INSERT** (`AddMessage`) on the first chunk
2. **UPDATE** (`UpdateMessage`) on every subsequent chunk
3. **UPDATE** again on `MessageAddedEvent` to finalise the row with the canonical payload

This behaviour is easy to regress when refactoring persistence or the store layer, but was not previously covered by a focused unit test or benchmark.

## Changes

New file: `pkg/runtime/persistence_observer_bench_test.go`

### Tests

| Test | What it pins |
|------|--------------|
| `TestPersistenceObserver_UpdateCountPerChunk` | 1 `AddMessage` + `(N-1)` `UpdateMessage` calls for N streaming chunks, plus one final `UpdateMessage` on `MessageAdded` |
| `TestPersistenceObserver_StreamingContentAccumulates` | Mid-stream reload shows accumulated assistant text (`"hel"` + `"lo"` → `"hello"`) |

Both tests use a `countingStore` wrapper around `InMemorySessionStore` to assert store call counts without mocking the observer.

### Benchmarks

Each iteration simulates a long assistant turn: **500 `AgentChoice` deltas + 1 `MessageAdded` finalisation** (`streamingBenchChunks = 500`).

| Benchmark | Store | Purpose |
|-----------|-------|---------|
| `BenchmarkPersistenceObserver_StreamingChunks` | `InMemorySessionStore` | CPU / alloc cost of observer logic + in-memory clone path |
| `BenchmarkPersistenceObserver_StreamingChunks_SQLite` | SQLite `:memory:` | Realistic cost including `json.Marshal` + `UPDATE session_items` per chunk |

### Sample results (darwin/arm64, Apple M3 Pro)

```
BenchmarkPersistenceObserver_StreamingChunks-12          ~2.0 ms/op    502 KB/op    2515 allocs/op
BenchmarkPersistenceObserver_StreamingChunks_SQLite-12   ~3.1 ms/op   1558 KB/op    8714 allocs/op
```

Per chunk (500 chunks/iter): ~4 µs in-memory, ~6 µs SQLite, ~5 vs ~17 allocs.

SQLite is ~1.5× slower and ~3× more alloc-heavy — expected given JSON marshal + SQL per update.

### Known benchmark caveats

- **In-memory drift**: the in-memory benchmark reuses the same session across `b.N` iterations, so `UpdateMessage` eventually scans an ever-growing message list (O(messages) per update). Later iterations are slower than the first. SQLite stays flat because updates are keyed by `message_id`.
- **Migration logs**: SQLite setup calls `NewSQLiteSessionStoreFromDB`, which logs migration info before `b.ResetTimer()` — noisy stdout but not included in `ns/op`.

These caveats are acceptable for a baseline but worth keeping in mind when comparing future numbers.

## Why now?

Streaming persistence is on the hot path for every assistant response when a session store is configured. Having explicit tests + benchmarks makes it safer to:

- change the flush strategy (per-chunk vs debounced)
- optimize `persistStreamingContent` (avoid `strings.Builder.String()` copies on every delta)
- compare in-memory vs SQLite store performance under realistic chunk counts

## Test plan

- [x] `go test ./pkg/runtime -run TestPersistenceObserver_UpdateCountPerChunk -v`
- [x] `go test ./pkg/runtime -run TestPersistenceObserver_StreamingContentAccumulates -v`
- [x] `go test ./pkg/runtime -run=^$ -bench=BenchmarkPersistenceObserver -benchmem -count=1`
- [ ] `task test` (full suite)
- [ ] `task lint`


Sayt-0 (commented):
Useful addition: the streaming persistence contract (1 `AddMessage` + N-1 `UpdateMessage` per turn) was not pinned by any focused test, and both characterization tests verify correctly against `persistStreamingContent`. Tests pass locally and follow the surrounding conventions (`t.Parallel`, `t.Context`, testify, `:memory:` + `SetMaxOpenConns(1)` pattern from `pkg/session/store_memory_test.go`).

Main issue: the in-memory benchmark result is a function of `b.N` (details inline), which defeats the stated goal of establishing a baseline. The SQLite benchmark is fine (reproduced ~2.9 ms/op, stable across benchtime values).

| Area | Status |
|---|---|
| Characterization tests | correct, verified against `persistence_observer.go` |
| SQLite benchmark | stable, no change needed |
| In-memory benchmark | numbers depend on `b.N`, needs fix (inline comment) |
| Conventions, vet | clean |

Non-blocking: the file name `persistence_observer_bench_test.go` also hosts two unit tests; moving them to `persistence_observer_streaming_test.go` would keep the bench file benchmark-only, matching `pkg/tui/components/message/bench_test.go`.

pkg/runtime/persistence_observer_bench_test.go
```diff
@@ -0,0 +1,161 @@
...
+func BenchmarkPersistenceObserver_StreamingChunks(b *testing.B) {
+	ctx := b.Context()
+	obs, sess, _ := setupPersistenceObserverBench(b)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for range b.N {
+		emitStreamingChunks(ctx, obs, sess, streamingBenchChunks)
+		finalizeStreamingMessage(ctx, obs, sess)
+	}
```
Sayt-0: The shared session makes `ns/op` a function of `b.N`: each iteration leaves one extra message row in `bench-session`, and `InMemorySessionStore.UpdateMessage` scans every message of every session on each call, so later iterations pay O(iterations) per chunk.

Measured on this branch (darwin/arm64):

| benchtime | ns/op |
|---|---|
| 200x | 195,198 |
| 2000x | 624,997 |
| 8000x | 2,398,435 |

The PR's headline number (~2.0 ms/op) is therefore an artifact of the iteration count picked by the harness, not a baseline that future optimizations can be compared against. The "in-memory drift" caveat in the description understates this: the number is not noisy, it is unbounded.

Keeping the store bounded flattens the result (~175 µs/op at 200x, 2000x and 8000x, verified):

```go
for i := range b.N {
    sess := session.New(session.WithID(strconv.Itoa(i)), session.WithUserMessage("hi"))
    if err := store.AddSession(ctx, sess); err != nil {
        b.Fatal(err)
    }
    emitStreamingChunks(ctx, obs, sess, streamingBenchChunks)
    finalizeStreamingMessage(ctx, obs, sess)
    if err := store.DeleteSession(ctx, sess.ID); err != nil {
        b.Fatal(err)
    }
}
```

Requires the `strconv` import and the store returned by `setupPersistenceObserverBench`. Note that a fresh session per iteration alone is not enough: `UpdateMessage` ranges over all sessions, so the store must not accumulate them, hence the `DeleteSession`. If per-iteration setup cost is a concern, it is ~2 allocations against ~2,500 per iteration, so it does not move the numbers.

pkg/runtime/persistence_observer_bench_test.go
```diff
@@ -0,0 +1,161 @@
...
+	s.addCalls.Add(1)
+	return s.InMemorySessionStore.AddMessage(ctx, sessionID, msg)
+}
+
+func (s *countingStore) UpdateMessage(ctx context.Context, messageID int64, msg *session.Message) error {
+	s.updateCalls.Add(1)
+	return s.InMemorySessionStore.UpdateMessage(ctx, messageID, msg)
+}
+
+func setupPersistenceObserverBench(tb testing.TB) (*PersistenceObserver, *session.Session, *session.InMemorySessionStore) {
```
Sayt-0: Nit: the third return value is discarded by the only caller. With the bounded-store fix suggested in the other comment, the session return becomes the unused one instead. Returning `(obs, store)` and letting benchmarks create their own sessions would keep the helper minimal.

pkg/runtime/persistence_observer_bench_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func openBenchSQLiteStore(b *testing.B) *session.SQLiteSessionStore {
+	b.Helper()
+	db, err := sql.Open("sqlite", ":memory:")
+	require.NoError(b, err)
+	b.Cleanup(func() { _ = db.Close() })
+	db.SetMaxOpenConns(1)
+
+	store, err := session.NewSQLiteSessionStoreFromDB(b.Context(), db)
```
Sayt-0, line 172: Nit: migration INFO logs are emitted on every harness re-invocation of the benchmark function, interleaving with the `-bench` output. Not counted in `ns/op` since setup runs before `b.ResetTimer`, but redirecting the default slog handler to `io.Discard` for the benchmark would keep the output clean for tools that parse it.

pkg/runtime/persistence_observer_bench_test.go
salignatmoandal: Good catch — confirmed in `InMemorySessionStore.UpdateMessage`: it `Range`s every session and scans `Messages`, so cost is O(sessions × messages). With a shared session the store grew unbounded with `b.N`, so `ns/op` was not a baseline.

Applied your suggested harness in a0ddcd4: each iteration does `AddSession` → work → `DeleteSession` (fresh session alone is not enough because of the session `Range`). Also simplified the helper to `(obs, store)`.

Re-measured on darwin/arm64 (Apple M3 Pro):

| benchtime | before (unbounded) | after (bounded) |
|---|---|---|
| 200x | ~203 µs | **165 µs** |
| 2000x | ~673 µs | **160 µs** |
| 8000x | ~2575 µs | **160 µs** |

`ns/op` is now flat (~160 µs), matching your ~175 µs result. Allocs stay ~2522/op.

Happy to follow up on the slog `io.Discard` nit and splitting unit tests into `persistence_observer_streaming_test.go` if you want those in this PR too.

pkg/runtime/persistence_observer_bench_test.go
salignatmoandal: Done in a0ddcd4 — `setupPersistenceObserverBench` now returns `(obs, store)`; the benchmark creates/deletes its own sessions per iteration.

pkg/runtime/persistence_observer_bench_test.go
salignatmoandal: Done in 29b4b7f — `openBenchSQLiteStore` now sets the default slog handler to `DiscardHandler` (restored via `b.Cleanup`) so migration INFO no longer interleaves with `-bench` output.

Numbers unchanged: in-memory stays flat (~160–180 µs/op across 200x/2000x/8000x), SQLite ~3.2 ms/op.

pkg/runtime/persistence_observer_bench_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/docker/docker-agent/pkg/session"
+)
+
+// streamingBenchChunks is the number of AgentChoice deltas exercised per
+// benchmark iteration — roughly the order of magnitude of a long assistant turn.
+const streamingBenchChunks = 500
+
+// countingStore wraps an in-memory store and records AddMessage / UpdateMessage
+// calls so tests can pin the per-chunk persistence contract.
+type countingStore struct {
```
aheritier, line 25: countingStore now collides with the helper in generated_file_test.go, so the pkg/runtime test binary fails to compile when this is rebased onto current main. Please rename this helper (for example, countingPersistenceStore).

pkg/runtime/persistence_observer_bench_test.go
```diff
@@ -0,0 +1,176 @@
...
+		InMemorySessionStore: session.NewInMemorySessionStore().(*session.InMemorySessionStore),
+	}
+}
+
+func (s *countingStore) AddMessage(ctx context.Context, sessionID string, msg *session.Message) (int64, error) {
+	s.addCalls.Add(1)
+	return s.InMemorySessionStore.AddMessage(ctx, sessionID, msg)
+}
+
+func (s *countingStore) UpdateMessage(ctx context.Context, messageID int64, msg *session.Message) error {
```
aheritier, line 42: This overrides the embedded method with the pre-sessionID signature. Current session.Store.UpdateMessage requires UpdateMessage(ctx, sessionID, messageID, msg), which means this wrapper no longer satisfies session.Store and breaks existing call sites. Please update and forward sessionID; embedding session.Store rather than the concrete in-memory store would also avoid the unnecessary type assertions.

pkg/runtime/persistence_observer_bench_test.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// streamingBenchChunks is the number of AgentChoice deltas exercised per
+// benchmark iteration — roughly the order of magnitude of a long assistant turn.
+const streamingBenchChunks = 500
+
+// countingStore wraps an in-memory store and records AddMessage / UpdateMessage
+// calls so tests can pin the per-chunk persistence contract.
+type countingStore struct {
+	*session.InMemorySessionStore
```
aheritier, line 26: embeddedstructfieldcheck requires a blank line between the embedded field and regular fields, so task lint currently fails.

aheritier (changes requested):
The benchmark behavior looks sound, but this branch is 592 commits behind main and does not compile against it due to the helper collision and stale UpdateMessage signature above. There is also a lint failure. Please rebase onto current main, address these items, and rerun CI.

pkg/runtime/persistence_observer_bench_test.go
salignatmoandal: Hi @aheritier — thanks for the review. I've rebased onto current main and addressed the items:

Renamed countingStore → countingPersistenceStore to avoid the collision with generated_file_test.go
Updated UpdateMessage to the current session.Store signature (ctx, sessionID, messageID, msg) and embedded session.Store instead of the concrete in-memory type
Added the blank line after the embedded field for embeddedstructfieldcheck
Locally: task lint and task test are green. Commits are re-signed (ED25519) and force-pushed. Happy to mark the PR ready for re-review whenever you are.

PR #4081, opened 2026-08-31
safety: close four classifier coverage gaps

trungutt:
## Why

The shell-safety classifier reports `unknown` — losing the blast-radius label and category tag it would otherwise expose — for a set of commands that are unambiguously destructive, and under-reports the blast radius on two `docker` commands where a critical flag is present but appears in a non-canonical position. Permissive safety modes end up asking for confirmation without saying *why*, and (for the flag-order cases) miss that the user is asking to delete named volumes.

Four independent gaps, one commit each:

### 1. Flag-order variants for `docker system prune --volumes` and `docker compose down -v`

Destructive patterns are matched contiguously. `docker system prune --volumes` (HIGH) does not cover `docker system prune -f --volumes` because `-f` sits between the two anchor tokens, so the command falls back to bare `docker system prune` (MEDIUM) even though the `--volumes` flag *is* present and named volumes will be wiped.

The same shape recurs on compose: `docker compose down --volumes` (HIGH) does not cover `docker compose down --remove-orphans -v` or `docker compose down -v --remove-orphans`.

Fix: enumerate the observed flag orderings for `-f` / `-a` / `-af` / `-fa` / `-v` / `--volumes` / `--remove-orphans`.

### 2. PowerShell destructive patterns

The taxonomy is Unix-only. Windows users running the shell tool in PowerShell produce destructive commands (`Remove-Item -Path <path> -Recurse -Force`, `Clear-Content`) that classify as `unknown`. Add fs-delete and fs-overwrite entries for the PowerShell equivalents, covering positional/`-Path`/trailing-path forms and both `-Recurse -Force` and `-Force -Recurse` orderings.

### 3. Destructive SQL, app-CLI, and key-value patterns

Destructive intent frequently arrives via a wrapper — `docker exec <ctr> …`, `kubectl exec …`, an SSH login — that hides the dangerous verb inside the inner command. Because the existing taxonomy only inspects the outer shell verb, real data-loss commands (`DROP DATABASE`, `TRUNCATE TABLE`, `redis-cli FLUSHALL`, `n8n db:reset`, `rails db:reset`, `bench new-site --force`, …) come back as `unknown`.

Add pattern groups that fire on the destructive verb wherever it appears, so wrapping through `docker exec` / `kubectl exec` / `ssh` inherits the classification automatically:

- **SQL DDL/DML**: `drop database`, `drop schema`, `drop table`, `truncate table`, `delete from` (covers `mysql -e`, `psql -c`, `sqlite3 db "…"`, `mariadb -e`).
- **App-framework resets**: `n8n db:reset`, `n8n user-management:reset`, `bench new-site --force`, `bench drop-site`, `rails db:drop` / `db:reset` / `db:schema:load`, `manage.py flush`, `prisma migrate reset`, `sequelize db:drop`.
- **Key-value flushes**: `redis-cli flushall`, `redis-cli flushdb`.

### 4. Whitespace anchor for shell-metachar patterns

The destructive-pattern regex prefix is `(?:^|.*\b)`, which fires on any word boundary. That's the right anchor for alphanumeric verbs (`rm`, `docker`), but wrong for patterns whose leading token is itself a shell metacharacter — `>`, `<`, `|`, `&`, `;`. A word boundary sits between any word char and the metachar, so:

- `--email <EMAIL> --firstName Admin` matched the truncate pattern `> <file>` via the `l>` boundary inside a redacted placeholder. This is the only case that reproduced on master; the placeholder shape (word-char, `>`, space) is what defeats the `\b` anchor.

Anchor shell-metacharacter patterns on `\s` instead: a real redirect, pipe, or chain operator is preceded by whitespace (or start of line) after `normalizeCommand` collapses runs to single spaces. Verbal patterns keep the `\b` anchor.

## Notes for the reviewer

Each commit is standalone and lands one gap. Regression tests are colocated in `safety_test.go` in each commit.


pkg/safety/safety_patterns.json
```diff
@@ -115,6 +143,32 @@
...
+      { "pattern": "drop schema <name>",             "blast_radius": "HIGH",        "category": "sql-drop",         "notes": "destroys schema and its objects" },
+      { "pattern": "drop schema if exists <name>",   "blast_radius": "HIGH",        "category": "sql-drop",         "notes": "idempotent form" },
+      { "pattern": "drop table <name>",              "blast_radius": "HIGH",        "category": "sql-drop",         "notes": "destroys table schema and rows" },
+      { "pattern": "drop table if exists <name>",    "blast_radius": "HIGH",        "category": "sql-drop",         "notes": "idempotent form" },
+      { "pattern": "truncate table <name>",          "blast_radius": "MEDIUM-HIGH", "category": "sql-truncate",     "notes": "removes all rows; typically non-transactional" },
+      { "pattern": "delete from <table>",            "blast_radius": "MEDIUM",      "category": "sql-delete",       "notes": "unqualified DELETE would remove every row; presence of WHERE cannot be checked from the command shape" }
```
Sayt-0: **[Blocking]** These verb-anchored patterns match anywhere in the command, including inside quoted arguments of safe-listed commands. Verified: `grep "drop table users" -r .` was safe on master and is destructive/high here; under `restricted` mode that flips ALLOW to DENY, and under `balanced` it turns an auto-approved search into a prompt. `delete from` and `drop table` are common strings in migrations and ORM code, so agents inspecting a schema will hit this often.

Options, in decreasing order of precision:
1. Require a client context: `mysql ... delete from`, `psql ... delete from`, and so on. Loses the wrapper-agnostic property the PR aims for.
2. Keep the patterns but add the known safe-list collisions as regression tests documenting the intended precedence (destructive wins over safe).
3. Keep as is with a `notes` caveat.

Option 2 seems like the minimum bar; if option 1 seems more appropriate, the `...` wildcard makes it cheap (`mysql ... delete from ...`).

pkg/safety/safety_patterns.json
```diff
@@ -37,6 +37,24 @@
       { "pattern": "chmod -R 777 <path>",          "blast_radius": "LOW",         "category": "fs-permissions",   "notes": "security regression, reversible" },
       { "pattern": "sed -i",                       "blast_radius": "MEDIUM",      "category": "fs-modify",        "notes": "in-place file edit; -i.bak variants also match" }
     ],
+    "powershell": [
+      { "pattern": "Remove-Item -Path <path> -Recurse -Force", "blast_radius": "HIGH",   "category": "fs-delete",      "notes": "PowerShell recursive force delete; equivalent to `rm -rf`" },
```
Sayt-0: `<path>` compiles to `\S+`, which cannot span a quoted path containing spaces. Because the destructive flags sit after the placeholder, contiguity breaks and the command falls through to the weakest pattern. Verified:

| command | matched pattern | blast radius |
|---|---|---|
| `Remove-Item "C:\tmp\foo" -Recurse -Force` | `Remove-Item <path> -Recurse -Force` | high |
| `Remove-Item "C:\Program Files\MyApp\cache" -Recurse -Force` | `Remove-Item <path>` | low |

Paths with spaces are the norm on Windows (`Program Files`, `OneDrive` folders). A wildcard form such as `Remove-Item ... -Recurse` / `Remove-Item ... -Recurse ... -Force` would cover every path shape and collapse most of the 14 `Remove-Item` entries at the same time.

pkg/safety/safety_patterns.json
```diff
@@ -94,7 +112,14 @@
       { "pattern": "docker system prune -a",                                 "blast_radius": "MEDIUM-HIGH", "category": "dk-multi-del",     "notes": "adds all unused images" },
       { "pattern": "docker system prune --all",                              "blast_radius": "MEDIUM-HIGH", "category": "dk-multi-del",     "notes": "long form of -a" },
       { "pattern": "docker system prune --volumes",                          "blast_radius": "HIGH",        "category": "dk-multi-del",     "notes": "adds named-volume deletion" },
+      { "pattern": "docker system prune -f --volumes",                       "blast_radius": "HIGH",        "category": "dk-multi-del",     "notes": "flag-order variant: force between prune and --volumes" },
+      { "pattern": "docker system prune -a --volumes",                       "blast_radius": "HIGH",        "category": "dk-multi-del",     "notes": "flag-order variant: -a between prune and --volumes" },
+      { "pattern": "docker system prune --volumes -f",                       "blast_radius": "HIGH",        "category": "dk-multi-del",     "notes": "flag-order variant: --volumes precedes -f" },
```
Sayt-0: The compiled regex ends with `(?:$|\b.*)`, so any existing pattern already covers commands that only append flags after it. Verified against master:

| new entry | already covered by | on master |
|---|---|---|
| `docker system prune --volumes -f` | `docker system prune --volumes` | high |
| `docker system prune --volumes -a` | `docker system prune --volumes` | high |
| `docker system prune --volumes -af` | `docker system prune --volumes` | high |
| `docker system prune --volumes -fa` | `docker system prune --volumes` | high |
| `docker system prune -a --volumes` | `docker system prune -a` | high |
| `docker compose down -v --remove-orphans` | `docker compose down -v` | high |

The classifier even reports the old pattern in its reason for these (verified: `docker system prune --volumes -f` matches `docker system prune --volumes`, not the new entry). The genuine gaps are only `-f --volumes`, `-fa --volumes`, `--remove-orphans -v`, `--remove-orphans --volumes`. Suggestion: drop the six redundant entries, or replace the whole set with wildcard entries (`docker system prune ... --volumes`, `docker compose down ... -v` plus the `--volumes` long form), which also covers orderings not enumerated here (`--filter`, `--timeout`, future flags).

pkg/safety/safety_patterns.json
```diff
@@ -115,6 +143,32 @@
...
+      { "pattern": "rails db:reset",                 "blast_radius": "HIGH",        "category": "app-reset",        "notes": "Rails: drop + create + migrate + seed" },
+      { "pattern": "rails db:schema:load",           "blast_radius": "HIGH",        "category": "app-reset",        "notes": "Rails: replaces the schema, wiping rows" },
+      { "pattern": "manage.py flush",                "blast_radius": "HIGH",        "category": "app-reset",        "notes": "Django: removes all data from the database" },
+      { "pattern": "prisma migrate reset",           "blast_radius": "HIGH",        "category": "app-reset",        "notes": "Prisma: drops schema and reruns migrations" },
+      { "pattern": "sequelize db:drop",              "blast_radius": "HIGH",        "category": "app-reset",        "notes": "Sequelize CLI: drops the database" }
+    ],
+    "key-value-store": [
+      { "pattern": "redis-cli flushall",             "blast_radius": "HIGH",        "category": "kv-flush",         "notes": "removes every key from every Redis database" },
```
Sayt-0: Connection flags break contiguity, and they are the norm for any non-local Redis. Verified: `redis-cli -h prod.example.com -p 6379 FLUSHALL` and `redis-cli -n 2 flushdb` both classify unknown. `redis-cli ... flushall` / `redis-cli ... flushdb` would cover these.

pkg/safety/safety_patterns.json
```diff
@@ -115,6 +143,32 @@
...
+      { "pattern": "drop table if exists <name>",    "blast_radius": "HIGH",        "category": "sql-drop",         "notes": "idempotent form" },
+      { "pattern": "truncate table <name>",          "blast_radius": "MEDIUM-HIGH", "category": "sql-truncate",     "notes": "removes all rows; typically non-transactional" },
+      { "pattern": "delete from <table>",            "blast_radius": "MEDIUM",      "category": "sql-delete",       "notes": "unqualified DELETE would remove every row; presence of WHERE cannot be checked from the command shape" }
+    ],
+    "app-cli": [
+      { "pattern": "n8n db:reset",                   "blast_radius": "HIGH",        "category": "app-reset",        "notes": "wipes the n8n instance database" },
+      { "pattern": "n8n user-management:reset",      "blast_radius": "MEDIUM",      "category": "app-reset",        "notes": "resets the n8n owner account and API keys" },
+      { "pattern": "bench new-site --force <site>",  "blast_radius": "HIGH",        "category": "app-reset",        "notes": "Frappe/ERPNext: --force overwrites an existing site's database in place" },
```
Sayt-0: The Frappe CLI accepts the flag on either side of the site name, and only `--force <site>` is covered. Verified: `bench new-site site1.local --force` classifies unknown. Given this PR's flag-order theme, `bench new-site ... --force` (wildcard) or the second ordering would close it.

pkg/safety/classify.go
```diff
@@ -300,9 +314,21 @@ func stringSlice(value any) []string {
...
+// `&`, `;`) is anchored on whitespace rather than a word boundary:
+// `>` between two word characters isn't a shell redirect, so a
+// redacted placeholder like `<EMAIL>` or an inline comparison like
+// `"1 > 0"` must not match the `> <file>` truncate pattern. The
+// `\b` prefix would have fired on either.
 func patternToRegexp(pattern string) string {
 	var b strings.Builder
-	b.WriteString(`(?i)(?:^|.*\b)`)
+	b.WriteString(`(?i)`)
+	if pattern != "" && isShellMetacharAnchor(pattern[0]) {
```
Sayt-0, line 330: The fix is correct for the reported false positive (verified: the `--email <EMAIL>` command was destructive/medium on master, unknown here). One side effect worth recording: fd-number redirects lose their label, since the byte before `>` is a digit, not whitespace.

| command | master | this PR |
|---|---|---|
| `echo hi > /etc/passwd` | destructive / medium | destructive / medium |
| `run 1> /tmp/out.log` | destructive / medium | unknown |
| `run 2> /dev/null` | destructive / medium (FP) | unknown |

Gating is identical for destructive and unknown, and the `2> /dev/null` false positive disappears, so the trade looks acceptable. A short note in the function comment (or a `1> <file>` pattern entry, which would take the `\b` anchor since it starts with a digit) would make the choice explicit.

pkg/safety/safety_test.go
```diff
@@ -124,6 +152,34 @@ func TestClassifyCommand_DenyFlagsAreNeverSafe(t *testing.T) {
...
 }
 
+// The truncate-redirect pattern `> <file>` must not fire when `>` sits
+// between two word characters — it isn't a shell redirect there, just a
+// placeholder or comparison operator. Anchoring on `\b` alone was too
+// permissive.
+func TestClassifyCommand_TruncateRedirectRequiresWhitespaceAnchor(t *testing.T) {
+	notDestructive := []string{
+		`docker exec n8n n8n user:create --email <EMAIL> --firstName Admin`,
+		`git commit -m "feat: 1>0 check"`,
```
Sayt-0, line 191: This case never matched on master: the truncate pattern compiles to `> \S+` with a literal space, and `1>0` has none. The test passes on both sides of the change, so it guards nothing specific to this fix. Fine to keep as a general guard, but the PR description's claim that it "matched via `1>0`" should be corrected; only the `<EMAIL>` placeholder case reproduces.

Sayt-0 (changes requested):
Solid direction: each gap is real, the commits are atomic, and the whitespace-anchor fix removes a confirmed false positive. Two behavior changes need attention before merge, and roughly half of the new pattern entries are already covered by existing ones.

**Verified behavior deltas (master vs this branch):**

| command | master | this PR | `restricted` mode impact |
|---|---|---|---|
| `grep "drop table users" -r .` | safe | destructive / high | ALLOW to DENY |
| `git log --grep "delete from cart"` | safe | destructive / medium | ALLOW to DENY |
| `rg "DELETE FROM users"` | safe | destructive / medium | ALLOW to DENY |
| `run 1> /tmp/out.log` | destructive / medium | unknown | none (ask either way) |
| `docker exec ... --email <EMAIL> ...` | destructive / medium (FP) | unknown | none |

**Blocking:**
- The unanchored SQL patterns flip safe-listed read-only searches to destructive (table above). Since destructive matching wins over the safe list, any `grep`/`rg`/`git log --grep` whose argument contains SQL text now prompts under `balanced` and is denied under `restricted`. This trade-off may be acceptable (precedent: `grep "rm foo"` already matches `rm <single-file>`), but it is a silent behavior change for a common agent workflow and deserves an explicit decision, a note in the taxonomy, and regression tests either way.

**Non-blocking but worth fixing in the same pass:**
- 6 of the 9 new docker entries are already covered by existing patterns (details inline); the pattern language's `...` wildcard can replace the enumeration entirely.
- `Remove-Item "C:\Program Files\..." -Recurse -Force` classifies LOW because `<path>` compiles to `\S+` and cannot span a quoted path with spaces (details inline).
- `redis-cli -h host FLUSHALL` and `bench new-site <site> --force` fall back to unknown (details inline).

**Notes:**
- Two claims in the PR description do not reproduce on master: `git commit -m "feat: 1>0 check"` never matched (the truncate pattern requires a space after `>`), and `docker compose down -v --remove-orphans` was already HIGH via `docker compose down -v` (the compiled suffix `(?:$|\b.*)` accepts trailing content). Worth correcting so the taxonomy history stays accurate.
- The `\s` anchor drops fd-number redirects (`1> file`, `2> file`) from the truncate pattern. Gating is unchanged (destructive and unknown gate identically), and the `2> /dev/null` false positive disappears, so this looks like an acceptable trade; a sentence in the commit message or pattern notes would make it deliberate.

pkg/safety/safety_patterns.json
trungutt: Went with option 1 + option 2 in 747151a: SQL patterns are now anchored on `mysql` / `mariadb` / `psql` / `sqlite3` via `<client> ...<verb>` wildcards. `grep`/`rg`/`git log --grep` searches for SQL text stay safe. Wrapper commands (`docker exec`, `kubectl exec`, `ssh`) still match because the classifier's leading `.*\b` sees the client further down the command.

The residual FP — searching for a string that itself contains a client name (`grep "mysql -e 'drop database ...'" file`) — flips to destructive. Documented in `TestClassifyCommand_SqlPatternsDoNotShadowTextSearch` as the accepted trade.

pkg/safety/safety_patterns.json
trungutt: Fixed in 747151a. Collapsed the 14 `Remove-Item` entries to two wildcard forms — `Remove-Item ...-Recurse` (HIGH) and `Remove-Item ...-Force` (MEDIUM) — plus a positional `Remove-Item <path>` base case. The wildcard spans quoted paths with spaces, so `Remove-Item "C:\\Program Files\\..." -Recurse -Force` now classifies HIGH. Added regression tests for `C:\\Program Files\\...` and `OneDrive - Corp\\...` shapes.

pkg/safety/safety_patterns.json
trungutt: Fixed in 747151a. Dropped the six redundant entries (`--volumes -f/-a/-af/-fa`, `-a --volumes`, `-v --remove-orphans`) and replaced the genuine gaps with wildcards: `docker system prune ...--volumes`, `docker compose down ...-v`, `docker compose down ...--volumes`. Future flags between `prune`/`down` and the destructive flag inherit automatically.

pkg/safety/safety_patterns.json
trungutt: Fixed in 747151a. `redis-cli ...flushall` / `redis-cli ...flushdb` — the wildcard spans `-h` / `-p` / `-n`. Regression tests added for `redis-cli -h prod.example.com -p 6379 FLUSHALL` and `redis-cli -n 2 flushdb`.

pkg/safety/safety_patterns.json
trungutt: Fixed in 747151a. `bench new-site ...--force` — matches whether `--force` precedes or follows the site name. Regression tests cover both orderings.

pkg/safety/classify.go
trungutt: Noted in b876c19 — extended the `patternToRegexp` comment to record that fd-number redirects (`1>`, `2>`) no longer match, the `2> /dev/null` FP goes away as a bonus, and gating is unaffected because destructive and unknown gate identically.

pkg/safety/safety_test.go
trungutt: Corrected the PR body — you're right, the truncate pattern needs a literal space after `>`, so only the `<EMAIL>` placeholder case actually reproduces. Kept the `1>0` test as a general guard since it costs nothing.

melmennaoui (approved):

Sayt-0 (approved):

PR #4093, opened 2026-09-01
feat: safely gate embedded skill commands

dgageot:
Ask permission to the user when running commands embedded in skills

Assisted-By: docker-agent


trungutt (approved):
Cool

PR #4105, opened 2026-09-01
refactor: update go version declaration format

aheritier:
Update the go.mod file to use the modern Go version
declaration format. The go directive now specifies the
minimum version (1.27) while the toolchain directive
explicitly pins the exact toolchain version (1.27.0).
This separation allows better control over version
management and aligns with Go 1.21+ conventions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/docker/docker-agent
 
-go 1.27.0
+go 1.27
+
+toolchain go1.27.0
 
 require (
 	charm.land/bubbles/v2 v2.2.1


aheritier (commented):
## Review: no functional change, and the stated rationale is technically incorrect

CI is green for `f409275` (10 successful check runs — build-and-test, lint, windows-tests, build-image linux/amd64 + linux/arm64, license-check, CodeQL + Analyze go/actions/js; 5 skipped publish jobs; 0 failing, 0 pending). Commit is signed and verified. No merge conflicts (`mergeable: MERGEABLE`; `BLOCKED` is just the required-review gate).

The diff is functionally inert, and the three claims in the commit message are wrong. Verified locally against the PR head with go1.27.0:

**[blocking] The commit message claims the `toolchain` directive "pins the exact toolchain version". It does not — it is a minimum, not a pin.**
Reproduced: a module with `go 1.25` / `toolchain go1.25.0` under `GOTOOLCHAIN=auto` with a local go1.27.0 runs **go1.27.0**, not go1.25.0. Real pinning requires `GOTOOLCHAIN=go1.27.0` (env or `go env -w`), or the existing `Dockerfile:3` `ARG GO_VERSION="1.27.0"`. As written, the message records a reproducibility guarantee that the repo does not actually have.

**[blocking] "modern Go version declaration format ... aligns with Go 1.21+ conventions" is backwards.**
Patch-precise `go 1.27.0` is the form Go 1.21+ introduced, and it is what Go's own tooling emits today: `go mod init` under go1.27.0 writes `go 1.27.0`. The bare `go 1.27` is the pre-1.21 style. This PR moves *away* from the current convention, and away from the deliberate choice made in 2bb1e4b26 ("chore: upgrade to Go 1.27.0"), which bumped go.mod and `Dockerfile` together to `1.27.0`.

**[blocking] "This separation allows better control over version management" — the new directive is a semantic no-op.**
`go 1.27` already implies a go1.27.0 toolchain floor, so `toolchain go1.27.0` adds nothing. Go's tooling agrees it is redundant: with `go 1.27.0` + `toolchain go1.27.0`, `go mod tidy` **deletes** the toolchain line. This repo has already been through that once — 75ec0d7dc added `toolchain go1.24.5` and the same-day `go mod tidy` commit removed it.

**[should-fix] Net effect is churn that de-syncs the two Go-version sources.**
`go.mod:3` drops to `1.27` while `Dockerfile:3` stays `1.27.0`, so the patch precision that CI and the image build previously shared via `go-version-file: go.mod` (ci.yml:27/53/88/111, codeql.yml:43, update-models.yml:26) now lives in only one of them. Nothing breaks today (`go mod tidy --diff` is clean, lint/build/test green), but the next `go get -u`/tidy pass can rewrite these two lines inconsistently.

**Suggestion:** unless there is a requirement I am missing, close this and keep `go 1.27.0`. If the goal is genuinely reproducible toolchain selection, set `GOTOOLCHAIN=go1.27.0` in CI / `Taskfile.yml` instead. If the PR is kept for another reason, please rewrite the commit message — the current one would mislead the next reader.

dgageot (approved):

PR #4074, opened 2026-08-28
fix(#4060): extend backoff gate to remote MCP HTTP errors

aheritier:
> 🤖 **Automated implementer agent** — *this comment was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

Stacked on PR #4062. Base retargets to `main` once that merges.

Refs #4060 (partial — A2A pacing and LSP crash-loop pacing deferred; see below)

## What

Remote MCP servers responding with 503/429/5xx during the initialize handshake previously triggered a fresh connect attempt on every agent turn (the #4060 burst pattern). This PR fixes that by wrapping the HTTP status from the remote server in a `*modelerrors.StatusError` so the `StartableToolSet` backoff gate arming logic can pace retries — including a server-supplied `Retry-After` hint, and correctly for responses with no body.

## Design

**Wrap point — `enrichConnectError` in `pkg/tools/mcp/remote.go`.** The `oauthTransport` already records the last HTTP error status via `logErrorResponse` (for any `>= 400` response). `enrichConnectError` wraps that status (regardless of whether the response carried a body — many rate-limit/load-balancer responses don't) via `modelerrors.WrapHTTPError`, surfacing it as a `*StatusError` in the chain.

**Retry-After is honored.** `pkg/tools/mcp/oauth.go`'s `oauthTransport` now also captures the raw `Retry-After` header value alongside the status/body it already tracked. `lastServerErrorSnapshot()` reads status, message, and Retry-After together under a single lock (not three separate accessor calls) so a caller can never pair a status from one response with a Retry-After header captured from a different concurrent response on the same transport — this transport's `RoundTrip` can run concurrently for one logical connect attempt (e.g. a standalone SSE probe racing the initialize call). `enrichConnectError` builds a minimal `*http.Response` carrying that header and passes it to `WrapHTTPError`, matching the handling already in place for model-provider adapters (PR #4062).

**What arms the gate.** `startBackoffRetryable` checks for a `*modelerrors.StatusError` with a retryable HTTP status (429/408/5xx) via `errors.As` — exactly as it already does for RAG embedding failures. No regex heuristics; no new classification logic in the gate itself.

**What does NOT arm (unchanged policy):**
- Local stdio MCP spawn failures (missing binary, connection refused) — never reach `enrichConnectError`.
- 4xx client errors (400/401/403) — wrapped in `*StatusError` for structured access but `RetryableHTTPStatus` returns false → fail promptly.
[...]

PR #4108, opened 2026-09-01
fix(#4060): wire LSP crash-loop pacing through backoff gate

### LINKED ISSUE #4099: LSP crash-loop failures not paced by StartableToolSet backoff gate (#4060 follow-up)
## Context
PR #4074 extended the `StartableToolSet` backoff gate to remote MCP HTTP errors and explicitly documented (in code comments and `docs/tools/lsp/index.md`) that LSP crash-loop pacing is not yet wired.

## Problem
`lifecycle.ErrServerCrashed` — the sentinel produced when an LSP server process crashes repeatedly — is only produced inside the supervisor's internal watcher goroutine, and is never surfaced through `supervisor.Start()`'s return value. Because of this, `startBackoffRetryable` never sees it, and a persistently-crashing LSP server gets relaunched at full speed on every turn rather than being paced by the backoff gate.

## Acceptance criteria
- [ ] Propagate `lifecycle.ErrServerCrashed` (or an equivalent classification) from the supervisor's watcher goroutine up through `supervisor.Start()`'s error return, so it reaches `StartableToolSet.startLocked`
- [ ] Decide whether `ErrServerCrashed` should arm the gate directly, or whether it needs translation into a `*modelerrors.StatusError`-compatible signal via the classifier
- [ ] Add focused tests: a simulated LSP crash-loop arms the backoff gate on the next turn; a clean LSP shutdown/non-crash failure does not
- [ ] Update `docs/tools/lsp/index.md` to remove the "not yet wired" caveat once implemented

aheritier:
🤖 **Automated implementer agent** — *this PR was opened by the implementer bot from Docker Agentic Platform, not by a human developer*

Fixes #4099 (follow-up to #4060/#4062/#4074): wires LSP crash-loop failures through the `StartableToolSet` backoff gate. `lifecycle.ErrServerCrashed` was only ever produced inside `Supervisor.watch()`'s background goroutine and never surfaced through `Start()`, so a persistently-crashing LSP server relaunched at full speed forever.

**Design (Option A, per issue): direct sentinel, not a synthetic `StatusError`.** The Supervisor now detects a sustained crash loop itself (`Policy.CrashLoop{Threshold, Window}`, default 3 crashes / 1 minute — a single crash or crashes spread wider than the window are unaffected, handled entirely by the existing restart policy) and reports a one-shot `lifecycle.ErrCrashLooping` through the *next* `Start()` call instead of reconnecting. `startBackoffRetryable` arms directly on `errors.Is(err, lifecycle.ErrCrashLooping)` — forcing this through an HTTP-status-shaped `StatusError` would be a lossy encoding of a non-HTTP condition.

The LSP `ToolSet` didn't implement `tools.StartReporter`, so `StartableToolSet` never revisited it after the first successful start; added `IsStarted() = !State.IsTerminal()` (deliberately looser than MCP's Ready/Degraded-only check, so an ordinary transient reconnect doesn't force a 35s `RestartAndWait` on every turn — only a full give-up gets the wrapper involved, which is exactly when the gate needs to pace).

**Review findings addressed** (full detail in the commit message):
- Fixed a `Stop()`-vs-crash-loop-detection race that could clobber `StateStopped` back to `StateFailed` (regression test included).
- Added `Supervisor.PendingCrashLoopError()` (non-consuming peek) so LSP's per-request lazy-start path (`ensureInitialized`) can't bypass the gate by consuming the one-shot report itself.
- **Known, documented, out-of-scope limitation:** a raw crash doesn't clear the handler's `initialized` fast-path flag (only a fresh `Connect`/`Close` does), so a tool call immediately after a crash can still slip past the check above in practice. Pre-existing, affects the ordinary exhausted-restart give-up identically, and deserves its own look at the crash-detection/session-teardown interaction — noted in the commit message rather than fixed here.
- Added a real-time test for the crash-loop window's pruning (crashes spread wider than the window never accumulate).

## Changed files

| File | Change |
|---|---|
| `pkg/tools/lifecycle/errors.go` | New `ErrCrashLooping` sentinel |
| `pkg/tools/lifecycle/supervisor.go` | `Policy.CrashLoop`, crash-time-window tracking, one-shot `Start()` report, `PendingCrashLoopError()` peek, `Stop()`/race hardening |
| `pkg/tools/lifecycle/supervisor_test.go` | Crash-loop detection, pruning, peek, and Stop-race regression tests |
| `pkg/tools/startable_backoff.go` | `startBackoffRetryable` arms on `ErrCrashLooping`; doc comment rewritten |
| `pkg/tools/startable_backoff_test.go` | Arm/non-arm gate tests |
| `pkg/tools/builtin/lsp/lsp.go` | `tools.StartReporter` (`IsStarted`), `ensureInitialized` consults `PendingCrashLoopError` |
| `pkg/tools/builtin/lsp/lsp_crashloop_test.go` | New — end-to-end tests via a self-re-exec fake LSP server subprocess |
| `docs/tools/lsp/index.md` | Replaced the "not currently paced" caveat with the new behaviour |

## Testing
- `task build`, `task lint` — clean.
- `task test` — passes; two unrelated pre-existing failures confirmed present on the base branch too and environmental to this sandbox: `pkg/rag/treesitter` (needs `CGO_ENABLED=1`/gcc, unavailable here) and `pkg/cache.TestFileCache_dedupSkipsRedundantWrite` (filesystem mtime-resolution flake).
- Reviewed by an independent reviewer sub-agent (two passes); all findings from the first pass addressed and confirmed resolved in the second, verdict `approve`.

Stacked on `fix/startable-toolset-backoff-mcp-lsp` (#4074, which includes #4062) — targeting that branch, not `main`.

### DIFF
--- a/docs/tools/lsp/index.md
+++ b/docs/tools/lsp/index.md
@@ -197,7 +197,7 @@ Available Capabilities:
 
 LSP toolsets are managed by the same supervisor as MCP toolsets, so a crashed `gopls` (or any other language server) is reconnected automatically with exponential backoff. Use the [`lifecycle`](../../configuration/tools/index.md#toolset-lifecycle) block to tune the policy per toolset — for example, mark `gopls` as `strict` if your CI flow requires it to be available, or use `/toolset-restart gopls` from the TUI to force a reconnect when the server gets stuck.
 
-**Startup failure behaviour:** missing-binary and bad-config failures fail fast — each turn retries immediately with no artificial delay. A language server that crash-loops is not currently paced by the backoff gate; the supervisor's own reconnect policy (controlled by the `lifecycle` block) is the primary throttle for crash recovery.
+**Startup failure behaviour:** missing-binary and bad-config failures fail fast — each turn retries immediately with no artificial delay. A language server that crash-loops (3 crashes within 1 minute, by default) is different: the supervisor stops auto-restarting and reports the loop instead. The next attempt to use the toolset (a turn's start, a tool call, or `/toolset-restart`) surfaces that report rather than relaunching the server, and the backoff gate then paces subsequent attempts (15s, doubling up to 5 minutes) the same way it paces a rate-limited MCP server — so a server that dies right after every restart no longer relaunches at full speed. This applies regardless of `profile`: even a `strict` toolset is retried on the next turn once its window elapses, rather than staying down until an explicit `/toolset-restart`.
 
 ```yaml
 toolsets:
--- a/pkg/tools/builtin/lsp/lsp.go
+++ b/pkg/tools/builtin/lsp/lsp.go
@@ -60,9 +60,10 @@ type ToolSet struct {
 
 // Verify interface compliance
 var (
-	_ tools.ToolSet      = (*ToolSet)(nil)
-	_ tools.Startable    = (*ToolSet)(nil)
-	_ tools.Instructable = (*ToolSet)(nil)
+	_ tools.ToolSet       = (*ToolSet)(nil)
+	_ tools.Startable     = (*ToolSet)(nil)
+	_ tools.Instructable  = (*ToolSet)(nil)
+	_ tools.StartReporter = (*ToolSet)(nil)
 )
 
 type lspHandler struct {
@@ -431,6 +432,20 @@ func (t *ToolSet) Stop(ctx context.Context) error {
 	return t.handler.supervisor.Stop(ctx)
 }
 
+// IsStarted implements tools.StartReporter: reports whether the supervisor
+// requires external action (Start/Restart) to serve requests again.
+// Deliberately looser than the MCP toolset's IsStarted (which tracks
+// Ready/Degraded only): a transient Restarting still reports true here,
+// since the supervisor already self-heals a one-off crash on its own
+// watcher goroutine and every per-request call (ensureInitialized) retries
+// eagerly regardless. Only a give-up — Failed (crash loop, exhausted
+// restarts) or Stopped — reports false, so StartableToolSet gets involved
+// (via Restart) exactly when the supervisor needs a caller-paced retry, not
+// on every ordinary transient reconnect.
+func (t *ToolSet) IsStarted() bool {
+	return !t.handler.supervisor.State().State.IsTerminal()
+}
+
 // State returns a snapshot of the underlying supervisor's lifecycle state,
 // suitable for the /tools dialog and lifecycle log messages.
 func (t *ToolSet) State() lifecycle.StateInfo {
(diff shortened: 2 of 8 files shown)


aheritier (commented):
🤖 **Automated implementer agent** — *this review was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

Self-review of record for #4099 (crash-loop pacing), following the same pattern used on #4062/#4074/#4104. This PR was independently reviewed twice by an automated reviewer agent before push (not GitHub-visible, since it runs in-sandbox); recording that review here since self-approval is blocked for the PR author.

## Findings and resolutions (all addressed pre-push, single squashed commit `30ec91e6`)

1. **`Stop()` vs. crash-loop-detection race** (`pkg/tools/lifecycle/supervisor.go`, `watch()`): a concurrent `Stop()` winning the race against a crash-loop declaration could leave `State` as `Failed` instead of `Stopped`. **Fixed**: the crash-loop branch re-checks `s.stopping` after re-acquiring `s.mu`, bailing out without overwriting state if `Stop` already won. Regression test `TestSupervisor_CrashLoopStopWinningRaceReportsStopped` — verified empirically to fail ~100% of the time with the fix reverted, and pass reliably with it restored.
2. **Gate bypass via `ensureInitialized`** (`pkg/tools/builtin/lsp/lsp.go`): the LSP handler's lazy-start path called `supervisor.Start()` directly, which would consume the one-shot `ErrCrashLooping` report outside `StartableToolSet`'s gate. **Fixed**: added `Supervisor.PendingCrashLoopError()` (non-consuming peek); `ensureInitialized` checks it first. Test: `TestLSPTool_CrashLoopPendingReportBlocksDirectToolCalls`. **Known, documented, out-of-scope residual limitation** (noted in the commit message): a raw crash doesn't clear `lspHandler.initialized`, so a tool call immediately following a crash can still occasionally hit the stale fast-path before the new check — pre-existing behavior, identically affects the ordinary exhausted-restart give-up path, not introduced by this change.
3. **Missing crash-loop window-pruning test**: added `TestSupervisor_CrashLoopWindowPrunesOldCrashes`, exercising real time via a short `Window` + `time.Sleep` (matches this test file's existing convention; `//nolint:forbidigo` with reason, no injectable clock added).
4. **Loose spawn-count assertion**: tightened `GreaterOrEqual` to exact `assert.Equal(t, 3, spawnsAtLoop, ...)` in `TestLSPTool_CrashLoopArmsBackoffGate`; confirmed exact and non-flaky across repeated runs.
5. **Undocumented `strict`-profile behavior change**: `docs/tools/lsp/index.md` updated to note a `strict`-profile LSP toolset is now retried on the next turn after a full give-up (consistent with MCP's pre-existing behavior), instead of staying down until `/toolset-restart`.

## Accepted as out-of-scope (not fixed, flagged for follow-up)

- A narrower pre-existing race between computing `crashLooping` and recording `crashLoopErr` that could orphan a session with no watcher — reproducible identically on pre-existing non-crash-loop `Failed` paths with no `CrashLoop` policy involved at all; not introduced or worsened by this diff.
- MCP toolsets silently inherit the same crash-loop protection (shared `Supervisor`), with no MCP-specific test/doc coverage and no YAML knob to tune `Threshold`/`Window`. Likely-intentional generalization consistent with `Supervisor`'s shared-implementation design; candidate for a follow-up ticket rather than expanding this PR's scope.
- An adjacent, unchanged `tracker.Fail(StateRestarting, waitErr)` line sits next to the new code with the same un-guarded-against-`Stop()` race class as finding 1, but predates this diff (present since before #4060) and is not part of the new crash-loop path.
- Base branch (`fix/startable-toolset-backoff-mcp-lsp`) had moved forward ~24 commits after this branch was first cut; addressed by rebasing onto the new tip.

## Verification
- `task build`, `task lint` (golangci-lint + `go run ./lint .` + `go mod tidy --diff`) — clean, 0 issues.
- `task test` — full suite passes; two pre-existing, unrelated, environmental failures confirmed present identically on the base branch and unaffected by this diff: `pkg/rag/treesitter` (needs `CGO_ENABLED=1`/gcc, unavailable in this sandbox) and `pkg/cache.TestFileCache_dedupSkipsRedundantWrite` (filesystem mtime-resolution flake).
- CI on this PR: 16 checks pass, 4 skip (image-push jobs gated on merge), 0 fail.

No blocking issues remain. Approving this changeset for merge (recorded here as a comment since GitHub blocks self-approval).

dgageot (approved):

PR #4110, opened 2026-09-02
ci(update-models): sign commits with bot identity, add semantic catalog diff

aheritier:

Two independent improvements to the weekly `update-models.yml` refresh (root-cause analysis and the full three-phase design were scoped directly with the requester in a design discussion prior to this PR; there is no filed GitHub issue to link). **Phase 3 (AI-generated PR summary via `docker/docker-agent-action`) is explicitly deferred to a follow-up PR** — no agent/LLM step, no `.github/agents/*.yaml`, no new secrets here.

## Signed commits
- `Create pull request` (`peter-evans/create-pull-request`) now passes `sign-commits: true` with `token: ${{ github.token }}` — this builds the commit through GitHub's API instead of a local `git push`, so it's **Verified** regardless of which token creates it. On the default `GITHUB_TOKEN`, the commit author/committer is consistently `github-actions[bot]`, rather than whoever happened to trigger the run.
- No App-token dependency: an earlier iteration of this PR used `docker/docker-agent-action/setup-credentials` to mint a `docker-read-write[bot]` App token, buying bot branding and CI auto-triggering on the created PR. That was deliberately dropped — not worth the added dependency for those two benefits alone. Accepted trade-off: commits show as `github-actions[bot]`, and the pre-existing weekly-manual-CI-rerun tax (GitHub's recursion guard doesn't fire workflows off PRs created by `GITHUB_TOKEN`) returns.
- Job permissions are `contents: write` + `pull-requests: write` (needed for the default `GITHUB_TOKEN` to do the signed-commit-via-API + PR creation itself). `persist-credentials: false` on checkout. A `concurrency` group and a canonical-repo job guard prevent an overlapping `workflow_dispatch` from racing the weekly schedule.
- The `on:` block (`workflow_dispatch` + weekly `schedule`) is untouched — byte-identical to `main`.

## Semantic catalog diff, no AI
- `.github/scripts/models-delta.jq` — structural added/removed/changed diff between two snapshot.json files.
- `.github/scripts/models-delta-render.jq` — renders that diff to markdown tables, with a `safe()`/`safe_diff()` sanitiser clamping every third-party (models.dev) string to a short, bounded, control-character-free charset before it reaches the PR body, since that body flows into this repo's automated PR-review pipeline.
- `scripts/models-delta.sh` — the single caller of both scripts; `task models-delta` (Taskfile) and the workflow both shell out to it, so the jq logic exists in exactly one place. With no args it diffs HEAD's `snapshot.json` against the working tree (what a local `task update-models` would change); `--json-out FILE` also emits the structural JSON.
- The workflow runs the diff after `task update-models` and uses the rendered markdown as the PR body/title/commit message. The delta step is `continue-on-error: true` with a static-body fallback, so a jq/shell failure degrades the PR text instead of silently skipping the weekly refresh.
- A refresh still opens a PR even when only `snapshot_date.txt` changed (no catalog content diff) — conservative choice, preserves today's behavior and doesn't interact with `check-models-fresh`/`TestSnapshotDateIsFresh`.
- `scripts/models-delta-test.sh` is a golden-fixture regression test for the two jq scripts. `.github/workflows/models-delta-lint.yml` runs it (plus `shellcheck`) as its own path-filtered workflow — mirrors this repo's existing `docs-lint.yml` pattern — rather than unconditionally on every PR via `ci.yml`'s blanket lint job.

## Testing
- `task build`, `task test`, `task lint` all pass (one pre-existing, environment-flaky `pkg/cache` test unrelated to this diff, reproduced identically on unmodified `main`).
- `./scripts/workflow-lint.sh`, `actionlint -shellcheck=shellcheck`, and `shellcheck scripts/models-delta.sh scripts/models-delta-test.sh` all clean.
- Self-diff of the real `pkg/modelsdev/snapshot.json` via `task models-delta` yields empty added/removed/changed.
- Reviewed by the `reviewer` sub-agent across many passes (verdict: approve) plus multiple rounds of human security review — see PR comments for the full history of findings addressed, including the deliberate App-token removal and the CI lint path-scoping.

Branch history was squashed into a single commit per request; the incremental "address review round N" commits are not individually preserved.


trungutt (dismissed):

aheritier (commented):
Security-focused review at `6a7150da`. CI is green on all required checks (`build-and-test`, `license-check`, `lint`, plus CodeQL incl. `Analyze (actions)`). Commenting rather than approving: there is no linked issue/discussion on this PR (`closingIssuesReferences` and the cross-reference timeline are both empty) although the description points at one, so the Phase 1/Phase 2 acceptance criteria can't be checked against a source of truth.

**Verified good:** `on:` is `workflow_dispatch` + `schedule` only and byte-identical to `main` — no fork code is ever checked out or run with these credentials. No script injection: the new `run:` block uses only `$RUNNER_TEMP`/`$GITHUB_OUTPUT`, `steps.delta.outputs.subject` is built solely from `jq length` integers, and the untrusted body is passed via `body-path` (file), not inline. All actions SHA-pinned, and `06e1767a…` does resolve to `v2.0.5` — the exact SHA `pr-review.yml` already trusts. `docker-read-write[bot]` is a real App already authoring PRs here (#4101), and a non-`GITHUB_TOKEN` credential does make CI trigger on the bot PR.

**[should-fix] Secrets sit in the job env while `go generate` runs** — `.github/workflows/update-models.yml:21-22`. `Setup credentials` is the *first* step and `exportVariable`s `GITHUB_APP_TOKEN` (a *stored* `pat` from AWS secret `docker-agent-action/github-app`, `src/credentials/github-app.ts:52` — not a per-run minted token, not repo/permission-scopable, not revocable at job end), `ORG_MEMBERSHIP_TOKEN`, and `ANTHROPIC_API_KEY_FROM_SSM`/`OPENAI_API_KEY_FROM_SSM` (`src/credentials/ai-keys.ts:40-45`) — AI billing keys unrelated to a snapshot refresh. All readable by every later step, including `setup-go` module downloads and `task update-models` → `go generate ./pkg/modelsdev/...`. Masking protects logs, not exfiltration. One-line fix with no behaviour change: move `Setup credentials` down to immediately before `Create pull request`. Longer term, ask `docker-agent-action` for opt-in inputs so callers can decline the AI keys / org token.

**[should-fix] Ambient `GITHUB_TOKEN` stays write-capable and persisted** — `update-models.yml:17-18`, `:24-25`. With `token: ${{ env.GITHUB_APP_TOKEN }}` + `sign-commits: true`, all writes go through the App token via the Git Database API, so `contents: write` / `pull-requests: write` are no longer needed, and `actions/checkout` still defaults to `persist-credentials: true`, leaving a write token in `.git/config` during `go generate`. Suggest `contents: read` + `id-token: write` only, drop `pull-requests: write`, add `persist-credentials: false`, then confirm with one `workflow_dispatch`.

**[should-fix] A delta failure silently kills the weekly refresh** — `update-models.yml:41-45`. Under `set -euo pipefail`, any jq/shell error means *no PR at all*, and the only symptom is a snapshot that quietly stops refreshing. Reachable: `esc_cell`/`esc_code` `gsub` whatever models.dev returns, so a non-string `name` aborts the run. Suggest `continue-on-error: true` on the delta step with a fallback to the old static subject/body when `steps.delta.outcome != 'success'`, plus `tostring` guards in the jq.

**[should-fix] No automated coverage for the new shell/jq** — `lint` runs `actionlint` + `workflow-lint.sh`, which cover workflow `run:` blocks but not `scripts/*.sh`, and nothing validates the jq (`ci.yml:30-42`). This code first executes for real at 06:00 Monday. Suggest two small snapshot fixtures with golden JSON + golden markdown wired into `task test`, and `shellcheck scripts/*.sh` in the lint job.

**[should-fix] Untrusted models.dev text reaches the PR body consumed by the AI review pipeline** — `models-delta-render.jq:28-32`. Path: `models.dev/api.json` → `snapshot.json` → delta → `pr-body.md` → bot PR → `pr-review-trigger.yml` (`pull_request: opened`) → `review-pr.yml` with `pull-requests: write`, `issues: write`, `actions: write`. Sanitization is only `|` and backticks; newlines, control chars, HTML and unbounded length pass through. Cheap hardening: `def safe: tostring | gsub("[^A-Za-z0-9 ._/+:-]"; "?") | .[0:80];` on every rendered id/name/value.

**[optional]** No `concurrency:` group — a manual dispatch overlapping the Monday schedule races two runs on `chore/update-models-snapshot` with a write-capable token. **[optional]** `models-delta-render.jq:32`: `.id` gets `esc_cell|esc_code`, `.model.name` only `esc_cell`, so a backtick in a name still opens an unterminated code span. **[optional]** Consider `dist/signed-commit.js` from `docker-agent-action` (reachable via `$DOCKER_AGENT_ACTION_ROOT`, exported by `setup-credentials`) so the stored App credential isn't handed to a third-party action — trade-off is that it only creates the commit, so branch reuse/PR update/`delete-branch` would move to `gh pr create`. Keeping SHA-pinned `peter-evans` is defensible; just make it explicit. **[optional]** The description says setup-credentials "mint[s]" an App token — it fetches a stored `pat` and also injects `ORG_MEMBERSHIP_TOKEN` + AI keys; worth correcting so reviewers reason about the right credential model. `task models-delta` is also missing from `AGENTS.md`'s command list.

Finally: "Verified commit" and "CI starts without a manual re-run" are unverifiable until this runs — worth one `workflow_dispatch` before merge to confirm the signature badge, the bot author, and the CI trigger.

aheritier (commented):
Re-review at `af34794` (previous pass: `6a7150da`). CI is green on this head — required checks `build-and-test`, `license-check`, `lint` all pass; no conflicts.

### Prior findings — all five should-fix items verified

| Finding | Status | Evidence |
|---|---|---|
| Secrets in job env during `go generate` | ✅ resolved | `Setup credentials` now at `update-models.yml:102-103`, immediately before `Create pull request` (`:105`); checkout/setup-go/`task update-models`/delta all run at `:24-100` |
| Over-broad job permissions | ✅ resolved | `:20-22` = `contents: read` + `id-token: write`; `persist-credentials: false` at `:26-27`. Verified the action still authenticates by itself: `create-pull-request.ts:98-101` calls `gitConfigHelper.configureToken(inputs.branchToken)` and `main.ts:47-48` defaults `branchToken`→`token`, so it installs its own `http.<origin>/.extraheader` — fetch/push are unaffected |
| Delta failure kills the weekly refresh | ✅ resolved | `continue-on-error: true` `:45`; fallback `:87-100`. Semantics check out: the guard uses `outcome` (correct under `continue-on-error`, where `conclusion` would always be `success`); the fallback **overwrites** `pr-body.md` with `>` (`:100`), so a partial body from a half-failed delta step can't survive; and a partially-set `steps.delta.outputs.subject` is never selected because the selection is outcome-gated (`:110-111`) |
| No coverage for the new shell/jq | ⚠️ partial | Fixtures + `models-delta-test.sh` wired into `ci.yml:44-47` — CI-enforced ✅. Not in `task test` (see below) |
| Untrusted models.dev text in the PR body | ✅ resolved | `safe` at `models-delta-render.jq:37-40` (`tostring` first, `[0:80]` cap), applied to `.id` (`:52`, `:65`), name (`:49`), and both `tojson` values (`:66`). Verified empirically: newline-in-name, numeric name, object-valued name, missing name, pipes and backticks all collapse without corrupting a table row; empty delta renders cleanly |

Optionals: concurrency group ✅ (`:6-8`, `cancel-in-progress: false` — right choice for a push-type job); `esc_code` on `.model.name` ✅ (`safe` runs first, so backticks can't survive); the peter-evans decision is documented and the pins verify (`22a9089…` = `v7.0.11` = `v7`, `06e1767a…` = `v2.0.5`, the same SHA `pr-review.yml:21` trusts).

On my earlier note that this adds the repo's first `workflow_dispatch`-reachable OIDC path — **withdrawn**: `ci.yml:176-181` (`build-and-push-image`) already has `id-token: write` under `workflow_dispatch` with only `if: github.event_name != 'pull_request' && !github.event.repository.fork`. Not something this PR introduces.

### New findings

**[should-fix] The golden test never exercises the `Removed` table.** `testdata/expected-delta.json:29` is `"removed": []`. Demonstrated: I replaced the entire `Removed` branch (`models-delta-render.jq:60-62`) with

```jq
+ (if (.removed|length) == 0 then "_none_\n" else
    "TOTALLY BROKEN REMOVED TABLE, UNSANITIZED\n"
    + ([.removed[] | .id] | join("\n")) + "\n" end)
```

— dropping the table header *and* all sanitisation — and `./scripts/models-delta-test.sh` still printed `models-delta-test.sh: OK` (exit 0). Adding one removed model to `old.json` closes the hole. The test is non-vacuous elsewhere (changing `safe`'s cap to `.[0:79]` does fail it), so this is specifically the removed-path.

**[should-fix] `task test` doesn't run the new test.** `Taskfile.yml:99-103` is `go test ./...`; the golden test only exists as `test-models-delta` (`:52`) and in `ci.yml:44-47`. But `AGENTS.md:42` ("All tests pass (`task test`)") and `:90` ("Run `task test` to execute the full test suite") tell both humans and agents that `task test` *is* the suite. `deps: [test-models-delta]` on `test` (or on `lint`) makes the docs true again.

**[should-fix] Dangling task name.** `scripts/models-delta-test.sh:9` says "run it directly or via `task test-scripts`" — no such task; it's `test-models-delta` (`Taskfile.yml:52`).

**[should-fix] "the linked issue/discussion" isn't linked.** The description points at a linked issue/discussion for the root cause and design, but the PR has no reference to one (no ref token in the body, no cross-reference events on the timeline). The Phase 1/2/3 rationale is unreachable from the PR — worth linking it or inlining the root cause, since "Phase 3 deferred" only makes sense with it.

**[optional] `.field` is the one un-`safe`d rendered leaf** (`models-delta-render.jq:66`). Not exploitable — those keys come from a typed Go struct (`pkg/modelsdev/types.go:20-24`), so they're a fixed set. Consistency only.

**[optional] `safe` is over-broad on `tojson` output** (`:66`). `tojson` already escapes control characters and quotes, so the only remaining markdown hazards are `` ` `` and `|`. A narrower variant for that path — ``gsub("[|`]"; "?") | .[0:80]`` — would render `{"input":5,"output":2}` instead of `??input?:5??output?:2?` at identical safety, which makes the trade-off documented at `:34-36` avoidable rather than inherent.

**[optional] `workflow_dispatch` bypasses the canonical-repo guard** (`update-models.yml:18`) for a job that now carries `id-token: write` (`:22`). It fails closed in a fork — `setup-credentials/action.yml:21-24` hard-errors when `GITHUB_APP_TOKEN` is unset — but the guard's intent reads better as `github.repository == 'docker/docker-agent'` unconditionally.

**[optional] Duplicated PR-body preamble** at `update-models.yml:60-61`, `:78-79`, `:95-96`.

**[optional / infra owner]** Independent of this PR: `workflow_dispatch` runs the workflow file from the chosen ref, so a write-access user can dispatch a modified `update-models.yml` from a branch and obtain the App token. Pre-existing (see `ci.yml` above), but worth confirming the AWS trust-policy `sub` condition — or gating on an `environment:` with a main-only deployment-branch policy.

### Also verified clean

No `${{ }}` inside any `run:` body (all four interpolations are `with:` inputs, `:108-112`) and the subject is built only from jq integer counts; `actionlint -shellcheck` and `shellcheck` both clean; `flatten_models` matches the real snapshot shape (top-level provider map, checked against `pkg/modelsdev/snapshot.json` and `snapshot.go:43`), so the "self-diff is empty" evidence isn't vacuous; `mtok`/`ctx` can't emit unsanitised strings because `Cost`/`Limit` are typed numerics (`types.go:42-51`), and a malformed value there would abort jq — which now fails open into the static body.

**No blocking findings.** The security posture of this workflow is now materially better than `main`'s. I can't approve my own PR, so this stays a comment.

.github/workflows/update-models.yml
```diff
@@ -11,14 +15,16 @@ on:
...
 jobs:
   refresh-snapshot:
-    if: github.event_name == 'workflow_dispatch' || github.repository == 'docker/docker-agent'
+    if: github.repository == 'docker/docker-agent'
     runs-on: ubuntu-latest
     permissions:
-      contents: write
-      pull-requests: write
+      contents: read
+      id-token: write # required: OIDC exchange for AWS Secrets Manager (setup-credentials)
```
aheritier:
```suggestion
      id-token: write # required
```

Taskfile.yml
```diff
@@ -87,6 +98,7 @@ tasks:
 
   test:
     desc: Run tests
+    deps: [test-models-delta]
```
aheritier:
```suggestion
```

trungutt (dismissed):

.github/workflows/ci.yml
```diff
@@ -41,6 +41,11 @@ jobs:
       - name: Lint GitHub Workflows
         run: ./scripts/workflow-lint.sh
 
+      - name: Lint models-delta scripts
+        run: |
+          shellcheck scripts/models-delta.sh scripts/models-delta-test.sh
+          ./scripts/models-delta-test.sh
+
```
aheritier: I don't think we should test this for every PR/commit

rumpl (approved):

PR #4145, opened 2026-09-03
fix: update stale Fireworks kimi-k2-instruct model reference

### LINKED ISSUE #4132: Fireworks default/example model reference (kimi-k2-instruct) doesn't exist in models.dev, silently skipped by CI
## What's wrong

`pkg/config/auto.go`'s `DefaultModels["fireworks"]` and `examples/fireworks.yaml` both reference:

```
accounts/fireworks/models/kimi-k2-instruct
```

This model ID does not exist in the `fireworks-ai` catalog on models.dev — and, as far as I can tell from both the current and previous embedded `pkg/modelsdev/snapshot.json`, it never has. models.dev's naming for this lineup uses a different scheme entirely (e.g. `accounts/fireworks/models/kimi-k2p6`, `.../kimi-k2p7-code`, `.../kimi-k3`).

Also referenced (with the same staleness) in:
- `pkg/config/auto_test.go` (two `assert.Equal` checks mirroring the value in `auto.go`, not validated independently)
- `docs/providers/fireworks/index.md` (`model:` snippets ×4, plus a stale "Available Models" table listing `accounts/fireworks/models/kimi-k2-instruct`, `.../llama-v3p3-70b-instruct`, `.../qwen3-235b-a22b` — none of which exist in the current `fireworks-ai` catalog; Llama has no successor there at all anymore)

## Why it's invisible to CI

`pkg/config/examples_test.go` maintains a `modelsDevAbsentProviders` map that skips models.dev catalog validation entirely for several providers, including `fireworks`, with the comment:

> `"fireworks": true, // models.dev catalogs Fireworks under the "fireworks-ai" id, not "fireworks"`

This is accurate — but the consequence is that **no test anywhere ever checks that the model string configured for the `fireworks` provider actually exists**, so this reference has silently drifted (or was simply wrong from the start) without any CI signal. The same blind spot applies to the other skipped providers (`together`, `moonshot`, `chatgpt`, `ovhcloud`, `vercel`, `opencode-zen`, `dmr`, `cloudflare-workers-ai`, `cloudflare-ai-gateway`), though this issue is specifically about the concrete Fireworks breakage found.

## Suggested fix direction

- Update the stale references to a model that currently exists in the `fireworks-ai` catalog (e.g. `accounts/fireworks/models/kimi-k3`) — tracked separately as part of the broader model-reference refresh for PR #4121.
- Longer-term: teach `examples_test.go` (or a dedicated test) to resolve the *actual* models.dev catalog id for these "absent" providers (e.g. `fireworks` → `fireworks-ai`, `together` → `togetherai`, `moonshot` → `moonshotai`) instead of skipping validation outright, so this class of staleness gets caught automatically going forward.

## References

- Found while auditing model references for docker/docker-agent#4121 (models.dev snapshot refresh).

aheritier:

Fixes #4132.

`accounts/fireworks/models/kimi-k2-instruct` does not exist in the `fireworks-ai` models.dev catalog (and never has). Replaces it with `accounts/fireworks/models/kimi-k3`, verified present in the current `pkg/modelsdev/snapshot.json`, across:

- `pkg/config/auto.go` (`DefaultModels["fireworks"]`)
- `examples/fireworks.yaml`
- `pkg/config/auto_test.go`, `pkg/model/provider/openai_alias_providers_test.go`, `pkg/model/provider/openai/system_message_merge_test.go`
- `docs/providers/fireworks/index.md` — 4 `model:` snippets, plus the "Available Models" table, which also listed two other nonexistent ids (`llama-v3p3-70b-instruct`, `qwen3-235b-a22b`); Llama has no successor in the `fireworks-ai` catalog at all, so it's dropped rather than remapped.

Also closes the CI blind spot the issue describes: `pkg/config/examples_test.go`'s `modelsDevAbsentProviders` unconditionally skipped models.dev validation for several providers whose docker-agent name diverges from their models.dev catalog id. Replaced with a `modelsDevCatalogProviders` map that resolves the real catalog id (`fireworks`→`fireworks-ai`, `together`→`togetherai`, `moonshot`→`moonshotai`, `chatgpt`→`openai`, `opencode-zen`→`opencode`) before validating — this is exactly the check that would have caught #4132 automatically. `vercel` now validates directly (models.dev catalogs it as-is). `dmr`, `ovhcloud`, `cloudflare-workers-ai`, `cloudflare-ai-gateway` remain skipped with corrected comments (e.g. `ovhcloud`'s example id is case-mismatched against models.dev's lower-cased ids, not "not yet catalogued").

**Known, deliberately out-of-scope follow-up:** `DefaultModels` itself isn't validated against models.dev for every provider (e.g. `github-copilot`'s default `gpt-5.6` isn't in that provider's catalog, which only has `gpt-5.6-luna/-sol/-terra`). Extending validation to `DefaultModels` would need separate model-choice review per provider, so left for a follow-up rather than bundled here.

**Testing:** `task build`, `task test` (green except the pre-existing, unrelated `pkg/rag/treesitter` CGO failures — no gcc in this sandbox), `task lint` (golangci-lint + `go run ./lint .` + `go mod tidy --diff`) all pass.

### DIFF
--- a/docs/providers/fireworks/index.md
+++ b/docs/providers/fireworks/index.md
@@ -11,7 +11,7 @@ _Use Fireworks AI models with Docker Agent._
 ## Overview
 
 [Fireworks AI](https://fireworks.ai/) is a fast inference host for open-weight
-models, serving Kimi K2, Llama, Qwen, DeepSeek, GLM and others through an
+models, serving Kimi, Qwen, DeepSeek, GLM and others through an
 OpenAI-compatible API. Docker Agent includes built-in support for Fireworks AI
 as an alias provider.
 
@@ -33,7 +33,7 @@ The simplest way to use Fireworks AI:
 ```yaml
 agents:
   root:
-    model: fireworks/accounts/fireworks/models/kimi-k2-instruct
+    model: fireworks/accounts/fireworks/models/kimi-k3
     description: Assistant using Fireworks AI
     instruction: You are a helpful assistant.
 ```
@@ -46,7 +46,7 @@ For more control over parameters:
 models:
   fireworks_model:
     provider: fireworks
-    model: accounts/fireworks/models/kimi-k2-instruct
+    model: accounts/fireworks/models/kimi-k3
     temperature: 0.7
     max_tokens: 8192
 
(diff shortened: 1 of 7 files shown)


trungutt (dismissed):
yes

dgageot (approved):

Issue #4106, opened 2026-09-01
Tool Schema Incorrectly Classified as Strict-Compatible


abirmingham:
### Description

When docker-agent imports MCP tools, it can classify valid MCP JSON Schemas as compatible with OpenAI strict function calling and send them with strict: true.

For example:

```
{
  "type": "object",
  "properties": {
    "value": {
      "oneOf": [
        { "type": "integer" },
        { "type": "string" }
      ]
    }
  }
}
```

Docker-agent’s compatibility check does not reject unsupported keywords such as oneOf and does not inspect definitions under $defs. The resulting schema is sent with `strict: true`, but OpenAI Structured Outputs supports only a subset of JSON Schema and rejects the entire request with HTTP 400 before any tool can be called.

### Expected Behavior

Schemas which do not meet the criteria of [Structured Outputs Supported Schemas](https://developers.openai.com/api/docs/guides/structured-outputs#supported-schemas) are sent with `strict: false`.

### Actual Behavior

Strict-incompatible schemas are sent to OpenAI's API with `strict: true`. The docker-agent criteria appears to be `pass strict:true if tool lacks 'additionalProperties: false' OR additionalProperties is a schema object`.

### Steps to Reproduce

1. Connect Docker Agent to an MCP server exposing a tool with this input schema:
```
{
  "type": "object",
  "properties": {
    "value": {
      "oneOf": [
        { "type": "integer" },
        { "type": "string" }
      ]
    }
  }
}
```
2. Use the OpenAI provider with a GPT-5 model.
3. Send any prompt to the agent.

### Docker Agent version

v1.130.0

### OS & terminal

Ubuntu 24.04.4 LTS, x86_64

### Model used

openai/gpt-5

### Error output

```shell
HTTP 400: Invalid schema for function 'search_devices'
```

### Additional context

This is a tough one, as it seems that neither OpenAI nor Anthropic provide any sort of tooling for MCP authors (nor harness authors) schemas as being strict-compatible. It's not clear to me what the best solution is here, short of docker-agent implementing the exact criteria specified by each model vendor before sending `strict: true`, or requiring the user to specify whether `strict: true` should be passed (and the user is unlikely to know). Worse, the criteria appears to vary by model vendor, which seems to be why pi-agent [opted against it](https://github.com/earendil-works/pi/issues/1086#issuecomment-3824818579).

abirmingham:
Adding a minimal stdio MCP server and an agent.yaml file to reproduce. See attached files.

[...]

PR #4161, opened 2026-09-04
fix(tools,openai): don't inject type:object onto anyOf nodes; handle Responses API response.incomplete/failed

simonferquel-clanker:
> 🤖 **Automated architect agent** — *this PR was opened by the architect bot from Docker Agentic Platform, not by a human engineer*

## Symptom
Any OpenAI model (`gpt-5.6-terra`/`sol`, `gpt-6-astra` via Responses) with the **Atlassian (Jira) or Notion remote MCP server** attached ends every turn in ~2 s with:

> Model openai/gpt-5.6-terra returned an empty response (stop reason: null)…

0 input / 0 output tokens, no error. Grafana MCP works; Anthropic works with all three — which hid the cause.

## Root cause (two bugs)
1. **`pkg/tools/schema.go` → `ensurePropertyTypes`** stamps `type: "object"` on every typeless property, including composition nodes. For `getConfluenceSpaces.expand` = `{"anyOf":[string, array<string>]}` the tool schema becomes `{"anyOf":[…], "type":["object","null"], "additionalProperties":false}` — unsatisfiable (`type` ANDs with `anyOf`). OpenAI's Responses API answers with an **instant `status=incomplete`, `incomplete_details.reason=max_output_tokens`, empty output, 0 tokens** — not a 400. The `isCompositionNode` guard added in #4106 lives downstream in the openai package and never gets the chance: the type is already injected.
2. **`response_stream.go`** has no case for `response.incomplete` / `response.failed`; both hit `default`, get logged at INFO and dropped → no usage, no finish reason → "stop reason: null".

## Fix
- `ensurePropertyTypes` skips `anyOf`/`oneOf`/`allOf` nodes (as it already does for `$ref`) and walks their variants.
- Responses adapter handles `response.incomplete` (usage + finish reason `length`, or `refusal` for `content_filter`; reason + `response_id` logged) and `response.failed` (returned as an error with code/message/response_id). Unhandled-event payloads logged at DEBUG.
- Gemini enum test fixture corrected (it pinned the buggy sibling `type`).
- Regression tests for both.

## Verification
Bisected live against api.openai.com with the real Atlassian tool list: swapping only the runtime-normalised `tools[]` into an otherwise-working raw request reproduces the failure; per-tool bisection isolates `getConfluenceSpaces`. With this branch, `cagent run` + Jira MCP and + Notion MCP on `gpt-5.6-terra` both answer normally (before: every turn `incomplete/max_output_tokens`, 0 tokens).

`go test ./pkg/tools/... ./pkg/model/... ./pkg/runtime/... ./pkg/teamloader/... ./pkg/config/...` green.

### DIFF
--- a/pkg/model/provider/openai/response_stream.go
+++ b/pkg/model/provider/openai/response_stream.go
@@ -389,14 +389,14 @@ func (a *ResponseStreamAdapter) Recv() (chat.MessageStreamResponse, error) {
 		// and the reason must reach the loop as a finish reason so the empty
 		// turn is explained rather than reported as "stop reason: null".
 		reason := event.Response.IncompleteDetails.Reason
-		slog.Warn("Response incomplete",
+		slog.Debug("Response incomplete",
 			"reason", reason,
 			"response_id", event.Response.ID,
 			"output_items", len(event.Response.Output),
 			"output_tokens", event.Response.Usage.OutputTokens,
 			"reasoning_tokens", event.Response.Usage.OutputTokensDetails.ReasoningTokens,
+			"response_raw", event.Response.RawJSON(),
 		)
-		slog.Debug("Incomplete response payload", "response_raw", event.Response.RawJSON())
 		u := event.Response.Usage
 		if u.TotalTokens > 0 {
 			response.Usage = &chat.Usage{
--- a/pkg/model/provider/openai/response_stream.go
+++ b/pkg/model/provider/openai/response_stream.go
@@ -2,6 +2,7 @@ package openai
 
 import (
 	"cmp"
+	"fmt"
 	"io"
 	"log/slog"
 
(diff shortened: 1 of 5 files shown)


pkg/model/provider/openai/response_stream.go
```diff
@@ -381,13 +382,61 @@ func (a *ResponseStreamAdapter) Recv() (chat.MessageStreamResponse, error) {
...
 	default:
 		slog.Info("Unhandled stream event type", "type", event.Type)
+		slog.Debug("Unhandled stream event payload", "type", event.Type, "raw", truncateForLog(event.RawJSON(), 4000))
 	}
 
 	return response, nil
 }
 
+// truncateForLog bounds a raw JSON payload for log output.
+func truncateForLog(s string, n int) string {
```
rumpl: Why truncate?

pkg/model/provider/openai/response_stream.go
```diff
@@ -381,13 +382,61 @@ func (a *ResponseStreamAdapter) Recv() (chat.MessageStreamResponse, error) {
...
 			},
 		}
+	case "response.incomplete":
+		// Terminal event: the model stopped before producing a complete
+		// response. incomplete_details.reason is "max_output_tokens" or
+		// "content_filter". Usage is still reported and must not be dropped,
+		// and the reason must reach the loop as a finish reason so the empty
+		// turn is explained rather than reported as "stop reason: null".
+		reason := event.Response.IncompleteDetails.Reason
+		slog.Warn("Response incomplete",
```
rumpl: Should be debug

pkg/model/provider/openai/response_stream.go
simonferquel-clanker: > 🤖 **Automated architect agent** — *this comment was posted by the architect bot from Docker Agentic Platform, not by a human engineer*

No good reason on reflection. My thinking was that the `response.incomplete` payload echoes the whole request `tools[]` back (~75 KB for the Atlassian server, ~150 KB for Notion), but the debug log already dumps the full request with the same tools a few lines earlier — so the cap only made the two inconsistent and risked cutting fields we'd actually want.

Removed in a255596: `truncateForLog` is gone, both DEBUG lines log `RawJSON()` as-is.

pkg/model/provider/openai/response_stream.go
simonferquel-clanker: > 🤖 **Automated architect agent** — *this comment was posted by the architect bot from Docker Agentic Platform, not by a human engineer*

Done in d6a60df — now a single `slog.Debug("Response incomplete", …)` carrying reason, response_id, token counts and the raw payload, matching the rest of the adapter. The incomplete state still reaches the caller through the finish reason (`length` / `refusal`), so nothing is lost at INFO.

rumpl (approved):

PR #4180, opened 2026-09-06
fix: make transfer_task concurrency-safe under parallel tool calls

### LINKED ISSUE #4156: transfer_task is not concurrency-safe: parallel calls race on current-agent pointer, later calls fail with "No agents are configured in this list"
## Symptom

When a model issues multiple `transfer_task` calls in one API response, `Dispatcher.Process` runs them in parallel goroutines (via `concurrent.MapSlice`). The first goroutine's `swapCurrentAgent` sets `r.current` to the target agent (e.g. "drafter"); subsequent goroutines' `handleTaskTransfer` → `resolveSessionAgent` read the mutated `r.current`, treat the target as the caller, and fail validation:

```
Agent drafter cannot transfer task to drafter: target agent not in sub-agents list. No agents are configured in this list.
```

## Real-World Impact

Broke the docker-agent-action PR review pipeline on docker/frontends#16712:
- Evidence run: https://github.com/docker/frontends/actions/runs/33804270624
- First delegation succeeded; the other two failed due to the race
- The orchestrator issued all three `transfer_task` calls in a single Anthropic API response (parallel tool use with a 3-chunk diff)

## Root Cause

**Concurrent mutation of `r.current`:**

```
Thread 1 (call 1):
  a = r.resolveSessionAgent(sess) → r.current = "orchestrator" → OK
  validateAgentInList(orchestrator, "drafter") → PASS
  runForwarding() → swapCurrentAgent: r.current = "drafter"  ← mutated

Thread 2 (call 2, concurrent):
  a = r.resolveSessionAgent(sess) → r.current = "drafter"    ← reads mutated value
  validateAgentInList(drafter, "drafter") → FAIL
```

The shared mutable field `r.current` is not safe for concurrent `swapCurrentAgent` + `handleTaskTransfer` calls. The caller agent identity must be captured before any concurrent dispatch, not re-read during validation.

## Affected Versions

At least v1.120.0–v1.131.0. The parallel dispatch via `concurrent.MapSlice` (and hence the race) pre-exists v1.121.0; no fix found in v1.121.0..v1.131.0 history.

## Suggested Fixes

1. **Store caller agent on session creation**: Record the "owner agent" in the session object at creation time; `resolveSessionAgent` reads the session-owned agent for root sessions, never `r.current`.
2. **Snapshot caller identity before dispatch**: Capture the caller agent at the top of `handleTaskTransfer` before any concurrent dispatch; pass it through to validation (do not re-read `r.current`).
3. **Serialize transfer_task in dispatcher**: Add a mutex that serializes all `transfer_task` calls; simplest runtime fix but less elegant.

## Workaround

Set `parallel_tool_calls: false` on the model in your agent config. This prevents the model from returning multiple `transfer_task` calls in a single API response, eliminating the race entirely.

## Key Code References

- `pkg/runtime/toolexec/dispatcher.go:257` — `concurrent.MapSlice(calls, ...)` parallel dispatch
- `pkg/runtime/agent_delegation.go:681` — `handleTaskTransfer` reads `r.current` via `resolveSessionAgent`
- `pkg/runtime/agent_delegation.go:326` — `runForwarding` / `swapCurrentAgent` mutates `r.current`
- `pkg/runtime/agent_router.go:76` — `ResolveSession` reads `r.current` for unpinned sessions

bernard-code-lab:
Fixes #4156

## Problem

When a model issues several `transfer_task` calls in one API response, `Dispatcher.Process` runs them in parallel goroutines via `concurrent.MapSlice`. Each goroutine re-resolved its caller through `resolveSessionAgent`, which reads the shared `agentRouter.current` that a sibling's `swapCurrentAgent` had already mutated:

```
T1: resolveSessionAgent(sess) -> "orchestrator"  -> validates OK
    runForwarding -> swapCurrentAgent -> current = "drafter"
T2: resolveSessionAgent(sess) -> "drafter"       <- reads the mutated value
    validateAgentInList("drafter", "drafter", ...) -> FAIL
```

The reported symptom is the loud one:

```
Agent drafter cannot transfer task to drafter: target agent not in sub-agents list.
No agents are configured in this list.
```

There is a quieter consequence too. `runForwarding` only pins the child session when the parent is already pinned, so a foreground delegation's child stays unpinned and `loop.go` re-resolves its agent from `current` on every iteration. With the deferred restores interleaving, a child could execute turns with another sibling's identity, toolset and model — no error, wrong output. `validateDelegation` inherits the same wrong caller, which can also produce a spurious `delegation cycle detected`.

`atomic.Pointer` keeps `current` free of data races, but it cannot make the read → validate → mutate → read sequence serializable.

## Fix

Two orthogonal changes.

**1. Resolve the caller from the batch snapshot.** `Dispatcher.Process` already resolves the caller exactly once, before the fan-out, and stores it on the `call` that backs every handler's `tools.Runtime`. It just was not reachable. `callRuntime` now exposes it, and `LocalRuntime.callerAgent` prefers it over session resolution, falling back for hosts that provide no snapshot. No handler signature changes — `rt tools.Runtime` was already a parameter of `ToolHandlerFunc`.

**2. Let one delegation own the shared current agent.** `runForwarding` claims it with a CAS; a sibling that loses the claim pins its child instead of stomping the swap. This is the same isolation mode background delegations already use.

`run_skill` shares `runForwarding` and had the same latent defect when it landed in a batch alongside `transfer_task`; the same change fixes it.

## Behavior change

With a single `transfer_task` in the batch — the common case — behavior is unchanged: the swap happens, the child stays unpinned, and the existing foreground tests pass untouched.

With N parallel transfers, the 2nd..Nth child is pinned. Pinned children do not take the `handoff` / `force_handoff` path (`loop.go` guards on `sess.AgentName == ""`). That is a pre-existing property of pinning, now reachable in a new scenario, and it is documented on `SwitchCurrentAgent`. If you would rather have no variance between the single and parallel cases, the alternative is to pin every child whenever the batch holds more than one delegation — same trade, but it gives up the untouched single-transfer path.

## Out of scope

`run_background_agent` resolves its caller the same way, but `agenttool.Handler.HandleRun` does not receive a `tools.Runtime`, so fixing it means changing a signature outside `pkg/runtime`. Left for a follow-up; its children are already pinned, which bounds the impact.

## Testing

Two regression tests, both verified to fail with the source changes reverted:

- `TestTransferTask_UsesBatchCallerSnapshotNotSharedCurrentAgent` — deterministic, reproduces the exact error string from the issue.
- `TestTransferTask_ConcurrentForegroundTransfersStayIsolated` — end-to-end through `RunStream` → `Dispatcher.Process` → `concurrent.MapSlice`, with three `transfer_task` calls in one assistant response. Without the fix two of the three children return no output and the race detector fires on a shared test stream, which is the cross-wiring made visible: two children had resolved to the same agent. With the fix it is stable over 20 runs under `-race`.

```
go build ./...                    Success
go test -race ./pkg/runtime/...   1141 passed
go test ./...                     6003 passed, 90 packages
golangci-lint run                 0 issues
go run ./lint .                   1929 files, no offenses
go mod tidy --diff                clean
```

`task lint` / `task test` were run as their underlying commands, since `task` is not installed locally.

### DIFF
--- a/pkg/runtime/agent_delegation.go
+++ b/pkg/runtime/agent_delegation.go
@@ -764,17 +764,22 @@ func (r *LocalRuntime) handleTaskTransfer(ctx context.Context, sess *session.Ses
 	})
 }
 
-func (r *LocalRuntime) handleHandoff(ctx context.Context, sess *session.Session, toolCall tools.ToolCall, _ EventSink, _ tools.Runtime) (*tools.ToolCallResult, error) {
+func (r *LocalRuntime) handleHandoff(ctx context.Context, sess *session.Session, toolCall tools.ToolCall, _ EventSink, rt tools.Runtime) (*tools.ToolCallResult, error) {
 	var params handoff.Args
 	if err := json.Unmarshal([]byte(toolCall.Function.Arguments), &params); err != nil {
 		return nil, fmt.Errorf("invalid arguments: %w", err)
 	}
 
-	ca := r.currentAgentName()
-	currentAgent, err := r.team.Agent(ca)
-	if err != nil {
-		return nil, fmt.Errorf("current agent not found: %w", err)
+	// Resolve the caller from the dispatcher's batch snapshot, never from the
+	// shared current agent: a transfer_task running beside this call in the
+	// same batch may already have swapped it, which would validate the handoff
+	// against the transfer target's handoffs list instead of the real
+	// caller's (#4156).
+	currentAgent := r.callerAgent(rt, sess)
+	if currentAgent == nil {
+		return nil, errors.New("no agent resolved for the calling session")
 	}
+	ca := currentAgent.Name()
 
 	if errResult := validateAgentInList(ca, params.Agent, "hand off to", "handoffs list", currentAgent.Handoffs()); errResult != nil {
 		return errResult, nil
--- a/pkg/runtime/agent_delegation.go
+++ b/pkg/runtime/agent_delegation.go
@@ -192,19 +192,29 @@ type SubSessionConfig struct {
 type delegationRequest struct {
 	SubSessionConfig
 
+	// CallerAgent is the agent that issued the delegating tool call,
+	// snapshotted by the caller before the dispatcher's parallel fan-out.
+	// runForwarding falls back to session resolution when it is nil.
+	CallerAgent *agent.Agent
+
+	// ConcurrentBatch reports that sibling delegations share this call's
+	// tool-call batch, so they run in parallel. Decided by the caller from
+	// the dispatcher's pre-fan-out tally, never by racing at runtime.
+	ConcurrentBatch bool
+
 	// SwitchCurrentAgent, when true, swaps r.currentAgent to AgentName
 	// for the lifetime of the call and emits AgentSwitching/AgentInfo
 	// events on entry and exit. Used by transfer_task. Mutually
 	// exclusive in spirit with PinAgent: pinning is for concurrent
 	// sub-sessions that must NOT share the runtime's mutable
-	// currentAgent, while switching is for sequential delegations where
-	// the parent loop is blocked anyway.
+	// currentAgent, while switching is for the one delegation that owns it.
 	//
-	// When the parent session is itself pinned (a background agent's
-	// session), runForwarding downgrades the switch to pinning the child
-	// to AgentName instead: the shared current agent belongs to the
-	// concurrent foreground loop and must not be mutated from a
-	// background task (#3886).
+	// runForwarding downgrades the switch to pinning the child to AgentName
+	// in two cases. When the parent session is itself pinned (a background
+	// agent's session), the shared current agent belongs to the concurrent
+	// foreground loop and must not be mutated from a background task (#3886).
+	// When sibling delegations share this call's tool-call batch, they all run
+	// in parallel and none may own the shared current agent (#4156).
 	SwitchCurrentAgent bool
 }
 
@@ -340,11 +350,14 @@ func (r *LocalRuntime) swapCurrentAgent(ctx context.Context, sessionID string, f
 func (r *LocalRuntime) runForwarding(ctx context.Context, parent *session.Session, evts EventSink, req delegationRequest) (*tools.ToolCallResult, error) {
 	span := trace.SpanFromContext(ctx)
 
-	// The caller resolves from the parent session, not the shared current
-	// agent: a nested transfer from a pinned background session must
-	// attribute events, hooks, and completion to the pinned agent, no
-	// matter where the concurrent foreground loop points (#3886).
-	callerAgent := r.resolveSessionAgent(parent)
+	// The caller never resolves from the shared current agent here: a nested
+	// transfer from a pinned background session must attribute events, hooks,
+	// and completion to the pinned agent (#3886), and a sibling call in the
+	// same parallel batch may already have swapped it (#4156).
+	callerAgent := req.CallerAgent
+	if callerAgent == nil {
+		callerAgent = r.resolveSessionAgent(parent)
+	}
 	if callerAgent == nil {
 		return nil, errors.New("no agent resolved for the parent session")
 	}
@@ -354,15 +367,23 @@ func (r *LocalRuntime) runForwarding(ctx context.Context, parent *session.Sessio
 	}
 
 	if req.SwitchCurrentAgent {
-		if parent.AgentName == "" {
-			defer r.swapCurrentAgent(ctx, parent.ID, callerAgent, child, evts)()
-		} else {
+		switch {
+		case parent.AgentName != "":
 			// Pinned parent (background delegation): the shared current
 			// agent belongs to the concurrent foreground loop and must not
 			// be mutated. Pin the child to the target instead — RunStream
 			// resolves pinned sessions directly, so the child still
 			// executes as the target agent, without switch events/hooks.
 			req.PinAgent = true
+		case req.ConcurrentBatch:
+			// Sibling delegations run beside this one, and the current agent
+			// is a single shared field: whoever swapped it would misroute the
+			// others' turns. Pin every member of the batch rather than letting
+			// one win a race — the parallel children are isolated, and which
+			// child keeps switch events no longer varies per run (#4156).
+			req.PinAgent = true
+		default:
+			defer r.swapCurrentAgent(ctx, parent.ID, callerAgent, child, evts)()
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/runtime/agent_delegation_test.go
```diff
@@ -1699,3 +1811,206 @@ func TestRunStream_NestedBackgroundAgents_EndToEnd(t *testing.T) {
...
+		setup            func(t *testing.T) (*LocalRuntime, *handoffRecordingProvider)
+		wantHandoffCalls int
+	}{
+		{
+			name: "solo delegation switches and honours the child's force_handoff",
+			setup: func(t *testing.T) (*LocalRuntime, *handoffRecordingProvider) {
+				t.Helper()
+				finisher, probe := handoffProbe("finisher")
+				worker := agent.New("worker", "worker agent",
+					agent.WithModel(&mockProvider{id: "test/mock-model",
```
aheritier, line 1887: [blocking] `lint` is red on this line:

```
pkg/runtime/agent_delegation_test.go:1887:1: File is not properly formatted (gofumpt)
```

Reproduced locally with golangci-lint 2.13.2. `task format` fixes it — the same `&mockProvider{...}` literal needs splitting in all four table cases.

pkg/runtime/agent_delegation_test.go
```diff
@@ -1699,3 +1811,206 @@ func TestRunStream_NestedBackgroundAgents_EndToEnd(t *testing.T) {
...
+	return agent.New(name, name+" agent",
+		agent.WithModel(prov),
+		agent.WithSubAgents(subAgents...),
+		agent.WithToolSets(transfertask.New()),
+	)
+}
+
+// handoffProbe returns an agent whose model invocations are counted, standing
+// in for a force_handoff target that must (or must not) be reached.
+func handoffProbe(name string) (*agent.Agent, *handoffRecordingProvider) {
```
aheritier, line 1844: [blocking] Second `lint` failure:

```
pkg/runtime/agent_delegation_test.go:1844:19: handoffProbe - name always receives "finisher" (unparam)
```

All four call sites pass `"finisher"`. Simplest fix is to drop the parameter and hardcode the name, since the probe is only ever the force_handoff target.

pkg/runtime/agent_delegation.go
```diff
@@ -354,15 +367,23 @@ func (r *LocalRuntime) runForwarding(ctx context.Context, parent *session.Sessio
...
 			// resolves pinned sessions directly, so the child still
 			// executes as the target agent, without switch events/hooks.
 			req.PinAgent = true
+		case req.ConcurrentBatch:
+			// Sibling delegations run beside this one, and the current agent
+			// is a single shared field: whoever swapped it would misroute the
+			// others' turns. Pin every member of the batch rather than letting
+			// one win a race — the parallel children are isolated, and which
+			// child keeps switch events no longer varies per run (#4156).
+			req.PinAgent = true
```
aheritier, line 384: [blocking] Pinning every member of the batch silently disables `force_handoff` for all of them — `loop.go:1035` gates `applyForceHandoff` on `sess.AgentName == ""`. Your own table encodes this at `agent_delegation_test.go:1961` (`wantHandoffCalls: 0`); on the merge base that case yields `1`.

That conflicts with shipped docs that promise the opposite:

- `docs/configuration/agents/index.md:119` — "**unconditionally** receives the conversation whenever this agent produces a final response … **guaranteeing deterministic pipelines**"
- `docs/concepts/multi-agent/index.md:147` — "`force_handoff` **removes that uncertainty** … the **runtime itself** routes the conversation"

So a `force_handoff` pipeline breaks with no error as soon as the model batches its transfers — the same quiet wrong-output class this PR is removing. Please either keep `force_handoff` reachable for pinned delegation children, or update both doc sites and the `SwitchCurrentAgent` comment in the same change to state the parallel-batch exception. (The PR body’s "the 2nd..Nth child is pinned" is also stale — all children are.)

pkg/runtime/toolexec/dispatcher.go
```diff
@@ -253,12 +253,21 @@ func (d *Dispatcher) Process(ctx context.Context, sess *session.Session, calls [
...
 	}
 
+	// Tally the batch per tool so a handler can tell a solo invocation from
+	// one whose siblings run beside it. transfer_task needs this to decide
+	// deterministically, before any of them starts, whether a delegation may
+	// own the runtime's shared current agent (#4156).
+	batchByTool := make(map[string]int, len(calls))
+	for _, tc := range calls {
+		batchByTool[tc.Function.Name]++
+	}
```
aheritier, line 263: [should-fix] Keying the tally on `tc.Function.Name` protects `transfer_task` from its own siblings, but not from the other tool that reaches `runForwarding`. With a batch of `[transfer_task, run_skill]` the transfer sees `concurrentSiblings == 1`, stays solo, and swaps `current`; `runSkillFork` leaves its child unpinned for an unpinned parent (`skill_runner.go:147`), so the fork resolves from the swapped field.

A reproducer on this head expected `SKILL-RAN-AS-ROOT` but received `SKILL-RAN-AS-WORKER`; the child had `AgentName == ""`. This is pre-existing rather than a regression, but the PR body’s claim that `run_skill` is fixed by the same change only holds for caller resolution, not child execution. Count the delegating tools together (or expose batch size so `runSkillFork` can pin when >1), and add a mixed `transfer_task` + `run_skill` batch test beside the `think` case.

pkg/runtime/agent_delegation.go
```diff
@@ -735,21 +758,28 @@ func (r *LocalRuntime) handleTaskTransfer(ctx context.Context, sess *session.Ses
...
-	ca := r.currentAgentName()
-	currentAgent, err := r.team.Agent(ca)
-	if err != nil {
-		return nil, fmt.Errorf("current agent not found: %w", err)
+	// Resolve the caller from the dispatcher's batch snapshot, never from the
+	// shared current agent: a transfer_task running beside this call in the
+	// same batch may already have swapped it, which would validate the handoff
+	// against the transfer target's handoffs list instead of the real
+	// caller's (#4156).
+	currentAgent := r.callerAgent(rt, sess)
```
aheritier, line 778: [should-fix] This changes behavior outside the parallel-batch scenario. `callerAgent` falls back to `resolveSessionAgent`, which prefers `sess.AgentName`, so a handoff from a pinned background session now validates against the pinned agent’s `handoffs` instead of the foreground agent’s — then reaches `setCurrentAgent(next.Name())` at line 815 with no pinned-session guard, mutating shared current state from a background task. That is what #3886 forbids, and this change appears to make the path newly reachable where it previously failed validation.

Intended? Please add a handoff-from-a-pinned-session test to make the intended behavior explicit. (#4242 covers the batch-clobber case, but not this one.)

aheritier:

Commits that are not verified: `4983673`, `7c9244f`, `5883d1e`, `d2f36bb`

aheritier:

PR #4190, opened 2026-09-07
fix: send x-opencode-session header to OpenCode providers

### LINKED ISSUE #4164: OpenCode Go/Zen do not send an x-opencode-session header (required by OpenCode from 2026-09-06)
### Description

OpenCode has started asking clients to send an `x-opencode-session` header ("one stable ID per conversation") on every request to OpenCode Go, and announced that requests without it may error starting 2026-09-06. docker-agent does not send it, neither via the built-in opencode-go / opencode-zen aliases nor via a custom provider, and there is no config-level way to produce a per-conversation value.

OpenCode's notice (received by email, 2026-09-04):
```
Some of your requests to OpenCode Go are missing an x-opencode-session header. If we don't have this we cannot properly optimize our service. Starting 09/06 requests missing this header may error.
Here are your useragents that are missing this header: Cagent/v1.126.0
We don't recognize this client — add x-opencode-session (one stable ID per conversation) or ask its maintainer to.
```

This was also announced on their X account [here](https://x.com/opencode/status/2095410501400289576).

Their [docs](https://opencode.ai/docs/go/#where-can-i-use-it) describe the header as the key OpenCode uses to optimize prompt caching.

### Expected Behavior

Inject a default `x-opencode-session` header for OpenCode providers, the same way chatgptAuthMiddleware (pkg/model/provider/openai/chatgpt.go) already generates a per-client uuid.NewString() and sets it as the session_id header when the user has not set one

### Actual Behavior

- pkg/model/provider/aliases.go: opencode-go and opencode-zen only set base_url and token_key; no default headers.
- pkg/model/provider/openai/headers.go (buildHeaderMap): the only provider-specific default header is Copilot-Integration-Id for github-copilot. User headers from provider_opts.http_headers are applied verbatim and fixed at client construction.

There is no way to reference the session ID from config: {{session_id}} is only expanded for commands.*.url in the TUI (pkg/tui/handlers.go), and ${env.X} / ${X} are not expanded in http_headers either.

The only workaround today is a hard-coded value in http_headers, i.e. one ID per deployment rather than per conversation, which defeats the purpose of the header for serve api / serve chat deployments that multiplex many conversations.

### Steps to Reproduce

Verified against a local fake OpenAI endpoint:

```
provider_opts:
  http_headers:
    x-opencode-session: "{{session_id}}"
```
the server receives the literal string {{session_id}}.

### Docker Agent version

v1.126.0

### OS & terminal

macOS

### Model used

_No response_

### Error output

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

IsmaelMartinez:
[OpenCode's docs](https://opencode.ai/docs/go/#where-can-i-use-it) ask clients to send a stable ID per conversation in `x-opencode-session`, which they use to optimise routing and prompt caching. The reporter of #4164 additionally quotes an OpenCode email saying requests missing the header may error from 2026-09-06; the docs themselves state no date, so treat the deadline as their account rather than a documented one.

A host-gated `http.RoundTripper` in `pkg/model/provider/base` sets the header on the direct HTTP client of the OpenAI, Anthropic and Gemini providers, so the built-in aliases and custom providers on any of the three clients send it, including the Anthropic-compatible and Google models Docker Agent's OpenCode Go and Zen pages configure. The gate is the base URL host alone, so an alias given its own `base_url` sends nothing, and OpenCode models stay on SSE even with `transport: websocket`, since a WebSocket dial never reaches the transport. The value is a UUIDv5 derived from the session ID already on the request context, salted so it cannot be correlated with the local session store or the gateway's `X-Cagent-Session-Id`. Requests with no session on the context fall back to a per-client UUID, a value pinned through `provider_opts.http_headers` on the OpenAI client always wins, and gateway traffic is untouched since it already carries `X-Cagent-Session-Id`.

This supersedes an earlier version of this PR that set the header once at client construction. @yunus25jmi1 pointed out in #4191 that the client is built once per process in `serve a2a`, `serve mcp` and the TUI, so a construction-time value collapses to one ID per deployment in exactly the multiplexed cases the issue is about. The per-request derivation is theirs, carried over with the commit co-authored; my part is the shared transport, the docs wording and the accuracy corrections against OpenCode's published text.

Fixes #4164

### DIFF
--- a/pkg/model/provider/openai/client.go
+++ b/pkg/model/provider/openai/client.go
@@ -133,6 +133,12 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		// required Copilot-Integration-Id) and any provider-specific defaults.
 		clientOptions = append(clientOptions, buildHeaderOptions(cfg)...)
 
+		// The session comes off the request context, so this cannot be a
+		// static header fixed at construction time.
+		if isOpenCodeProvider(cfg) {
+			clientOptions = append(clientOptions, option.WithMiddleware(opencodeSessionMiddleware()))
+		}
+
 		// Preserve full error details from non-OpenAI providers (e.g. GitHub
 		// Copilot returns a bare "400 Bad Request" whose body explains the
 		// actual cause); without this the SDK discards it.
--- a/docs/providers/opencode-go/index.md
+++ b/docs/providers/opencode-go/index.md
@@ -132,6 +132,10 @@ This means OpenCode Go uses the same client as OpenAI, making it fully compatibl
 
 For Anthropic-compatible models (MiniMax, Qwen), Docker Agent uses a custom provider pointing to the Anthropic client at `https://opencode.ai/zen/go` with the same token.
 
+### Session Header
+
+OpenCode asks clients to send an `x-opencode-session` header carrying one stable ID per conversation, which it uses to optimise routing and prompt caching. Docker Agent sends it automatically on every request to `opencode.ai` (built-in aliases and custom providers alike), deriving an opaque value from the agent session so each conversation keeps the same ID, including in `serve api` / `serve chat` deployments that multiplex many conversations. To pin your own value, set `provider_opts.http_headers.x-opencode-session`.
+
 ## Example: Code Assistant
 
 ```yaml
--- a/docs/providers/opencode-zen/index.md
+++ b/docs/providers/opencode-zen/index.md
@@ -204,6 +204,10 @@ The same API key works for both OpenCode Go and OpenCode Zen — they are part o
 
 For Anthropic-compatible models, Docker Agent uses a custom provider pointing to the Anthropic client at `https://opencode.ai/zen` with the same token. For Google models, a custom provider points to the Google client at `https://opencode.ai/zen` (the Google SDK appends its own `/v1beta/models/...` path segment).
 
+### Session Header
+
+OpenCode asks clients to send an `x-opencode-session` header carrying one stable ID per conversation, which it uses to optimise routing and prompt caching. Docker Agent sends it automatically on every request to `opencode.ai` (built-in aliases and custom providers alike), deriving an opaque value from the agent session so each conversation keeps the same ID, including in `serve api` / `serve chat` deployments that multiplex many conversations. To pin your own value, set `provider_opts.http_headers.x-opencode-session`.
+
 ### Differences from OpenCode Go
 
 | Aspect | OpenCode Zen | OpenCode Go |
--- a/pkg/model/provider/openai/opencode.go
+++ b/pkg/model/provider/openai/opencode.go
@@ -1,0 +1,66 @@
+package openai
+
+import (
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/google/uuid"
+	"github.com/openai/openai-go/v3/option"
+
+	"github.com/docker/docker-agent/pkg/config/latest"
+	"github.com/docker/docker-agent/pkg/httpclient"
+)
+
+// OpenCode asks clients to send one stable ID per conversation; it keys their
+// routing and prompt caching.
+const (
+	opencodeSessionHeader = "x-opencode-session"
+	opencodeHost          = "opencode.ai"
+)
+
+// opencodeSessionNamespace salts the hash so the header cannot be mapped back
+// to the local session ID or the gateway's X-Cagent-Session-Id.
+var opencodeSessionNamespace = uuid.MustParse("6f0c2a1e-8d4b-4c7f-9a3e-2b5d7e9f1c03")
+
+// isOpenCodeProvider reports whether requests target OpenCode, either through
+// the built-in aliases or a custom provider pointed at opencode.ai.
+func isOpenCodeProvider(cfg *latest.ModelConfig) bool {
+	if cfg == nil {
+		return false
+	}
+	switch cfg.Provider {
+	case "opencode-go", "opencode-zen":
+		return true
+	}
+	u, err := url.Parse(cfg.BaseURL)
+	if err != nil {
+		return false
+	}
+	host := strings.ToLower(u.Hostname())
+	return host == opencodeHost || strings.HasSuffix(host, "."+opencodeHost)
+}
+
+// opencodeSessionID derives the header from the agent session, so multiplexed
+// deployments get one ID per conversation rather than per client, and a
+// resumed session keeps its ID across processes.
+func opencodeSessionID(sessionID string) string {
+	return uuid.NewSHA1(opencodeSessionNamespace, []byte(sessionID)).String()
+}
+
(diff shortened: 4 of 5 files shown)


pkg/model/provider/openai/client.go
```diff
@@ -133,6 +133,12 @@ func NewClient(ctx context.Context, cfg *latest.ModelConfig, env environment.Pro
 		// required Copilot-Integration-Id) and any provider-specific defaults.
 		clientOptions = append(clientOptions, buildHeaderOptions(cfg)...)
 
+		// The session comes off the request context, so this cannot be a
+		// static header fixed at construction time.
+		if isOpenCodeProvider(cfg) {
```
aheritier, line 138: **[blocking]** This installs the middleware only for the OpenAI client. The changed provider docs also promise automatic handling for custom providers, but their documented Anthropic and Google configurations are dispatched to separate factories (`pkg/model/provider/providers/providers.go:35-36`) and never reach this code. Consequently, those requests to `opencode.ai` still omit `x-opencode-session`. Please either install the host-scoped behavior in a transport path shared by those providers and test it, or explicitly narrow the documented and intended scope to OpenAI-compatible providers.

aheritier (commented):
Reviewed #4190 at `d4fa1553528e18cbdac802fa7f0cbb0cbd9eb01f` against #4164 and the linked discussion.

The per-request implementation for OpenAI-compatible OpenCode providers is sound: it derives an opaque, stable value from the conversation context, uses a per-client fallback when no session is available, preserves explicit user overrides, and avoids sending the header to unrelated OpenAI providers. The tests cover those behaviors.

**[blocking] OpenCode's documented Anthropic and Google custom-provider paths still do not receive the header.** The middleware is installed only inside `openai.NewClient` (`pkg/model/provider/openai/client.go:138`), while custom providers with `provider: anthropic` or `provider: google` are dispatched to separate factories (`pkg/model/provider/providers/providers.go:35-36`). The changed documentation nevertheless says Docker Agent sends the header on every request to `opencode.ai`, "built-in aliases and custom providers alike," immediately after documenting those Anthropic/Google configurations (`docs/providers/opencode-go/index.md:133-137`, `docs/providers/opencode-zen/index.md:205-209`). Those requests can therefore still be rejected for a missing `x-opencode-session` header. Please either apply the host-scoped middleware at a transport layer shared by these providers and add coverage, or narrow the documented and intended scope explicitly to OpenAI-compatible providers.

CI is not green because it has not run: all five workflow runs for the head SHA are `completed/action_required`, there are zero check runs and no commit status contexts, and the required `gate` check is absent. The workflows need approval and a green rerun before this is merge-ready.

Local validation of the exact head passed:

- `task lint`
- `task test`
- `task build`
- `git diff --check 9b7d2b5fe30890944046c62b79acf1f1a3864442 d4fa1553528e18cbdac802fa7f0cbb0cbd9eb01f`

pkg/model/provider/openai/client.go
IsmaelMartinez: Good catch, thanks Arnaud. Pushed the transport route: the host check and header now live in `pkg/model/provider/base` as an `http.RoundTripper` on the OpenAI, Anthropic and Gemini direct paths, with a capture test per provider. Gateway traffic is untouched (it already carries `X-Cagent-Session-Id`) and the docs now say so, along with `http_headers` pinning being OpenAI-only. Separate issue to follow: the documented Anthropic/Google configs use `token_key: OPENCODE_API_KEY`, but those clients only read their native env vars. If you'd prefer a different shape, just let me know.

IsmaelMartinez: The separate issue is up: #4284, with the fix in #4285 (the Anthropic and Gemini clients now honour `token_key`, so the documented OpenCode custom-provider configs authenticate as written).

docker-agent (commented):
### Assessment: 🟢 APPROVE

- pkg/model/provider/base/opencode.go:22 — Static hardcoded UUIDv5 namespace provides only weak session-ID obfuscation (verifier mitigation: The header is used for routing/caching only — not auth. Underlying session IDs are random UUIDs with ~122 bits of entropy, so an attacker cannot precompute the derived value. Even with the namespace known, no privilege or access is gained by knowing the header value.)

aheritier (changes requested):
Two correctness issues should be addressed before merge:

1. **Host-gating can send `x-opencode-session` to a non-OpenCode endpoint** (`pkg/model/provider/base/opencode.go:28`). `isOpenCodeProvider` returns true for `opencode-go` / `opencode-zen` by name before checking `BaseURL`; a custom base URL or shadowed provider alias can therefore leak the derived identifier to another endpoint. Please gate solely on the intended OpenCode host.

2. **`transport: websocket` bypasses the header** (`pkg/model/provider/openai/client.go:209`). WebSocket dialing does not pass through `http.RoundTripper`, so OpenCode requests silently omit the header this PR introduces. Please disable WebSockets for OpenCode using the existing wrapped-transport precedent, or set the header during WebSocket dialing.

The prior Anthropic/Gemini coverage concern is resolved; CI and local validation are green. The remaining observations are non-blocking.

docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34945615906) for details. Re-request a review from `docker-agent` to retry.

docker-agent (commented):
### ⚠️ Review incomplete

This review covers only the commits pushed since `6058b0637c49d6db6fd41704361655bb18781cc1`.

aheritier (approved):
Re-reviewed the current head after the previous review round.

Issue-to-implementation checklist for #4164 and linked context (#4191, #4284, and merged #4285):
- [x] Direct OpenCode Go/Zen requests send `x-opencode-session` through the OpenAI, Anthropic, and Gemini clients.
- [x] Custom providers targeting `opencode.ai` receive the same behavior.
- [x] The value is stable per conversation, distinct across conversations, and does not expose the raw session ID.
- [x] Requests without a session use a stable per-client fallback.
- [x] Explicit OpenAI `http_headers` values take precedence.
- [x] Host-only gating prevents aliases redirected to unrelated endpoints from receiving the header.
- [x] OpenCode WebSocket requests fall back to SSE so the RoundTripper cannot be bypassed.
- [x] Gateway traffic remains unchanged.
- [x] Provider-level and shared transport tests cover the required behavior and non-leakage cases.
- [x] Documentation matches the implemented scope and transport behavior.

The previous Anthropic/Google coverage, host-gating, and WebSocket findings are resolved. No Jira references were found. The required `ci / gate` check and every reported non-skipped check run are green; targeted tests for the four affected provider packages also pass locally. The separate `pr-review-trigger` action-required run is the failed automated review retry and is not a failing required check.

The PR is non-draft and conflict-free. It is implementation-ready; the current `blocked` merge state reflects the outstanding prior changes-requested review and should clear when that review state is superseded or dismissed.

PR #4199, opened 2026-09-08
feat(plan)!: remove the session_plan toolset

trungutt:
## Why

The `session_plan` toolset existed to support a per-session "draft, review, execute" plan-mode workflow driven by a host application. Downstream consumers have retired that workflow and migrated off the toolset, leaving it with no remaining users — but with a wide footprint: runtime-owned tool handlers, a dedicated stream event, a session scope threaded through the host-facing plans service, read-only session rows in the TUI `/plans` browser, and `--session`/`--scope` addressing on the plans CLI.

This PR removes the toolset and all of that surface in one pass, so the plans stack is shared-plans-only:

```
before                                    after

session_plan toolset ──┐
  write_session_plan   │
  read_session_plan    ├─► removed
  exit_plan_mode       │
session_plan_updated ──┘

plan toolset (shared) ──► unchanged
  /plans TUI browser  ──► shared plans only (all rows now mutable)
  docker agent plans  ──► addressed by name only (no --session/--scope)
  pkg/plans Service   ──► single scope; UpdateSession/UnsupportedError gone
```

## Compatibility notes

- **Agent configs**: `session_plan` is removed from the schema's toolset-type enum. Configs still using it fail validation — that is the breaking change, flagged in the commit.
- **Event stream**: `session_plan_updated` is no longer emitted. Clients ignore unknown event types, and no event with this type will ever arrive again, so older clients are unaffected.
- **Plans CLI JSON**: the v1 wire contract is preserved — `scope` remains in every document (always `"shared"`); the `unsupported` error code disappears along with the only operation that produced it.

The shared `plan` toolset, its optimistic-locking semantics, and the `/plans` browser behaviour for shared plans are untouched.


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/docker-agent/actions/runs/34207876181).

dgageot (approved):

PR #4201, opened 2026-09-08
feat(gateway): carry encrypted agent config in request body and config-load envelope

gtardif:
The encrypted agent config previously travelled in the `X-Cagent-Encrypted-Config` header, both on gateway-bound model requests and on config-load responses from a trusted Docker source. A large encrypted config can exceed header-size limits, so it now rides in the body in both directions:

- **Request path** (docker-agent -> gateway): injected as a top-level `encrypted_agent_config` field into the JSON request body at the transport layer, gateway-bound (`X-Cagent-Forward`) and JSON-only, retry-safe. The proxy strips it before forwarding upstream.
- **Config-load path** (gateway -> docker-agent): a trusted Docker source returns a JSON envelope (`application/vnd.docker.agent.config+json`) carrying both the agent YAML and the encrypted config on a 200; the client unwraps it and caches the inner YAML. Unencrypted agents are still served as raw YAML. The 304 path keeps sending only the small digest header for cache revalidation.

Removes the now-unused `EncryptedConfigHeader` constant.

## Testing

- `task build` succeeds.
- Tests for all affected packages pass: `pkg/config/...`, `pkg/httpclient/...`, `pkg/model/provider/base/...`.

### DIFF
--- a/pkg/config/hcl/source.go
+++ b/pkg/config/hcl/source.go
@@ -46,3 +46,20 @@ func (s source) Read(ctx context.Context) ([]byte, error) {
 	}
 	return converted, nil
 }
+
+// encryptedConfigSource mirrors config.EncryptedConfigSource so this package
+// can forward the capability without importing pkg/config.
+type encryptedConfigSource interface {
+	EncryptedConfig() string
+}
+
+// EncryptedConfig forwards the inner source's captured encrypted agent config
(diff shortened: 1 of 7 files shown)


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/docker-agent/actions/runs/34205189781).

trungutt (approved):

PR #4203, opened 2026-09-08
fix: drop Gemini keepalive SSE events before SDK parsing

dgageot:
The Gemini SDK rejects upstream SSE heartbeat events that arrive as `event: keepalive` with `data: {}`. These frames are valid at the transport layer but cause the SDK to fail parsing when it encounters them unexpectedly.

The fix sits in the shared SSE transport layer: `sseFilter` now drops any event whose SSE `event:` field is exactly `keepalive` before the bytes reach the SDK. All other named and unnamed events pass through unchanged, so no existing behavior is affected beyond silencing those heartbeats.

New unit tests cover the filter directly (named keepalive events are dropped, anonymous data events are preserved), and new Gemini integration tests reproduce the original parse failure on both the direct and gateway paths, then confirm that content, usage metadata, and normal stream completion are correct after the fix.

### DIFF
--- a/pkg/httpclient/sse_filter.go
+++ b/pkg/httpclient/sse_filter.go
@@ -10,7 +10,7 @@ import (
 
 // sseFilterTransport wraps a base RoundTripper and, when the response is a
 // `text/event-stream`, replaces the body with one that strips SSE events
-// containing no `data:` lines.
+// containing no `data:` lines or named `keepalive`.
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### ⚠️ Review incomplete

aheritier (approved):

PR #4204, opened 2026-09-08
fix: snapshot caller agent before parallel tool dispatch to eliminate transfer_task race

derekmisler:
## what

When the root (foreground, unpinned) session emits two or more `transfer_task` calls in a single LLM response, the dispatcher runs them concurrently via `concurrent.MapSlice`. The second goroutine called `resolveSessionAgent(sess)` after the first had already invoked `swapCurrentAgent`, so it resolved the caller as the target agent instead of the root agent, failing validation with:

```
Agent drafter cannot transfer task to drafter:
target agent not in sub-agents list. No agents are configured in this list.
```

This produced a 19.2% "Review incomplete" spike in the review pipeline after bumping docker-agent from v1.121.0 to v1.131.0 (where multi-chunk PRs caused the LLM to emit parallel `transfer_task` calls more frequently).

## fix

Two cooperating changes:

1. **`toolexec.ToolHandler` and `runtime.ToolHandlerFunc`** now include `*agent.Agent` as the second parameter. `Dispatcher.Process` already resolves `a := d.AgentFor(sess)` once before the parallel loop; passing it to every handler ensures each goroutine uses the same pre-dispatch snapshot regardless of concurrent `r.current` mutations.

2. **`handleTaskTransfer`** uses the passed `*agent.Agent` directly for validation and passes it as `delegationRequest.Caller` so `runForwarding` also uses the pre-resolved identity for hook attribution and `SubSessionCompleted` events.

All other handlers accept the parameter but ignore it (marked `_`).

## test

`TestTransferTask_ConcurrentFromUnpinnedForegroundSession` - two concurrent `transfer_task` calls from the same unpinned foreground session (the exact pattern the pr-review root agent emits for multi-chunk diffs). Before the fix: second call fails with the drafter-self-delegation error. After: both pass validation and complete successfully.

Verified existing `TestTransferTask_ForegroundSwitchesAndRestoresCurrentAgent` and all delegation tests still pass.

Closes: the transfer_task race condition reported in the v2.0.6 "Review incomplete" regression.

### DIFF
--- a/pkg/runtime/agent_delegation.go
+++ b/pkg/runtime/agent_delegation.go
@@ -192,6 +192,14 @@ type SubSessionConfig struct {
 type delegationRequest struct {
 	SubSessionConfig
 
+	// Caller, when non-nil, is used as the calling agent instead of
+	// re-resolving via resolveSessionAgent inside runForwarding. Callers
+	// that have already resolved the agent (e.g. handleTaskTransfer,
+	// which validates it before calling runForwarding) must set this to
+	// prevent a concurrent swapCurrentAgent from corrupting the identity
+	// read for a second, parallel tool call from the same LLM response.
+	Caller *agent.Agent
+
 	// SwitchCurrentAgent, when true, swaps r.currentAgent to AgentName
 	// for the lifetime of the call and emits AgentSwitching/AgentInfo
 	// events on entry and exit. Used by transfer_task. Mutually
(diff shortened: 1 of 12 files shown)


docker-agent (commented):
### ⚠️ Review incomplete

chunk 1: checking closure capture bug in the for loop

The drafter agent did not return a complete review — it ran out of context before finishing analysis of the diff. The diff (497 lines across 12 files) was provided as a single chunk; re-triggering the review or splitting into smaller chunks may resolve this.

No findings were confirmed or dismissed in this run.

PR #4205, opened 2026-09-09
docs: add tip for injecting session ID with a session_start hook

dgageot:
Docker Agent does not automatically expose the session ID as an environment variable, so users who need it in the model's context have to reach for hooks. This tip documents the pattern.

The new section shows how to attach a `session_start` command hook that reads `.session_id` from the JSON payload on stdin via `jq` and prints a human-readable line that Docker Agent injects as model-visible context. It notes the `jq` dependency, clarifies that no `$SESSION_ID` env var is set automatically, and mentions `turn_start` as an alternative when fresh context before every call is needed.

No behavior change; documentation only.


aheritier (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

PR #4206, opened 2026-09-09
feat(hooks): add add_context builtin for dependency-free template context injection

dgageot:
The existing way to inject hook-derived context into the model's conversation requires an external tool such as `jq` to parse the JSON payload on stdin and format the output. This is fragile in environments where `jq` is absent, and it forces users to write shell quoting around Go template-style strings.

`add_context` is a new builtin hook that takes a list of Go `text/template` strings as `args`, renders each one against the typed hook input (using Go field names like `.SessionID`, `.AgentName`, `.Cwd`), and joins non-blank results with newlines into `additional_context`. No shell, no JSON parser, no external binary. Template parse and execution errors are strict: they discard the hook's output and follow the configured `on_error` policy. Rendered output is plain text only; it is not re-evaluated or executed.

The tips guide's session-ID example, which used a `jq` command hook, is updated to use `add_context` instead. A new standalone `examples/context_hooks.yaml` shows the builtin in action; `examples/hooks.yaml` gains a matching entry. Schema descriptions and `types.go` comments are updated to document the new builtin. Unit tests cover template rendering, blank-result filtering, and error propagation; a runtime integration test covers the end-to-end path through the hooks engine.


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/docker-agent/actions/runs/34321660161).

trungutt (approved):

PR #4207, opened 2026-09-09
feat(hooks): sequential pipeline for pre_tool_use, before_llm_call, and tool_response_transform

dgageot:
Hooks for `pre_tool_use` (default lane), `before_llm_call`, and `tool_response_transform` previously ran concurrently and used a first-rewrite-wins strategy. This made it impossible for one hook to build on another's output: a redaction hook and a normalization hook on the same event would race, and whichever wrote first would see its work discarded or its changes overwritten silently. The only workaround was to bundle all logic into a single monolithic hook.

These three events now execute hooks sequentially in configuration order. Each hook receives the accumulated result of all prior accepted rewrites — the updated `tool_input`, `messages` array, or `tool_response` — and can further transform it. Verdicts continue to aggregate across the full sequence (`deny > ask > allow`), and a blocking verdict does not short-circuit remaining hooks. All other events, including `preempt_yolo` checks and `before_compaction`, remain concurrent.

The `updated_input` field for `pre_tool_use` changes from a full replacement to a shallow-merge patch: only the top-level keys present in the returned object are applied; omitted keys are preserved from the current input. This is a compatibility change — previously any omitted key was silently dropped. Key deletion is not supported by this patch protocol.

The automatic `limit_large_tool_results` builtin is now appended *after* all configured `tool_response_transform` hooks and automatic `redact_secrets` injection, so spill files written to disk already contain scrubbed content rather than raw secrets. This restores the correct ordering that was previously broken because the limiter was prepended before user-configured transforms.

The new `pkg/hooks/pipeline.go` implements the sequential executor. Tests added cover the argument-preservation invariant (a second hook sees the first hook's patch merged over the original), the redact-before-spill invariant (spill file on disk is clean), and the ordering of builtin injection in `ApplyAgentDefaults`. The docs and the `redact_secrets_hooks.yaml` example are updated to reflect the new composition semantics.

### DIFF
--- a/docs/configuration/hooks/index.md
+++ b/docs/configuration/hooks/index.md
@@ -394,7 +394,7 @@ The `hook_specific_output` for `pre_tool_use` (and `permission_request`) support
 | ---------------------------- | ------ | --------------------------------------- |
 | `permission_decision`        | string | `allow`, `deny`, or `ask`               |
 | `permission_decision_reason` | string | Explanation for the decision            |
-| `updated_input`              | object | Modified tool input (replaces original) |
+| `updated_input`              | object | Top-level patch to the current tool input (`pre_tool_use` default lane only); omitted keys are preserved |
 | `metadata`                   | object | (`permission_request` and `pre_tool_use` entries with `preempt_yolo: true` only) string key/value annotations merged onto the tool-call confirmation prompt — see below |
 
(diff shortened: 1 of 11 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

hamza-jeddad (approved):

PR #4209, opened 2026-09-09
feat(safety): expand shell safety patterns and harden substitution checks

dgageot:
The shell safety classifier accumulates coverage debt over time as new command patterns are exercised and either misclassified or left unclassified. This change makes a substantial pass at closing that gap: 56 new safe entries and 27 new destructive entries bring the totals to 133 safe / 167 destructive patterns, covering Git read operations, GitHub CLI queries, Go tooling, `rg`/`ripgrep`, and common inspection commands that were previously unclassified.

Beyond raw pattern count, several correctness issues are fixed. Numeric-argument patterns for `sed`, `head`, and `tail` are now constrained with anchors so they cannot match arbitrary suffixes. The denied-flag logic in the classifier is hardened to handle quoted flags and clustered short flags (e.g. `-rf`), which previously bypassed denial. A shared metacharacter guard now blocks zsh `=(...)` process substitution and fish `(...)` command substitution in both the safe classifier and the session-grant path. The `gh run list -w` workflow-filter flag is corrected so `gh` patterns stop incorrectly classifying that common invocation as unsafe.

Shell grant logic in `shell_grant.go` is simplified — redundant branches are removed and the metachar check is unified with the classifier's guard. New regression tests cover quoted/clustered flag handling, the metachar variants, and the corrected `gh` patterns. Documentation in `docs/tools/shell/index.md` is updated to reflect the current pattern counts accurately.


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/docker-agent/actions/runs/34335121761).

aheritier (approved):

PR #4211, opened 2026-09-09
feat(hooks): tool hook phases — mandatory guards, transforms, dedup and strict output

dgageot:
Hook execution for tool calls had no guaranteed ordering between transform, guard and approval hooks, deduplication was shallow (missing `strict_output` and ignoring nil-vs-empty equivalence), and error semantics were inconsistent across cancellation, bad exit codes and malformed JSON.

This PR introduces a fixed, mandatory execution order: `tool_input_transform` always runs first and may rewrite the tool's input; `tool_guard` then runs mandatory approval with guards and rules rechecked against any rewritten input; normal approval hooks follow. Security guards and `pre_tool_use` hooks fail closed on unexpected outcomes. Parent-context cancellations are now reported as *canceled* rather than denied. Unexpected exit codes, malformed JSON and invalid verdicts all produce consistent, structured errors. `strict_output` is an opt-in per-hook flag that validates hook scripts to only emit recognized JSON event types; `stop` and `post_tool_use` contexts are non-consuming and strict mode detects that. A shared, stdlib-only `pkg/hooks/events` catalog drives config validation, execution-strategy selection and result aggregation, replacing scattered repeated event lists. Hook deduplication now covers all fields and treats nil and empty `args`/`env` as equivalent, retaining the first occurrence. Invalid regex patterns, unknown policies and misplaced `preempt_yolo` are caught at config-load time.

A few compatibility notes: successful no-op scripts must exit 0 rather than 1; `on_error:block` is rejected at config load for events that cannot block; customized or named built-ins are treated as distinct from the automatic defaults; experimental WASM does not yet implement the native mandatory tool phases (documented). Depends on #4207 (sequential pipeline), already merged.


docker-agent (commented):
### ⚠️ Review incomplete

chunk 1: analyzing
chunk 2: Drafter did not complete

aheritier (approved):

PR #4214, opened 2026-09-09
feat(openai): support provider_opts.service_tier for Fast mode

dgageot:
OpenAI's [Fast mode](https://developers.openai.com/api/docs/guides/fast-mode) lets callers trade response quality for lower latency and cost by setting `service_tier` to `"fast"` (alias: `"priority"`). There was no way to forward this hint through docker-agent, so any agent running on an OpenAI-compatible backend couldn't opt into it.

This adds a `service_tier` field under `provider_opts` that is forwarded unchanged to the underlying API. The field is applied consistently across chat completions, Responses API (SSE and WebSocket), and reranking requests via the native OpenAI SDK field. An empty or absent value is omitted from the request entirely; an unrecognized value is silently dropped with a debug log rather than failing the call. The accepted values are `"auto"`, `"default"`, `"flex"`, `"fast"`, and `"priority"` — the last two are equivalent aliases for Fast mode.

The schema, docs, and a worked example (`examples/openai-service-tier.yaml`) are all updated. Docs note provider compatibility (the field is ignored by non-OpenAI-compatible servers) and the cost implications of `"priority"`. Table-driven mock HTTP/WebSocket wire tests cover the forwarding path.


docker-agent (commented):
### ⚠️ Review incomplete

aheritier (approved):

PR #4218, opened 2026-09-10
feat(sandbox): add kubectl and aws cli to sbx-templates image

maxcleme:
Installing these at each sandboxes creation is quite annoying, I'd say it makes sense to include these in the template. WDYT?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,6 +86,7 @@ apt-get clean
 rm -rf /var/lib/apt/lists/*
 EOF
 COPY --chmod=0755 --from=docker/mcp-gateway:v2 /docker-mcp /usr/local/lib/docker/cli-plugins/docker-mcp
+COPY --chmod=0755 --from=registry.k8s.io/kubectl:v1.37.0 /bin/kubectl /usr/local/bin/kubectl
 USER agent
 ENV DOCKER_AGENT_NO_TOUR=1 \
     DOCKER_AGENT_HIDE_TELEMETRY_BANNER=1 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,6 +87,9 @@ rm -rf /var/lib/apt/lists/*
 EOF
 COPY --chmod=0755 --from=docker/mcp-gateway:v2 /docker-mcp /usr/local/lib/docker/cli-plugins/docker-mcp
 COPY --chmod=0755 --from=registry.k8s.io/kubectl:v1.37.0 /bin/kubectl /usr/local/bin/kubectl
+COPY --chmod=0755 --from=amazon/aws-cli:2.36.42 /usr/local/aws-cli/ /usr/local/aws-cli/
+RUN ln -s /usr/local/aws-cli/v2/current/bin/aws /usr/local/bin/aws && \
+    ln -s /usr/local/aws-cli/v2/current/bin/aws_completer /usr/local/bin/aws_completer
 USER agent
 ENV DOCKER_AGENT_NO_TOUR=1 \
     DOCKER_AGENT_HIDE_TELEMETRY_BANNER=1 \
--- a/docs/configuration/sandbox/index.md
+++ b/docs/configuration/sandbox/index.md
@@ -207,6 +207,7 @@ layers onto `docker/sandbox-templates:shell-docker` and adds:
 - The `docker-agent` binary.
 - `vim` and `tmux`, for interactive debugging inside the VM.
 - `kubectl`, for interacting with Kubernetes clusters from inside the VM.
+- The **AWS CLI** (`aws`), for working with AWS from inside the VM.
 - The **`docker-mcp`** Docker CLI plugin, installed at `~/.docker/cli-plugins/docker-mcp`.
 
 The image carries the label `com.docker.sandboxes.flavor=docker-agent-docker`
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -86,6 +86,7 @@ apt-get clean
 rm -rf /var/lib/apt/lists/*
 EOF
 COPY --chmod=0755 --from=docker/mcp-gateway:v2 /docker-mcp /usr/local/lib/docker/cli-plugins/docker-mcp
+COPY --chmod=0755 --from=registry.k8s.io/kubectl:v1.37.0 /bin/kubectl /usr/local/bin/kubectl
```
docker-agent, line 89: **[medium/security] Mutable image tag used for kubectl COPY source — no digest pin**

The `COPY --from=registry.k8s.io/kubectl:v1.37.0` instruction references a mutable version tag. If this tag is ever overwritten on the registry, future builds will silently pull a different binary, which is a supply-chain integrity risk. While `registry.k8s.io` is the official, well-governed Kubernetes registry and the realistic risk of tag mutation for a versioned release tag is low, pinning to an immutable digest fully eliminates the risk.

To fix, pin the image to a specific digest alongside the version tag:
```
COPY --chmod=0755 --from=registry.k8s.io/kubectl:v1.37.0@sha256:<digest> /bin/kubectl /usr/local/bin/kubectl
```
You can obtain the digest by running:
```sh
crane digest registry.k8s.io/kubectl:v1.37.0
```
or by checking the image manifest on the registry.

> Note: The existing `docker/mcp-gateway:v2` line above uses the same mutable-tag pattern, so this is consistent with current project practice — but both lines carry this risk.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

Dockerfile
```diff
@@ -86,6 +86,10 @@ apt-get clean
...
+COPY --from=amazon/aws-cli:2.36.42 /usr/local/aws-cli/ /usr/local/aws-cli/
```
docker-agent: **[medium] Missing `--chmod` on aws-cli COPY — file permissions are uncontrolled**

The kubectl COPY on the line above uses `--chmod=0755`, but the aws-cli COPY does not:

```
COPY --chmod=0755 --from=registry.k8s.io/kubectl:v1.37.0 /bin/kubectl /usr/local/bin/kubectl
COPY --from=amazon/aws-cli:2.36.42 /usr/local/aws-cli/ /usr/local/aws-cli/   ← no --chmod
```

Without `--chmod`, Docker copies the files with whatever permissions they carry inside the source image. If those permissions are wider than necessary (e.g. world-writable) or narrower than expected (e.g. not executable), the result is non-deterministic and inconsistent with the adjacent pattern. The symlinks created in the subsequent `RUN` layer point into this copied tree, so executability of `aws` and `aws_completer` depends entirely on the source image's permissions.

Adding `--chmod=0755` enforces a predictable, least-privilege permission set and is consistent with how kubectl is handled:

```suggestion
COPY --chmod=0755 --from=amazon/aws-cli:2.36.42 /usr/local/aws-cli/ /usr/local/aws-cli/
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):

One medium-severity finding was identified in the new aws-cli integration lines.

> **Note:** The `amazon/aws-cli:2.36.42` Docker Hub tag is also pinned by a mutable tag (same pattern as the open thread on `registry.k8s.io/kubectl:v1.37.0` — deduplicated against that existing thread). Consider pinning both to immutable digests together.

dgageot (dismissed):

dgageot (dismissed):

Dockerfile
aheritier: I think it makes sense @maxcleme

Dockerfile
aheritier: I am afraid of the compatibility. K8s is still ensuring the compat with 2 minor versions up and down ?

Dockerfile
docker-agent: Thanks for the confirmation! Adding `--chmod=0755` will ensure the aws-cli binaries have consistent, predictable permissions regardless of how they're packaged upstream.

Dockerfile
docker-agent: Good point about Kubernetes version skew policy. You're right that digest pinning locks you to the exact binary, preventing even patch-level updates that might improve compatibility or fix bugs.

However, digest pinning doesn't prevent you from updating — it just makes updates explicit. When you need to bump kubectl (for compatibility, security, or features), you'd update both the tag and digest together. The trade-off is: mutable tags give silent updates (convenient but risky), while digest pins give explicit updates (safer but require active maintenance).

For a kubectl client used by Docker Agent, the skew policy means v1.37.0 works with clusters running 1.35–1.38. If you're comfortable with the supply-chain risk from mutable tags (especially since registry.k8s.io is well-governed), and you value the flexibility of automatic patch updates, keeping the tag-only reference is a reasonable choice. Just be aware that if the tag is ever republished, your builds will silently use the new artifact.

docker-agent (commented):
### Assessment: 🟢 APPROVE

- [medium] Dockerfile:91 — Symlink chain through `current` may be broken — COPY --from does not dereference symlinks (confidence: weak 40/100)
- [medium] Dockerfile:90 — AWS CLI binary may be incompatible with base image's libc or target architecture (confidence: weak 40/100)

> **Note**: A mutable-tag supply-chain finding for `amazon/aws-cli:2.36.42` (the same class of risk as the open thread on `kubectl:v1.37.0` at line 89) was identified but suppressed by deduplication since it is already being tracked.

aheritier (approved):

PR #4222, opened 2026-09-10
revert: remove foreign-OS path rejection from filesystem tool

dgageot:
The previous commit introduced logic to detect and reject file paths intended for a different operating system (e.g., Windows-style paths on Linux), along with working-directory instructions and path resolution hints surfaced to the agent. After further consideration, this approach adds unnecessary complexity and makes the filesystem tool more brittle without a clear benefit — legitimate paths can match the heuristics, and the agent already handles working-directory context through other means.

This reverts commit cec4ba4923a8a07393e7730392982c86ab159218. The filesystem tool is restored to its prior behavior: it accepts any path without OS-validation gates. The related test cases covering foreign-path rejection are removed, and the test cassettes are updated to reflect the simpler tool output format.

No behavior changes outside the reverted feature. Builds, tests, and linter all pass.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34482889246) for details. Re-request a review from `docker-agent` to retry.

gtardif (approved):

PR #4235, opened 2026-09-11
feat: add secure HTTP relay packages

dgageot:
Two new packages handle the different relay concerns that had previously been mixed together. `pkg/config/httprelay` provides a thin, generic handler that fetches agent configuration from a trusted URL callback — the caller decides what URL is valid, and the handler ensures it is well-formed, uses an explicit scheme, and carries no embedded credentials. `pkg/modelsgateway/relay` handles the richer model-API forwarding path: request validation, upstream credential injection, SSE streaming with line-by-line flush, and optional stream-completion verification.

Both packages share a small set of design constraints applied consistently. The `Target`/`Source` callbacks are the only entry points for resolving upstream URLs, which means no user-supplied URL ever reaches the outbound client. Request and response headers are passed through strict allowlists that exclude `Authorization`, `Cookie`, credential-style keys, and the internal `X-Cagent-Forward` header. Body sizes are bounded before any bytes are sent upstream. Redirects are unconditionally refused at the `http.Client` level so that a redirect response from a misbehaving upstream cannot replay credentials to a third-party host. Errors sent to browsers are deliberately vague; detail stays on the server side.

The relay package adds one further layer: when the upstream returns 401, the token is invalidated and fetched exactly once more. If the fresh token matches the rejected one, the request fails immediately rather than looping. Streaming responses are forwarded line by line with an `http.Flusher` flush on each SSE boundary, and an optional `StreamObserver` can validate that the stream ended in a recognized completion state — emitting a caller-supplied error event if it did not.

Validation: `task lint`, `task test`, `task build` all pass.

### DIFF
--- a/pkg/config/httprelay/httprelay.go
+++ b/pkg/config/httprelay/httprelay.go
@@ -1,0 +1,155 @@
+// Package httprelay exposes a remote agent configuration through an HTTP handler.
+package httprelay
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strconv"
+	"time"
+)
+
+const (
+	defaultMaxBytes = 1 << 20
+	defaultTimeout  = 30 * time.Second
+)
+
+// Source resolves the trusted upstream URL for a request. Implementations must
+// not return unvalidated user input.
+type Source func(*http.Request) (*url.URL, error)
+
+// Option configures a Handler.
+type Option func(*Handler)
+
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34581616054) for details. Re-request a review from `docker-agent` to retry.

aheritier (approved):

PR #4238, opened 2026-09-11
feat(share): sign a DSSE-wrapped in-toto statement, not just the YAML bytes

gtardif:
`share push --key` signed the agent YAML directly, so the signature proved "a key holder published these bytes" and nothing else. It carried no metadata: no indication of where the artifact was published or when, and re-tagging a signed layer under a new reference verified happily.

The signature now covers an [in-toto Statement v1](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md) carried in a [DSSE envelope](https://github.com/secure-systems-lab/dsse), recorded in clear in a new `io.docker.agent.attestation` annotation, with the statement's `predicateType` advertised alongside it in `in-toto.io/predicate-type` so a consumer can tell what an attestation is about without base64-decoding it (the same annotation key BuildKit puts on its in-toto attestation layers).

```json
{
  "_type": "https://in-toto.io/Statement/v1",
  "subject": [
    {
      "name": "index.docker.io/myorg/agent:v1",
      "digest": { "sha256": "889871ef…" }
    }
  ],
  "predicateType": "https://docker.com/docker-agent/share/publication/v1",
  "predicate": {
    "registry": "index.docker.io",
    "repository": "myorg/agent",
    "tag": "v1",
    "created": "2026-09-11T08:30:00Z"
  }
}
```

Three checks have to hold: the signature is valid over the exact stored payload, the statement is an in-toto statement we understand, and a subject digest matches the YAML that was pulled. The digest binds statement to payload, so neither a swapped layer nor a swapped statement verifies. The subject binds it to a location: `Verification.CheckSubject` compares the attested reference with the one actually requested, which is what makes a copied artifact detectable. `share pull` and `ocisource` both call it.

`--key` and `--encrypt` are unchanged — no CLI changes.

## Why interoperable instead of bespoke

The wire format is the one cosign and in-toto tooling consume rather than a custom JSON blob: a plain DSSE envelope (`payload`, `payloadType`, `signatures[].sig`) of type `application/vnd.in-toto+json`, whose signatures cover the DSSE pre-authentication encoding `PAE(payloadType, body)` — replacing the previous NUL-delimited `domainInput` framing for signatures. The annotation value is exactly an `application/vnd.dsse.envelope.v1+json` object (exported as `protect.EnvelopeMediaType`), and subject digests are bare hex, as in-toto requires.

A standalone Python verifier using nothing but `cryptography` validates the artifacts this produces, and the PAE is pinned by the DSSE spec's own test vector.

## Strict where it binds, lenient where it describes

This is the core design decision, and it is worth reviewing explicitly:

- **The statement is strict.** `_type`, `subject` (each entry needs a name and a well-formed sha256 digest) and `predicateType` are what tie an artifact to its bytes and its location, so unknown fields there are refused rather than ignored.
- **The predicate is lenient.** Unknown fields — and known fields carrying an unexpected JSON type — are surfaced as opaque key/values (`Predicate.Unknown`) instead of failing the artifact, and are re-emitted verbatim on marshal so a round-trip never drops them. An unknown `predicateType` yields "signature valid, predicate not understood", an honest outcome rather than a hard failure.

So publication metadata can be added later without breaking already-deployed verifiers. A test asserts exactly that, and it was also checked end to end against Docker Hub with a publisher emitting extra predicate fields — the older parser verified the artifact **and** displayed the unknown fields.

The `predicateType` URI is the version; there is no separate version integer to keep in sync. A breaking change to the predicate shape gets a new URI, and old verifiers keep checking the signature and the subject binding while reporting the predicate as not understood.

## Notes on the implementation

- **No new dependencies.** PAE is five lines and the two structs are plain JSON; pulling in `github.com/in-toto/attestation` (protobuf) plus a DSSE library would add real dependency weight for types this small. Conformance is pinned by tests instead: the spec's own PAE vector, the envelope shape, and a verification done by hand from the raw primitive.
- The payload is verified **before** it is parsed, and the signature covers the bytes as stored, never a re-marshaled copy: DSSE explicitly forbids re-reading the payload out of the envelope after verification, and canonicalization must not depend on this process's JSON encoder.
- `keyid` is recorded as a hint but never consulted: every signature in the envelope is tried, so a wrong or missing keyid neither rejects a valid signature nor lends weight to an invalid one. Both base64 alphabets (standard and URL-safe, padded or not) are accepted on the way in.
- `in-toto.io/predicate-type` is likewise a filter hint, not evidence: it sits outside the signature, so verification reports the `predicateType` from the signed statement and ignores the annotation. Forging or stripping it changes no outcome — asserted by tests, and checked against a live registry by rewriting the annotation on a pushed manifest.
- The creation date in the predicate is the same value as `org.opencontainers.image.created`, so the advertised and signed dates cannot drift.
- Signatures are the only thing that moved to PAE framing; the AEAD path is untouched, its algorithm-binding header is now `encryptAAD` (same bytes as the old `domainInput("encrypt", alg, nil)`).
- Encrypt mode with a symmetric secret still records only the AEAD copy — it is proof by itself — so such artifacts carry no signature and no attestation, and `CheckSubject` is a no-op for them.

Tampering with the layer now surfaces as `ErrStatementMismatch` rather than `ErrInvalidSignature`: the signature over the statement is still valid, it is the digest that no longer matches. Tests assert the new error accordingly.

## What is *not* covered, documented rather than glossed over

The subject digest covers the YAML layer, not the manifest. Annotations are part of the manifest, so a manifest digest could not be recorded inside one without a circular dependency. The consequence is that only the agent YAML and the statement's own contents are authenticated — the surrounding annotations (`authors`, `licenses`, `revision`, `tags`, the advertised `created`) can be rewritten by anyone who can push to the repository while the signature still verifies. I confirmed this against a live registry.

The docs now list exactly which annotations are covered and which are not, and say not to build policy on the latter; the package doc carries the same caveat for embedders.

**Open question for reviewers:** authenticating the whole manifest means publishing the envelope as a referring artifact (OCI Referrers API) so the subject can be the manifest digest. That is also what would make the attestation discoverable by `cosign verify-attestation` — the bytes here are conformant, but tools that expect attestations as referring artifacts will not find one in an annotation. Docker Hub does support the Referrers API (`GET /v2/<repo>/referrers/<digest>` returns 200), so it is viable. I left it out as a larger, separable change; the format itself survives such a migration unchanged, since only the storage layout differs.

## Breaking change

This changes the signature format. `io.docker.agent.signature` is gone, so artifacts signed by an earlier version carry no attestation and are rejected with "artifact is neither signed nor encrypted" instead of silently verifying, and older clients reject artifacts produced here the same way. **Both directions fail closed** (verified against Docker Hub in both directions). Re-push signed artifacts.

## Testing

- `task lint` — 0 issues; `go test ./...` — green.
- New tests in `pkg/protect` (envelope/statement/predicate, PAE vector, keyid handling, base64 variants, strict-statement rejections, forward-compat), plus `pkg/oci`, `pkg/config/ocisource` and `cmd/root`.
- End-to-end against Docker Hub with Ed25519, ECDSA and RSA keys, in sign and `--encrypt` modes: signed artifacts verify, the re-tag attack is detected by both `share pull` and an independent Python verifier, a symmetric-secret encrypt-only artifact carries no attestation, and a wrong key is rejected.
- Verified the tests are load-bearing: stubbing out the signature check fails exactly the tests that assert it, and nothing else.

### DIFF
--- a/pkg/protect/attestation.go
+++ b/pkg/protect/attestation.go
@@ -1,0 +1,423 @@
+package protect
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/name"
+)
+
+const (
+	// PayloadType is the DSSE payloadType of the signed body: an in-toto
+	// Statement. It is part of the pre-authentication encoding (see
+	// paeEncode), so it cannot be swapped without invalidating the signature.
+	PayloadType = "application/vnd.in-toto+json"
+	// StatementType is the in-toto Statement v1 `_type`.
+	StatementType = "https://in-toto.io/Statement/v1"
+	// PredicateTypePublication identifies the predicate holding the
+	// publication metadata of a shared agent. The URI is the version: a
+	// breaking change to the predicate shape gets a new URI, and verifiers
+	// that do not know a URI report the predicate as not understood instead of
+	// rejecting the artifact.
+	PredicateTypePublication = "https://docker.com/docker-agent/share/publication/v1"
+	// digestAlgorithm is the only digest algorithm we produce in a subject's
+	// digest set. Values are bare hex, as in-toto requires.
+	digestAlgorithm = "sha256"
+)
+
+var (
+	ErrMalformedAttestation = errors.New("malformed attestation annotation")
+	ErrUnknownStatementType = errors.New("unsupported in-toto statement type")
+	ErrStatementMismatch    = errors.New("attestation does not describe this artifact")
+	ErrSubjectMismatch      = errors.New("attested subject does not match the reference being read")
+)
+
+// Envelope is a DSSE envelope (Payload holds the base64 in-toto Statement).
+// It is stored verbatim in [AnnotationAttestation].
+type Envelope struct {
+	// Payload is the base64 SERIALIZED_BODY. Verification covers the exact
+	// bytes it decodes to, never a re-serialized copy.
+	Payload string `json:"payload"`
+	// PayloadType must be [PayloadType]; it is authenticated through the PAE.
+	PayloadType string `json:"payloadType"`
+	// Signatures holds one entry per signer. At least one must verify.
+	Signatures []Signature `json:"signatures"`
+}
+
+// Signature is one DSSE signature over PAE(payloadType, body).
+type Signature struct {
+	// KeyID is an unauthenticated hint identifying the signing key. It must
+	// never drive a security decision: verification tries every signature.
+	KeyID string `json:"keyid,omitempty"`
+	// Sig is the base64 raw signature (or MAC).
+	Sig string `json:"sig"`
+}
+
+// Subject is one in-toto subject: a name and a digest set of bare-hex digests.
+type Subject struct {
+	Name   string            `json:"name"`
+	Digest map[string]string `json:"digest"`
+}
+
+// Statement is an in-toto Statement v1: the signed body of the envelope.
+//
+// The security-critical part is parsed strictly: `_type`, `subject` (each with
+// its digest set) and `predicateType` bind the artifact to a location and to
+// its bytes, so unknown fields there are refused rather than ignored. The
+// predicate is parsed leniently — see [Predicate].
+type Statement struct {
+	// Type is always [StatementType].
+	Type string
+	// Subject lists the artifacts the predicate is about. A statement attests
+	// all of its subjects equally: a match against any entry is a match.
+	Subject []Subject
+	// PredicateType names the predicate shape, and is its version.
+	PredicateType string
+	// Predicate is the publication metadata, populated only when
+	// PredicateUnderstood is set.
+	Predicate Predicate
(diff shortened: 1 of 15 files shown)


trungutt (approved):

pkg/protect/attestation.go
```diff
@@ -0,0 +1,423 @@
...
+//
+// The statement envelope is strict — unknown fields, a foreign `_type` or a
+// subject without a usable digest are refused. The predicate is not: an
+// unknown predicateType or unknown predicate fields yield a statement whose
+// metadata is (partly) opaque, not an error.
+func ParseStatement(raw []byte) (Statement, error) {
+	var wire statementWire
+	decoder := json.NewDecoder(bytes.NewReader(raw))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&wire); err != nil {
```
aheritier, line 205: Non-blocking: `ParseStatement` and `ParseEnvelope` each perform a single `decoder.Decode` without verifying EOF, so they accept a valid JSON value followed by a second JSON value (for example, `<valid JSON> {"extra":true}`). This is not a signature/trust bypass, but it conflicts with the documented strict-parser posture. Could we reject trailing values and add malformed-input coverage?

PR #4241, opened 2026-09-11
fix: always clone RuntimeConfig in LoadWithConfig to prevent race

### LINKED ISSUE #4239: fix(teamloader): LoadWithConfig races on shared RuntimeConfig
## Summary

The `ci / test-race` job on `main` is intermittently failing due to a real data race in `pkg/teamloader.LoadWithConfig`. The behavior is schedule-dependent, but the underlying concurrent access is deterministic—not runner or network flakiness.

## Observed CI behavior

Recent `main` runs have alternated between pass and failure at effectively the same head, including failures in runs `34603115677`, `34601273521`, and `34579008635`, followed by a passing run `34603490548`.

The race report consistently points to concurrent access in `LoadWithConfig`:

- Read: `pkg/teamloader/teamloader.go:297` (`config.MergeGlobalProviders` reading `runConfig.Providers`)
- Write: `pkg/teamloader/teamloader.go:378` (mutating `runConfig.Providers`; the function also mutates `Models` and `ProviderRegistry`)

A representative race trace is:

```
WARNING: DATA RACE
Read at ... by goroutine 473:
  LoadWithConfig() teamloader.go:297
  strict_test.go:185

Previous write at ... by goroutine 472:
  LoadWithConfig() teamloader.go:378
  strict_test.go:178
```

## Trigger

`pkg/teamloader/strict_test.go`'s `TestLoadExternalAgent_InheritsLoaderOptions` creates one shared `*config.RuntimeConfig` and passes it into two `t.Parallel()` subtests. Each calls `teamloader.Load`, with recursive loading through `loadExternalAgent`.

`LoadWithConfig` mutates the supplied `RuntimeConfig`. It has a partial clone safeguard when a caller provides a working directory different from `runConfig.WorkingDir`, but does not clone when concurrent callers share the same config and working directory.

The race detector may attribute the failure to different concurrently-running tests in the package; those varying test names are collateral, rather than separate root causes.

## Impact

- `ci / test-race` on `main` is unreliable.
- The current implementation may have a production thread-safety gap if API-server sessions or other callers can invoke `LoadWithConfig` concurrently with the same `RuntimeConfig` and working directory.

## Proposed remediation

1. **Immediate test isolation:** update `TestLoadExternalAgent_InheritsLoaderOptions` so each parallel subtest creates or receives its own `RuntimeConfig`; audit nearby teamloader tests for equivalent shared-config patterns.
2. **Decide and enforce the API contract:** if `LoadWithConfig` is expected to support concurrent callers sharing one config, defensively clone the `RuntimeConfig` on every call (or otherwise stop mutating the caller-owned object). If not, document and enforce the ownership/concurrency requirement at call sites.
3. Re-run and monitor `ci / test-race` after this repair. Track the independent E2E chat-server timeout and historical Docker-cancel race separately if they persist.

## Context

PR #4228 addressed earlier `pkg/evaluation` and `pkg/tui` race findings, but did not modify `pkg/teamloader` or the relevant E2E tests; the teamloader race continued afterward.

dgageot:
## Summary

- clone `RuntimeConfig` for every `LoadWithConfig` invocation so resolved models, providers, provider registry, and encrypted config remain per-load
- add direct regression coverage for concurrent loads sharing one runtime config
- clarify comments describing per-load state and working-directory overrides

Closes #4239.

## Validation

- `task test`
- `task lint`
- `task build`
- `go test -race -shuffle=on -count=10 ./pkg/teamloader/...`

### DIFF
--- a/pkg/config/runtime.go
+++ b/pkg/config/runtime.go
@@ -27,9 +27,9 @@ type RuntimeConfig struct {
 	modelsDevStoreOnce     sync.Once
 
 	// ProviderRegistry instantiates model providers for toolsets that build
-	// providers at load time (e.g. RAG embeddings/reranking). It is populated
-	// by the team loader with the same registry used for agent models. When
-	// nil, ProviderRegistryOrDefault falls back to provider.DefaultRegistry.
+	// providers at load time (e.g. RAG embeddings/reranking). The team loader
+	// sets it on its per-load clone with the registry used for agent models.
+	// When nil, ProviderRegistryOrDefault falls back to provider.DefaultRegistry.
 	ProviderRegistry *provider.Registry
 }
 
--- a/pkg/teamloader/teamloader.go
+++ b/pkg/teamloader/teamloader.go
@@ -56,11 +56,9 @@ type loadOptions struct {
 
 type Opt func(*loadOptions) error
 
-// WithWorkingDir overrides the working directory toolsets are built with,
-// without touching the caller's RuntimeConfig. Callers that share one
-// RuntimeConfig across concurrent loads (the API server, one per session)
-// need this to keep each session's shell, filesystem and git tools rooted in
-// that session's directory.
+// WithWorkingDir overrides the working directory toolsets are built with.
+// Callers that share one RuntimeConfig across sessions use it to root each
+// session's shell, filesystem and git tools in that session's directory.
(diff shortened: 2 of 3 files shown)


docker-agent (commented):
### ⚠️ Review incomplete

aheritier (approved):

PR #4244, opened 2026-09-11
feat: add GET /api/sessions?active=true for lightweight attached-session listing

dgageot:
`GET /api/sessions` reads the full session history from disk on every call. On a project with many past sessions this can easily take over 15 seconds, which makes it impractical for supervising clients — like a board UI — that need to periodically poll for active TUI tabs and aggregate their streaming state.

This adds an `?active=true` query parameter to the endpoint. When set, the handler skips history entirely and returns only the runtimes that are currently attached in memory, each one carrying `working_dir` and the live streaming flag. The response is immediate because no disk I/O is involved. The existing unfiltered path is unchanged, so nothing breaks for callers that need the full history.

Validation: `go test -race ./pkg/server ./pkg/api ./pkg/runtime`, `task lint`, `task build`.

### DIFF
--- a/docs/features/api-server/index.md
+++ b/docs/features/api-server/index.md
@@ -59,7 +59,7 @@ For an agent loaded from a remote HTTP(S) configuration source, endpoints that n
 
 | Method   | Path                                | Description                                             |
 | -------- | ----------------------------------- | ------------------------------------------------------- |
-| `GET`    | `/api/sessions`                     | List all sessions                                       |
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34626115050) for details. Re-request a review from `docker-agent` to retry.

aheritier (approved):

PR #4255, opened 2026-09-14
refactor(tui): consolidate tab UI ownership

dgageot:
The TUI previously tracked per-tab state through six separate maps keyed by tab index — editors, chat pages, session state, pending restores, sidebar state, and stashed dialogs — alongside three top-level aliases (`activeEditor`, `activeChatPage`, `activeSession`) that mirrored whatever the current tab pointed to. This scattered ownership made it hard to reason about which fields belonged to a tab's lifetime vs. the model's lifetime, and easy to miss a map when adding or removing a tab.

This refactor introduces `tabModel`, a small struct that bundles all six fields, and replaces the maps with a single `[]tabModel` slice. The three active-UI aliases are removed; callsites now reach through `m.tabs[m.activeTabIdx]` (or the helper that returns a pointer to that entry), which makes ownership explicit. Runtime application binding — the step that wires a live `Runtime` into a tab's fields after restore — is deliberately left as a separate pass so that staged ordering of pending restores, extension callbacks, close-last-tab failure handling, and side-effect timing are unchanged.

A new `tab_lifecycle_test.go` covers the cases most likely to regress: draft and component preservation across tab switches, atomicity of a failed tab switch, lazy restore success and failure paths, startup restore, tab closing, the distinction between runtime tab IDs and persisted session IDs, and settings/cleanup behavior on tabs that have not yet been loaded.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34813232565) for details. Re-request a review from `docker-agent` to retry.

trungutt (approved):

PR #4257, opened 2026-09-14
feat: toolset graph traversal and explicit RuntimeHandler identity

dgageot:
Previously, toolset capabilities were discovered with a single type-assertion (`tools.As`), which only checked the immediate toolset. Composite or decorator-wrapped toolsets silently dropped inner capabilities, so handlers like elicitation, sampling, and OAuth were never wired up for nested providers. Separately, the tool dispatcher selected host-owned handlers by matching the model-facing tool *name* against a `Handlers` map, meaning any external tool that happened to share a name with a privileged handler (e.g. `transfer_task`) could shadow or inadvertently invoke it.

This change introduces a `Walk`/`Find`/`FindAll` graph API in `pkg/tools/graph.go` that traverses a toolset tree depth-first, visiting decorator wrappers before composite children and visiting shared nodes only once to break cycles. All previous `tools.As` call sites are replaced with `Find` or `FindAll` so capabilities nested anywhere in a toolset graph are discovered automatically. `ConfigureHandlers` now walks the full graph and wires handlers onto every matching provider, not just the outermost one.

To prevent name-collision attacks, `Tool` gains a `RuntimeHandler` field. Host-owned handlers are registered under opaque identifiers in `Dispatcher.Handlers`, and a tool's `RuntimeHandler` field must carry the matching identifier for the dispatcher to route to it. Tool names are never consulted when selecting entries from that map, so a user-supplied tool cannot invoke a privileged handler regardless of its name.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34813303787) for details. Re-request a review from `docker-agent` to retry.

gtardif (approved):

PR #4259, opened 2026-09-14
perf(tui): avoid redundant renders on height-only resize

dgageot:
When the TUI is resized, `SetSize` is called on every message component even when only the terminal height changes. Because message rendering depends only on width, those renders produced identical output and discarded the work immediately — but they still allocated fresh buffers for every cached message.

The fix is two-part. First, each message now tracks the width it was last rendered at and skips re-rendering when `SetSize` is called with the same width, returning the cached view directly. Second, the transcript pre-allocates its lines slice from the known total of cached per-message heights rather than growing it one line at a time, eliminating the dominant allocation on every layout pass.

Benchmarked on a 500-message mixed-Markdown history (Apple M4 Max, Go 1.27, `GOMAXPROCS=4`, `-benchtime=20x`, six alternating before/after runs):

```text
height-only resize  median latency  14.9 ms  →  0.13 ms
                    allocated       8.42 MiB →  173 KiB
                    allocations     291 011  →  40
```

A CLI smoke test of 40 live height resizes (`--cpuprofile`/`--memprofile`) confirmed the direction: CPU samples 1.17 s → 0.60 s, total allocated 565 MB → 146 MB. Width-resize and streaming-response paths show no measurable change, which is expected — width genuinely requires a re-render.

Both commits are behavior-preserving: child `SetSize`/render ordering and scroll-reset behavior are unchanged, and the existing resize test suite was extended to cover the new caching logic.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34817150761) for details. Re-request a review from `docker-agent` to retry.

gtardif (approved):

PR #4260, opened 2026-09-14
feat(lint): add HookBuiltinsDocumented cop and document missing builtins

dgageot:
Several hook builtins shipped by the runtime — `http_post`, `limit_large_tool_results`, `safer_shell`, and `snapshot` — were absent from both `agent-schema.json`'s type description and the hooks reference table in `docs/configuration/hooks/index.md`. This made them invisible to schema-driven tooling and to anyone reading the documentation, even though they work fine at runtime.

This PR introduces the `Lint/HookBuiltinsDocumented` cop, which statically verifies that every exported string constant in `pkg/hooks/builtins/` (i.e. every wire name) appears in both documentation surfaces. It runs over `builtins.go`'s directory, parses the schema's `HookDefinition.type.description` for single-quoted names, and scans the hooks reference for rows in the builtins table — failing the lint run if anything is missing. Two focused unit tests cover the schema parser and the Markdown table scanner independently.

The four previously-undocumented builtins are also added to both surfaces in the same commit, so the new cop passes on the existing codebase immediately. No runtime behaviour changes.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34817659918) for details. Re-request a review from `docker-agent` to retry.

trungutt (approved):

gtardif (approved):

PR #4261, opened 2026-09-14
feat(lint): add EnvironmentVariablePrefix cop and migrate pprof to DOCKER_AGENT_PPROF_ADDR

dgageot:
The codebase historically used `CAGENT_*` environment variables, which leaked an internal project name into the public interface. This adds an `EnvironmentVariablePrefix` custom cop to the agent-config linter that enforces the `DOCKER_AGENT_` prefix for any new public environment variable references in YAML configs, while exempting the private askpass IPC variables and their tests.

As a concrete migration, `CAGENT_PPROF_ADDR` is replaced by `DOCKER_AGENT_PPROF_ADDR` as the canonical name for the pprof listener address. The old name is kept as a legacy fallback so existing deployments continue to work without changes. The cop's exemption list covers the askpass socket variables, which are internal IPC details that intentionally keep the `CAGENT_` prefix. Docs for the API server and CLI are updated to reflect the new variable name.

No breaking changes: anything reading `CAGENT_PPROF_ADDR` today will continue to work.

## Validation

- `task lint` — passes
- `task test` — passes
- `task build` — passes


docker-agent (commented):

lint/environment_variable_prefix.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		literals := stringLiterals(p.File)
+		values := literalsByValue(literals)
+		for lit, value := range literals {
+			if !isLegacyEnvVar(value) || internalCagentEnvVars[value] {
+				continue
+			}
+			current := currentEnvPrefix + strings.TrimPrefix(value, legacyEnvPrefix)
+			if values[current] {
```
docker-agent, line 44: **[medium] Any `DOCKER_AGENT_*` string literal suppresses the alias check — including flag help text**

The cop's alias check (`values[current]`) is satisfied by **any** string literal in the file whose value equals the expected `DOCKER_AGENT_*` name, regardless of its syntactic role. `literalsByValue` collects every `*ast.BasicLit` of kind `token.STRING`, including flag description strings.

In `cmd/root/api.go` today the flag registration contains:

```go
cmd.PersistentFlags().StringVar(&flags.pprofAddr, "pprof-addr", "",
    "TCP host:port ... also set via DOCKER_AGENT_PPROF_ADDR")
```

That description string `"... also set via DOCKER_AGENT_PPROF_ADDR"` — substring or full value — is itself a string literal. If a future contributor removes the `envPprofAddr` constant (`"DOCKER_AGENT_PPROF_ADDR"`) but leaves the flag help text unchanged, the cop will still find `"DOCKER_AGENT_PPROF_ADDR"` in `values` (because the description string contains a literal that, after `Unquote`, equals exactly `"DOCKER_AGENT_PPROF_ADDR"` when `StringVar` is passed the string from the constant — wait, the description actually uses the string `"... also set via DOCKER_AGENT_PPROF_ADDR"` which does NOT equal `"DOCKER_AGENT_PPROF_ADDR"` as an exact value).

**Correcting scope:** after re-reading `stringLiterals`, it collects only literals whose `Unquote`d value exactly matches `current` (i.e., the full `"DOCKER_AGENT_PPROF_ADDR"` string). The flag description `"TCP host:port ... also set via DOCKER_AGENT_PPROF_ADDR"` does NOT equal `"DOCKER_AGENT_PPROF_ADDR"`, so it would **not** satisfy `values[current]`. However, any string literal whose exact value is `"DOCKER_AGENT_PPROF_ADDR"` — whether it is a constant, a variable assignment, or a bare string passed directly to `os.Getenv("DOCKER_AGENT_PPROF_ADDR")` — would satisfy the check.

The real false-negative risk: a call like `os.Getenv("DOCKER_AGENT_PPROF_ADDR")` present in the same file (without a named constant) would suppress the offense for `"CAGENT_PPROF_ADDR"`. This is arguably correct behaviour (the migration literal exists), but it also means the check conflates "a constant declaring the alias" with "an arbitrary usage of the alias string." If the `os.Getenv` call is removed and only the legacy constant remains, the cop correctly fires — so the practical risk is limited.

**Recommendation:** if the intent is to require a *named constant* alias rather than any string occurrence, restrict `literalsByValue` to literals that are the direct value of a `const` declaration by inspecting `*ast.GenDecl` / `*ast.ValueSpec` nodes with `token.CONST` kind instead of collecting all string literals across the file.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

trungutt (approved):

gtardif (approved):

PR #4262, opened 2026-09-14
feat(lint): enforce TUI key bindings

dgageot:
The TUI code in `pkg/tui` compares key events using raw `msg.String()` comparisons — either inside `switch` statements or with `==`. This is error-prone: named keys and modified keys can have multiple string representations, so a comparison that looks correct can silently miss events. A linter check makes the pattern impossible to sneak back in.

This PR adds a `TUIKeyBindings` custom cop that detects both switch-based and equality-based `KeyPressMsg.String()` comparisons anywhere under `pkg/tui`. It registers alongside the existing cops in `lint/main.go` and ships with a focused test suite covering true-positive cases (switch, `==`, `!=`) and true-negative cases (comparisons outside `pkg/tui`, non-key string calls). All existing call-sites in the package are migrated to `key.Matches` with the appropriate `key.Binding` values, so the diff reflects zero behavior change.

`task build` and `task lint` pass cleanly. Full `task test` passes for all TUI tests; the only failures observed in the isolated worktree are environment-only Docker Model Runner issues in `TestLoadExamples` (port 12434 already in use, `ai/qwen3` model not pulled) that are unrelated to this change.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34819091942) for details. Re-request a review from `docker-agent` to retry.

gtardif (approved):

trungutt (approved):

PR #4265, opened 2026-09-14
feat(lint): enforce shared tool argument decoding

dgageot:
Tool handlers receive model-generated JSON as a raw `Arguments` string. Several call sites decoded it directly with `json.Unmarshal`, bypassing the `aijson` repair layer that fixes truncated or otherwise malformed model output, and the telemetry that tracks how often repairs are needed.

This adds a `Lint/ToolArgumentsViaAIJSON` cop to the in-tree linter. It detects any `json.Unmarshal([]byte(call.Function.Arguments), ...)` pattern and reports it as an error, directing callers to `tools.UnmarshalToolArguments` instead. The cop is scoped to exclude packages that legitimately decode other JSON (model-layer, TUI, renderer example, and a known sampling helper). All existing violations — spread across the runtime, WASM adapter, ACP, built-in tools, MCP, A2A, and the Go library example — are migrated in the same commit.

No behaviour changes are expected at runtime; `tools.UnmarshalToolArguments` already wraps `aijson` and was the intended path for all of these sites.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34828561899) for details. Re-request a review from `docker-agent` to retry.

trungutt (approved):

PR #4266, opened 2026-09-14
feat(lint): add ErrorStringMatching cop and fix call sites

dgageot:
Branching on `err.Error()` text is fragile: error messages can be reworded at any time, the check silently breaks when an error is wrapped, and it bypasses the type system entirely. The right approach is to expose sentinel errors and compare with `errors.Is`, or wrap structured types and unwrap with `errors.As`.

This adds the `Lint/ErrorStringMatching` cop, which reports any `strings.Contains`, `strings.HasPrefix`, or `strings.HasSuffix` call whose first argument is an `err.Error()` invocation. The cop is type-aware (it checks that the receiver actually implements the `error` interface), skips test files, and skips frozen versioned config packages where the schema is intentionally locked.

All in-module violations are fixed: `pkg/config/latest` now uses a dedicated `ErrUnsupportedModel` sentinel; `pkg/runtime`, `pkg/server`, `pkg/model/provider/dmr`, `pkg/desktop`, and `pkg/memory/database/sqlite` are updated similarly. The three remaining suppressions are legitimate third-party cases — an overlong Unix socket path check, a SQLite DDL migration string test, and Docker CLI unknown-flag detection — each annotated with an inline `//nolint` comment explaining why the text match is unavoidable.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34828619344) for details. Re-request a review from `docker-agent` to retry.

trungutt (approved):

PR #4269, opened 2026-09-14
feat(lint): add EnvironmentVariablePrefix cop and migrate pprof to DOCKER_AGENT_PPROF_ADDR

dgageot:
The codebase accumulated environment variable names under the legacy `CAGENT_` prefix. Now that the product ships as Docker Agent, new public variables should use `DOCKER_AGENT_` instead. Without an automated check, `CAGENT_*` names can silently creep back in.

This adds a new `Lint/EnvironmentVariablePrefix` cop to the custom linter. It walks AST string literals in non-test files and flags any `CAGENT_*` variable name that lacks a corresponding `DOCKER_AGENT_*` alias in the same file. Internal subprocess-protocol variables (`CAGENT_ASKPASS_SOCKET` and `CAGENT_ASKPASS_TOKEN`) are exempted because they are never user-facing. The cop is registered in `lint/main.go` alongside the existing cops.

To satisfy the new rule, `cmd/root/api.go` introduces `DOCKER_AGENT_PPROF_ADDR` as the canonical name for the pprof listen address, with `CAGENT_PPROF_ADDR` kept as a legacy fallback via `cmp.Or`. The two docs files that previously advertised `CAGENT_PPROF_ADDR` are updated to lead with `DOCKER_AGENT_PPROF_ADDR` and note the legacy alias for users upgrading from older deployments.


lint/environment_variable_prefix.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		literals := stringLiterals(p.File)
+		values := literalsByValue(literals)
+		for lit, value := range literals {
+			if !isLegacyEnvVar(value) || internalCagentEnvVars[value] {
+				continue
+			}
+			current := currentEnvPrefix + strings.TrimPrefix(value, legacyEnvPrefix)
+			if values[current] {
```
docker-agent, line 44: **[medium] Alias check is file-scoped — false positives when `DOCKER_AGENT_*` and `CAGENT_*` literals live in different files**

The cop builds its lookup map from a single `*ast.File` (`p.File`) and checks `values[current]` against only that file's string literals. If a package spreads the two sibling names across files — for example, `"DOCKER_AGENT_PPROF_ADDR"` defined as a constant in `cmd/root/api.go` and `"CAGENT_PPROF_ADDR"` referenced in a separate helper — the cop will fire a false-positive offense on the file that contains the legacy name, even though the required alias is correctly declared elsewhere in the same package.

The inverse is also true: a file that contains both string literals purely in a comment or as test data passes the check even though neither is an actual env-var exposure.

Consider widening the scope to a `cop.Pass` that covers the entire package (if the framework supports it), or document that the cop intentionally requires both aliases to be co-located in the same source file.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

cmd/root/api.go
```diff
@@ -103,7 +108,7 @@ func (f *apiFlags) runAPICommand(cmd *cobra.Command, args []string) (commandErr
 		return errors.New("--pull-interval flag can only be used with OCI or URL references, not local files")
 	}
 
-	if pprofAddr := cmp.Or(f.pprofAddr, os.Getenv("CAGENT_PPROF_ADDR")); pprofAddr != "" {
+	if pprofAddr := cmp.Or(f.pprofAddr, os.Getenv(envPprofAddr), os.Getenv(cagentEnvPprofAddr)); pprofAddr != "" {
```
docker-agent, line 111: **[medium] `cmp.Or` treats `DOCKER_AGENT_PPROF_ADDR=""` as absent and silently falls back to `CAGENT_PPROF_ADDR`**

`cmp.Or` returns the first non-zero (non-empty) value among its arguments. This means that if a user explicitly sets `DOCKER_AGENT_PPROF_ADDR=""` — intending the new canonical variable to override or disable the legacy one — the empty string is silently skipped and `CAGENT_PPROF_ADDR` takes effect instead. In a containerised deployment that inherits a legacy `CAGENT_PPROF_ADDR` from a base image or secret store, this can unexpectedly start a pprof server on a port the operator believed they had disabled.

A more robust approach checks whether the new variable is present in the environment at all, regardless of its value:

```suggestion
	newAddr, newAddrSet := os.LookupEnv(envPprofAddr)
	pprofAddr := cmp.Or(f.pprofAddr, func() string {
		if newAddrSet {
			return newAddr
		}
		return os.Getenv(cagentEnvPprofAddr)
	}())
	if pprofAddr != "" {
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):

Two findings in the new code introduced by this PR: one in the lint cop's alias-scoping logic, one in the env-var fallback precedence.

trungutt (approved):

PR #4271, opened 2026-09-14
refactor(tools): establish StartableToolSet as single lifecycle owner

dgageot:
Previously, two separate mechanisms could mutate lifecycle state for a toolset: `StartableToolSet` managed its own `started` flag and streaks, while `RestartToolset` in the runtime dispatched directly to the underlying `Restartable` interface, bypassing the wrapper entirely. Code Mode maintained its own parallel state machine (`innerIdle`, `innerStarted`, `innerFailed`, `innerLost`) alongside the wrappers. This meant that an explicit restart could leave the wrapper's latch, failure streaks, and retry gate out of sync with the actual supervisor state.

This change makes `StartableToolSet` the single owner of lifecycle state for its wrapped toolset. It adds `CanRestart` and `RestartIfSupported` methods that perform the restart under the wrapper's lifecycle lock, keeping `started`, `recovering`, `attempted`, and all streaks consistent with the outcome. The runtime's `RestartToolset` now calls through the wrapper instead of reaching past it. Code Mode drops its hand-rolled state machine entirely and delegates each child's lifecycle to a dedicated `StartableToolSet` wrapper; `availableToolsets`, `IsStarted`, `Start`, and `Stop` all read through `TryIsAvailable`, `TryIsHealthy`, and the child wrapper's own `Start`/`Stop`. The test that previously verified "external recovery clears the backoff gate" is inverted: the wrapper must *not* adopt externally mutated inner state, since the wrapper is now the authoritative source of truth.

No behavior changes outside the lifecycle ownership boundary; the degraded-mode contract (partial vs. total start errors, `LostAfterStart` signaling) is preserved.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34830948376) for details. Re-request a review from `docker-agent` to retry.

gtardif (approved):

PR #4272, opened 2026-09-14
refactor(provider): centralize model provider interfaces in contracts package

dgageot:
The `Provider`, `EmbeddingProvider`, `BatchEmbeddingProvider`, and `RerankingProvider` interfaces were independently redeclared in three different packages (`provider`, `rulebased`, `vertexai`). Any drift between copies was a latent bug, and adding a method to the canonical contract meant hunting down every duplicate.

This change introduces `pkg/model/provider/contracts`, a lightweight package with no construction logic, that owns the single authoritative definition of all four interfaces. The `provider` package re-exports them as type aliases so all existing callers continue to compile unchanged. In `rulebased` and `vertexai`, the now-redundant local declarations are replaced with aliases to `contracts.Provider`.

No behavior changes. The `contracts` package sits below every provider implementation in the dependency graph, so there is no import cycle risk.


docker-agent (commented):
### Assessment: 🟢 APPROVE

gtardif (approved):

PR #4273, opened 2026-09-14
feat(lint): sync Toolset schema enum with DefaultToolsetCreators

dgageot:
The `agent-schema.json` Toolset `type` enum is the source of truth for editor completion and config validation, but it had drifted from the runtime registry: the `environment` toolset type was registered in `DefaultToolsetCreators` but absent from the enum, so schema-aware tools silently rejected configs that used it even though the runtime handled it fine.

`Lint/ToolsetSchemaRegistrySync` is a new rubocop-go cop that runs on `pkg/teamloader/toolsets/toolsets.go`. It parses the `map[string]teamloader.ToolsetCreator{...}` literal with the Go AST to extract every registered key, then reads `agent-schema.json` and collects every string value that appears in any `type.enum` array under `definitions.Toolset`. Both directions are checked: a key present in the registry but missing from the schema is flagged as undocumented, and a schema value with no corresponding registry entry is flagged as dead. The cop is registered in `lint/main.go` alongside the other structural checks.

The immediate fix is adding `"environment"` to the Toolset type enum in `agent-schema.json`. No runtime behaviour changes.


gtardif (dismissed):

docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34832466772) for details. Re-request a review from `docker-agent` to retry.

gtardif (dismissed):

gtardif (approved):

PR #4274, opened 2026-09-14
feat(lint): add StatePathViaPathsPackage cop and fix call sites

dgageot:
Hard-coded `.cagent` string literals scattered across production packages are a latent correctness problem for embedders. Anything that calls `filepath.Join(homeDir, ".cagent", ...)` directly bypasses `paths.SetRoot()`, so an embedder that relocates all docker-agent state (e.g. Gordon running inside Docker Sandboxes) silently leaks files into the host's `~/.cagent` instead of the configured root.

The new `Lint/StatePathViaPathsPackage` cop catches this at lint time. It scans every non-test `.go` file outside `pkg/paths` for string literals equal to `".cagent"` or prefixed with `".cagent/"` / `".cagent\"`, and reports them with a message pointing at `paths.GetDataDir()`, `GetCacheDir()`, or `GetConfigDir()`. The cop is excluded from its own source file with the standard `//rubocop:disable-file` directive.

The two clean call sites in `pkg/content/store.go` and `pkg/modelsdev/store.go` are updated to use `paths.GetCacheDir()` directly. The two remaining literals in `pkg/history/history.go` are intentional — one is the live history path that takes a caller-supplied root, and the other is a legacy migration path — and each carries a `//rubocop:disable` suppression with a brief explanation so the reason is on record.


gtardif (dismissed):

docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34832540474) for details. Re-request a review from `docker-agent` to retry.

gtardif (dismissed):

gtardif (dismissed):

trungutt (approved):

PR #4275, opened 2026-09-14
feat(lint): require atomicfile.Write for marshalled state

dgageot:
Writing marshalled bytes with `os.WriteFile` is a common source of state corruption: if the process crashes mid-write, the file is left truncated or partially overwritten. The project already has `atomicfile.Write` to address this, but nothing enforced its use.

This adds a new `AtomicStateWrite` cop to the custom linter. It detects any call to `os.WriteFile` where the data argument is the result of `json.Marshal`, `json.MarshalIndent`, or `yaml.Marshal` in the same function body, and requires `atomicfile.Write` instead. Four call sites were migrated: `pkg/content/store.go`, `pkg/modelsdev/store.go`, `pkg/skills/cache.go`, and `pkg/sandbox/kit/kit.go`. The one remaining `os.WriteFile` in `pkg/evaluation/save.go` writes evaluation output rather than application state, so it carries a `//nolint:atomicstatewrite` suppression.


gtardif (approved):

docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34832602799) for details. Re-request a review from `docker-agent` to retry.

PR #4276, opened 2026-09-14
feat(lint): add HTTPClientTransport cop to flag bare &http.Client{} without Transport

dgageot:
Without an explicit `Transport`, `&http.Client{}` falls back to `http.DefaultTransport`, which bypasses three things the codebase relies on: OTel trace correlation (injected by the tracing transport in `pkg/httpclient`), the Desktop-aware PAC proxy transport, and the SSRF guards in `pkg/httpclient`. Bare clients have quietly crept into call sites that should be using the shared helpers, making it easy to introduce a new one without noticing.

This adds a `Lint/HTTPClientTransport` cop (Warning severity) that flags any `&http.Client{}` composite literal that omits a `Transport` field, excluding `pkg/httpclient` (where the conventions are defined), `pkg/tools/mcp/oauthflow` (OAuth flows that configure their own transport stack), `pkg/fake` (test infrastructure), and `pkg/desktop` (raw Unix-socket dialing). Existing call sites that intentionally skip the shared transport — DMR, relay, runtime, selfupdate, federation, and hubauth clients — receive a `//rubocop:disable Lint/HTTPClientTransport` annotation with a brief explanation of why the default transport is appropriate there. The telemetry client is suppressed with a note that wrapping it in the OTel transport would create trace loops.

No behaviour is changed; this is purely a lint guardrail and annotation pass.


gtardif (dismissed):

docker-agent (commented):
### ⚠️ Review incomplete

gtardif (approved):

PR #4277, opened 2026-09-14
refactor(tools): unify toolset startup coordination

dgageot:
Previously, the agent and runtime each maintained their own ad-hoc logic for launching toolsets concurrently, classifying start outcomes, and deciding what to surface in the UI or log. This duplication meant bugs and improvements had to be applied in two places, and the outcome taxonomy lived implicitly inside each caller.

`StartToolSets` in `pkg/tools` now owns the full startup lifecycle: it fans out independent toolsets concurrently, waits for them before launching peer-dependent ones, applies a shared timeout, and returns a per-toolset channel so callers receive each result as soon as it is ready. `StartOutcome` and `StartOutcomeKind` provide a single vocabulary for ready, in-flight, timed-out, canceled, authorization-required, failed, and partial states. Both the agent turn path and the runtime initialization path now delegate entirely to this coordinator, discarding several dozen lines of duplicated scheduling and classification logic.

The new `start_coordinator_test.go` covers the coordinator's sequencing guarantees and outcome classification for the main outcome kinds.


docker-agent (commented):
### Assessment: 🟢 APPROVE

gtardif (approved):

PR #4278, opened 2026-09-14
refactor(provider): replace ProviderRegistry any with typed RebuildProviderFunc

dgageot:
The `base.Config.ProviderRegistry` field was typed as `any` to avoid a package cycle. Callers reconstructing a provider had to type-assert it to `*Registry`, which silently returned the base provider on any mismatch and made the cloning path hard to follow and test.

This change replaces that field with a `RebuildProviderFunc` closure that is injected by the factory at construction time. The closure captures the registry, environment, and model map (for routers), so `CloneWithOptions` can call it directly without knowing anything about the registry. The dead `Config.Models` field is removed; routers capture the map in the closure instead.

To break the `contracts → base` import cycle introduced when `Config` and the result types moved, all of that state is defined once in `contracts` and re-exported from `base` as type aliases, keeping all existing import sites unchanged.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34836769305) for details. Re-request a review from `docker-agent` to retry.

gtardif (approved):

PR #4279, opened 2026-09-14
fix(tui): scope asynchronous results and dialogs to their originating page

dgageot:
When a user submits a message or triggers a model refresh, the TUI may be in a different state by the time the response arrives — the tab may have been closed, cleared, restored, branched, or reloaded. Previously, results were routed to whatever page happened to be active at delivery time, causing responses, error toasts, and follow-on dialogs to appear in the wrong tab or on stale UI state.

This change gives each page a stable identity at submission time. Steer and follow-up results — success, failure, tool calls — are tagged with the originating page's identity and discarded if that identity no longer matches when the result arrives. The same guard applies to model-refresh dialogs: the result and any follow-on toast or confirmation dialog are scoped to the page that initiated the refresh, including the edge case where the tab was closed and its UI is still retained in memory. Commands snapshot the app and session state they need at dispatch time rather than closing over a live reference, so captures are consistent even if the session is replaced mid-flight. Page teardown cancels any in-flight submissions and makes a best-effort withdrawal of queue entries that succeeded locally but have not yet been acknowledged by `Update`.

A broad redesign of the general effect-routing architecture is intentionally out of scope. New tests cover delayed success and failure delivery, hidden-tab fallback execution, tab switches, close, clear, restore, branch, restore-back, same-page reload, attachment handling, cancellation, and second-hop dialog delivery.


rumpl (approved):

docker-agent (commented):
### ⚠️ Review incomplete

PR #4280, opened 2026-09-14
feat(provider): add request-time TokenSource authentication for OpenAI and Vertex AI

dgageot:
The OpenAI provider previously had no way to supply a short-lived bearer token at request time: callers either set a static API key, relied on the `token_key` env-var lookup, or went through the ChatGPT or gateway paths. Vertex AI Model Garden worked around this by wrapping the environment provider with a private `tokenEnv` type that cached a GCP access token and refreshed it on each `Get` call — a fragile indirection that mixed credential lifecycle management into the environment layer.

This change introduces `options.TokenSource`, a `func(context.Context) (string, error)` callback registered via `WithTokenSource`. The OpenAI client checks for it before `token_key`, using it to set the `Authorization` header on both HTTP and WebSocket requests. Vertex AI Model Garden drops `tokenEnv` entirely and passes an `oauth2.TokenSource`-backed callback directly through `WithTokenSource`, so GCP credentials are refreshed by the standard Go oauth2 machinery without any environment-layer indirection. Static API-key, ChatGPT, and gateway auth paths are unaffected. `FromModelOptions` is updated to round-trip the token source so provider cloning preserves it.

No breaking changes. The only callers of `newOpenAIClient` in Vertex AI are internal, and the new `newOpenAIClientWithTokenSource` entry point is unexported and tested directly.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly introduces a `TokenSource` callback mechanism for the OpenAI provider and refactors Vertex AI Model Garden to use standard oauth2 machinery. The design is sound and no high or medium severity bugs were found.

- [low] `pkg/model/provider/openai/client.go:85` — Construction-time token fetch is discarded, causing one extra token-source invocation per `NewClient` call (confidence: unverified/low)
  - The eager validation call to `tokenSource(ctx)` at construction time produces a token that is immediately discarded. For pure stateless sources this is harmless; for rate-limited or side-effectful sources it causes a spurious invocation per client construction. The test confirms this: token-1 is wasted at construction, token-2 and token-3 are the actual request tokens.
- [low] `pkg/model/provider/openai/auth.go:13` — `tokenAuthMiddleware` returns `nil` `*http.Response` on token fetch error (confidence: unverified/low)
  - Consistent with the existing `chatgptAuthMiddleware` pattern; safe if the SDK handles nil response + error correctly.
- [low] `pkg/model/provider/vertexai/modelgarden.go:200` — Token refresh failure now propagates as a request error instead of falling back to cached token (confidence: unverified/low)
  - The old `tokenEnv` cached and fell back to the last good token on refresh failure. The new code propagates the error directly. The `oauth2.ReuseTokenSource` is thread-safe and caches internally, so transient refresh failures will now surface as request errors rather than be silently absorbed. This is arguably more correct behavior.

aheritier (approved):

PR #4281, opened 2026-09-14
refactor(provider): centralize per-request gateway client setup

dgageot:
OpenAI, Anthropic, and Gemini providers each duplicated the same boilerplate for building a per-request gateway HTTP client: token refresh, URL construction, auth-retry transport wrapping, and SDK-specific credential injection. This redundancy made the logic harder to audit and meant any future change had to be applied in three separate places.

This change extracts that shared plumbing into `base.NewGatewayClient`, which every gateway-backed provider now calls. The function centralises token rotation, URL concatenation (including trailing-slash handling), error timing, the retry-on-401 transport, and SDK authentication — while keeping each call lazy so tokens are still fetched per request. Direct-auth paths and DMR are untouched.

New focused tests cover request headers and body passthrough, token rotation, error timing, URL edge cases, retry behaviour, and Gemini keepalive handling. All tests were validated against the original implementation before the refactor, and an independent review confirmed behaviour preservation.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34849297454) for details. Re-request a review from `docker-agent` to retry.

trungutt (approved):

PR #4282, opened 2026-09-14
feat(tools): request-scoped callbacks, safe MCP routing, and multi-subscriber event delivery

dgageot:
MCP toolset callbacks (elicitation, sampling, sampling-with-tools, and OAuth) were previously stored as global state on the toolset, meaning a callback registered by one in-flight request could be invoked or overwritten by another. This becomes a real problem under concurrent use — wrong callers receive dialogs, tokens, or results.

This change makes those callbacks request-scoped via a new `HandlerScope` type that is created per request and torn down when the request completes. The MCP session client now routes in-flight callbacks safely to the scope that registered them, so concurrent requests can no longer interfere. OAuth follows the same pattern, replacing the previous `SetOAuthCallback` / `GetOAuthCallback` accessors with scope-aware lookup.

The `tools-changed` and RAG events that toolsets emit now support multiple subscribers. The runtime subscribes to both, and tests cover fan-out delivery, ensuring no event is silently dropped when more than one consumer is registered.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34849302122) for details. Re-request a review from `docker-agent` to retry.

trungutt (approved):

PR #4283, opened 2026-09-14
refactor(tui): replace routed-timer contract with explicit UpdateEffects

dgageot:
The `discard-background-commands` / `TakeRoutedTimers` contract that PR #4279 introduced couples the TUI host to chat page internals and makes it hard to reason about when side effects are applied. This refactor replaces it with a small `UpdateEffects` value returned from `chat.Update` that classifies commands into `Local`, `Visible`, and `Global` buckets, and a follow-up fix that correctly preserves those effects for hidden tabs and page-owned globals.

The refactor commit (`eed822f3e`) keeps the `chat.Update` return type behaviorally identical while making the intent explicit: local commands run regardless of visibility, visible commands only run on the active tab, and global commands (e.g. exit-after-response) are guarded by tab, page, and app identity so they cannot fire for a stale origin. The fix commit (`d625a1251`) closes two related gaps: async work submitted from a hidden tab was being silently dropped, and page-owned global effects were not surviving tab switches.

Regression tests cover FIFO ordering with no duplicate dialogs, Bubble Tea command composition, background queue progression and cancellation, media stale-ID rejection, async origin routing, and retired global effects. All pass clean under `go test -race -shuffle=on`.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/docker-agent/actions/runs/34850085254) for details. Re-request a review from `docker-agent` to retry.

trungutt (approved):

PR #4285, opened 2026-09-14
fix: read token_key in the Anthropic and Gemini clients

### LINKED ISSUE #4284: Custom providers with `provider: anthropic` or `provider: google` ignore `token_key`
Custom providers typed `anthropic` or `google` cannot authenticate through `token_key`: those two clients never read it and only look at their native variables, while startup validation only asks for the `token_key` variable, so the failure surfaces at client construction. The documented OpenCode examples (`docs/providers/opencode-go/index.md:104-109`, `docs/providers/opencode-zen/index.md:145-150` and `:176-181`) and the `provider: anthropic` examples in `docs/providers/custom/index.md` (lines 52, 122, 216) all rely on it.

Reproduced on `main` (d1d596f) with the OpenCode Go example and only `OPENCODE_API_KEY` set:

```
$ docker-agent run opencode-anthropic.yaml --dry-run
Error: failed to get models: ANTHROPIC_API_KEY environment variable is required
```

The Gemini client fails the same way with `GOOGLE_API_KEY or GEMINI_API_KEY environment variable is required`. Workaround today: export the client's native variable with the same key.

Cause: `pkg/model/provider/anthropic/client.go:203-205` reads only `ANTHROPIC_API_KEY` and `pkg/model/provider/gemini/client.go:107-114` only `GEMINI_API_KEY`/`GOOGLE_API_KEY`; neither references `cfg.TokenKey`, although `defaults.go:188-189` copies the provider's `token_key` into the model config and the OpenAI, Bedrock and Model Garden clients honour it. `pkg/config/gather.go:219-221` marks only the `token_key` variable as required, so `docker agent` never prompts for the native one. Origin: #2343 made `token_key` the credential for every provider type at the config layer but did not update these two clients, and the OpenCode examples arrived in #3211, reviewed for URL shape only.

Fix: read `cfg.TokenKey` in the two direct paths, falling back to the native variables when it is unset, as the OpenAI client does. Found while working on #4190; unrelated to the session header.

IsmaelMartinez:
The Anthropic and Gemini clients now read the model's `token_key` before falling back to their native variables, the precedence the OpenAI client already uses, so the documented custom-provider configs (OpenCode Go and Zen, the `provider: anthropic` examples) authenticate as written instead of failing with `ANTHROPIC_API_KEY environment variable is required`. Verified with the binary against api.anthropic.com and opencode.ai; with both variables exported the `token_key` value now wins, as the docs already say.

Fixes #4284


aheritier (approved):
Approved. The direct Anthropic and Gemini clients now correctly honor configured `token_key` values while preserving their existing fallback and non-API-key authentication paths. CI is green.

aheritier (approved):

PR #4286, opened 2026-09-14
refactor(mcp): make GatewayToolset construction side-effect free

dgageot:
Previously, `NewGatewayToolset` created the secrets and config temp files immediately on construction, meaning any error writing files would surface before the toolset was ever started, and a crash or `SIGKILL` between construction and `Stop` would leave plaintext secrets on disk with no cleanup path tied to the subprocess lifecycle.

This change defers temp-file materialization to `Start` and `Restart`, so files are only written immediately before the subprocess is launched. If the subprocess fails to start, the files are cleaned up in the same call. `Stop` also removes them, making the full lifecycle symmetric. A mutex guards `prepare`/`cleanUp` so concurrent `Start`/`Restart`/`Stop` calls are safe. `NewGatewayToolset` itself becomes a pure, infallible constructor (aside from the config YAML marshal, which cannot be deferred).

The gateway tests are extended to cover the new lifecycle paths: successful start/stop, failed-start cleanup, and restart behavior.


docker-agent (commented):

> ⚠️ **Verification inconclusive**: The verifier returned mismatched finding IDs (a duplicate `finding_id` violated the one-to-one pairing requirement). The findings below are surfaced unverified from the drafter's analysis so the author can evaluate them.

- [medium] `pkg/tools/mcp/gateway.go`:115 — **Potential data race: `client.args` written in `prepare()` vs read by supervisor watcher goroutine** (confidence: ⚪ negligible 0/100 — unverified)

  `prepare()` assigns a new `[]string` slice to `client.args` while holding `GatewayToolset.mu`. The supervisor's background watcher goroutine reads `c.args` in `Initialize` → `exec.CommandContext(ctx, c.command, c.args...)` without holding that mutex. If a background reconnect is already in-flight when `prepare()` runs (e.g. triggered by an unexpected disconnect racing with a manual `Restart`), the concurrent read of `c.args` in the watcher's `Initialize` call could be a data race detectable by `go test -race`. The sequential call structure (`prepare()` → `Toolset.Restart` → `RestartAndWait`) suggests the write typically completes before the watcher reconnects, but the happens-before guarantee via the supervisor's internal channels may not be sufficient under Go's memory model for the write to `client.args` itself.

- [medium] `pkg/tools/mcp/gateway.go`:111 — **Dead `!ok` type-assertion branch is unreachable but masks a concrete-type dependency** (confidence: ⚪ negligible 0/100 — unverified)

  After writing temp files and assigning `t.secretsFile`/`t.configFile`, `prepare()` asserts `t.mcpClient.(*stdioMCPClient)`. The `!ok` branch calls `t.cleanUp(ctx)` and returns an error. This path is unreachable today since `NewGatewayToolset` always uses `NewToolsetCommand` which constructs a `*stdioMCPClient`. The drafter notes this is a theoretical concern only; `cleanUp` handles the error path correctly if it were ever reached.

trungutt (approved):

PR #4287, opened 2026-09-14
refactor(provider): compose capability-specific tracing wrappers

dgageot:
The tracing wrappers for embedding, batch-embedding, and reranking each followed the same pattern as the chat wrapper but were copy-pasted in full, duplicating five method bodies (`ID`, `BaseConfig`, `Unwrap`, `rebuilder`, and the capability guard) across three types.

This change extracts a shared `instrumentedBase` struct that embeds the common fields and implements those five methods once. Each capability wrapper now embeds `instrumentedBase` and only carries the logic specific to its own span: the `InstrumentX` constructor, the span helper call, and the interface assertion. The six capability-flag combinations produced by `InstrumentProvider` are preserved exactly, span helpers and the chat implementation are untouched, and the public API is unchanged.

Characterization tests covering all combinations were added alongside to lock in the behavior before and after the refactor.


docker-agent (commented):
### Assessment: 🔴 CRITICAL

One compile error found in the new test file.

pkg/model/provider/instrument_test.go
```diff
@@ -0,0 +1,317 @@
...
+			for _, outcome := range []string{"success", "nil non-streaming results", "empty non-streaming results", "error with result"} {
+				t.Run(outcome, func(t *testing.T) {
+					exporter.Reset()
+					ctx := genai.WithConversationID(t.Context(), "conversation")
+					ctx, parent := tp.Tracer("test").Start(ctx, "parent")
+					defer parent.End()
+					p := &instrumentTestProvider{
+						Config: base.Config{ModelConfig: latest.ModelConfig{
+							Provider: "openai", Model: "before-wrapping",
+							MaxTokens: new(int64(42)), Temperature: new(0.0), TopP: new(0.8),
```
docker-agent, line 130: **[high] `new(int64(42))`, `new(0.0)`, `new(0.8)` are invalid Go — compile error**

The built-in `new` function takes a **type**, not a value expression. Passing value expressions like `int64(42)`, `0.0`, or `0.8` causes the compiler to reject the file with an error such as `int64(42) is not a type`.

Because this is in the new test file introduced by this PR, the entire test package will fail to compile, blocking `task test` from running at all.

Fix: use a small helper or pointer-variable idiom:

```suggestion
						MaxTokens: ptr(int64(42)), Temperature: ptr(0.0), TopP: ptr(0.8),
```

*(where `ptr[T any](v T) *T { return &v }` is a one-line generic helper; alternatively assign each value to a local variable and take its address.)*

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

gtardif (approved):

PR #4290, opened 2026-09-14
fix(tui): serialize attention delivery on the owning tab

dgageot:
Attention dialogs were opened from subscription goroutines and emitted as
unscoped deferred commands, meaning they could race with tab switches, stack
up out-of-order, and survive the page lifetime they belonged to. This pair
of commits fixes that and cleans up the code around it.

The first commit is a pure extraction: all attention-dialog construction and
the `RouteScope` delivery logic are moved from `tui.go` into a new
`attention.go` file with a single `openAttentionDialogs` helper. No behavior
changes — just untangling the logic so the real fix is readable.

The second commit makes delivery intentional. Tab status and attention state
are now applied synchronously inside the root `Update` loop rather than from
the subscription goroutine. Dialogs open FIFO on the correct tab instead of
firing deferred unscoped opens. Background dialog instances are parked on
successful tab switches (preserving typed input), and prompts are retired
synchronously on close, reload, cancel, and the relevant stream boundaries,
while detached elicitations are kept alive. `RouteScope` rejects runtime
deliveries that belong to a retired page lifetime, and startup emissions are
rotated before replacement app startup so the scope ordering is correct.
Tests cover five prompt kinds across arrival/switch orderings, FIFO replay,
delayed init delivery, multiple typed dialogs alongside an unrelated modal,
close/clear/reload/restore/app-replacement lifecycles, visible and parked
root-child-cancel boundaries, unloaded tabs, startup scope ordering, and
standalone chat behavior.

Follows #4283. Deferred response-command and `StreamCancelledMsg` leaf
scoping are not addressed here and remain a follow-up; not all async TUI
commands are scoped, and not all stale producer events are rejected.


docker-agent (commented):
### ⚠️ Review incomplete

trungutt (approved):

aheritier (approved):
Approve — with follow-ups noted

The change is merge-ready. Required `ci / gate` is green; the commits are signed; and the implementation correctly serializes attention delivery through the owning tab/event loop while adding useful stale-page protection.

Two follow-ups are worth tracking, but neither blocks this PR:

1. `pkg/tui/service/supervisor/supervisor.go:140` — A runtime event buffered before `RetirePage` can be stamped with the replacement page’s scope when it is dequeued, potentially allowing a stale delivery through. This is not a regression from the base behavior, and the PR description explicitly defers complete stale-producer rejection; please track it separately.

2. `pkg/tui/service/supervisor/supervisor.go:148` — The comment says retirement invalidates deliveries already queued for the prior page. That overstates the current guarantee given the buffered-event case above; a wording adjustment would keep it accurate.

Optional: the CI `test-race` job was skipped, although local race testing passed.

PR #4291, opened 2026-09-14
fix(server): forward the agent config envelope in API-server mode

gtardif:
## Problem

The gordon proxy serves the agent config together with a DSSE attestation signing it, in the `encrypted_agent_config` field of the `/gordon-agent` response. The client is expected to forward that opaque envelope back in every gateway-bound request body so the gateway can verify the system prompt against the signed config.

With the feature fully deployed in production — artifact correctly signed, `serve-encrypted-prompt` on, `verify-encrypted-prompt-llm-judge` on — a conversation started from **Docker Desktop** produced traces with no prompt-verification attributes at all: no `prompt_verified`, no `agentConfigHash`, no `promptVerifyDuration`. The gateway was never given anything to verify.

The Desktop agent log shows the envelope arriving and then going nowhere:

```
DEBUG Captured encrypted agent config from Docker source YAML field  url=…/proxy/gordon-agent?gordonTag=v15-dev…
# "Forwarding encrypted agent config to Docker gateway" → 0 hits
```

## Cause

`teamloader.LoadWithConfig` discovers the envelope by **type-asserting** the source it is handed:

```go
// pkg/teamloader/teamloader.go:275
if runConfig.EncryptedConfig == "" {
    if ecs, ok := agentSource.(config.EncryptedConfigSource); ok { … }
}
```

In API-server mode (`docker agent serve api`, i.e. Docker Desktop) `NewSessionManager` wraps every source in a `sourceLoader` for caching and periodic refresh:

```go
// pkg/server/session_manager.go:157
loaders[name] = newSourceLoader(ctx, source, refreshInterval)
```

`sourceLoader` implemented only `Name`/`ParentDir`/`Read`, so the capability was **lost at the decorator**: the assertion failed, `runConfig.EncryptedConfig` stayed `""`, `WithEncryptedConfigBody` was never applied, and no envelope reached the gateway.

Nothing errored, because the gateway deliberately **fails open** on a request carrying no config — so verification silently never ran instead of failing loudly. The CLI path was unaffected: it passes the source to `teamloader` undecorated, which is why this only reproduced from Desktop.

`pkg/config/hcl` already forwards this exact capability for its own decorator (`hcl/source.go:60`) for the same reason; `sourceLoader` never got the same treatment.

## Fix

Forward `EncryptedConfig()` through to the inner source, mirroring the `hcl` passthrough.

It **reads through** rather than caching the value: the envelope is refreshed by the very same `Read` the existing `refreshLoop` performs, so an artifact repushed under a new signature is picked up without restarting the daemon. Sources without the capability (local files, bytes) return `""`, leaving request bodies untouched.

A compile-time `var _ config.EncryptedConfigSource = (*sourceLoader)(nil)` assertion pins it, since removing the method would otherwise silently stop forwarding again rather than fail the build — the failure mode that made this hard to spot.

## Tests

Four tests in `source_loader_test.go`, using the file's existing `mockSource` conventions:

| Test | Covers |
|---|---|
| `TestSourceLoaderForwardsEncryptedConfig` | satisfies `config.EncryptedConfigSource` (the interface, not just a same-named method) |
| `TestTeamloaderAdoptsEncryptedConfigThroughSourceLoader` | real `teamloader.LoadWithConfig` end-to-end — the assertion that actually failed |
| `TestSourceLoaderEncryptedConfigFollowsRefresh` | read-through, not stale-cached |
| `TestSourceLoaderEncryptedConfigAbsent` | source without an envelope → `""` |

Verified non-vacuous: reverting the fix turns both behavioural tests red (`teamloader must adopt the envelope through the sourceLoader decorator`).

## Verification

Clean on `golangci-lint` and `go run ./lint .`. `pkg/server`, `pkg/config/...` and `pkg/teamloader/toolsets` pass. `TestLoadExamples` fails on `dmr.yaml` / `unload_on_switch.yaml`, which reproduces on unmodified `main` (needs Docker Model Runner locally) and is unrelated.

Manual check once built into Desktop: `Forwarding encrypted agent config to Docker gateway` appears in `cagent.debug.log`, and proxy traces gain `agentConfigHash` + `promptVerifyDuration`.

## Note for reviewers

There are now two hand-written `EncryptedConfigSource` passthroughs (this one and `pkg/config/hcl`), with nothing obliging the next source decorator to remember. If a third appears, a shared wrapper or a note on the interface itself would be the structural fix — happy to follow up separately.


docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (changes requested):
The implementation is correct and well-tested. The required `ci / gate` is currently red because `test-windows` failed in `pkg/hooks` with command-hook timeouts. This appears unrelated to this PR: the same failure occurred on an unrelated `main` run, while `pkg/server` passed on Windows. Please rerun the Windows job; if it repeats, link or file the appropriate flaky-test issue.

One code-quality request: please trim the comment volume around `sourceLoader.EncryptedConfig()` (`pkg/server/source_loader.go:14-18, 62-79`). Keep the concise local rationale—this decorator must remain transparent because `teamloader` type-asserts the capability—but move the gateway/incident narrative to the PR description or commit message.

The data/envelope refresh-pairing observation and interface documentation are optional follow-ups, not merge blockers.

aheritier (approved):

PR #4297, opened 2026-09-14
fix(leantui): finalize interrupted tool calls

rumpl:
## Summary

- finalize in-flight lean TUI tool calls when a stream stops without a tool response
- render interrupted tools as errors before the cancellation marker
- add a regression test for pressing Esc during an `edit_file` call

## Validation

- `task build`
- `task test`
- `task lint`


docker-agent (commented):
### ⚠️ Review incomplete

aheritier (approved):

Issue #4309, opened 2026-09-16
thinking_budget: none is silently ignored on OpenAI-compatible endpoints, and provider_opts cannot pass the vendor off switch


IsmaelMartinez:
Open-weight reasoning models (Qwen3, DeepSeek, GLM) think by default, and every OpenAI-compatible server or vendor exposes a request field to turn it off: llama.cpp, vLLM, SGLang and mlx_lm accept `chat_template_kwargs: {"enable_thinking": false}`; llama.cpp, vLLM, SGLang, Ollama, Groq and Cerebras accept `reasoning_effort: "none"`; DeepSeek uses `thinking: {"type": "disabled"}`. docker agent can send none of them: `thinking_budget` is only acted on for OpenAI reasoning model names (`modelinfo.UsesReasoningEffort`), a disabled budget is normalised to nil in `applyModelDefaults` before any client sees it (which also leaves the documented DMR `reasoning-budget: 0` path dead), and `provider_opts` forwarding is limited to the sampling allowlist. The model therefore reasons on every call and `max_tokens` can be spent entirely on reasoning.

Measured with v1.139.0 on macOS against mlx_lm.server 0.31.3 serving mlx-community/Qwen3.6-35B-A3B-8bit:

```yaml
models:
  local:
    provider: openai
    model: mlx-community/Qwen3.6-35B-A3B-8bit
    base_url: http://localhost:8080/v1
    temperature: 0
```

`docker agent run --exec --json agent.yaml -` with a 7 kB prompt took 80–200 s per call and emitted 4.5k–13k `agent_choice_reasoning` events; the same prompt with `chat_template_kwargs` sent directly returns in about 3 s. `thinking_budget: none`, `provider_opts: {chat_template_kwargs: {enable_thinking: false}}` and a `/no_think` prefix all changed nothing (209–231 reasoning events on a three-word probe); a proxy confirms the body carries only `model`, `messages`, `stream`, `temperature`. With `max_tokens: 2048` the whole budget went to reasoning: zero `agent_choice` events, exit 0, a `warning` event, no `error`. `provider: dmr` with `runtime_flags: ["--reasoning-budget", "0"]` on qwen3.6:35b-a3b-q8_0 still produced 218 reasoning events (whether the flag reached llama.cpp is unconfirmed; a failed `_configure` is only logged at debug level).

Proposed:
[...]

PR #4310, opened 2026-09-16
feat(openai,dmr): switch thinking off on OpenAI-compatible endpoints and add provider_opts.extra_body

IsmaelMartinez:
Open-weight reasoning models served through OpenAI-compatible endpoints think by default and each server exposes a request field to turn it off, but the openai client only honoured `thinking_budget` for OpenAI reasoning model names, `applyModelDefaults` dropped a disabled budget before any client saw it, and no `provider_opts` beyond the sampling allowlist reached the wire, so a Qwen3 model on mlx_lm, llama.cpp, vLLM or SGLang reasoned on every call and could spend the whole `max_tokens` budget before answering. This adds `provider_opts.extra_body`, an object merged verbatim into the chat-completions body for the openai and dmr clients on any provider (an explicit key wins over anything derived), and makes `thinking_budget: none`/`0` on a user-supplied `base_url` send `chat_template_kwargs.enable_thinking=false` on Chat Completions with the existing 256-token output floor; the dmr client sends the same per request alongside `llamacpp.reasoning-budget`, which now actually reaches `_configure` for `none`. The switch is strictly opt-in: it needs the user's own `none`/`0` (resolved once in the factory as `options.WithChatTemplateThinkingOff`, so title-generation and compaction clones inherit it but cannot trigger it alone), compares the resolved `base_url` against the alias default so built-in aliases and their clones are unaffected, and is never sent for `azure`, `chatgpt` or OpenAI model names behind a proxy. Docs, `agent-schema.json` and `examples/thinking_budget.yaml` are updated, including a note that reasoning tokens count against `max_tokens`.

Refs #4309


lyonsno:
I reproduced a DMR side effect of preserving the disabled budget: MCP sampling clones now send `llamacpp.reasoning-budget: 0` to the model-wide `_configure` endpoint.

Using the real registry and `CloneWithOptions` with the options from `samplingModelOptionsFor`, an agent configured with budget 4096 produces:

```text
base 032d705: initial configure 4096; sampling configure omits budget
head 6e0a73b: initial configure 4096; sampling configure sets budget 0
```

Title/compaction clones skip that second configure call; sampling clones don't. Model Runner stores accepted configuration by model, so on the llama.cpp route this can leave the main agent using a zero reasoning budget after sampling. The original provider's config still says 4096, and its completion path doesn't reconfigure the runner.

Could sampling clones also avoid model-wide reconfiguration and use the new per-request thinking-off field? I verified the HTTP writes on both revisions; the downstream effect is traced through Model Runner's scheduler/loader, not a live-model run. The parent already reconfigures with the budget omitted—the newly introduced part is explicitly setting it to zero.

aheritier:

IsmaelMartinez:
Applied in `a38a765dc`. Model Runner replaces a model's stored configuration on every `_configure`, so the parent's budget-omitted reconfigure already wiped the agent's 4096 on main (`docker model configure show` → `Config: {}`, verified live on llama.cpp b9879); this PR changed the wiped value to 0. `dmr.NewClient` now skips `_configure` for `NoThinking` clones, matching title and compaction clones, and the sampling request keeps its per-request `chat_template_kwargs`. `WithNoThinking` is set only by the sampling and title clones, so nothing else changes. `TestNewClientClonesSkipConfigure` covers the main agent (configures 4096) and the three clones (no configure); through the real registry the scenario now produces one configure call instead of two.

aheritier:
Resolved by @dgageot in #4397

PR #4314, opened 2026-09-16
perf(tools/fetch): avoid HTML escaping in JSON results and trim fetch prompt overhead

dgageot:
`encoding/json` HTML-escapes characters like `<`, `>`, and `&` by default, which is harmless for JSON consumers but inflates token counts for no benefit when the encoded value is passed directly to an LLM. This change makes `ResultJSON` in `pkg/tools/tools.go` use a buffer encoder with `SetEscapeHTML(false)`, so tool results no longer carry unnecessary escape sequences. Existing callers get smaller payloads; the decoded data is identical.

The fetch toolset's tool descriptions also contained boilerplate that restated what the parameters do in ways an LLM already infers from the parameter names. Those descriptions are trimmed to the essential contract, saving roughly 39–40 tokens per invocation. Additionally, a new `escape_html` config field (default `false`) on the fetch tool lets callers opt back into the original encoding when byte-identical JSON output is required — for example, when piping results into a system that relies on the HTML-escaped form. The field is documented in `docs/tools/fetch/index.md`, reflected in `agent-schema.json`, and demonstrated in `examples/fetch_escape_html.yaml`. Regression tests cover both encoding paths and the trimmed descriptions.

The `escape_html` field is additive and the default preserves the existing behavior for all new configs; no existing agent YAML needs updating.


docker-agent (commented):
### Assessment: 🟢 APPROVE

gtardif (approved):

PR #4315, opened 2026-09-16
perf(urldetect): avoid O(n²) string conversion in URL prefix scan

dgageot:
`findURLSpans` builds `remaining := string(runes[i:])` on every scanned rune
position to test whether it starts with `"http://"` or `"https://"`. Because both
prefixes are pure ASCII, this re-encodes the entire remaining suffix at each step,
producing O(n²) allocations and bytes over a line with no URL — a 300-rune line
triggers 300 full-suffix re-encodings.

The fix replaces the `string(runes[i:])` + `strings.HasPrefix` pattern with a small
`hasRunePrefix` helper that compares the fixed-length ASCII prefix directly against
the rune slice — no allocation, no re-encoding. All other logic (word-boundary guard,
body scanning, trailing-punctuation stripping, paren balancing, display-column
accounting) is untouched.

Measured on a 7-entry batch covering plain ASCII, Unicode/CJK, emoji, invalid UTF-8,
and OSC 8 / ANSI-colour inputs (Apple M4 Max, `-benchmem -count=5`):

```text
                       before          after       delta
findURLSpansBatch      159 KiB/op      5.0 KiB/op  −97 %
                       799 allocs/op   22 allocs/op −97 %
findAllURLSpansBatch   161 KiB/op      8.6 KiB/op  −94 %
                       806 allocs/op   46 allocs/op −94 %
```

B/op and allocs/op are deterministic across runs; wall-clock figures are not quoted
because they vary with host load.

Behaviour equivalence is verified by `TestFindURLSpansRunePrefixEquivalence` and
`TestFindURLSpansRunePrefixEquivalenceRandom`, which compare the new implementation
against a verbatim copy of the old function using `reflect.DeepEqual` (so nil vs
empty-slice differences are also caught), over a hand-written corpus covering Unicode,
invalid UTF-8, surrogate bytes, boundary conditions, punctuation, and OSC 8 / ANSI
sequences, plus 20 000 seeded random strings. The existing `TestFindURLSpans`,
`TestURLAtPosition`, and related tests continue to pass unchanged.


docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (approved):

PR #4316, opened 2026-09-16
perf(tui): avoid discarded strings.Join in scrollview.compose

dgageot:
The scrollbar and reserved-space branches of `scrollview.compose` built a viewport-sized `contentView` string via `strings.Join(lines, "\n")` and then used `len(contentView)` only as a `strings.Builder.Grow` hint before discarding it entirely. The actual output was assembled by a separate loop. Only the default branch returned `contentView` directly, so the allocation was wasted for the two more common render paths.

This change eliminates the intermediate allocation by accumulating `contentLen` during the existing pad/truncate loop: each iteration adds `len(lines[i])` after the line is finalised, and `max(0, len(lines)-1)` is added afterward for the newline separators. Both `Grow` calls now use `contentLen`, and `strings.Join` is moved to the `default` branch where it's actually needed. Output is identical across all three code paths.

A regression test (`TestComposeNodiscardviewEquivalence`) checks all seven scenario combinations — scrollbar active, reserved-space only, and plain join — against an oracle copied from the pre-change implementation, covering the memoized-width path, the re-measure path, restyled lines, and empty input. A benchmark (`BenchmarkComposeNodiscardview`) exercises all scenarios to make allocation changes visible in `go test -bench -benchmem`. Note: full `task build` and `task lint` were blocked by unrelated environmental issues (Xcode license for CGO, Docker Model Runner not running for e2e fixtures); the scrollview package itself builds, vets, and lints cleanly under `CGO_ENABLED=0`.


docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (approved):

PR #4317, opened 2026-09-16
perf(filesystem): use strings.SplitSeq to avoid allocating line slice in search

dgageot:
`search_files_content` called `strings.Split(string(content), "\n")` for every scanned file, materialising the entire content as a `[]string` before testing a single line for matches. On a corpus of 40 files × 2000 lines, this allocates roughly 26.7 MiB per search call. `strings.SplitSeq` (Go 1.24) iterates lazily over the same newline-delimited segments and avoids that allocation, bringing the figure to ~24.2 MiB — a ~9.5% reduction verified with `go test -bench=. -benchmem`. Wall-clock timing was too noisy across runs to claim a latency improvement; the allocation reduction is the concrete, reproducible win.

The line-number counter moves from a 0-based range index (`lineNum+1` in the format string) to an explicit int that is pre-incremented before each line is processed (`lineNum` in the format string). The output is byte-for-byte identical: first line of a file still reports as line 1.

Two focused regression tests are added alongside the existing suite: `TestFilesystemTool_SearchFilesContent_LineNumbering` checks 1-based numbering for files with and without trailing newlines and for empty files; `TestFilesystemTool_SearchFilesContent_RegexEmptyLine` verifies that `^$` matches blank lines (including the trailing empty segment produced by a file ending in `\n`) with correct line numbers. `BenchmarkSearchFilesContent_Batch` is added so the allocation improvement can be reproduced locally.


docker-agent (commented):
### Assessment: 🟢 APPROVE

docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (approved):

PR #4318, opened 2026-09-16
perf: reduce attachment data URI allocations

dgageot:
Building a data URI with `fmt.Sprintf("data:%s;base64,%s", mimeType, b64)` creates an intermediate string for each format argument before joining them. For small payloads that cost is negligible, but for the megabyte-range attachments that providers send and receive it shows up as measurable heap pressure — the extra copy lives until the next GC.

This change replaces three `Sprintf` calls in the `oaistream` `DataURI` helper and in the `openai` Responses image and PDF branches with plain string concatenation. The resulting string is constructed in one allocation instead of several, and the surrounding logic is untouched: error paths, the payload output branch, and the evaluation branch are all unchanged.

## Tests

New table-driven tests cover the exact URI format for nil data, empty data, base64-padded values, and case/literal MIME preservation. Benchmarks run against a 1 MiB payload and confirm the reduction; the numbers below were collected on Go 1.27 darwin/arm64 (Apple M4 Max) with `-p 1 -cpu=1 -benchtime=200ms -count=6` and compared with `benchstat`:

```text
                                                │   before    │         after          │
                                                │    B/op     │    B/op       vs base  │
DataURI/1MiB                                    │  5.344Mi    │  4.008Mi   -25.00%     │
ConvertDocumentResponseInputB64/image/png       │  5.344Mi    │  4.008Mi   -25.00%     │
ConvertDocumentResponseInputB64/application/pdf │  5.344Mi    │  4.008Mi   -25.00%     │

                                                │ allocs/op   │ allocs/op  vs base     │
DataURI/1MiB                                    │    11       │     3      -72.73%     │
ConvertDocumentResponseInputB64/image/png       │    14       │     5      -64.29%     │
ConvertDocumentResponseInputB64/application/pdf │    14       │     5      -64.29%     │
```

Reproducible with:

```bash
go test -p 1 -cpu=1 ./pkg/model/provider/oaistream ./pkg/model/provider/openai \
  -run '^$' -bench 'Benchmark(DataURI|ConvertDocumentResponseInputB64)$' \
  -benchmem -benchtime=200ms -count=6
```

Validated with `go test -count=1 -race` on the affected packages and a `CGO_ENABLED=0 GOFLAGS=-tags=no_audio` build and lint run (0 issues). Native `task build`, `task lint`, and the full `task test` suite cannot complete in this environment because the Xcode licence is unaccepted and Docker Model Runner is unavailable; those are pre-existing environment constraints unrelated to this change.


docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (approved):

PR #4325, opened 2026-09-16
refactor: parse reference names and import suffixes with strings.CutLast

dgageot:
Go 1.27 added `strings.CutLast`, which returns the two halves around the last occurrence of a separator together with a found boolean — eliminating the common `LastIndex` / slice arithmetic pattern. Two places in this repository used that pattern, and this change replaces them.

In `pkg/config/references.go`, `externalRefBaseName` previously called `strings.LastIndex` three times to strip a digest, strip a tag (guarding against registry ports), and isolate the final path segment. Each call is replaced with `strings.CutLast`, using the found flag to replace the `>= 0` guard and the returned suffix to replace the manual offset arithmetic. The logic and semantics are identical.

In `lint/configpath.go`, `versionFromImport` used `strings.LastIndex` to locate `"pkg/config/"` and then sliced from `idx+len(marker)` onwards. `strings.CutLast` returns that suffix directly, removing both the offset calculation and the intermediate index variable.

Both files gain table-driven tests covering the cases relevant to each function: registry ports, tags, digests, empty and repeated separators, and the last-marker / missing-marker import suffix scenarios. No behaviour changes.


dgageot:
Validation:
- `task build` and `task lint` pass.
- Tests and `go vet` for `pkg/config` and `lint` pass. New table tests also passed against the original implementations before refactoring.
- `task test` fails only on the `dmr.yaml` and `unload_on_switch.yaml` subtests of `TestLoadExamples`: Docker Model Runner is installed but not running locally. Both failures also reproduce with the original `origin/main` implementations.
- With `.env.test` loaded, the suite passes excluding exactly those two subtests:
  ```sh
  go test ./... -skip '^TestLoadExamples/../../examples/(dmr|unload_on_switch)[.]yaml$'
  ```
  The first excluded run encountered a fake-process startup timeout in `TestRunDockerAgentInContainerCancelInterruptsDocker`; the rerun passed without code changes.

aheritier (approved):

PR #4327, opened 2026-09-17
chore(deps): bump openai-go/v3 to v3.61.0

dgageot:
The openai-go v3 SDK removed `ResponseStreamEventUnion.Name` between v3.54.0 and v3.61.0, which broke compilation. This bump upgrades the dependency and fixes the callsite.

The response stream handler now reads the optional top-level event name directly from `event.RawJSON()` instead of the removed field. Valid names take precedence over any per-item name; missing, null, empty, or non-string values fall back gracefully. A regression test covers the main cases including escaped names and buffered tool-argument flushing.

No behavioural change is expected for correctly-formed streams; the fallback path only activates for events where the SDK no longer exposes the field.


docker-agent (commented):
### Assessment: 🟢 APPROVE

gtardif (approved):

PR #4328, opened 2026-09-17
fix: preserve final TUI responses across deferred message transitions

dgageot:
When the TUI is scrolled up, incoming assistant text is buffered by a deferred tail component and flushed only when the user scrolls back to the bottom. The bug was that this buffer was never cleared when a *different* assistant message became the active owner — for example after a tool call, a reasoning block, a sub-agent return, or an agent switch. The old content would bleed into the new owner's output, visually truncating the final response even though the persisted session was correct.

The fix guards the buffer flush path: before accepting new content, it checks whether the owner has changed and, if so, flushes the stale buffer under the old owner first. It also ensures the pending image-load command is propagated correctly through that transition so no image attachment is silently dropped.

A regression test covers all four transition types (tool call, reasoning, agent return, agent switch) across three completion paths (normal stop, cancel, scroll-to-bottom), plus image command propagation. The full test suite passes; the two failures in `pkg/teamloader` (`TestLoadExamples` for `unload_on_switch.yaml` and `dmr.yaml`) are pre-existing and require Docker Model Runner to be running — they are unrelated to this change.


docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (approved):

trungutt (approved):

PR #4341, opened 2026-09-17
fix: report per-call costs in telemetry

dgageot:
Telemetry received `sess.TotalCost` *before* the current call was charged, and the accumulator sums all values it receives, so four \$2 calls emitted `[0, 2, 4, 6]` and reported \$12 instead of the true \$8.

The fix changes the telemetry call site to emit exactly one per-call cost, computed after the effective model is resolved (including fallback) and priced using `computeMessageCost` — the same helper already shared with budgets, hooks, and message accounting. Nil usage is silently skipped; models with no pricing entry emit 0. The obsolete model and cost plumbing that was threading through the stream and fallback paths is removed now that it is no longer needed there.

Tests cover repeated calls on top of existing spend, all cached-token buckets, and a range of edge cases: EOF with no usage, reasoning-only responses, free and unpriced models, missing usage, fallback models, and custom per-agent pricing. The full suite passes with the exception of two `TestLoadExamples` subtests (`dmr.yaml` and `unload_on_switch.yaml`) that require Docker Model Runner to be running; the same two subtests fail against base `HEAD` under identical conditions, confirming they are pre-existing environment failures unrelated to this change.


docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (dismissed):

aheritier (approved):

PR #4342, opened 2026-09-17
perf(tui): defer frame composition while blurred

dgageot:
## Summary

- Reuse the last composed frame while blurred, without dropping runtime events or commands. Keep window-title and progress metadata current on a copy of the cached view.
- Compose the accumulated changes on focus. Startup, resize, fatal errors, and the final lean-mode frame still render when necessary.
- Add regression coverage for streamed chunks, tool completion, startup, resize, errors, and exit. A real Bubble Tea test verifies that 101 chunks produce zero frame compositions and terminal writes while blurred.

## Dependency and scope

Draft pending the separate focus-reporting/initial detached-session detection change (recommendation **1** from the CPU investigation). This PR implements recommendation **2**; it does not enable `ReportFocus` itself. Without that prerequisite, ordinary terminal runs do not receive the focus events needed to activate this optimization. Initial detached-session visibility will also need to be wired into rendering deferral when integrating the two changes.

The current policy defers on `BlurMsg`, which can also describe a **visible but unfocused split pane**. That UX tradeoff needs confirmation before marking ready. This skips root composition and downstream terminal rendering; it does not eliminate all rendering work performed during event processing.

## Validation

- `CI=true GOMAXPROCS=4 GOFLAGS=-p=2 task build`
- `GOMAXPROCS=4 GOFLAGS=-p=2 task lint`
- `MODEL_RUNNER_HOST=http://127.0.0.1:1 GOMAXPROCS=4 GOFLAGS=-p=2 task test`
- `GOMAXPROCS=4 go test -p 2 -race ./pkg/tui -run 'TestRootBlur|TestActualProgramBlur|TestBlurFocus|TestActualProgramScrollCancel' -count=3`

Plain `task test` failed only in the existing `TestLoadExamples` cases for `dmr.yaml` and `unload_on_switch.yaml`: Docker Model Runner was not running. The explicit host above bypasses CLI discovery/auto-pulls for that run; no services or live agents were restarted. `CI=true` prevents the build script from changing the user's binary symlink.


gtardif (dismissed):

aheritier:

trungutt (approved):

PR #4346, opened 2026-09-17
fix(openai): lease WebSocket connections exclusively

dgageot:
The OpenAI WebSocket pool allowed a single idle connection to be shared across concurrent streaming requests. Because each WebSocket carries exactly one in-flight exchange at a time, two concurrent callers could interleave reads on the same connection, silently corrupting both responses.

A new `Lint/ExclusiveStreamLease` analyzer was written first and run against the codebase to confirm the problem. It uses resolved-type, intra-procedural CFG and origin tracking via the existing `prog.Cop` framework to flag WebSocket wrapper values that alias a retained receiver field and are returned to callers — covering branch paths and named-return cases. The analyzer flagged exactly two violations at `ws_pool.go:103` and `ws_pool.go:127`. It is intra-procedural only and does not prove whole-program concurrency safety; helper side-effects, receiver aliases in callers, container tracking, and general lease protocols are out of scope.

The fix gives the idle slot take-and-clear semantics: consuming it transfers exclusive connection ownership to the caller for the lifetime of its streaming request. A second concurrent request dials its own connection rather than blocking or sharing. Active connections are tracked so pool shutdown closes both idle and in-flight connections. `Close` is idempotent; connections that have not finished are closed to unblock the reader; a request already cancelled at entry leaves the idle slot unchanged.

`task build` and `task lint` pass. The targeted race tests pass with `-race -count=3`. `task test` fails only on `TestLoadExamples` entries that require Docker Model Runner to be running locally; this failure reproduces identically on the pre-fix commit and is unrelated to this change.

The rest of the suite passes with `go test ./... -skip '^TestLoadExamples$'`. Pool-wide `lastResponseID` semantics and holding the pool mutex during dialing predate this PR and remain out of scope.


gtardif (dismissed):

aheritier:

trungutt (approved):

PR #4350, opened 2026-09-17
feat: support native Gemini text embeddings

dgageot:
Google's genai SDK exposes a dedicated embeddings API that is meaningfully different from the generic chat path, and wiring it up properly enables RAG use-cases on Gemini without routing through a chat model or falling back to zero-vectors. This PR adds that native embedding support to the Gemini provider.

The implementation lives in `pkg/model/provider/gemini/embed.go` and handles both the direct (`batchEmbedContents`) and gateway (`predict`/`Embedding2`) transport paths. Single and batch embedding share the same entry point; responses are reordered to match input order so callers never see shuffled results. Input validation rejects empty slices, cardinality mismatches, and unexpected dimension counts early, and the existing retry/error plumbing from the rest of the Gemini provider is reused unchanged. `output_dimensionality` is exposed as an optional config field in the schema and the `latest` config package. A worked RAG example (`examples/rag/gemini_embeddings.yaml`) and updated provider and RAG docs show how to wire it together.

Embeddings are only invoked when the configured backend supplies them; nothing changes for chat, multimodal, default model selection, or task-type prefixes. Switching embedding model or dimensions requires a full reindex — this is called out in the docs. No live paid API calls are made in tests; coverage is via targeted unit and race tests against the Gemini and RAG packages.


pkg/model/provider/gemini/embed.go
```diff
@@ -0,0 +1,153 @@
...
+func (c *Client) embedConfig() (*genai.EmbedContentConfig, error) {
+	config := &genai.EmbedContentConfig{}
+	if _, set := c.ModelConfig.ProviderOpts[outputDimensionalityOpt]; !set {
+		return config, nil
+	}
+	dims, ok := providerutil.GetProviderOptInt64(c.ModelConfig.ProviderOpts, outputDimensionalityOpt)
+	if !ok || dims <= 0 || dims > math.MaxInt32 {
+		return nil, fmt.Errorf("provider_opts.%s must be a positive integer, got %v", outputDimensionalityOpt, c.ModelConfig.ProviderOpts[outputDimensionalityOpt])
+	}
+	config.OutputDimensionality = new(int32(dims))
```
docker-agent, line 151: **[high] `new(int32(dims))` is a Go compilation error — `new` requires a type, not a value expression**

The `embedConfig()` function passes a type-conversion expression to `new`, which Go's type system rejects at compile time. `new(T)` requires `T` to be a type literal; `int32(dims)` is a value expression, not a type. The Go compiler confirms this: `int32(dims) is not a type`. This prevents the entire `gemini` package from building, so any attempt to use `output_dimensionality` in `provider_opts` will result in a build failure.

```suggestion
	v := int32(dims)
	config.OutputDimensionality = &v
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

docker-agent (commented):
### Assessment: 🟢 APPROVE

aheritier (approved):

PR #4353, opened 2026-09-17
fix: drain runtime streams before releasing turn ownership

dgageot:
A code audit identified that several call sites consumed a `RunStream` partially and then returned, leaving the stream abandoned. This matters because the server's streaming lock and the ACP turn token are held until `RunSession` and `runAgent` return — if the caller exits before the stream is fully drained, the lock or token is released while the producer goroutine may still be running, breaking the intended ordering guarantee.

The fix cancels the child context to signal the producer to stop, then drains and discards any remaining items from the stream before returning. This preserves the normal close-then-unlock sequence: the streaming lock in `session_manager.go` and the ACP turn token in `agent.go` are only released after the stream has actually closed. The per-turn CLI child context is kept alive until the stream is drained so that follow-up prompts remain possible during teardown. In the A2A adapter, an early consumer exit now also stops the detached producer goroutine rather than letting it run to completion unobserved.

`RunStream` ownership semantics are documented on the type itself. A new lint cop (`drain_run_stream_before_release`) detects `return`, `break`, and `goto` that abandon a locally consumed stream; it was deliberately run against the unfixed codebase first and found exactly seven violations across `pkg/server`, `pkg/acp`, `pkg/a2a`, `pkg/cli`, and `pkg/runtime`. After the fixes it reports zero. The cop uses real package types with a local `CFG`, handles comma-ok closure patterns, and supports deferred empty-range drains; it does not attempt to prove cancellation ordering or track cross-function transfers — the regression tests cover those behavioral properties instead.

Regression tests use `synctest` and channel-driven synchronization to give deterministic before/after coverage: they prove the pre-fix code fails (stream not drained, lock released early) and the fixed code passes, including tails that exceed channel capacity, handler errors, iterator exit, and CLI follow-up sequences. `golang.org/x/tools` is promoted from indirect to direct in `go.mod` to support the cop; no version was bumped.


trungutt (dismissed):

aheritier:

trungutt (approved):

PR #4354, opened 2026-09-17
feat(anthropic): native compaction, progress updates, strict tools, cache diagnostics, and secret redaction

dgageot:
The Anthropic provider has accumulated several independent capabilities that together make it a first-class participant in the agent lifecycle: native conversation compaction, opt-in thinking progress updates, strict tool argument enforcement, per-request cache diagnostics, and request-context secret redaction.

**Native compaction** is the central piece. When `provider_opts.native_compaction: true` is set, the runtime calls `CompactConversation` on the conversation model instead of routing through the generic LLM summarization path. Anthropic returns a signed `compaction` block that the SDK verifies on replay; the block is stored in a new `chat.CompactionResult` attached to the synthetic summary message. A strict prefix guard (`ReplayableCompaction`) prevents the raw block from being replayed if the summary text was edited or hook-redacted after compaction, keeping what the model sees consistent with what is stored. The runtime writes the block to disk before it replaces the in-memory history so a crash between the two is recoverable; post-compaction hooks fire after the commit.

The remaining features are independent opt-ins. `thinking_display: updates` streams thinking deltas as progress events on claude-3-7+ and claude-opus-4. `strict_tools` sets `tool_choice: {type: "auto", disable_parallel_tool_use: false, strict: true}` so the model cannot omit required fields. Cache diagnostics log hit/miss stats per request at debug level. Turn-scoped system, tool, and effort context is preserved across turns and through compaction via `CompactionResult.RequestContext`; the `redact_secrets` hook scrubs that payload in the same pass as message content so captured API keys or tokens are not persisted in plaintext.

Two known limitations: compaction is synchronous and whole-history (no background execution or keep-tail strategy), and the test suite uses HTTP/SSE fixtures rather than a live Anthropic API.


aheritier:

aheritier (dismissed):

aheritier:

gtardif (approved):

PR #4357, opened 2026-09-18
feat(wasm): shared runtime, portable tools, in-memory storage, egress proxy, and cloud provider builds

dgageot:
The WebAssembly build previously ran a bespoke event loop that reimplemented much of the agent runtime inline. Every new feature — RAG, plans, memory, multi-agent teams, MCP OAuth — had to be wired in twice: once for the desktop path and once inside `runtime_wasm.go`. This series replaces that loop with the same `embeddedchat`-backed runtime used by every other host, so wasm inherits all current and future capabilities without a second implementation.

The changes land in layers. `embeddedchat` now accepts nonlocal hosts and forwards interactive events (approval, elicitation, cancellation) across the JS/Go boundary via a new typed event bus. The JS bridge gains long-lived session objects, so multiple turns share conversation state, tool registries, and per-session stores. Per-session scoping is extended to OAuth token stores (`mcp`), environment discovery (`config`), todo/plan storage (new in-memory backend), and memory (`memory` package now supports SQLite-free backends). RAG gains an in-memory strategy store and a document-indexing path that never touches the filesystem, making all three retrieval strategies (BM25, semantic, chunked) available in the browser. Fuzzy deferred tool matching is extracted into `pkg/fuzzy` (MIT-licensed fzf algorithm, no CGo) so it compiles to wasm without the `fzf` binary dependency.

HTTP egress from the browser runs through a trusted proxy rather than the browser's own `fetch`; the proxy enforces strict SSRF policy (private-range blocking, fail-closed), omits cookies on forwarded requests, and refuses to follow redirects on behalf of the agent. Hooks that previously assumed a local filesystem (`add_environment_info`, `limit_large_tool_results`) are split into `_js` / `_other` build variants. An opt-in `docker_agent_wasm_cloud` build tag enables Bedrock and Vertex/Gemini provider builds, with host-supplied credentials injected at request time (no long-lived keys in the binary); the Bedrock bearer-auth scheme misconfiguration is also fixed here.

The series is tested at three levels: Go unit and race tests cover every new package including the in-memory stores, egress policy, and fzf parity against the upstream library; 17 Node.js bridge tests exercise the JS API end-to-end with a mock Go runtime; and 13 Chrome browser tests (`cmd/wasm/browser/smoke_test.js`) run the compiled wasm against real browser APIs via Playwright. CI is extended to run all three suites.

Existing embedders that relied on the old `InitSession`/`RunSession` API will need to migrate to the new session object model. The `runtime_wasm.go` monolith is removed; its surface is replaced by `runtime.go`, `session.go`, `handle.go`, `events.go`, and `toolsets.go`, each with focused responsibilities. Cloud provider support requires the `docker_agent_wasm_cloud` build tag and is off by default.


aheritier:

aheritier (dismissed):

aheritier (approved):

Issue #2741, opened 2026-05-11
Custom/aliased OpenAI-compatible providers: attachment caps may degrade to text-only


simonferquel-clanker:
## Problem

Providers like xAI, Mistral, Nebius, Ollama, and other custom OpenAI-compatible endpoints route through `pkg/model/provider/oaistream` with whatever `Provider` string is set in the model config (e.g. `"xai"`, `"mistral"`, `"ollama"`). The `modelcaps` system looks these up in models.dev, but the provider IDs in models.dev may not match the provider strings used in docker-agent configs.

For example:
- An xAI model config with `provider: xai, model: grok-2-vision-1212` produces a lookup for `"xai/grok-2-vision-1212"`
- An Ollama config with `provider: ollama, model: llava` produces `"ollama/llava"`

If models.dev doesn't have those exact keys, capability detection silently falls back to text-only and image/PDF attachments are dropped.

Additionally, user-configured aliases (via `display_model`) are already handled correctly by `ModelCapsID()` (which uses `Model`, not `DisplayOrModel()`), but the underlying provider identifier mapping to models.dev still depends on the config's `provider` field matching models.dev exactly.

## Possible fixes

- Maintain a mapping table from docker-agent provider strings to models.dev provider IDs
- Allow per-model capability overrides in the agent config YAML
- Warn (rather than silently drop) when a model ID is not found in models.dev, so users can diagnose the issue

## Context

Identified during review of PR #2738 (fix: pass fully-qualified provider/model ID to modelcaps.Load).

aheritier:
**Triage note:** This is the third issue in a `modelcaps` cluster opened today — all stemming from the same root cause (capability lookup relying on `models.dev` provider IDs that don't match docker-agent's provider strings):

| Issue | Provider affected |
|-------|------------------|
| #2737 ✅ (PR #2738) | Anthropic, OpenAI, DMR, Bedrock, Gemini — bare model name bug |
| #2739 | DMR local models — `models.dev` doesn't know DMR at all |
| #2740 | Vertex AI Model Garden — provider field mismatch |
| #2741 (this) | Custom/aliased OpenAI-compatible providers — provider string mismatch |

PR #2738 fixes the low-hanging fruit (bare name → qualified ID). The remaining three issues (#2739, #2740, this one) require a longer-term design: either a local capability override mechanism, a provider→models.dev mapping table, or a per-model config escape hatch. Marking `status/needs-design`.

PR #4365, opened 2026-09-19
fix(oaistream): replay ReasoningContent when converting messages

### LINKED ISSUE #4363: OpenAI custom provider does not send reasoning from previous turns
### Description

Reasoning is received and persisted correctly for OpenAI-compatible custom providers, but it is not included when assistant messages are converted back into the next Chat Completions request.

For models such as Qwen that expose reasoning via `reasoning` / `reasoning_content`, Docker Agent correctly:

- captures the streamed reasoning in `oaistream`
- accumulates it as `ReasoningContent`
- stores it on the assistant `chat.Message`
- displays it in the TUI

However, when the conversation history is converted for the next request, `oaistream.ConvertMessages()` serializes the assistant content and tool calls but does not serialize `msg.ReasoningContent`.

As a result, reasoning is visible and stored in the Docker Agent session, but is silently omitted from subsequent model requests.

This means a reasoning model cannot see its previous reasoning after a tool call. In long-running agent workflows this can cause the model to repeatedly reconstruct decisions and analysis it has already performed, substantially increasing reasoning tokens and execution time.

I confirmed this by patching the OpenAI-compatible message conversion to include the stored reasoning. An existing session that Docker Agent previously considered within a ~120K context window immediately produced a ~277K context request once the previously stored reasoning was actually replayed. After compaction/accounting was adjusted, the model retained reasoning correctly across tool turns and showed substantially better continuity.

This appears specific to the OpenAI-compatible/custom-provider conversion path; the reasoning state is already retained and replayed appropriately by other provider implementations.

### Expected Behavior

We should at least have an option to preserve reasoning for models using this provider. I appreciate this may need to be configurable rather than enabled by default, since OpenAI-compatible providers can behave differently.

### Actual Behavior

Reasoning is never preserved when using OpenAI custom provider.

### Steps to Reproduce

Use the custom AI provider and monitor context usage, it will only go up for output and not reasoning.

### Docker Agent version

1.141.0

### OS & terminal

Linux

### Model used

Qwen 3.8 flash next via llama.cpp

### Error output

```shell
No specific error, its subtle to spot
```

### Screenshots

_No response_

### Additional context

_No response_

20csm2k25-netizen:
Fixes #4363

## Root cause
`convertMessagesWithCaps` in `pkg/model/provider/oaistream/messages.go` built the outgoing assistant message param from `Content`, `FunctionCall`, and `ToolCalls` only. `msg.ReasoningContent` was never read, so reasoning captured from OpenAI-compatible custom providers (e.g. Qwen via llama.cpp) was silently dropped when the conversation history was replayed on the next request.

## Fix
`openai-go`'s typed `ChatCompletionAssistantMessageParam` has no field for `reasoning_content` — it's a non-standard, provider-specific extension (see [openai/openai-go#558](https://github.com/openai/openai-go/issues/558)). This PR attaches the stored reasoning via `SetExtraFields` when `msg.ReasoningContent` is non-empty, so it's a no-op for every model that doesn't produce reasoning.

## Open question for maintainers
The issue notes: *"We should at least have an option to preserve reasoning when defining models against this provider, or have it set by default."* This PR makes it default-on for the OpenAI-compatible path. Given the reporter's own finding that replaying reasoning inflated a request from ~120K to ~277K tokens, happy to add a per-model opt-out (e.g. via the existing `modelinfo.CapsOverride` mechanism from #2741) if that's preferred instead.

## Testing
Added `repro_issue4363_test.go` covering: reasoning is carried onto the replayed message, no field is added when reasoning is absent, and reasoning survives alongside tool calls. Full package test suite passes.


aheritier:

Commits that are not verified: `630ad45`

MysticalMount:
Hi - original issue reporter here. Thanks for picking this up. A couple of things stood out to me on this PR:

- `gofmt` hasn’t been run on the test file, which makes the diff a little difficult to read.
- In my local testing, simply restoring `ReasoningContent` on replay wasn’t sufficient. The token estimation used for compaction also needed to account for the restored reasoning tokens; otherwise the compaction threshold was substantially underestimated once reasoning started being replayed.

I hit this immediately with a long-running Qwen session: restoring the reasoning caused the effective context size to jump significantly, so without the corresponding accounting change the compaction behaviour became incorrect.

Happy to share the relevant part of my local patch if useful.

MysticalMount:
A clarification after investigating further: the compaction problem appears to be a transition edge case for existing sessions, rather than something this PR needs to address.

My session had accumulated stored reasoning while it was still omitted from requests. Enabling replay suddenly added that history to the next prompt, while provider-reported usage still reflected the smaller pre-fix requests. That caused the jump from roughly 71K to 277K tokens.

For new sessions where reasoning is replayed from the beginning, usage should progressively reflect that context, so the focused change here looks like the correct scope for #4363. The stale-usage and llama.cpp overflow-recovery cases can be handled separately.

PR #4371, opened 2026-09-21
chore: mark models.dev snapshot files as generated in .gitattributes

aheritier:

Marks `pkg/modelsdev/snapshot.json` and `pkg/modelsdev/snapshot_date.txt` as generated in `.gitattributes` (`linguist-generated=true -diff`).

Both files are refreshed weekly by `.github/workflows/models-update.yml` (see e.g. #4369). `linguist-generated=true` collapses their diffs on GitHub; `-diff` does the same for local `git diff`/`git show`, which is fine since the refresh PR body already carries a rendered added/removed/changed delta of the catalog (`task models-delta`) rather than relying on a line-by-line diff of the 2.4MB JSON blob.

`merge=ours` was deliberately omitted: the workflow always opens a short-lived branch off `main` and deletes it after merge, so there's no long-lived branch to protect, and `merge=ours` on the target branch could cause a legitimate incoming refresh to be silently discarded.

No Go code touched — `task lint`/`task test`/`task build` not applicable.


.gitattributes
```diff
@@ -26,3 +26,8 @@
 *.zip binary
 *.tar binary
 *.gz binary
+
+# Bot-refreshed models.dev snapshot: regenerated weekly, reviewed via the
+# rendered delta in the PR body rather than a line-by-line diff.
```
aheritier: too much detailled

aheritier: 🤖 **Automated implementer agent** — *this comment was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

Fair point — shortened to a single line in cf87bb4: `# Bot-refreshed weekly by task update-models`.

trungutt (approved):

PR #4372, opened 2026-09-21
docs(attachment): state the real termination argument for the defuse loop

### LINKED ISSUE #4054: maxDefusePasses documents a termination argument that is false
### Description

`pkg/attachment` bounds its replace-until-stable loop with `maxDefusePasses = 8`, justified by
this comment:

```go
// maxDefusePasses bounds the replace-until-stable loop. Each pass strictly
// shortens the body (a match is always longer than nothing and is replaced by a
// constant), so this converges quickly; the bound only exists so a pathological
// input cannot spin.
const maxDefusePasses = 8
```

The premise is false. `delimiterPlaceholder` is 41 bytes; `</document-x>` is 13. Replacing the
latter with the former **grows** the body. Passes do not shorten it.

A second comment on `defuseDelimiters` is wrong in the opposite direction:

```go
// Replacement repeats until the output is stable, because one pass can leave a
// delimiter-shaped residue behind: `</TAG</TAG>>` collapses to `[…removed]>` only
// after the second pass.
```

`envelopeTagRe` runs to the **first** `>`, so `</TAG</TAG>>` is a single match and collapses in
one pass, not two.

**This is a documentation defect, not a behaviour defect.** The loop does terminate and no
delimiter escapes — that was verified before filing (see below). But the reasoning is what
someone will rely on when tuning the constant or changing the placeholder, and it is wrong.

### Expected Behavior

The comment states the argument that actually guarantees termination.

### Actual Behavior

It states an argument that is measurably false, and a worked example whose pass count is wrong.

### Steps to Reproduce

Measure a single pass over `</document-x>` with the envelope's own tag:

```
pass 1: len 13 -> 42   (GREW)
```

And over the residue shape `</document-x</document-x>>`:

```
pass 1: len 26 -> 43   (GREW)   — and it is already stable, i.e. one pass, not two
```

### Docker Agent version

_No response_

### OS & terminal

_No response_

### Model used

_No response_

### Error output

```shell
case 0 pass 1: len 13 -> 42  (GREW)
case 1 pass 1: len 26 -> 43  (GREW)
case 2 pass 1: len 117 -> 50  (shrank)


Only inputs with many delimiters shrink; the common single-delimiter case grows.
```

### Screenshots

_No response_

### Additional context

**No break-out is possible today.** Eighteen adversarial bodies — deep nesting, split brackets,
mixed case, trailing attributes, self-closing forms, whitespace padding, prefix-extending tags,
a body pre-seeded with the placeholder — were run through `defuseDelimiters`. Every one came back
with zero live delimiters, and every one converged in a single pass. The escaping is sound; this
issue is strictly about the stated reasoning.

The real termination argument is: `delimiterPlaceholder` contains neither `<` nor `>`, so it can
never form part of a new match. Every pass therefore strictly reduces the number of
delimiter-shaped tokens left in the body, and that count is a non-negative integer.

Why this is worth fixing rather than ignoring: the loop **returns whatever it has** if the bound
is ever hit — silently, with no error. So if someone later gives the placeholder angle brackets
(a plausible cosmetic change, e.g. `<removed>`), the real invariant breaks, the loop can exhaust
its bound, and a live delimiter is returned inside the envelope. The current comment gives no
hint that the placeholder's character set is load-bearing.

PerryLink:
﻿Fixes #4054.

`maxDefusePasses` justified its bound with an argument that is measurably false, and the issue is right about that. While correcting it I found the argument wrong in a third way the issue does not mention, so this records the real invariant instead of a patched-up version of the old one.

### What was wrong

1. **"Each pass strictly shortens the body"** — false. The placeholder is 42 bytes and a short delimiter like `</document-x>` is 13, so defusing *grows* the body. Measured: `13 -> 42`, and `26 -> 43`.
2. **"One pass can leave a delimiter-shaped residue behind"** — false. `envelopeTagRe` runs to the first `>`, so `</TAG</TAG>>` is a single match and any nested delimiter is consumed with it. One pass is always sufficient.
3. **The issue's own third data point doesn't reproduce.** It reports `117 -> 50 (shrank)` for many delimiters. I measured `</document-x>` × 9 (117 bytes) as `117 -> 378 (grew)`. Shrinking happens when a match span is *longer* than 42 bytes, not when there are many delimiters.

### What actually holds

Termination rests on one property of the replacement text: `delimiterPlaceholder` contains no `<`. A replacement can therefore never introduce a delimiter start, while every match consumes at least one `<`. Each changing pass strictly reduces the body's `<` count, which is a non-negative integer — so the loop cannot run forever.

I verified "one pass is always enough" rather than asserting it: brute-forced 846,383 bodies (all `{<,>}` strings up to length 14, plus token-alphabet enumeration to depth 5). With `maxPasses=1`, zero delimiters survived and the `<` count never increased.

The placeholder's character set is worth recording as load-bearing rather than cosmetic: the loop returns whatever it has once the bound is exhausted, silently and with no error, so a placeholder that gained an angle bracket (say `<removed>`) could let a live delimiter through inside the envelope.

### Also

`envelope_test.go:95`'s comment repeated the same false premise, so it is corrected too. The assertion itself is correct and unchanged — a nested delimiter must genuinely leave no residue, whether it is neutralised in the matching pass or a later one.

### Scope

Comment-only, 2 files, +23/−9. Verified no code line changes: every added/removed line is a `//` comment.

```
go test -count=1 ./pkg/attachment/   ok
gofmt -l pkg/attachment/             clean
go vet ./pkg/attachment/             exit 0
go run ./lint ./pkg/attachment       no offenses
```

`task test` (`go test ./...`) has unrelated failures in `pkg/rag/treesitter` (needs `CGO_ENABLED=1`) and `pkg/workspacemedia` (Windows symlink privilege); both reproduce with my changes stashed.


aheritier:

Commits that are not verified: `386a0b5`

